PR #42, opened 2020-03-05
Fix instructions

gmuraru:
Currently, the instructions from the README redirects you to http://www.unikraft.org/getting-started.md which is not valid. It should be without the ```.md``` suffix.


felipehuici:
Resubmitted via mini-os (Unikraft) mailing list.

PR #65, opened 2020-09-28
Introduce POSIX socket driver microlibrary

### LINKED ISSUE #11: Support for UNIX domain sockets

### LINKED ISSUE #97: Socket Support

nderjung:
This patch series introduces the posix-socket microlibrary within the Unikraft core in order to facilitate the use of multiple POSIX-compliant socket implementations which target specific AF family numbers.

This library allows for the registration of socket interfaces, listed within `struct posix_socket_ops`, and exposes unikernel-wide  prototypes for `socket()`, `accept()`, `bind()`, `listen()`, `connect()`, `send()`, `recv()` and friends in order to access them via a  corresponding AF family number.  The implementing library simply needs to register against the desired interfaces with a glue function:

```C
#include <uk/socket.h>
#include "my_socket.h"

static struct posix_socket_ops mysock_socket_ops = {
  /* The initialization function on socket registration. */
  .init        = mysock_lib_init,
  /* POSIX interfaces */
  .create      = mysock_glue_create,
  .accept      = mysock_glue_accept,
  .bind        = mysock_glue_bind,
  .shutdown    = mysock_glue_shutdown,
  .getpeername = mysock_glue_getpeername,
  .getsockname = mysock_glue_getsockname,
  .getsockopt  = mysock_glue_getsockopt,
  .setsockopt  = mysock_glue_setsockopt,
  .connect     = mysock_glue_connect,
  .listen      = mysock_glue_listen,
  .recvfrom    = mysock_glue_recvfrom,
  .recvmsg     = mysock_glue_recvmsg,
  .sendmsg     = mysock_glue_sendmsg,
  .sendto      = mysock_glue_sendto,
  /* vfscore ops */
  .read        = mysock_glue_read,
  .write       = mysock_glue_write,
  .close       = mysock_glue_close,
  .ioctl       = mysock_glue_ioctl,
};
```

Each glue function accepts traditional parameters for the POSIX interface as side from the file descriptor, traditionally `int fd` or `int sock`, which has instead been cast as `void *`. This allows the implementing library to use internal socket file descriptor or reference.  The creation of a socket via `socket()` and `accept()` have  also been cast to return `void *` in line with internal needs of the  implementing library.  This makes sense for implementing libraries which use a structure instead of a file descriptor identification integer.  An additional parameter, `struct posix_socket_driver`, is passed to each interface which can store private data as well as a preferred memory allocator.  The glue for each interface simply needs to call the appropriate method internally, for example:

```C
static int
mysock_glue_bind(struct posix_socket_driver *d,
        void *sock, const struct sockaddr *addr, socklen_t addr_len)
{
  struct my_socket *mysock;
  int ret;

  mysock = (struct my_socket *)sock;
  ret = mysock_bind(mysock, addr, addr_len);
  if (ret < 0)
    ret = -errno;

  return ret;
}
```

The library selects the appropriate implementation based on the AF family number and registers the required interfaces using a newly exposed macro:

```C
POSIX_SOCKET_FAMILY_REGISTER(AF_INET,  &mysock_socket_ops);
```

Implementing libraries can register as many AF numbers as desired or specify new families which are not listed in the `<sys/socket.h>` header.  To access the relevant implementation, the application simply needs to call the `socket()` method with the relevant AF family number.

This PR is now part of a larger group of dependent PRs which must be merged in the following order:
1. https://github.com/unikraft/lib-newlib/pull/23
2. https://github.com/unikraft/unikraft/pull/484
3. https://github.com/unikraft/unikraft/pull/65  (this PR)
4. https://github.com/unikraft/lib-lwip/pull/17
5. https://github.com/unikraft/unikraft/pull/485

> **Warning** Without having merged all of these PRs, network is no longer functional. All PRs should thus be checked and reviewed first and then merged in quick succession.


nderjung:
I need to remove the uk ring commits from this series.  I will force push an update.

nderjung:
This PR is ready for review :)

lib/posix-socket/driver.c
```diff
@@ -0,0 +1,154 @@
...
+		return 0;
+
+	return d->ops->init(d);
+}
+
+/* Returns the number of successfully initialized device sockets */
+static int
+posix_socket_family_lib_init(void)
+{
+	uk_pr_info("Initializing socket handlers...\n");
```
ungps: Just a nitpick: I would move this message after the variable declarations. Again, it's just a nitpick, in the worst case scenario it will cause a warning.

lib/posix-socket/socket_vnops.c
```diff
@@ -0,0 +1,323 @@
...
+
+	/* Put things together, and fill out necessary fields */
+	vfs_file->fd = vfs_fd;
+	vfs_file->f_flags = UK_FWRITE | UK_FREAD;
+	vfs_file->f_count = 1;
+	vfs_file->f_data = file;
+	vfs_file->f_dentry = s_dentry;
+	vfs_file->f_vfs_flags = UK_VFSCORE_NOPOS;
+
+	// s_vnode->v_data = file;
```
ungps: Just a nitpick, but I would drop this line.

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -91,15 +91,44 @@ struct posix_socket_driver {
...
 	posix_socket_driver_init_func_t   init;
+	/* vfscore ops */
+	posix_socket_close_func_t         close;
 };
 
+
+static inline int
+posix_socket_do_close(struct posix_socket_file *sock)
+{
+	UK_ASSERT(sock);
```
ungps, line 416: More of a question rather than a suggestion, but is there a reason to keep these `UK_ASSERT(sock);` in the `posix_socket_do_*()` functions? This point will not be reached because of the `if (unlikely(!sock))` in the `posix_socket_*()`.

ungps (commented):
Hello.

I looked two times over the commits during the past week and there wasn't much to say, I just left some comments (mostly nitpicks). It looks good, although there is a compilation error. I fetched the branch and tried to build it, but header `sys/socket.h` is missing.
```
In file included from /home/ubuntu/unikraft/lib/posix-socket/driver.c:38:
/home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49:10: fatal error: sys/socket.h: No such file or directory
   49 | #include <sys/socket.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

This header can be found in `lwip` library, but that is an external library. It doesn't make sense to me to have an internal library dependant on an external one. Even if we were to compile with `lwip`, there are another errors and warning (such as conflicting types for a couple of functions):
```
/home/ubuntu/unikraft/lib/posix-socket/driver.c: In function ‘posix_socket_get_family’:
/home/ubuntu/unikraft/lib/posix-socket/driver.c:149:29: warning: unused parameter ‘sock’ [-Wunused-parameter]
  149 | posix_socket_get_family(int sock)
      |                         ~~~~^~~~
  CC      libposix_socket: socket_vnops.o
In file included from /home/ubuntu/unikraft/lib/nolibc/include/fcntl.h:19,
                 from /home/ubuntu/unikraft/lib/vfscore/include/vfscore/fs.h:4,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:38:
/home/ubuntu/unikraft/lib/nolibc/arch/x86_64/bits/fcntl.h:6: warning: "O_NONBLOCK" redefined
    6 | #define O_NONBLOCK    04000
      | 
In file included from /home/ubuntu/libs/lwip/include/sys/socket.h:46,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:35:
/home/ubuntu/apps/app-httpreply/build/liblwip/origin/lwip-2.1.2/src/include/lwip/sockets.h:478: note: this is the location of the previous definition
  478 | #define O_NONBLOCK  1 /* nonblocking I/O */
      | 
In file included from /home/ubuntu/unikraft/lib/vfscore/include/vfscore/fs.h:4,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:38:
/home/ubuntu/unikraft/lib/nolibc/include/fcntl.h:40: warning: "O_RDONLY" redefined
   40 | #define O_RDONLY  00
      | 
In file included from /home/ubuntu/libs/lwip/include/sys/socket.h:46,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:35:
/home/ubuntu/apps/app-httpreply/build/liblwip/origin/lwip-2.1.2/src/include/lwip/sockets.h:484: note: this is the location of the previous definition
  484 | #define O_RDONLY    2
      | 
In file included from /home/ubuntu/unikraft/lib/vfscore/include/vfscore/fs.h:4,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:38:
/home/ubuntu/unikraft/lib/nolibc/include/fcntl.h:41: warning: "O_WRONLY" redefined
   41 | #define O_WRONLY  01
      | 
In file included from /home/ubuntu/libs/lwip/include/sys/socket.h:46,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:35:
/home/ubuntu/apps/app-httpreply/build/liblwip/origin/lwip-2.1.2/src/include/lwip/sockets.h:487: note: this is the location of the previous definition
  487 | #define O_WRONLY    4
      | 
In file included from /home/ubuntu/unikraft/lib/vfscore/include/vfscore/fs.h:4,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:38:
/home/ubuntu/unikraft/lib/nolibc/include/fcntl.h:42: warning: "O_RDWR" redefined
   42 | #define O_RDWR    02
      | 
In file included from /home/ubuntu/libs/lwip/include/sys/socket.h:46,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:35:
/home/ubuntu/apps/app-httpreply/build/liblwip/origin/lwip-2.1.2/src/include/lwip/sockets.h:490: note: this is the location of the previous definition
  490 | #define O_RDWR      (O_RDONLY|O_WRONLY)
      | 
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:45:31: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct vnode *, struct vattr *)’ [-Wcast-function-type]
   45 | #define posix_socket_getattr ((vnop_getattr_t) vfscore_vop_einval)
      |                               ^
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:54:17: note: in expansion of macro ‘posix_socket_getattr’
   54 |  .vop_getattr = posix_socket_getattr,
      |                 ^~~~~~~~~~~~~~~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:46:32: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct vnode *)’ [-Wcast-function-type]
   46 | #define posix_socket_inactive ((vnop_inactive_t) vfscore_vop_nullop)
      |                                ^
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:55:18: note: in expansion of macro ‘posix_socket_inactive’
   55 |  .vop_inactive = posix_socket_inactive,
      |                  ^~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:58:28: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct mount *, struct vnode *)’ [-Wcast-function-type]
   58 | #define posix_socket_vget ((vfsop_vget_t) vfscore_nullop)
      |                            ^
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:62:14: note: in expansion of macro ‘posix_socket_vget’
   62 |  .vfs_vget = posix_socket_vget,
      |              ^~~~~~~~~~~~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c: In function ‘posix_socket_vfscore_close’:
/home/ubuntu/unikraft/lib/posix-socket/socket_vnops.c:205:24: warning: unused parameter ‘vfscore_file’ [-Wunused-parameter]
  205 |   struct vfscore_file *vfscore_file)
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
  CC      libposix_socket: socket.o
In file included from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:43:2: error: unknown type name ‘uint32_t’
   43 |  uint32_t type;
      |  ^~~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c: In function ‘getnameinfo’:
/home/ubuntu/unikraft/lib/posix-socket/socket.c:378:45: warning: unused parameter ‘sa’ [-Wunused-parameter]
  378 | getnameinfo(const struct sockaddr *restrict sa, socklen_t sl,
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:378:59: warning: unused parameter ‘sl’ [-Wunused-parameter]
  378 | getnameinfo(const struct sockaddr *restrict sa, socklen_t sl,
      |                                                 ~~~~~~~~~~^~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:379:18: warning: unused parameter ‘node’ [-Wunused-parameter]
  379 |   char *restrict node, socklen_t nodelen, char *restrict serv,
      |   ~~~~~~~~~~~~~~~^~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:379:34: warning: unused parameter ‘nodelen’ [-Wunused-parameter]
  379 |   char *restrict node, socklen_t nodelen, char *restrict serv,
      |                        ~~~~~~~~~~^~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:379:58: warning: unused parameter ‘serv’ [-Wunused-parameter]
  379 |   char *restrict node, socklen_t nodelen, char *restrict serv,
      |                                           ~~~~~~~~~~~~~~~^~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:380:13: warning: unused parameter ‘servlen’ [-Wunused-parameter]
  380 |   socklen_t servlen, int flags)
      |   ~~~~~~~~~~^~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:380:26: warning: unused parameter ‘flags’ [-Wunused-parameter]
  380 |   socklen_t servlen, int flags)
      |                      ~~~~^~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c: At top level:
/home/ubuntu/unikraft/lib/posix-socket/socket.c:467:1: error: conflicting types for ‘recv’
  467 | recv(int sock, void *buf, size_t len, int flags)
      | ^~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:95:5: note: previous declaration of ‘recv’ was here
   95 | int recv(int s, void *mem, size_t len, int flags);
      |     ^~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:505:1: error: conflicting types for ‘recvfrom’
  505 | recvfrom(int sock, void *buf, size_t len, int flags,
      | ^~~~~~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:96:5: note: previous declaration of ‘recvfrom’ was here
   96 | int recvfrom(int s, void *mem, size_t len, int flags,
      |     ^~~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:544:1: error: conflicting types for ‘recvmsg’
  544 | recvmsg(int sock, struct msghdr *msg, int flags)
      | ^~~~~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:98:5: note: previous declaration of ‘recvmsg’ was here
   98 | int recvmsg(int s, struct msghdr *msg, int flags);
      |     ^~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:582:1: error: conflicting types for ‘send’
  582 | send(int sock, const void *buf, size_t len, int flags)
      | ^~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:99:5: note: previous declaration of ‘send’ was here
   99 | int send(int s, const void *dataptr, size_t size, int flags);
      |     ^~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:620:1: error: conflicting types for ‘sendmsg’
  620 | sendmsg(int sock, const struct msghdr *msg, int flags)
      | ^~~~~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:100:5: note: previous declaration of ‘sendmsg’ was here
  100 | int sendmsg(int s, const struct msghdr *message, int flags);
      |     ^~~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:658:1: error: conflicting types for ‘sendto’
  658 | sendto(int sock, const void *buf, size_t len, int flags,
      | ^~~~~~
In file included from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket_driver.h:49,
                 from /home/ubuntu/unikraft/lib/posix-socket/include/uk/socket.h:56,
                 from /home/ubuntu/unikraft/lib/posix-socket/socket.c:35:
/home/ubuntu/libs/lwip/include/sys/socket.h:101:5: note: previous declaration of ‘sendto’ was here
  101 | int sendto(int s, const void *dataptr, size_t size, int flags,
      |     ^~~~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c: In function ‘socketpair’:
/home/ubuntu/unikraft/lib/posix-socket/socket.c:741:1: warning: label ‘EXIT’ defined but not used [-Wunused-label]
  741 | EXIT:
      | ^~~~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:704:33: warning: unused variable ‘u2’ [-Wunused-variable]
  704 |  struct posix_socket_file *u1, *u2;
      |                                 ^~
/home/ubuntu/unikraft/lib/posix-socket/socket.c:704:28: warning: unused variable ‘u1’ [-Wunused-variable]
  704 |  struct posix_socket_file *u1, *u2;
      |                            ^~
make[3]: *** [/home/ubuntu/unikraft/support/build/Makefile.build:27: /home/ubuntu/apps/app-httpreply/build/libposix_socket/socket.o] Error 1
make[2]: *** [Makefile:990: sub-make] Error 2
make[1]: *** [Makefile:32: _all] Error 2
make[1]: Leaving directory '/home/ubuntu/unikraft'
make: *** [Makefile:6: all] Error 2
```

lib/posix-socket/include/uk/socket_driver.h
marcrittinghaus: I would keep this `UK_ASSERT`. Asserts should be placed where the checked condition has to be true. So here it is directly before dereferencing `sock`. You can argue that in this specific case the caller already ensures that `sock` is valid. But I would respond that checking here does not hurt and is good design. The function works on its own and it is safe to call it from other functions without having to worry that necessary conditions are not properly checked for.

lib/posix-socket/socket_vnops.c
```diff
@@ -0,0 +1,323 @@
...
+			file->sock_data);
+
+	ret = posix_socket_read(file, buf->uio_iov->iov_base, buf->uio_iov->iov_len);
+	if (ret < 0) {
+		ret = errno;
+		goto EXIT;
+	}
+
+	buf->uio_resid -= ret;
+
```
marcrittinghaus, line 269: There is a bug here. The function should return an error code (or 0) and not the bytes read.

Add:
`ret = 0;`

nderjung:
Hey @marcrittinghaus, I have a few updates to this series out of this PR.  I will incorporate your suggestions (thanks btw!) when i force-push to this PR.

lib/posix-socket/socket_vnops.c
nderjung: This line should not be commented out, turns out it is very important and an artifact of previous debugging :)

marcrittinghaus:
I already did a ton of changes to this library internally. I think it makes sense to update it before making any reviews. It also needs some more clean up.

razvand:
@marcrittinghaus , maybe it makes sense that you update the PR contents (I think you can push to @nderjung 's brach (`nderjung/posix-socket`), and I will assign a reviewer and assignee to it. What do you think?

marcrittinghaus:
Hi @nderjung,

I have spent some hours yesterday to further clean up and fix some small things in posix socket. At the moment the patch series contains your original changes and then a lot of patches from my side. Are you ok with the solution that I discard the original commits and just create a few commits that provide the current version and have us two as authors?

nderjung:
Hi @marcrittinghaus, yes that's fine.  You should have the ability to force push to my branch for this PR.

marcrittinghaus:
@razvand, @nderjung,

I force pushed the new version. You may now assign a reviewer 😃

marcrittinghaus:
Updated the PR as discussed in the community meeting (https://github.com/unikraft/meeting-notes/blob/staging/discussions/community/2022/2022-03-24.md)

lib/posix-socket/driver_list.ld
```diff
@@ -0,0 +1,10 @@
+SECTIONS
+{
+	. = ALIGN(0x8);
```
craciunoiuc, line 3: Didn't this cause problems? I know there was a discussion a while ago and @skuenzer told me to drop `ALIGN`.

craciunoiuc: If I remember correctly, the bug was that it created broken images when certain flags were used. Still, this was half a year ago so everything is a bit foggy 😓

craciunoiuc: found it 🎉 #197

lib/posix-socket/include/uk/socket.h
```diff
@@ -0,0 +1,63 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UK_SOCKET__
+#define __UK_SOCKET__
+
+struct vfscore_file;
+struct posix_socket_driver;
+
+struct posix_socket_file {
+	/** The fd or data used internaly by the socket implementation */
```
craciunoiuc:
```suggestion
	/** The fd or data used internally by the socket implementation */
```

lib/posix-socket/socket_vnops.c
```diff
@@ -0,0 +1,315 @@
...
+struct vnops posix_socket_vnops = {
+	.vop_close = posix_socket_vfscore_close,
+	.vop_write = posix_socket_vfscore_write,
+	.vop_read = posix_socket_vfscore_read,
+	.vop_ioctl = posix_socket_vfscore_ioctl,
+	.vop_getattr = posix_socket_vfscore_getattr,
+	.vop_inactive = posix_socket_vfscore_inactive
+};
+
+#define posix_socket_vget ((vfsop_vget_t) vfscore_nullop)
```
craciunoiuc: These three defines give off warnings:

```
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:291:39: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct vnode *, struct vattr *)’ [-Wcast-
function-type]                                                                                                                                                                                   
  291 | #define posix_socket_vfscore_getattr ((vnop_getattr_t) vfscore_vop_einval)                                                                                                               
      |                                       ^                                                                                                                                                  
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:300:17: note: in expansion of macro ‘posix_socket_vfscore_getattr’                                                                  
  300 |  .vop_getattr = posix_socket_vfscore_getattr,                                                                                                                                            
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                             
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:292:40: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct vnode *)’ [-Wcast-function-type]  
  292 | #define posix_socket_vfscore_inactive ((vnop_inactive_t) vfscore_vop_nullop)                                                                                                             
      |                                        ^                                                                                                                                                 
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:301:18: note: in expansion of macro ‘posix_socket_vfscore_inactive’
  301 |  .vop_inactive = posix_socket_vfscore_inactive                                          
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                          
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:304:28: warning: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(struct mount *, struct vnode *)’ [-Wcast-
function-type]
  304 | #define posix_socket_vget ((vfsop_vget_t) vfscore_nullop)                               
      |                            ^                                                            
/home/cez/.unikraft/unikraft/lib/posix-socket/socket_vnops.c:308:14: note: in expansion of macro ‘posix_socket_vget’
  308 |  .vfs_vget = posix_socket_vget,                                                         
      |              ^~~~~~~~~~~~~~~~~   
```
There shouldn't be any problem with this related to functionality.
Still, it would be good if we didn't add more build warnings knowingly.

craciunoiuc (changes requested):
Hello @marcrittinghaus @nderjung,

I added my review and my ideas. The PR is pretty mature so I think it looks good.

I have some questions though:
 * What would be a clear use case for this library? From what I could see, it is not compatible with `newlib`, as it does not have `<sys/socket.h>`. Also, it is not compatible with `lwip` as it implements much of the same stuff as this (but with different function signatures).
 * Could this be used eventually with nginx? If it were to swap `lwip` for this, I can't see how it would be done, as `nginx` has a dependency on `lwip` right now.

I'm asking this because I'm not sure what I should test except that the functions work as expected (and I was hoping that using it with `nginx` would save me some time).

skuenzer:
Hi @craciunoiuc,

with this interface library we want to replace the current socket implementation. Basically it should enable us better flexibility and support for multiple socket types. With lwip we will still provide UDP and TCP sockets. But Unix Domain Sockets, for instance, could be provided by a different library at the same time. Common socket calls are fetched by posix-socket and redirected to the according socket implementation.
Together with this PR there should be another PR for lwip that binds the UDP and TCP socket families to posix-socket. Is this one available, @nderjung , @marcrittinghaus ?

Thanks,
Simon

craciunoiuc:
> Hi @craciunoiuc,
> 
> with this interface library we want to replace the current socket implementation. Basically it should enable us better flexibility and support for multiple socket types. With lwip we will still provide UDP and TCP sockets. But Unix Domain Sockets, for instance, could be provided by a different library at the same time. Common socket calls are fetched by posix-socket and redirected to the according socket implementation. Together with this PR there should be another PR for lwip that binds the UDP and TCP socket families to posix-socket. Is this one available, @nderjung , @marcrittinghaus ?
> 
> Thanks, Simon

Okay, It makes sense now. I was wondering if I was doing something wrong in regard to `lwip`. Thanks!

Cezar

marcrittinghaus:
Hi @craciunoiuc, @skuenzer,

To avoid misunderstanding, I'd like to add that this library provides (at least at the moment) only the driver API and the socket syscalls. So `posix-socket` does not actually implement the socket API but only functions as the POSIX socket interface to applications and other libraries. It requires `lwip` to register a driver for `AF_INET` and throw out its entire syscall interface. However, we still depend on `lwip`'s socket implementation (except `epoll`/`select`).

I have internally a modified `lwip` that connects to `posix-socket` but I have not yet created a PR. I doubt it will make it in the release tomorrow.

lib/posix-socket/socket_vnops.c
marcrittinghaus: This throws warnings only without https://github.com/unikraft/unikraft/pull/405. The problem is that the `nop` style functions where defined as prototyped functions, which causes the compiler to warn about the mismatching function signatures. PR 405 changes the `nop` functions to be non-prototyped so that the function signatures are not checked anymore. This is the way to go, so I would not change anything in this PR.

But a good catch! @skuenzer We should upstream https://github.com/unikraft/unikraft/pull/405 before this one to avoid the warnings from popping up. @mschlumpp already reviewed the PR, but you have been assigned to it (because of your plenty free time! 😏)

craciunoiuc:
> Hi @craciunoiuc, @skuenzer,
> 
> To avoid misunderstanding, I'd like to add that this library provides (at least at the moment) only the driver API and the socket syscalls. So `posix-socket` does not actually implement the socket API but only functions as the POSIX socket interface to applications and other libraries. It requires `lwip` to register a driver for `AF_INET` and throw out its entire syscall interface. However, we still depend on `lwip`'s socket implementation (except `epoll`/`select`).
> 
> I have internally a modified `lwip` that connects to `posix-socket` but I have not yet created a PR. I doubt it will make it in the release tomorrow.

Yes, that's what I thought. Maybe then we should add somewhere in the config a mention that it's at-the-moment not usable with lwip?

Or should we just move fast and integrate that too into lwip?

lib/posix-socket/driver_list.ld
marcrittinghaus: > > Hi @craciunoiuc, @skuenzer,
> > To avoid misunderstanding, I'd like to add that this library provides (at least at the moment) only the driver API and the socket syscalls. So `posix-socket` does not actually implement the socket API but only functions as the POSIX socket interface to applications and other libraries. It requires `lwip` to register a driver for `AF_INET` and throw out its entire syscall interface. However, we still depend on `lwip`'s socket implementation (except `epoll`/`select`).
> > I have internally a modified `lwip` that connects to `posix-socket` but I have not yet created a PR. I doubt it will make it in the release tomorrow.
> 
> Yes, that's what I thought. Maybe then we should add somewhere in the config a mention that it's at-the-moment not usable with lwip?
> 
> Or should we just move fast and integrate that too into lwip?

Let me see if I can make a PR for it as the code should be mostly ok.

marcrittinghaus:
> > Hi @craciunoiuc, @skuenzer,
> > To avoid misunderstanding, I'd like to add that this library provides (at least at the moment) only the driver API and the socket syscalls. So `posix-socket` does not actually implement the socket API but only functions as the POSIX socket interface to applications and other libraries. It requires `lwip` to register a driver for `AF_INET` and throw out its entire syscall interface. However, we still depend on `lwip`'s socket implementation (except `epoll`/`select`).
> > I have internally a modified `lwip` that connects to `posix-socket` but I have not yet created a PR. I doubt it will make it in the release tomorrow.
> 
> Yes, that's what I thought. Maybe then we should add somewhere in the config a mention that it's at-the-moment not usable with lwip?
> 
> Or should we just move fast and integrate that too into lwip?

Let me see if I can make a PR for it as the code should be mostly ok.

lib/posix-socket/driver_list.ld
marcrittinghaus: I have not had any problems with it yet and I tested with a lot of other libraries and also on KVM and XEN. If you remove the alignment the driver list will be put at a byte-aligned location which is not good for performance and can cause problems if the architecture does not support unaligned memory accesses. So I would keep it for now.

lib/posix-socket/include/uk/socket.h
marcrittinghaus: I fixed it 👍🏻

lib/posix-socket/driver_list.ld
craciunoiuc: I didn't have problems with `ALIGN` either when I wanted to use it. After that bug, I removed it as per @skuenzer 's requests.

@skuenzer should `ALIGN` stay here or not? For me it's fine either way

lib/posix-socket/socket_vnops.c
craciunoiuc: Perfect then. Well @skuenzer is assigned to both, so he can control the order or merges

razvand:
@marcrittinghaus , also make sure conflicts are resolved.

marcrittinghaus:
> @marcrittinghaus , also make sure conflicts are resolved.

The conflict was due upstreaming of PR #407 that also changed the file. Conflict resolved.

marcrittinghaus:
I checked what has to be done for also publishing the `lwip` side. I think it does not make sense. The problem is that the `select` and `poll` calls depend on the new `eventpoll` infrastructure that is not published either, yet. I looked into the possibility to have these calls bypass `posix-socket` and be directly implemented in `lwip` but it becomes a bit of a mess, where I also have to write new code. I think in such an important area like network, we should not unnecessarily rush for putting untested code upstream, just for the sake of having a "half-baked" modern version. 

We should instead aim at publishing the `eventpoll` infrastructure and then add `posix-socket` and `lwip` all at once for the next release. It is also fine to upstream `posix-socket` now, but it won't be of any particular use as @craciunoiuc correctly pointed out.

craciunoiuc:
@razvand @nderjung 

Should we go for this release or the next? You are the guys with the milestones 😄

marcrittinghaus:
I updated the PR to include the polling interface.

This PR is now part of a larger group of dependent PRs which must be merged in the following order:
1. https://github.com/unikraft/lib-newlib/pull/23
2. https://github.com/unikraft/unikraft/pull/484
3. https://github.com/unikraft/unikraft/pull/65  (this PR)
4. https://github.com/unikraft/lib-lwip/pull/17
5. https://github.com/unikraft/unikraft/pull/485

⚠️ Without having merged all of these PRs, network is no longer functional. All PRs should thus be checked and reviewed first and then merged in quick succession⚠️

craciunoiuc (commented):
I already looked at this once, but I checked it again.

If everything else discussed above was fixed, I don't have anything more to add now.

lib/posix-socket/socket_vnops.c
```diff
@@ -0,0 +1,341 @@
...
+	UK_ASSERT(vnode->v_type == VSOCK);
+
+	sock = (struct posix_socket_file *)vnode->v_data;
+
+	ret = posix_socket_poll(sock, revents, ecb);
+	if (ret < 0)
+		return -ret;
+	if (unlikely(ret < 0)) {
+		PSOCKET_ERR("poll on socket %d failed: %d\n", fp->fd,
+			    (int)ret);
```
craciunoiuc: `fp` is missing here from the signature. Is it lingering in the error?

marcrittinghaus: Yap! That needs to be `sock->vfs_file->fd`

marcrittinghaus:
@craciunoiuc

Thanks for the feedback regarding `EAGAIN`. I have added checks in some places.

marcrittinghaus:
@craciunoiuc I changed the `EAGAIN` checks for the VNOPS to only affect the output and not the actual return values so `EAGAIN` is still regarded as an error condition and the application is notified correctly but without spamming the log.

craciunoiuc:
All good on my side now

I'll do one more pass/test just in case to make sure nothing pops up.
Edit: Yup, all good

Your turn now @skuenzer ;)

marcrittinghaus:
@craciunoiuc, @razvand Just noticed that if the socket error message is off, the single-line `if` statement introduced in the last push would break. This fixes the issue.

Other than that, I tested this again with `nginx` and everything works. However, as @craciunoiuc also mentions on discord, we have to update the various network-enabled applications/libraries to reference `posix-socket` and `posix-event` instead of `liblwip`. But I would not do this as part of this PR series.

lib/nolibc/include/sys/socket.h
```diff
@@ -0,0 +1,207 @@
...
+
+int socket(int family, int type, int protocol);
+int socketpair(int family, int type, int protocol, int usockfd[2]);
+
+int shutdown(int sock, int how);
+
+int bind(int sock, const struct sockaddr *addr, socklen_t addr_len);
+int connect(int sock, const struct sockaddr *addr, socklen_t addr_len);
+int listen(int sock, int backlog);
+int accept(int sock, struct sockaddr *restrict addr,
```
razvand, line 197: The `accept4` function declaration is missing. Is this intentional?

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -0,0 +1,633 @@
...
+ *
+ * @param sock Reference to the socket.
+ * @param addr The address of the peer socket.
+ * @param addr_len Specifies the size, in bytes, of the address structure
+ *    pointed to by addr
+ *
+ * @return NULL or a valid pointer to driver-specific data on success,
+ *    ERR2PTR(-errno) otherwise
+ */
+typedef void *(*posix_socket_accept_func_t)(struct posix_socket_file *sock,
```
razvand: The `posix_socket_accept4_func_t` type is missing. Is this intentional?

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -0,0 +1,633 @@
...
+
+/**
+ * A structure containing the functions exported by a Unikraft socket driver
+ */
+struct posix_socket_ops {
+	/* The initialization function on socket registration. */
+	posix_socket_driver_init_func_t   init;
+	/* POSIX interface */
+	posix_socket_create_func_t        create;
+	posix_socket_accept_func_t        accept;
```
razvand: The comment with `accept4`.

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -0,0 +1,633 @@
...
+		    int protocol)
+{
+	UK_ASSERT(d);
+	UK_ASSERT(d->ops->create);
+
+	return d->ops->create(d, family, type, protocol);
+}
+
+static inline void *
+posix_socket_accept(struct posix_socket_file *sock,
```
razvand: The comment with `accept4` again.

lib/posix-socket/include/uk/socket.h
```diff
@@ -0,0 +1,63 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
razvand, line 35:
```suggestion
 */
 
```

lib/posix-socket/include/uk/socket_vnops.h
```diff
@@ -0,0 +1,67 @@
...
+#include <vfscore/file.h>
+#include <vfscore/vnode.h>
+
+struct posix_socket_driver;
+
+/**
+ * Return the socket file structure used by a file descriptor.
+ *
+ * @param sock_fd
+ *  The vfscore file descriptor number.
```
razvand:
```suggestion
 * @param sock_fd The vfscore file descriptor number.
```

lib/posix-socket/include/uk/socket_vnops.h
```diff
@@ -0,0 +1,67 @@
...
+ */
+struct posix_socket_file *
+posix_socket_file_get(int sock_fd);
+
+/**
+ * Allocate a file descriptor in vfscore for the given socket driver.
+ *
+ * @param d
+ *  The driver used to create the socket.
```
razvand:
```suggestion
 * @param d The driver used to create the socket.
```

lib/posix-socket/include/uk/socket_vnops.h
```diff
@@ -0,0 +1,67 @@
...
+ * @param type
+ *  The socket type.
```
razvand:
```suggestion
 * @param type The socket type.
```

lib/posix-socket/include/uk/socket_vnops.h
```diff
@@ -0,0 +1,67 @@
...
+ * @param sock_data
+ *  The socket implementation's private data.
```
razvand:
```suggestion
 * @param sock_data The socket implementation's private data.
```

lib/posix-socket/socket.c
```diff
@@ -0,0 +1,639 @@
...
+
+	trace_posix_socket_listen(sock, backlog);
+
+	file = posix_socket_file_get(sock);
+	if (unlikely(PTRISERR(file))) {
+		ret = PTR2ERR(file);
+		goto EXIT_ERR;
+	}
+
+	/* Listen to the socket */
```
razvand:
```suggestion
	/* Listen on the socket */
```

lib/posix-socket/socket.c
```diff
@@ -0,0 +1,639 @@
...
+		goto EXIT_ERR_FREE_FD1;
+	}
+
+	usockfd[0] = vfs_fd1;
+	usockfd[1] = vfs_fd2;
+
+	trace_posix_socket_socketpair_ret(ret);
+	return ret;
+EXIT_ERR_FREE_FD1:
+	/* TODO: Broken error handling. Leaking vfs_fd1! */
```
razvand, line 630: Why can't `vfs_fd1` be freed? Isn't there a `free()` equivalent to the `posix_socket_alloc_fd()` function?

razvand (changes requested):
Hi, @marcrittinghaus . Thanks for the work. I used it before for binary compatibility and it did the trick. I will test it again.

I did a first read pass. See my suggestions as inline comments.

Apart from this, some other comments below.

There is an inconsistency between macros in `sys/socket.h` (as I understand, this is imported from OpenBSD) and [the macros in Linux](https://elixir.bootlin.com/linux/latest/source/include/linux/socket.h#L179) for address families. Some examples:
* `AF_BLUETOOTH` is 32 - in Linux: 31
* `AF_ROUTE` is 17 - in Linux: 16
* `AF_INET6` is 24 - in Linux: 10
* `AF_IPX` is 23 - in Linux: 4

This may be an issue when using binary compatibility mode.

For Doxygen comments, there is an inconsistency around the use of dot (`.`) at the end of a comment line or not. Either should be OK, as long as it's consistent.

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -0,0 +1,633 @@
...
+ */
+typedef int (*posix_socket_socketpair_func_t)(struct posix_socket_driver *d,
+		int family, int type, int protocol, void *sockvec[2]);
+
+/**
+ * Write to a socket file descriptor.
+ *
+ * @param sock Reference to the socket
+ * @param buf The pointer to the buffer to write to the socket
+ * @param count The number of bytes to be written
```
razvand: These two lines need to be updated to `iov` and `iovcnt`.

lib/posix-socket/include/uk/socket_driver.h
```diff
@@ -0,0 +1,633 @@
...
+ */
+typedef ssize_t (*posix_socket_write_func_t)(struct posix_socket_file *sock,
+		const struct iovec *iov, int iovcnt);
+
+/**
+ * Read from a socket file descriptor.
+ *
+ * @param sock Reference to the socket
+ * @param buf The buffer to read the data from the socket into
+ * @param count The number of bytes to be read
```
razvand: These two lines need to be updated to `iov` and `iovcnt`.

razvand (changes requested):

lib/posix-socket/include/uk/socket_driver.h
marcrittinghaus: @razvand Yes, this was intentional. This is the driver interface. It is thus not bound to the POSIX standard but only resembles the interface for simplicity. You may notice that there is no `recv` and `send` either. Instead both are realized by expecting the driver to implement the more powerful `recvfrom` and `sendto` calls and `posix-socket` internally calls these methods.

I do not want a driver having to implement `accept` and `accept4`, so I changed the interface so that a driver needs to provide `accept4` only.

marcrittinghaus:
> There is an inconsistency between macros in `sys/socket.h` (as I understand, this is imported from OpenBSD) and [the macros in Linux](https://elixir.bootlin.com/linux/latest/source/include/linux/socket.h#L179) for address families. Some examples:
> 
> * `AF_BLUETOOTH` is 32 - in Linux: 31
> * `AF_ROUTE` is 17 - in Linux: 16
> * `AF_INET6` is 24 - in Linux: 10
> * `AF_IPX` is 23 - in Linux: 4
> 

I redefined all of the address families by taking the values from musl. This should increase compatibility.

razvand:
@marcrittinghaus , all is OK. @craciunoiuc , please approve this PR and then I'll approve it as well. I'll be careful to do the approval in the order signaled by @marcrittinghaus .

craciunoiuc (approved):
All good on my side!

razvand (approved):
Thanks, @marcrittinghaus, @nderjung  . This works.

PR #71, opened 2020-11-27
solve ukswrand does not generate random numbers

### LINKED ISSUE #70: ukswrand does not generate random numbers
if you configure `ukswrand` with the platform timestamp as initial seed, and you'll try to use `getrandom` function or try to read from `/dev/urandom` you won't get random numbers. This is because the initialization of the random number generator, which is located in the swrand.c file, is done earlier than the initialization of TSC clock.
To solve this problem you have to make the following changes in swrand.c: 
- include <uk / init.h>
- replace the registration of the constructor `UK_CTOR_PRIO (_uk_swrand_ctor, UK_SWRAND_CTOR_PRIO);`  with  `uk_early_initcall (_uk_swrand_ctor);`

dragosargint:
If you configure ukswrand with the platform timestamp as initial seed, and you'll try to use getrandom function or try to read from /dev/urandom you won't get random numbers. This is because the initialization of the random number generator, which is located in the swrand.c file, is done earlier than the initialization of TSC clock.

### DIFF
--- a/lib/ukswrand/swrand.c
+++ b/lib/ukswrand/swrand.c
@@ -31,7 +31,6 @@
  */
 #include <string.h>
 #include <uk/swrand.h>
-#include <uk/ctors.h>
 #include <uk/config.h>
 #include <uk/print.h>
 #include <uk/init.h>
@@ -76,7 +75,7 @@ ssize_t uk_swrand_fill_buffer(void *buf, size_t buflen)
 	return buflen;
 }
 
-static void _uk_swrand_ctor(void)
+static int _uk_swrand_init(void)
 {
 	unsigned int i;
 #ifdef CONFIG_LIBUKSWRAND_CHACHA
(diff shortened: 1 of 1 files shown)


b4laur (changes requested):
Hi, Dragos. I have a few comments regarding your PR:
1. The include of the `uk/ctors.h` header is unnecessary since we will no longer use the `UK_CTOR_PRIO` macro.
2. Since the `uk_early_initcall` macro expects an `int fn(void)` function type, we should change the signature of the `_uk_swrand_ctor()` function and because the `int` type requests a return, we can return `seedc`. Also, for consistency, we should rename it `_uk_swrand_init`.

b4laur (approved):
Looks good to me.

Reviewed-by: Laurentiu Barbulescu <lrbarbulescu@gmail.com>

PR #78, opened 2020-12-11
plat/xen: remove #ifndef CONFIG_PARAVIRT code

### LINKED ISSUE #27: remove code under #ifndef CONFIG_PARAVIRT
We inherited bits of PVH code from Minios. The problem is this code did not work even for Minios. There is little point in keeping these bits. However it might be good to keep the -DCONFIG_PARAVIRT, as hints for future

maruthgoyal:


nderjung:
Hi!

Thanks for the PR to solve #27.  It looks like you've included changes which also do heavy reformats which are not in the scope of removing compile-time `CONFIG_PARAVIRT` blocks which subsequently give warnings from the checkpatch tool.

Could you please force-push to this branch (and PR) which only make these following changes or create separate commits which do correct refactors.  Also don't forget to `Signed-off-by:` ;)

Thanks!

razvand:
@maruthgoyal , could you please take a look in this PR? Would you want it to be owned by someone else?

razvand:
@StefanJum , please see if this PR is still useful / interesting to us. If so, please own it. If not, we'll close it.

PR #109, opened 2021-02-10
lib/vfscore: Split readv/write from preadv/pwritev

hisharanwithu:
The patch series split the readv/writev implementation
from preadv/pwritev. We split them up to separate the
different error checks required on the different flavors
of read/write call. The patch also make it consistent on
the different patches of the read/write to return negative
error values and let the syscall shim library to the conversion
of error value to appropriate errno values.

Change in V1
- lib/vfscore: UK_VFSCORE_NOPOS flag to check seek
- lib/vfscore: Perform readv and writev operations
- lib/vfscore: Implement readv and writev
- lib/vfscore: Neg error and use UK_SYSCALL_R_DEFINE

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -337,7 +337,7 @@ UK_SYSCALL_R_DEFINE(ssize_t, preadv, int, fd, const struct iovec*, iov,
 		goto out_error;
 
 	/* Check if the file is indeed seekable. */
-	if (!(fp->f_vfs_flags & UK_VFSCORE_NOPOS)) {
+	if (fp->f_vfs_flags & UK_VFSCORE_NOPOS) {
 		error = ESPIPE;
 		goto out_error_fdrop;
 	}
@@ -427,7 +427,7 @@ UK_SYSCALL_R_DEFINE(ssize_t, pwritev, int, fd, const struct iovec*, iov,
 		goto out_error;
 
 	/* Check if the file is indeed seekable. */
-	if (!(fp->f_vfs_flags & UK_VFSCORE_NOPOS)) {
+	if (fp->f_vfs_flags & UK_VFSCORE_NOPOS) {
 		error = ESPIPE;
 		goto out_error_fdrop;
 	}
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -353,9 +353,29 @@ out_errno:
 
 LFS64(pread);
 
+UK_TRACEPOINT(trace_vfs_read, "%d %p %d", int, void *, int);
+UK_TRACEPOINT(trace_vfs_read_ret, "0x%x", ssize_t);
+UK_TRACEPOINT(trace_vfs_read_err, "%d", int);
+
 UK_SYSCALL_DEFINE(ssize_t, read, int, fd, void *, buf, size_t, count)
 {
-	return pread(fd, buf, count, -1);
+	ssize_t bytes;
+
+	UK_ASSERT(buf);
+
+	struct iovec iov = {
+			.iov_base	= buf,
+			.iov_len	= count,
+	};
+
+	trace_vfs_read(fd, buf, count);
+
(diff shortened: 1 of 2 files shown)


lib/vfscore/main.c
```diff
@@ -73,6 +73,19 @@ static inline int libc_error(int err)
...
+// error, so the system call appear a successful partial read/write.
+// In FreeBSD, dofilewrite() and dofileread() (sys_generic.c) do this too.
+static inline int has_error(int error, int bytes)
+{
+	/* TODO: OSv checks also for ERESTART */
+	return error && (
+		(bytes == 0) ||
+		(error != EWOULDBLOCK && error != EINTR));
+}
+
```
nderjung, line 88: Hey Sharan, in [a previous commit](https://github.com/unikraft/unikraft/commit/d04d5f4a10884844275af00e6605d595e3453e26#diff-50257766dd062f03b631cf9789a11b13c936be36b9423bca03ec8277b79264b0L293-L306) I removed this particular error handler in favour of inline error handling as it was simple enough to be inline.  

However, this PR is fixing the error introduced from said commit series of negated (and thus invalid use of) `errno` and return values (my bad!) by re-introducing this error handler with an even more verbose approach to `pread`(`v`) and `pwrite`(`v`).  So it may be worth, therefore, to increase the scope of this error handler even further by abstracting the check for seekablility and whether the file has already been read or is from a char device in [`preadv`](https://github.com/unikraft/unikraft/pull/109/files#diff-50257766dd062f03b631cf9789a11b13c936be36b9423bca03ec8277b79264b0R396-R408) and [`pwritev`](https://github.com/unikraft/unikraft/pull/109/files#diff-50257766dd062f03b631cf9789a11b13c936be36b9423bca03ec8277b79264b0R564-R576) since it feels as though there is a lot of repeated code.

What do you think?  The original intention of the aforementioned series was to simplify the code by reducing `pread` and `preadv` (and for `pwrite` and `pwritev`) which had a bunch of overlapping functionality.

hisharanwithu: Hello @nderjung,

Thank you for the review. 

The original intention of the has_error was to check both the error as well as the nr of bytes returned. I don't think error < 0 captures this completely as there are cases where EINTR is returned but certain amount of data is returned. Plus the condition of error < 0 is not valid as sys_read and sys_write return positive error.

As far as increasing the scope of error handler is concerned, I would not support this idea because the current error check is done once the read and write operation is performed, while the other error check you would want to perform is for the input parameter check. In my opinion they are meant to performed separately at different point in the code so keeping them separate would be better. 

If your concern is about duplicate code checks being performed on preadv and readv, maybe we could extend prototype of do_preadv
```
static ssize_t do_preadv(struct vfscore_file *fp, const struct iovec *iov,
			 int iovcnt, off_t offset, bool has_offset, ssize_t *bytes)
```
 and perform the offset checks within do_preadv based on the has_offset flag.

nderjung: Hi Sharan,

> The original intention of the has_error was to check both the error as well as the nr of bytes returned. I don't think error < 0 captures this completely as there are cases where EINTR is returned but certain amount of data is returned. Plus the condition of error < 0 is not valid as sys_read and sys_write return positive error.

This makes sense and it was a mistake to simplify it.

> If your concern is about duplicate code checks being performed on preadv and readv, maybe we could extend prototype of do_preadv

This was my main concern, I think if `has_error` should be paired with another abstract error hanldler for this:

https://github.com/unikraft/unikraft/pull/109/files#diff-50257766dd062f03b631cf9789a11b13c936be36b9423bca03ec8277b79264b0R390-R408

Which is duplicate for `preadv` and `pwritev`.

nderjung: As discussed offline, I will approve this PR to get it upstream faster and we can discuss these changes in a seperate PR.

nderjung (approved):
Thanks Sharan :)  Just tested these proper and all works as expected.

PR #135, opened 2021-02-16
lib/uksignal: Register system call from uksignal

hisharanwithu:
Register the system calls implemented in the uksignal library. The signal
library provides support for the standard signals and do not yet support
real time signals. The `rt_sigtimedwait` system call does not yet support
time out.

[[-]] lib/uksignal: Register rt_sigaction
- lib/uksignal: Register rt_sigprocmask
- lib/uksignal: Register rt_sigpending
- lib/uksignal: Register rt_sigsuspend
- lib/uksignal: Register rt_sigtimedwait
- lib/uksignal: Register tkill


lib/uksignal/Config.uk
```diff
@@ -6,3 +6,11 @@ menuconfig LIBUKSIGNAL
 	select LIBUKALLOC
 	select LIBUKSCHED
 	select LIBPOSIX_PROCESS
+	select LIBSYSCALL_SHIM
+	select LIBSYSCALL_SHIM_LIBCSTUBS
+
+if LIBUKSIGNAL
+	config LIBUKSIGNAL_LIBCFNS_ENABLE
+		bool "signal: Provide libc signal implementation"
+		default n
```
mogasergiu, line 15: Just a nitpick, but shouldn't this be ```default y```? So that apps such as ```nginx``` can compile by default without errors, as without this option enabled, ```nginx``` fails to compile.

mogasergiu (changes requested):
It all looks good to me, but requires a rebase and some changes with respect to the updated ```checkpatch```.

razvand:
@razvanvirtan will create a new PR rebasing the current one.

PR #141, opened 2021-03-02
mritting/fix_ssp

marcrittinghaus:
The current random stack canary implementation is broken. The canary is 0 after initialization. Furthermore, only the lower 32-bits of the canary are used on 64-bit platforms. Using terminator characters would also make sense, especially as part of the default values.

```gdb
Breakpoint 1, ukplat_entry (argc=1, argv=0x7fffffffe498) at unikraft/lib/ukboot/boot.c:193
193             UKSP_INIT_CANARY();
(gdb) p/x __stack_chk_guard
$4 = 0xdeadbeef
(gdb) n
198             uk_pr_info("Unikraft constructor table at %p - %p\n",
(gdb) p/x __stack_chk_guard
$5 = 0x0
```

### DIFF
--- a/lib/ukboot/boot.c
+++ b/lib/ukboot/boot.c
@@ -63,7 +63,7 @@
 #ifdef CONFIG_LIBUKLIBPARAM
 #include <uk/libparam.h>
 #endif /* CONFIG_LIBUKLIBPARAM */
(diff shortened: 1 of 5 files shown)


vladandrew (approved):
Great Patch!

PR #150, opened 2021-03-14
lib/vfscore: Fix `pipe_read` to return when data is available

dragosargint:
The problems that this PR is trying to solve are:
- if we write n bytes in the pipe and try to read more, the program hangs. So, we should return if we managed to read available data.
-  when we close the write end of the pipe and there's no available data, the program is also looping.
We should exit if all the references of the write descriptor are deleted.

### DIFF
--- a/lib/vfscore/pipe.c
+++ b/lib/vfscore/pipe.c
@@ -137,6 +137,11 @@ static int pipe_buf_can_read(struct pipe_buf *pipe_buf)
 	return pipe_buf_get_available(pipe_buf) > 0;
 }
 
+static int pipe_file_can_read(struct pipe_file *pipe_file)
+{
+	return pipe_buf_can_read(pipe_file->buf) || !(pipe_file->w_refcount);
+}
+
 static unsigned long pipe_buf_write(struct pipe_buf *pipe_buf,
 		struct iovec *iovec, size_t iovec_off)
 {
@@ -335,7 +340,11 @@ static int pipe_read(struct vnode *vnode,
 					while (!pipe_buf_can_read(pipe_buf)) {
 						uk_mutex_unlock(&pipe_buf->rdlock);
 						uk_waitq_wait_event(&pipe_buf->rdwq,
-							pipe_buf_can_read(pipe_buf));
+							pipe_file_can_read(pipe_file));
+
+						if (!pipe_file->w_refcount)
+							return 0;
+
 						uk_mutex_lock(&pipe_buf->rdlock);
 					}
 				}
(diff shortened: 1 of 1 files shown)


skuenzer (changes requested):
ab5d56f: Hum... We had a reason why we moved `_uk_swrandr_ctor` as early as possible during boot and I think this was because of `libuksp`. Alternatively, you could move it into an init function only if the timestamp was selected as source. Can you also hide that option from `Config.uk` when `libuksp` is selected for a build?

46dc32e: Just for me to make sure: Does reading from the pipe block when there is no data available?

General comments:
- Please make sure that `checkpatch` does not throw any avaiodable error.
- Make sure that the commit subject format follows our guidelines. See `CONTRIBUTING.md`
- You should merge commit 46dc32e with 93ec8a0
- Please rebase your branch to the latest staging instead of merging. This avoids commit c9dff52

dragosargint:
The PR for ukswrand was already closed here [1]
I think I did something wrong when I merged. I will rebase and update the PR.

Regarding reading from the pipe, it blocks when there is no written data.
So if no write has been done before, the read call is blocked. 
I didn't change that, because that's the default on linux as well.
(However, if the writing end of the pipe is closed, we must return from the read even if no data is available.) 
The problem was that it would block if you did something like this
write (p [1], ..., 10);
read (p [0], ..., n)); where n > 10

Thank you for your comments!

[1] https://github.com/unikraft/unikraft/pull/71

dragosargint:
I rebase my branch and I also merged these commits:
https://github.com/unikraft/unikraft/commit/46dc32ef05bfa6f1c2aa40f961804816e74ca9e1
https://github.com/unikraft/unikraft/commit/93ec8a0326e6acaa1ca5bb1ab9a3f6722ac0d479

lib/vfscore/pipe.c
```diff
@@ -211,6 +216,7 @@ static unsigned long pipe_buf_read(struct pipe_buf *pipe_buf,
 
 out:
 	return to_read;
+
```
skuenzer: Can you avoid introducing this blank line from the commit? I guess it is here by mistake.

skuenzer (changes requested):
I have just a minor comment to your patch, otherwise it looks fine.

lib/vfscore/pipe.c
dragosargint: Yes, It was there by mistake. I removed it. 

Thank you for your comments!

dragosargint:
I would like to say in advance that I did not solve the checkpatch.pl warnings to maintain consistency with the code already written. There are several lines that are over 80 characters long.

mogasergiu (approved):
This looks fine.

skuenzer (approved):

PR #151, opened 2021-03-15
lib/isrlib: "Interrupt-service-routine"-safe standard functions

skuenzer:
Introduces libisr, a library for providing "interrupt-service-routine"-
safe variants of standard functions so that they can be used in
interrupt handlers and early boot code. This commit provides an initial
set of standard functions normally found with `<string.h>`. The
implementation is derived from nolibc and compiled with the `|isr`
variant which disables the use of extended CPU features, like
vector units (e.g., SSE, AVX) and floating point units.

Signed-off-by: Simon Kuenzer <simon.kuenzer@neclab.eu>

Note: This series is a follow-up version of patch [738374](https://patchwork.unikraft.org/patch/738374/) on patchwork.


lib/isrlib/Config.uk
```diff
@@ -0,0 +1,4 @@
+config LIBISRLIB
+    bool "isrlib: Interrupt-safe standard routines"
+    default n
+    select LIBNOLIBC if !HAVE_LIBC
```
hisharanwithu, line 4: should we select libc? If we need it for headers, should we move header into the library?

skuenzer: We depend on standard libc-provided headers, especially for the data types provided with `<stddef.h>`. The idea is that the prototypes should not differ from standard libc definitons (except, slightly renamed with `*_isr`) and we should be able to re-use things like `uint8_t`, `NULL`, and error numbers.

hisharanwithu: Yeah it make sense to use those header.

The other discussion on including the libc, need more support from the build system to prevent call from this library to the function that aren't to be called from a interrupt routine. As this extends the scope of this patch, we should treat it separate set.

hisharanwithu:
Patch looks good to me.

Reviewed-by: Sharan Santhanam <sharan.santhanam@neclab.eu>

PR #152, opened 2021-03-15
build: Warn about empty libraries and platforms

skuenzer:
Print a warning on the console when libraries and/or platforms are dropped
from the build because they are empty. The message is shown when verbose
mode is active (`V=1`).


b4laur (approved):
Hi Simon,

It looks good.

Reviewed-by: Laurentiu Barbulescu <lrbarbulescu@gmail.com>

PR #166, opened 2021-03-27
support: checkpatch: Prefer raw system call definitions

skuenzer:
Introduces a warning for system call definitions when `UK_SYSCALL_DEFINE` and `UK_LLSYSCALL_DEFINE` was used:
Raw system call definitions are preferred because `errno` is not used.


nderjung (approved):
Tested. Looks good to me! Thanks!

PR #169, opened 2021-03-29
support: checkpatch: Unikraft subject format

skuenzer:
Checks and fails with an error when the subject line format is not
following the Unikraft scheme.
This check should be disabled for repositories that are not the
Unikraft core with `--ignore UNIKRAFT_SUBJECT_FORMAT`.


nderjung (approved):
Tested.  Looks good to me!  Thanks! :)

PR #172, opened 2021-03-29
Skuenzer/ukdebug printonce

skuenzer:
Provide `uk_printd_once()` and `uk_printk_once()` that prints messages just once. This is useful for code paths where otherwise the message printing would cause flooding of the console.


razvand (approved):
I tested the `uk_*_once()` functions. Everything is OK.

PR #175, opened 2021-03-31
Improvements to the timer driver and timer functions

rene:
This pull request provides fixes and improvements to the timer driver and timer functions:

- plat/common/arm/generic_timer.c: Fix time conversion functions
- plat/common/arm/generic_timer.c: Keep consistence of delay calculation
- plat/common: Fix time argument types to unsigned
- plat: Fix timer IRQ number provided to _ukplat_irq_handle()
- plat/common/arm: Export function to get system clock ticks

Signed-off-by: Renê de Souza Pinto <Rene.deSouzaPinto@opensynergy.com>

### DIFF
--- a/plat/common/arm/generic_timer.c
+++ b/plat/common/arm/generic_timer.c
@@ -56,6 +56,9 @@ static uint32_t ns_per_tick;
 /* Multiplier for converting nsecs to counter ticks */
 static uint32_t tick_per_ns;
 
+/* Total (absolute) number of nanoseconds per tick */
+static uint64_t tot_ns_per_tick;
+
 /*
  * The maximum time range in seconds which can be converted by multiplier
  * and shift factors. This will guarantee the converted value not to exceed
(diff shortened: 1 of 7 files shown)


rene:
checkpatch errors are fixed now.

plat/common/arm/generic_timer.c
```diff
@@ -225,14 +231,13 @@ static inline uint64_t generic_timer_epochoffset(void)
...
 
 	UK_ASSERT(ukplat_lcpu_irqs_disabled());
 
-	/* Record current ns and until_ticks for timer */
-	now_ns = ukplat_monotonic_clock();
-	until_ticks = generic_timer_get_ticks()
-				+ ns_to_ticks(until_ns - now_ns);
+	/* Record current ns (and ticks) and calculate until_ticks for timer */
+	now_ns      = generic_timer_monotonic_ticks(&now_ticks);
+	until_ticks = now_ticks + ns_to_ticks(until_ns - now_ns);
```
razvanvirtan: This looks good, but I think we should move the `until_ticks` computation inside the if statement.
Without this change, we will have a conflict with the ARM virtio_mmio PR: https://github.com/unikraft/unikraft/pull/208.
See this commit as a reference: https://github.com/unikraft/unikraft/pull/208/commits/de3da1672cd476136f32ae8db061da7b96b86476.

plat/common/lcpu.c
```diff
@@ -48,7 +48,7 @@ void ukplat_lcpu_halt_irq(void)
 	ukplat_lcpu_disable_irq();
 }
 
-void ukplat_lcpu_halt_to(__snsec until)
+void ukplat_lcpu_halt_to(__nsec until)
```
razvanvirtan, line 43: Can you update your working branch according to the latest `staging`, please? Otherwise, we will get a conflict here with this PR, that has been already merged: https://github.com/unikraft/unikraft/pull/257/commits/210ff11d7bac31801359a57ce010639fd0f6eeff

plat/common/arm/time.c
```diff
@@ -149,3 +153,8 @@ void ukplat_time_init(void)
 	/* Enable timer */
 	generic_timer_enable();
 }
+
+uint32_t ukplat_time_get_irq(void)
```
razvanvirtan, line 162: I think we should add a `ukplat_time_get_irq` implementation  in `plat/kvm/x86/time.c` too. At this moment, the build process fails for x86 images at linking, because this function is called in ` plat/kvm/irq.c`, but not defined. Something like this should work for x86_64 (I've tested it on my local):
```
uint32_t ukplat_time_get_irq(void)
{
	return 0;
}
```

razvanvirtan (changes requested):

plat/common/arm/time.c
rene: Done.

plat/common/lcpu.c
rene: Done.

plat/common/arm/generic_timer.c
rene: ok, it's done.

razvanvirtan (approved):

razvanvirtan:
All good from my side.

PR #176, opened 2021-03-31
lib/ubsan: Add Undefined Behavior Sanitizer Library

vladandrew:
This patch adds an internal library that implements the functions needed
by UBSAN. The flag can either be enabled globally by enabling the
LIBUBSAN_GLOBAL config option or manually by adding the
-fsanitize=undefined flag. Our current implementation
crashes when undefined behavior is detected.

UBSAN can catch runtime bugs such as dereferencing NULL or non-canonical
addresses, certain undefined overflow errors, shifting or multiplying
data which is out of bounds, and other errors.

For example, ubsan will catch the following addition overflow:
int k = 0x7fffffff;
k += 127;

Changes since v1:
* Added type_mismatch_info_v1
* New implementation of __ubsan_handle_type_mismatch_v1 that uses type_mismatch_info_v1
* Added ubsan_overflow_pointer_data and fixed __ubsan_handle_pointer_overflow
* Updated __ubsan_handle_out_of_bounds output message
* Added a TODO for missing prints
* Added index_raw to __ubsan_handle_nonnull_arg
* Added __ubsan_handle_nonnull_return_v1 and ubsan_nonnull_return_data_v1


danield20 (approved):
Seems fine. The only thing that I find a bit strange is why do we need another option in the submenu. If we enable LIBUBSAN isn't it already enabled globally?

vladandrew:
Hi Daniel,

We have that option in case someone wants to enable the -fsanitize=undefined per library and not globally.

danield20:
Reviewed-by: Daniel Dinca <dincadaniel97@gmail.com>

PR #178, opened 2021-04-07
[RFC] Spinlock implementation

rene:
Provides a common spinlock implementation for all ARM64 platforms:

- Provide implementation at plat/common/arm
- Functions imported from ARM Trusted Firmware: git@github.com:ARM-software/arm-trusted-firmware.git
- Included on KVM platform 

In platforms with multicore support, spinlock variables shall be placed in a common section to all cores, which is not a problem for SMP implementations where .data/.rodata/.bss segments are shared. However, some implementations might not share these segments, so that's why this RFC proposes a new section called slocks. All spinlock variables shall be placed in this section. Thus, they can be easily shared throughout different SMP implementations.

Signed-off-by: Renê de Souza Pinto Rene.deSouzaPinto@opensynergy.com
CC: @razvand


include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
-typedef struct {} spinlock_t;
+#ifdef CONFIG_UKPLAT_LCPU_MULTICORE
+/** spinlock struct */
+typedef struct {
+	volatile uint32_t lock;
+} __spinlock_t;
+
+/** section attribute cannot be used with pointers */
+typedef __spinlock_t *spinlock_ptr_t;
+
```
marcrittinghaus, line 45: It would be good to add the following macro:
```C
#define UK_SPINLOCK_INITIALIZER() { 0 }
```
and remove the `DEFINE_SPINLOCK` from the public header. This makes the type and name more explicit in definitions and also keeps the standard C-way of how variable definitions look like.

plat/common/arm/spinlock64.S
```diff
@@ -0,0 +1,71 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
marcrittinghaus, line 1: To enable seamless inlining and LTO, it is better to move these implementations into a separate header `spinlock.h` in `arch/arm/arm64/include/uk/asm` with all functions declared as `static inline`, using GCC inline assembly. Do not use macros, because functions make is easier to trace spinlock invocations.

include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
+/**
+ * Release lock previously acquired by spin_lock.
+ * Use store-release to unconditionally clear the spinlock variable.
+ * Store operation generates an event to all cores waiting in WFE
+ * when address is monitored by the global monitor.
+ * @param [in/out] lock Pointer to spinlock object
+ */
+void ukarch_spin_unlock(spinlock_ptr_t lock);
+
+/* Defines a pre-initialized spin_lock in unlocked state */
```
marcrittinghaus, line 72:
```suggestion
/* Defines a pre-initialized spinlock in unlocked state */
```

Also fix in the non-SMP branch.

include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-typedef struct {} spinlock_t;
+#ifdef CONFIG_UKPLAT_LCPU_MULTICORE
+/** spinlock struct */
+typedef struct {
+	volatile uint32_t lock;
+} __spinlock_t;
```
marcrittinghaus, line 41: As this file will be part of the core API, it should not be dependent on `stdint.h`. You can use a simple `int`. The `include` statement can then be removed. For the basic data types, we also would like to remove the `_t` just as `__atomic`.
```suggestion
typedef struct {
	volatile int lock;
} __spinlock;
```

include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
+
+/** section attribute cannot be used with pointers */
+typedef __spinlock_t *spinlock_ptr_t;
```
marcrittinghaus, line 44: We actually prefer pointer types to be explicitly declared as such. So we would prefer to go without this `typedef` and use actual pointers in the function signatures.
```suggestion

```

include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
+
+/* spinlock type for declaration */
+#define spinlock_t __attribute__((section(".slocks"))) __spinlock_t
```
marcrittinghaus, line 47: We prefer to keep spinlocks in the regular data sections, so please remove this define and also the changes to the `sections.h` and `link64.lds.S`.
```suggestion

```

include/uk/arch/spinlock.h
```diff
@@ -21,20 +23,62 @@
...
+
+/**
+ * Initialize a spinlock object (put in unlocked state)
+ * @param [in/out] lock Pointer to spinlock object
+ */
+void ukarch_spin_lock_init(spinlock_ptr_t spinlock);
```
marcrittinghaus, line 53: According to `__atomic` we would rename the functions as

```C
void __spinlock_init(__spinlock *lock);
void __spinlock_lock(__spinlock *lock);
void __spinlock_unlock(__spinlock *lock);
int __spinlock_is_locked(__spinlock *lock);
int __spinlock_trylock(__spinlock *lock);
```

Having the word `spinlock` in it also makes it more intuitive to `grep` for spinlock functions.

marcrittinghaus (changes requested):
Hi Rene,

thank you very much for providing an implementation for the spinlocks! 😃 Sorry, for taking so long to review. I added a few comments.

I think the most notable aspect is storing spinlocks in a dedicated section for easy sharing. Considering the way OpenSynergy currently uses Unikraft, we see this as an elegant solution to the limited SMP support. However, the direction into which we want the project to go is to have all relevant components SMP-safe so that spinlocks can reside in the regular `.data`/`.bss` section which is shared among all cores. Nevertheless, we like the idea of having the option to add the attribute. So a way for OpenSynergy could be to provide an extra OpenSynergy-internal header that declares the removed `DEFINE_SPINLOCK`-type macro. This macro can then include the attribute to store the respective spinlock in the `.slocks` section as part of the variable definition.

```C
#include <uk/essentials.h>
#include <uk/arch/spinlock.h>

#define DEFINE_SPINLOCK(lock)  \
    __spinlock __section(".slocks") lock
```

plat/common/arm/spinlock64.S
michpappas: Would lib/uklock/ be a more suitable place for `spinlocks.h` and its implementation?

marcrittinghaus: We also thought about this and we had a 50/50-feeling when deciding where to put it. At the moment all primitives in `uklock` interact with the scheduler for thread blocking, so `uklock` has a dependency on it. We therefore thought that it might be better to see the spinlock as some kind of basic data type/primitive and allow it in the architectural headers (hence the change in the naming to `__spinlock`).

Of course, we could also make the dependency in `uklock` more differentiated so that the scheduler is only involved if `uk_mutex` or `uk_semaphore` are added per config. But that would create a hard dependency of most low-level SMP-ready code (e.g., platforms) on `uklock`.

What are your thoughts on this?

include/uk/arch/spinlock.h
michpappas: > However, the direction into which we want the project to go is to have all relevant components SMP-safe so that spinlocks can reside in the regular .data/.bss section which is shared among all cores.

Makes sense, and it's the right approach.

> Nevertheless, we like the idea of having the option to add the attribute. So a way for OpenSynergy could be to provide an extra OpenSynergy-internal header that declares the removed DEFINE_SPINLOCK-type macro. This macro can then include the attribute to store the respective spinlock in the .slocks section as part of the variable definition.

Regarding the proposed way to override the default behavior, I can see some challenges: Overriding the macro with our own version does not cover the case where a spinlock is declared as member of another structure; you cannot use the macro there as gcc will complain that you're trying to place one member to a different section than the rest of the structure. For simple declarations of spinlocks as global variables, there is no guarantees that one would use the macro either, but that's the least of the two problems :)

marcrittinghaus: That is a fair point. Don't you have to mark those structures manually, anyway? At least your proposed solution is equivalent in this regard. How do you do it at the moment?

But you are right that the `DEFINE_SPINLOCK` macro is not a good solution in that it would not cover spinlocks that we add upstream and it introduces a potential for mistakes. But at least, as long as you use it internally, your spinlocks would be covered. Of course, you could internally also define a `spinlock_t` type via a macro as your PR suggests.

michpappas: Well, the ideal would be that we somehow replaced the spinlock implementation with ours (perhaps by overriding uklock - see also https://github.com/unikraft/unikraft/discussions/261) and not patch unikraft core at all. I'm not sure if its possible though :)

plat/common/arm/spinlock64.S
michpappas: I think in that case uklock would be a dependency of SMP with our without a scheduler, but since uklock is an internal library does it make a difference?

rene:
Consider the outcome of this RFC PR, we have proposed another approach for spinlock implementation:

- https://github.com/unikraft/unikraft/pull/289

Thus, I'm closing this PR. Appreciate all comments and discussion.

PR #181, opened 2021-04-15
build: Support second download URLs and checksums

skuenzer:
Introduces build system support for
- Alternative download URLs for archives
- Checksum tests for downloaded archives


nderjung (approved):
Hey! Thanks for this! This will ensure the integrity of libraries and is well-received!  I've tested this, and this works as expected.

PR #183, opened 2021-04-16
plat/ukplatkvm: Provide ACPI structures and functionalities

cristian-vijelie:
This series provides functions for finding and parsing some of the ACPI tables, namely RSDP, RSDT and MADT. It also provides structures for storing them. I followed the standard, and so you will see the Microsoft naming style in some places.


plat/Config.uk
```diff
@@ -12,6 +12,13 @@ config EARLY_PRINT_PL011_UART_ADDR
 	help
 		Pl011 serial address used by early debug console.
 
+config UKPLAT_ACPI
+	bool "Support for using the ACPI tables"
```
skuenzer: You may implicitly just integrate ACPI code when needed. I would not add an extra menu item for it.

plat/kvm/Makefile.uk
```diff
@@ -64,6 +64,9 @@ LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/intctrl.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/tscclock.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/time.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/memory.c|x86
+ifeq ($(CONFIG_UKPLAT_ACPI),y)
```
skuenzer: You could configure it depending on a feature flag, like `CONFIG_HAVE_SMP`

plat/kvm/x86/setup.c
```diff
@@ -287,6 +292,13 @@ void _libkvmplat_entry(void *arg)
 	uk_pr_info("     stack top: %p\n",
 		   (void *) _libkvmplat_cfg.bstack.start);
 
+#ifdef CONFIG_UKPLAT_ACPI
+	find_rsdt();
+#ifdef CONFIG_LIST_ACPI
```
skuenzer: I would list it as part of find rsdt and you could configure the code in only when `UK_DEBUG` is set. This way we do not compile anything when debugging is off

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+/*
+ * Find the ACPI Root System Descriptor Table (RSDT), and check if it's valid.
+ * For this, the Root System Descriptor Pointer (RSDP) must be found, in the
+ * physical memory area 0xe0000 -> 0xfffff.
+ */
+
+void find_rsdt(void)
+{
+	__u64 start_addr = 0xe0000, end_addr = 0xfffff;
+	__u64 i;
```
skuenzer: Because you use it as integer storing an address, I recommend to use `__uptr` for `start_addr`, `end_addr` and `i`. This type automatically adopts between 32bit and 64but addressing mode.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+#include <x86/acpi/madt.h>
+
+__u8 acpi_version;
+struct RSDPDescriptor *rsdp;
+struct RSDPDescriptor20 *rsdp2;
+struct RSDT *rsdt;
+struct XSDT *xsdt;
+struct MADT *madt;
+
+static inline __u8 detect_acpi_version(__u64 *ptr)
```
skuenzer: Is there a reason for `__u64 *` for the pointer? You could just use `void *` since you anyway cast it

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+{
+	struct RSDPDescriptor *rsdp = (struct RSDPDescriptor *)ptr;
```
skuenzer: Please also add a "is not-NULL" check:
``` c
UK_ASSERT(ptr);
```
our assertions are automatically removed on productions builds. They should be used to check for programming errors or forbidden API usage.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+		if (!memcmp(ptr, "RSD PTR ", 8)) {
+			uk_pr_info("Found RSDP at %p\n", ptr);
+			break;
+		}
+	}
+
+	uk_pr_info("ACPI version detected: ");
+	if (!detect_acpi_version(ptr)) {
+		uk_pr_info("1.0\n");
+		if (!rsdp_checksum(ptr)) {
```
skuenzer: Same here - I read it as if the checksum is not there or invalid, instead that it is valid:
``` c
if (rdsp_checksum(ptr) == 0) {
```

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+	for (i = start_addr; i < end_addr; i += 16) {
+		ptr = (__u64 *)i;
+		if (!memcmp(ptr, "RSD PTR ", 8)) {
+			uk_pr_info("Found RSDP at %p\n", ptr);
+			break;
+		}
+	}
+
+	uk_pr_info("ACPI version detected: ");
+	if (!detect_acpi_version(ptr)) {
```
skuenzer: This if condition I would change to explicitly compare with a value, otherwise you read if the ACPI version is not detected then we assume version 1.0. In general I prefer if conditions with the `!` operator to check that something is not given or failed (e.g., NULL often means something is not given) but here we expect a certain number.
``` c
if (detect_acpi_version() == 0) {
}
```

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+				(struct ACPISDTHeader *)rsdp->RsdtAddress)) {
+				rsdt = (struct RSDT *)rsdp->RsdtAddress;
+				uk_pr_info("RSDT is valid and present at %p\n",
+					   rsdt);
+			} else {
+				uk_pr_err("RSDT is not valid\n");
+				return;
+			}
+		} else {
+			uk_pr_err("RSDP is not valid\n");
```
skuenzer: This looks to me that you could flatten the nested `if` statements when you exit (or `goto` to an error exit label) when you check for failing cases, e.g.:
```  c
if (rsdp_checksum(ptr) != 0)
   uk_pr_err("Invalid RSDP\n");   
   return;
}

acpi_version = 1;
rsdp = (struct RSDPDescriptor *) ptr;
rsdt = (struct RSDT *) rsdp->RsdtAddress;

if (acpi_checksum(rsdt) != 0) {
   uk_pr_err("Invalid RSDT\n");
   return;
}

uk_pr_info("RSDT present at %p\n", rsdt);
```

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+			return;
+		}
+	} else {
+		uk_pr_info("> 2");
+		if (!rsdp2_checksum(ptr)) {
```
skuenzer: In general, I would remove code that we do not support yet since you could not test it. Exit with an error message instead that you found an RDSP version that is currenlty unsupported or not implemented and continue booting without ACPI. With a `/* TODO */` comment you can describe what needs to be done to support it.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+	     i++)
+		checksum += *i;
+
+	return (__u8)checksum == 0 ? 0 : -1;
+}
+
+/*
+ * Not yet implemented, because I have not yet detected ACPI version 2 or
+ * greater.
+ */
```
skuenzer, line 49: Does `qemu-system` has an option to change the ACPI version so that you could test it? Maybe `-machine q35` does it? I am not sure...

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,182 @@
...
+		if (!memcmp(h->Signature, "APIC", 4)) {
+			uk_pr_info("Found MADT at %p. ", h);
+			if (!acpi_checksum(h)) {
+				uk_pr_info("The table is valid\n");
+				madt = (struct MADT *)h;
+			} else
+				uk_pr_info("The table is not valid\n");
+			break;
+		}
+	}
```
skuenzer, line 187: How about the following flattened loop?
``` c
for (i = 0; i < entries; i++) {
	h = (struct ACPISDTHeader *)rsdt->PointerToOtherSDT[i];
	if (memcmp(h->Signature, "APIC", 4) != 0)
		continue; /* Not an APIC entry */

	if (acpi_checksum(h) != 0) {
		uk_pr_err("Unsupported or invalid MADT at %p\n", h)
 		break;
	}

	uk_pr_info("MADT present at %p\n", h);
	break;
}
```

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,78 @@
...
+	__u16 Flags;
+	__u8 LINT;
+};
+
+struct MADTType5Entry {
+	struct MADTEntryHeader eh;
+	__u16 Reserved;
+	__u64 LAPICOverride;
+};
+
```
skuenzer: Is it maybe safer to add the packed attribute to these structs? You can check out `<uk/essentials.h>`

plat/common/include/x86/acpi/rsdt.h
```diff
@@ -0,0 +1,37 @@
...
+	char Signature[4];
+	__u32 Length;
+	__u8 Revision;
+	__u8 Checksum;
+	char OEMID[6];
+	char OEMTableID[8];
+	__u32 OEMRevision;
+	__u32 CreatorID;
+	__u32 CreatorRevision;
+};
```
skuenzer: packed?

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,78 @@
...
+void find_madt(void);
```
skuenzer: Maybe add a description comment for this function (e.g., that is searches within the first bytes of the address space). In general: Do you want to return a status code to indicate if you found it?

skuenzer (changes requested):
I think this patch series looks really good. I have minor comments to the code style but this is a really minimal ACPI starting set which is perfect for what we want to achieve.

plat/kvm/x86/setup.c
cristian-vijelie: It's not a debug function. RSDT must be found, in order for the rest of the components to work properly

plat/kvm/x86/acpi.c
cristian-vijelie: https://stackoverflow.com/questions/60137506/qemu-support-for-acpi-2-0

cristian-vijelie: The -acpitables option is very weird, complicated, and badly documented. Maybe I'll try after I finish with the SMP.

danbrodsky:
Later when the ptsupport patch is merged (https://github.com/unikraft/eurosys21-artifacts/blob/master/support/patches-unikraft-eurosys21/ptsupport.patch) enabling `CONFIG_PT` will disable static pagetables and cause reads of RSDT to segfault. It would be good to add a change to that patch before it's merged to map regions conditionally if `CONFIG_SMP` is set:
0xe000 - 0xffff for BIOS memory
0x3fe0000 - 0x3fe3000 for RSDT area
0xfee00000 - 0xfee01000 for LAPIC
0x0 - 0x1000 for initial AP trampoline after SIPI
0x8000 - 0x9000 for AP trampoline area (or whichever address is used for the trampoline)

Adding this near `__mmap_kvm_areas` would probably be ideal.

plat/common/include/x86/acpi/acpi.h
```diff
@@ -0,0 +1,69 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#ifndef __ACPI_H__
+#define __ACPI_H__
```
marcrittinghaus: Use the convention for include guards set by the other headers in the common platform code:
```C
#ifndef __PLAT_CMN_X86_ACPI_H__
#define __PLAT_CMN_X86_ACPI_H__
```
Don't forget to match the comment at the end of the file for the `#endif /* __PLAT_CMN_X86_ACPI_H__ */`.

This also applies to the other headers.

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+
+/* Multiprocessor Wakeup Structure */
+struct __packed MADTType10Entry {
+	struct MADTEntryHeader eh;
+	__u16 MailBoxVersion;
+	__u32 Reserved;
+	__u64 MailBoxAddress;
+};
+
+struct MADT *madt;
```
marcrittinghaus: Missing `extern` keyword, if this needs to be exported in the first place. See comment on using `static` keyword in `acpi.c`

```C
extern struct MADT *madt;
```

plat/common/include/x86/acpi/rsdt.h
```diff
@@ -0,0 +1,60 @@
...
+	__u8 Revision;
+	__u8 Checksum;
+	char OEMID[6];
+	char OEMTableID[8];
+	__u32 OEMRevision;
+	__u32 CreatorID;
+	__u32 CreatorRevision;
+} __packed;
+
+
```
marcrittinghaus: Remove empty line

plat/kvm/x86/acpi.c
marcrittinghaus: The same applies to the other functions that receive a pointer.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+	__uptr i;
+	__u64 *ptr;
+
+	for (i = start_addr; i < end_addr; i += 16) {
+		ptr = (__u64 *)i;
+		if (!memcmp(ptr, "RSD PTR ", 8)) {
+			uk_pr_info("Found RSDP at %p\n", ptr);
+			break;
+		}
+	}
```
marcrittinghaus: This is a runtime operation whose outcome depends on undefined input. You should expect this operation to fail and handle this case gracefully. That means, you have to detect if you actually found the RSDP signature and bail out with `return -1` if not. You current implementation will crash the system with the following pointer dereference in `detect_acpi_version`.

I would change the `uk_pr_info` to a `uk_pr_debug`. Printing such detailed information just slows down boot time and does not provide any additional value in a production build.

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+	struct ACPISDTHeader h;
+	__u32 LocalAPICAddress;
+	__u32 Flags;
+	__u8 Entries[];
+};
+
+struct __packed MADTEntryHeader {
+	__u8 Type;
+	__u8 Length;
+};
```
marcrittinghaus: By convention attributes are supplied after the actual struct body, e.g.:

```C
struct MADTEntryHeader {
	__u8 Type;
	__u8 Length;
} __packed;
```
This applies to all structure declarations.

plat/common/include/x86/acpi/rsdt.h
```diff
@@ -0,0 +1,60 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#ifndef __RSDT_H__
+#define __RSDT_H__
+
```
skuenzer: You should include the data type definitons within this header. Because you use `__packed` you also need `essentials.h`
```c
#include <uk/arch/types.h>
#include <uk/essentials.h>
```

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#ifndef __MADT_H__
+#define __MADT_H__
+
+#include <uk/arch/types.h>
+#include <x86/acpi/rsdt.h>
```
skuenzer: I would also include `<uk/essentials.h>` becuase of the `__packed` attribute usage.

plat/common/include/x86/acpi/acpi.h
```diff
@@ -0,0 +1,69 @@
...
+	__u64 XsdtAddress;
+	__u8 ExtendedChecksum;
+	__u8 reserved[3];
+} __packed;
+
+/*
+ * Find the ACPI Root System Descriptor Table (RSDT), and check if it's valid.
+ * For this, the Root System Descriptor Pointer (RSDP) must be found, in the
+ * physical memory area 0xe0000 -> 0xfffff.
+ */
```
skuenzer, line 65: Can you document what the function returns? Is there a reason for `__s8`? Why not `int`? `int` would be more inline with our other functions and you would more expect success/error when seeing `int` at a function prototype.

plat/common/include/x86/acpi/acpi.h
```diff
@@ -0,0 +1,69 @@
...
+__s8 find_rsdt(void);
+
+/*
+ * Print the detected ACPI tables, for debug.
```
skuenzer: Maybe also say that you print to debug output or actually do not export this function because you print the tables anyways when debugging is enabled.

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+
+struct MADT {
```
skuenzer, line 41: I guess this struct needs also be packed?

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+};
+
+struct MADT *madt;
+
+/*
+ * Find the Multiple APIC Descriptor Table (MADT), in the RSDT, and check if
+ * it's valid. MADT can be found by searching for the string "APIC", in the
+ * first 4 bytes of each table entry
+ */
+__s8 find_madt(void);
```
skuenzer: `int` as return code? Please document also the return values.

plat/kvm/Makefile.uk
```diff
@@ -64,6 +64,9 @@ LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/intctrl.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/tscclock.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/time.c
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/memory.c|x86
+ifeq ($(CONFIG_UKPLAT_SMP),y)
```
skuenzer: Can you use `HAVE_SMP`? This would be inline with the other feature flags that we have: HAVE_SYSCALL, etc. You may also want to update `include/uk/plat/lcpu.h` accordingly.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+
+	return 0;
+}
+
+/*
+ * Print the detected ACPI tables, for debug.
+ */
+
+void list_acpi_tables(void)
+{
```
skuenzer, line 138: When you want to keep it exported with the header (which I think you don't need) you can optimize this function by guarding its body with:
```c
#if UK_DEBUG
```
Otherwise I would declare this function `static`.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+	int entries = (rsdt->h.Length - sizeof(rsdt->h)) / 4;
+	int i;
+	struct ACPISDTHeader *h;
+
+	uk_pr_debug("%d ACPI tables found\n", entries);
+
+	for (i = 0; i < entries; i++) {
+		h = (struct ACPISDTHeader *)rsdt->PointerToOtherSDT[i];
+		uk_pr_debug("%p: %.4s\n", h, h->Signature);
+	}
```
skuenzer, line 150:
```c
#endif
```

plat/common/include/x86/acpi/acpi.h
```diff
@@ -0,0 +1,69 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#ifndef __ACPI_H__
+#define __ACPI_H__
+
+#include <x86/acpi/rsdt.h>
+#include <x86/acpi/madt.h>
```
marcrittinghaus, line 38: As you are using the `__packed` attribute defined in `<uk/essentials.h>` you should add the include.

```C
#include <uk/essentials.h>
```

This also applies to the other header files, where needed includes are missing. Generally, all headers should be self-contained in their includes and do not depend on other headers being manually included in the target C file for your header to work.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+
+	uk_pr_info("ACPI version detected: ");
+	if (detect_acpi_version(ptr) == 0) {
+		uk_pr_info("1.0\n");
+		if (rsdp_checksum(ptr) != 0) {
+			uk_pr_err("Invalid RSDP\n");
+			return -1;
+		}
```
marcrittinghaus: The name `detect_acpi_version` suggests that the function does more than what it actually does. In fact, I would prefer to see the checksum verification prior to using the revision value.

The ACPI specification states that the first 20 bytes of the table that are covered by the checksum are backward compatible. So you can move the checksum test into `detect_acpi_version` and first compute the checksum and if it matches return the proper ACPI revision. You may return -1 in case no valid RSDP structure has been found.

Adjust the log output accordingly. That means only output the info on which ACPI version you detected if the checksum has been verified.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+		rsdp = (struct RSDPDescriptor *)ptr;
+		rsdt = (struct RSDT *)rsdp->RsdtAddress;
+
+		if (acpi_checksum(&rsdt->h) != 0) {
+			uk_pr_err("Invalid RSDT\n");
+			rsdt = NULL;
+			return -1;
+		}
+
+		uk_pr_info("RSDT present at %p\n", rsdt);
```
marcrittinghaus: Change this to a `uk_pr_dbg` output. Use the same general message for this and the previous RSDP message. This makes interpreting log messages much easier.

plat/common/include/x86/acpi/madt.h
```diff
@@ -0,0 +1,230 @@
...
+	__u32 Flags;
+	__u8 Entries[];
+};
+
+struct __packed MADTEntryHeader {
+	__u8 Type;
+	__u8 Length;
+};
+
+/* I included all possible types of entries here, so you won't have to. */
```
skuenzer: I would leave this comment out.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+#include <uk/assert.h>
+
+__u8 acpi_version;
+struct RSDPDescriptor *rsdp;
+struct RSDPDescriptor20 *rsdp2;
+struct RSDT *rsdt;
+struct XSDT *xsdt;
+struct MADT *madt;
+
+static inline __u8 detect_acpi_version(void *ptr)
```
skuenzer: I would suggest that you handover `struct RSDPDescriptor *ptr` instead of `void *`. It makes the function prototype clearer on what you need to handover.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+{
+	struct RSDPDescriptor *rsdp = (struct RSDPDescriptor *)ptr;
+
+	UK_ASSERT(ptr);
+
+	return rsdp->Revision;
+}
+
+static inline __s8 rsdp_checksum(__u64 *ptr)
```
skuenzer: With your checksum function I was wondering if code gets easier to understand if you name them something with `test_` or `_isvalid`. This sounds less that this function computes a checksum and returns you this one. It sounds that you are checking the validity. I would also use `int` as return type. Effectively you don't save anything by using `__s8`.
I would also prefer if function arguments are already in the target type instead of casting it within. It simplifies the API.

```c
static inline int test_rdsp_checksum(struct RSDPDescriptor *rdsp)
{
   /* ... */
}
```

plat/kvm/x86/acpi.c
skuenzer: Okay, in such a case, please remove the rdsp2_checksum stub.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+static inline __s8 rsdp2_checksum(__u64 *ptr)
+{
+	return 0;
+}
+
+/*
+ * Do checksum for any ACPI table, except RSDP.
+ */
+
+static inline __s8 acpi_checksum(struct ACPISDTHeader *h)
```
skuenzer:
```c
static inline int test_acpi_checksum(struct ACPISDHeader *h) {}
/* or */
static inline int acpi_checksum_isvalid(struct ACPISDHeader *h) {}
```

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <uk/print.h>
+#include <string.h>
+#include <x86/acpi/acpi.h>
+#include <x86/acpi/madt.h>
+#include <uk/assert.h>
+
```
skuenzer, line 39: I would declare all of them `static`. These are internal variables, I don't expect that we should access them directly - right? Make sure that the function calls that need one of these filled do an `UK_ASSERT` to test that the initialization function was executed before (like `find_rsdt()`).

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+	return checksum == 0 ? 0 : -1;
+}
+
+/*
+ * Find the ACPI Root System Descriptor Table (RSDT), and check if it's valid.
+ * For this, the Root System Descriptor Pointer (RSDP) must be found, in the
+ * physical memory area 0xe0000 -> 0xfffff.
+ */
+
+__s8 find_rsdt(void)
```
skuenzer: I would prefer returning the result as `int`.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+		 * TODO: add support for ACPI version 2 or greater, when/ if it
+		 * will be detected
+		 */
+
+		uk_pr_info(">= 2\n");
+		uk_pr_err("ACPI version not yet supported!\n");
+
+		return -1;
+	}
+#ifdef UK_DEBUG
```
skuenzer: Think about if you want to keep `list_acpi_tables` exported or not. I would prefer not.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+
+/*
+ * Find the Multiple APIC Descriptor Table (MADT), in the RSDT, and check if
+ * it's valid. MADT can be found by searching for the string "APIC", in the
+ * first 4 bytes of each table entry
+ */
+
+__s8 find_madt(void)
+{
+	if (!rsdt)
```
skuenzer: Maybe turn this into an `UK_ASSERT`. I consider this case as an API mis-use. What do you think?

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+/*
+ * Find the Multiple APIC Descriptor Table (MADT), in the RSDT, and check if
+ * it's valid. MADT can be found by searching for the string "APIC", in the
+ * first 4 bytes of each table entry
+ */
+
+__s8 find_madt(void)
```
skuenzer: `int`

plat/kvm/x86/pagetable.S
```diff
@@ -41,7 +43,8 @@ cpu_zeropt:
 	   0x09000 - 0x09fff -> multiboot info @ 0x09500 (read-only)
 	   0xb8000 - 0xbffff -> VGA buffer (read+write)
```
skuenzer, line 42: Can you extend this comment and add information about your mapped ACPI range?

plat/kvm/x86/setup.c
```diff
@@ -287,6 +292,10 @@ void _libkvmplat_entry(void *arg)
 	uk_pr_info("     stack top: %p\n",
 		   (void *) _libkvmplat_cfg.bstack.start);
 
+#ifdef CONFIG_UKPLAT_SMP
```
skuenzer: Try to use `CONFIG_HAVE_SMP`. The other name was introduced earlier where we did not had this naming scheme with `HAVE_*` for platform features.

skuenzer (changes requested):
Hi Christian, thanks a lot for your work! I put some comments to your implementation. I tried considering this as a library that multiple platforms could include to enable ACPI support. This is why I checked on the API part a bit more. I guess, later, the SMP commits will add functions to this library that return the number of cores, etc. Right?

Thanks
Simon

plat/common/include/x86/acpi/acpi.h
marcrittinghaus: I would not export it.

plat/common/include/x86/acpi/acpi.h
marcrittinghaus: I agree, I don't see a reason here to return an `__s8` instead of the standard `int`, too.

Btw: Although this is not a concern in this case, note that using non-standard data types can make the compiled code slower. For example, a signed 8-byte has to be sign-extended to 32-bit if the result value is (implicitly) cast to int by C introducing an extra instruction after the call:
```ASM
107cef:	e8 03 0d 00 00       	call   1089f7 <find_rsdt>
107cf4:	0f be c0             	movsx  eax,al
``` 

You can change the comment to include information on the return value like so:
```C
/**
 * ...
 *
 * @return X
 */
 ```

The same applies to `find_madt()`.

File: plat/common/include/x86/acpi/madt.h (hunk 33 above)
Comment (marcrittinghaus) [reply plat/common/include/x86/acpi/madt.h]: Agree, remove. Such comments are better placed in the commit message that explains why you have done the change, i.e., include more structures than currently needed?!

Generally, keep your comments formal. Use _we_ instead of _I_, or use passive. In this case, a more suitable comment could, for example, look like
```C
/* 
 * The following structures are declared according to ACPI X.X.
 * [Source] <- Long-living reference, i.e., prefer document title and version over URL
 */
 ```

File: plat/kvm/x86/acpi.c (hunk 35 above)
Comment (marcrittinghaus) [reply plat/kvm/x86/acpi.c]: I agree. Since the function not only computes the checksum but also verifies it, I also suggest to use a clearer name that expresses this circumstance. I would use the term **verify**, though.

With this being said, I would rewrite `rsdp_checksum` like this:
```C
static inline int verify_rsdp_checksum(struct RSDPDescriptor *rdsp)
{
	__u8 checksum = 0;
	__u8 *ptr = (__u8 *)rdsp;

	UK_ASSERT(rdsp);

	while (ptr < (__u8 *)(rdsp + 1))
		checksum += *ptr++;

	return (checksum == 0) ? 0 : -1;
}
```

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+ * Find the ACPI Root System Descriptor Table (RSDT), and check if it's valid.
+ * For this, the Root System Descriptor Pointer (RSDP) must be found, in the
+ * physical memory area 0xe0000 -> 0xfffff.
+ */
+
+__s8 find_rsdt(void)
+{
+	__uptr start_addr = 0xe0000, end_addr = 0xfffff;
+	__uptr i;
+	__u64 *ptr;
```
marcrittinghaus: You could use a `__u8 *` to be more inline with the other code in this file as there is no specific reason to choose a pointer to a 64-bit data type.

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,202 @@
...
+	if (detect_acpi_version(ptr) == 0) {
+		uk_pr_info("1.0\n");
+		if (rsdp_checksum(ptr) != 0) {
+			uk_pr_err("Invalid RSDP\n");
+			return -1;
+		}
+
+		acpi_version = 1;
+		rsdp = (struct RSDPDescriptor *)ptr;
+		rsdt = (struct RSDT *)rsdp->RsdtAddress;
```
marcrittinghaus: This will throw `warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]`.

As I do not see a technical problem here, you can just use an intermediate cast to `__uptr` like so:
```C
rsdt = (struct RSDT *)((__uptr)rsdp->RsdtAddress);
```

That applies to all locations, where you directly cast an integer from one of the data structures to a pointer.

Also consider to set the global variables only (specifically `acpi_version`) if everything looks good. This way a single check for example with `acpi_version > 0` would allow to detect if ACPI initialized successfully. Otherwise, we have to check all data structures for correct initialization.

plat/kvm/x86/setup.c
marcrittinghaus: Does the build system automatically set `CONFIG_HAVE_SMP`? Did not find the option to enable or disable it in menuconfig.

plat/common/include/x86/acpi/madt.h
cristian-vijelie: I thought that, because of the variable-length Entries, it should not be packed.

plat/kvm/x86/acpi.c
marcrittinghaus: Agree. Would guard it and make it static (i.e., not export it).

Concerning coherence with the coding style, I also suggest to remove the `rsdt` test as commented by Simon for `find_madt` and put the variable definitions at the top. Then add an `UK_ASSERT` and after that compute the number of entries. The same applies to `find_madt`.

```C
#ifdef UK_DEBUG
void list_acpi_tables(void)
{
	int i, entries;
	struct ACPISDTHeader *h;

	UK_ASSERT(rsdt);
 	entries = (rsdt->h.Length - sizeof(rsdt->h)) / 4;

	uk_pr_debug("%d ACPI tables found\n", entries);

	for (i = 0; i < entries; i++) {
		h = (struct ACPISDTHeader *)rsdt->PointerToOtherSDT[i];
		uk_pr_debug("%p: %.4s\n", h, h->Signature);
	}
}
#endif
```

Also consider to use a macro for the computation of the number of entries, since you use it also in `find_madt`.

cristian-vijelie:
> I guess, later, the SMP commits will add functions to this library that return the number of cores, etc. Right?

Yes, that's part of the SMP API, as we discussed. This part only provides the ACPI tables to the user.

plat/kvm/x86/setup.c
```diff
@@ -40,6 +42,9 @@
 #include <uk/plat/console.h>
 #include <uk/assert.h>
 #include <uk/essentials.h>
+#include <x86/acpi/acpi.h>
+
+#include <uk/plat/lcpu.h>
```
marcrittinghaus: Why is this additional include required?

plat/kvm/x86/acpi.c
```diff
@@ -0,0 +1,191 @@
...
+static inline __u8 detect_acpi_version(void *ptr)
+{
+	struct RSDPDescriptor *rsdp = (struct RSDPDescriptor *)ptr;
+
+	UK_ASSERT(ptr);
+
+	return rsdp->Revision;
+}
+
+static inline int rsdp_checksum(__u64 *ptr)
```
marcrittinghaus: Somehow, Simon comments got lost in my github for this function. So in case my addition is missing. I agree with Simon, but would use the term `verify`. I would rewrite the function as:
```C
static inline int verify_rsdp_checksum(struct RSDPDescriptor *rdsp)
{
	__u8 checksum = 0;
	__u8 *ptr = (__u8 *)rdsp;

	UK_ASSERT(rdsp);

	while (ptr < (__u8 *)(rdsp + 1))
		checksum += *ptr++;

	return (checksum == 0) ? 0 : -1;
}
```

marcrittinghaus (commented):
Thank you very much for your effort! I have added some additional remarks.

plat/kvm/x86/setup.c
cristian-vijelie: Leftover from the first versions. Removed it

marcrittinghaus (approved):
Everything looks fine now. Thank you very much for you hard work 👍🏻 We really appreciate it!

PR #184, opened 2021-04-16
plat/kvm: Provide udelay and mdelay implementations

cristian-vijelie:
Delay functions, inspired from the Linux kernel.


include/uk/plat/lcpu.h
```diff
@@ -99,6 +101,9 @@ void ukplat_lcpu_halt_to(__snsec until);
  */
 void ukplat_lcpu_halt_irq(void);
 
+void uk_udelay(__u16 usec);
+void uk_mdelay(__u16 msec);
```
skuenzer: Since this is a purely KVM platform internal function, you can remove `uk_` prefix. We will change the build system sooner or later in order to mask platform symbols towards a common API (issue #200).

skuenzer (changes requested):

marcrittinghaus:
For a function to be placed in the generic `include/uk/plat/` it has to be really general and applicable to all architectures. Now, we do not have such instructions with deterministic delay on each architecture and especially you do not supply an implementation for all currently supported architectures. That is the reason why Simon and I would rather put this somewhere to the x86 platform code.

Then there we have to move it to kvm, because you cannot do the out operation in linuxu and probably don't need it in Xen due to paravirtualization and different SMP start up. This leaves us with putting the function declarations in a new header in `plat/kvm/kvm-x86`. The question is, where could we potentially use these delay functions besides in your SMP code. Some drivers could perhaps make use of it, but on the other side, we do not have come across this situation, yet. Nevertheless, we can make a `delay.h` in the mentioned directory and put the functions there.

What we have to consider is that the function signatures suggest a precisely configurable delay. But, in fact, it's all more like "about" X usecs. This is ok for the SMP startup, but may not be in other cases. So this needs to be documented in the function comment together with a hint that these functions can be used to wait for short durations in places where setting up a timer is not worth the effort or cannot be safely done.

You have done a good job in cleaning up the code :thumbsup:

It might be an option to just supply a header file with two static inline functions and move the `outb` into the `udelay` function. The advantage would be that the functions can be easily inlined.

marcrittinghaus (approved):
Thank you Cristian for this patch! Looks all fine 👍🏻 

PR #186, opened 2021-04-18
support/build: Extend preprules for external language support

vladandrew:
We extend the build system to load preparation rules from external
libraries. This enables Unikraft to support preparation rules for
different formats without modifying the core Makefile.rules.

Usage: If we want to add a new external preparation rule, we create
in the root of the library a  Makefile.rules. Inside Makefile.rules
we define a preprule_{format}.

The preprule should be similar to the preprule_m4 defined in
support/build/Makefile.rules.

### DIFF
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -647,9 +647,10 @@ endef
 #
 # preprule $libname,$source,$preptarget,$extraflags(optional)
 define preprule =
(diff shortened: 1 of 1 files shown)


skuenzer:
This feature is covered by patch series [1604](https://patchwork.unikraft.org/project/unikraft/list/?series=1604&state=*) on patchwork.

PR #138, opened 2021-02-23
Skuenzer/ukthreadcb

skuenzer:
Utilizes `uksched` callback mechanism to create and delete pthread-embedded meta data for threads. This enables using pthread-embedded API calls on threads that were created with the native `uksched` API.

Changes v2:
------------
- Rebased to latest staging
- Register init/uninit functions of `uksignal`

PR #188, opened 2021-04-22
plat/common: Fix struct name in `sw_ctx_size()`

skuenzer:
By mistake, a wrong struct name (`ukplat_ctx`) was used instead
of `sw_ctx` in `sw_ctx_size()`. The issue was introduced with
commit 9ca0659f6 and caused a Unikraft build to fail.

Signed-off-by: Simon Kuenzer <simon.kuenzer@neclab.eu>


nderjung (approved):
Thanks for fixing the bug introduced in #138.

Builds are passing again! :D

PR #192, opened 2021-04-26
Provide additional interrupt context safe functions

cristian-vijelie:
The following patches provide isr-safe functions for uklock and ukmpi libraries. The patches were initially on patchwork:

https://patchwork.unikraft.org/patch/738373/
https://patchwork.unikraft.org/patch/738372/
https://patchwork.unikraft.org/patch/738371/


lib/uklock/include/uk/isr/mutex.h
```diff
@@ -0,0 +1,3 @@
+#include <uk/mutex.h>
+
+#define uk_mutex_trylock_isr(x) uk_mutex_trylock(x)
```
skuenzer, line 8: It probably helps to add a `/* NOTE: ... */` comment in this header that says that `uk_mutex_trylock()` is a static inline function and this is why we can just provide an alias: The implementation of `uk_mutex_trylock_isr()` becomes part of the compilation unit where it is used. Because of this it will be compiled with the same compile flags as the unit.

lib/uklock/include/uk/isr/semaphore.h
```diff
@@ -0,0 +1,3 @@
+#include <uk/semaphore.h>
+
+#define uk_semaphore_down_try_isr(x) uk_semaphore_down_try(x)
```
skuenzer, line 8: I would at a similar comment here as suggested for the mutex function.

lib/ukmpi/Makefile.uk
```diff
@@ -4,3 +4,4 @@ CINCLUDES-$(CONFIG_LIBUKMPI)   += -I$(LIBUKMPI_BASE)/include
 CXXINCLUDES-$(CONFIG_LIBUKMPI) += -I$(LIBUKMPI_BASE)/include
 
 LIBUKMPI_SRCS-$(CONFIG_LIBUKMPI_MBOX) += $(LIBUKMPI_BASE)/mbox.c
+LIBUKMPI_SRCS-$(CONFIG_LIBUKMPI_MBOX) += $(LIBUKMPI_BASE)/isr.c|isr
```
skuenzer: I would suggest to not call the file `isr.c` because it implements the mbox components as isr-safe version. What do you think of `mbox_isr.c`?

lib/ukmpi/isr.c
```diff
@@ -0,0 +1,74 @@
+#include <uk/mbox.h>
+#include <uk/isr/mbox.h>
+
+struct uk_mbox {
```
skuenzer: In order to avoid duplicates, it would be much better to place the struct definition in a shared and library-internal header that is used by `mbox.c` and `mbox_isr.c`. You can place it next to the source files, e.g.,
```c
#include "mbox_defs.h"
```

skuenzer: Any duplicate function (like `_do_mbox_recv()` could be placed as `static inline` function within this header, too.

skuenzer (changes requested):
Hi Christian,
thanks for this work! I think, we can upstream this after adding some comments and descriptions, e.g., could you add one or two sentences to the commit message? This is to make the implementation choice more transparent.
Thanks,
Simon

cristian-vijelie:
@skuenzer, I made the changes you requested

lib/ukmpi/mbox_isr.c
```diff
@@ -0,0 +1,28 @@
...
+#include <uk/isr/mbox.h>
+#include "mbox_defs.h"
+
+int uk_mbox_recv_try_isr(struct uk_mbox *m, void **msg)
+{
+	void *rmsg;
+
+	UK_ASSERT(m);
+
+	if (!uk_semaphore_down_try(&m->readsem))
```
skuenzer: It would be better to use the ISR version of `semaphore_down_try`.
Can you include `<uk/isr/semaphore.h>` and use the isr definiton?

lib/ukmpi/mbox_defs.h
```diff
@@ -0,0 +1,61 @@
+#ifndef __MBOX_DEFS_H__
+#define __MBOX_DEFS_H__
+
+#include <stddef.h>
+#include <uk/semaphore.h>
+
```
skuenzer, line 6: Maybe add also a note here that says that the definitions below are included by ISR-safe and normal compilation units. Moving an inline function out from here requires special care.

lib/ukmpi/mbox_isr.c
```diff
@@ -0,0 +1,28 @@
...
+	if (msg)
+		*msg = rmsg;
+	return 0;
+}
+
+int uk_mbox_post_try_isr(struct uk_mbox *m, void *msg)
+{
+	UK_ASSERT(m);
+
+	if (!uk_semaphore_down_try(&m->writesem))
```
skuenzer: ...same here

skuenzer (changes requested):
Hey Cristian,
this PR looks great! The only thing that we should change is that we should use isr-safe functions within `mbox_isr.c`. I am also suggesting another comment in `mbox_defs.h` to avoid potential mistakes for potential PRs that do modifications in this file.
Thanks,
Simon

skuenzer (approved):
Hi Cristian, thanks a lot for the changes. The PR looks great. Finally we can start adopting the interrupt handler code in `lib/uknetdev` and `lib/ukblkdev` and make them ISR-safe.

Issue #72, opened 2020-11-28
Relocations leading to crashes for Rust helloworld on KVM


razvand:
**From Cyril Soldani**

In the [helloworld-rust example](https://github.com/cffs/app-helloworld-rust), for the KVM platform, some relocations lead to crashes (at least on my machine). For example, if you look at the function `<_ZN3std10sys_common12thread_local22register_dtor_fallback17hc5ad1095357fa8faE` (name mangling might be slightly different on other machines), you'll see at offset `0x36` a call to `pthread_key_create()`.

With `objdump -dr helloworld-rust_kvm-x86_64.o`, it looks like
```
36:   ff 15 00 00 00 00       callq  *0x0(%rip)
            38: R_X86_64_GOTPCREL   pthread_key_create-0x4
```
The `R_X86_64_GOTPCREL` relocation should make the linker replace the `0x0` offset by the actual offset to the GOT entry corresponding to `pthread_key_create()` (you can ignore the `-0x4`, which is normal). Alternatively, since our build is static and `pthread_key_create()` address is known at link time, the linker could also optimise out that GOT indirection, and just put the offset to the `pthread_key_create()` function there (32 bits are available for that offset).

With the `linuxu` platform, objdump -dr helloworld-rust_linuxu-x86_64.dbg gives
```
4f0606:       ff 15 4c f9 07 00       callq  *0x7f94c(%rip)
                    # 56ff58 <.got+0x148>
```
The call is indirect, through the GOT (where `pthread_key_create()` is stored at offset `0x148`), but it works.

For the KVM target, `objdump -dr helloworld-rust_kvm-x86_64.dbg` gives
```
1f8c16:       ff 15 74 4f 08 00       callq  *0x84f74(%rip)
                    # 27db90 <_edata+0x148>
```
where we see `_edata+0x148` in place of `.got+0x148`. It should still be correct because `_edata` marks the end of the `.data` section, and `.got` is put just after `.data`. `readelf --section-headers helloworld-rust_kvm-x86_64.dbg` gives
```
  [43] .data             PROGBITS         0000000000274000  00175000
       0000000000009a48  0000000000000000  WA       0     0     32
  [44] .got              PROGBITS         000000000027da48  0017ea48
       0000000000000158  0000000000000000  WA       0     0     8
```
and `0x274000 + 0x009a48 = 0x27da48`.

However, that call crashes at runtime. The GOT seems to be incorrectly initialized, so that this call tries to call address `0x0`.
[...]

PR #201, opened 2021-05-06
lib/uksched: Fixes for thread inittab

### LINKED ISSUE #197: Unable to run Unikraft hello-world since thread context updates
The [hello-world](https://github.com/unikraft/app-helloworld) application, presumably others, don't work since thread context updates.

Commit https://github.com/unikraft/unikraft/commit/56943ad23981b201466ef9b9e983235b6cb3f0c8 works, but nothing after that. Actually commits between https://github.com/unikraft/unikraft/commit/56943ad23981b201466ef9b9e983235b6cb3f0c8 and https://github.com/unikraft/unikraft/commit/a4295f7c90a9ae664813966e1cc54b32997c7d3d don't compile due to this issue. So something inserted between those commits broke the code:
```
  CC      libkvmplat: sw_ctx.common.o
/home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/sw_ctx.c: In function ‘sw_ctx_size’:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/sw_ctx.c:54:16: error: invalid application of ‘sizeof’ to incomplete type ‘struct ukplat_ctx’
  return sizeof(struct ukplat_ctx) + arch_extregs_size();
                ^~~~~~
/home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/sw_ctx.c:55:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
 /home/razvan/projects/unicore/unikraft-my/unikraft.git/support/build/Makefile.build:50: recipe for target '/home/razvan/projects/unicore/unikraft-my/apps/hello-world-from-scratch/build/libkvmplat/sw_ctx.common.o' failed
```

Since the commit that doesn't work (https://github.com/unikraft/unikraft/commit/a4295f7c90a9ae664813966e1cc54b32997c7d3d) to the current `staging` head, the `linuxu` build ends in Segmentation fault, and the KVM build hangs:
```
razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ ./build/hello-world-from-scratch_linuxu-x86_64
Segmentation fault
razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ ./qemu-guest -k build/hello-world-from-scratch_kvm-x86_64
Trying to get root privileges...
**************************************************************************
 QEMU:
   Name:                  408bdc57-2799-46ba-9bfb-bc212adfe166
   PID:                   17126
   Monitor socket:        /run/qemu-guest-17109_monitor.socket
   VNC listen:            5900

   SMP configuration:     sockets=1,cores=1,threads=1
   Memory:                64 MB
   Kernel:                build/hello-world-from-scratch_kvm-x86_64
   Parameters:            console=ttyS0

   Warning: No networking for guest!
   Warning: Guest will be terminated with CTRL+C!
**************************************************************************
[...]
SeaBIOS (version 1.10.2-1ubuntu1)
Booting from ROM...
```

skuenzer:
This series fixes two issues when building the thread inittab:
- An unnecessary alignment in the linker script caused some broken binaries. This is a fix for issue #197.
- `UK_THREAD_INIT()` or `UK_THREAD_INIT_PRIO` could only be specified once for each priority level. This is fixed by generating unique names for each entry.

### DIFF
--- a/lib/uksched/extra.ld
+++ b/lib/uksched/extra.ld
@@ -1,7 +1,6 @@
 SECTIONS
 {
 	.uk_thread_inittab : {
-		. = ALIGN(0x8);
 		PROVIDE(_uk_thread_inittab_start = .);
 		KEEP (*(.uk_thread_inittab0))
 		KEEP (*(.uk_thread_inittab0.*))
--- a/lib/uksched/include/uk/thread.h
+++ b/lib/uksched/include/uk/thread.h
@@ -152,7 +152,7 @@ struct uk_thread_inittab_entry {
 #define __UK_THREAD_INITTAB_ENTRY(init_fn, fini_fn, prio)		\
 	static const struct uk_thread_inittab_entry			\
 	__used __section(".uk_thread_inittab" # prio) __align(8)	\
-		__uk_thread_inittab ## prio ## _ ## entry = {		\
+		__uk_thread_inittab ## prio ## _ ## init_fn ## _ ## fini_fn = {\
 		.init = (init_fn),					\
 		.fini = (fini_fn)					\
 	}


jschlumpp:
While this PR fixes the problem in #197, it does not fix the underlying problem. The bootloader (of QEMU) simply copies the contents of the file into the memory at the location specified by the [multiboot header](https://github.com/unikraft/unikraft/blob/9d6d85530967184de97ab9b07d4d803ba7d8b2a6/plat/kvm/x86/entry64.S#L46). However, this ignores the fact that the layout in the file does not match layout in the memory. 
In this case the emitted padding caused a section filled with zeros to appear. The linker decided to use the `NOBITS` section type instead of actually storing the zeros. An ELF loader would automatically add this gap based on the program headers (note that the offsets and sizes do not match):
```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000001000 0x0000000000100000 0x0000000000100000
                 0x00000000000184a7 0x00000000000184a8  RWE    0x1000
  LOAD           0x00000000000194c0 0x0000000000119000 0x0000000000119000
                 0x00000000000061f0 0x000000000000c000  RW     0x20
  LOAD           0x0000000000020000 0x0000000000130000 0x0000000000130000
                 0x0000000000020000 0x0000000000020000  RW     0x10000
[...]
```

This PR makes the issue disappear because without padding the section will never appear as `NOBITS`. The workaround proposed in #72 fixes a similar issue by moving the broken section (`.got`) before a `NOBITS` section (`.tbss`).
A simple solution would be to use `opjcopy` to create an "image" of the address space as if loaded by an actual ELF loader. For example: `objcopy -O binary build/something-kernel_kvm-x86_64 build/something-kernel.bin_kvm-x86_64`. The resulting file can be loaded as usual with QEMU. While this results in larger binaries, it requires at no ELF parsing and segment relocation at boot time.

cc @marcrittinghaus

razvand:
@mschlumpp , this does not explain why the `linuxu` build fails as well:
```
razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ ./linuxu_not_works
Segmentation fault

razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ strace ./linuxu_not_works
execve("./linuxu_not_works", ["./linuxu_not_works"], 0x7fff204f0f70 /* 70 vars */) = -1 EINVAL (Invalid argument)
--- SIGSEGV {si_signo=SIGSEGV, si_code=SI_KERNEL, si_addr=NULL} ---
+++ killed by SIGSEGV +++
Segmentation fault (core dumped)

razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -l linuxu_not_works

Elf file type is EXEC (Executable file)
Entry point 0x400130
There are 3 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x00000000000000f0 0x00000000004000f0 0x00000000004000f0
                 0x000000000000be11 0x000000000000be18  RWE    0x10
  LOAD           0x000000000000bf20 0x000000000040c000 0x000000000040c000
                 0x0000000000002c20 0x0000000000003008  RW     0x20
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RWE    0x10

 Section to Segment mapping:
  Segment Sections...
   00     .text .uk_thread_inittab
   01     .uk_inittab .rodata .data .bss
   02

razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -S linuxu_not_works
There are 10 section headers, starting at offset 0xebc0:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .text             PROGBITS         00000000004000f0  000000f0
       000000000000be11  0000000000000000 WAX       0     0     16
  [ 2] .uk_thread_initta NOBITS           000000000040bf01  0000bf01
       0000000000000007  0000000000000000  WA       0     0     1
  [ 3] .uk_inittab       PROGBITS         000000000040c000  0000bf20
       0000000000000008  0000000000000000   A       0     0     8
  [ 4] .rodata           PROGBITS         000000000040c020  0000bf40
       0000000000002b85  0000000000000000   A       0     0     32
  [ 5] .tbss             NOBITS           000000000040eba8  0000eac5
       0000000000000000  0000000000000000 WAT       0     0     1
  [ 6] .data             PROGBITS         000000000040ebb0  0000ead0
       0000000000000070  0000000000000000  WA       0     0     16
  [ 7] .bss              NOBITS           000000000040ec20  0000eb40
       00000000000003e8  0000000000000000  WA       0     0     32
  [ 8] .comment          PROGBITS         0000000000000000  0000eb40
       0000000000000029  0000000000000001  MS       0     0     1
  [ 9] .shstrtab         STRTAB           0000000000000000  0000eb69
       0000000000000052  0000000000000000           0     0     1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  l (large), p (processor specific)
```

I'm investigating it.

If I placed the `.uk_thread_inittab` section after the `.uk_inittab` section (instead of after the `.text`) section, it would work. i.e. updating the linker script in `lib/uksched/extra.ld` to:
```
-INSERT AFTER .text;
+INSERT AFTER .uk_inittab;
```

jschlumpp:
> ```
>  Type           Offset             VirtAddr           PhysAddr
>                 FileSiz            MemSiz              Flags  Align
>  [...]
>  LOAD           0x000000000000bf20 0x000000000040c000 0x000000000040c000
>                 0x0000000000002c20 0x0000000000003008  RW     0x20
>  [...]
> ```

According to elf(5):
> Loadable process segments must have congruent values for p_vaddr and p_offset, modulo the page size. 

If this condition is not satisfied, Linux is unable to memory map the segment causing `execve` to return `EINVAL`. While Linux does not specify the exact reason, it is possible to use QEMU to verify this:
```
$ qemu-x86_64 -strace -trace 'enable=*' build/linuxu-test_linuxu-x86_64.dbg
[...]
target_mmap start=0x400000 len=0x100b680 prot=0x0 flags=0x4032 fd=-1 offset=0x0
target_mmap_complete retaddr=0x400000
target_mmap start=0x400000 len=0x9000 prot=0x7 flags=0x12 fd=3 offset=0x0
target_mmap_complete retaddr=0x400000
target_mmap start=0x409000 len=0x3000 prot=0x3 flags=0x12 fd=3 offset=0x88e0
qemu-x86_64: build/linuxu-test_linuxu-x86_64.dbg: Error mapping file: Invalid argument
```

The problem seems to be caused by the `--omagic` linker flag set in [Makefile.uk](https://github.com/unikraft/unikraft/blob/6f3f684f2e1ca147257c81938db980c8543f3658/Makefile.uk#L30): 
> Set the text and data sections to be readable and writable.  Also, do not page-align the data segment, [...]

Removing the flag fixes (this) problem, but I do not know why the flag was used.

razvand:
@mschlumpp , great, thanks for the finding.

The `--omagic` option is required for the KVM build. Without that option, KVM won't start. We stumbled into that while trying to add randomization support. If I recall this is because bootstrapping magic requires it.

jschlumpp:
@razvand I just removed the flag and the hello world application started in KVM without any problems. Or does it only happen for some applications?

razvand:
@mschlumpp , on my side, if I keep the initial version of the `lib/uksched/extra.ld` file (i.e. with `. = ALIGN(0x8);`), removing `--omagic` allows the `linuxu` build to work, but not the `kvm` build.

This happens for the [`RELEASE-0.5` version](https://github.com/unikraft/unikraft/tree/RELEASE-0.5) as well: removing `--omagic` allows `linuxu` to work, but `kvm` no longer works.

Maybe it's related to the QEMU version. On my side, the QEMU trace command you used doesn't show anything from the broken `linuxu` ELF:
```
$ qemu-x86_64 -strace -trace 'enable=*' linuxu_not_works 
linuxu_not_works: Invalid argument
```

razvand:
@skuenzer , I propose to merge this PR to address issue #197 . I would start a separate issue to discuss a solution for the underlying problem pointed out by @mschlumpp for KVM: the inconsistency between the expected in-memory layout (loaded from ELF) and the actual in-memory layout (simply copying the ELF file contents).

On this PR, an alternate solution would be to use `INSERT AFTER .uk_inittab;` as the last line in `lib/uksched/extra.ld`.

jschlumpp:
@razvand 
> This happens for the [`RELEASE-0.5` version](https://github.com/unikraft/unikraft/tree/RELEASE-0.5) as well: removing `--omagic` allows `linuxu` to work, but `kvm` no longer works.

I think you might have hit the issue described previously. Can you try to `objcopy -O binary` the KVM file and load the resulting file in QEMU? All the mentioned combinations seem to work for me...

razvand:
@mschlumpp , yes, using `objcopy` to extract the binary works. Though I'm not sure what issue I may have hit that you didn't. The build should be more or less the same. The only `NOBITS` sections are the `.tbss` (empty) and the `.bss` that I guess will use the `.intrstack` section. Why would it work on my side and not on your side?

```
razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -l ./build/hello-world-from-scratch_kvm-x86_64

Elf file type is EXEC (Executable file)
Entry point 0x107420
There are 2 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000140000 0x0000000000140000  RWE    0x200000
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RWE    0x10

 Section to Segment mapping:
  Segment Sections...
   00     .text .uk_ctortab .uk_inittab .rodata .data .bss .intrstack
   01

razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -S ./build/hello-world-from-scratch_kvm-x86_64
There are 11 section headers, starting at offset 0x140080:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .text             PROGBITS         0000000000100000  00100000
       00000000000147b7  0000000000000000  AX       0     0     4096
  [ 2] .uk_ctortab       PROGBITS         0000000000115000  00115000
       0000000000000010  0000000000000000   A       0     0     8
  [ 3] .uk_inittab       PROGBITS         0000000000115010  00115010
       0000000000000008  0000000000000000   A       0     0     8
  [ 4] .rodata           PROGBITS         0000000000116000  00116000
       00000000000034a0  0000000000000000   A       0     0     32
  [ 5] .tbss             NOBITS           00000000001194a0  001194a0
       0000000000000000  0000000000000000 WAT       0     0     1
  [ 6] .data             PROGBITS         000000000011a000  0011a000
       0000000000000100  0000000000000000  WA       0     0     32
  [ 7] .bss              NOBITS           000000000011b000  0011a100
       0000000000005000  0000000000000000  WA       0     0     32
  [ 8] .intrstack        PROGBITS         0000000000120000  00120000
       0000000000020000  0000000000000000  WA       0     0     65536
  [ 9] .comment          PROGBITS         0000000000000000  00140000
       0000000000000029  0000000000000001  MS       0     0     1
  [10] .shstrtab         STRTAB           0000000000000000  00140029
       0000000000000056  0000000000000000           0     0     1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  l (large), p (processor specific)
```

razvand:
When `--omagic` is present, the `.intrstack` section is placed in a new segment. I'm not sure how the linker decides to do that. And I'm still in the dark on why it `--omagic` works on your side and not on my side.

```
razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -l kvm_omagic

Elf file type is EXEC (Executable file)
Entry point 0x107420
There are 3 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000001000 0x0000000000100000 0x0000000000100000
                 0x000000000001a100 0x0000000000020000  RWE    0x1000
  LOAD           0x0000000000020000 0x0000000000120000 0x0000000000120000
                 0x0000000000020000 0x0000000000020000  RW     0x10000
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RWE    0x10

 Section to Segment mapping:
  Segment Sections...
   00     .text .uk_ctortab .uk_inittab .rodata .data .bss
   01     .intrstack
   02

razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -S kvm_omagic
There are 11 section headers, starting at offset 0x40080:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .text             PROGBITS         0000000000100000  00001000
       00000000000147b7  0000000000000000 WAX       0     0     4096
  [ 2] .uk_ctortab       PROGBITS         0000000000115000  00016000
       0000000000000010  0000000000000000   A       0     0     8
  [ 3] .uk_inittab       PROGBITS         0000000000115010  00016010
       0000000000000008  0000000000000000   A       0     0     8
  [ 4] .rodata           PROGBITS         0000000000116000  00017000
       0000000000003490  0000000000000000   A       0     0     32
  [ 5] .tbss             NOBITS           0000000000119490  0001a490
       0000000000000000  0000000000000000 WAT       0     0     1
  [ 6] .data             PROGBITS         000000000011a000  0001b000
       0000000000000100  0000000000000000  WA       0     0     32
  [ 7] .bss              NOBITS           000000000011b000  0001b100
       0000000000005000  0000000000000000  WA       0     0     32
  [ 8] .intrstack        PROGBITS         0000000000120000  00020000
       0000000000020000  0000000000000000  WA       0     0     65536
  [ 9] .comment          PROGBITS         0000000000000000  00040000
       0000000000000029  0000000000000001  MS       0     0     1
  [10] .shstrtab         STRTAB           0000000000000000  00040029
       0000000000000056  0000000000000000           0     0     1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  l (large), p (processor specific)
```

jschlumpp:
> @mschlumpp , yes, using `objcopy` to extract the binary works. 

Did you try to load the output into QEMU?

> Though I'm not sure what issue I may have hit that you didn't. The build should be more or less the same. 

The problem is that any small variation can result in breakages because of the alignments. A different compiler version in the toolchain could be enough to cause problems.

> The only `NOBITS` sections are the `.tbss` (empty) and the `.bss` that I guess will use the `.intrstack` section. Why would it work on my side and not on your side?

Completely empty sections do not break anything in the current state. And `.bss` is also fine, because the multiboot AOUT_KLUDGE model expects the `.bss` section at the end of the image.

> ```
> razvan@yggdrasil:~/.../unikraft-my/apps/hello-world-from-scratch$ readelf -l ./build/hello-world-from-scratch_kvm-x86_64
>   Type           Offset             VirtAddr           PhysAddr
>                  FileSiz            MemSiz              Flags  Align
>   LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
>                  0x0000000000140000 0x0000000000140000  RWE    0x200000

This segments looks really weird: the first segment starts at 0x0 instead of 0x100000. 

> When `--omagic` is present, the `.intrstack` section is placed in a new segment. I'm not sure how the linker decides to do that. 

`.intrstack` requires a larger alignment than the previous segment. Therefore, it must emit a new one.

> And I'm still in the dark on why it `--omagic` works on your side and not on my side.

I guess you meant to write "removing `--omagic`". Double check that removing the flag and creating a bare memory image using objcopy really does not work for you. If it still does not work there might be another issue.

razvand:
@mschlumpp , yes, the resulting `objcopy` file works in QEMU / KVM. Using `objcopy` makes all images work under QEMU / KVM, irrespective of the use of `--omagic`.

In summary, the situation where the default ELF build doesn't work are (separately):
* the removal of the `--omagic` linker field (for any Unikraft commit / version / release)
* the initial alignment + `NOBITS` issue fixed by this PR

> I guess you meant to write "removing --omagic". Double check that removing the flag and creating a bare memory image using objcopy really does not work for you. If it still does not work there might be another issue.

Yes, I meant removing `--omagic`. From my understanding, removing `--omagic` still allows the default ELF build for QEMU / KVM to work (load and run). It doesn't on my side; if `--omagic` is absent, all default ELF builds for QMEU / KVM (for all Unikraft versions that I tested) fail.

razvand:
@mschlumpp , it the `--omagic` removal issue reproducible on your side? If you remove `--omagic`, do the default ELF builds still work in QEMU / KVM?

jschlumpp:
@razvand yes it still works, but probably just by luck. 

The main problem about omagic is that `ld` seems to sometimes output broken ELF binaries. The side effect of the flag is that it also shifts sections around due to alignment (which can fix _or_ break stuff). On the other hand the flag makes it much more likely to build a working memcpy-able kernel (as it basically removes a large amount of padding).

razvand (approved):
There are some fundamental issues on the linking process as pointed out by @mschlumpp . Let's merge this for now and we'll open up a specific issue regarding linking.

PR #202, opened 2021-05-07
lib/ukstore: Introduce the `ukstore` library

### LINKED ISSUE #5: Library for generic unikernel control and monitoring (e.g., REST API)

craciunoiuc:
These commits introduce a new internal library to Unikraft, `ukstore`.
The role of this new library is to provide a standard way of passing
information getters/setters from one place to another.

The function pointers are stored in entries and entries are stored in
folders. These are structured in two different ways, statically and
dynamically. The static ones are used for information that does not
change at runtime. The dynamic ones can be added and deleted at runtime
and they are structured differently (section array of folders vs. list
of folders).

To generate sections per library the `AWK` tool is used together with
the build system. This library has a dependency only to `nolibc`.

In order to work, this, already merged, commit is needed: 9d6d85530967184de97ab9b07d4d803ba7d8b2a6

This library is needed for the porting of Prometheus and other tracks
in the future.


lib/ukstore/Config.uk
```diff
@@ -0,0 +1,4 @@
+menuconfig LIBUKSTORE
```
skuenzer: For the cases where you do not have additional library-specific coinfigurations you can use `config` instead of `menuconfig`

lib/ukstore/Config.uk
```diff
@@ -0,0 +1,4 @@
...
+	bool "ukstore: Tree buffer storage"
```
skuenzer: You probably want to give it a different title, like: `"ukstore: Run-time configuration and monitoring database"`?

lib/ukstore/Makefile.uk
```diff
@@ -0,0 +1,55 @@
...
+		echo $(foreach P,$(UK_PLATS) $(UK_PLATS-y),\
+		     $(if $(call qstrip,$($(call uc,$(P))_LIBS) $($(call uc,$(P))_LIBS-y)),\
+		     $(foreach L,$($(call uc,$(P))_LIBS) $($(call uc,$(P))_LIBS-y), \
+		     $(if $(call qstrip,$($(call vprefix_lib,$(L),SRCS)) $($(call vprefix_lib,$(L),SRCS-y))), \
+		     $(L) \
+		     )))) \
+		     $(UK_LIBS) $(UK_LIBS-y) \
+		| tr ' ' '\n' > $@)
+
+$(LIBUKSTORE_GEN_INCLUDES_PATH)/uk/_store_libs.h: $(LIBUKSTORE_BASE)/gen_libs.h.awk $(LIBUKSTORE_BUILD)/uk_libs.in
```
skuenzer, line 36: We usually call this initial commit `ukstore: Introduce library skeleton` (your description is fine) and register this library and to the build system (merge your last commit with this one). In order that all builds work when you add the commits piece by piece from your series, can you make sure that extra `Makefile.uk` entries are added with the commits that introduce the according source files? This would mean that you would generate at most `uk_libs.in` with this commit and add the other rules later.

lib/ukstore/exportsyms.uk
```diff
@@ -0,0 +1,6 @@
+uk_store_set_value
+uk_store_get_value
+uk_store_get_entry
+_uk_store_add_folder
+_uk_store_get_static_folder
+_uk_store_get_dynamic_folder
```
skuenzer, line 22: Can you add the entries to `exportsyms.uk` as soon as you introduce them with the source file that gets compiled?

lib/ukstore/gen_libs.ld.awk
```diff
@@ -0,0 +1,24 @@
...
+	print "/* Automatically generated file; DO NOT EDIT */"
+	print ""
+    print "SECTIONS"
+    print ""
+    print "{"
+}
+
+/[a-zA-Z0-9]+/{
+    printf "\t.uk_store_lib_%s : {\n", $1;
+    print  "\t\t. = ALIGN(0x8);"
```
skuenzer: I just had an issue caused by an `ALIGN` like this: #197 . It is better to remove this line. It should still work.

lib/ukstore/store.ld
```diff
@@ -0,0 +1,11 @@
+SECTIONS
+{
+	.uk_store_entries_list : {
+		. = ALIGN(0x8);
```
skuenzer: Remove this align. Currently it can cause issues with the generated ELF files #197

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+#endif /* CONFIG_LIBUKSTORE */
+
+/**
+ * Returns the name of the library
+ *
+ * @return this libraries' name
+ */
+#define uk_store_self_name()	\
+	__LIBNAME__
```
skuenzer: I am not sure how useful this macro is. I would remove it because it does not return a string.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+/* Library-specific information generated at compile-time */
+#include <uk/_store_libs.h>
+
+/* Do not call directly */
+#define _uk_store_libid(libname)	\
+	__UK_STORE_ ## libname
+
+/**
+ * Returns the id of a given library.
+ *
```
skuenzer, line 110: You should also document the parameter, it is also not a string (like `"`) - it is part of a macro name for a library.
One could think of
```
int id = uk_store_libid("vfscore");
```
although it is
```
int id = uk_store_libid(LIBVFSCORE);
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+#define uk_store_self_name()	\
+	__LIBNAME__
+
+/**
+ * Returns the id of the place it was called from.
+ *
+ * @return this libraries' id
+ */
+#define uk_store_libid_self()	\
+	uk_store_libid(uk_store_self_name())
```
skuenzer:
```
#define uk_store_libid_self()	\
	uk_store_libid(__LIBNAME__)
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+/* All basic types that exist - use these when interacting with the API */
+enum uk_store_entry_basic_type {
```
skuenzer: I find `uk_store_entry_basic_type` really long. I would prefer `uk_store_entry_type`, but that is up to you.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+	} set;
+
+	/* The entry name */
+	const char *entry_name;
+
+	/* Entry flags */
+	__u16 flags;
+
+	/* Entry getter/setter type */
+	__u16 type;
```
skuenzer: Isn't this ...?
``` c
enum uk_store_entry_basic_type type;
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+		uk_store_set_s64_func_t  s64;
+		uk_store_set_u64_func_t  u64;
+		uk_store_set_uptr_func_t uptr;
+	} set;
+
+	/* The entry name */
+	const char *entry_name;
+
+	/* Entry flags */
+	__u16 flags;
```
skuenzer: Maybe you want to move the flag definition closeer to this struct... I often saw that just `int` is used for flags
``` c
#define UK_STORE_ENTRY_ISSTATIC (1 << 1)
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+		uk_store_set_u16_func_t  u16;
+		uk_store_set_s32_func_t  s32;
+		uk_store_set_u32_func_t  u32;
+		uk_store_set_s64_func_t  s64;
+		uk_store_set_u64_func_t  u64;
+		uk_store_set_uptr_func_t uptr;
+	} set;
+
+	/* The entry name */
+	const char *entry_name;
```
skuenzer: I would just take `name` for this field.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+struct uk_store_folder {
+	/* List for the dynamic folders only */
+	struct uk_list_head folder_head;
+
+	/* List for the entries in the folder */
+	struct uk_list_head entry_head;
+
+	/* The folder name */
+	const char *folder_name;
```
skuenzer: You can call this also just `name` - it would be still clear.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+	none,
+	s8,
+	u8,
+	s16,
+	u16,
+	s32,
+	u32,
+	s64,
+	u64,
+	uptr
```
skuenzer: Would these names conflict with `uklibparams` when you include both headers?
I would suggest to prefix all entries to avoid clashes, like:
```
enum uk_store_entry_type {
     UK_STORE_ENTRY_TYPE_undefined = 0,
     UK_STORE_ENTRY_TYPE_s8,
     UK_STORE_ENTRY_TYPE_u8,
     <...>
}
```
You could use pre-processor concatenation so that a programmer does not have to type the whole typename.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+#define UK_STORE_FLAG_STATIC	1
+
+/**
+ * Retuns the folder_entry of an entry
+ *
+ * @param entry_ptr the entry
+ */
+#define	uk_store_get_folder_entry(entry_ptr)	\
+	__containerof(entry_ptr, struct uk_store_folder_entry, entry)
```
skuenzer: Maybe do a static inline function that also has an assertion that checks that the flag STATIC is not set.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * Adds a dynamic folder to a library
+ *
+ * @param folder the folder to add it to
+ */
+#define uk_store_add_folder(folder) \
+	_uk_store_add_folder(uk_store_libid_self(), folder)
+
+extern int
```
skuenzer: Since this is a function prototype, you do not need the `extern` keyword. This is true also for the following function prototypes.

lib/Makefile.uk
```diff
@@ -39,4 +39,5 @@ $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukmmap))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukblkdev))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/posix-process))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/uksp))
+$(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukstore))
```
skuenzer, line 41: Do this with the first patch

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+#endif
+
+#if !CONFIG_LIBUKSTORE
+#define uk_store_libid(libname) (-1)
+
+#define UK_STORE_INITREG_FOLDER(fldr)
+
+#define UK_STORE_INITREG_ENTRY(entry, e_type, e_get, e_set)
+
+#define uk_store_entry_init(entry, e_name, e_type, getter, setter)
```
skuenzer: for empty function-like declarations use
``` c
do {} while (0)
```
This makes sure that when you use them within if cases without `{` `}` blocks is still working fine, e.g.,:

``` c
if (new)
   uk_store_entry_init();
else
   /* else */
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+typedef __s16 (*uk_store_get_s16_func_t)(void);
+typedef __u16 (*uk_store_get_u16_func_t)(void);
+typedef __s32 (*uk_store_get_s32_func_t)(void);
+typedef __u32 (*uk_store_get_u32_func_t)(void);
+typedef __s64 (*uk_store_get_s64_func_t)(void);
+typedef __u64 (*uk_store_get_u64_func_t)(void);
+typedef __uptr (*uk_store_get_uptr_func_t)(void);
+
+/* Setter definitions */
+typedef void (*uk_store_set_s8_func_t)(__s8);
```
skuenzer: The setters should probably return `int` in order to forward error codes in case setting did not work. Anything positive and `0` could mean success, any negative number could carry an error code `-errno`.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if !CONFIG_LIBUKSTORE
+#define uk_store_libid(libname) (-1)
+
+#define UK_STORE_INITREG_FOLDER(fldr)
+
+#define UK_STORE_INITREG_ENTRY(entry, e_type, e_get, e_set)
```
skuenzer: This macro may still need to define something but mark it with `__unused` so that you do not get compiler warnings when your getter or setter was defined as `static` and is not used anymore. You also want to keep the ability that the function types are checked by the compiler.

Imagine:
``` c
static __u8 my_getter(void) {
   return u8val;
}

UK_STORE_INITREG_ENTRY(u8val, u8, my_getter, NULL);
```

It could probably be done with something like this:
``` c
#define UK_STORE_INITREG_ENTRY(entry, e_type, e_get, e_set) \
	static const struct uk_store_entry __unused		\
	__uk_store_entries_list ## _ ## entry  = {			\
		.type       = (e_type),					\
		.get.e_type = (e_get),					\
		.set.e_type = (e_set),					\
	}
```
The compiler should optimize it out when DCO is enabled.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * Adds an entry to the entry section.
+ *
+ * @param entry the entry in the section
+ * @param e_type the type of the function
+ * @param e_get getter pointer
+ * @param e_set setter pointer
+ */
+#define UK_STORE_INITREG_ENTRY(entry, e_type, e_get, e_set)		\
```
skuenzer: I would rather call it `UK_STORE_STATIC_ENTRY()` or `UK_STORE_SENTRY()`.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+ * Gets the value returned by the saved function and puts it in `out`
+ * Use only if you really know the type.
+ *
+ * @param entry the entry to use
+ * @param e_type the type of the function
+ * @param out the place where to store the result
+ */
+#define uk_store_get_value_typed(entry, e_type, out)			\
+	do {								\
+		if (unlikely(!(entry) || !(e_type)))			\
```
skuenzer: I would use `UK_ASSERT` for this check since it is clearly a programming mistake.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+ * Sets the value from `in` with the saved function.
+ * Use only if you really know the type.
+ *
+ * @param entry the entry to use
+ * @param e_type the type of the function
+ * @param in the value to give to the setter
+ */
+#define uk_store_set_value_typed(entry, e_type, in)			\
+	do {								\
+		if (unlikely(!(entry) || !(e_type)))			\
```
skuenzer: I would use `UK_ASSERT` for this check since it is clearly a programming mistake.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Saves a new setter function in the entry
+ *
+ * @param entry the place where to save the new function
+ * @param e_type the new type of the function
+ * @param func the function
+ */
+#define uk_store_update_setter(entry, e_type, func)	\
+	do {						\
+		if (unlikely(!(entry) || !(e_type)))	\
```
skuenzer: I would use `UK_ASSERT` for this check since it is clearly a programming mistake.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Saves a new getter function in the entry
+ *
+ * @param entry the place where to store the function
+ * @param e_type the new type of the function
+ * @param func the function
+ */
+#define uk_store_update_getter(entry, e_type, func)	\
+	do {						\
+		if (unlikely(!(entry) || !(e_type)))	\
```
skuenzer: I would use `UK_ASSERT` for this check since it is clearly a programming mistake.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __UK_STORE_H__
+#define __UK_STORE_H__
+
+#include <uk/config.h>
+#include <uk/essentials.h>
+#include <string.h>
```
skuenzer, line 36: We started including libc headers before Unikraft headers as best practice (you couldn't know). Especially `<uk/essentials.h>` should be included as last header because of potential duplicate definitions that some libc's do not handle well.

lib/ukstore/store.c
```diff
@@ -0,0 +1,241 @@
...
+ */
+int
+uk_store_set_value(struct uk_store_entry *entry, void *in)
+{
+	if (unlikely(!entry))
+		return -EINVAL;
+
+	switch (entry->type) {
+	case uptr:
+		entry->set.uptr(*((__uptr *) in));
```
skuenzer: Let the setters also return error values
``` c
case uptr:
   return entry->set.uptr(*((__uptr *) in));
```

lib/ukstore/store.c
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * Calls the saved setter with the value in `in`
+ *
+ * @param entry the entry to call from
+ * @param in the value to give to the setter
+ */
+int
+uk_store_set_value(struct uk_store_entry *entry, void *in)
```
skuenzer: I was wondering if it is better to provide one function for each setter type, like
``` c
int uk_store_set_u8(struct uk_store_entry *e, __u8 val);
int uk_store_set_u16(struct uk_store_entry *e, __u16 val);
int uk_store_set_u32(struct uk_store_entry *e, __u32 val);
int uk_store_set_charp(struct uk_store_entry *e, const char *val);
```
The function would do the corresponding casting to the target type. Especially when you set a `const char *` to an `__u8` entry, you may want to do the appropriate casting with `atoi()` etc. You could also make sure that a value is within range and return `-ERANGE` otherwise (we should have min/max defined for each type).
You could do a generic switch with a wrapper macro in the header, like:

``` c
#define uk_store_set_value(entry, intype, in) \
   uk_store_set_value_ ## intype ((entry), (in))
```

lib/ukstore/store.c
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * Calls the saved getter and saves the returned value in out
+ *
+ * @param entry the entry to call from
+ * @param out the value to return
+ */
+int
+uk_store_get_value(struct uk_store_entry *entry, void *out)
```
skuenzer: I would do the something similar for the getters. Also here you can check if the returned number is still in range:
``` c
int uk_store_get_u8(struct uk_store_entry *e, __u8 *out);
int uk_store_get_u16(struct uk_store_entry *e, __u16 *out);
int uk_store_get_u32(struct uk_store_entry *e, __u32 *out);
int uk_store_get_charp(struct uk_store_entry *e, char *buf, size_t buflen); /* uses snprintf() for number entries */
```

lib/ukstore/include/uk/store.h
skuenzer: I would add a type called `charp` for setting and returning strings. `uptr` is just enpugh room to contain an address and should not be used for strings. It is also fine to add handling for strings later if it is not needed now.

lib/ukstore/include/uk/store.h
skuenzer: I can imagine to use the setters also to trigger system actions in some cases (e.g., change the size of a cache, requesting a system shutdown, requesting a rescan of PCI devices). In all of these cases you want to be able to return errors.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+	const char *entry_name;
+
+	/* Entry flags */
+	__u16 flags;
+
+	/* Entry getter/setter type */
+	__u16 type;
+} __align8;
+
+struct uk_store_folder {
```
skuenzer: Maybe you want to make this header private to the library and define it within the C file. Here you would only do the prototype:
``` c
struct uk_store_folder;
```

I was actually wondering if we should allocate all dynamic libraries on registration within the library with malloc. THis is making the alloc and release scheme easier.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+	struct uk_list_head folder_head;
+
+	/* List for the entries in the folder */
+	struct uk_list_head entry_head;
+
+	/* The folder name */
+	const char *folder_name;
+} __align8;
+
+struct uk_store_folder_entry {
```
skuenzer: I would make this also private and keep the prototoype here for type checking:
``` c
struct uk_store_folder_netry;
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+ */
+#define	uk_store_get_folder_entry(entry_ptr)	\
+	__containerof(entry_ptr, struct uk_store_folder_entry, entry)
+
+/* Static entry array start+end points */
+extern struct uk_store_entry *uk_store_entries_start;
+extern struct uk_store_entry *uk_store_entries_end;
+
+/* Do not call directly */
+#define __UK_STORE_INITREG_FOLDER(fldr, fldr_string)			\
```
skuenzer: I would avoid folders for the static entries - I think it makes it more complicated and I can imagine that we can live without it. It would have also one benefit when dealing with lookups, like:
``` c
/* lookup for a static entry */
struct uk_store_entry *_uk_store_sentry_find(unsigned int libid, const char *entryname);
/* lookup for a dynamic (at run-time allocated) entry */
struct uk_store_entry *_uk_store_dentry_find(unsigned int libid, const char *foldername, const char *entryname);

/* Do the switch as static inline so that the compiler can optimize to the direct call in case the foldername is hard-coded in the caller code */
static inline struct uk_store_entry *
_uk_store_entry_find(unsigned int libid, const char *foldername, const char *entryname)
{
   if (!foldername)
      return _uk_store_sentry_find(libid, entryname);
   return _uk_store_dentry_find(libid, foldername, entryname);
}

/* API interface */
#define uk_store_entry_find(libname, foldername, entryname) \
   _uk_store_entry_find(uk_store_libid(libname), foldername, entryname)
```

For static entry lookups you just set `foldername` to `NULL`:
``` c
static_entry = uk_store_entry_find(LIBUKALLOC, NULL, "free");
```

Dynamic entries would have a folder name
``` c
dynamic_entry = uk_store_entry_find(LIBUKALLOC, "1", "free");
```

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Initializes an entry
+ *
+ * @param entry the entry to initialize
+ * @param e_name the name of the entry
+ * @param e_type the type of the entry
+ * @param getter getter pointer
+ * @param setter setter pointer
+ */
+#define uk_store_entry_init(entry, e_name, e_type, getter, setter)	\
```
skuenzer: Hum... I am thinking in the meantime that the library usage is easier if we move memory management of the entries to the library itself. Encapsulating of the entries would be consistent (thus it is safer to use `containerof` internally) and registering code. Also, refcounting and releasing of memory gets easier.
In such a case, you would not have `init` functions anymore. What do you think?

I am thinking of an interface like this:
``` c
struct uk_folder *uk_store_folder_create(const char *name);

#define uk_store_dentry_create(folder, name, type, getter, setter) \
   uk_store_dentry_create_ ## type((folder), (name), (getter), (setter))

struct uk_entry_entry *uk_store_dentry_create_u8(struct uk_store_folder *f, uk_store_get_u8_func_t s, uk_store_set_u8_func_t f);
struct uk_entry_entry *uk_store_dentry_create_u16(struct uk_store_folder *f, uk_store_get_u16_func_t s, uk_store_set_u16_func_t f);
struct uk_entry_entry *uk_store_dentry_create_u32(struct uk_store_folder *f, uk_store_get_u32_func_t s, uk_store_set_u32_func_t f);
/* ... */
```

⚠️ One note regarding the name strings when creating folders/entries: Because an entry lives longer than the caller and we take over memory management of entries to the library, you should create a duplicate allocation for the name filed. Either you embedded it to the struct (like 8 or 16 chars max) or do a `strdup()` or `strndup()`. We should not expect that the caller keeps the string alive.

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * Releases an entry (decreases the refcount and sets the reference to NULL)
+ * If the refcount is 0 (it was also released by the creator), it is removed
+ * from the list and memory can be free.
+ *
+ * @param entry pointer to the entry to release
+ */
+static inline void
+uk_store_release_entry(struct uk_store_entry **entry) {
```
skuenzer: If we do memory management within the library, I would then call it `free` instead of `release`

lib/ukstore/include/uk/store.h
```diff
@@ -0,0 +1,452 @@
...
+ * @param out the place where to store the result
+ */
+#define uk_store_get_value_typed(entry, e_type, out)			\
+	do {								\
+		if (unlikely(!(entry) || !(e_type)))			\
+			break;						\
+		*((__##e_type *) (out)) = (entry)->get.e_type();	\
+	} while (0)
+
+extern int uk_store_get_value(struct uk_store_entry *entry, void *out);
```
skuenzer: Please read my comments about the non-typed getters/setters in the C file. I think `void *` is risky and can cause easy mis-use and programming mistakes.

skuenzer (changes requested):
Hey, the patch series looks really good. I like the minimalism of the implementation. I would change  the following points (of course open for discussion):
* I would change a bit of the structure of the patch series. I would introduce things feature-like, for example: (1) skeleton, (2) libid generating, (3) static entries, (4) dynamic entries
* I would not allow a folder structure for static entries. I think this restriction would not harm and would simplify the library usage even further
* Since I saw the API, I am thinking that it is probably best if we move memory managenemt to the library and should not expose it to the users. This is making also enabling/disabling `uk_store` from the build easier. Especially when switched off, there is no dangling allocation of entries happening in the libraries...

lib/ukstore/Config.uk
craciunoiuc: I went with `"ukstore: Run-time configuration and monitoring"`, as the one you suggested seemed a bit too long

lib/ukstore/include/uk/store.h
craciunoiuc: Thanks! There were two enums before and I forgot to edit this.

lib/ukstore/include/uk/store.h
craciunoiuc: I tested it and there aren't any clashes. The ones in `uklibparams` have this form: `_LIB_PARAM___s32`. Still I am up for this change if you think it would help in the long run.

lib/ukstore/store.c
craciunoiuc: This seems fine. Though, I don't really get why we should check if values are in a range.  If someone calls a `__u8` function with an `__s32` that's on them, and they should take care to not lose information when casting.

lib/ukstore/store.c
craciunoiuc: Same thing here like the setters.

lib/ukstore/store.c
craciunoiuc: Also, what purpose does `enum uk_store_entry_type type;` have if you give the type to it like `#define uk_store_set_value(entry, intype, in)`. Or did you suggest this macro for ease of use?

craciunoiuc: To sum it up: the current `uk_store_set_value` function will do the switch based on the type. Afterwards it uses your suggested macro to call the specific function based on the type and return its value.

Did I get this right? I am unsure what will happen to the current `uk_store_set_value` function.

lib/ukstore/include/uk/store.h
```diff
@@ -46,23 +47,47 @@ extern "C" {
...
+	__unused __section(".uk_store_entries_list") __align8		\
+	__uk_store_entries_list ## _ ## entry = {			\
+		.name = STRINGIFY(entry),				\
+		.type       = (e_type),					\
+		.get.e_type = (e_get),					\
+		.set.e_type = (e_set),					\
+		.flags      = UK_STORE_ENTRY_ISSTATIC			\
+	}
 
 #define uk_store_entry_init(entry, e_name, e_type, getter, setter)
```
skuenzer: Because this is a multiline macro, make sure that you add a `\` at the end of the line ;-)

lib/ukstore/include/uk/store.h
```diff
@@ -187,15 +212,20 @@ struct uk_store_folder_entry {
 	__atomic refcount;
 } __align8;
 
-#define UK_STORE_FLAG_STATIC	1
+#define UK_STORE_ENTRY_ISSTATIC	(1 << 1)
```
skuenzer: Oh, I just see that I was putting a bit of weird worded example in the previous review regarding the naming. I think what makes sense is:
``` c
#define UK_STORE_ENTRY_FLAG_STATIC    (1 << 0)

/* test macro */
#define UK_STORE_ENTRY_ISSTATIC(entry) \
   ((entry)->flags & UK_STORE_ENTRY_FLAG_STATIC)
```

lib/ukstore/include/uk/store.h
```diff
@@ -379,8 +409,8 @@ uk_store_release_entry(struct uk_store_entry **entry) {
  */
 #define uk_store_update_getter(entry, e_type, func)	\
 	do {						\
-		if (unlikely(!(entry) || !(e_type)))	\
-			break;				\
+		UK_ASSERT(entry);			\
```
skuenzer: Be careful with the whitespaces that your IDE produces ;-)

lib/ukstore/include/uk/store.h
```diff
@@ -431,12 +473,24 @@ extern int uk_store_get_value(struct uk_store_entry *entry, void *out);
  */
 #define uk_store_set_value_typed(entry, e_type, in)			\
 	do {								\
-		if (unlikely(!(entry) || !(e_type)))			\
-			break;						\
+		UK_ASSERT(entry);					\
+		UK_ASSERT(e_type);					\
```
skuenzer: I am not sure how much this assertions helps. Because the type is used for casting later and there is no defintion, the compiler would anyways fail. If you want to avoid that the type NONE is not used you may want to consider a compile-time assertion `UK_CTASSERT`.

lib/ukstore/store.c
```diff
@@ -34,6 +34,90 @@
...
+{
+	return e->set.u32(*val);
+}
+int uk_store_set_s32(struct uk_store_entry *e, __s32 *val)
+{
+	return e->set.s32(*val);
+}
+int uk_store_set_u64(struct uk_store_entry *e, __u64 *val)
+{
+	return e->set.u64(*val);
```
skuenzer: My idea of these functions was to do the right conversion between the given and target entry type. Imagine you set a `__u64` to an `__s8` entry. You may want to return `-ERANGE` if you input is greater than `__S8_MAX`. Those limits differ per target type. If you set you `__u64` to a `charp` (string) you need to do a integer to string conversion.
``` c
int uk_store_set_s64(struct uk_store_entry *e, __s64 val)
{
   UK_ASSERT(e);
   switch (e->type) {
   case s8:
      if (unlikely((val > __S8_MAX) || (val < __S8_MIN)))
         return -ERANGE;
      return e->set.s8((__s8) val);

   case u8:
      if (unlikely((val > __U8_MAX) || (val < 0)))
         return -ERANGE;
      return e->set.u8((__u8) val);
   }

   /* ... */
}
```

lib/ukstore/store.c
```diff
@@ -34,6 +34,90 @@
 #include <errno.h>
 
 
+
+int uk_store_set_u8(struct uk_store_entry *e, __u8 *val)
```
skuenzer: For setting integers, you can just handover *by value*, you do not need the *by reference* indirection. This is not making the code more expensive since both fit into a single CPU register.

lib/ukstore/include/uk/store.h
craciunoiuc: There are macros that use fields from `struct uk_store_folder;` inside the header. Same thing for functions inside the `store.c` file. I would keep it like this to avoid complications.

lib/ukstore/include/uk/store.h
craciunoiuc: Same thing here. There are macros that use fields from `struct uk_store_folder_entry;` inside the header. I would keep it like this to avoid complications.

lib/ukstore/include/uk/store.h
craciunoiuc: Sure, if you think it makes things easier.

lib/ukstore/store.c
```diff
@@ -34,6 +34,90 @@
...
+{
+	return e->set.uptr(*val);
+}
+
+int uk_store_get_u8(struct uk_store_entry *e, __u8 *out)
+{
+	*out = e->get.u8();
+	return 0;
+}
+int uk_store_get_s8(struct uk_store_entry *e, __s8 *out)
```
skuenzer: Similar to the setters you could do some checks that the target type is big enough to take the number.

``` c
int uk_store_get_s8(struct uk_store_entry *e, __s8 *out)
{
   int ret;

   UK_ASSERT(e);

   switch (e->type) {
   /* ... */

   case s64:
      __s64 val;
      ret = e->get.s64(&val);
      if (ret < 0)
         return ret;
      if (unlikely((val > __S8_MAX) || (val < __S8_MIN)))
         return -ERANGE;
      *out = (__s8) val;
      return ret;

   /* ... */
}

**Review by skuenzer — COMMENTED** [review]
Hey, thanks for your work and update. I think you are still working on it right? I provided you some comments already. ;-)

File: lib/ukstore/include/uk/store.h (hunk 15 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Seems fine to differentiate between random pointers and strings.
Although, I don't see why it should also take the size of the string (as it is null terminated) `int uk_store_get_charp(struct uk_store_entry *e, char *buf, size_t buflen);` (taken from a previous comment). Input checks should be done elsewhere I think.

I will add `charp`, but it will be similar to `uptr` (in usage). Is it ok?

File: lib/ukstore/include/uk/store.h (hunk 40 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: I think I will just remove them. I don't know why I checked them in the first place (except to be extra safe).

File: lib/ukstore/store.c (hunk 41 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c]: It seems there will be a ton of duplicate code (get/set with 10 switches with 10 cases each). Other than this it's okay. I will modify it now.

File: lib/ukstore/store.c (hunk 29 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c]: Discussed in a newer thread.

File: lib/ukstore/include/uk/store.h (hunk 15 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: I got what you were saying from another comment. 👍

File: lib/ukstore/include/uk/store.h (hunk 34 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Sure, sounds good, I'll do the edit now.

File: lib/ukstore/include/uk/store.h (hunk 31 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: I could switch the macros to functions that are defined within the C file. But I don't know if it's 100% worth yet. (and move the definitions afterwards)

File: lib/ukstore/include/uk/store.h (hunk 34 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: It seems to me you are suggesting that the entry create function also creates the folder_entry. This is fine too.

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Almost done with this.

File: lib/ukstore/include/uk/store.h (hunk 35 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Almost done with this.

File: lib/ukstore/include/uk/store.h (hunk 31 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: It might be doable, with the new changes.

File: lib/ukstore/include/uk/store.h (hunk 32 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: It might be doable, with the new changes.

File: lib/ukstore/include/uk/store.h (hunk 31 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Done.

File: lib/ukstore/include/uk/store.h (hunk 32 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Done.

File: lib/ukstore/include/uk/store.h (hunk 15 above)
Comment (skuenzer) [reply lib/ukstore/include/uk/store.h]: I think you should be prepared for the worst case. The reason is that your library will likely be included by many other libraries that we start integrating. They may have own or (more likely) no naming conventions. This means that a clash becomes much more likely. I am fine in calling them for instance `_LIB_UKSTORE__s32` so that they are inline with `uklibparams` but we should namespace.

**Comment by craciunoiuc** [comment]
Hey @skuenzer,

I implemented all the things you suggested and I restructured the commits. Please make another pass and check everything. I would suggest to be thorough on the functions containing switches (there is a high chance I missed something with that much repetition).

Cezar

File: lib/ukstore/include/uk/store.h [hunk 44]
```diff
@@ -29,3 +29,404 @@
...
+ * Removes the dynamic folder from the library and frees its memory and
+ * the memory of the entries.
+ *
+ * @param folder the folder to free
+ */
+#define uk_store_free_folder(folder)	\
+	_uk_store_free_folder((folder))
+
+void
+_uk_store_free_folder(struct uk_store_folder **folder);
```
skuenzer, line 368: Hum... Try to keep the reverse workflow for free'ing similar to adding a folder.
For creating, you do (1) `alloc` and then (2) `add`.
Freeing I would split: (1) `rm/del` and then (2) `free`.
With this, you keep corresponding pairs. To split it makes sense to keep atomicy on ukstore. You can now imagine that you add a folder only when your are done with populating all entries.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * @param libname the name of the library to search in
+ * @param foldername the name of the folder to search in
+ * @param entryname the name of the entry to search for
+ * @return the found entry or NULL
+ */
+#define uk_store_get_entry(libname, foldername, entryname)	\
+	_uk_store_get_entry(uk_store_libid(libname), foldername, entryname)
+
+void
+_uk_store_free_entry(struct uk_store_entry **entry);
```
skuenzer: Name it `release` instead of `free`? Maybe we should add another call to remove an entry from the folder (like `rm/del`)? The `release` would be corresponding to `get/find` and `rm/del` would correspond to `create`? `release` would then only free the memeory when there is no reference to a folder anymore: This could happen either the folder was deleted or the entry was already requested to be deleted... what do you think?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * @param library_id the id returned by uk_store_libid
+ * @return the folder searched for
+ */
+#define uk_store_get_folder(library_id, folder_name)	\
+	_uk_store_get_folder(library_id, folder_name)
+
+struct uk_store_folder *
+_uk_store_get_folder(unsigned int library_id, const char *name);
+
+struct uk_store_entry *
```
skuenzer: Can we return it with `const struct uk_store_entry *`?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * Decreases the refcount. When it reaches 0, the memory is freed.
+ *
+ * @param entry the entry to free
+ */
+#define uk_store_free_entry(entry)	\
+	_uk_store_free_entry((entry))
+
+
+/**
+ * Saves a new getter function in the entry
```
skuenzer: Did you use `update_getter` and `update_setter` already? If not, I am not sure if it makes sense keeping it. They can be dangerous in terms of race condition. Imagine one context reading while another is updating the pointers. Atomicity would need to be guaranteed. But I think we can live without updating uk_store_entries, this means that we can return a store_entry as `const` always.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+#define uk_store_update_setter(entry, e_type, func)			\
+	do {								\
+		if ((entry)->type != uk_store_get_entry_type(e_type))	\
+			(entry)->get.e_type = NULL;			\
+		(entry)->type = uk_store_get_entry_type(e_type);	\
+		(entry)->set.e_type = func;				\
+	} while (0)
+
+int
+_uk_store_get_u8(struct uk_store_entry *e, __u8 *out);
```
skuenzer: the struct argument can be `const struct uk_store_entry *e` here, too

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * @param entry the entry to call the getter from
+ * @param outtype the type of the variable used
+ * @param out space to store the result of the call
+ * @return 0 or error on fail
+ */
+#define uk_store_get_value(entry, outtype, out)	\
+	_uk_store_get_ ## outtype((entry), (out))
+
+int
+_uk_store_set_u8(struct uk_store_entry *e, __u8 val);
```
skuenzer: `const struct uk_store_entry *e`?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+int
+_uk_store_get_s32(struct uk_store_entry *e, __s32 *out);
+int
+_uk_store_get_u64(struct uk_store_entry *e, __u64 *out);
+int
+_uk_store_get_s64(struct uk_store_entry *e, __s64 *out);
+int
+_uk_store_get_uptr(struct uk_store_entry *e, __uptr *out);
+int
+_uk_store_get_charp(struct uk_store_entry *e, char **out);
```
skuenzer: Hum... the problem with such a getter is the memory allocation for the result. Who owns it, who should free it again?
In case the user hands over a buffer where the string should be written to (which I think would make sense), we should handover a size argument (and maybe the function mallocs something when size is set to 0 and hands it over to the caller?). I would do `get_charp` separately from the other getters because I expect a different signature because of a size argument.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if !CONFIG_LIBUKSTORE
+
+#define uk_store_libid(libname) (-1)
+
+#define uk_store_dynamic_entry_create(folder, name, type, get, set)	\
```
skuenzer: If you use empty `static inline` functions you can keep the compiler doing all type checks even when the library is off.

skuenzer: I just noticed that this probably makes only sense for the registration of all static entries. Dynamic entries probably should not be defined at all since the API requires an interaction (like saving a reference and then interact with it). It probbaly makes sense to not provide stubs for them. You could add a comment for code that uses dynamic entries: Such code should be guarded with `#if CONFIG_LIBUKSTORE`. Actually an option that enables/disables just dynamic entries and keeps the static ones is also useful. You may want to add another preprocessor value that code should check for, e.g., `UK_STORE_DYNAMIC`.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+/**
+ * Returns the id of the place it was called from.
+ *
+ * @return this libraries' id
+ */
+#define uk_store_libid_self()	\
+	uk_store_libid(__LIBNAME__)
+
+/* All basic types that exist - use these when interacting with the API */
+enum uk_store_entry_type {
```
skuenzer, line 128: You may need to make this defintion available independent if the lib is selected or not: Just to make the stubbed entries working and type-checked.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+#include <uk/config.h>
+#include <uk/assert.h>
+#include <uk/list.h>
+#include <uk/arch/atomic.h>
+#include <uk/essentials.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
skuenzer, line 47: You may want to add a struct prototype here: Independent if the library is there or not
```c
struct uk_store_entry;
```
This is for adding the empty stubs that do type checking later for the static entries.

lib/ukstore/include/uk/store.h
skuenzer: So, I expect the commit introducing the static entries may add stubs here.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+	char *name;
+
+	/* Entry flags */
+	int flags;
+
+	/* Entry getter/setter type */
+	enum uk_store_entry_type type;
+} __align8;
+
+/* Used internally for storage */
```
skuenzer, line 215: Did we have an API that uses `struct uk_store_folder_entry`? - If not we can jsut keep it within the C file.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+struct uk_store_entry *
+		_uk_store_dynamic_entry_create_charp(
+		struct uk_store_folder *folder, const char *name,
+		uk_store_get_charp_func_t get, uk_store_set_charp_func_t set);
+
+/**
+ * Creates a new dynamic entry
+ *
+ * @param folder the folder where to add the entry
+ * @param name the name of the new entry
```
skuenzer: Add some words regarding memory management of this string. Probably we dup it wihtin the lib

lib/ukstore/include/uk/store.h
skuenzer: Btw, is there also a refcounting for folders?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+	_uk_store_get_folder(library_id, folder_name)
+
+struct uk_store_folder *
+_uk_store_get_folder(unsigned int library_id, const char *name);
+
+struct uk_store_entry *
+_uk_store_get_dynamic_entry(unsigned int libid, const char *f_name,
+				const char *e_name);
+
+static inline struct uk_store_entry *
```
skuenzer: `const`?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+int
+_uk_store_set_s32(struct uk_store_entry *e, __s32 val);
+int
+_uk_store_set_u64(struct uk_store_entry *e, __u64 val);
+int
+_uk_store_set_s64(struct uk_store_entry *e, __s64 val);
+int
+_uk_store_set_uptr(struct uk_store_entry *e, __uptr val);
+int
+_uk_store_set_charp(struct uk_store_entry *e, char *val);
```
skuenzer: `const char *val`?

lib/ukstore/Makefile.uk
```diff
@@ -0,0 +1,34 @@
...
+
+LIBUKSTORE_PHONY_SRC     := $(LIBUKSTORE_BUILD)/uk_libs.in
+LIBUKSTORE_PHONY_SRC_NEW := $(addsuffix .new, $(LIBUKSTORE_PHONY_SRC))
+
+UK_PREPROCESS-$(CONFIG_LIBUKSTORE) += $(LIBUKSTORE_PHONY_SRC) $(LIBUKSTORE_GEN_SRC)
+
+$(LIBUKSTORE_PHONY_SRC): %: %.new
+	@cmp -s $^ $@; if [ $$? -ne 0 ]; then cp $^ $@; fi
+
+$(LIBUKSTORE_BUILD)/uk_libs.in.new:
```
skuenzer, line 25: I was recently wondering if we want to move the library ID routines to an own lib `lib/uklibid`. I can imagine more places where we want to solve  library IDs into names and vice versa (e.g., for the per-library memory allocation stats, debug prints (`lib/ukdebug`) could save a lot of strings in the code). You could split it as part of your patch series already or we do it later after we have your patches upstream. It depends what you prefer.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	return 0;
+}
+
+/**
+ * Frees a folder and all its contents
+ *
+ * @param folder pointer to the folder to free
+ */
+void
+_uk_store_free_folder(struct uk_store_folder **folder)
```
skuenzer, line 210: Why the double reference?

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+}
+
+/**
+ * Creates a folder
+ *
+ * @param name the folder to create
+ * @return the created folder
+ */
+struct uk_store_folder *
+_uk_store_dynamic_folder_create(const char *name)
```
skuenzer: You could add `struct uk_alloc *a` to allow specifying an own allocator. This could be done for optimizations later. You would need to save the reference on the struct itself so that the free operation can work without.

skuenzer: ...similar to `uk_store_folder_entry`

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+_uk_store_free_folder(struct uk_store_folder **folder)
+{
+	struct uk_store_folder_entry *iter, *sec;
+
+	uk_list_del(&(*folder)->folder_head);
+	free((*folder)->name);
+
+	uk_list_for_each_entry_safe(iter, sec,
+				&(*folder)->entry_head, list_head) {
+		uk_list_del(&iter->list_head);
```
skuenzer, line 219: We need to be very careful with freeing. Actually, we should call the entry free operation. This one should reduce its refcount and free itself when it became 0. An entry should also release itself from the folder list (if the folder still exists). For this purpose an `uk_store_folder_entry` should have a reference to is folder. When we remove the folder before an entry, we should set the entry back-reference to `NULL` and just decrease the refcount.

skuenzer: Do we do refocunting for a folder, too? In principle we should...

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+/**
+ * Searches for a folder in a library. Then search for an entry in the folder.
+ * Increases the refcount.
+ *
+ * @param libid the library to search in
+ * @param f_name the name of the folder to search for
+ * @param e_name the name of the entry to search for
+ * @return the found entry or NULL
+ */
+struct uk_store_entry *
```
skuenzer: Think about if we can return the entry as `const`

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+
+	/* Entry getter/setter type */
+	enum uk_store_entry_type type;
+} __align8;
+
+/* Used internally for storage */
+struct uk_store_folder;
+struct uk_store_folder_entry;
+
+struct uk_store_entry *
```
skuenzer: Can we make them all return `const struct uk_store_entry *e`?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * Creates a new dynamic folder
+ *
+ * @param name the name of the new folder
+ * @return pointer to the created folder
+ */
+#define uk_store_dynamic_folder_create(name)	\
+	_uk_store_dynamic_folder_create((name))
+
+struct uk_store_folder *
+_uk_store_dynamic_folder_create(const char *name);
```
skuenzer: Maybe call the `create` for the folders `alloc` so that we differentiate their interaction: `create` on entries do already register them to a folder. Folder need to be explicitly added to the library - which makes sense for atomic reasons.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	struct uk_list_head folder_head;
+
+	/* List for the entries in the folder */
+	struct uk_list_head entry_head;
+
+	/* The folder name */
+	char *name;
+} __align8;
+
+struct uk_store_folder_entry {
```
skuenzer, line 54: I would save a reference to `uk_store_folder` in order to implement the `rm/del` and `release` logic. As well as the case  when a folder is deleted while a entry still has a reference.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+typedef int (*uk_store_set_u16_func_t)(__u16);
+typedef int (*uk_store_set_s32_func_t)(__s32);
+typedef int (*uk_store_set_u32_func_t)(__u32);
+typedef int (*uk_store_set_s64_func_t)(__s64);
+typedef int (*uk_store_set_u64_func_t)(__u64);
+typedef int (*uk_store_set_uptr_func_t)(__uptr);
+typedef int (*uk_store_set_charp_func_t)(char *);
+
+struct uk_store_entry {
+	/* Function getter pointer */
```
skuenzer, line 174: Another `void *cookie;` field would  be especially useful for dynamic entries to store a reference when the same getter and setter function is registered.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+	_LIB_UK_STORE___u64   = 8,
+	_LIB_UK_STORE___uptr  = 9,
+	_LIB_UK_STORE___charp = 10
+};
+
+/* Transforms a simple data type to a ukstore specifice one */
+#define uk_store_get_entry_type(simple_type)	\
+	(_LIB_UK_STORE___ ## simple_type)
+
+/* Getter definitions */
```
skuenzer, line 146: We should add a cookie to allow saving a reference for a library. This one should be saved on the entry and should automatically get hand-over to the getter and setters:
```c
typedef int (*uk_Store_get_s8_func_t)(void *, __s8 *);
```

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+ * @param folder the folder to free
+ */
+#define uk_store_free_folder(folder)	\
+	_uk_store_free_folder((folder))
+
+void
+_uk_store_free_folder(struct uk_store_folder **folder);
+
+/**
+ * Get a dynamic folder from a library
```
skuenzer, line 384: You should also describe that all entries are deleted.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	/* Saved entry pointer */
+	struct uk_store_entry entry;
+
+	/* The linked list head */
+	struct uk_list_head list_head;
+
+	/* Refcount */
+	__atomic refcount;
```
skuenzer, line 71: Hum... since getters and setters can still be called while we requested deleting (refcount > 0 but entry deleted  from folder or folder deleted), we have two options:
(1) We check on the getters and setters if the entry was already deleted and return -ENOENT. This requires making getters, setters, and delete functions calls to be mutual exclusive (a mutex lock within uk_store_entry).
(2) We add here another callback that  is called as soon the entry is finally deleted (refcount became zero):
```c
typedef void (*uk_store_entry_freed_func_t)(void *);
```
The extra argument is the user-provided cookie. This way, final clean-up can be done (and getters/setters keep working, or the library does ENOENT itself) until really released.

I think (1) is easier for one to use but we would need to make sure that we do not have macros that directly allow calling the callbacks with byassing the mutex. However (2) would is more lightweight, potentially performs faster but is more complicated to implement on the library that provides dynamic entries.
I am not sure which one I would prefer at the moment. What do you think?

Independent to this, I think we should anyways add later an option to register a callback to entries to get notified when a folder/entry is deleted or modified. This allows implementing something like a procfs easily and that can be kept in sync. So, in general, (2) would be an approach that is fine for such a low-level store library.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	SETCASE_UPCAST(e, s32, S32, val, u8);
+	SETCASE_UPCAST(e, u64, U64, val, u8);
+	SETCASE_UPCAST(e, s64, S64, val, u8);
+	SETCASE_UPCAST(e, uptr, PTR, val, u8);
+
+	case uk_store_get_entry_type(charp): {
+		int ret;
+		char to_set[UK_STORE_NR_MAX_LEN];
+
+		sprintf(to_set, "%hhu", val);
```
skuenzer: We have general applicable scan and print formats defined. You should preferably use those since u, d etc are not necessarily the same size across architectures:
snprintf(to_set, `"%"__PRu16`, val)

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	_uk_store_dynamic_entry_create_ ## gen_type(			\
+		struct uk_store_folder *folder,				\
+		const char *name,					\
+		uk_store_get_ ## gen_type ## _func_t get,		\
+		uk_store_set_ ## gen_type ## _func_t set)		\
+	{								\
+		struct uk_store_folder_entry *new_folder_entry;		\
+									\
+		UK_ASSERT(folder || name);				\
+									\
```
skuenzer: Why two blank lines?

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	default:
+		return -EINVAL;
+	}
+	return 0;
+}
+
+int
+_uk_store_set_charp(struct uk_store_entry *e, char *val)
+{
+	return _uk_store_set_s64(e, atoi(val));
```
skuenzer: Oh, I think this could be to easy. u64 would not be fully covered. Can you try to use `sscanf` and the appropriate `__SCN*` format definitions?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,404 @@
...
+	_LIB_UK_STORE___u32   = 6,
+	_LIB_UK_STORE___s64   = 7,
+	_LIB_UK_STORE___u64   = 8,
+	_LIB_UK_STORE___uptr  = 9,
+	_LIB_UK_STORE___charp = 10
+};
+
+/* Transforms a simple data type to a ukstore specifice one */
+#define uk_store_get_entry_type(simple_type)	\
+	(_LIB_UK_STORE___ ## simple_type)
```
skuenzer, line 144: If you remove the update getter&setter you can move this definiton into the C file

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+	GETCASE_DOWNCASTUU(e, uptr, U8, out, u8);
+
+	case uk_store_get_entry_type(charp): {
+		char *val = NULL;
+		__s64 nr;
+
+		ret = e->get.charp(&val);
+		if (ret < 0)
+			return ret;
+		nr = atoi(val);
```
skuenzer: Can you use `sscanf` and the appropriate`__SCN*` defintions? An s16 should be also enough to detect an ERANGE.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+		return -ENOMEM;
+
+	switch (e->type) {
+	case uk_store_get_entry_type(u8): {
+		__u8 val;
+
+		ret = e->get.u8(&val);
+		if (ret < 0)
+			return ret;
+		sprintf(str, "%hhu", val);
```
skuenzer: Use `__PRIu8` etc.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1194 @@
...
+
+int
+_uk_store_get_charp(struct uk_store_entry *e, char **out)
+{
+	int ret;
+	char *str;
+
+	UK_ASSERT(e);
+
+	str = calloc(UK_STORE_NR_MAX_LEN, sizeof(char));
```
skuenzer: You probably should allocate within the cases and have as many digits a number can maximum have including the sign. If we enable user provided buffers with size, we can use `snprintf`

lib/ukstore/include/uk/store.h
```diff
@@ -301,10 +373,14 @@ _uk_store_get_folder(unsigned int library_id, const char *name);
 struct uk_store_entry *
 _uk_store_get_dynamic_entry(unsigned int libid, const char *f_name,
 				const char *e_name);
+struct uk_store_entry *
+_uk_store_get_static_entry(unsigned int libid, const char *e_name);
 
 static inline struct uk_store_entry *
 _uk_store_get_entry(unsigned int libid, const char *f_name, const char *e_name)
 {
+	if (!f_name)
```
skuenzer, line 406: You should also add checks on the release/free functions when it is a static entry. delete should fail, release can be totally ignored since we do not do refcounting.

lib/ukstore/include/uk/store.h
```diff
@@ -45,9 +45,41 @@ extern "C" {
 #endif
 
 #if !CONFIG_LIBUKSTORE
-
 #define uk_store_libid(libname) (-1)
 
+#define __UK_STORE_STATIC_ENTRY(entry, lib_str, e_type, e_get, e_set)	\
+	static const struct uk_store_entry				\
+	__used __section(".uk_store_lib_" lib_str) __align8		\
```
skuenzer: If the UK_STORE library is not there, you can still do this but mark it as __unused and do not put it in a special section.

skuenzer (changes requested):
Hey, thanks a lot for your work. The series looks really good. I have comment for finer details of the API. We especially need to re-visit some of the function to make them safe in respect of race conditions (e.g., an entry is deleted but still accessed). Maybe we also want to allow custom allocators to be used... this would be inline with other Unikraft native libs. Let me know what you think.

Thanks!

lib/ukstore/store.c
```diff
@@ -0,0 +1,1246 @@
...
+
+/* Generate a function creator for each type */
+_UK_STORE_DYNAMIC_CREATE_TYPED(u8);
+_UK_STORE_DYNAMIC_CREATE_TYPED(s8);
+_UK_STORE_DYNAMIC_CREATE_TYPED(u16);
+_UK_STORE_DYNAMIC_CREATE_TYPED(s16);
+_UK_STORE_DYNAMIC_CREATE_TYPED(u32);
+_UK_STORE_DYNAMIC_CREATE_TYPED(s32);
+_UK_STORE_DYNAMIC_CREATE_TYPED(u64);
+_UK_STORE_DYNAMIC_CREATE_TYPED(uptr);
```
nderjung: Missing `s64`:
```diff --git a/lib/ukstore/store.c b/lib/ukstore/store.c
index c0bc586..dd841b7 100644
--- a/lib/ukstore/store.c
+++ b/lib/ukstore/store.c
@@ -308,6 +308,7 @@ _UK_STORE_DYNAMIC_CREATE_TYPED(s16);
 _UK_STORE_DYNAMIC_CREATE_TYPED(u32);
 _UK_STORE_DYNAMIC_CREATE_TYPED(s32);
 _UK_STORE_DYNAMIC_CREATE_TYPED(u64);
+_UK_STORE_DYNAMIC_CREATE_TYPED(s64);
 _UK_STORE_DYNAMIC_CREATE_TYPED(uptr);
 _UK_STORE_DYNAMIC_CREATE_TYPED(charp);
 ```
 
 Caught by @flapjack321

File: lib/ukstore/include/uk/store.h (hunk 47 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: They have a limited use case, true. I guess I'll just remove them.

File: lib/ukstore/include/uk/store.h (hunk 73 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:144]: This is also used by `__UK_STORE_STATIC_ENTRY` in the header. If I move it that will stop working

File: lib/ukstore/store.c (hunk 79 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c]: Phew, thanks!

File: lib/ukstore/include/uk/store.h (hunk 54 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:215]: It is used by `__UK_STORE_STATIC_ENTRY`. I don't think it can be moved to the C file.

File: lib/ukstore/include/uk/store.h (hunk 44 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:368]: No refcounting for folders. I kept the double pointer to mark it as `NULL` to ensure no uses after free.

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:368]: Also, it will be a bit harder to split the `free` into two completely separate parts (`remove`/`free`) as entries that are removed will be lost (for freeing) if they are not freed in the same iteration.

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:368]: For this I will move the `uk_list_del` of the folder inside the `remove` and keep the rest inside the `free`. Does this sound good?

File: lib/ukstore/include/uk/store.h (hunk 45 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Sounds good. Please check again after the fixes are done that I did it correctly.

File: lib/ukstore/store.c (hunk 59 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c:210]: To mark it as `NULL` when freeing to guard against its usage after freeing. It is optional, I can remove it if you think it is necessary to do so

File: lib/ukstore/Makefile.uk (hunk 58 above)
Comment (craciunoiuc) [reply lib/ukstore/Makefile.uk:25]: I think it is easier to split all the ID-using libraries later (when creating uklibid) 🙂

File: lib/ukstore/include/uk/store.h (hunk 51 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: True. Though, I wouldn't do the dynamic/static separation. In the last fixes we made the access to static/dynamic entries seamless. It would be strange to separate them again. Also, i think it gets cumbersome to use both `CONFIG_LIBUKSTORE` and `CONFIG_LIBUKSTORE_DYNAMIC`.

The other suggestions are all good.

File: lib/ukstore/include/uk/store.h (hunk 66 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:174]: I will do this, but I don't really get what it will be used for. Please check that I implemented it correctly

File: lib/ukstore/include/uk/store.h (hunk 67 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:146]: I don't get this either, but I will try to implement it

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h:146]: So now the saved getters/setters need to receive one extra argument, which is the cookie

File: lib/ukstore/store.c (hunk 61 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c:219]: I don't think refcounting for folders is necessary

File: lib/ukstore/include/uk/store.h (hunk 50 above)
Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: I agree, I asked myself this too. But, by changing the signature it will kind of break the design. Maybe it is easier to just add a comment for the getter to ensure that the caller does not give a buffer, but frees the one received afterwards.

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Right now the user should give a pointer and Ukstore will allocate a buffer to it which needs to be freed afterwards.

Comment (craciunoiuc) [reply lib/ukstore/include/uk/store.h]: Do you think it is better to switch to user given buffer+size? (is it worth?)

File: lib/ukstore/store.c (hunk 69 above)
Comment (craciunoiuc) [reply lib/ukstore/store.c:71]: I agree, I'll try to do the second one. Please check that I did it correctly

Comment (craciunoiuc) [reply lib/ukstore/store.c:71]: To sum it up, the dynamic entries will take another parameter (a cleanup function). This will be called right before an entry is removed and freed.

What I don't get is how will this solve the use-after-free problem? (when an entry is still available but it does not belong to a folder anymore)

**Comment by craciunoiuc** [comment]
Hey @skuenzer,

I did most of the fixes you suggested. Some items still remain. Please go through the discussions marked as resolved (and the ones that are not) and check my answers/questions there.

Thanks for the review!

**Comment by craciunoiuc** [comment]
Hey @skuenzer,

I think it's time to get back to this Please answer the old questions above and provide additional feedback.

PS: I will fix the typo from checkpatch don't worry

Cezar

**Comment by marcrittinghaus** [comment]
Hi @craciunoiuc,

I am looking at `ukstore` now the first time (as a user, not reviewer 😉). Could you give a short example on how you would use the functionality to publish a value and then access it with the API? I looked at #282 but don't see how `ukstore` would eventually fit in there. Also, concerning static values: for example, if I would do a HW probe at boot time to let's say get the processor model with `cpuid`. Could this be a static field or is static compile-time static and I would need a dynamic one?

Thank you very much! Hope we get this in soon. Looks good! 👍🏻

**Comment by craciunoiuc** [comment]
Hey @marcrittinghaus,

I have a usage example somewhere that I can post soon.

Inside the `mutex` statistics the only things to add will be a macro call (or multiple ones) to register the entry (like you register a `ctor`) and another getter (or multiple ones) to retrieve the data in the format of  `ukstore`.
This was there before, but I asked for it to be removed so that we can have the metrics in early.

As for the use case you presented:
 * Static fields inside `ukstore` are compile-time static. Basically each entry is a structure inside their section for that file. 
Of course, the access is done at run-time, but everything regarding "the setup" is done at compile time.
 * Dynamic fields have a behavior almost identic to the linked list inside unikraft (it actually uses these lists)
The dynamic lists have this form:
```
folder_head --> folder1 --> folder2 --> folder3 --> ...
                   |          |             |
                   |          |             |
                   V          V             V
                 entry11     entry21      entry31
                   |          |
                   |          |
                   V          V
                 entry12     entry22 
```

craciunoiuc:
Hey @marcrittinghaus,

I finally found the usage example (it was hidden inside an `archive.tar.gz` :laughing: )

It basically covers all the usual use cases for `ukstore`.

I pasted it here, let me know what you think: https://pastebin.com/119GYzNY

marcrittinghaus:
> Hey @marcrittinghaus,
> 
> I finally found the usage example (it was hidden inside an `archive.tar.gz` 😆 )
> 
> It basically covers all the usual use cases for `ukstore`.
> 
> I pasted it here, let me know what you think: https://pastebin.com/119GYzNY

Hi @craciunoiuc,

thank you very much! Will look into it 😃

craciunoiuc:
This is going to be split up into 2 `PRs`: one for static entries and one for dynamic. This PR will become the dynamic one.

craciunoiuc:
Actually, I'll move this as a Draft.

razvand:
This PR has been split into a static side PR (#459) and a dynamic side PR (#460).

PR #211, opened 2021-05-18
Support dumping backtrace in Unikraft

### LINKED ISSUE #210: Support backtrace dumping for debugging purpose

linking, 3rd linking is used.

justin-he:


nderjung:
Hi @justin-he,

Thanks for this PR! This is super useful feature, I'm sure it will end up helping many Unikraft developers! :)

Just a few notes regarding licensing and copyright from my side after quickly glancing over this PR:

 * Please can you remove any instances of `THIS HEADER MAY NOT BE EXTRACTED OR MODIFIED IN ANY WAY.` in header license preambles as [this line is a mistake](https://github.com/unikraft/unikraft/commit/7fd6797bd5917acc515ef6ddbfa85621f4aacf5f);
 * I just want to confirm with @skuenzer or @felipehuici if there a licensing issue here since it includes code from the Linux kernel, which is licensed under GPL-2.0, namely the new file: `support/kallsyms/kallsyms.c`?

justin-he:
> Hi @justin-he,
> 
> Thanks for this PR! This is super useful feature, I'm sure it will end up helping many Unikraft developers! :)
> 
> Just a few notes regarding licensing and copyright from my side after quickly glancing over this PR:
> 
> * Please can you remove any instances of `THIS HEADER MAY NOT BE EXTRACTED OR MODIFIED IN ANY WAY.` in header license preambles as [this line is a mistake](https://github.com/unikraft/unikraft/commit/7fd6797bd5917acc515ef6ddbfa85621f4aacf5f);
> * I just want to confirm with @skuenzer or @felipehuici if there a licensing issue here since it includes code from the Linux kernel, which is licensed under GPL-2.0, namely the new file: `support/kallsyms/kallsyms.c`?

That is a good question! Thanks Simon or Felipe for making  it clear. I remembered there was a policy that unikraft building *can* use the GPLv2. Is it correct?

marcrittinghaus:
I already integrated symbol resolution into the bigger debugging update (GDB stub etc.) I am currently working on, that will also provide a uniform crash screen for x86-64 and arm64. That code does not borrow from Linux. I will look into the patch and see if we can merge both patches so that we do not have copyright issues but all the great stuff from both our contributions.

@justin-he 
You can take a look at the current status for the debugging stuff in my dev branch ([https://github.com/marcrittinghaus/unikraft/tree/mritting/libukdebug-gdb])

plat/kvm/Linker.uk
```diff
@@ -38,6 +56,57 @@ $(KVM_DEBUG_IMAGE): $(KVM_ALIBS) $(KVM_ALIBS-y) $(KVM_OLIBS) $(KVM_OLIBS-y) \
 		     $(KVM_LDFLAGS) $(KVM_LDFLAGS-y) \
 		     $(KVM_LD_SCRIPT_FLAGS) \
 		     $(KVM_IMAGE).o -o $@)
+# If dumping backtrace is supported, we should link the image 3 times after
+# getting the symbol name assembly src file.
+ifeq ($(CONFIG_DEBUG_BACK_TRACE),y)
+# 2nd time link
+	$(NM) -n $@ | $(ALLSYMS_TOOL) > $(SYMBOL_SRC)
+	$(HOSTCC) -I$(CONFIG_UK_BASE)/include $(HOST_EXTRACFLAGS) $(HOSTCFLAGS) -c $(SYMBOL_SRC) -o $(SYMBOL_OBJ)
```
razvanvirtan, line 64: I think we have an issue here, because using `$(HOSTCC)` will cause build failure on any application that enables backtrace support, if we try to cross-compile on x86 systems. $(HOSTCC) translates to `/usr/bin/gcc`, which works fine when we build the image on an aarch64 host. However, when trying to build an aarch64 image on a x86 host, `usr/bin/gcc` will be a x86 compiler and the build process will fail with a linking error like:
``` 
LD      app-helloworld_kvm-arm64.ld.2.o
/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: /home/razvan/Documents/Unikraft/ARM64/backtrace_tests2/apps/app-helloworld/build/libkvmplat/symbol.o: Relocations in generic ELF (EM: 62)
/home/razvan/Documents/Unikraft/ARM64/backtrace_tests2/apps/app-helloworld/build/libkvmplat/symbol.o: error adding symbols: File in wrong format
```
My testcase was app-helloworld, but a similar behaviour will be encountered on any application.

justin-he: Thanks, Razvan.  I will try to set up a cross building environment to fix this. But if you replace $HOST_GCC with the cross-complier, can you please continue your test?

razvanvirtan: Yes, using the cross compiler directly instead of HOSTCC works.

razvanvirtan:
I've taken a look on the debug system work made by @marcrittinghaus and it integrates the backtrace functionality on both x86_64 and ARM64. Also, it doesn't raise licensing issues. Therefore, I think we could close this PR.

justin-he:
closed as this is superceded

PR #220, opened 2021-06-01
lib/nolibc: Add strncat

vladandrew:

### DIFF
--- a/lib/nolibc/string.c
+++ b/lib/nolibc/string.c
@@ -686,3 +686,20 @@ char *strerror(int errnum)
 
 	return strerror_r(errnum, buf, sizeof(buf));
 }
+
+char *strncat(char *dest, const char *src, size_t n)
+{
+	char *a = dest;
+
+	dest = dest + strlen(dest);
+
+	if (src != NULL) {
+		while (n && *src) {
+			n--;
+			*dest++ = *src++;
+		}
+	}
+
+	*dest++ = 0;
+	return a;
+}
--- a/lib/nolibc/exportsyms.uk
+++ b/lib/nolibc/exportsyms.uk
@@ -75,6 +75,7 @@ strlcpy
 strlcat
 strerror_r
 strerror
+strncat
 
 # strsignal
 strsignal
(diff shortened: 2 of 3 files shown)


lib/nolibc/string.c
```diff
@@ -686,3 +686,17 @@ char *strerror(int errnum)
 
 	return strerror_r(errnum, buf, sizeof(buf));
 }
+
+char *strncat(char *dest, const char *src, size_t n)
+{
+	char *a = dest;
+
```
skuenzer, line 693: Should we add `UK_ASSERT()`s to catch the case that one parameter is `__NULL`?
What is the standard saying when `n` is `0` and `src` or `dest` is `__NULL`?

skuenzer (changes requested):

lib/nolibc/string.c
vladandrew: According to the standard, If dest is not large enough, program behavior is unpredictable. Basically, we don't have to add any checks for `dst`. Although. maybe in the future it would make sense to have a hardened C library with such checks.

The case where `src` is `NULL` has defined behaviour since the while loop does not execute. (e.g. you only execute *dest++ = 0;)

skuenzer: Undefined behavior just means that behavior is not defined by manual, meaning the function could even cause a crash, like `UK_ASSERT()` would do. However, the other functions in this file  do not do a `NULL` check, so it would be out of style to add it  here.
Regarding the while loop, if behavior `src == NULL` is  documented, you still need to add a  check for `src`: In your  code right now you are de-referencing `src` which would cause undefined behavior.

skuenzer (changes requested):

vladandrew:
Hi Simon,

Right, I've added a check for `src` such that we don't end dereferecing `NULL`. What do you think about this new updated version?

lib/nolibc/string.c
vladandrew: Hi Simon,

Thanks for the feedback. Right, I've added a check for src such that we don't end dereferencing NULL. What do you think about this new updated version?

skuenzer (approved):
Looks good to me, thanks!

PR #222, opened 2021-06-02
build: Option to treat compiler warnings as errors

skuenzer:
Introduces an option to treat compiler warnings as errors. The build
system stops compilation if a compiler warning is displayed. This
option is intended to ensure clean builds.


nderjung:
This is about to break many builds ;)

skuenzer:
Yes, if you enable it, it does break some builds. However, as default the option is off so we don't get any issues if we integrate this patch now. The idea is that the CI/CD system sets the option as part of the patch tests - of course after someone provided clean-up patches to make our base compile without warnings - which we anyways should always aim.

nderjung (changes requested):
Hey thanks for this, can you please rebase?

razvand (approved):

nderjung (approved):
Reviewed-by: Alexander Jung <alex@unikraft.org>

PR #226, opened 2021-06-06
lib/uknetdev: Add network interface instrumentation

craciunoiuc:
This commit series adds support for instrumenting the network
interfaces in order to extract metrics from them. The metrics extracted
follow the fields from `/proc/net/dev`. Right now, only the most common
ones are implemented. The rarer ones will be added later.

Interaction with the metrics is thread-safe and the overhead from the
instrumentation should be acceptable.

In order to enable the metrics, a new entry was added inside the
`uknetdev` `KConfig` option.

### DIFF
--- a/lib/uknetdev/Config.uk
+++ b/lib/uknetdev/Config.uk
@@ -30,4 +30,10 @@ if LIBUKNETDEV
 			When this option is enabled a dispatcher thread is
 			allocated for each configured receive queue.
 			libuksched is required for this option.
+	config LIBUKNETDEV_METRICS
+		bool "Interface metrics for statistics"
+		default n
+		help
+			The metrics instrumented are the same with the ones from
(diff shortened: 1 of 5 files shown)


craciunoiuc:
Needs 2 lines changed after fccc041

craciunoiuc:
@skuenzer 

I have a rebased/recreated PR offline using atomic variables instead of spinlocks.
Let me know if you would prefer that and I can push it.

I'm not sure of the good version. It seems that spinlocks are actually faster than atomic variables:
https://news.ycombinator.com/item?id=17338767

mandrei12:
Looks good to me. @skuenzer , could you take a look?

craciunoiuc:
@skuenzer 

Ping :tennis:

craciunoiuc:
Closed in favor of #943

PR #227, opened 2021-06-11
include/event: Add uk_event for new trap interface

### LINKED ISSUE #210: Support backtrace dumping for debugging purpose

marcrittinghaus:
This PR adds the uk_event object that is used in the extensible trap interface.

This PR is the first part of a larger trap/debugging update. Please see the attached presentation for more information (also on how uk_event works): [ukdbg.pdf](https://github.com/unikraft/unikraft/files/6639916/ukdbg.pdf)


marcrittinghaus:
Checkpatch seems unhappy with it, however, performing the suggested changes will break the code. Should I redesign the macros?

nderjung:
It's worth double checking with @skuenzer, however, you can trick the bot into ignoring it by adding:
```
Checkpatch-Ignore: COMPLEX_MACRO
```
As a git trailer to the patches which introduce these errors.

marcrittinghaus:
> It's worth double checking with @skuenzer, however, you can trick the bot into ignoring it by adding:
> 
> ```
> Checkpatch-Ignore: COMPLEX_MACRO
> ```
> 
> As a git trailer to the patches which introduce these errors.

Thanks, good to know. I will ask Simon on Monday.

michpappas:
Hi @marcrittinghaus, I was playing around with this one on aarch64 / KVM and experienced some failures due to - what that seems to be - gcc generating invalid code under certain circumstances (yes seriously). I explain below.

Consider a simple application such as:
```
#include <stdio.h>
#include <uk/event.h>

int myhandler(void *data)
{
	(void)data;
	printf("Handled\n");
	return UK_EVENT_HANDLED;
}

UK_EVENT(myevent);
UK_EVENT_HANDLER(myevent, myhandler);

{
	uk_raise_event(myevent, NULL);
}
```

Running this application with qemu triggers the assertion on `*itr`:
```
bash$ qemu-system-aarch64 -M virt -cpu cortex-a57 -nographic -kernel build/app-helloworld_kvm-arm64
[    0.002126] CRIT: [apphelloworld] <main.c @  200> Assertion failure: *itr
```

While debugging I added some `printf()` in `uk_raise_event_ptr`, and experienced a bizarre behavior where the output changes depending on different combinations of function calls. For example, with these changes:
```
	uk_event_handler_foreach(itr, e) {
		printf("0x%x, 0x%x\n", itr, *itr);
		printf("0x%x, 0x%x\n", itr, *itr);
		//__event_assert(*itr);
		ret = ((*itr)(data));
		if (ret)
			break;
	}
```
You get this output:
```
bash$ qemu-system-aarch64 -M virt -cpu cortex-a57 -nographic -kernel build/app-helloworld_kvm-arm64
0x40112010, 0x0
0x40112010, 0x40103b70
Handled
```
But as soon as you enable __event_assert(*itr), it fails again(!) There are other combinations of `printf()`, `__event_assert()`, `volatile` and what not, that cause one result or the other.

Anyway, to get to the point, I had a look at the code generated by gcc. In the working case (not included here) everything looks normal. In the failing case gcc generates an invalid instruction where the handler starts (0x40110b4c):
```
bash$ ~/local/aarch64-linux-gnu-objdump -D build/app-helloworld_kvm-arm64.dbg
...
0000000040110af0 <main>:
    40110af0:   d0000000        adrp    x0, 40112000 <uk_ctortab_end>
    40110af4:   91002000        add     x0, x0, #0x8
    40110af8:   d0000001        adrp    x1, 40112000 <uk_ctortab_end>
    40110afc:   91002000        add     x0, x0, #0x8
    40110b00:   91006021        add     x1, x1, #0x18
    40110b04:   eb01001f        cmp     x0, x1
    40110b08:   54000202        b.cs    40110b48 <main+0x58>  // b.hs, b.nlast
    40110b0c:   a9bf7bfd        stp     x29, x30, [sp, #-16]!
    40110b10:   f0000005        adrp    x5, 40113000 <_rodata>
    40110b14:   f0000004        adrp    x4, 40113000 <_rodata>
    40110b18:   913d00a5        add     x5, x5, #0xf40
    40110b1c:   9123c084        add     x4, x4, #0x8f0
    40110b20:   f0000002        adrp    x2, 40113000 <_rodata>
    40110b24:   f0000001        adrp    x1, 40113000 <_rodata>
    40110b28:   913d2042        add     x2, x2, #0xf48
    40110b2c:   913d4021        add     x1, x1, #0xf50
    40110b30:   528018c3        mov     w3, #0xc6                       // #198
    40110b34:   910003fd        mov     x29, sp
    40110b38:   52800000        mov     w0, #0x0                        // #0
    40110b3c:   97ffce15        bl      40104390 <_uk_printk>
    40110b40:   52800060        mov     w0, #0x3                        // #3
    40110b44:   97ffc73f        bl      40102840 <ukplat_terminate>
    40110b48:   d65f03c0        ret
    40110b4c:   00000000        .inst   0x00000000 ; undefined
    40110b50:   a9bf7bfd        stp     x29, x30, [sp, #-16]!
    40110b54:   90000020        adrp    x0, 40114000 <__utc+0x958>
    40110b58:   91010000        add     x0, x0, #0x40
    40110b5c:   910003fd        mov     x29, sp
    40110b60:   97ffd584        bl      40106170 <printf>
    40110b64:   128002a0        mov     w0, #0xffffffea                 // #-22
    40110b68:   a8c17bfd        ldp     x29, x30, [sp], #16
    40110b6c:   d65f03c0        ret
...
```

Didn't look any further into exactly what triggers gcc to do so (who enjoys fixing gcc bugs :P) but got it working by forcing gcc not to optimize `uk_raise_event_ptr()`:
```
diff --git a/include/uk/event.h b/include/uk/event.h
index 3ab28ed..b9958e1 100644
--- a/include/uk/event.h
+++ b/include/uk/event.h
@@ -187,7 +187,7 @@ struct uk_event {
  * @return
  *   One of the UK_EVENT_* macros on success, errno on < 0
  */
-static inline int uk_raise_event_ptr(const struct uk_event *e, void *data)
+static inline __attribute((optimize("O0"))) int uk_raise_event_ptr(const struct uk_event *e, void *data)
 {
        const uk_event_handler_t *itr;

```

The issue is reproducible when cross-compiling the application with all major releases of gcc between 9.2 and 10.3 (which is the latest version available in developer.arm.com).

michpappas:
@marcrittinghaus I also attach a [zipfile with my setup](https://github.com/unikraft/unikraft/files/7448800/uk_event.zip) in case you wanted to reproduce with minimal effort. Just get a toolchain from developer.arm.com (eg https://developer.arm.com/-/media/Files/downloads/gnu-a/10.3-2021.07/binrel/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf.tar.xz) and update the top-level Makefile with the toolchain location.

marcrittinghaus:
Hi @michpappas,

very cool finding 👍🏻 That makes me really curious what is happening there. I will look into this tomorrow and see if there is anything I have missed. Thank you very much for putting all the effort into the debugging and providing the sample 😃

I know we had some problems in `uktest` with the approach I used here with optimization enabled. The problem there, however, was that the ordering of the sections is not the same for performance and debug builds. I always develop with debug builds so I missed that one in the first prototype. Perhaps we have a similar issue here 🤔 Or it is really a bug in gcc, who knows. But from my experience, the compiler was alright in most cases 😆 

I would like to go over the code once more (rebase, perhaps do some polishing, and fixing of course 😄) and then it would be super to get it upstream together with the trap code changes. I am currently working on the paging code and the trap interface would allow to implement demand paging for the heap and in the future things like copy-on-write.

michpappas:
Hi Marc, you're welcome it's been fun :) I had a closer look at the case where the two `printf()` show different addresses, as I found it to be the most curious one. In the snippet below I only comment the parts relevant to the actual flow.
```
0000000040110af0 <main>:
    40110af0:   a9bd7bfd        stp     x29, x30, [sp, #-48]!
    40110af4:   d0000000        adrp    x0, 40112000 <uk_ctortab_end>
    40110af8:   91006000        add     x0, x0, #0x18                // x0 -> 0x40112018 (past the end of .uk_eventtab)
    40110afc:   910003fd        mov     x29, sp
    40110b00:   a90153f3        stp     x19, x20, [sp, #16]
    40110b04:   d0000013        adrp    x19, 40112000 <uk_ctortab_end>
    40110b08:   91002273        add     x19, x19, #0x8
    40110b0c:   91002273        add     x19, x19, #0x8              // x19 -> _uk_eventtab_myevent_9_myhandler
    40110b10:   eb00027f        cmp     x19, x0                     // if (x19 > x0)
    40110b14:   54000322        b.cs    40110b78 <main+0x88>        //     goto return;

    40110b18:   d1000414        sub     x20, x0, #0x1
    40110b1c:   d2800002        mov     x2, #0x0                    // x2 -> 0
    40110b20:   cb130294        sub     x20, x20, x19

    40110b24:   f90013f5        str     x21, [sp, #32]
    40110b28:   927df294        and     x20, x20, #0xfffffffffffffff8
    40110b2c:   f0000015        adrp    x21, 40113000 <_rodata>
    40110b30:   8b140274        add     x20, x19, x20
    40110b34:   913d02b5        add     x21, x21, #0xf40
    40110b38:   14000004        b       40110b48 <main+0x58>        // jump to first print unconditionally

    40110b3c:   eb14027f        cmp     x19, x20
    40110b40:   540001a0        b.eq    40110b74 <main+0x84>

    /* This should be the beginning of the first print block, but actually 0x40110b38
     * skips the next instruction and jumps directly to 0x40110b48 after setting x2 to 0
     */
    40110b44:   f8408e62        ldr     x2, [x19, #8]!
    40110b48:   aa1303e1        mov     x1, x19                      // arg1 -> _uk_eventtab_myevent_9_myhandler
    40110b4c:   aa1503e0        mov     x0, x21                      // arg0 -> "0x%x, 0x%x"
    40110b50:   97ffd588        bl      40106170 <printf>            // printf(arg0, arg1, arg2)

    /* prints 0x40112010, 0x40103b70 */
    40110b54:   f9400262        ldr     x2, [x19]                    // arg2 -> *_uk_eventtab_myevent_9_myhandler
    40110b58:   aa1303e1        mov     x1, x19                      // arg1 -> _uk_eventtab_myevent_9_myhandler
    40110b5c:   aa1503e0        mov     x0, x21                      // arg0 ->  "0x%x, 0x%x"
    40110b60:   97ffd584        bl      40106170 <printf>            // printf(arg0, arg1, arg2)

    40110b64:   f9400261        ldr     x1, [x19]
    40110b68:   d2800000        mov     x0, #0x0
    40110b6c:   d63f0020        blr     x1                           // myhandler()
    40110b70:   34fffe60        cbz     w0, 40110b3c <main+0x4c>     // ret == UK_EVENT_HANDLED ?
    40110b74:   f94013f5        ldr     x21, [sp, #32]

return:
    40110b78:   a94153f3        ldp     x19, x20, [sp, #16]
    40110b7c:   a8c37bfd        ldp     x29, x30, [sp], #48
    40110b80:   d65f03c0        ret

    /* myhandler() */
    40110b90:   a9bf7bfd        stp     x29, x30, [sp, #-16]!
    40110b94:   90000020        adrp    x0, 40114000 <__utc+0x958>
    40110b98:   9100c000        add     x0, x0, #0x30
    40110b9c:   910003fd        mov     x29, sp
    40110ba0:   97ffd574        bl      40106170 <printf>
    40110ba4:   128002a0        mov     w0, #0xffffffea
    40110ba8:   a8c17bfd        ldp     x29, x30, [sp], #16
    40110bac:   d65f03c0        ret

0000000040112000 <uk_ctortab_end>:
    40112000:   4010f530        .word   0x4010f530
    40112004:   00000000        .word   0x00000000

Disassembly of section .uk_eventtab:

0000000040112008 <uk_event_myevent>:
    40112008:   40112018        .word   0x40112018
    4011200c:   00000000        .word   0x00000000

0000000040112010 <_uk_eventtab_myevent_9_myhandler>:
    40112010:   40103b70        .word   0x40103b70
    40112014:   00000000        .word   0x00000000
```
So unless I'm making some grave mistake on my interpretation, it looks like GCC generates a bunch of nonsense.
One part that makes me suspicious is this:
```
    40110af8:   91006000        add     x0, x0, #0x18                // x0 -> 0x40112018 (past the end of .uk_eventtab)
```
It makes me think it may all be related to alignment, ~~as the entries in `.uk_eventtab` are 32-bits wide, while they should be 64-bits given that they contain pointers, and that's what confuses gcc after all.~~ Edit: Sorry I got confused by objdump's word representation, the entries are actually 64-bit.

marcrittinghaus:
This is crazy! I will now try to reproduce everything on my side.

marcrittinghaus:
Hi @michpappas

yes, I can reproduce the issue. This seems indeed to be a problem with gcc. I get the following output and code for the two `printf` and no assert in the loop:

```
0x40116010, 0x0
0x40116010, 0x40103ee8
Handled
```

```ASM
	const uk_event_handler_t *itr;

	__event_assert(e);

	uk_event_handler_foreach(itr, e) {
    40111084:	b0000020 	adrp	x0, 40116000 <__uk_inittab20_uk_bus_lib_init>
    40111088:	91006000 	add	x0, x0, #0x18             <- Points after uk_eventtab.
    4011108c:	910003fd 	mov	x29, sp
    40111090:	a90153f3 	stp	x19, x20, [sp, #16]
    40111094:	b0000033 	adrp	x19, 40116000 <__uk_inittab20_uk_bus_lib_init>
    40111098:	91002273 	add	x19, x19, #0x8
    4011109c:	91002273 	add	x19, x19, #0x8
    401110a0:	eb00027f 	cmp	x19, x0
    401110a4:	54000322 	b.cs	40111108 <main+0x88>  // b.hs, b.nlast
    401110a8:	d1000414 	sub	x20, x0, #0x1
    401110ac:	d2800002 	mov	x2, #0x0                  // #0  <- This is the 0 output for *itr
    401110b0:	cb130294 	sub	x20, x20, x19
    401110b4:	f90013f5 	str	x21, [sp, #32]
    401110b8:	927df294 	and	x20, x20, #0xfffffffffffffff8
    401110bc:	f0000035 	adrp	x21, 40118000 <__func__.1768+0x810>
    401110c0:	8b140274 	add	x20, x19, x20
    401110c4:	911362b5 	add	x21, x21, #0x4d8
    401110c8:	14000004 	b	401110d8 <main+0x58>
    401110cc:	eb14027f 	cmp	x19, x20
    401110d0:	540001a0 	b.eq	40111104 <main+0x84>  // b.none
    401110d4:	f8408e62 	ldr	x2, [x19, #8]!          <- First printf begins here for following iterations
		printf("0x%x, 0x%x\n", itr, *itr);
    401110d8:	aa1303e1 	mov	x1, x19                 <- First printf begins here for first iteration
    401110dc:	aa1503e0 	mov	x0, x21
    401110e0:	97ffd5a8 	bl	40106780 <printf>
		printf("0x%x, 0x%x\n", itr, *itr);
    401110e4:	f9400262 	ldr	x2, [x19]
    401110e8:	aa1303e1 	mov	x1, x19
    401110ec:	aa1503e0 	mov	x0, x21
    401110f0:	97ffd5a4 	bl	40106780 <printf>
		//__event_assert(*itr);
		ret = ((*itr)(data));
    401110f4:	f9400261 	ldr	x1, [x19]
    401110f8:	d2800000 	mov	x0, #0x0                   	// #0
    401110fc:	d63f0020 	blr	x1
		if (ret)
    40111100:	34fffe60 	cbz	w0, 401110cc <main+0x4c>
```

The table looks like this:
```
0x40116008 <uk_event_myevent>:                     0x0000000040116018 ---------------+
0x40116010 <_uk_eventtab_myevent_9_myhandler>:     0x0000000040103ee8 <myhandler>    |
0x40116018:                                        0x0000000000000000 <--------------+
```

It is generally fine that the code first computes `0x40116018` at `40111088`. This is the offset that points after the table that gcc uses to determine if the foreach has finished. However, the whole code between `40111084` and `401110a4` is in fact useless, as it is constant and always evaluates to false. A hint on what is wrong is given at `401110ac`:
```ASM
    401110ac:	d2800002 	mov	x2, #0x0                  // #0  <- This is the 0 output for *itr
```
This is the third argument to the first `printf` call. Obviously gcc assumes `*itr` to be always `NULL`, **BUT ONLY** in the first use as argument in `printf`. For the second call, gcc emits code that actually does the memory access and thus the different output. If you enable the following assert, gcc assumes `*itr` to be always 0 again.  I am not quite sure, why gcc sometimes emits a constant `x2 = 0` and sometimes an actual memory access, but the problem originates from the constant definition of `uk_event`:
```C
#define _EVT_SECTION_HEADER(event)					\
	extern const uk_event_handler_t _EVT_HLIST_END(event)[];	\
	---> const <--- struct uk_event					\
	__used __section(".uk_eventtab_" #event "~")			\
	_EVT_EVENT(event) = {						\
		_EVT_HLIST_END(event)					\
	}

#define _EVT_IMPORT_EVENT(event)					\
	extern ---> const <--- struct uk_event _EVT_EVENT(event)
```

GCC assumes that the struct does not change and while the code is compiled the list end label (defined in the ASM snippet) is unresolved and thus assumed 0 (as GCC does for all unresolved symbols). The optimizer then kicks in and thinks it can do some magic shortcuts, but totally screws it. Unfortunately for gcc and its optimizer, the linker assigns a non-zero address to the list end label. Hence the discrepancy between the code and the actual data.

In fact, if you just put the assert in (without the printf) the optimizer thinks it is very clever and even throws away the handler call altogether, assuming it can never be called.

Long story short: Yes, this is a bug in gcc. A fix is to remove the `const` keyword so gcc does not make any assumptions about the value of the label.

marcrittinghaus:
First time, I have written code that not even the compiler understands 😆 Perhaps we should provide your sample zip to ARM or the gcc guys?!

michpappas:
Hi @marcrittinghaus

> GCC assumes that the struct does not change and while the code is compiled the list end label (defined in the ASM snippet) is unresolved and thus assumed 0 (as GCC does for all unresolved symbols). The optimizer then kicks in and thinks it can do some magic shortcuts, but totally screws it. Unfortunately for gcc and its optimizer, the linker assigns a non-zero address to the list end label. Hence the discrepancy between the code and the actual data.
>
> In fact, if you just put the assert in (without the printf) the optimizer thinks it is very clever and even throws away the handler call altogether, assuming it can never be called.
>
> Long story short: Yes, this is a bug in gcc. A fix is to remove the const keyword so gcc does not make any assumptions about the value of the label.

Nice, thanks for the interpretation :+1: 

> It is generally fine that the code first computes 0x40116018 at 40111088. This is the offset that points after the table that gcc uses to determine if the foreach has finished.

It is indeed. Back then I didn't fully understand how the list is created and terminated. Now I do. 

In retrospect, I find this implementation to provide an elegant way to allow libraries to register an arbitrary number of events, without having to resort to pre-allocated arrays of a finite number of elements. Yet, in all its ingenuity it contains subtle points that cannot be grasped by just reading through the code. I guess you win some you lose some :)

> First time, I have written code that not even the compiler understands laughing Perhaps we should provide your sample zip to ARM or the gcc guys?!

It's a good idea :)

michpappas:
Hi @marcrittinghaus, hope you're well!  I finally got to to have another look at this PR. 

First some general questions:
* I see in many macros follow a pattern of one or more levels of indireciton. For example: `UK_EVENT_HANDLER` -> `_UK_EVENT_HANDLER_PRIO` -> `__UK_EVENT_HANDLER_PRIO`. What is the purpose of that?
* What is the purpose of the double underscore in `_uk_eventtab_<event>__end`?
* The presentation states "Tilde (~) enforces correct ordering - no matter the event name", yet I cannot to see how it makes a difference if it's the final character. Could you please elaborate?
* When events are raised, iteration terminates on the first event that returns success. Isn't that in conflict with the ability to define more than one handlers? Also what is the intended purpose of priority levels?
* Is there any specific reason of a dedicated section for events?

Questions aside, I still find the code quite complex for the task and I am concerned about maintainability. Also macronization makes debugging harder; you can't set breakpoints etc. I put together a simplified version, you can find it at https://github.com/michpappas/unikraft/tree/uk_event_simple. It is not as sophisticated, and I'm not sure if it fits your original requirements, or whether you have interest to replace / rework the current version whatsoever. If you would prefer to get it merged the current version instead, all is missing is the workaround for the GCC error :)

**Update:** I think I got the reasoning behind using sections and why the simple implementation won't work. You want to be able to register handlers from different libraries for a given event. Is that right?

marcrittinghaus:
Hi @michpappas,

Thank you for taking the time to look into this PR. Hope you are well, too! 😃

> * I see in many macros follow a pattern of one or more levels of indireciton. For example: `UK_EVENT_HANDLER` -> `_UK_EVENT_HANDLER_PRIO` -> `__UK_EVENT_HANDLER_PRIO`. What is the purpose of that?

The purpose of the indirections is to allow macro expansion for cases like this:
```C
#define CONCAT(first, last) first ## last

UK_EVENT(CONCAT(my, event));

static int myhandler(void *data) { return UK_EVENT_HANDLED; }
UK_EVENT_HANDLER_PRIO(CONCAT(my, event), CONCAT(my, handler), UK_PRIO_LATEST);
```

But you are right. I checked the definitions and removed some unnecessary steps (e.g.,  `__UK_EVENT_HANDLER_PRIO`) 👍🏻 

> * What is the purpose of the double underscore in `_uk_eventtab_<event>__end`?

When I started working on this, I first did not use the priorities that UK already defines but would also allow other literals as priorities. The underscore was meant to ensure that the ordering would still be right. I removed it, as it is no longer needed 👍🏻 

> * The presentation states "Tilde (~) enforces correct ordering - no matter the event name", yet I cannot to see how it makes a difference if it's the final character. Could you please elaborate?

The tilde is necessary so that handlers for the same event are grouped and not mixed up with handlers from other events when sorting alphabetically in the linking step.

Consider you would have the following sections:
```
.uk_event_myevent2 << from UK_EVENT(myevent2)
.uk_event_myevent << from UK_EVENT(myevent)
.uk_event_myevent_end << from UK_EVENT(myevent)
.uk_event_myevent_9_myhandler << from UK_EVENT_HANDLER(myevent, myhandler)
```
If you sort them you get:
```
.uk_event_myevent
.uk_event_myevent2
.uk_event_myevent_9_myhandler
.uk_event_myevent_end
```
This would obviously break things in the final binary as the handlers are supposed to come right after the event definition and must also not be mixed up with other events. If you work with the tilde you have:
```
.uk_event_myevent2~ << from UK_EVENT(myevent2)
.uk_event_myevent~ << from UK_EVENT(myevent)
.uk_event_myevent~~ << from UK_EVENT(myevent)
.uk_event_myevent~9_myhandler << from UK_EVENT_HANDLER(myevent, myhandler)
```
Sorting this gives you:
```
.uk_event_myevent2~
.uk_event_myevent~
.uk_event_myevent~9_myhandler
.uk_event_myevent~~
```
So no matter what C identifier you choose for myevent2 it will always sort correctly because the tilde is not a valid character for a C identifier. It is a bit complex, but a necessary evil to allow arbitrary event names.

> * When events are raised, iteration terminates on the first event that returns success. Isn't that in conflict with the ability to define more than one handlers? Also what is the intended purpose of priority levels?

Iteration does only end if the handler returns `UK_EVENT_HANDLED`, otherwise iteration continues. But I see the problem and I added a 3rd return value that allows for continuing the iteration while treating this as a success for the caller of `uk_raise_event` in the end (`UK_EVENT_HANDLED_CONT`) no matter the return value of the other handlers :+1:

Priority levels are meant to allow giving certain handlers precedence over others. For example, I register a page fault handler in `uknofault` which allows probing memory addresses without causing a fatal crash. Now I may want this page fault handler to have low priority as I expect it to be called in debugging/crash handling code only. Instead, I want the page fault handler for memory mapped files to kick in earlier as I assume this to be the more common page fault reason. Unfortunately, I cannot know in advance what the reason is in the trap so I have to give all handlers a chance to see if the instruction pointer or probed memory address falls into their responsibility. So priorities allow to optimize this a bit.

> * Is there any specific reason of a dedicated section for events?

No. I just did this to be inline with the other code that registers global handlers (e.g., the constructors in `.uk_ctortab`). But it would be simple to move this into the `.data` (or `.rodata` section) like so:

`common.lds.h`
```C
#define EVENT_DEFINITIONS						\
	. = ALIGN(0x8);							\
	KEEP(*(SORT_BY_NAME(.uk_eventtab_*)))
```
Linker script:
```
	/* Read-write data (initialized) */
	. = ALIGN(__PAGE_SIZE);
	_data = .;
	.data :
	{
		*(.data)
		*(.data.*)
		EVENT_DEFINITIONS
	}
	_edata = .;
```

I could change the PR accordingly as I also dislike section bloat. Then we need to also manually define the `.data` section in the `linuxu` target.

> Questions aside, I still find the code quite complex for the task and I am concerned about maintainability. Also macronization makes debugging harder; you can't set breakpoints etc. I put together a simplified version, you can find it at https://github.com/michpappas/unikraft/tree/uk_event_simple. It is not as sophisticated, and I'm not sure if it fits your original requirements, or whether you have interest to replace / rework the current version whatsoever. If you would prefer to get it merged the current version instead, all is missing is the workaround for the GCC error :)

Hey, thanks for putting together a simpler version. Unfortunately, it is not satisfying the original requirement. The idea is to define an event in one library (or the platform) and being able to install handlers without modification to the event definition. Otherwise, you cannot transparently extend event handling functionality. For example, to support the `uknofault` library I would have to change the page fault event in the platform and create an (optional) dependency on the (potentially external) library. This is not inline with UK's requirement to be as modular as possible. I agree that the code is not that clear on the first sight. I can extend the comments, especially for the `~` part if that would help? 

I incorporated the workaround for the GCC error by removing the `const` keywords. It is also a nice thing because we can then easily extend `uk_event` with stats to count events in the future, for example, the number of page faults. But this also means the  structs should be placed in the `.data` section and not the `.rodata` section.

michpappas:
@marcrittinghaus many thanks for the elaborate reply.

> But I see the problem and I added a 3rd return value that allows for continuing the iteration while treating this as a success for the caller of uk_raise_event in the end (UK_EVENT_HANDLED_CONT) no matter the return value of the other handlers

Sounds like a good idea. I would imagine most handlers would use that just in case, and only those that "know what they're doing" will explicitly stop processing. Except from those that don't handle the event, that is.

> and not mixed up with handlers from other events

From *other* events. That's the part I didn't think of 🤦🏼  With that in mind of course it makes sense 😓 

> Hey, thanks for putting together a simpler version. Unfortunately, it is not satisfying the original requirement.

Yeah I figured it out after watching your talk in FOSDEM this morning, and it all made sense. I posted an update to my initial comment probably at the same time you were writing these lines :)

> I agree that the code is not that clear on the first sight. I can extend the comments, especially for the ~ part if that would help?

It's fine, I think it has been extensively documented already through this discussion :)

> I incorporated the workaround for the GCC error by removing the const keywords. It is also a nice thing because we can then easily extend uk_event with stats to count events in the future, for example, the number of page faults. But this also means the structs should be placed in the .data section and not the .rodata section.

That would be really great. But I guess we can do it later, right. It will be a backwards compatible change anyway.

michpappas:
One last thing I noticed is that handlers within a given priority won't be sorted.
Consider the following:
```
int myhandler1(void *data)    
{    
        (void)data;    
        printf("myhandler1\n");    
        return UK_EVENT_NOT_HANDLED;    
}    
    
int myhandler2(void *data)    
{    
        (void)data;    
        printf("myhandler2\n");    
        return UK_EVENT_HANDLED_CONT;    
}    
    
int myhandler3(void *data)    
{    
        (void)data;    
        printf("myhandler3\n");    
        return UK_EVENT_HANDLED_CONT;    
}    
    
UK_EVENT(myevent);
UK_EVENT_HANDLER(myevent, myhandler1);
UK_EVENT_HANDLER(myevent, myhandler2);
UK_EVENT_HANDLER(myevent, myhandler3);

{    
        (void)argc;    
        (void)argv;    
            
        uk_raise_event(myevent, NULL);
    
        return 0;
}
```
During execution handlers are called in the reverse order:
```
bash$ qemu-system-aarch64 -machine virt -cpu max -monitor none -serial stdio -nographic -kernel build/app-helloworld_kvm arm64
Powered by
                    Mimas 0.7.0~4daef00
myhandler3
myhandler2
myhandler1
```

While looking at the object files I noticed that the problem starts at the individual object files:
```
bash$ readelf -sW build/apphelloworld/main.o | grep handler
    16: 0000000000000000     8 OBJECT  LOCAL  DEFAULT   32 _uk_event_myevent_9_myhandler3
    17: 0000000000000008     8 OBJECT  LOCAL  DEFAULT   32 _uk_event_myevent_9_myhandler2
    18: 0000000000000010     8 OBJECT  LOCAL  DEFAULT   32 _uk_event_myevent_9_myhandler1
    92: 0000000000000000    32 FUNC    GLOBAL DEFAULT   22 myhandler1
    94: 0000000000000020    32 FUNC    GLOBAL DEFAULT   22 myhandler2
    95: 0000000000000040    32 FUNC    GLOBAL DEFAULT   22 myhandler3
```
I think this is due to `.pushsection` which, as the name indicates, uses some sort of stack that presumably ld pops while creating the relevant sections. The documentation also hints towards that direction:
```
This directive pushes the current section (and subsection) onto the top of the section stack, and then replaces the current section and subsection with name and subsection
```
I can't think of a way to get this fixed. While it can cause some unpleasant surprises, I think in practice it's unlikely that we'll stumble into it (famous last words). Yet it's good to keep it in mind that one must not rely into it, and always use priorities. Anyway, if you agree that we can live with it, I can approve the PR.

marcrittinghaus:
Hi @michpappas,

yes, right. Ordering within the same priority class is non-deterministic. It becomes even more surprising, if you change between `-O0` and `-O3`. Somehow, `ld` decides to reverse handler ordering then. We could leave it that way and generally say that there is no guarantee for handlers of the same class to be called in any particular order. However, we can also easily remove some of the surprise and enforce ordering of handlers from the same object file. The trick is to use the line number as subsection index. I also do it in `uktest`. I will look into it later this day. Should be a quick fix. Then it is probably more intuitive and also less error-prone as the behavior (within the same C file) does not change with the optimization level.

What about the dedicated `.uk_eventtab`? Do you think, we should move it do `.data`?

michpappas:
@marcrittinghaus, 

> We could leave it that way and generally say that there is no guarantee for handlers of the same class to be called in any particular order. However, we can also easily remove some of the surprise and enforce ordering of handlers from the same object file. The trick is to use the line number as subsection index. I also do it in uktest. I will look into it later this day. Should be a quick fix. Then it is probably more intuitive and also less error-prone as the behavior (within the same C file) does not change with the optimization level.

If it is documented it should be okay. If it can be fixed, even better :) I didn't get the chance to look at the line numbers trick in uk_test yet, but I incidentally discovered an issue in self tests that might be related - haven't had the chance to look at it deeper though, so I can't tell with certainty. For more info see #393.

> What about the dedicated .uk_eventtab? Do you think, we should move it do .data?

I don't have a strong opinion about it 🤐

michpappas:
Hi @marcrittinghaus, looking briefly at how uk_test works with line numbers, I think that using numbered subsections will make execution deterministic with respect to the order of declaration, but it won't help with the sorting problem. That is declaring handler1, handler 3, handler 2 in that order will result into execution in that order, but won't result in execution by name order.

Perhaps the only way to avoid confusion is to clearly state that one must use priorities for ordering, and that there are no guarantees in the execution order of handlers of the same priority level. So if you agree, maybe add a comment or even better some documentation on how to use the events interface in the header file, and we can go ahead and get this merged? :)

marcrittinghaus:
Hi @michpappas,

yes, it would only enforce declaration order. I have added a more thorough explanation with usage example at the top and also added a short sentence on ordering to the `UK_EVENT_HANDLER` macro. I figured that adding line ordering would probably require to also use inline assembly for the event definition so we can use subsections. So it is probably not worth the added complexity. I also keep the separate section for now. We can think of consolidating some of the extra sections with `.data` and `.rodata` in the future.

michpappas:
@marcrittinghaus thanks! Except from the small typo caught by checkpatch, and a review by @skuenzer, from my side it's ready to merge :)

include/uk/event.h
```diff
@@ -0,0 +1,296 @@
...
+#ifdef CONFIG_LIBUKDEBUG
+#include <uk/assert.h>
+#endif /* CONFIG_LIBUKDEBUG */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef CONFIG_LIBUKDEBUG
+#define __event_assert(x) UK_ASSERT(x)
```
skuenzer: To be sure there will be no conflict, maybe call this `__uk_event_assert`

include/uk/event.h
```diff
@@ -0,0 +1,296 @@
...
+ * UK_EVENT_HANDLER_PRIO(handler1, UK_PRIO_EARLIEST);
+ * UK_EVENT_HANDLER_PRIO(handler2, UK_PRIO_LATEST);
+ * UK_EVENT_HANDLER_PRIO(handler3, UK_PRIO_LATEST);
+ * ```
+ */
+
+#include <uk/essentials.h>
+#include <uk/prio.h>
+
+#ifdef CONFIG_LIBUKDEBUG
```
skuenzer, line 116: You should probably include `<uk/config.h>` to make sure you have config options populated in this file.

include/uk/event.h
```diff
@@ -0,0 +1,296 @@
...
+/* Internal macros to create per-event sections. The linker will sort these
+ * sections and merge them in the .uk_eventtab section in the final binary. So
+ * that we are able to arbitrarily name events (including numbers, underscores,
+ * etc.) and still guarantee proper ordering in the presents of multiple events,
+ * we use the tilde (~) to mark header and end sections as well as to separate
+ * the event name from handler priority. The tilde has a large ASCII value (126)
+ * and thus allows us to keep sections in order, irrespective of the chosen
+ * event name (and conceptionally also the priority's textual representation).
+ */
+#define _EVT_EVENT(event)	uk_event_ ## event
```
skuenzer: For avoiding any potential conflict, I would name the `_EVT` prefix with `_UKEVT` or `_UK_EVT`

skuenzer (commented):
Hey Marc,
this PR looks really good. I would just rename some definitions to make sure that we never hit the case where we have a conflict because of a name. Obviously it is a matter on how we consider `EVENT` or `EVT` being potentially a conflicting name. Other from this, I am fine with this PR.
Thanks,
Simon

skuenzer (commented):
Can you also include the `EVENTTAB_SECTION` to Xen on Arm's linker script?

include/uk/event.h
marcrittinghaus: Done :+1:

include/uk/event.h
marcrittinghaus: I stole this from `/include/uk/refcount.h`, where it does not use the `uk` prefix. I have changed it anyways 😃

But we should have a discussion about this because I heavily dislike `UK_ASSERT` not being part of the generic headers and having to check for `ukdebug` all the time. This is especially stupid as we add dependencies for `ukdebug` everywhere, also in the platform code. If we were stringent, we would have to add checks everywhere also for the presence of the `uk_pr_*` functions or simply accept that you cannot build without `ukdebug`, which would make such checks obsolete. I already had a short discussion about this with @michpappas here: https://github.com/unikraft/unikraft/pull/338#discussion_r829961496

include/uk/event.h
marcrittinghaus: Good point. Thanks! 👍🏻

marcrittinghaus:
> Can you also include the `EVENTTAB_SECTION` to Xen on Arm's linker script?

I have not added it because I have no possibility to test it and to be honest, I regard Xen ARM32 as dead and would remove it 😄 

Anyways. Done :+1: 😃

include/uk/event.h
skuenzer: Good points. I would now also say that we should change it within refcount, too. In general, I agree with you, `UK_ASSERT` became such an important macro for us. Together with the compile-time assertions, it might make sense to move it out from `ukdebug` and place it under the global `/include/uk` folder. However, the problem that we have is that the assertions currently depend on `ukdebug's` printing system :thinking:... this is why we stick them with `ukdebug` at the moment. In general, we have only problems with headers that are in `/include` which we explicitly do not want to imply any library dependency so that we keep highest flexibility. Maybe we can do an exception for assertions to let print messages only if `UKDEBUG` is available... similar to this assertion availability helper here. What do you think? It would be anyways a different PR. :smile:

skuenzer:
> > Can you also include the `EVENTTAB_SECTION` to Xen on Arm's linker script?
> 
> I have not added it because I have no possibility to test it and to be honest, I regard Xen ARM32 as dead and would remove it 😄
> 
> Anyways. Done 👍 😃

Great, thanks a lot. I understand. However, we still have Xen for Arm32 as supported upstream platform. We should have in general a maintainers discussion if we really want to remove the code. It is possible that we want to keep it as basis for Xen on Arm64 support. An option could be to move it out from the main tree... :thinking:

skuenzer (approved):
:+1: Xen on Arm32 could be still compiled and run. At the moment still without scheduler, and DCO enabled. Actually, Xen's `depends on` line in `Config.uk` should be updated for Arm32 so that a conflicting configuration cannot be created...

michpappas:

PR #228, opened 2021-06-14
plat/common/x86: Fix control protection trap

marcrittinghaus:
The current trap number for the control protection exception (in Unikraft called security trap) is wrong and the asm trap is missing. This patch fixes these issues.

From the Intel manual:
image
image


skuenzer (approved):
Thanks a lot for this fix.

PR #230, opened 2021-06-15
plat/x86: Use event-based trap interface

### LINKED ISSUE #210: Support backtrace dumping for debugging purpose

marcrittinghaus:
This PR changes the trap handling in KVM and XEN to use the event-based trap interface. This means, it adds code to raise the corresponding trap event in the architectural trap. A list of implemented trap events and a context data structure that is supplied to handlers is provided in `uk/arch/traps.h`.

For example, adding a page fault handler can be done in the following way:
```
#include <uk/arch/traps.h>
static int my_pagefault_handler(void *data)
{

	return UK_EVENT_HANDLED;
}
UK_EVENT_HANDLER_PRIO(UKARCH_TRAP_PAGE_FAULT, my_pagefault_handler, UK_PRIO_LATEST);
````
The same code will work with other architectures (e.g., arm64). However, the `ukarch_trap_ctx` structure may change.

**NOTE**: This PR depends on #227.

### DIFF
--- a/arch/x86/x86_64/include/uk/asm/traps.h
+++ b/arch/x86/x86_64/include/uk/asm/traps.h
@@ -1,0 +1,65 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Author(s): Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *
+ * Copyright (c) 2021, Karlsruhe Institute of Technology. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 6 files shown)


razvand:
@marcrittinghaus , could you please rebase this?

marcrittinghaus:
Changed `TRAP_*` to `UKARCH_TRAP_*` to avoid potential name clashes.

skuenzer (changes requested):
Could you adopt the title since this PR just adopts KVM and xen, not linuxu.
Something like `plat/kvm, /plat/xen: x86: Use event-based trap interface`

marcrittinghaus:
> Could you adopt the title since this PR just adopts KVM and xen, not linuxu. Something like `plat/kvm, /plat/xen: x86: Use event-based trap interface`

This would actually be too long.

skuenzer (approved):
Thanks a lot for the updates. I am good with this PR @razvand .

razvand (approved):

PR #231, opened 2021-06-15
plat/arm64: Use event-based trap interface

### LINKED ISSUE #210: Support backtrace dumping for debugging purpose

marcrittinghaus:
This PR changes the trap handling in KVM for arm64 to use the event-based trap interface. This means, it adds code to raise the corresponding trap event in the architectural trap. A list of implemented trap events and a context data structure that is supplied to handlers is provided in `uk/arch/traps.h`.

For example, adding a page fault handler can be done in the following way:
```
#include <uk/arch/traps.h>
static int my_pagefault_handler(void *data)
{

	return UK_EVENT_HANDLED;
}
UK_EVENT_HANDLER_PRIO(UKARCH_TRAP_PAGE_FAULT, my_pagefault_handler, UK_PRIO_LATEST);
````
The same code will work with other architectures (e.g., x86). However, the `ukarch_trap_ctx` structure changes.

**NOTE**: This code depends on #227

### CODE REVIEW

**Comment by razvand** [comment]
Hi, @marcrittinghaus . Please rebase this PR.

**Comment by michpappas** [comment]
Hi @marcrittinghaus, I'm quite excited about this change as it paves the way for gdb support :) 

Please find some general comments below, along with a few more inline:

As this PR still depends on #227 I assume that the two commits that introduce events in this PR will be eventually removed in a later revision.

From what I see there are some additional commits for cleanups here too. I think it's okay to have cleanup work included in a PR with a different scope, but maybe at least the commits that implement changes for events should be grouped together.

**Comment by michpappas** [comment]
In the commit message of e4d7664:
> This commit does multiple things:

Generally it's preferable to do one thing per commit, also see below.

> 1) Removes the EL0 stack pointer from the arm64 __regs structure.
> There is no meaningful use for the EL0 stack pointer as all code
> runs in EL1. Even if we were entering from EL0, the new trap code
> saves EL0_SP in the shared sp field. As there is no alignment of EL1_SP
> when coming from EL0, EL1_SP can be restored simply by subtracting
> __REGS_SIZEOF (i.e., there is no need to explicitly store EL_SP1 in
> __regs). That also makes it more intuitive to use the sp field in __regs
> (e.g., in crash code, debuggers, etc.), as it always represents the
> stack pointer of the trapped code, no matter if this was running in
> EL0 or EL1. In addition, in the old code SP1 would erroneously
> point below the trap frame (e.g., this breaks stack traces).

Actually in a given exception level (EL1 for the sake of this discussion), when taking an exception the default is to use SP_ELx. For Exception Levels other than EL0, AArch64 provides the ability to choose between SP_EL0 and SP_ELx:

https://developer.arm.com/documentation/ddi0500/d/programmers-model/armv8-a-architecture-concepts/stack-pointer-selection

There is no hard rule on which SP to use. Some platforms use SP_EL0 for runtime. Others use SP_EL0 for early exception handling and use SP_ELx for runtime. Others use SP_ELx in all cases etc. The conculusion is that SP_EL0 is not exclusively used by EL-0 :)

> Removes the ESR_EL1 field from the arm64 __regs structure and
> supplies it via an extra parameter to the trap functions.
> There is no use case for restoring ESR_EL1 when leaving the trap, it
> can thus be removed, which also alignes the __regs structure to
> 16-bytes again, after removing SP0.

Ok, but s/alignes/aligns/

> 3) Fixes a bug, where x19 is not being restored after a trap.
> 4) Fixes a bug, where EL1_SP is incorrectly restored, leaking

Ideally these would require separate PRs, but at a minimum let's have bug fixes in separate commits, so that it's easier to spot them out in git history. Also makes life easier when backporting :)

File: arch/arm/arm64/include/uk/asm/lcpu.h [hunk 1]
```diff
@@ -47,25 +46,25 @@
  */
 struct __regs {
 	/* General-purpose registers, from x0 ~ x29 */
-	uint64_t x[30];
+	unsigned long x[30];
```
michpappas: As this is not portable code, but architecture specific, it would be better to use fixed sizes. I would instead replace uint64_t with _u64 etc rather than using generic types.

michpappas:
The commit message of 70b4468 states that
> This commit changes the arm64 platform code for kvm

actually the changes do not modify any files under plat/kvm/

plat/common/arm/traps.c
```diff
@@ -32,6 +36,121 @@ static const char *exception_modes[]= {
 	"Error"
 };
 
+enum aarch64_trap {
```
michpappas: `trap.c` is a generic file. If we add Armv7 support it will bloat. It would be better if we declared these to separate per-arhictecture files.

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -51,32 +49,25 @@ struct __regs {
 	/* Link Register (x30) */
 	unsigned long lr;
 
-	/* Exception Link Register */
-	unsigned long elr_el1;
+	/* Program Counter */
```
michpappas: I was initially confused by this as in AArch64 there is no way to get access to the PC. Of course ELR does contain a previous value of the PC, yet I think it would be more precise if this was called elr_el1. If you think I'm being too pedantic, feel free to ignore this comment though ;)

michpappas (changes requested):
@marcrittinghaus, I'll stop reviewing for now cause I'm not sure how much will change based on your views on the SP_EL0 topic. Will do another round on the next revision.

plat/common/arm/traps.c
marcrittinghaus: Yes, good point. I renamed the file to `traps_arm64.c` for now. I think, we may want to move all the 64-bit platform files into a `arm64` subdirectory and remove the `_arm64` suffix at some point. Instead we should have a separate `arm32` or so folder for the 32-bit stuff. All common code could remain where it is.

Note that the `traps.c` is not a new file, although I changed a lot.

marcrittinghaus:
@michpappas

Thank you very much for your feedback 😃 Very much appreciated and very valuable as always! You are totally right that the cleanup part of this PR needs some more discussion and I would like to move it to a separate PR to get the trap interface into this release. I thus removed everything that is not strictly related to the trap interface and will do a separate cleanup PR.

marcrittinghaus:
> The commit message of [70b4468](https://github.com/unikraft/unikraft/commit/70b44687b2ce47a4ea023f4ff691e4521ac61d0d) states that
> 
> > This commit changes the arm64 platform code for kvm
> 
> actually the changes do not modify any files under plat/kvm/

Now, it does 😄

plat/common/arm/traps_arm64.c
```diff
@@ -0,0 +1,210 @@
...
+
+	AARCH64_TRAP_MAX
+};
+
+/* Mapping of the fault status code (FSC) for instruction and data aborts to
+ * trap type (either page fault or bus error). Zero means invalid. The map
+ * takes 64 bytes but saves a ton of comparisons.
+ */
+#define NUM_ABORT_MAP_ENTRIES 64
+static const __u8 _abort_map[NUM_ABORT_MAP_ENTRIES] = {
```
michpappas: It is not impossible for the architecture to add new FCSs in a future revision. I'm afraid that someone may add a fault into the array and forget to update `NUM_ABORT_MAP_ENTRIES`. Would it be better to declare `_abort_map[NUM_ABORT_MAP_ENTRIES]` as `_abort_map[]` instead and use `ARRAY_SIZE` when iterating?

plat/common/arm/traps_arm64.c
```diff
@@ -0,0 +1,210 @@
...
+	case ESR_EL1_EC_STEP_EL1:
+	case ESR_EL1_EC_WATCHP_EL1:
+	case ESR_EL1_EC_BRK64:
+		return AARCH64_TRAP_DEBUG;
+	}
+
+	return AARCH64_TRAP_MAX;
+}
+
+static void dump_registers(struct __regs *regs, uint64_t far)
```
michpappas: `uint64_t` should be `__u64`. There are more stdint types below.

plat/common/arm/traps_arm64.c
```diff
@@ -0,0 +1,210 @@
...
+
+	uk_pr_crit("Unikraft: EL1 sync trap caught\n");
+
+	dump_registers(regs, far);
+	ukplat_crash();
+}
+
+void trap_el1_irq(void)
+{
+	if (!gic) {
```
michpappas: I think it would be worth to use `unlikely` to optimize interrupt handling for the common case :)

plat/common/arm/traps_arm64.c
```diff
@@ -0,0 +1,210 @@
+/* SPDX-License-Identifier: ISC */
+/*
+ * Authors: Wei Chen <Wei.Chen@arm.com>
+ *
+ * Copyright (c) 2018 Arm Ltd.
```
michpappas, line 6: The copyright needs to be updated.

plat/common/include/arm/arm64/cpu_defs.h
```diff
@@ -292,4 +294,103 @@
 		(SECT_ATTR_DEFAULT | ATTR_XN | \
 		ATTR_IDX(DEVICE_nGnRnE))
 
+/*
```
michpappas, line 297: It's okay to add these here for now. In the virtual memory PR for Arm I'm doing some cleanups and will move these along with others under arch/arm/arm64/.

plat/kvm/Makefile.uk
```diff
@@ -88,7 +88,7 @@ LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/smccc.c|comm
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/smccc_invoke.S|common
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/time.c|common
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/generic_timer.c|common
-LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/traps.c|isr
```
michpappas, line 94: Hmmm I was concerned whether this will break ARM_32, but a quick grep doesn't show anything other than Xen's own implementation of `trap.c`.

michpappas (changes requested):
Hi @marcrittinghaus, thanks a lot for separating the fixes from this PR. Saying that, it would be nice to get them merged at some point; they were both appreciated and needed 😄 I don't have any major comments other than a couple of small issues below. I also ran a quick test with UKARCH_TRAP_PAGE_FAULT, everything looks good 👍🏼

plat/common/arm/traps_arm64.c
marcrittinghaus: Yes, good idea. I wasn't aware of `ARRAY_SIZE` actually 😄

plat/common/arm/traps_arm64.c
marcrittinghaus: Yes. I have not touched these functions in the PR because I replaced them in the cleanup anyways. But I have fixed these issues 👍🏻 😃

plat/common/arm/traps_arm64.c
marcrittinghaus: I replaced this with a simple `UK_ASSERT`. I think it can be considered a programming error if the GIC has not been initialized and the boot process was allowed to continue.

plat/common/arm/traps_arm64.c
marcrittinghaus: Done 👍🏻 Actually, when the trap functions are replaced as part of the cleanup and debugging stuff, I will replace the header as there will be no ARM code left anymore.

plat/common/include/arm/arm64/cpu_defs.h
marcrittinghaus: Great! Thank you very much 😃

plat/kvm/Makefile.uk
marcrittinghaus: Yes, the `traps.c` was exclusively used by KVM despite its position in `common` 🤷🏻‍♂️

marcrittinghaus:
@michpappas,

thanks a lot! I have made the requested changes.

michpappas:
@marcrittinghaus, thanks a lot! 🚀

michpappas (approved):
Reviewed-by: Michalis Pappas <mpappas@fastmail.fm>

skuenzer (approved):
Thanks @marcrittinghaus!

razvand (approved):

PR #232, opened 2021-06-17
lib/ukdebug: Fix overrun in outf for OUTDEV_BUFFER

marcrittinghaus:
This patch fixes two bugs:
1) The return value of vsnprintf does not include the null-terminator.
Subtracting one is thus not necessary.
2) In case the buffer is too small, vsnprintf returns the number of
characters that would have been written if the buffer had
enough space. The return value must therefore be tested against
the remaining buffer space.


vladandrew (approved):
Hi Marc,

This patch looks good. Now vsnprintf should work accordingly.

PR #234, opened 2021-06-21
Provide GIC v3 support

rene:
This patch series implements GIC v3 device driver support:

* Enable affinity IRQ routing
* Implement basic functions from GICv2 to GICv3
* Introduces a common GIC driver infrastructure in order to allow both GICv2 and GICv3 drivers on the same binary. The corresponding version is chosen according to device tree configuration.

Note: LPI's are not supported.

GICv3 can be tested with qemu using _-machine gic-version=3_ argument, for instance:
qemu-system-aarch64 -M virt-2.7 -cpu cortex-a57 -machine gic-version=3 -nographic -kernel <unikraft-kernel.dbg>

### DIFF
--- a/plat/Config.uk
+++ b/plat/Config.uk
@@ -37,3 +37,19 @@ config FPSIMD
 	depends on ARCH_ARM_64
 	help
 		Enable support FPU usage in application
+
+menuconfig LIBGIC
+	bool "Arm GIC (generic interrupt controller) support"
+	select LIBOFW
+	default y if ARCH_ARM_64
+	depends on ARCH_ARM_64
+	depends on PLAT_KVM
+if LIBGIC
+config LIBGICV2
+	bool "Version 2"
+	default y
+
+config LIBGICV3
+	bool "Version 3"
+	default y
+endif
--- a/plat/drivers/gic/gic-v2.c
+++ b/plat/drivers/gic/gic-v2.c
@@ -39,37 +39,22 @@
 #include <uk/asm.h>
 #include <uk/plat/lcpu.h>
 #include <uk/plat/common/irq.h>
-#if defined(CONFIG_PLAT_KVM)
+#ifdef CONFIG_PLAT_KVM
 #include <kvm/irq.h>
 #endif
 #include <uk/plat/spinlock.h>
 #include <arm/cpu.h>
-#include <gic/gic.h>
 #include <gic/gic-v2.h>
 #include <ofw/fdt.h>
 
 /* Max CPU interface for GICv2 */
 #define GIC_MAX_CPUIF		8
 
-#define GIC_DIST_REG(r)	((void *)(gic_dist_addr + (r)))
-#define GIC_CPU_REG(r)	((void *)(gic_cpuif_addr + (r)))
-#define IRQ_TYPE_MASK	0x0000000f
-
-static uint64_t gic_dist_addr, gic_cpuif_addr;
-static uint64_t gic_dist_size, gic_cpuif_size;
-#ifdef CONFIG_HAVE_SMP
-static char gic_is_initialized;
-__spinlock gic_dist_lock;
-inline void dist_lock(void) { ukarch_spin_lock(&gic_dist_lock); };
-inline void dist_unlock(void) { ukarch_spin_unlock(&gic_dist_lock); };
-#else
-inline void dist_lock(void) {};
-inline void dist_unlock(void) {};
-#endif
+#define GIC_CPU_REG(gdev, r)	((void *)(gdev.cpuif_mem_addr + (r)))
 
 #ifdef CONFIG_HAVE_SMP
 __spinlock gicv2_dist_lock;
-#endif
+#endif /* CONFIG_HAVE_SMP */
 
 /** GICv2 driver */
 struct _gic_dev gicv2_drv = {
@@ -83,7 +68,7 @@ struct _gic_dev gicv2_drv = {
 	.cpuif_mem_size = 0,
 #ifdef CONFIG_HAVE_SMP
 	.dist_lock      = &gicv2_dist_lock,
-#endif
+#endif /* CONFIG_HAVE_SMP */
 };
 
 static const char * const gic_device_list[] = {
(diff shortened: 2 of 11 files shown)


plat/drivers/gic/gic-v2.c
```diff
@@ -418,6 +427,17 @@ int _dtb_init_gic(const void *fdt)
 {
 	int fdt_gic, ret;
 
+#ifdef UKPLAT_LCPU_MULTICORE
+	if (gic_is_initialized) {
```
razvanvirtan: @marcrittinghaus In one of the comments from my SMP draft, you mentioned that this check is not SMP safe and that we should use instead the bspid. However, `ukplat_lcpu_start()` shouldn't be called before the bootstraping cpu has reached `ukplat_entry()`, so secondary cpus should only reach this part of the code after the boostraping one. Should we still change this condition?

marcrittinghaus: no, then it is fine 👍🏻

razvanvirtan (approved):
All good.

razvanvirtan:
@rene The PR with the architecture independent SMP API is created [1], can you change `UKPLAT_LCPU_MULTICORE` in this PR with `CONFIG_HAVE_SMP`, according to it?

[1] https://github.com/unikraft/unikraft/pull/309

rene:
> @rene The PR with the architecture independent SMP API is created [1], can you change `UKPLAT_LCPU_MULTICORE` in this PR with `CONFIG_HAVE_SMP`, according to it?
> 
> [1] #309

It's done @razvanvirtan .

plat/Config.uk
```diff
@@ -28,3 +28,23 @@ config FPSIMD
...
+config LIBGICV2
+	bool "Version 2"
+	default y
+
+config LIBGICV3
+	bool "Version 3"
+	default y
+endif
+
+config HAVE_SMP
```
razvanvirtan: `HAVE_SMP` will by introduced in `Config.uk` by https://github.com/unikraft/unikraft/pull/309, so we should remove it here.

rene: ok. I will do it.

rene: @razvanvirtan , it's done. Also fixed the maximum line size in commit messages.

rene:
Updates in this PR:
- Rebase with staging
- Fix spinlocks according the new API (recently merged)
- Fix checkpatch error

marcrittinghaus (changes requested):
Hi @rene,

thank you very much for submitting the GIC PR! Sorry that it took so long to review, but I admit I had to study the specs first 😄 As usual I like to blow the party and have prepared a patch file. Please apply it and we take it from there. The patch is mostly clean up but also fixes some small bugs (e.g., lock not released on exit, wrong mask on linenum register, missing EOI for special but non-spurious interrupts).

I also have a couple of open issues: 

1) the GICv2 implements the `gic_sgi_gen_*` functions. Are these used only for GICv2? In that case they should be renamed to `gicv2_*`. However, then I don't know where they could be practically used in the first place as we want the GIC implementation to be hidden by the device object. So that makes me think that these functions are used generically for GICv2 and GICv3 (which would be broken btw. due to the wrong lock being used!! ⚠️). But then they should be in the common part.

2) Generally, we could think about increasing code sharing as some functions are identical to both versions except using a different lock. So probably the v2 and v3 locks should be merged and moved to the common part so that identical functions can follow.

3) In the v2 there is an `ukplat_lcpu_save_irqf()`/`ukplat_lcpu_restore_irqf()`-pair in the `gic_sgi_gen_*` functions. What is the purpose of this guard?

Thank you!

[gic.zip](https://github.com/unikraft/unikraft/files/7610624/gic.zip)

rene:
Hi @marcrittinghaus ,

Thank you very much to be so kind and provide the patch with your changes. Actually your patch provides substantial amount of changes, so I believe we could go through a regular code review instead of receive all your suggestions at once, since it doesn't allow us to discuss in an incremental way, don't give a chance to others to highlight their thoughts about it and eventually will not give you a chance to hear my thinking about some specific implementation decision, approaches, etc. But again, I really appreciate your time and work on providing the patch ready for apply.

I saw you provided a lot of changes on GICv2 driver (~ 38% of the patch). The PR introduces the GICv3 support, so my idea was to touch on GICv2 as less as possible since it was a driver already present, used and tested on upstream. it doesn't mean I don't agree with your changes, actually I think all of them must be committed. However, don't you think those should be made in a separated PR?  We could keep here only changes on GICv2 that are strictly related with the introduction of GICv3. Although I'm not opposed to keep these changes here into a separated commit. What do you think?

Regarding your questions:

1. Currently these functions are used only for GICv2 (they were already present on the upstream driver), but SGI register is different on GICv3, so they cannot be shared, a GICv3 version must implemented.

2. Good point and I agree with you. Remember that when GICv3 driver was developed we didn't have the final spinlock implementation and we had the problem about keep the lock in the struct while locks are put into a dedicated segment. I will check the functions again and try to figure out those that eventually can be shared between both drivers.

3. These are on the GICv2 original driver, AFAIK they are not needed, my guess is they were there because there was no spinlocks when the driver was introduced... I think we can safely replace with spinlocks, but again, these are improvements on GICv2 and are not strictly related with this PR.

I'm reviewing your patch and I will update this PR, coming back to you if I have questions or diverge at some point...

Thank you once again for your work and time on this PR...

rene:
Hi @marcrittinghaus 

I'm applying most of your changes for gic-v3. I have only one diversion so far. Description is below:

In your patch you did the following modifications:

```diff
 /**
- * \brief Set IRQ affinity routing
- * \param [in] irq IRQ number
- * \param [in] affinity
+ * Set interrupt affinity
+ *
+ * @param irq interrupt number [GIC_SPI_BASE..GIC_MAX_IRQ-1]
+ * @param target an 8-bit bitmap with 1 bit per CPU 0-7. A `1` bit
+ *    indicates that the SGI should be forwarded to the respective CPU
  */
-static void gic_set_irq_affinity(uint32_t irq, uint8_t affinity)
+static void gicv3_set_irq_affinity(uint32_t irq, uint32_t affinity)
 {
-	uint64_t irouter_val = 0; /* Interrupt_Routing_Mode = 0 */
-
-	if (irq < GIC_SPI_BASE)
-		UK_CRASH("Bad irq number: should not less than %u",
-				GIC_SPI_BASE);
-
-	irouter_val = ((affinity << 8) & MPIDR_AFF3_MASK) |
-		(affinity & 0xffffff);
+	UK_ASSERT(irq >= GIC_SPI_BASE && irq < GIC_MAX_IRQ);
 
 	dist_lock(gicv3_drv);
-	write_gicd64(GICD_IROUTER(irq), irouter_val);
+	write_gicd64(GICD_IROUTER(irq), GIC_AFF_TO_ROUTER(affinity, 0));
 	dist_unlock(gicv3_drv);
 }
```
However the format of IROUTER<n> registers are:
```
63     40     32 30     24    16      8 
| RES0  | Aff3 | | RES0 | Aff2 | Aff1 | Aff0 |
```
Edit:  That will be converted from 32 bits affinity by GIC_AFF_TO_ROUTER macro. 

The  "target" argument description is not true.

marcrittinghaus:
Hi @rene,

yes. Makes absolutely sense to move the changes for GICv2 to a different PR. I sometimes tend to just work with the code, do cleanup etc. all at once, not considering the resulting git history very much 😄 

I decided against a github-based review process at that time as I think it is very ineffective and leads to many details being untouched due to the amount of work it takes to comment (not meaning code quality is bad!). I therefore like to provide a patch file so that trivial things like whitespaces or matching code style with the project etc. are not eating time. I am glad that in the end you agree with most changes from the patch. Please, just leave out changes that you think need a more thorough debate or clarification or are just a misunderstanding/mistake/unnecessary. We can then discuss further on this basis.

Concerning the affinity issue. Yes, the description is not right. I think, I first added the description and then figured out that the `uint8_t` is actually wrong and should be a `uint32_t` which is converted to 64 bits.

rene:
Hi @marcrittinghaus , as we agreed today, I've updated this PR:

- I've reviewed your patch and applied basically all changes with minor fixes to GICv3 and gic common driver code
- GICv2: only major changes required to work along with GICv3 were introduced in this PR
- Tests were made with QEMU

Feature work to be done on GIC:
- Apply your improvements to GICv2 code (to be pushed by you)
- Introduce functions to SGI generation on GICv3 and in the common driver struct (I will do into a separated PR)

marcrittinghaus:
Hi @rene,

thank you very much for your great work. It is very appreciated. Again, apologizes for the hassle with the huge patch file!

marcrittinghaus (approved):

PR #237, opened 2021-06-25
lib/ukboot: add bindings for Mimalloc allocator

hlef:
This commits allows users to select Mimalloc as default system
allocator. Previously users had to patch the main tree to achieve this.
While adding more an more "hardcoded" allocators like this is certainly
not a long term solution, this is the only one we have for now.

Signed-off-by: Hugo Lefeuvre <hugo.lefeuvre@manchester.ac.uk>


skuenzer (approved):
Looks good! Thanks a lot!

PR #239, opened 2021-06-25
build: Option to compile as PIE

danield20:
This patch adds the option to compile the unikernel as
a position-independent executable so we can have ASLR.
If the unikernel is compiled as PIE, it cannot run on it's
own. A bootloader will be needed that will come in a future PR.

Signed-off-by: Daniel Dinca <dincadaniel97@gmail.com>

### DIFF
--- a/Config.uk
+++ b/Config.uk
@@ -86,6 +86,15 @@ config OPTIMIZE_LTO
 		will increase overall building time but creates more efficient
 		Unikraft binaries.
 
+config ENABLE_PIE
+	bool "PIE - Postition-Independent Executable"
+	default n
+	help
+		Enables the unikernel to be compiled as a position-independent executable.
+		Note that certain parts of the entry64.S file will be eliminated, and the unikernel
+		will need to be loaded into memory from initrd by the bootloader (which will be compiled
+		separately).
+
 choice
 	prompt "Debug information level"
 	default DEBUG_SYMBOLS_LVL3
--- a/Makefile.uk
+++ b/Makefile.uk
@@ -27,7 +27,7 @@ LIBLDFLAGS-$(call have_gcc)	+= -nostdinc
 # Set the text and data sections to be readable and writable. Also,
 # do not page-align the data segment. If the output format supports
 # Unix style magic numbers, mark the output as OMAGIC.
-LDFLAGS	+= -nostdlib -Wl,--omagic -Wl,--build-id=none
+LDFLAGS	+= -nostdlib -Wl,--build-id=none
 LDFLAGS-$(call have_gcc)	+= -nostdinc
 
 COMPFLAGS-$(CONFIG_OPTIMIZE_NONE)         += -O0 -fno-optimize-sibling-calls -fno-tree-vectorize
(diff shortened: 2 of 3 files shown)


skuenzer:
@danield20 Can you fix the checkpatch error message?

Config.uk
```diff
@@ -86,6 +86,15 @@ config OPTIMIZE_LTO
 		will increase overall building time but creates more efficient
 		Unikraft binaries.
 
+config ENABLE_PIE
```
skuenzer, line 89: Hey, can you call this option OPTIMIZE_PIE ? This would be inline with the other optimization flags and could be catched by our build script that watches changes of compiler optimization flags.

Config.uk
```diff
@@ -86,6 +86,15 @@ config OPTIMIZE_LTO
...
+	bool "PIE - Postition-Independent Executable"
+	default n
+	help
+		Enables the unikernel to be compiled as a position-independent executable.
+		Note that certain parts of the entry64.S file will be eliminated, and the unikernel
+		will need to be loaded into memory from initrd by the bootloader (which will be compiled
+		separately).
```
skuenzer, line 96: This description is somewhat KVM specific. Could you provide a more general description that just says that code is compiled position independent.

Makefile.uk
```diff
@@ -58,13 +58,20 @@ M4FLAGS      += -DUK_VERSION=$(UK_VERSION).$(UK_SUBVERSION)
 # If GCC supports "-no-pie" flag, we will add this flag to link flags to
 # override "pie" option, because some distributions will set
 # "--enable-default-pie" by default.
+ifeq ($(CONFIG_ENABLE_PIE),y)
+COMPFLAGS-$(call gcc_version_ge,6,1)	+= -fPIC
+LDFLAGS-$(call gcc_version_ge,6,1)	+= -static-pie
+else
+# "--omagic" is not compatible with the -static-pie option
```
skuenzer, line 65: Do you know if we could get rid of this flag in general?

plat/kvm/x86/entry64.S
```diff
@@ -40,6 +40,8 @@
 #define MYMULTIBOOT_FLAGS \
     (MULTIBOOT_PAGE_ALIGN | MULTIBOOT_MEMORY_INFO | MULTIBOOT_AOUT_KLUDGE)
 
+#ifndef CONFIG_ENABLE_PIE
+
```
skuenzer, line 44: Instead of modifying the KVM platform that still can't be booted with this change, I would disable the KVM platform as soon PIE is enabled. As long as we do not have a first stage ASLR bootloader for KVM I would consider the platform incompatible. Linuxu should work with PIE, do you know if Xen works?

skuenzer (changes requested):
Hey @danield20, I left you some comments in-line. I think this series could be integrated soon. Thanks, Simon

vladandrew:
@danield20 Any update on this?

danield20:
Hey @vladandrew, I haven't been looking into this actively, but I will try to look at it over the weekend and make an update.

razvand:
@gaulthiergain and one of his students (Terry) are working on this. They aim to provide a PR in early May.

danield20:
The status on this is that it currently works only on `gcc 8.4.0`  and `ldd 2.27.0`. I took a look but I can't tell what changes on newer versions of gcc and ldd that it doesn't work. I was in touch with both @gaulthiergain and Terry (his student) and they are both aware of this.

The behavior on newer versions is that it compiles successfully but the final executable is no longer PIE.

gaulthiergain:
@danield20 @razvand Yes, we are currently investigating this topic with @TerryLos .

We realized that there are several things not optimal with this implementation (micro-libs are not shuffled, it only works with a specific version of the linker, etc). As a first step, we will provide a P/R with ASLR at the linking level.

danield20:
@razvand can we close this?

gaulthiergain:
@razvand @danield20 Terry has implemented another approach concerning ASLR (P/R #492).

gabrielmocanu:
@razvand @danield20 
What is the status of this PR?
Will this be deprecated and we will continue with #492 ?

marcrittinghaus:
This PR will be superseded by #772

PR #240, opened 2021-06-26
Do not allocate IRQ handler entries dynamically

hlef:
IRQ handler entries are currently allocated dynamically. This has two undesired consequences: (1) this requires the dynamic memory allocator to be initialized when `ukplat_irq_init()` is called, which happens quite early, and (2) this makes the registering of IRQ handlers quite expensive.

Pre-allocate IRQ handles statically. Their number can be configured via a new config variable `KVM_MAX_IRQ_HANDLER_ENTRIES`.

This PR contains patches for KVM and linuxu but not Xen.

This PR is a simple re-upload of the fairly old [patch series 1431](https://patchwork.unikraft.org/project/unikraft/list/?series=1431). I rebased on top of staging, but it probably deserves a closer look and a bit more testing.

### DIFF
--- a/plat/kvm/Config.uk
+++ b/plat/kvm/Config.uk
@@ -77,6 +77,11 @@ endif
 
 endmenu
 
+config KVM_MAX_IRQ_HANDLER_ENTRIES
+       int "Maximum number of handlers per IRQ"
+       default 8
+       depends on (ARCH_X86_64 || ARCH_ARM_64)
+
 config KVM_PCI
        bool "PCI Bus Driver"
        default y
--- a/plat/kvm/irq.c
+++ b/plat/kvm/irq.c
@@ -27,46 +27,54 @@
 
 #include <stdlib.h>
 #include <uk/alloc.h>
-#include <uk/list.h>
 #include <uk/plat/lcpu.h>
 #include <uk/plat/time.h>
-#include <uk/plat/common/cpu.h>
+#include <uk/plat/irq.h>
 #include <uk/plat/common/irq.h>
 #include <kvm/irq.h>
 #include <kvm/intctrl.h>
 #include <uk/assert.h>
+#include <uk/print.h>
 #include <errno.h>
 #include <uk/bitops.h>
 
-static struct uk_alloc *allocator;
-
+/* IRQ handlers declarations */
 struct irq_handler {
 	irq_handler_func_t func;
 	void *arg;
-
-	UK_SLIST_ENTRY(struct irq_handler) entries;
 };
 
-UK_SLIST_HEAD(irq_handler_head, struct irq_handler);
-static struct irq_handler_head irq_handlers[__MAX_IRQ];
+static struct irq_handler irq_handlers[__MAX_IRQ]
+				[CONFIG_KVM_MAX_IRQ_HANDLER_ENTRIES];
+
+static inline struct irq_handler *allocate_handler(unsigned long irq)
+{
+	UK_ASSERT(irq < __MAX_IRQ);
+	for (int i = 0; i < CONFIG_KVM_MAX_IRQ_HANDLER_ENTRIES; i++)
+		if (irq_handlers[irq][i].func == NULL)
+			return &irq_handlers[irq][i];
+	return NULL;
+}
 
 int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
 {
 	struct irq_handler *h;
 	unsigned long flags;
 
-	UK_ASSERT(irq < __MAX_IRQ);
-	UK_ASSERT(allocator != NULL);
+	UK_ASSERT(func);
+	if (irq >= __MAX_IRQ)
+		return -EINVAL;
 
-	h = uk_malloc(allocator, sizeof(struct irq_handler));
-	if (!h)
+	flags = ukplat_lcpu_save_irqf();
+	h = allocate_handler(irq);
+	if (!h) {
+		ukplat_lcpu_restore_irqf(flags);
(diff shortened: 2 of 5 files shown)


hlef:
The review should probably be done by @skuenzer (he was the assigned reviewer on Patchwork).

marcrittinghaus (changes requested):
Hi Hugo, thank you very much :smiley: 👍🏻 

The patch already looks good. I have added some details and changed some parts and I would like to incorporate this. It makes sense to put some of the more "cleanup" type changes into one or two separate commits.

**KVM**:
- Callers of `ukplat_irq_register` expect to receive an `errno`. So I added the runtime check for `irq` from linuxu and replaced the fatal crash.
- Moved setting of h->arg. Considering that this function is called during initialization, having the additional overhead of a single mem write is not critical in the IRQ blocked section. However, it makes it more clear that these fields belong together and it ensures that there cannot be a situation in which `func` is set but not `arg` (the IRQ might already be active if another handler has been registered for that IRQ).
- Added an extra `UK_ASSERT` to `_ukplat_irq_handle` to check `irq`

**Linuxu**:
- Removed the unnecessary `uk/list.h` include.
- Changed `IRQS_NUM` to `__MAX_IRQ` to be consistent across platforms
- Removed the `RESERVED` state. Instead, call `sigaction` with signals blocked, however, only for the first handler of each IRQ. I don't think this is a serious issue as Linuxu is primarily a development platform and the IRQ handler registration happens very early in the boot phase. What do you think?
- Removed the `oldaction` field. The current code does not provide any means to deregister a handler. Furthermore, when you would deregister the first handler if there are further handlers configured, restoring the old action would disable the signal and break IRQ handling. It is not defined how to properly restore state in this case.

**Both**:
- Moved the existing or added a new `UK_ASSERT` to the new `allocate_handler` function because `irq` is actually used only here to dereference the array. This also makes the function 'safe' to use irrespective of the call chain (e.g., possible future reuse)
- I added an assert for `func` (we can expect this to be invalid API usage on all platforms and not a runtime issue, so no if).
- Removed all references to `uk_alloc`, also in the initialization function, the corresponding header, and also in the implementation in XEN.
- Added an assert in `ukplat_irq_init` that IRQs are disabled on all platforms so that we have consistent behavior.

You find my changes to you PR in this patch
[pull_240_patch_v1.zip](https://github.com/unikraft/unikraft/files/6790216/pull_240_patch_v1.zip)

hlef:
Hi @marcrittinghaus,

Thanks for the review! Sorry for taking so much time to reply.

I have reworked my patch with some of your comments.

I left out a couple of things:
- I am reluctant to remove the `struct uk_alloc *a __unused` argument from `ukplat_irq_init`. Doing so has little (immediate) benefits, and will definitely break external platforms. If someone wants to take a look at them later and decide to remove this argument, fine. I don't think that it has to be done in this PR.
- I left out most of the linuxu refactoring. It's nice, but might contain bugs, and is not really related to this PR. We should probably merge it as part of another, unrelated, cleanup PR.

I did not have time to test these changes yet, if you cannot give it a shot let me know and I'll try to free up some time end of next week :)

nderjung:
Hi @hlef! Thanks for the PR :) The CI/CD pipeline for `arm64-kvm` will always build now (before there was a continious error for all PRs), so it's worth checking its output if it fails.  In this case, it is and it's failing because of the changes in this PR:

```
/tmp/build/80754af9/built/unikraft/plat/kvm/arm/lcpu.c: In function 'ukplat_lcpu_halt_irq':
/tmp/build/80754af9/built/unikraft/plat/kvm/arm/lcpu.c:49:2: warning: implicit declaration of function 'halt' [-Wimplicit-function-declaration]
   49 |  halt();
      |  ^~~~
CC      libkvmplat: intctrl.o
CC      libkvmplat: shutdown.o
CC      libkvmplat: memory.o
CC      libkvmplat: irq.o
/tmp/build/80754af9/built/unikraft/plat/kvm/irq.c:48:6: error: 'CONFIG_KVM_MAX_IRQ_HANDLER_ENTRIES' undeclared here (not in a function)
   48 |     [CONFIG_KVM_MAX_IRQ_HANDLER_ENTRIES];
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/build/80754af9/built/unikraft/plat/kvm/irq.c:47:27: warning: 'irq_handlers' defined but not used [-Wunused-variable]
   47 | static struct irq_handler irq_handlers[__MAX_IRQ]
      |                           ^~~~~~~~~~~~
make[1]: *** [/tmp/build/80754af9/built/unikraft/support/build/Makefile.build:50: /tmp/build/80754af9/built/apps/helloworld/build/libkvmplat/irq.o] Error 1
make: *** [Makefile:1015: sub-make] Error 2
make: Leaving directory '/tmp/build/80754af9/built/unikraft'
```

hlef:
Thanks @nderjung, I need to find time and make a proper pass on this...

cristian-vijelie:
The PR now builds for linuxu and kvm, for x86_64, after the rebase.

cristian-vijelie (approved):
Good to go
Reviewed-by: Cristian Vijelie <cristianvijelie@gmail.com>

marcrittinghaus:
Xen was not building anymore due to a missing include of `<uk/assert.h>`, so I quickly fixed this. Also arm64 needed a fix. 

I also tested on KVM and Xen x86_64 to serve a webpage with `nginx`. Works as expected.

marcrittinghaus (approved):
All fine. @hlef Thank you very much!

PR #241, opened 2021-06-27
Add initial embedded rust support

vladandrew:
This patch adds embedded rust support. Only the core crate is available. Internal libraries may be written in rust now.
Right now, the sources should have #![no_std] and a panic handler but future work will integrate this in a rust internal library.

### DIFF
--- a/Config.uk
+++ b/Config.uk
@@ -168,6 +168,15 @@ config CROSS_COMPILE
 		need to set this unless you want the configured build
 		directory to select the cross-compiler automatically.
 
+
+config LLVM_TARGET_ARCH
+	string "Custom cross-compiler LLVM target (optional)"
+	help
+		Same as running 'make LLVM_TARGET_ARCH=target' but stored for
+		default make runs in this build directory.  You don't
+		need to set this unless you want the configured build
+		directory to select the cross-compiler automatically.
+		This is used for LLVM based compilers such as clang or rustc.
 #config PARALLEL_JOBS
 #	int "Number of jobs to run simultaneously (0 for auto)"
 #	default "0"
--- a/Makefile
+++ b/Makefile
@@ -319,6 +319,8 @@ CXXINCLUDES :=
 CXXINCLUDES-y :=
 GOCFLAGS :=
 GOCFLAGS-y :=
+RUSTCFLAGS :=
+RUSTCFLAGS-y :=
 GOCINCLUDES :=
 GOCINCLUDES-y :=
 DBGFLAGS :=
@@ -568,6 +570,13 @@ CC		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CPP		:= $(CC)
 CXX		:= $(CPP)
 GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
+# We use rustc because the gcc frontend is experimental and missing features such
+# as borrowing checking
+ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
+RUSTC		:= rustc --target=$(CONFIG_LLVM_TARGET_ARCH)
+else
+RUSTC		:= rustc
+endif
 AS		:= $(CC)
 AR		:= $(CONFIG_CROSS_COMPILE)gcc-ar
 NM		:= $(CONFIG_CROSS_COMPILE)gcc-nm
--- a/Makefile
+++ b/Makefile
@@ -515,6 +515,7 @@ ifeq ($(sub_make_exec), 1)
 ifeq ($(UK_HAVE_DOT_CONFIG),y)
 # Hide troublesome environment variables from sub processes
 unexport CONFIG_CROSS_COMPILE
+unexport CONFIG_LLVM_TARGET_ARCH
 unexport CONFIG_COMPILER
 #unexport CC
 #unexport LD
@@ -546,6 +547,10 @@ unexport MACHINE
 
 ifneq ("$(origin CROSS_COMPILE)","undefined")
 CONFIG_CROSS_COMPILE := $(CROSS_COMPILE:"%"=%)
+endif
+
+ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
+CONFIG_LLVM_TARGET_ARCH := $(LLVM_TARGET_ARCH:"%"=%)
 endif
 
 ifneq ("$(origin COMPILER)","undefined")
--- a/Makefile.uk
+++ b/Makefile.uk
@@ -21,6 +21,30 @@ CINCLUDES    += -I$(CONFIG_UK_BASE)/include
 CXXINCLUDES  += -I$(CONFIG_UK_BASE)/include
 GOCINCLUDES  += -I$(CONFIG_UK_BASE)/include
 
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -563,6 +565,7 @@ CC		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CPP		:= $(CC)
 CXX		:= $(CPP)
 GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
+RUSTC		:= rustc
```
skuenzer, line 578: How would cross compiling work with rust? I would have expected something like
```Makefile
RUSTC		:= $(CONFIG_CROSS_COMPILE)rustc
```

support/build/Makefile.rules
```diff
@@ -540,6 +540,21 @@ $(eval $(call vprefix_lib,$(1),OBJS-y) += $(4))
 $(eval $(call vprefix_lib,$(1),CLEAN-y) += $(call build_clean,$(4)) $(call out2dep,$(4)))
 endef
 
+define buildrule_rs =
+$(4): $(2) | preprocess
+	$(call build_cmd,RUSTC,$(1),$(4),\
+		$(RUSTC) $$(RUSTCFLAGS) $$(RUSTCFLAGS-y) \
```
skuenzer: Can we also include the general compiler flags that we have, like the optimization flags, flags for ISR code and the file-variant specific flags?
Is `rustc` GCC flag compatible?

skuenzer (commented):
Hey, this looks really cool. I have some question regarding the integration and adoption of our flags that are set for the compilation units. Most importantly, DCE, LTO and ISR-safe code.

vladandrew:
Hi Simon, thank you for your review, I've solved all the comments and pushed the next version of the patch!

Makefile
vladandrew: Rustc is using --target for specifying the target, I've modified the code accordingly.

support/build/Makefile.rules
vladandrew: Rustc is not GCC compatible, I've adapted the debug and optimization flags accordingly.

Makefile
skuenzer: I think `$(CONFIG_CROSS_COMPILE)` includes a minus. We probably just need to remove the trailing `-` from the variable. However, do you know if the naming is exactly the same or do we need translating it? A tiny support script could do the job, e.g., `support/scripts/rustc-cross`

Makefile.uk
```diff
@@ -21,6 +21,23 @@ CINCLUDES    += -I$(CONFIG_UK_BASE)/include
...
+RUSTCFLAGS-$(CONFIG_OPTIMIZE_SIZE)         += -Copt-level="s"
+RUSTCFLAGS-$(CONFIG_OPTIMIZE_PERF)         += -Copt-level="2"
+
+RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL0)     += -Cdebuginfo=0
+RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL1)     += -Cdebuginfo=1
+RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL2)     += -Cdebuginfo=2
+# There is not level 3 in rustc
+RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL3)     += -Cdebuginfo=2
+
+# Works only with clang
```
skuenzer: Please add this as a note.
```Makefile
# NOTE: rustc supports LTO only with clang
```

support/build/Makefile.rules
```diff
@@ -540,6 +540,24 @@ $(eval $(call vprefix_lib,$(1),OBJS-y) += $(4))
 $(eval $(call vprefix_lib,$(1),CLEAN-y) += $(call build_clean,$(4)) $(call out2dep,$(4)))
 endef
 
+# We are not using most of the flags such as COMPFLAGS due to incompatibilities between rustc and GCC.
```
skuenzer:
```Makefile
# NOTE: ...
```

support/build/Makefile.rules
skuenzer: Are there includes that need to be considered for `rustc`?
Can `rustc` generate something for dependency management?

vladandrew:
Hi Simon, I've pushed a new version of this series. I've added a new patch for LLVM_TARGET_ARCH.

Makefile
```diff
@@ -568,6 +570,13 @@ CC		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CPP		:= $(CC)
 CXX		:= $(CPP)
 GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
+# We use rustc because the gcc frontend is experimental and missing features such
+# as borrowing checking
+ifneq ("$(origin CROSS_COMPILE)","undefined")
+RUSTC		:= rustc --target=$(CONFIG_CROSS_COMPILE)
```
skuenzer: Shouldn't this be set to `LLVM_TARGET_ARCH`?

Makefile
```diff
@@ -515,6 +515,7 @@ ifeq ($(sub_make_exec), 1)
 ifeq ($(UK_HAVE_DOT_CONFIG),y)
 # Hide troublesome environment variables from sub processes
 unexport CONFIG_CROSS_COMPILE
+unexport LLVM_TARGET_ARCH
```
skuenzer:
```diff
+unexport CONFIG_LLVM_TARGET_ARCH
```

Config.uk
```diff
@@ -168,6 +168,14 @@ config CROSS_COMPILE
...
 		directory to select the cross-compiler automatically.
 
+
+config LLVM_TARGET_ARCH
+	string "Custom cross-compiler LLVM target (optional)"
+	help
+		Same as running 'make LLVM_TARGET_ARCH=target' but stored for
+		default make runs in this build directory.  You don't
+		need to set this unless you want the configured build
+		directory to select the cross-compiler automatically.
```
skuenzer, line 178: Maybe add another sentence that this is used when compiling with LLVM compiles, like clang or rustc.

skuenzer (changes requested):

Makefile
vladandrew: Solved :100:

Makefile
vladandrew: Done

support/build/Makefile.rules
skuenzer: Can you also add `$$(RUSTCFLAGS_EXTRA)` to get the functionality as introduced with PR #255 ?

skuenzer (changes requested):
Can you add an extra flag called `$$(RUSTCFLAGS_EXTRA)` to get the functionality as introduced with PR #255 ?

skuenzer (approved):
Looks good to me. Thanks!

nderjung:
Hi,

@skuenzer can you re-approve this PR following @vladandrew new rebase?  I tried re-running the merge pipeline but it fails because it's re-using the state at of unikraft's `staging` branch of when you sent the original approval.

Thanks!

skuenzer (approved):
Looks good, thanks a lot!

PR #243, opened 2021-06-30
lib/uknetdev: Define and implement burst transactions for netdevice

hisharanwithu:
This patch series enables batching the send/receive operation on
a network packets on a netdev. The implementation extends the
virtio-net driver to support batch operation.

- plat/virtio: Remove warning in virtio_netdev_recv
- lib/uknetdev: Support burst tx
- lib/uknetdev: Support to check UK_NETDEV_UNDERRUN status
- lib/uknetdev: Support burst receive
- plat/virtio: Support ANY_LAYOUT on virtio_net
- lib/uknetdev: Release a burst of packet from ring

### DIFF
--- a/lib/uknetdev/Config.uk
+++ b/lib/uknetdev/Config.uk
@@ -30,4 +30,11 @@ if LIBUKNETDEV
 			When this option is enabled a dispatcher thread is
 			allocated for each configured receive queue.
 			libuksched is required for this option.
+
+	config LIBUKNETDEV_MAX_PKT_BURST
+		int "Netdev burst transaction"
+		default 64
+		help
+			The maximum number of packets to be sent/received on a
+			single burst transaction.
 endif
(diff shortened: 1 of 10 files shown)


Zack-Will:
I see that netdevice supports burst in earlier versions, why was this feature removed?

razvand:
> I see that netdevice supports burst in earlier versions, why was this feature removed?

@skuenzer, could you answer this?

skuenzer:
> I see that netdevice supports burst in earlier versions, why was this feature removed?

Hi @Zack-Will,
in fact, burst behavior was never upstream but planned for uknetdev. We planned to update this PR with a slight different API, however, we did not manage to do so yet because of different priorities. On the other hand, you also need networking stacks supporting bursts, the use is limited to custom network functions. Are you interested in such a functionality?

Thanks,

Simon

PR #244, opened 2021-06-30
plat/kvm/x86: Add SMP support

cristian-vijelie:
This series of patches enable Unikraft to utilize more than one core. First, the SMP API is defined, then the functions needed to signal a core to wake up and to make it able to run code from Unikraft are implemented.


include/uk/plat/lcpu.h
```diff
@@ -1,104 +1,164 @@
...
+	void *user;
+};
+
+struct ukplat_cpu {
+	ukplat_lcpu_entry_t entry;
+	void *stackp;
+	enum lcpu_states state;
+	struct uk_list_head fnlist;
+	struct __regs *trap_frame;
+};
```
marcrittinghaus: The requirements on what has to be stored in the struct are very architecture-specific. For example, on x86, we need to add the exceptions stacks as part of the struct. So please move this to the `smp.c` file. The cpu struct should not be publicly visible. The same goes for the states.

include/uk/plat/lcpu.h
```diff
@@ -1,104 +1,164 @@
...
+
+/**
+ * Return the number of logical CPUs present on the system
+ */
+__lcpuid ukplat_lcpu_count(void);
+
+/**
+ * Return whether the current logical CPU is the Bootstrapping one
+ */
+int ukplat_lcpu_is_bsp(void);
```
marcrittinghaus: We should keep the API minimal and do not introduce architecture specific terms. Is there a specific reason, you want this function to be part of the API? Where is it needed?

include/uk/plat/lcpu.h
```diff
@@ -1,104 +1,164 @@
...
+
+/**
+ * Wait for the given lcpus to enter the "halted" state, or until the timeout
+ * expires. If lcpuid[] is NULL, wait for all the lcpus, except the current
+ * one.
+ * @return 1, if the timeout expired, 0 otherwise
+ */
+
+int ukplat_lcpu_wait(__lcpuid lcpuid[], int num, __nsec timeout);
```
marcrittinghaus: Typo: should be idle state and not halted

lib/ukdebug/print.c
```diff
@@ -272,13 +272,18 @@ void _uk_vprintd(const char *libname, const char *srcname,
 #endif
 }
 
+static int lock;
+
```
marcrittinghaus: In the end, we should make this PR depend on the spinlock PR to synchronize printing. Also I think this is a more general problem and console printing needs to be synchronized on a lower level. Think about, for example, the `vfprintf` function in `lib/nolibc` which calls `ukplat_coutk` and `ukplat_coutd` and would thus bypass synchronization. We could also have an output device that does some form of caching/batching because it sends console output to a file that would not need full synchronization all the time (reserve memory atomically and then copy line). So I suggest to move synchronization into the respective implementations of `ukplat_coutk` and `ukplat_coutd`. A spinlock is fine for console printing.

plat/Config.uk
```diff
@@ -12,6 +12,15 @@ config EARLY_PRINT_PL011_UART_ADDR
...
 		Pl011 serial address used by early debug console.
 
+config HAVE_SMP
+	bool "Symmetric multi-processor support"
+	default n
+
+config MAX_CPUS
+	int
+	depends on HAVE_SMP
+	prompt "The maximum nuber of supported CPUs"
```
marcrittinghaus: Typo: number

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+
+	/* get the AP id */
+	mov	$1, %eax
+	cpuid
+	shrl	$24, %ebx
+
+	/* Notify the BSP that the core has started successfully */
+	lock	incq smp_aprunning
+
+	/* Taken from entry64.S */
```
marcrittinghaus: I don't like that the code is replicated. You could refactor this to reside in its own function that is called for both the BSP and APs. Even better, merge the .S files for the BSP and AP into one file and share code where possible.

plat/kvm/x86/traps.c
```diff
@@ -129,49 +129,49 @@ static void idt_init(void)
 	 */
```
marcrittinghaus, line 201: We have a more complicated problem here. We must not use IST for switching to interrupt and exception stacks in general. This is because, IST will always set `rsp` to the stack base not matter if we are already executing on the very same stack. That means, if we have for example a debug exception and produce a page fault because the kernel debugger probes an invalid address for access, the stack gets corrupted (the page fault handler overwrites state already stored on the stack by the debug handler). Another example would be to intentionally use a trap in order to get a consistent register snapshot when failing an `UK_ASSERT`. If the assertion is placed in the code path of a trap handler, we also get nested calls.

The very same problem is true for interrupts. At the moment, we enable and disable interrupts all at once. However, it is perfectly valid to just mask certain interrupts and still allow others to be delivered if they don't break synchronization. For example, it would be ok to have the network interrupt delivered while being in the handler for the disk driver. So in the long run, we don't want to use IST for interrupts as well.

Generally, IST is designed for cases in which you have to guarantee that a functional stack is present. Therefore, IST always resets the `rsp` to the stack base, no matter if we are already using that stack. Because otherwise, we could run into a stack overflow (i.e., end up with *NOT* having a valid stack despite IST). But that means, IST cannot be used whenever nested invocations may be possible.

Where we can use IST is for `double fault`, `nmi`, and `machine check`, which usually represent critical situations that we currently do not handle and perhaps won't in the near future. Except for the `NMI` flag, where we should add the scan for functions to execute. Otherwise, we are just crashing in these situations at the moment. I think it is fine to reserve one common stack for these three cases. This will be our `panic` stack, which should be switched to via IST.

But also for the traps and interrupts, we would like to use two separate stacks and do not use the current application stack (as done with setting IST to `0`). Depending on the application and the current runtime state, the stack might provide too little space to handle the interrupt (for example, GO may use very small stacks for the go coroutines). Or we could have an interrupt during a stack switch in a user-level threading library where we do not even have a valid `rsp` configured. To gracefully handle those cases, we need a similarly atomic HW stack switch like with IST because the interrupt handler already pushes data onto the stack (that is, before we could switch manually). However, unlike IST, the stack switch must respect the case, where we have nested invocations.

1) Please investigate how this is generally done in other operating systems and what x86 features could be used for this. It could be that a task switch (TSS) is used for this as it also allows to switch stacks. However, I am not sure if this can handle nested cases. Also in many cases, stack switches are only done during privilege changes, that we do not have. It could also be that in our case, we need a manual stack switch in the trap handler and a careful handling of double faults to cope with corner cases (e.g., stack overflow). Ideally, we would like to have the option to avoid stack switches altogether and only do them if we run into trouble (i.e., on a double fault). But this would definitely be an option, because we might not be able to detect stack overflows in all cases; for example, with GO where stacks allocates its own stacks from a large heap segment.

2) You can add the `panic`, `trap`, and `int` stacks to the cpu struct. For now, just use `STACK_SIZE`, but we probably will define a custom size to not waste too much memory. 

3) Implement IST for the `panic` stack. This includes properly configuring TSS for APs. It would be ideal to remove the dynamic table initialization and just provide prepared tables in the `.data` segment (just like with the fixed page table). This saves initialization time and makes it easier to bring up secondary CPUs. However, this also depends on what you find out for (1) and how this should be solved. For example, if we need differently configured data structures for each CPU (e.g., TSS) we would have to generate them dynamically at build or run time and store them as part of the cpu struct. However, we are probably able to share at least some of them (e.g., IDT, GDT).

4) Implement the NMI handler that executes scheduled NMI routines.

plat/kvm/x86/setup.c
```diff
@@ -289,7 +293,12 @@ void _libkvmplat_entry(void *arg)
 		   (void *) _libkvmplat_cfg.bstack.start);
 
 #ifdef CONFIG_HAVE_SMP
-	acpi_init();
+	ret = acpi_init();
+	if (ret != 0)
+		uk_pr_err("ACPI Init failed with error %d\n", ret);
+	ret = smp_init();
+	if (ret != 0)
+		uk_pr_err("SMP init failed\n");
```
marcrittinghaus, line 465: It won't make sense to initialize smp, when ACPI initialization failed.

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+	cmpb 	$0, bspdone
+	jz 	spin
+
+	/* get the AP id */
+	mov	$1, %eax
+	cpuid
+	shrl	$24, %ebx
+
+	/* Notify the BSP that the core has started successfully */
+	lock	incq smp_aprunning
```
marcrittinghaus: This variable should be removed. We have this information from the per-cpu state.

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+nopku:
+#endif /* CONFIG_HAVE_X86PKU */
+	/* Load IDT */
+	lidt 	(idtptr)
+
+	/* Use CPU ID * cpustruct_size for indexing the cpu structure */
+	movq 	%r9, %rax
+	movq	cpustruct_size, %r10
+	mulq 	%r10
+	movq 	%rax, %r10
```
marcrittinghaus: Please add a define at the cpu struct that has the size of the CPU struct as a plain number and use this define here. Add a comment at the cpu struct that informs other programmers that they have to update the define if they change the struct size. This is not very nice, either, but is in line with the current design (e.g., see the trap frame `__regs`) and is faster at runtime, because you safe the memory access.

include/uk/plat/lcpu.h
```diff
@@ -1,104 +1,164 @@
...
+ */
+void ukplat_lcpu_halt_irq(void);
+
+#ifdef CONFIG_HAVE_SMP
+#include <uk/list.h>
+
+typedef void (*ukplat_lcpu_entry_t)(void) __noreturn;
+typedef __u32 __lcpuid;
+
+enum lcpu_states { offline, busy, idle, waiting };
```
marcrittinghaus: Since you set the CPU states in the startup code, please use `DEFINE`s for the states values. You may use an enum on the C-side where the value of each enum value is set to the `DEFINE`. Since enum values are added to the global namespace in C, please use distinctive names such as `LCPU_STATE_OFFLINE` etc.

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+
+	test	%rax, %rax
+	jz 	halt_prep
+
+	/* Set the AP as running */
+	movq	$1, 16(%rsi, %r10, 1)
+	jmp	*%rax
+
+halt_prep:
+	lock 	incl aps_halted
```
marcrittinghaus: This variable is not needed and can be removed. Halting CPUs will be done by executing a halt routine using `ukplat_lcpu_run()`. Waiting for CPUs to reach `HLT` state can then be done easily in C using the `user` attribute for the function.

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+halt:
+	hlt
+	jmp 	halt
+END(_lcpu_start64)
```
marcrittinghaus: 1) Please use `DEFINE`s as suggested for the state enumeration and also add offset `DEFINE`s (just like as for the `__regs` definition) to refer to the individual fields of the cpu struct.
2) It should be `idle` state in line 226
3) Move setting the idle state into the halt loop. So you do not change state twice in case we have an entry function. It also means, we can be sure that the state is always set to idle before going into the `hlt`.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+	intctrl_init();
+
+	ukplat_lcpu_enable_irq();
+
+	ret = ukplat_irq_register(UKPLAT_X86_IPI_RUN_IRQ,
+				  ukplat_lcpu_ipi_run_handler, NULL);
+	if (ret)
+		uk_pr_crit("Interrupt handler for IRQ %d failed\n",
+			   UKPLAT_X86_IPI_RUN_IRQ);
+	ret = ukplat_irq_register(UKPLAT_X86_IPI_WAKEUP_IRQ,
```
marcrittinghaus: The purpose of the wake up handler was to avoid costly trap frame generation and calling handle interrupt etc. just for kicking the CPU out of halt. So you have to write a dedicated trap handler here in ASM that just does an `iret` without calling any C functions.

plat/kvm/include/kvm-x86/smp.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __PLAT_KVM_X86_SMP_H__
+#define __PLAT_KVM_X86_SMP_H__
+
+#include <uk/arch/types.h>
+
+#define UKPLAT_X86_IPI_FIXED	0
+#define UKPLAT_X86_IPI_NMI	1
```
marcrittinghaus: As this should be used in a flags field and not an enum or type designator it is enough to just define the `NMI` flag. Please rename the flag to something like `X86_LCPU_RUN_FLAG_NMI`

plat/kvm/include/kvm-x86/smp.h
```diff
@@ -0,0 +1,15 @@
...
+
+#define UKPLAT_X86_IPI_RUN_IRQ		5
+#define UKPLAT_X86_IPI_WAKEUP_IRQ	6
```
marcrittinghaus: Is there a  reason for choosing these specific IRQ numbers?

plat/kvm/irq.c
```diff
@@ -77,8 +77,9 @@ int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
  * interrupts in order to stop waiting for interrupts with deadline.
  */
 extern unsigned long sched_have_pending_events;
+extern struct ukplat_cpu cpus[CONFIG_MAX_CPUS];
 
-void _ukplat_irq_handle(unsigned long irq)
+void _ukplat_irq_handle(unsigned long irq, struct __regs *regs)
```
marcrittinghaus, line 90: 1) This is an architecture independent function. This means that a patch must include all architectures. 

2) As the cpu struct is architecture-dependent you cannot access it here. We should therefore investigate if it is possible to add the `__regs` parameter to the interrupt handler function type without having to touch too much code. This would also eliminate the `trap_frame` variable from the cpu struct.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+{
+	int i, j;
+	__u32 eax, edx;
+	__lcpuid id;
+
+	smp_aprunning = 0;
+	if (smp_numcores == 0)
+		ukplat_lcpu_count();
+
+	memcpy((void *)0x8000, &_lcpu_start16, 4096);
```
marcrittinghaus: I don't like having to copy the function code. It should be possible to place the function into its own ELF section that is mapped at the corresponding address.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+	uk_pr_debug("Copied AP boot code to 0x8000\n");
+
+	/* Store the CR3 register */
+	__asm__ __volatile__("movq %%cr3, %0" : "=r"(cr3_val));
+
+	for (i = 0; i < num; i++) {
+		if (lcpuid != NULL)
+			id = lcpuid[i];
+		else
+			id = i;
```
marcrittinghaus: A more compact way: `id = (lcpuid) ? lcpuid[i] : i;`

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+
+		if (id == bspid)
+			continue;
+
+		if (apic_ids[id] == 0)
+			continue;
+
```
marcrittinghaus: Add comments why you exclude these CPUs

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+	mdelay(10);
+
+	return smp_aprunning;
+}
+
+__lcpuid ukplat_lcpu_id(void)
+{
+	__lcpuid id;
+
+	__asm__ __volatile__("mov $1, %%eax; cpuid; shrl $24, %%ebx;"
```
marcrittinghaus: It is a bit easier to work with if you just use the `cpuid` function and perform the rest of the steps in C.

On the long run, it we should have CPU-local storage, where we store the index and perhaps a direct pointer into the CPU struct. Calling `cpuid` is a costly operation.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+
+	return 0;
+}
+
+int ukplat_lcpu_wakeup(__lcpuid lcpuid[], int num)
+{
+	int i, eax, edx;
+
+	for (i = 0; i < num; i++) {
+		edx = lcpuid[i];
```
marcrittinghaus: wake up should work just like the other functions concerning a `NULL` lcpuid

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+	return 1;
+}
+
+int ukplat_lcpu_ipi_wakeup_handler(__attribute__((unused)) void *args)
+{
+	wrmsr(x2APIC_EOI, 0, 0);
+	return 1;
+}
+
+void __noreturn _lcpu_entry_default(void)
```
marcrittinghaus: Don't we have to do these initialization steps for all APs? In that case, it might make sense to always jump from the start IPI to this C function and implement jumping to the specified entry function (if any) as well as the HLT loop here.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+		if (lcpuid != NULL)
+			id = lcpuid[i];
+		else
+			id = i;
+
+		/* Don't wait for the current CPU */
+		if (id == ukplat_lcpu_id())
+			continue;
+
+		while (cpus[id].state != idle) {
```
marcrittinghaus: You have to be careful when spinning on a variable, as the compiler may deduce that its value cannot change in the body of the loop and optimize away the actual memory access. Mark the state field as `volatile` .

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+	/* set the x2APIC enable bit */
+	eax |= (1 << APIC_BASE_EXTD);
+	wrmsr(IA32_APIC_BASE, eax, edx);
+	uk_pr_debug("x2APIC is enabled\n");
+
+	return 0;
+}
+
+int ukplat_lcpu_ipi_run_handler(__attribute__((unused)) void *args)
+{
```
marcrittinghaus: Functions and structures that are not part of the official platform API should not posses the `ukplat_` prefix. Instead use per-module prefixes such as `x86_lcpu_ipi_run_handler` (also the cpu struct -> `x86_cpu`).

Remember to use the `static` keyword for all things that are not directly referenced from outside of the object file. If you supply a function as pointer and it should not be called otherwise (as it is in this case) you also mark it as `static`. This ensures that it cannot be called by other means than the explicitly given function pointer; that is under other circumstances than anticipated.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+
+	int id = ukplat_lcpu_id();
+
+	fnlist = &cpus[id].fnlist;
+	regs = cpus[id].trap_frame;
+
+	uk_list_for_each_entry(fn_entry, fnlist, lentry)
+	    fn_entry->fn(regs, fn_entry);
+
+	wrmsr(x2APIC_EOI, 0, 0);
```
marcrittinghaus: Please ensure that all CPUs are configured the same. That means for the BSP to also have the same APIC configuration.

Define a `static inline` function for the ack operation in the appropriate APIC header file.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+			      | APIC_ICR_DMODE_FIXED
+			      | (32 + UKPLAT_X86_IPI_RUN_IRQ);
+
+			wrmsr(x2APIC_ICR, eax, edx);
+		} else if (flags == UKPLAT_X86_IPI_NMI) {
+			eax = APIC_ICR_TRIGGER_LEVEL | APIC_ICR_LEVEL_ASSERT
+			      | APIC_ICR_DESTMODE_PHYSICAL | APIC_ICR_DMODE_NMI;
+
+			wrmsr(x2APIC_ICR, eax, edx);
+		} else
```
marcrittinghaus: To make this easier to comprehend, you should provide a `static inline` function for issuing an IPI with the IRQ number, possible additional flags and the CPUID as arguments. You could add another `static inline` function that calls this function to issue an NMI. You can the use this functions in all the locations where you trigger IPIs/NMIs.

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,462 @@
...
+#include <uk/plat/time.h>
+#include <uk/plat/irq.h>
+
+__u8 bspdone;
+__u64 smp_aprunning, cr3_val, idt_addr;
+volatile int aps_halted;
+
+static __lcpuid bspid;
+static __u32 smp_numcores;
+static __u32 apic_ids[256] = {0};
```
marcrittinghaus: Intergate the APIC id into the cpu struct

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,241 @@
...
+	lidt 	(idtptr)
+
+	/* Use CPU ID * cpustruct_size for indexing the cpu structure */
+	movq 	%r9, %rax
+	movq	cpustruct_size, %r10
+	mulq 	%r10
+	movq 	%rax, %r10
+
+	/* Load the stack pointer and the entry address */
+	movq	cpusptr, %rsi
```
marcrittinghaus: use the `lea` instruction and the actual cpu struct array symbol to calculate the address. Delete the `cpusptr` variable.

include/uk/plat/lcpu.h
```diff
@@ -1,104 +1,164 @@
...
+	struct uk_list_head lentry;
+	void (*fn)(struct __regs *regs, struct ukplat_lcpu_func *fn);
+	void *user;
+};
+
+struct ukplat_cpu {
+	ukplat_lcpu_entry_t entry;
+	void *stackp;
+	enum lcpu_states state;
+	struct uk_list_head fnlist;
```
marcrittinghaus: We could save a bit of memory by using a union. Something like this:
```C
struct x86_lcpu_startup_args {
	/* add field description */
	ukplat_lcpu_entry_t entry;
	/* add field description */			
	void *stackp;
};

struct x86_cpu {
	/* add comment why we can share memory */
	union {
		/* add field description */
		struct x86_lcpu_startup_args s_args;
		/* add field description */
		struct uk_list_head fn_list;
	};
	/* add field description */
	volatile enum x86_lcpu_states state;
};
```

Don't forget to initialize the list head using `UK_INIT_LIST_HEAD` after using the arguments.

marcrittinghaus (changes requested):
Hey Cristian,

thank you very much for your work on the x86 SMP implementation :fire: It is very appreciated and an important feature for Unikraft :+1: I suggest a number of changes and also want to use the review to start a discussion on how to handle stack switching. I am curious about your findings and ideas. I plan this review to be the first in a couple of rounds.

Please feel free to contact me anytime if you have questions regarding my comments or if you need a second opinion on how things could be implemented.

plat/kvm/include/kvm-x86/smp.h
cristian-vijelie: No, I didn't know which numbers to use, and I didn't find a mechanism to allocate them dynamically.

include/uk/plat/lcpu.h
cristian-vijelie: It came up in the ARM discussions, and I thought that it could be needed.

marcrittinghaus: Ok, I will look into this to get a better understanding of the reasoning behind this method.

plat/kvm/include/kvm-x86/smp.h
marcrittinghaus: Ok, then we have to think about this a bit more so we do not end up having IRQ number clashes

plat/kvm/x86/smp.c
cristian-vijelie: That would require always providing a stack to the AP, or them defining their own boot stacks somewhere

plat/kvm/x86/smp.c
```diff
@@ -0,0 +1,539 @@
...
+		end = ukplat_monotonic_clock() + timeout;
+
+	for (i = 0; i < num; i++) {
+		id = (lcpuid) ? lcpuid[i] : (__lcpuid)i;
+
+		/* Don't wait for the current CPU */
+		if (id == ukplat_lcpu_id())
+			continue;
+
+		while (cpus[id].state != LCPU_STATE_IDLE) {
```
razvanvirtan, line 320: @cristian-vijelie I had a synchronization issue with this function in the ARM64 implementation. There is a situation when a cpu makes the following calls:
 ```
 ukplat_lcpu_run(target_cpu....);
 ukplat_lcpu_wait(target_cpu....);
 ukplat_lcpu_run(target_cpu....);
 ```
 Now, there is a scenario when the source cpu can reach this state check before the target one updated the state to `LCPU_STATE_BUSY`. Therefore, the source will exit `ukplat_lcpu_wait()` before the target core finished it's job, will send the `LCPU_RUN` interrupt while the target core has interrupts still disabled, so the second `ukplat_lcpu_run()` call will be lost.
 I have solved this using a lock associated to each cpu [1], here is a figure with the race condition:
image
[1] https://github.com/unikraft/unikraft/pull/373

marcrittinghaus: Hey @razvanvirtan,

thanks for pointing out the issue and for the detailed explanation 🚀 Also for you @cristian-vijelie, I am currently working on this PR internally and will address all of these issues. I am also working on merging initialization code for BSP and AP and also fix some issues concerning missing initialization of CPU features (e.g., at the moment the CPU caches are disabled for APs). @cristian-vijelie Please do not push any updates until further notice.

@razvanvirtan I am also looking at the ARM version. When I finished my first iteration on the x86 SMP code, I will take a look at the ARM one. Afterwards, we can see if there is some more stuff we can define/move/merge in a platform/architecture agnostic way to avoid duplicate code.

marcrittinghaus:
As the code and the overall PR has been significantly changed for adopting the common SMP API, I close this PR and reopen a new one. This way all SMP related PRs also have consecutive PR numbers, which makes navigation easier.

PR #245, opened 2021-06-30
Add DLang Runtime Support

CristiBM:
This includes support for compilation of D sources and support for aligned pages in mmap()/munmap() - feature that is assumed to be available by the D runtime library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -321,6 +321,14 @@ GOCFLAGS :=
 GOCFLAGS-y :=
 GOCINCLUDES :=
 GOCINCLUDES-y :=
+DMDFLAGS :=
+DMDFLAGS-y :=
+DMDINCLUDES :=
+DMDINCLUDES-y :=
+GDCFLAGS :=
+GDCFLAGS-y :=
+GDCINCLUDES :=
+GDCINCLUDES-y :=
 DBGFLAGS :=
 DBGFLAGS-y :=
 LDFLAGS :=
@@ -563,6 +571,8 @@ CC		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CPP		:= $(CC)
 CXX		:= $(CPP)
 GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
+DMD		:= $(CONFIG_CROSS_COMPILE)dmd
+GDC		:= $(CONFIG_CROSS_COMPILE)gdc
 AS		:= $(CC)
 AR		:= $(CONFIG_CROSS_COMPILE)gcc-ar
 NM		:= $(CONFIG_CROSS_COMPILE)gcc-nm
@@ -598,6 +608,7 @@ ASFLAGS		+= -DCC_VERSION=$(CC_VERSION)
 CFLAGS		+= -DCC_VERSION=$(CC_VERSION)
 CXXFLAGS	+= -DCC_VERSION=$(CC_VERSION)
 GOCFLAGS	+= -DCC_VERSION=$(CC_VERSION)
+GDCFLAGS	+= -DCC_VERSION=$(CC_VERSION)
 
 # ensure $(BUILD_DIR)/kconfig, $(BUILD_DIR)/include and $(BUILD_DIR)/include/uk exists
 $(call mk_sub_build_dir,kconfig)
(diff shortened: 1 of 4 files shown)


CristiBM:
@skuenzer I'm referencing here the conversation thread from patchwork so that it may be continued.
Simon Kuenzer:
        "Hi Marius-Cristian,

        thanks a lot for your work! I am actually wondering if we should rather 
        move this to a D language support library instead of maintaining it with 
        the core Unikraft repository - especially if a language runtime library 
        is anyway required (is this true?). It is also planned that we move the 
        GO compile rules out of the main repository. I am concerned that at some 
        point we have too many languages to maintain within the core Unikraft 
        repository. What do you think?
        In general, a language support library, would have a file called 
        `Makefile.rules`. It will be automatically loaded by the build system 
        and you can take your compile rules for D sources there.
        
        Sorry for this naive question (I am obviously not an D expert): Can the 
        D compiler be used to compile our C sources (e.g., by using the betterC 
        subset)? In such a case, we could provide a D compiler option with a 
        mechanism that is currently suggested for supporting clang compiler:
        https://patchwork.unikraft.org/project/unikraft/list/?series=1584&state=*
        
        In case that this could be even completely done without needing any 
        runtime library, then we should have enough justifaction to integrate D 
        support into the core repository.
        
        Let me know what you think.
        
        Thanks,
        
        Simon
     "

CristiBM:
At least from what I know, the situation looks something like this:
There is the option of using the gdc D compiler, which does not have an explicit support for a subset of the language that would work without the runtime (although it does allow for not including the standard or runtime libraries in the build, but making the build actually work is another story); since this is a gcc frontend, you can actually compile C sources with it, but indeed, i believe it would not be of much help if it needs the runtime.
On the other hand, there is dmd -betterC, which doesn't need the runtime, but it doesn't seem to recognize C sources.
The resulting object files (from both gdc and dmd -betterC compilation of D sources) can, without problems, be linked by gcc to other object files resulted from C sources - something that already happens with this patch series.
The motivation of not wanting the runtime as a dependency for the core is clear :) but, from what I understand, it is also not desirable to add compilation support to the core even if the runtime is not needed, right? (more of a curiosity, does this also mean that core components could not be re-written in D?). If this is the case, I think would be most useful to look a little bit more into how gdc behaves when removing the runtime (I didn't manage to get anywhere with this, but I believe Cosmin Ularu is looking into it) and if not, a language support library sounds good enough :)

vladandrew:
Hi Cristi, I think that we should keep D support in a separate repo, like we did for Go. Basically, we'll have a repo with the runtime and the build rules. This is because D cannot easily be integrated with the rest of the Unikernel to write internal libraries.

skuenzer:
> At least from what I know, the situation looks something like this:
> There is the option of using the gdc D compiler, which does not have an explicit support for a subset of the language that would work without the runtime (although it does allow for not including the standard or runtime libraries in the build, but making the build actually work is another story); since this is a gcc frontend, you can actually compile C sources with it, but indeed, i believe it would not be of much help if it needs the runtime.
> On the other hand, there is dmd -betterC, which doesn't need the runtime, but it doesn't seem to recognize C sources.
> The resulting object files (from both gdc and dmd -betterC compilation of D sources) can, without problems, be linked by gcc to other object files resulted from C sources - something that already happens with this patch series.
> The motivation of not wanting the runtime as a dependency for the core is clear :) but, from what I understand, it is also not desirable to add compilation support to the core even if the runtime is not needed, right? (more of a curiosity, does this also mean that core components could not be re-written in D?). If this is the case, I think would be most useful to look a little bit more into how gdc behaves when removing the runtime (I didn't manage to get anywhere with this, but I believe Cosmin Ularu is looking into it) and if not, a language support library sounds good enough :)

Hum... the only working mode and reason why to put it into the core repo is to write libraries with D code that are independent of any D runtime (like we can do with embedded rust). If this is possible and if this mode is not creating any conflicting compile rules for libraries in D that depend on the language environment, then we have a reason why to put it into the core.
Related to this you can have a look to PR #241.

vladandrew:
@CristiBM any update on this?

razvand:
@Alex-deVis and myself have been working on this to update it for the newer version of Unikraft. We suspended it due to @Alex-deVis 's requirement to work on his thesis project.

gabrielmocanu:
@razvand @Alex-deVis  Any updated on this?

Alex-deVis:
Hi, there are no updates. I will attempt to update the PR or at least provide a detailed list with my findings.

razvand:
This is superseded by #783

PR #257, opened 2021-07-14
plat/kvm/x86: Fix lost IRQs

marcrittinghaus:
There is a race condition in `ukplat_lcpu_halt_irq()`, where an IRQ
may fire in between the short window after which interrupts are
enabled and before the halt is done. This may halt the caller
forever (or much longer until the next interrupt).

As this issue is platform and architecture specific, this commit
moves `ukplat_lcpu_halt_irq()` from the common code to the
plat/arch-specific implementation. For x86 the race condition
is avoided by ensuring that there are no instructions between
`STI` and `HLT`.

The PR does not provide fixes for the other platforms or
architectures other than KVM/x86_64.

### DIFF
--- a/plat/common/include/x86/irq.h
+++ b/plat/common/include/x86/irq.h
@@ -49,6 +49,11 @@
 	asm volatile("sti" : : : "memory"); \
 })
 
+#define __sti_hlt() \
+({ \
+	asm volatile("sti; hlt" : : : "memory"); \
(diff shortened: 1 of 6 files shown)


skuenzer (approved):
Thanks! Really good finding. We should double-check that this also properly handled on Xen and LinuxU.

PR #258, opened 2021-07-29
build: Move GO build rules in the libgo repo

vladandrew:
Since we cannot integrate GO sources without the libgo external library,
we move the GO build rules outside the main repo, in the libgo library.


skuenzer (approved):
Thanks a lot for this move.

PR #262, opened 2021-07-30
support/scripts: Helper script to create CPIO images

skuenzer:
Adds a helper script for creating CPIO images under Linux that can be loaded by `lib/ukcpio`. `lib/ukcpio` currently supports only the `newc` format from BSD and requires that folders entries are listed before file entries. The script can be find under `support/scripts/mkcpio`.

### DIFF
--- a/support/scripts/mkcpio
+++ b/support/scripts/mkcpio
@@ -1,0 +1,36 @@
+#!/bin/bash
+CPIOTOOL="$( which bsdcpio )"
+
+usage()
+{
+	echo "$0 [output.cpio] [CPIO root path]" 1>&2
+	exit 1
+}
+
+[ -z "$1" -o -z "$2" ] && usage
+if [ -e "$1" -a ! -f "$1" ]; then
+	echo "Output '$1' does already exist and is not a file" 1>&2
+	usage
+fi
+if [ ! -d "$2" ]; then
+	echo "'$2' is not a directory" 1>&2
+	usage
+fi
+
+OUT_BASE="$( readlink -f "$( dirname "$1" )" )"
+OUT_FILE="$( basename "$1" )"
+
+if [ ! -d "${OUT_BASE}" ]; then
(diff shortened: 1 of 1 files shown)


marcrittinghaus (commented):
Hey Simon, very useful script 😃 I have a similar one but without all the whistles and bells 😏 

I noticed that the `-print0` replaces the `\n` with `\0` so you can have files listed that have newlines in their name. Unfortunately, `tac` uses `\n` as separator so it won't reverse the output anymore, which breaks the mentioned requirement to list directories before files. I admit to not having actually tested the script, but if it works I would be confused. Or did I miss something here?

hlef:
This is really useful IMO, we should have this in the release. If this works, let's merge it :)

skuenzer:
I have no explanation how this happened but `-print0` does in fact not work as intended because of `tac`. I pushed an update where just `-print` is used. Sorry. Thanks for spotting the issue.

marcrittinghaus:
Does it make sense to have the `-L` option for `find` and the cpio tool to follow symlinks or do we have support for symlinks in ramfs?

skuenzer:
> Does it make sense to have the `-L` option for `find` and the cpio tool to follow symlinks or do we have support for symlinks in ramfs?

No, even if `ramfs` may not support symlinks (I would need to check, actually), I would keep symlinks as symlinks. As user, I would not assume that symlinks would get replaced and would prefer rather mounting fails in the image than the script replacing them. A user could still prepare the source folder not to have symlinks in such a case. I find this behavior more transparent. What do you think?

hlef:
Agree with Simon. That's the behavior I'd expect.

marcrittinghaus (approved):

hlef:
Approved-by: Hugo Lefeuvre <hugo.lefeuvre@manchester.ac.uk>

hlef:
Thanks Marc and Simon :)

PR #279, opened 2021-09-03
lib/ukalloc: Register allocator statistics to `ukstore`

craciunoiuc:
This adds the allocator statistics to `ukstore` as an example for the USoC21 hackathon.
The implementation is just for showcase purposes and is not final/ready for `staging`.

Please merge the `ukstore` PR first.

### DIFF
--- a/lib/ukalloc/stats.c
+++ b/lib/ukalloc/stats.c
@@ -31,6 +31,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <uk/store.h>
 #include <uk/alloc_impl.h>
 
 #if CONFIG_LIBUKALLOC_IFSTATS_GLOBAL
@@ -48,6 +49,13 @@ void uk_alloc_stats_get(struct uk_alloc *a,
 	uk_preempt_enable();
 }
 
+static int get_cur_mem_free(void *cookie __unused, __u64 *out)
+{
+	*out = (__u64) uk_alloc_availmem_total();
+	return 0;
+}
+UK_STORE_STATIC_ENTRY(cur_mem_free, u64, get_cur_mem_free, NULL, NULL);
+
 #if CONFIG_LIBUKALLOC_IFSTATS_GLOBAL
 void uk_alloc_stats_get_global(struct uk_alloc_stats *dst)
 {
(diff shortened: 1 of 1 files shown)


albisorua (approved):
Looks good. Tested.

Signed-off-by: Andrei Albisoru andreialbisoru98@gmail.com

nderjung:
Depends-on: #202

skuenzer (changes requested):
Hi @craciunoiuc ,

thanks for your work. I think we should do it slightly different: `ukalloc` already provides the interface `uk_alloc_stats_get_global()` globally, I see less of an advantage in forwarding the struct also through `ukstore`. Users would need to include ukalloc headers to get the struct defintion and then they actually could just use the `ukalloc`API directly. I would also avoid doing any allocations in the entry callbacks, if possible.
Overall, I think that it is probably more meaningful if we export some individual and selected numbers from the stats directly, like:
```c
static int get_mem_free(void *cookie __unused, __u64 *out)
{
   *out = (u64) uk_alloc_availmem_total(void);
}
UK_STORE_STATIC_ENTRY(mem_free, u64, get_mem_free, NULL);

#if CONFIG_LIBUKALLOC_IFSTATS_GLOBAL
static int get_mem_used(void *cookie __unused, __u64 *out)
{
   struct uk_alloc_stats stats;

   uk_alloc_stats_get_global(&stats);
   *out = (u64) cur_mem_use;
   return 0;
}
UK_STORE_STATIC_ENTRY(mem_used, u64, get_mem_used, NULL);

static int get_nb_allocs(void *cookie __unused, __u64 *out)
{
   struct uk_alloc_stats stats;

   uk_alloc_stats_get_global(&stats);
   *out = (u64) cur_nb_allocs;
   return 0;
}
UK_STORE_STATIC_ENTRY(nb_allocs, u64, get_nb_allocs, NULL);
#endif /* CONFIG_LIBUKALLOC_IFSTATS_GLOBAL */
```

craciunoiuc:
Hey, yes, this is what I wanted to do initially. This PR was more like a Draft until now, I kind of forgot about it too 😓

I will do an ukstore entry for each stat

skuenzer (approved):
Thanks a lot for the changes and work! This looks good.

PR #287, opened 2021-09-15
arch/x86_64: Fix TLS alignment issue

cffs:
According to the x86_64 ELF ABI, the thread pointer in FS register should be aligned according to the alignment of the TLS program header in the ELF file. As this pointer should be just after the TLS area, the TLS area itself has to be properly aligned and padded.

### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Change the TLS area setup so that the thread pointer always has sufficient (i.e. 32-byte) alignment. This amounts to increase alignment and padding of the TLS area. Additionally, we ensure there is always at least one variable in `.tbss` section, to work around a ld limitation (ld ignores the `.tbss` section if it contains only padding).

### DIFF
--- a/arch/x86/x86_64/include/uk/asm/tls.h
+++ b/arch/x86/x86_64/include/uk/asm/tls.h
@@ -50,7 +50,10 @@ static inline __sz ukarch_tls_area_size(void)
 
 static inline __sz ukarch_tls_area_align(void)
 {
-	return 8;
+	/* The minimal required TLS area alignment should ideally come from the
+	 * ELF header, but 32 should be fine for all use X86_64 instructions.
+	 */
+	return 32;
 }
 
 static inline void ukarch_tls_area_copy(void *tls_area)
--- a/arch/x86/x86_64/Makefile.uk
+++ b/arch/x86/x86_64/Makefile.uk
@@ -45,3 +45,6 @@ ARCHFLAGS-$(CONFIG_MARCH_X86_64_BTVER1)         += -march=btver1
 ISR_ARCHFLAGS-$(CONFIG_MARCH_X86_64_BTVER1)     += -march=btver1
 ARCHFLAGS-$(CONFIG_MARCH_X86_64_BTVER2)         += -march=btver2
 ISR_ARCHFLAGS-$(CONFIG_MARCH_X86_64_BTVER2)     += -march=btver2
+
+$(eval $(call addlib,libx86_64arch))
+LIBX86_64ARCH_SRCS-y += $(CONFIG_UK_BASE)/arch/x86/x86_64/tls.c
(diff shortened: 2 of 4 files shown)


arch/x86/x86_64/include/uk/asm/tls.h
```diff
@@ -67,5 +81,8 @@ static inline void ukarch_tls_area_copy(void *tls_area)
 
 static inline void *ukarch_tls_pointer(void *tls_area)
 {
-	return tls_area + (_tls_end - _tls_start);
+	__sz tls_len = _tls_end - _tls_start;
+	__sz padding = tls_len % ukarch_tls_area_align();
+
+	return tls_area + padding + tls_len;
```
mogasergiu: I believe it should have been `return tls_area + ukarch_tls_area_size();`. Otherwise `fs:offset` seems to pick wrong values.

cffs: `ukarch_tls_area_size()` returns the full TLS area size, including **8 bytes for the TLS pointer itself** appended at the end of it. It does seem to me the TLS pointer address should be `tls_area + ukarch_tls_area_size() - 8`, which is the same as `tls_area + padding + tls_len`. It was already the case before my changes, which amounted to add the padding both to `ukarch_tls_area_size()` and to `ukarch_tls_pointer()`.
Can you provide a [MWE](https://www.wikiwand.com/en/Minimal_working_example) that triggers wrong values for `fs::offset`, so I can have a look? It seems to work on my setup.
That being said, I think returning `tls_area + ukarch_tls_area_size() - 8` would be clearer and less redundant than what is currently present. I'll update the PR with that version.

cffs: I found some issue on my own, I will investigate and come back to you.

mogasergiu: I have provided [here](https://github.com/cs-pub-ro/app-helloworld/tree/TLS-alignment) the example which fails on my end. With this commit, the output is `0x0.0x0.0x0`, instead of `0x22.0x4.0x4`.

cffs: There is some problem unrelated to this patch, that also occurs with current staging. I created a new issue: #290. Let's revisit the alignment issue once #290 is fixed.

cffs:
I found a simpler way to fix the alignment issue than in the original commit, by asking for more alignment in the linker script rather than doing manual padding in the code. However, there is still issue #290 when the `.tbss` section should be missing, but is not.

mogasergiu (approved):
Everything looks fine now.

cffs:
Rather than adding a thread-local variable in `uksched` (which calls TLS initialisation functions), I added it in the x86_64 architecture code, since it should only be necessary on that architecture.

vladandrew:
Approved-by: Vlad-Andrei Badoiu <vlad_andrei.badoiu@upb.ro>

PR #289, opened 2021-09-16
Provide spinlock implementation for arm64 and x86_64

rene:
## Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): kvm
 - Application(s): N/A

### Additional configuration

CONFIG_UKPLAT_LCPU_MULTICORE will need to be set when SMP become available.

### Description of changes

This pull request provide a patch series that provide spinlock implementation for ARM64. The following functions are implemented:

void ukarch_spin_lock_init(spinlock_t *spinlock);
void ukarch_spin_lock(spinlock_t *lock);
void ukarch_spin_unlock(spinlock_t *lock);
int ukarch_spin_trylock(spinlock_t *lock);
int ukarch_spin_is_locked(spinlock_t *lock);

FYI: @razvand @skuenzer @marcrittinghaus

### DIFF
--- a/arch/arm/arm64/include/uk/asm/spinlock.h
+++ b/arch/arm/arm64/include/uk/asm/spinlock.h
@@ -1,0 +1,93 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2013-2019, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2021 OpenSynergy GmbH. All rights reserved.
+ * Copyright (c) 2021 Karlsruhe Institute of Technology (KIT).
+ *               All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKARCH_SPINLOCK_H__
+#error Do not include this header directly
+#endif
+
+#include <uk/arch/atomic.h>
+
+struct __spinlock {
+	volatile int lock;
+};
+
+/* Initialize a spinlock to unlocked state */
+#define UKARCH_SPINLOCK_INITIALIZER() { 0 }
+
+static inline void ukarch_spin_init(struct __spinlock *lock)
+{
+	lock->lock = 0;
+}
+
+static inline void ukarch_spin_lock(struct __spinlock *lock)
+{
+	register int r, locked = 1;
+
+	__asm__ __volatile__(
+		"	sevl\n"			/* set event locally */
+		"1:	wfe\n"			/* wait for event */
+		"2:	ldaxr	%w0, [%1]\n"	/* exclusive load lock value */
+		"	cbnz	%w0, 1b\n"	/* check if already locked */
+		"	stxr	%w0, %2, [%1]\n"/* try to lock it */
+		"	cbnz	%w0, 2b\n"	/* jump to l2 if we failed */
+		: "=&r" (r)
+		: "r" (&lock->lock), "r" (locked));
+}
+
+static inline void ukarch_spin_unlock(struct __spinlock *lock)
+{
+	__asm__ __volatile__(
+		"stlr	wzr, [%0]\n"		/* unlock lock */
+		"sev\n"				/* wake up any waiters */
+		:
+		: "r" (&lock->lock));
+}
+
+static inline int ukarch_spin_trylock(struct __spinlock *lock)
+{
+	register int r, locked = 1;
+
+	__asm__ __volatile__(
+		"	ldaxr	%w0, [%1]\n"	/* exclusive load lock value */
+		"	cbnz	%w0, 1f\n"	/* bail out if locked */
+		"	stxr	%w0, %2, [%1]\n"/* try to lock it */
+		"1:\n"
+		: "=&r" (r)
+		: "r" (&lock->lock), "r" (locked));
+
+	return !r;
+}
+
+static inline int ukarch_spin_is_locked(struct __spinlock *lock)
+{
+	return UK_READ_ONCE(lock->lock);
+}
--- a/arch/x86/x86_64/include/uk/asm/spinlock.h
+++ b/arch/x86/x86_64/include/uk/asm/spinlock.h
@@ -1,0 +1,92 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2021 Karlsruhe Institute of Technology (KIT).
+ *               All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKARCH_SPINLOCK_H__
+#error Do not include this header directly
+#endif
+
+#include <uk/arch/atomic.h>
+
+struct __spinlock {
+	volatile int lock;
+};
+
+/* Initialize a spinlock to unlocked state */
+#define UKARCH_SPINLOCK_INITIALIZER() { 0 }
+
+static inline void ukarch_spin_init(struct __spinlock *lock)
+{
+	lock->lock = 0;
+}
+
+static inline void ukarch_spin_lock(struct __spinlock *lock)
+{
+	register int locked = 1;
+
+	__asm__ __volatile__(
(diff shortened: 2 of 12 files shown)


marcrittinghaus:
Hi Rene,

as discussed in our call, I supply the following patches for your PR. Please apply them and we are continuing discussion from there, if you agree with the changes. Each patch should be a separate commit.

To summarize, I have done the following changes:
+ I moved the actual spinlock implementation into header files and made the functions `static inline`. I also optimized the code a bit by letting the compiler decide which registers to use in the ASM snippets. This makes the snippets fit much better into the surrounding code when inlining. I added barriers where needed.
+ I moved the `__spinlock` structure to the architecture header allowing for architecture-dependent extensions (e.g., using a different word length than `int`)
+ I replaced the empty spinlocks (i.e., with disabled `CONFIG_UKPLAT_LCPU_MULTICORE`) so that they still use barriers where appropriate and code in critical sections is not mixed up with code outside. So as a developer we can expect spinlocks to have always the same effect. I also changed the empty handling for `ukarch_spin_trylock` and `ukarch_spin_is_locked` so they can still be transparently used in the code. Statements like `if (ukarch_spin_is_locked(..)) {}` do not compile for `(void)(lock)`.
+ I fixed existing uses of `spinlock_t` to use `__spinlock` instead.
+ I added an x86 spinlock
+ I added the `uk_spinlock` alias definitions so you can write `struct uk_spinlock` and use `uk_spin_lock()`, which are just `#define`s for the ukarch variants.

[spinlock_patches.zip](https://github.com/unikraft/unikraft/files/7211673/spinlock_patches.zip)

rene:
> Hi Rene,
> 
> as discussed in our call, I supply the following patches for your PR. Please apply them and we are continuing discussion from there, if you agree with the changes. Each patch should be a separate commit.
> 
> To summarize, I have done the following changes:
> 
>     * I moved the actual spinlock implementation into header files and made the functions `static inline`. I also optimized the code a bit by letting the compiler decide which registers to use in the ASM snippets. This makes the snippets fit much better into the surrounding code when inlining. I added barriers where needed.
> 
>     * I moved the `__spinlock` structure to the architecture header allowing for architecture-dependent extensions (e.g., using a different word length than `int`)
> 
>     * I replaced the empty spinlocks (i.e., with disabled `CONFIG_UKPLAT_LCPU_MULTICORE`) so that they still use barriers where appropriate and code in critical sections is not mixed up with code outside. So as a developer we can expect spinlocks to have always the same effect. I also changed the empty handling for `ukarch_spin_trylock` and `ukarch_spin_is_locked` so they can still be transparently used in the code. Statements like `if (ukarch_spin_is_locked(..)) {}` do not compile for `(void)(lock)`.
> 
>     * I fixed existing uses of `spinlock_t` to use `__spinlock` instead.
> 
>     * I added an x86 spinlock
> 
>     * I added the `uk_spinlock` alias definitions so you can write `struct uk_spinlock` and use `uk_spin_lock()`, which are just `#define`s for the ukarch variants.
> 
> 
> [spinlock_patches.zip](https://github.com/unikraft/unikraft/files/7211673/spinlock_patches.zip)

Thanks @marcrittinghaus , I'm looking into it.....

rene:
Fix commit messages.

marcrittinghaus:
Hi @rene 

the build system has found that the change from `ukarch_spin_lock_init()` to `ukarch_spin_init()` is missing in my patches for 9pfs and the xenbus driver. Could you update the relevant code locations as part of the respective commits?

image

There are also two `DEFINE_SPINLOCK`s used in 9p.

rene:
> Hi @rene
> 
> the build system has found that the change from `ukarch_spin_lock_init()` to `ukarch_spin_init()` is missing in my patches for 9pfs and the xenbus driver. Could you update the relevant code locations as part of the respective commits?
> 
> image
> 
> There are also two `DEFINE_SPINLOCK`s used in 9p.

@marcrittinghaus , it's done.

marcrittinghaus:
@rene 
You forgot to change some of the references. Please see the attached patch file for the missing changes. Sorry for having missed that in the first place.

[spinlock.zip](https://github.com/unikraft/unikraft/files/7467251/spinlock.zip)

rene:
> @rene You forgot to change some of the references. Please see the attached patch file for the missing changes. Sorry for having missed that in the first place.
> 
> [spinlock.zip](https://github.com/unikraft/unikraft/files/7467251/spinlock.zip)

no worries, I've updated the PR. Please, check if it's building at your side now....

include/uk/arch/spinlock.h
```diff
@@ -30,22 +29,30 @@
...
+/* N.B.: For singlecore systems we remove spinlocks by mapping functions to
+ * void operations. No state is saved for the spinlock but we assume that the
+ * CPU is always able to get the lock. Trying to acquire a spinlock thus always
+ * succeeds and asserting that a lock is held returns false.
+ */
+typedef struct __spinlock {
+	/* empty */
+} __spinlock;
+
+#define UKARCH_SPINLOCK_INITIALIZER()
```
michpappas: This won't compile when CONFIG_HAVE_SMP is not enabled:
```
spinlock_t mylock = UK_SPINLOCK_INITIALIZER();
```

michpappas:
The public interface is not documented, which makes it unclear on how to use the `UK_SPINLOCK_INITIALIZER` macro, or what are the return values of `is_locked()` and `trylock()`. There are two different parts, one is the abstracted interface in `lib/uklock/include/uk/spinlock.h` and then there is the interface in `include/uk/arch/spinlock.h` that can be included directly.

While we're there, I found it also a bit unclear that in the relevant menuconfig item, while you get options for mutexes and semaphores, spinlocks are not mentioned anywhere. It's not  a big deal, given that when we introduce ticketlocks (I have a PR for Arm64 on the way) there will be a menu item for choosing between the two. So I guess it's okay to leave it for now as is.

arch/x86/x86_64/include/uk/asm/spinlock.h
```diff
@@ -0,0 +1,116 @@
...
+		"	test	%%eax, %%eax\n"	/* bail out if locked */
+		"	jnz	1f\n"
+		"	lock; cmpxchg %2, (%1)\n" /* try to acquire spinlock */
+		"	cmove	%2, %0\n"	/* store if successful */
+		"1:\n"
+		: "+&r" (r)
+		: "r" (&lock->lock), "r" (locked)
+		: "eax");
+
+	return r;
```
michpappas, line 105: This actually returns zero on successful acquisition, but the documentation states otherwise. Btw, I found this interface inconsistent with `is_locked()`. Maybe we should leave the return value as is and change the interface instead?

arch/arm/arm64/include/uk/asm/spinlock.h
```diff
@@ -0,0 +1,117 @@
...
+
+	__asm__ __volatile__(
+		"	ldaxr	%w0, [%1]\n"	/* exclusive load lock value */
+		"	cbnz	%w0, 1f\n"	/* bail out if locked */
+		"	stxr	%w0, %2, [%1]\n"/* try to lock it */
+		"1:\n"
+		: "=&r" (r)
+		: "r" (&lock->lock), "r" (locked));
+
+	return r;
```
michpappas: This actually returns zero on successful acquisition, but the documentation states otherwise. ~~Btw, I found this interface inconsistent with is_locked(). Maybe we should leave the return value as is and change the interface instead?~~ My bad, the interface is fine, it's only the return value that needs to be fixed here.

include/uk/arch/spinlock.h
marcrittinghaus:
```suggestion
#define UKARCH_SPINLOCK_INITIALIZER()	{}
```

yep, good catch!

arch/arm/arm64/include/uk/asm/spinlock.h
marcrittinghaus:
```suggestion
	return !r;
```

I missed this one 👍🏻 Very good that you are testing everything!

marcrittinghaus:
> The public interface is not documented, which makes it unclear on how to use the `UK_SPINLOCK_INITIALIZER` macro, or what are the return values of `is_locked()` and `trylock()`. There are two different parts, one is the abstracted interface in `lib/uklock/include/uk/spinlock.h` and then there is the interface in `include/uk/arch/spinlock.h` that can be included directly.

Yes, this is a good point. The question is how to solve it. We could put declarations into `arch/spinlock.h` like so:
image
And perhaps remove the redundant documentation from the actual implementations in `asm/spinlock.h`. However this is not easily done for the macro. What do you think?

We could repeat the doc in `uk_spinlock`.

> While we're there, I found it also a bit unclear that in the relevant menuconfig item, while you get options for mutexes and semaphores, spinlocks are not mentioned anywhere. It's not a big deal, given that when we introduce ticketlocks (I have a PR for Arm64 on the way) there will be a menu item for choosing between the two. So I guess it's okay to leave it for now as is.

yep, sounds good 👍🏻

michpappas:
Yes, all I had in mind was actually repeating the doc in both public headers, sorry I didn't make that clear. Or copy it on one, and leave a comment with a reference on the other. This also formalizes the interface that needs to be used by other architectures or algorithms.

marcrittinghaus:
> Yes, all I had in mind was actually repeating the doc in both public headers, sorry I didn't make that clear. Or copy it on one, and leave a comment with a reference on the other. This also formalizes the interface that needs to be used by other architectures or algorithms.

I would prefer to not have a copy of the doc at multiple locations, just to avoid inconsistencies when updating it. I have prepared a patch that
- puts the documentation and function declarations into `uk/arch/spinlock.h`
- removes the redundant documentation from the implementation headers
- adds a note to `uk/spinlock.h` where to find the documentation
- fixes the two errors

[spinlock_fix_and_doc.zip](https://github.com/unikraft/unikraft/files/7475469/spinlock_fix_and_doc.zip)

michpappas:
Hi @marcrittinghaus, thanks for that. I notice some additional changes than the ones described above:
* Replace the opaque `__spinlock` type with `struct __spinlock`
* Introduce a lock in `weak_main.c`
* Introduce a hard-coded `#define CONFIG_SMP` in `include/uk/arch/spinlock.h`

The last two look like accidental inclusions of test code. How about the first one? Is it intentional?

Also, would it be preferable to have the docs in the more abstract `lib/uklock/include/uk/spinlock.h`, so that it covers other variants such as ticket locks too? Otherwise, once we introduce `include/uk/arch/ticketlock.h` we would have to add a link to refer to the documentation of `include/uk/arch/spinlock.h` (or duplicate which is what we wanted to avoid in the first place). Anyway, either way it's not a big deal.

marcrittinghaus:
> Hi @marcrittinghaus, thanks for that. I notice some additional changes than the ones described above:
> 
> * Replace the opaque `__spinlock` type with `struct __spinlock`
> * Introduce a lock in `weak_main.c`
> * Introduce a hard-coded `#define CONFIG_SMP` in `include/uk/arch/spinlock.h`
> 
> The last two look like accidental inclusions of test code. How about the first one? Is it intentional?
> 
> Also, would it be preferable to have the docs in the more abstract `lib/uklock/include/uk/spinlock.h`, so that it covers other variants such as ticket locks too? Otherwise, once we introduce `include/uk/arch/ticketlock.h` we would have to add a link to refer to the documentation of `include/uk/arch/spinlock.h` (or duplicate which is what we wanted to avoid in the first place). Anyway, either way it's not a big deal.

Ah, yes. Of course, the last two points are from debugging. I removed these parts but must have forgotten to save the files. With the github plugin in VSCode this happens easily, because you can revert changes in the GUI via GIT but it only changes the files and you still have to explicitly save them 🙄  

Regarding the first point, I thought that it has to be a struct anyway. Otherwise, you cannot write `struct uk_spinlock`, which I would like to have for coherence with the other synchronization primitives from `uklock`. To somewhat enforce this and since we liked the spinlock to be more generically documented, I thought having a "forward" declaration in the publicly documented part would be a good idea. I also think that restricting `__spinlock` to be a `struct` of some sort is not a limitation in anyway.

My plan for the `ticketlock` was to have the implementation entirely in `lib/uklock` (not ukarch) and provide a `CONFIG_` option to decide the implementation of `struct uk_spinlock` as either `__spinlock` (ukarch) or `uk_ticketlock` (uklock), with the API obviously being the same. This way, you could use `uk_ticketlock` wherever explicitly desired and also switch the default implementation for `uk_spinlock`. Then, for cases where you do not want any more complex spinlock types (perhaps some IRQ/exception handler) or cannot efford a dependency on `uklock`, you can always use `__spinlock` (as a built-in type).

So from my perspective, the API to which all spinlocks have to adhere to is given by `__spinlock`. This is why I put the documentation there and a reference in `uklock`.

michpappas:
> Regarding the first point, I thought that it has to be a struct anyway. Otherwise, you cannot write struct uk_spinlock, which I would like to have for coherence with the other synchronization primitives from uk_lock. To somewhat enforce this and since we liked the spinlock to be more generically documented, I thought having a "forward" declaration in the publicly documented part would be a good idea.

By all means.

> My plan for the ticketlock was to have the implementation entirely in lib/uklock (not ukarch) and provide a CONFIG_ option to decide the implementation of struct uk_spinlock as either __spinlock (ukarch) or uk_ticketlock (uklock), with the API obviously being the same. This way, you could use uk_ticketlock wherever explicitly desired and also switch the default implementation for uk_spinlock. Then, for cases where you do not want any more complex spinlock types (perhaps some IRQ/exception handler) or cannot efford a dependency on uklock, you can always use __spinlock (as a built-in type).
>
> So from my perspective, the API to which all spinlocks have to adhere to is given by __spinlock. This is why I put the documentation there and a reference in uklock.

Okay, I see. If ticket locks go in uklock then it makes sense to keep the docs where they are, thanks.

rene:
@marcrittinghaus , @michpappas I've pushed the changes except by the debug code left on Marc's patch.

michpappas:
Hi @rene, it all looks good to me, except from one small thing in one of the commit messages:

> Fix spinlock declarations on device drivers according new definition
> introduced by commit 7bcc2f1 ("include/uk/arch/spinlock.h: Rewrite
> spinlock header").

The commit hash is no longer valid, as it refers to a previous version of this patch set. Since it refers to a commit that has not been merged yet, even if you fix it now, it may still change at merge time due to a rebase, so I would probably avoid referring by hash anything that is not already merged.

marcrittinghaus:
> The commit hash is no longer valid, as it refers to a previous version of this patch set. Since it refers to a commit that has not been merged yet, even if you fix it now, it may still change at merge time due to a rebase, so I would probably avoid referring by hash anything that is not already merged.

Right. In fact, the hashes will always change because the commit messages are rewritten to add the git trailers (e.g., Reviewed-by etc.)

rene:
> > The commit hash is no longer valid, as it refers to a previous version of this patch set. Since it refers to a commit that has not been merged yet, even if you fix it now, it may still change at merge time due to a rebase, so I would probably avoid referring by hash anything that is not already merged.
> 
> Right. In fact, the hashes will always change because the commit messages are rewritten to add the git trailers (e.g., Reviewed-by etc.)

ok, so there is no reason to let commit hash in this case. @michpappas , @marcrittinghaus I've updated the commit message.

michpappas:
@rene thanks :) From my side it's good now.

marcrittinghaus:
Hi Rene, Hi Michalis, 

thank you very much! 🎉  

@michpappas Have you actually approved the changes (i.e., start review -> approve) or just left a comment? Also please ensure that the git trailer is just plain text (i.e., no `Reviewed-by` but just Reviewed-by). 

BTW: For the SMP API (#309) please wait with the final approval. @nderjung is changing the approval process so that a reviewer and a maintainer have approved the changes and I like that PR to have both our names on it, so it is clear that I was involved (and can be made responsible 😄)

michpappas (approved):

michpappas:
> @michpappas Have you actually approved the changes (i.e., start review -> approve) or just left a comment? Also please ensure that the git trailer is just plain text (i.e., no `Reviewed-by` but just Reviewed-by).

I did now :) Earlier I used backticks for easy copy-paste (github provides a copy widget on the right on these), as in some projects RBs are applied manually. Anyway, will be using github's review process from now on.
 
> BTW: For the SMP API (#309) please wait with the final approval. @nderjung is changing the approval process so that a reviewer and a maintainer have approved the changes and I like that PR to have both our names on it, so it is clear that I was involved (and can be made responsible smile)

Okay good to know :smile: I expected that more people can add their RBs, so you would be able to have yours once you approved.

PR #291, opened 2021-09-23
Modify buildsystem to integrate cargo

TrueDoctor:
### Prerequisite checklist

### Base target

 - Architecture(s): [all]
 - Platform(s): [all]

### Additional configuration

cargo needs to be available on the target system to take advantage of the changes

### DIFF
--- a/Makefile.uk
+++ b/Makefile.uk
@@ -29,7 +29,7 @@ RUSTCFLAGS-y	+= --emit=obj --crate-type=rlib --edition=2018 \
 
 RUSTCFLAGS-$(CONFIG_OPTIMIZE_NONE)         += -Copt-level="0"
 RUSTCFLAGS-$(CONFIG_OPTIMIZE_SIZE)         += -Copt-level="s"
-RUSTCFLAGS-$(CONFIG_OPTIMIZE_PERF)         += -Copt-level="2"
+RUSTCFLAGS-$(CONFIG_OPTIMIZE_PERF)         += -Copt-level="3"
 
 RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL0)     += -Cdebuginfo=0
 RUSTCFLAGS-$(CONFIG_DEBUG_SYMBOLS_LVL1)     += -Cdebuginfo=1
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -72,6 +72,10 @@ fileext = $(subst .,,$(suffix $(1)))
 # libname2olib $libname
 libname2olib = $(addprefix $(BUILD_DIR)/,$(addsuffix .o,$(1)))
 
+# converts a list of library names to paths pointing to their corresponding static library file
+# libname2alib $libname
+libname2alib = $(addprefix $(BUILD_DIR)/,$(addsuffix .a,$(1)))
+
 libname2preolib = $(addprefix $(BUILD_DIR)/,$(addsuffix .ld.o,$(1)))
 
 # converts a list of sources to paths pointing to their corresponding destination files
@@ -88,11 +92,16 @@ endef
 define src2obj =
 $(call src2dst,$(1),$(2),.o,$(3))
 endef
+# cargo2alib $libname,$source(s),$variant(optional)
+# converts a list of .toml files to library out files
+define cargo2alib =
+$(call src2dst,$(1),$(1),.a,$(3))
+endef
 
 # converts a list of output file (object files or linker scripts) paths
 # to paths pointing to their corresponding dependency file
 # out2dep $objects
-out2dep = $(patsubst %.o,%.d,$(patsubst %.lds,%.lds.d,$(1)))
+out2dep = $(patsubst %.o,%.d,$(patsubst %.lds,%.lds.d,$(patsubst %.o,%.d,$(patsubst %.a,%.d,$(1)))))
 
 # converts a list of sources (.S, .c, .cc, .cpp) to paths pointing to their corresponding dependency file
 # src2dep $libname,$source(s),$variant(optional)
@@ -789,7 +798,9 @@ $(if $(filter %.lds.S,$(2)),\
 		$(eval target=$(call src2lds,$(1),$(2))),\
 	$(if $(filter %.dts,$(2)),\
 		$(eval target=$(call dts2dtb,$(1),$(2))),\
-		$(eval target=$(call src2obj,$(1),$(2),$(3)))))
+		$(if $(filter %Cargo.toml,$(2)),\
+			$(eval target=$(call cargo2alib,$(1),$(2),$(3))),\
+			$(eval target=$(call src2obj,$(1),$(2),$(3))))))
 $(eval $(call buildrule,$(1),$(2),$(3),$(target),))
 endef
 


Makefile.uk
```diff
@@ -21,16 +21,18 @@ CINCLUDES    += -I$(CONFIG_UK_BASE)/include
 CXXINCLUDES  += -I$(CONFIG_UK_BASE)/include
 GOCINCLUDES  += -I$(CONFIG_UK_BASE)/include
 
-RUSTCFLAGS-y	+= --emit=obj --crate-type=rlib --edition=2018 \
+RUSTCFLAGS-y	+= +nightly --emit=obj --crate-type=rlib --edition=2018 \
```
jschlumpp: The "+<toolchain>" syntax is rustup specific and IIRC does not work with a normal rust installation (from the package manager or a manual installation). It also requires installing the toolchain manually (The error message does not seem to give instructions how to do that).
As an alternative there are [rust-toolchain](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) files, which rustup will use to automatically install the required rust version.

support/build/Makefile.rules
```diff
@@ -592,6 +601,36 @@ $(eval $(call vprefix_lib,$(1),OBJS-y) += $(4))
 $(eval $(call vprefix_lib,$(1),CLEAN-y) += $(call build_clean,$(4)) $(call out2dep,$(4)))
 endef
 
+# Build rust projects via Cargo
```
vladandrew: Hi Dennis,

I think that that **only** the code from this file, `Makefile.rules` should be moved in a separate repo dedicated to this[1]. You'll create a file with the same name in this repo. You may look into this repo[1] for a similar approach. So we'll basically split this intwo two pull requests, one on the lib-rust repo and this one with the remaining build changes. What do you think?
[1] https://github.com/unikraft/lib-rust
[2] https://github.com/unikraft/lib-dafny

.gitignore
```diff
@@ -34,6 +34,11 @@ GRTAGS
 GSYMS
 GTAGS
 
+# Carog build dirs
+**/target
+TODO: Decide whether to include lock files
+**/Cargo.lock
```
vladandrew: Do we still this in the main repo if we moved some parts to the new repo?

Makefile.uk
TrueDoctor: Thanks didn't know that, will remove that then

vladandrew (changes requested):

Makefile.uk
cffs: Ideally, one should be able to keep using stable rustc for regular user-space applications, preferably without rustup. That being said, as we currently require the use of the musl target for those, rustup is by far the easiest way to install that toolchain (whether manually or automatically).

support/build/Makefile.rules
cffs: I agree that for now, it is more scalable if each language stores its building rules in its own `lib-lang` external repository. In the long run, it could change if we were to rewrite or add new core components written in rust, rather than C.
Note that if we put rust/cargo Makefile rules in `lib-rust`, other rust-related stuff (e.g. bindings to core unikraft libs) should be behind options, or in yet other repositories, so that one can use *only* the building rules for his user-space regular cargo-based application.

Makefile.uk
TrueDoctor: The rust embedded support for the build system requires nightly rust due to unstable compiler flags being used, that is why in initially put it there.
The cargo approach runs for the most part on stable and does not require nightly iirc

.gitignore
TrueDoctor: This is only really relavant when developing internal libs with rust, I can safely remove this, someone will notice if this becomes an issue again

support/build/Makefile.rules
TrueDoctor: How do I split the modifications to the Makefile rules into a separate file? (the hackathon was my first time working with make so I am quite the noob)

vladandrew: You simply create a `Makefile.rules` in the new repo and paste the code there, check the dafny example :D.

TrueDoctor: So I just copy paste the entire function to which I made changes and that will just shadow the existing implementation?

vladandrew: Actually, we should move only this part of the changes to Makefile.rules in the external repo lib-rust.
```suggestion
# NOTE: We are not using most of the flags such as COMPFLAGS due to incompatibilities between rustc and GCC.
			#/--out-dir=$(call sub_build_dir,$(1)) \
	#$(call libname2olib,$(1))
# buildrule_cargo $libname,$source,$variant,$target,$extraflags
# $$(RUSTCFLAGS) $$(RUSTCFLAGS-y) $$(RUSTCFLAGS_EXTRA)
define buildrule_toml =
$(4): $(2) | preprocess
	$(call build_cmd,CARGO,$(1),$(4),\
		UK_CINCLUDES="$$(CINCLUDES) $$(CINCLUDES-y) $$($(call vprefix_lib,$(1),CINCLUDES)) $$($(call vprefix_lib,$(1),CINCLUDES-y))" \
		$(CARGO) $$($(call vprefix_lib,$(1),RUSTCHANNEL)) rustc \
		 	$$(CARGOFLAGS) $$(CARGOFLAGS-y) $$(CARGOFLAGS_EXTRA) \
			--manifest-path=$(2) \
			$$($(call vprefix_lib,$(1),CARGOFLAGS)) $$($(call vprefix_lib,$(1),CARGOFLAGS-y)) \
			-- \
			$$($(call vprefix_lib,$(1),RUSTCFLAGS)) $$($(call vprefix_lib,$(1),RUSTCFLAGS-y)) \
			$$($(call vprefix_src,$(1),$(2),$(3),FLAGS)) $$($(call vprefix_src,$(1),$(2),$(3),FLAGS-y)) \
			--cfg '__LIBNAME__="$(1)"' $(if $(3),--cfg '__VARIANT__="$(3)"') \
			--emit link=$(4)
	)

UK_SRCS-y += $(2)
UK_DEPS-y += $(call out2dep,$(4))
UK_ALIBS-y += $(4)
$(eval $(call vprefix_lib,$(1),OBJS-y) += $(4))
$(eval $(call vprefix_lib,$(1),CLEAN-y) += $(call build_clean,$(4)) $(call out2dep,$(4)))
.PHONY: $(4)
endef
```

TrueDoctor: That is not Ideal either because the other changes depend on being able to use `Cargo.toml` as a source. 
I am not a make expert but I would assume things would break if your folder contains a `Cargo.toml` file but the build system does not know what to do with it.

TrueDoctor: But I guess we could live with non descriptive error messages but that is certainly not ideal

vladandrew: Let me recap to see that I understand correctly, if `lib-rust` is not selected, then we'd get a non-descriptive error for any library that has a `Config.toml`, right? The error that I'm seeing is: ` ... is not defined: Failed to install rule for toml`. I think that we could live with it that for now. Maybe in the future, we could come back to this and see how we can improve upon it.

.gitignore
vladandrew: Hi Dennis, actually this should remain, but could you move it in a separate commit?

Makefile
```diff
@@ -573,10 +576,10 @@ GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
 # We use rustc because the gcc frontend is experimental and missing features such
 # as borrowing checking
 ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
-RUSTC		:= rustc --target=$(CONFIG_LLVM_TARGET_ARCH)
-else
-RUSTC		:= rustc
+RUST_TARGET		:= --target=$(CONFIG_LLVM_TARGET_ARCH)
```
vladandrew: What happens when $(CONFIG_LLVM_TARGET_ARCH) is empty? What do you think about:
```
ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
RUST_TARGET		:= --target=$(CONFIG_LLVM_TARGET_ARCH)
else
RUST_TARGET
endif
```

Makefile.uk
```diff
@@ -27,10 +27,12 @@ RUSTCFLAGS-y	+= --emit=obj --crate-type=rlib --edition=2018 \
 		-Cforce-unwind-tables=n -Ccodegen-units=1 \
 		-Dunsafe_op_in_unsafe_fn -Drust_2018_idioms
 
+CARGOFLAGS-y	+= #--offline #-q
```
vladandrew: Not sure if this is a typo, but I think that # marks a comment.

vladandrew (changes requested):

razvand:
@TrueDoctor , @vladandrew , should this be closed, since cargo support will likely be moved to an outside library?

TrueDoctor:
> @TrueDoctor , @vladandrew , should this be closed, since cargo support will likely be moved to an outside library?

As discussed on discord this pr will partially be moved to an exteral library and replaced with a pr that just adds the necesseary modifications to the build system. IIrc @vladandrew it working on that together with the effort to support both cargo and rustc + unicraft build system for building rust apps

vladandrew (approved):
I think this is minimum number of changes to Unikraft core for the Cargo support. We can't move them to `lib-cargo` since we are modifying existing rules.

marcrittinghaus:
Let me quickly jump into this discussion as I only see it now. It would be optimal if we could keep language-specific extensions out of the generic build system, as you obviously agree with this minimal PR. 

I have gone a different way with integrating the build system of GO, which uses a single file as entry point and then automagically finds out all other sources. I have extended the build system only in that it allows libraries to define own `Makefile.build` files (https://github.com/marcrittinghaus/unikraft/tree/mritting/makefile_build).

The library then defines all necessary files. The only missing piece at the moment is cleanup. You may want to have a look (https://github.com/marcrittinghaus/lib-libgo).

You basically define a go library by using `addgolib` instead of `addlib`. Everything else works as normal. In the background `addgolib` will add the library to an own list of libs that are processed differently by the custom `Makefile.build`. They are linked to the rest of the build simply by adding the object files.

vladandrew:
@razvand we can close this pull request. It is superseded by #680

razvand:
Roger, @vladandrew , I'm closing it.

PR #292, opened 2021-09-23
Add unikraft rust binding generation

TrueDoctor:
### Prerequisite checklist

 - [ x] Read the [contribution guidelines](https://docs.unikraft.org/contribute.html) regarding submitting new changes to the project;

### Base target

### DIFF
--- a/lib/ukrust-sys/Cargo.toml
+++ b/lib/ukrust-sys/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "unikraft-sys"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dependencies]
+
+[build-dependencies]
+bindgen = {version= "0.59", default-features = false}
--- a/lib/ukrust-sys/build.rs
+++ b/lib/ukrust-sys/build.rs
@@ -1,0 +1,38 @@
+extern crate bindgen;
+
+use std::env;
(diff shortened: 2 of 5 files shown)


vladandrew:
Hi Denis,

Would it make sense to add this to the external lib-rust repo? And maybe, we could progressively integrate more and more from that repo into the main repo?

cffs:
I have the same question: what prevent generating those bindings from an external library?
Additionally, if we use `lib-rust` both for Makefile rules allowing to use cargo for user-space apps (see #291), and for bindings to unikraft core, then need to put it behind an option. One might want to use cargo integration `without` the bindings.

cffs:
Assuming we put this in core, what would be the relation between `lib/ukrust` (see #293) and `lib/ukrust-sys`? Shouldn't those be merged?

TrueDoctor:
@vladandrew @cffs 
Yes, this can be an external repo, it is not even integrated as an internal library, rather just a crate that you can depend on.
The one tricky thing is that It depends on the list of header files being exposed by the buildsystem, so it is not completely self sustained. We might be able to factor out the required changes to `Makefile.rules` into this repo.
We would not need to put the binding generation behind options because it is not actually executed by the build system directly but rather by the dependencies.
ukrust-sys provides raw bindings while ukrust will be an opinionated easier to use wrapper.
they don't have to live in separate repos, but it is rust convention (and quite sensible I think) to split that into two crates.

TrueDoctor:
https://github.com/unikraft/lib-rust/pull/2

PR #293, opened 2021-09-23
Add wrapper for ukrust-sys

TrueDoctor:
### Prerequisite checklist

### Base target


cffs:
What prevent this to be in external `ukrust-sys` only? Why does it have to be in core?

TrueDoctor:
> What prevent this to be in external `ukrust-sys` only? Why does it have to be in core?

I don't think it has to be a core lib, should I put it in a different repo then?

vladandrew:
> > What prevent this to be in external `ukrust-sys` only? Why does it have to be in core?
> 
> I don't think it has to be a core lib, should I put it in a different repo then?

I was thinking of using the same lib-rust external repo. What do you think?

cffs:
> I was thinking of using the same lib-rust external repo. What do you think?

Seems sensible to me. We could have a single `lib-rust` external repository with:
- the `Makefile.rules` for handling `.rs` and `Cargo.toml` files (see #291);
- the `ukrust-sys` crate (see #292);
- this wrapper crate.

TrueDoctor:
Closing in favor of https://github.com/unikraft/lib-rust/pull/3

PR #297, opened 2021-09-28
arch/arm: Introduce SMCCC implementation

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): common, kvm
 - Application(s): N/A

### Additional configuration

- CONFIG_ARCH_ARM_64

### Description of changes

This PR introduces an implementation of SMCCC v1.3 as defined in ARM DEN 0028D.


include/uk/arch/arm/smccc.h
```diff
@@ -0,0 +1,274 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __INLCUDE_UKARCH_SMCCC_H__
```
razvanvirtan: Maybe `#ifndef __UKARCH_SMCCC_H__` would look cleaner?

plat/common/arm/smccc_invoke.S
```diff
@@ -0,0 +1,90 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <uk/asm.h>
+#include <uk/arch/arm/smccc.h>
+
+#ifdef CONFIG_ARCH_ARM_64
+.macro smccc conduit
+
+	mov x18, x0
+
```
razvanvirtan, line 39: Shouldn't we preserve x4-x18, the same way we do for x4-x8 in the SMC32/HVC32  case below?

razvanvirtan (commented):
I think the source files should be added to the build in `plat/kvm/Makefile.uk`.

Also, I have tried to run a simple helloworld app with the following code:
```
	struct smccc_config config;
	config.version = SMCCC_VERSION_1_3;
	config.conduit = SMCCC_CONDUIT_HVC;
	smccc_init(&config);
	unsigned long result = smccc_arch_soc_id(SMCCC_ARCH_SOC_VERSION);
```
and I get to an EL1 sync trap. The same thing happens for other calls too, like:
```
unsigned long result = smccc_version();
```
or
```
unsigned long result = smccc_arch_features(0x80000000);
```
(Maybe I got something wrong about how to test the PR)

include/uk/arch/arm/smccc.h
michpappas: Ooops, sorry bad habit from elsewhere. Fixed.

plat/common/arm/smccc_invoke.S
michpappas: In theory it should be caller saved, but I just had a look at the [AAPCS64](https://github.com/ARM-software/abi-aa/releases/download/2021Q3/aapcs64.pdf) and it turns out that using `x18` can be problematic. I'll update my PR to use (preserved) `x19` instead.

michpappas:
> I think the source files should be added to the build in `plat/kvm/Makefile.uk`.
> 
> Also, I have tried to run a simple helloworld app with the following code:
> 
> ```
> 	struct smccc_config config;
> 	config.version = SMCCC_VERSION_1_3;
> 	config.conduit = SMCCC_CONDUIT_HVC;
> 	smccc_init(&config);
> 	unsigned long result = smccc_arch_soc_id(SMCCC_ARCH_SOC_VERSION);
> ```
> 
> and I get to an EL1 sync trap. The same thing happens for other calls too, like:
> 
> ```
> unsigned long result = smccc_version();
> ```
> 
> or
> 
> ```
> unsigned long result = smccc_arch_features(0x80000000);
> ```
> 
> (Maybe I got something wrong about how to test the PR)

Hi @razvanvirtan, thanks for looking at this! I'm not sure what platform did you use for your test, but I assume you're using QEMU with `virt`. In any case, if there's no one to handle the SMC it's normal that you get an abort. If you're issuing an `hvc`, you would require a hypervisor that implements a handler for that request. If you are issuing an `smc`, you would need to have at least Secure Monitor in EL3, with the corresponding handler.

I have (successfully) tested this in real hardware, but I intend to provide at some point a qemu-based setup for testing.

marcrittinghaus:
Sorry, won't have the chance to look into this in detail today. Have to move it to 0.7.

michpappas:
> Sorry, won't have the chance to look into this in detail today. Have to move it to 0.7.

That's okay, no worries 👍🏼

plat/common/arm/smccc_invoke.S
michpappas: Fixed

michpappas:
Hi, @razvanvirtan. I completely forgot that qemu actually supports PSCI :) The following works:
```
#include <cpu_defs.h>
#include <stdio.h>
#include <uk/arch/arm/smccc.h>

{
	(void)argc;
	(void)argv;

	struct smccc_config config;
	struct smccc_args args = {0};

	/* Use default SMCCC version (ie 1.2) and set
	 * the conduit to HVC for qemu. Notice that you
	 * can alternatively ommit smc_init(), which use
	 * the defaults (SMCCC v1.2 and SMC), and 
	 * use HVC explicitly by calling smccc_hvc() below.
	 */
	config.conduit = SMCCC_CONDUIT_HVC;
	smccc_init(&config);

	/* Test PSCI Version */
	args.a0 = PSCI_FNID_VERSION;
	smccc_invoke(&args);
	printf("PSCI version: 0x%lx\n", args.a0);

	/* Test PSCI System Reset */
	args.a0 = PSCI_FNID_SYSTEM_RESET;
	smccc_invoke(&args);

	return 0;
}
```
For any SMC / HVC other than PSCI (including FIDs with the SVE bit set when using SMCCC v1.3) qemu will trigger an exception.

michpappas:
@razvanvirtan, the error you will see on reset is caused by #360

razvanvirtan:
@mpp-opsy I've tested again using `PSCI_FNID_VERSION` and it works as expected for SMCC version 1.1 and 1.2. (Problems appear indeed for version 1.3, when the FPSIMD is not enabled in the configuration menu). This is great, since we can integrate it in the SMP support.

Also, I think there is one more point to discuss here. We already have a PSCI interface implementation in Unikraft (see `smcc_psci_hvc_call()` and `smcc_psci_smc_call()`, declared in `arm64/cpu.h` and implemented in `plat/common/arm/psci_arm64.S`). I think we have three ways to go:
- keep the `smcc_psc_hvc/smc_call()` functions and replace their implementations with `smc_invoke()` calls
- completely remove the existing functions and just use `smc_invoke()` with corresponding parameters wherever we need it (e.g. in the SMP implementation)
- keep the old implementation as it is (I don't know if this would be a good choice, since it would introduce some duplicate work)

What do you think about this?

michpappas:
Hey @razvanvirtan good thing you brought up the PSCI topic 😄  I was planning to come up with a follow-up PR once generic SMCCC support was merged, but since we're at it we might as well discuss it now.

On the proposed approaches:

> * keep the smcc_psc_hvc/smc_call() functions and replace their implementations with smc_invoke() calls

We need to have a separate conduit for PSCI. That is because there are systems that use TrustZone, along with a hypervisor in EL2 that needs to handle PSCI calls to implement features like vm reboot. This type of configuration would justify configuring smccc_invoke() to issue SMCs for most calls, but at the same time require issuing HVCs for PSCI. [The `psci` node type in the dts provides its own `method` property](https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/psci.txt) for that purpose. Notice that technically it is possible for a hypervisor to trap all SMCs in EL2, and determine whether it is a PSCI call from the FID, but it's better that we're flexible here.

Given the above, the current implementation that allows the platform to choose explicitly which method to use for PSCI looks good to me. A platform can check whether there are conduits for PSCI and TEE in the device-tree, and otherwise fall-back to platform-specific defaults.

> * completely remove the existing functions and just use smc_invoke() with corresponding parameters wherever we need it (e.g. in the SMP implementation)

This is possible, but it would still require using the right conduit on every call. The current PSCI interface saves us from that, as you initialize and forget.

> keep the old implementation as it is (I don't know if this would be a good choice, since it would introduce some duplicate work)

So what I had in mind was to retire `smcc_psci_hvc_call()` / `smcc_psci_smc_call()`, and update the platform to register `smccc_smc()` / `smccc_hvc()`, which will solve the duplication problem you mention. What do you think?

razvanvirtan:
> 

Sounds good to me, I think we should provide flexibility regarding the psci method at the platform level (I've also realized that we are always using hvc calls to boot cpus in the current SMP draft, I will add a check for the psci "method" property).

razvanvirtan:
> Hey @razvanvirtan good thing you brought up the PSCI topic smile I was planning to come up with a follow-up PR once generic SMCCC support was merged, but since we're at it we might as well discuss it now.
> 
> On the proposed approaches:
> 
> > * keep the smcc_psc_hvc/smc_call() functions and replace their implementations with smc_invoke() calls
> 
> We need to have a separate conduit for PSCI. That is because there are systems that use TrustZone, along with a hypervisor in EL2 that needs to handle PSCI calls to implement features like vm reboot. This type of configuration would justify configuring smccc_invoke() to issue SMCs for most calls, but at the same time require issuing HVCs for PSCI. [The `psci` node type in the dts provides its own `method` property](https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/psci.txt) for that purpose. Notice that technically it is possible for a hypervisor to trap all SMCs in EL2, and determine whether it is a PSCI call from the FID, but it's better that we're flexible here.
> 
> Given the above, the current implementation that allows the platform to choose explicitly which method to use for PSCI looks good to me. A platform can check whether there are conduits for PSCI and TEE in the device-tree, and otherwise fall-back to platform-specific defaults.
> 
> > * completely remove the existing functions and just use smc_invoke() with corresponding parameters wherever we need it (e.g. in the SMP implementation)
> 
> This is possible, but it would still require using the right conduit on every call. The current PSCI interface saves us from that, as you initialize and forget.
> 
> > keep the old implementation as it is (I don't know if this would be a good choice, since it would introduce some duplicate work)
> 
> So what I had in mind was to retire `smcc_psci_hvc_call()` / `smcc_psci_smc_call()`, and update the platform to register `smccc_smc()` / `smccc_hvc()`, which will solve the duplication problem you mention. What do you think?

@michpappas Sounds good to me, I agree that we should provide flexibility regarding the psci method at the platform level (I've also realized that we are always using hvc calls to boot cpus in the current SMP draft, I will add a check for the psci "method" property).

razvanvirtan (approved):
Looks good to me.

michpappas:
@razvanvirtan I just pushed an update that moves `smccc.h` to `plat/common/include/arm/`. Everything else is unchanged, so if you're still happy with it perhaps it's time for @marcrittinghaus to have a look? :)

@skuenzer with this update it is not possible for applications to access the SMCCC API. This should be okay for the time being, but I think we should eventually come up with a formalized approach on how applications that implement kernel-like / baremetal logic can access low-level APIs.

razvanvirtan:
> @razvanvirtan I just pushed an update that moves `smccc.h` to `plat/common/include/arm/`. Everything else is unchanged, so if you're still happy with it perhaps it's time for @marcrittinghaus to have a look? :)
> 
> @skuenzer with this update it is not possible for applications to access the SMCCC API. This should be okay for the time being, but I think we should eventually come up with a formalized approach on how applications that implement kernel-like / baremetal logic can access low-level APIs.

@michpappas Yes, everything seems still ok to me :)

marcrittinghaus:
ok, I will have a look, hopefully until our meeting next week. I send a patch file then. No, just kidding 😄

plat/common/arm/smccc_invoke.S
```diff
@@ -0,0 +1,94 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <arm/smccc.h>
+#include <uk/asm.h>
+
+#ifdef CONFIG_ARCH_ARM_64
+.macro smccc conduit
+
+	str   x19, [sp, #-16]!
```
marcrittinghaus: Is there a specific reason why the indentation for the register is different here (and in the corresponding `ldr`) than for the `ldp`/`stp`?

plat/common/arm/smccc_invoke.S
```diff
@@ -0,0 +1,94 @@
...
+
+	mov x19, x0
+
+	ldp x0, x1, [x19]
+	ldp x2, x3, [x19, #16]
+	ldp x4, x5, [x19, #32]
+	ldp x6, x7, [x19, #48]
```
marcrittinghaus: Please adapt indentation for the `[x19...]`-part so it is same for all loads (just as with the stores).

plat/common/include/arm/smccc.h
```diff
@@ -0,0 +1,274 @@
...
+#define SMCCC_FNUM_SHIFT	0U
+
+#define SMCCC_SVE_STATE_ON	0U
+#define SMCCC_SVE_STATE_OFF	1U
+
+#define SMCCC_FID(_type, _cc, _svc, _fnum)				\
+	((((_type) & SMCCC_TYPE_MASK) << SMCCC_TYPE_SHIFT) |		\
+	 (((_cc) & SMCCC_CALL_CONV_MASK) << SMCCC_CALL_CONV_SHIFT) |	\
+	 (((_svc) & SMCCC_SERVICE_MASK) << SMCCC_SERVICE_SHIFT) |       \
+	 ((_fnum) & SMCCC_FNUM_MASK) << SMCCC_FNUM_SHIFT)
```
marcrittinghaus: Please change the spaces in line 78 at the end to a tab 🔥 😄 

The macro is broken. The masks will kill the arguments as the macro receives unshifted values. It should be:
```suggestion
#define SMCCC_FID(_type, _cc, _svc, _fnum)				\
	((((_type) << SMCCC_TYPE_SHIFT) & SMCCC_TYPE_MASK) |		\
	 (((_cc) << SMCCC_CALL_CONV_SHIFT) & SMCCC_CALL_CONV_MASK) |	\
	 (((_svc) << SMCCC_SERVICE_SHIFT) & SMCCC_SERVICE_MASK) |	\
	 ((_fnum) << SMCCC_FNUM_SHIFT) & SMCCC_FNUM_MASK)
```

plat/common/include/arm/smccc.h
```diff
@@ -0,0 +1,274 @@
...
+#define SMCCC_CC_32		0U
+#define SMCCC_CC_64		1U
+
+#define SMCCC_TYPE_MASK		0x80000000U
+#define SMCCC_TYPE_SHIFT	31U
+#define SMCCC_CALL_CONV_MASK	0x40000000U
+#define SMCCC_CALL_CONV_SHIFT	30U
+#define SMCCC_SERVICE_MASK	0x3f000000U
+#define SMCCC_SERVICE_SHIFT	24U
+#define SMCCC_SVE_STATE_MASK	0x10000U
```
marcrittinghaus:
```suggestion
#define SMCCC_SVE_STATE_MASK	0x00010000U
```

plat/common/include/arm/smccc.h
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * Requests the firmware to apply workaround for CVE-2017-5715
+ *
+ * This function should be called on every PE that requires a firmware
+ * mitigation for CVE-2017-5715. Should be called only when feature
+ * discovery returns a non-negative value. For more information see
+ * ARM DEN0028C Sect. 7.5
+ */
+void smccc_arch_workaround_1(void);
```
marcrittinghaus, line 240: Originally, I thought that this should be called differently. Perhaps `smccc_arch_workaround_2017_5715`. But then I found that the manual uses this name so it is probably better to keep it although a bit ugly.

marcrittinghaus (changes requested):
Hi @michpappas,

Looks very good. I think the `SMCCC_FID` macro is broken. I also left some minor style hints.

I assume custom calls fall into the `TRUSTED_APP` - `TRUSTED_APP_END` / `TRUSTED_OS` - `TRUSTED_OS_END` ranges. Do you use these and do you call them frequently? 

Perhaps it might make sense to optimize the interface in the future, if this is the case. I am thinking about a syscall shim like macro magic that would set and read only the registers necessary for the specific call. Although this might also introduce a lot of call gate code. For now, it is fine I think.

marcrittinghaus:
> So what I had in mind was to retire `smcc_psci_hvc_call()` / `smcc_psci_smc_call()`, and update the platform to register `smccc_smc()` / `smccc_hvc()`, which will solve the duplication problem you mention. What do you think?

I agree. @razvanvirtan Do you want to do this after the new SMCCC code is upstream?

razvanvirtan:
> > So what I had in mind was to retire `smcc_psci_hvc_call()` / `smcc_psci_smc_call()`, and update the platform to register `smccc_smc()` / `smccc_hvc()`, which will solve the duplication problem you mention. What do you think?
> 
> I agree. @razvanvirtan Do you want to do this after the new SMCCC code is upstream?

@marcrittinghaus Sure, sounds good

plat/common/arm/smccc_invoke.S
michpappas: None other than sheer carelessness :) Will fix thanks.

plat/common/arm/smccc_invoke.S
michpappas: Yup

plat/common/include/arm/smccc.h
michpappas: Ooops you're right, thanks for that

plat/common/include/arm/smccc.h
michpappas: Ok

plat/common/include/arm/smccc.h
michpappas: Yeah, it's how Arm calls it in the spec so I thought it would be a good idea to stick to the standard terminology despite being ugly 👹

michpappas:
@marcrittinghaus I updated the PR. All your comments have been addressed, thanks.

marcrittinghaus:
@michpappas Super, thank you very much! 😃 Just one last question. As I have yet to actually test the code, I just recognized that there is no modification to the platform Makefile to include the new `.c`/`.S.` files. Is this intentional or does this change is missing in the PR?

michpappas:
@marcrittinghaus, that's true, I didn't mean to add these in this PR. I thought we could introduce support in KVM in the next PR when @razvanvirtan updates the implementation of psci calls. But if you prefer to do it now, it's also okay, it's just two lines in the Makefile.

marcrittinghaus:
@michpappas Ok. As it should not break anything I would include the Makefile changes so that this functionality is self-contained. I see the update to the PSCI calls more like a cleanup PR that removes the two duplicate functions.

michpappas:
@marcrittinghaus I just pushed an update. It introduces an additional commit to add support into KVM, along with a small change on `smccc.h` to fix a warning I didn't notice earlier:
```
In file included from /home/mpp/devel/unikraft_upstream/uk_smccc/unikraft/plat/common/arm/smccc.c:32:
/home/mpp/devel/unikraft_upstream/uk_smccc/unikraft/plat/common/arm/smccc.c: In function 'smccc_svc_query':
/home/mpp/devel/unikraft_upstream/uk_smccc/unikraft/plat/common/include/arm/smccc.h:79:33: warning: suggest parentheses around arithmetic in operand of '|' [-Wparentheses]
   79 |   ((_fnum) << SMCCC_FNUM_SHIFT) & SMCCC_FNUM_MASK)
      |                                 ^
/home/mpp/devel/unikraft_upstream/uk_smccc/unikraft/plat/common/arm/smccc.c:100:12: note: in expansion of macro 'SMCCC_FID'
  100 |  args.a0 = SMCCC_FID(SMCCC_TYPE_FAST, SMCCC_CC_32,
      |            ^~~~~~~~~
```

marcrittinghaus (approved):

PR #299, opened 2021-10-07
build: `make V=2` warns about undefined variables

skuenzer:
### Description of changes

This PR introduces verbosity level 2 for the build system. The output is like level 1 with additional warnings for referenced but undefined variables. This can be a useful tool for debugging and developing Unikraft build recipes.


skuenzer:
This PR is re-opened as PR #305

PR #300, opened 2021-10-07
build: print-vars: Value only for simply expanded variables

skuenzer:
### Description of changes

This PR makes `make print-vars` work again by only showing the value of variables that are simply expanded (`:=` or `::=`). Printing the value of recursive variables caused errors for some variables that we use as function to generate build recipes.
This PR adopts the  output format: For each variable, the variable origin is printed before the variable name. Afterwards the value is shown for simply expanded variables. Otherwise the flavor is shown.

Example:
```
[file] SIMPLE_VAR := Hello World!
[file] RECURSIVE_VAR = <recursive>
```


skuenzer:
This PR is re-opened as PR #304

PR #308, opened 2021-10-19
uknetdev, plat/kvm: Support for partial checksumming / checksum offloading

skuenzer:
For checksum offloading for TCP and UDP flows, typically, a partial checksum needs to be computed. This checksum covers only the pseudo header of the protocols. By specifying a start pointer and another pointer to the checksum field, a hardware network device is able to complete the computation of the checksum.
This PR introduces support for `lib/uknetdev` for such a feature. Two new flags and two fields for pbufs are introduced to specify information needed by the driver to offload the computation for a packet. Additionally, a new `uknetdev` driver state is introduced: `UK_NETDEV_UNPROBED`. This state indicates that network devices features weren't negotiated with the device yet. During device initialization, a call to `uk_netdev_probe()` is becoming mandatory to turn the device into `UK_NETDEV_UNCOFNIGURED` state. For some devices (like virtio-net) this is required to figure out if the backend supports partial checksumming.
This PR also implements partial checksumming for `virtio-net`.

### DIFF
--- a/lib/uknetdev/include/uk/netbuf.h
+++ b/lib/uknetdev/include/uk/netbuf.h
@@ -101,9 +101,33 @@ typedef void (*uk_netbuf_dtor_t)(struct uk_netbuf *);
  * uk_netbuf_prepare_buf() are placing all these three regions into a single
  * allocation.
  */
+
+/* Packet is validated (non-corrupted, checksum okay)
+ * When sending or receiving a packet with a partial checksum, the
+ * DATA_VALID bit should also be set in order to confirm that the
+ * payload is not corrupted.
+ */
+#define UK_NETBUF_F_DATA_VALID_BIT   0
+#define UK_NETBUF_F_DATA_VALID       (1 << UK_NETBUF_F_DATA_VALID_BIT)
+
+/* Indicates that a protocol Checksum was only partially computed (e.g, TCP,
+ * UDP). `csum_start` and `csum_offset` fields indicate over which bytes
+ * the checksum needs to be computed and to which offset the result
+ * needs to be written. On transmit this is typically done by a physical
+ * interface card when leaving the host.
+ * NOTE: The defacto standard is that IP header checksums are always
+ * computed in software). In the case of TCP and UDP, a softrware computed
+ * partial checksum (just the pseudo header) is filled into the checksum
+ * field.
+ */
+#define UK_NETBUF_F_PARTIAL_CSUM_BIT 1
+#define UK_NETBUF_F_PARTIAL_CSUM     (1 << UK_NETBUF_F_PARTIAL_CSUM_BIT)
+
 struct uk_netbuf {
 	struct uk_netbuf *next;
 	struct uk_netbuf *prev;
+
+	uint8_t flags;         /**< Flags for this netbuf */
 
 	void *data;            /**< Payload start, is part of buf. */
 	uint16_t len;          /**< Payload length (should be <= buflen). */
@@ -113,6 +137,15 @@ struct uk_netbuf {
 
 	void *buf;             /**< Start address of contiguous buffer. */
 	size_t buflen;         /**< Length of buffer. */
+
+	uint16_t csum_start;   /**< Used if UK_NETBUF_F_PARTIAL_CSUM is set;
+				 * Offset within this netbuf's data segment to
+				 * begin checksumming
+				 */
+	uint16_t csum_offset;  /**< Used if UK_NETBUF_F_PARTIAL_CSUM is set;
+				 * Number of bytes starting from `csum_start`
+				 * pointing to the checksum field
+				 */
 
 	uk_netbuf_dtor_t dtor; /**< Destructor callback */
 	struct uk_alloc *_a;   /**< @internal Allocator for free'ing */
@@ -520,6 +553,14 @@ static inline int uk_netbuf_header(struct uk_netbuf *head, int16_t len)
 
 	head->data   -= len;
 	head->len    += len;
+
+	if (head->flags & UK_NETBUF_F_PARTIAL_CSUM) {
+		/* When header gets removed, delete PARTIAL_CSUM flag */
+		if (unlikely((len < 0) && (-len > head->csum_start)))
+			head->flags &= ~UK_NETBUF_F_PARTIAL_CSUM;
+		else
+			head->csum_start += len;
+	}
 	return 1;
 }
 
--- a/plat/drivers/virtio/virtio_net.c
+++ b/plat/drivers/virtio/virtio_net.c
@@ -1054,7 +1054,7 @@ static void virtio_net_info_get(struct uk_netdev *dev,
 	dev_info->nb_encap_tx = sizeof(struct virtio_net_hdr_padded);
 	dev_info->nb_encap_rx = sizeof(struct virtio_net_hdr_padded);
 	dev_info->ioalign = sizeof(void *); /* word size alignment */
-	dev_info->features = UK_FEATURE_RXQ_INTR_AVAILABLE;
+	dev_info->features = UK_NETDEV_F_RXQ_INTR;
 }
 
 static int virtio_net_start(struct uk_netdev *n)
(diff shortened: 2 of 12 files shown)


lib/uknetdev/include/uk/netbuf.h
```diff
@@ -101,10 +101,34 @@ typedef void (*uk_netbuf_dtor_t)(struct uk_netbuf *);
...
+ * field.
+ */
+#define UK_NETBUF_F_PARTIAL_CSUM_BIT 1
+#define UK_NETBUF_F_PARTIAL_CSUM     (1 << UK_NETBUF_F_PARTIAL_CSUM_BIT)
+
 struct uk_netbuf {
 	struct uk_netbuf *next;
 	struct uk_netbuf *prev;
 
+	uint8_t flags;         /**< Flags for this netbuf */
```
craciunoiuc, line 130: Is this flags element placed correctly? Considering 8 byte alignment (for 64bit), this might cause unnecessary padding.

If I remember correctly, the compiler might move elements in a structure to keep padding to a minimum. Is this true?

Furthermore, if padding is inevitable, maybe it's worth moving to `uint16_t` so the flags storage is more than enough in the future.

lib/uknetdev/include/uk/netbuf.h
```diff
@@ -520,6 +553,14 @@ static inline int uk_netbuf_header(struct uk_netbuf *head, int16_t len)
 
 	head->data   -= len;
 	head->len    += len;
+
+	if (head->flags & UK_NETBUF_F_PARTIAL_CSUM) {
+		/* When header gets removed, delete PARTIAL_CSUM flag */
+		if (unlikely((len < 0) && (-len > head->csum_start)))
```
craciunoiuc, line 559:
```suggestion
		if (unlikely((len < 0) && (len < head->csum_start)))
```
Isn't it more concise like this? That `-` might be easy to miss.

craciunoiuc (commented):
Hey @skuenzer,

I looked over your commits and I left a couple of comments.
All in all, the patches look good. I will let @mogasergiu review the patches first on the `lwip` part.

There is a typo in the second commit description `flagson` -> `flags on`.

Afterwards I'll come back and we'll continue the review/approval.

Cezar

lib/uknetdev/include/uk/netbuf.h
```diff
@@ -101,10 +101,34 @@ typedef void (*uk_netbuf_dtor_t)(struct uk_netbuf *);
...
+#define UK_NETBUF_F_DATA_VALID       (1 << UK_NETBUF_F_DATA_VALID_BIT)
+
+/* Indicates that a protocol Checksum was only partially computed (e.g, TCP,
+ * UDP). `csum_start` and `csum_offset` fields indicate over which bytes
+ * the checksum needs to be computed and to which offset the result
+ * needs to be written. On transmit this is typically done by a physical
+ * interface card when leaving the host.
+ * NOTE: The defacto standard is that IP header checksums are always
+ * computed in software). In the case of TCP and UDP, a softrware computed
+ * partial checksum (just the pdeudo header) is filled into the checksum
```
craciunoiuc: There is a typo here: `pdeudo` -> `pseudo`

plat/drivers/virtio/virtio_net.c
```diff
@@ -372,8 +368,23 @@ static int virtio_netdev_xmit(struct uk_netdev *dev,
 	/**
 	 * Fill the virtio-net-header with the necessary information.
 	 * Zero explicitly set.
+	 * NOTE: We do not set VIRTIO_NET_HDR_F_DATA_VALID and
+	 *       VIRTIO_NET_HDR_F_RSC_INFO because we aren't allowed
+	 *       according to virtio specification during sending.
+	 * TODO: We assume that the PARTIAL_CSUM flag is only set with
+	 *       the first netbuf of a queue. If this is not the case,
+	 *       (e.g., due to encapsulation of protocol headers with
+	 *        prepending netbufs) we we need to replace the call
```
craciunoiuc: Duplicate `we` here.

plat/drivers/virtio/virtio_net.c
```diff
@@ -830,10 +858,15 @@ static __u16 virtio_net_mtu_get(struct uk_netdev *n)
 	return d->mtu;
 }
 
-static int virtio_netdev_feature_negotiate(struct virtio_net_device *vndev)
+static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 {
 	__u64 host_features = 0;
+	__u64 drv_features  = 0x0;
```
craciunoiuc: No need to write it in hex if it's 0.

plat/drivers/virtio/virtio_net.c
```diff
@@ -843,17 +876,59 @@ static int virtio_netdev_feature_negotiate(struct virtio_net_device *vndev)
...
+	if (!VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_STATUS)) {
+		uk_pr_err("%p: Host system does not offer MTU feature\n", n);
+		rc = -EINVAL;
+		goto err_negotiate_feature;
+	}
+	VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_STATUS);
+
+	/**
+	 * Gratuitous ARP
+	 * NOTE: We tell that we will do gratuitous ARPs ourself.
```
craciunoiuc: Typo here: `ourself` -> `ourselves`

craciunoiuc (changes requested):
Hey @skuenzer,

I reviewed your new changes and came back with some more comments.

Still, the patch series looks good and it works for me. As soon as Sergiu finishes
his review I will also approve this.

Cezar

lib/uknetdev/include/uk/netbuf.h
skuenzer: Since this struct is never used to interact with an actual device directly, padding doesn't matter. A driver will parse the fields and will feed device specific descriptors. Whatever is making accessing fields in this struct faster is actually good for the networking subsystem - so I prefer leaving it up to the compiler. I don't expect many flags in the future but we can still extend its data type to `uint16_t` when needed since everything is recompiled together. The `uint8_t` was followed the `pbuf` example in lwip.

plat/drivers/virtio/virtio_net.c
skuenzer: Sure not needed, I can remove it if you prefer - I don't mind. I find it supportive to the understanding that we are expecting bit flags to be set when using the `0x` notation - kind of: "this is not an ordinary number, we will do bit-wise operations on it."

lib/uknetdev/include/uk/netbuf.h
skuenzer: This change is would not be the same expression anymore.
It is either
`-len > head->csum_start`
or
`len < -head->csum_start`
but in both cases one minus is staying.

skuenzer:
The pushed update fixes the comments from your review, @craciunoiuc . Thanks for your feedback!

lib/uknetdev/include/uk/netbuf.h
craciunoiuc: Whoops, you're right

lib/uknetdev/include/uk/netbuf.h
craciunoiuc: Sounds good. I just wanted to highlight that we have space for more if needed 😉

craciunoiuc (approved):
Hey @skuenzer @marcrittinghaus,

The PR looks good to me now. Please, @marcrittinghaus, take care when approving this PR such that it is introduced in the correct order with respect to the other PRs related (https://github.com/unikraft/lib-lwip/pull/12 https://github.com/unikraft/fork-lwip/pull/1 https://github.com/unikraft/lib-lwip/pull/10)

marcrittinghaus (approved):
I did not do a detailed review but I have been using this code for quite some time internally and the code is of high quality. The PR depends on https://github.com/unikraft/lib-lwip/pull/10 being merged, too. Otherwise, network breaks.

PR #309, opened 2021-10-21
arch, plat: Define the common SMP API

cristian-vijelie:
### Prerequisite checklist

### Base target

 - Architecture(s): [`ARM`, `x86_64`]
 - Platform(s): [`linuxu`, `kvm`, `xen`]
 - Application(s): [N/A]

### Description of changes

This Pull Request defines the common API for the SMP support in Unikraft, adds the configuration variables that will be used by the SMP component, and disables all the platforms, if SMP is enabled, until the implementation of SMP for those platforms will be provided.

### DIFF
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -2,7 +2,6 @@
 /*
  * Authors: Simon Kuenzer <simon.kuenzer@neclab.eu>
  *
- *
  * Copyright (c) 2017, NEC Europe Ltd., NEC Corporation. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -35,17 +34,11 @@
 #define __UKPLAT_LCPU_H__
 
 #include <uk/arch/time.h>
+#include <uk/arch/lcpu.h>
+#include <uk/essentials.h>
 
 #ifdef __cplusplus
 extern "C" {
-#endif
-
-#if UKPLAT_LCPU_MULTICORE
-__u8 ukplat_lcpu_id(void);
-__u8 ukplat_lcpu_count(void);
-#else
-#define ukplat_lcpu_id()    (0)
-#define ukplat_lcpu_count() (1)
 #endif
 
 /**
@@ -60,22 +53,28 @@ void ukplat_lcpu_disable_irq(void);
 
 /**
  * Returns current interrupt flags and disables them
- * @return interrupt flags (Note that the format is unspecified)
+ *
+ * @return interrupt flags (the format is unspecified)
  */
 unsigned long ukplat_lcpu_save_irqf(void);
 
 /**
  * Loads interrupt flags
- * @param flags interrupt flags (Note that the format is unspecified)
+ *
+ * @param flags interrupt flags (the format is unspecified)
  */
 void ukplat_lcpu_restore_irqf(unsigned long flags);
 
 /**
  * Checks if interrupts are disabled
+ *
  * @return non-zero value if interrupts are disabled
  */
 int ukplat_lcpu_irqs_disabled(void);
 
+/**
+ * TODO: This is Xen specific. Remove from public API?
+ */
 void ukplat_lcpu_irqs_handle_pending(void);
 
 /**
@@ -84,18 +83,115 @@ void ukplat_lcpu_irqs_handle_pending(void);
 void ukplat_lcpu_halt(void);
 
 /**
- * Halts the current logical CPU execution
- * Execution is returned when an interrupt/signal arrived or
- * the specified deadline expired
+ * Halts the current logical CPU. Execution is resumed when an interrupt/signal
+ * arrives or the specified deadline expires
+ *
  * @param until deadline in nanoseconds
  */
 void ukplat_lcpu_halt_to(__nsec until);
 
 /**
- * Halts the current logical CPU execution
- * Execution is returned when an interrupt/signal arrived
+ * Halts the current logical CPU. Execution is resumed when an interrupt/signal
+ * arrives.
  */
 void ukplat_lcpu_halt_irq(void);
+
+#ifdef CONFIG_HAVE_SMP
+#include <uk/list.h>
+
+typedef void (*ukplat_lcpu_entry_t)(void) __noreturn;
+typedef __u32 __lcpuid;
+
+struct ukplat_lcpu_func {
+	struct uk_list_head lentry;
+
+	/**
+	 * Function to execute.
+	 *
+	 * @param regs pointer to a snapshot of the current CPU state. Changes
+	 *   to the state are applied after the function returns
+	 * @param fn pointer to this structure. The structure is not touched
+	 *   after function invocation and can be freed if necessary during
+	 *   function execution
+	 */
+	void (*fn)(struct __regs *regs, struct ukplat_lcpu_func *fn);
+
+	/* Optional user-supplied pointer. */
+	void *user;
+};
+
+/**
+ * Returns the ID of the current logical CPU
+ */
+__lcpuid ukplat_lcpu_id(void);
+
+/**
+ * Returns the number of logical CPUs present on the system
+ */
+__lcpuid ukplat_lcpu_count(void);
+
+/**
+ * Starts multiple logical CPUs and assigns them the given stacks. The logical
+ * CPUs execute the entry functions if supplied or enter a low-power wait state
+ * otherwise.
+ *
+ * @param lcpuid array with the IDs of the logical CPUs that are to be started
+ * @param sp array of stack pointers, one for each specified logical CPU
+ * @param entry optional array of entry function pointers, one for each
+ *   specified logical CPU. Provided functions must not return. If the
+ *   parameter or individual elements are NULL the respective logical CPUs
+ *   enter a low-power wait state after startup
+ * @param num number of logical CPU IDs given
+ *
+ * @return 0 on success, an errno-type error value otherwise
+ */
+int ukplat_lcpu_start(__lcpuid lcpuid[], void *sp[],
+		      ukplat_lcpu_entry_t entry[], unsigned int num);
+
+/**
+ * Waits for the specified logical CPUs to enter idle state, or until the
+ * timeout expires.
+ *
+ * @param lcpuid array with the IDs of the logical CPUs to wait for. Can be
+ *   NULL to include all logical CPUs except the one executing the function
+ * @param num number of logical CPU IDs given
+ * @param timeout timeout in nanoseconds for the wait to be satisfied. Can be
+ *   0 to wait indefinitely
+ *
+ * @return 0 if the wait for all specified logical CPUs has been satisfied,
+ *   an errno-type error value otherwise (e.g., timeout)
+ */
+int ukplat_lcpu_wait(__lcpuid lcpuid[], unsigned int num, __nsec timeout);
+
+/**
+ * Executes a function on the specified logical CPUs. The run function does not
+ * wait for the execution to start or complete. Multiple functions can be run
+ * at the same time without having to wait for their completion.
+ *
+ * @param lcpuid array with the IDs of the logical CPUs that should execute the
+ *   function. Can be NULL to execute the function on all logical CPUs except
+ *   the current one
+ * @param num number of logical CPU IDs given
+ * @param fn the function to be executed
+ * @param flags (architecture-dependent) flags that specify how the function
+ *   should be executed (see UKPLAT_LCPU_RFLG_* flags)
+ *
+ * @return 0 on success, an errno-type error value otherwise
+ */
+int ukplat_lcpu_run(__lcpuid lcpuid[], unsigned int num,
+		    struct ukplat_lcpu_func *fn, int flags);
+
+/**
+ * Wakes up the specified logical CPUs from a halt or low-power sleep state.
+ *
+ * @return 0 on success, an errno-type error value otherwise
+ */
+int ukplat_lcpu_wakeup(__lcpuid lcpuid[], unsigned int num);
+
+#else /* CONFIG_HAVE_SMP */
+#define ukplat_lcpu_id()	(0)
+#define ukplat_lcpu_count()	(1)
+#endif /* CONFIG_HAVE_SMP */
 
 #ifdef __cplusplus
 }
--- a/plat/Config.uk
+++ b/plat/Config.uk
@@ -12,6 +12,15 @@ config EARLY_PRINT_PL011_UART_ADDR
 	help
 		Pl011 serial address used by early debug console.
 
+config UKPLAT_LCPU_MAXCOUNT
+        int "Maximum number of supported logical CPUs"
+        range 1 256
+        default 1
+
+config HAVE_SMP
+        bool
+        default y if UKPLAT_LCPU_MAXCOUNT > 1
+        default n
 endmenu
 
 config HZ
--- a/plat/kvm/Config.uk
+++ b/plat/kvm/Config.uk
@@ -1,7 +1,7 @@
 menuconfig PLAT_KVM
        bool "KVM guest"
        default n
-       depends on (ARCH_X86_64 || (ARCH_ARM_64 && !HAVE_SYSCALL))
+       depends on ((ARCH_X86_64 && !HAVE_SMP) || (ARCH_ARM_64 && !HAVE_SMP && !HAVE_SYSCALL))
        select LIBUKDEBUG
        select LIBUKALLOC
        select LIBUKTIMECONV
--- a/plat/linuxu/Config.uk
+++ b/plat/linuxu/Config.uk
@@ -1,8 +1,7 @@
 menuconfig PLAT_LINUXU
        bool "Linux user space"
        default n
-       depends on (ARCH_X86_64) || (ARCH_ARM_32)
-       depends on !HAVE_SYSCALL
+       depends on ((ARCH_X86_64 && !HAVE_SYSCALL && !HAVE_SMP) || (ARCH_ARM_32 && !HAVE_SYSCALL && !HAVE_SMP))
        select LIBUKDEBUG
        select LIBNOLIBC if !HAVE_LIBC
        help
--- a/plat/xen/Config.uk
+++ b/plat/xen/Config.uk
@@ -1,9 +1,9 @@
 menuconfig PLAT_XEN
        bool "Xen guest image"
        default n
-       depends on (ARCH_X86_32 || ARCH_X86_64 || ARCH_ARM_32)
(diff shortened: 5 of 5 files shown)


razvanvirtan:
Hey, here is a patch file with some older `CONFIG_HAVE_SMP` placeholders that should be changed in this PR.
[placeholders.zip](https://github.com/unikraft/unikraft/files/7391145/placeholders.zip)

marcrittinghaus (dismissed):
Hi Cristian, hi Razvan,

thank you very much for pushing the SMP API 👍🏻 SMP is making good progress, thanks to you guys! I have done further cleanup, removed the X86/KVM specific header file (there should only be a new flag for x86 in `asm/lcpu.h`, but not as part of this PR. All other stuff is private to the x86 platform), and extended/clarified the documentation. Please apply the following patch:

[smp_api_v1.zip](https://github.com/unikraft/unikraft/files/7395769/smp_api_v1.zip)

There are also some open points we have to discuss further:
* Currently it is not entirely clear how to communicate errors. For example, if we call `start` or `run` it may be that we cannot issue the interrupt (to one or more CPUs) or that an lcpuid is invalid.
* I dislike the `is_bsp` function as I would prefer all CPUs to be treated the same. Why is this function needed? If I remember correctly, the function is not strictly needed on ARM either.

cristian-vijelie:
The `is_bsp` was proposed By the people from OpenSynergy. I put it there because I saw it in their proposal. It may be useful when it is not guaranteed that the BSP has the ID 0.

razvanvirtan:
> Hi Cristian, hi Razvan,
> 
> thank you very much for pushing the SMP API 👍🏻 SMP is making good progress, thanks to you guys! I have done further cleanup, removed the X86/KVM specific header file (there should only be a new flag for x86 in `asm/lcpu.h`, but not as part of this PR. All other stuff is private to the x86 platform), and extended/clarified the documentation. Please apply the following patch:
> 
> [smp_api_v1.zip](https://github.com/unikraft/unikraft/files/7395769/smp_api_v1.zip)
> 
> There are also some open points we have to discuss further:
> 
>     * Currently it is not entirely clear how to communicate errors. For example, if we call `start` or `run` it may be that we cannot issue the interrupt (to one or more CPUs) or that an lcpuid is invalid.
> 
>     * I dislike the `is_bsp` function as I would prefer all CPUs to be treated the same. Why is this function needed? If I remember correctly, the function is not strictly needed on ARM either.
I think `is_bsp` could only be useful when we boot APs, since that is the only moment when the difference between the boot core and the APs matters [1]. However,  I haven't used it in my initial boot code and, since the function would only check  for `ukplat_lcpu_id() == bspid`,  I think we can safely remove it.

[1] https://github.com/razvanvirtan/unikraft/pull/2/files#diff-47b8b12220e68c052272b4198ee33f962e5c9f8cf63b95fda9b47c0ac4fde9d7R181

cristian-vijelie:
I removed the `is_bsp` function.

michpappas:
Hi, just to clarify about `is_bsp`: We did use it before, but I believe it won't be needed under this threading model :)

marcrittinghaus:
Thank you @michpappas for the clarification 😃

So the remaining issue is how to properly report errors. For the paging code I currently have the design that the documentation just says "a non-zero value" but in fact I return error codes from `<errno.h>`. So if you want to create the dependency to libc wherever you are using the API you can interpret the error codes. Otherwise, you know something went wrong when a non-zero value is returned. However, I would also like to hear you guys opinion on this topic as this approach is a bit intransparent and I don't like that the caller does not have the guarantee from the specification that these will be `errno` values. So it is probably better to just state in the documentation that error values are of `errno` type.

What do you think?

include/uk/plat/lcpu.h
```diff
@@ -84,19 +84,116 @@ void ukplat_lcpu_irqs_handle_pending(void);
...
+ * Halts the current logical CPU. Execution is resumed when an interrupt/signal
+ * arrives.
  */
 void ukplat_lcpu_halt_irq(void);
 
+#ifdef CONFIG_HAVE_SMP
+#include <uk/list.h>
+
+typedef void (*ukplat_lcpu_entry_t)(void) __noreturn;
+typedef __u32 __lcpuid;
```
michpappas, line 103: Out of curiosity, why is lcpuid prefixed with underscores? AFAIK the C standard reserves these for the implementation (eg C99 7.1.3). Similarly for `__u32`, shouldn't it be replaced with the standard `uint32_t`? Or is there a UK-specific convention?

marcrittinghaus: The original idea was to not have a dependency on any libc headers in the basic APIs to draw a clear line between UK and the optional C library. Therefore, datatypes such as `uint32_t` are not desired here. Of course, in practice separation is not that easy and you quickly get a hard dependency on the C library. So there have been discussions on throwing away these own data types and just always include a C library. However, Simon does not like that idea and I agree, as I prefer the architectural split at this level.

Why the internal data types use double underscore, I don't know. This decision was made long before I joined the project. You are right, it is against the C standard. I see it as a convention at this point (thus also the `__spinlock`)

michpappas:
> Thank you @michpappas for the clarification smiley
> 
> So the remaining issue is how to properly report errors. For the paging code I currently have the design that the documentation just says "a non-zero value" but in fact I return error codes from `<errno.h>`. So if you want to create the dependency to libc wherever you are using the API you can interpret the error codes. Otherwise, you know something went wrong when a non-zero value is returned. However, I would also like to hear you guys opinion on this topic as this approach is a bit intransparent and I don't like that the caller does not have the guarantee from the specification that these will be `errno` values. So it is probably better to just state in the documentation that error values are of `errno` type.
> 
> What do you think?

Using errno codes sounds like a good idea. They are standard, they are also used in the linux kernel, and overall adhere to the [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). No strong opinion regarding the documentation, perhaps if they are commonly used across UK core it can be safely implied. In any case, if you get errors at such a low level state, you'll need to dive into the sources anyway :)

include/uk/plat/lcpu.h
michpappas: > The original idea was to not have a dependency on any libc headers in the basic APIs to draw a clear line between UK and the optional C library.

The way I understand this is that the requirement also includes nolibc, which defines `uint32_t`, right? Where is the line of separation? Perhaps we need to update some of our PRs under plat/ to comply with that. I believe it's safe to assume that internal libraries don't have this dependency?

> So there have been discussions on throwing away these own data types and just always include a C library.

Being able to build nearly bare-metal unikernels is a strong feature of Unikraft for virtualization, so maybe the rationale of separation is not too bad and it leaves the option of requiring something like nolibc to the application :)

michpappas:
While we're there, isn't also `errno` part of libc? If these are used in core functionality, then these values should be part of uk core, and be overriden if a standard library is used.

include/uk/plat/lcpu.h
marcrittinghaus: > The way I understand this is that the requirement also includes nolibc, which defines uint32_t, right? Where is the line of separation? Perhaps we need to update some of our PRs under plat/ to comply with that. I believe it's safe to assume that internal libraries don't have this dependency?

The minimum line of separation is that all public headers in `unikraft/include/uk` and `unikraft/arch/` should not use libc types, but only the internal `__X` types. It is less strict for the platform code, although I personally prefer the internal types or native C types (e.g., `unsigned long`) where possible just to not introduce unnecessary dependencies on the C library. It is perfectly fine for (internal) libraries to use libc data types, although in the rare cases, where there is no other dependency on libc (e.g., `memcpy`) it makes sense to at least considering to only use internal UK types and native C types.

This is why I removed the `uintX_t` types from the ARM headers in one of the outstanding PRs  (#231).

marcrittinghaus:
Yes, `errno.h` is part of libc. Thus my reluctance in returning `errno.h` codes in public UK APIs as we do not have a `uk/errno.h` in the base headers now. But I like the idea of having this header added and overridden, when there is a libc available. I would like to hear @skuenzer's opinion on this. It would give us the opportunity to globally return uniform error codes. In fact, in the paging code that I am currently working on, the implementation of the `munmap` syscall is now just a call to the internal paging `ukplat_page_unmap` API with special flags. Error codes and behavior are matched.

Another point is that when @cristian-vijelie introduced the first ACPI code, he liked to return error codes. So in the absence of `errno.h` as a standard, we had to define custom codes, which I still somewhat dislike.

marcrittinghaus:
I had a short discussion with @skuenzer this morning. We settled on having it properly documented that `errno` codes are returned. However, the `errno.h` header file should not be included and we prefer to not have a `uk/errno.h`. Simon thinks that the header file in nolibc fulfills this task well enough. I can live with this decision as it makes the API specification precise while not introducing any hard dependencies. Furthermore, chances are high that any code that wants to interpret the error codes and perhaps even report the error with `strerror` has a dependency on at least nolibc anyways.

@cristian-vijelie Could you change the documentation appropriately? I would then take a last look and we could get everything upstream.

@michpappas Any other suggestions, comments from your side? Are you happy with the solution?

michpappas:
> I had a short discussion with @skuenzer this morning. We settled on having it properly documented that errno codes are returned. However, the errno.h header file should not be included and we prefer to not have a uk/errno.h. 

So if I understand the idea correctly, for that to compile it would still require that someone uses at least nolibc.
Wouldn't that effectively make SMP a dependant of a libc? Unless you hardcode integers, that is :laughing:

cristian-vijelie:
> @cristian-vijelie Could you change the documentation appropriately? I would then take a last look and we could get everything upstream.

Sure. I'll do it by tomorrow.

marcrittinghaus:
> > I had a short discussion with @skuenzer this morning. We settled on having it properly documented that errno codes are returned. However, the errno.h header file should not be included and we prefer to not have a uk/errno.h.
> 
> So if I understand the idea correctly, for that to compile it would still require that someone uses at least nolibc. Wouldn't that effectively make SMP a dependant of a libc? Unless you hardcode integers, that is 😆

No no 😄 The code that returns `errno` codes can and should of course include `<errno.h>` and is thus dependent on at least the header being present. But you could technically also use your own defines for the codes, although that would obviously be a bit dirty. It is just that the API does not require you to include `errno.h`, if you are not interested in parsing the error codes. Not more. You may see this as having a static library that implements some API and the dependency on nolibc/libc is an implementation detail and not a strict requirement by the API (which would be different if `errno.h` would be included in the API header itself). So IF you wanted to, you could write an implementation of the API that does not use `errno.h` and no other dependency on libc.

In practice of course, you are right. If you are using the `errno.h` header file in your implementation of the API (which you should to avoid redundant definitions) then you already create a dependency on nolibc for the whole unikernel. But we have this for more complex APIs in most cases anyways due to `memcpy` etc. This is why I am looking at this as a compromise between architectural clarity and practice. Although from a pure architectural point of view, I personally would have preferred a standalone `uk/errno.h`.

michpappas:
@marcrittinghaus understood, thanks for that. For the current PR it should be fine, let's go ahead with it. Yet, this approach feels like we are migrating the problem to the implementation, rather than addressing it on its core. If the implementation is at the platform level one could argue that platforms should be free to depend on nolibc or whatever else implements these error codes, but the same thing cannot be claimed for implementations under arch/ or lib/. Do we have a plan on how to proceed with the implementation here?

There is another thing, specific to `errno.h`: If the API returns errno.h values and calls into components that do not use `errno.h`, we lose information. The API would have to guess or translate error codes into `errno.h`. Linux and BSDs use `errno` codes universally at the kernel level, without meaning that their kernels are dependant on libc :laughing: It's just that the error codes match those with libc's.

Anyway, I think that this needs to be discussed in the broader scope of such dependencies and establish a more formal approach on how to move forward.

include/uk/plat/lcpu.h
```diff
@@ -60,22 +53,28 @@ void ukplat_lcpu_disable_irq(void);
...
 void ukplat_lcpu_restore_irqf(unsigned long flags);
 
 /**
  * Checks if interrupts are disabled
+ *
  * @return non-zero value if interrupts are disabled
  */
 int ukplat_lcpu_irqs_disabled(void);
 
+/**
```
michpappas, line 75: Looks like it's been used in lib/ukschedcoop/schedcoop.c

marcrittinghaus: Yes, but it seems the functionality to look for pending IRQs is only used in Xen. If I remember correctly I checked the code and thought this could also be done differently without having to introduce a public API that makes no sense for other platforms. But I would leave this as TODO for a future PR.

marcrittinghaus:
> @marcrittinghaus understood, thanks for that. For the current PR it should be fine, let's go ahead with it. Yet, this approach feels like we are migrating the problem to the implementation, rather than addressing it on its core. If the implementation is at the platform level one could argue that platforms should be free to depend on nolibc or whatever else implements these error codes, but the same thing cannot be claimed for implementations under arch/ or lib/. Do we have a plan on how to proceed with the implementation here?
> 
> There is another thing, specific to `errno.h`: If the API returns errno.h values and calls into components that do not use `errno.h`, we lose information. The API would have to guess or translate error codes into `errno.h`. Linux and BSDs use `errno` codes universally at the kernel level, without meaning that their kernels are dependant on libc 😆 It's just that the error codes match those with libc's.
> 
> Anyway, I think that this needs to be discussed in the broader scope of such dependencies and establish a more formal approach on how to move forward.

I agree. I am not really happy with this either. Maybe we should include this topic in our next meeting. I would also vote for introducing a `uk/errno.h` that should universally be used to express error conditions. I think Simon just sees this already fulfilled with having the header in nolibc and accepting that if you want to interpret the codes you have to include nolibc. But he also agreed that this is not a perfect solution. For this PR, I think it is good, that we already have the specification for `errno` codes in.

marcrittinghaus:
@cristian-vijelie 
Thanks for updating. Perhaps we should exclude `spinlock.h` from the PR, because the spinlock PR (#289) already updates the code to `CONFIG_HAVE_SMP` and we will run into a merge conflict otherwise.

cristian-vijelie:
> @cristian-vijelie Thanks for updating. Perhaps we should exclude `spinlock.h` from the PR, because the spinlock PR (#289) already updates the code to `CONFIG_HAVE_SMP` and we will run into a merge conflict otherwise.

Done

plat/Config.uk
```diff
@@ -12,6 +12,15 @@ config EARLY_PRINT_PL011_UART_ADDR
 	help
 		Pl011 serial address used by early debug console.
 
+config UKPLAT_LCPU_MAXCOUNT
+        int "Maximum number of supported logical CPUs"
+        range 1 256
+        default 1
+
+config HAVE_SMP
```
michpappas, line 20: Is this option intentionally invisible?

marcrittinghaus: Yes, the option is enabled automatically if you set LCPU_MAXCOUNT to a value > 1. Then it is in line with the other `HAVE_` options but without having to activate SMP first and then select the number of supported CPUs.

michpappas: Ah okay. The truth is that I find it a bit against my intuition :) If I was trying to enable SMP I would look for an explicit option similarly to Linux (Processor type and features -> Symmetric multi-processing support). Btw, would it make more sense to have `LCPU_MAXOUNT` as a runtime parameter? That would allow one to test and release a single image that by default starts all cores, but also allows applying a limit on a per-deployment basis.

marcrittinghaus: ok, I see. This comes from the fact that in UK the `HAVE_` variables have always been hidden in kconfig (see HAVE_LIBC etc.). So when Razvan asked how to do it, Simon came up with this solution and I liked it.

Having `LCPU_MAXCOUNT` as runtime parameter is not possible because it is used to statically scale the CPUs array in the SMP implementations. Of course you can add a runtime parameter that restricts the number of CPUs actually initialized, thus the name `MAXCOUNT` (i.e., you are not forced to start all). Btw: For Unikraft, we generally should prefer compile-time constants vs. runtime parameters where possible, so compile often and specialize as much as possible.

michpappas: Hmmm okay. On the one hand I understand the motivation. One the other hand, this has quite some implications in corporate environments, as the bigger the organization the higher the cost a release comes with. As such, people don't want to have separate releases for different configuration variants. Anyway, as long as everyone is aware of the potential implications it's just one choice over another. For `LCPU_MAXCOUNT` in particularly it's not really important.

Having said the above,
```
```

justin-he:
Based on this pr #309, I wrote the draft version to support smp boot on arm64 kvm plat.
**Please feel free to add/delete/update my patches.**

justin-he:
[0006-plat-kvm-smp-Add-psci-call-psci_cpu_on.txt](https://github.com/unikraft/unikraft/files/7535261/0006-plat-kvm-smp-Add-psci-call-psci_cpu_on.txt)
[0007-plat-kvm-arm64-Add-a-parameter-to-start_mmu.txt](https://github.com/unikraft/unikraft/files/7535262/0007-plat-kvm-arm64-Add-a-parameter-to-start_mmu.txt)
[0008-plat-kvm-arm64-Add-entry-point-for-APs-secondary-cpu.txt](https://github.com/unikraft/unikraft/files/7535264/0008-plat-kvm-arm64-Add-entry-point-for-APs-secondary-cpu.txt)
[0009-plat-kvm-arm-Implement-smp-boot-on-arm64-kvm-plat.txt](https://github.com/unikraft/unikraft/files/7535265/0009-plat-kvm-arm-Implement-smp-boot-on-arm64-kvm-plat.txt)
[0010-plat-kvm-arm-Add-simple-percpu-variable-support.txt](https://github.com/unikraft/unikraft/files/7535266/0010-plat-kvm-arm-Add-simple-percpu-variable-support.txt)
[0011-plat-kvm-arm-Enable-scheduler-on-secondary-APs.txt](https://github.com/unikraft/unikraft/files/7535267/0011-plat-kvm-arm-Enable-scheduler-on-secondary-APs.txt)
[0005-plat-kvm-smp-Add-smp-definitions-in-smp.h.txt](https://github.com/unikraft/unikraft/files/7535268/0005-plat-kvm-smp-Add-smp-definitions-in-smp.h.txt)

plat/Config.uk
marcrittinghaus: Hi @michpappas,

we want to have this in for the 0.6 release next week. Everything is set up, so you can now approve the PR and add your Reviewed-by tag 😃 Thanks!

michpappas (approved):

plat/Config.uk
michpappas: Hey @marcrittinghaus, done thanks a lot!

marcrittinghaus (approved):
Looks good. Thanks everyone 👍🏻 

PR #314, opened 2021-10-30
lib/uksignal: Register system call from uksignal

razvanvirtan:
Register the system calls implemented in the uksignal library. The signal
library provides support for the standard signals and do not yet support
real time signals. The `rt_sigtimedwait` system call does not yet support
time out.
Also, remove the uksignal LIBSYSCALL_SHIM dependency.

Signed-off-by: Răzvan Vîrtan <virtanrazvan@gmail.com>
Signed-off-by: Sharan Santhanam <sharan.santhanam@neclab.eu>


mogasergiu (approved):
Looks good to me.

mogasergiu (approved):
Looks good to me.

lib/uksignal/signal.c
```diff
@@ -141,30 +142,18 @@ sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
...
-sighandler_t signal(int signum, sighandler_t handler)
+UK_SYSCALL_R_DEFINE(int, rt_sigprocmask,
+		    int, how,
+		    const sigset_t *, set,
+		    sigset_t *, oldset,
+		    size_t, sigsetsize)
 {
-	/* SA_RESTART <- BSD signal semantics */
-	return uk_signal(signum, handler, SA_RESTART);
+	return  uk_thread_sigmask(how, set, oldset);
```
marcrittinghaus:
```suggestion
	return uk_thread_sigmask(how, set, oldset);
```

lib/uksignal/exportsyms.uk
```diff
@@ -9,13 +9,40 @@ uk_thread_sigmask
 # signal.h
 sigaction
 signal
+rt_sigaction
+uk_syscall_e_rt_sigaction
+uk_syscall_r_rt_sigaction
+
+
```
marcrittinghaus: Look for proper spacing in the files. Here is one empty line too many. It is the same in line 36

razvand:
@marcrittinghaus , see if this is OK with you and approve it.

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
...
+	if (rt_sigaction(signum, &act, &old, (_NSIG / 8)) < 0)
+		return SIG_ERR;
+
+	if (old.sa_flags & SA_SIGINFO)
+		return NULL;
+	else
+		return old.sa_handler;
+}
+
+#ifdef CONFIG_LIBUKSIGNAL_LIBCFNS_ENABLE
```
marcrittinghaus: We already have a config option: `UK_LIBC_SYSCALLS`, see http://docs.unikraft.org/developers-app.html#syscall-shim-layer. Please adopt the outlined scheme:

```C
UK_SYSCALL_R_DEFINE(int, rt_sigaction, ...)
{
   /* Implement syscall */
}

#if UK_LIBC_SYSCALLS
int sigaction(...)
{
      return rt_sigaction(...)
}
#endif /* UK_LIBC_SYSCALLS */
```

For a more stringent version, you can also use `UK_LLSYSCALL_R_DEFINE` and call `uk_syscall_e_rt_sigaction`. This is a bit cleaner from the perspective that `rt_sigaction` is not a defined libc function and this would prevent its definition. But I am also fine with keeping `UK_SYSCALL_R_DEFINE` because for calling the low-level version all parameters must be cast to `long` which is not so nice.

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
...
+sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
+{
+	int error;
+
+	error = rt_sigaction(signum, act, oldact, (_NSIG / 8));
+	if (error < 0) {
+		errno = -error;
+		error = -1;
+	}
+	return error;
```
marcrittinghaus: The `rt_sigaction`/`uk_syscall_e_rt_sigaction` already takes care of setting `errno` and the return value as part of the `UK_SYSCALL_R_DEFINE` macro. So this additional code actually breaks the return values everywhere!

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
...
+}
+
+sighandler_t signal(int signum, sighandler_t handler)
+{
+	/* SA_RESTART <- BSD signal semantics */
+	return uk_signal(signum, handler, SA_RESTART);
+}
```
marcrittinghaus: Since `uk_signal` is not used elsewhere, please put its code here and remove the function. This is cleaner and also otherwise this gives a warning when libc declarations are deactivated as `uk_signal` is unused then.

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
...
+
+	return error;
+}
+
+int sigprocmask(int how, const sigset_t *set, sigset_t *oldset)
+{
+	int error;
+	int ret = 0;
+
+	error = uk_thread_sigmask(how, set, oldset);
```
marcrittinghaus: You must not call the function that actually implements the syscall here because that would prevent the system call hook to be triggered, for example, when logging system calls. So you have to call `rt_sigprocmask`/`uk_syscall_e_rt_sigprocmask` here.

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
...
+	}
+	return error;
+}
+
+int sigwait(const sigset_t *set, int *sig)
+{
+	int error;
+	siginfo_t si;
+
+	error = rt_sigtimedwait(set, &si, 0, (_NSIG / 8));
```
marcrittinghaus: Use `NULL` instead of `0` for pointer values everywhere

lib/nolibc/musl-imported/include/signal.h
```diff
@@ -8,6 +8,15 @@
...
 
+#ifndef __NEED_time_t
+#define __NEED_time_t
+#endif
+#ifndef __NEED_suseconds_t
+#define __NEED_suseconds_t
+#endif
+#ifndef __NEED_struct_timespec
+#define __NEED_struct_timespec
+#endif
```
marcrittinghaus: As these definitions are not used by any of the public functions, these definitions should be removed from the `signal.h`. Also `time_t` and `suseconds_t` is not actually used even internally.

So place something like this directly in `signal.c` in the include section:
```C
#ifndef __NEED_struct_timespec
#define __NEED_struct_timespec
#endif
#include <sys/types.h>
```

lib/uksignal/Config.uk
```diff
@@ -6,3 +6,11 @@ menuconfig LIBUKSIGNAL
 	select LIBUKALLOC
 	select LIBUKSCHED
 	select LIBPOSIX_PROCESS
+	select LIBSYSCALL_SHIM
+	select LIBSYSCALL_SHIM_LIBCSTUBS
+
+if LIBUKSIGNAL
+	config LIBUKSIGNAL_LIBCFNS_ENABLE
```
marcrittinghaus: This setting should be removed. See below.

lib/uksignal/Config.uk
```diff
@@ -6,3 +6,11 @@ menuconfig LIBUKSIGNAL
 	select LIBUKALLOC
 	select LIBUKSCHED
 	select LIBPOSIX_PROCESS
+	select LIBSYSCALL_SHIM
+	select LIBSYSCALL_SHIM_LIBCSTUBS
```
marcrittinghaus: `LIBSYSCALL_SHIM_LIBCSTUBS` is not required and should be removed.

`LIBSYSCALL_SHIM` can also be omitted. For this please replace the `uk_syscall_r_getpid()` call in `killpg` with 
`-pgrp` as suggested by `man killpg`.

lib/uksignal/signal.c
```diff
@@ -297,14 +286,129 @@ int sigwait(const sigset_t *set, int *sig)
 	ptr->wait.status = UK_SIG_NOT_WAITING;
 
 	/* do not execute handler, set received signal */
-	*sig = ptr->wait.received_signal.si_signo;
+	memcpy(info, &ptr->wait.received_signal,
+	       sizeof(*info));
```
marcrittinghaus:
```suggestion
	*info = ptr->wait.received_signal;
```

Then you can remove the `include <string.h>`.

lib/uksignal/exportsyms.uk
```diff
@@ -9,13 +9,40 @@ uk_thread_sigmask
...
+uk_syscall_e_rt_sigpending
+uk_syscall_r_rt_sigpending
+
 sigwait
+rt_sigtimedwait
+uk_syscall_e_rt_sigtimedwait
+uk_syscall_r_rt_sigtimedwait
+
+
 kill
```
marcrittinghaus, line 36: Is there a reason why `kill` is not exposed as system call?

lib/uksignal/exportsyms.uk
```diff
@@ -9,13 +9,40 @@ uk_thread_sigmask
...
 killpg
+
 raise
+tkill
+uk_syscall_e_tkill
+uk_syscall_r_tkill
+
```
marcrittinghaus, line 43: Please group these with the other kill calls

marcrittinghaus (changes requested):
Hi @razvanvirtan ,

thank you for the PR. Before we can merge this PR we need to do some changes. Please see the comments.

Also please take care that the individual commits are in good shape. 
* For example, commit 83e5a514ba breaks the definition of `sigpending` which is then repaired in another commit. 
* Remove all changes that are undone anyways like the `__sigaction`-style definitions. Commits should always be as close to the final code as possible and not reflect the development steps.
* Take care that a commit does not do more than what its description and title suggests.

Do you think you can make this until the release tomorrow?

marcrittinghaus:
@mogasergiu 
Please also take a look at the comments when you have time to get an idea of additional things to consider when reviewing 😃 PR quality is still very poor and it should not have gotten an approval.

razvanvirtan:
@marcrittinghaus This is a rebase of an older PR, maybe it would  be better to close it and create a new one? I will try to do this in the next few days, so that we don't postpone it a lot after the release, but also make sure we don't add low quality code into the upstream.

marcrittinghaus:
Hi @razvanvirtan,

As we have the whole discussion here, I would keep the PR open and just do a force push.

lib/uksignal/signal.c
```diff
@@ -360,7 +426,7 @@ int killpg(int pgrp, int sig)
 		return -1;
 	}
 
-	return kill(uk_syscall_r_getpid(), sig);
+	return kill(-pgrp, sig);
 }
 
 int raise(int sig)
```
mogasergiu: This definition will conflict with the one above if libc wrappers are enabled. I guess you could remove this one here.

lib/uksignal/signal.c
```diff
@@ -297,20 +339,44 @@ int sigwait(const sigset_t *set, int *sig)
...
 	uk_sig_handle_signals();
 
 	return 0; /* returns positive errno */
 }
 
+#if UK_LIBC_SYSCALLS
+int sigwait(const sigset_t *set, int *sig)
+{
+	int error;
+
```
mogasergiu, line 354: What `si`? It does not seem defined to me.
```suggestion
        siginfo_t si;
```

lib/uksignal/signal.c
```diff
@@ -297,20 +339,44 @@ int sigwait(const sigset_t *set, int *sig)
...
+	error = rt_sigtimedwait(set, &si, NULL, (_NSIG / 8));
+	*sig = si.si_signo;
+
+	return error;
+}
+#endif
+
+UK_SYSCALL_R_DEFINE(int, tkill, pid_t __unused, pid, int, sig)
```
mogasergiu: This issues a big preproccessing error on my end. For whatever reason, it does not like the name `tkill`. If I were to place some random name here, say `gkill`, that error does not occur anymore. Is this just on my end?

mogasergiu:
By the way, I think that, conventionally, there should be two separate `Signed-off-by` tags instead of one with the authors being separated by a comma.

lib/uksignal/signal.c
razvanvirtan: Apparently the tkill symbol was not added to UK_PROVIDED_SYSCALLS, sorry for this, I think something was messed up during the rebase.

lib/uksignal/signal.c
```diff
@@ -297,20 +339,45 @@ int sigwait(const sigset_t *set, int *sig)
...
+}
+#endif
+
 /*
  * Search for a thread that does not have the signal blocked
  * If all of the threads have the signal blocked, add it to process
  * pending signals
  */
-int kill(pid_t pid, int sig)
+UK_SYSCALL_R_DEFINE(int, kill, pid_t, pid, int, sig)
```
marcrittinghaus: This function still sets `errno` in lines 398 and 403. This should be:
```C
	if (pid != 1 && pid != 0 && pid != -1)
		return -ESRCH;

	if (!uk_sig_is_valid(sig))
		return -EINVAL;
```

You can also remove the extra empty line 396.

lib/uksignal/signal.c
```diff
@@ -176,12 +186,33 @@ int sigpending(sigset_t *set)
...
 
-int sigsuspend(const sigset_t *mask)
+#if UK_LIBC_SYSCALLS
+int sigprocmask(int how, const sigset_t *set, sigset_t *oldset)
+{
+	return rt_sigprocmask(how, set, oldset, (_NSIG / 8));
+}
+#endif
+
+
```
marcrittinghaus, line 211: Here is an empty line too many.

lib/uksignal/signal.c
```diff
@@ -141,30 +147,34 @@ sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
...
-	if (sigaction(signum, &act, &old) < 0)
+	if (rt_sigaction(signum, &act, &old, (_NSIG / 8)) < 0)
 		return SIG_ERR;
 
 	if (old.sa_flags & SA_SIGINFO)
 		return NULL;
 	else
 		return old.sa_handler;
 }
+#endif
```
marcrittinghaus: Please always provide the macro you tested as comment with the `endif`
```suggestion
#endif /* UK_LIBC_SYSCALLS */
```

marcrittinghaus (commented):
Hi @razvanvirtan, thanks for updating the PR! I have some more feedback 😏

Do you have some application or code snipped for testing?

razvanvirtan:
@marcrittinghaus I have updated according to your comments. Here you can find a test source file, I changed it so that it shows a `PASSED` message for each working syscall. The `rt_sigaction + rt_kill` scenario is the only one that doesn't show a message, I think this happens because of the `TO DO` item in the code. The output I am getting right now is the following:
```
rt_sigaction + tkill PASSED
rt_sigpending PASSED
rt_sigsuspend PASSED
rt_sigtimedwait PASSED
```
[Test_signal.zip](https://github.com/unikraft/unikraft/files/9354927/Test_signal.zip)

lib/uksignal/signal.c
```diff
@@ -297,20 +338,45 @@ int sigwait(const sigset_t *set, int *sig)
...
+	siginfo_t si;
+
+	error = rt_sigtimedwait(set, &si, NULL, (_NSIG / 8));
+	*sig = si.si_signo;
+
+	return error;
+}
+#endif /* UK_LIBC_SYSCALLS */
+
+UK_SYSCALL_R_DEFINE(int, tkill, pid_t __unused, pid, int, sig)
```
marcrittinghaus: I recognized that the signature is wrong. It should be:
```suggestion
UK_SYSCALL_R_DEFINE(int, tkill, int __unused, tid, int, sig)
```

marcrittinghaus (commented):
Hey @razvanvirtan,

I just recognized that the signature for `tkill` is wrong. Otherwise, thanks for putting all the work in! I have tested the code with the `main.c` that you supplied (thanks!) and I get the same output. I also tested the code without the PR and it is basically the same. So it is not induced by the PR 👍🏻

As signals break with the `musl` changes, I guess we have to update the code anyways and can then fix this broken test.

razvanvirtan:
@marcrittinghaus Thank you for looking into this! Shall we then fix this today or after the release? (Indeed, I haven't tested yet with the musl changes)

marcrittinghaus:
> @marcrittinghaus Thank you for looking into this! Shall we then fix this today or after the release? (Indeed, I haven't tested yet with the musl changes)

Please fix the signature today. Everything else, we do after the release. It would also be outside of the scope of this PR.

Actually, I have not tested with musl either, but I know from Simon that signals break. So there will probably be changes necessary.

razvanvirtan:
@marcrittinghaus Done, I have updated the header.

marcrittinghaus (approved):
@razvanvirtan

Great thanks! Looks good then 😃👍🏻

razvanvirtan:
@marcrittinghaus I did the rebase to remove the merge conflicts. Also updated a bit the commit messages. Now, some commits are not only registering a new syscall, but also unregister other syscalls, that become libc wrappers.

razvanvirtan:
Last update: changed commit message for [3d7c820](https://github.com/unikraft/unikraft/pull/314/commits/3d7c820e20726d375012e0e9cb48f263ff852c11).

marcrittinghaus (approved):
@razvanvirtan Thank you very much! Tested it again, all fine 😃👍🏻

PR #326, opened 2021-11-08
lib/uksignal: Update struct sigset_t and macros to 1024bits and solve compilation errors

dragosargint:

### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]

### Additional configuration

Include `uksignal` from menuconfig.
### Description of changes

As newlib uses a 64bits mask for signals and musl uses 128bits mask we need a unified way for those to work
with the underlying signal processing. This is what this PR intends to do.

This PR is part of a group that aims to solve compiling errors for both musl and newlib:
* https://github.com/unikraft/lib-newlib/pull/15
* https://github.com/unikraft/lib-musl/pull/4
* https://github.com/unikraft/unikraft/pull/327
* https://github.com/unikraft/unikraft/pull/326

### DIFF
--- a/lib/uksignal/include/uk/uk_signal.h
+++ b/lib/uksignal/include/uk/uk_signal.h
@@ -45,6 +45,16 @@
 extern "C" {
 #endif
 
+#ifndef is_sig_dfl
+#define is_sig_dfl(ptr) \
+	(!((ptr)->sa_flags & SA_SIGINFO) && (ptr)->sa_handler == SIG_DFL)
+#endif
+
+#ifndef is_sig_ign
+#define is_sig_ign(ptr) \
+	(!((ptr)->sa_flags & SA_SIGINFO) && (ptr)->sa_handler == SIG_IGN)
+#endif
+
 #define _UK_TH_SIG uk_crr_thread_sig_container()
 
 struct uk_thread;
--- a/lib/uksignal/include/uk/bits/sigset.h
+++ b/lib/uksignal/include/uk/bits/sigset.h
@@ -43,16 +43,16 @@ extern "C" {
 
 /* TODO: do we have gnu statement expression?  */
 /* internal use */
-#define uk_sigemptyset(ptr)	(*(ptr) = 0)
-#define uk_sigfillset(ptr)  (*(ptr) = ~((__sigset_t) 0))
-#define uk_sigaddset(ptr, signo) (*(ptr) |= (1 << ((signo) - 1)))
-#define uk_sigdelset(ptr, signo) (*(ptr) &= ~(1 << ((signo) - 1)))
-#define uk_sigcopyset(ptr1, ptr2) (*(ptr1) = *(ptr2))
-#define uk_sigandset(ptr1, ptr2)  (*(ptr1) &= *(ptr2))
-#define uk_sigorset(ptr1, ptr2)	  (*(ptr1) |= *(ptr2))
-#define uk_sigreverseset(ptr)	  (*(ptr) = ~(*(ptr)))
-#define uk_sigismember(ptr, signo) (*(ptr) & (1 << ((signo) - 1)))
-#define uk_sigisempty(ptr) (*(ptr) == 0)
+#define uk_sigemptyset(ptr) (((ptr)->__bits[0]) = 0)
+#define uk_sigfillset(ptr) (((ptr)->__bits[0]) = ~(0))
+#define uk_sigaddset(ptr, signo) (((ptr)->__bits[0]) |= (1 << ((signo) - 1)))
+#define uk_sigdelset(ptr, signo) (((ptr)->__bits[0]) &= ~(1 << ((signo) - 1)))
+#define uk_sigcopyset(ptr1, ptr2) (((ptr1)->__bits[0]) = ((ptr2)->__bits[0]))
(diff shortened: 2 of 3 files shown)


craciunoiuc (changes requested):
Hey @dragosargint,

Please update your review as discussed to switch from 128bits to 1024bits.

Cezar

dragosargint:
> Hey @dragosargint,
> 
> Please update your review as discussed to switch from 128bits to 1024bits.
> 
> Cezar

Hi, @craciunoiuc,

The signal mask in both musl and glibc does have 1024 bits, but this is done for signals that can be added later in Linux. In reality, this mask needs only 64 bits, equivalent to the 62 signals we have on Linux (signo 32, and 33 do not exist). On the other hand, newlib uses only a 64-bit mask. So this PR aims to provide support for both musl and newlib.

craciunoiuc:
Hey @dragosargint,

All good now. I tested the PR and it works.

Reviewed-By: Cezar Craciunoiu <cezar.craciunoiu@gmail.com>

craciunoiuc (approved):
Hey @dragosargint,

All good now. I tested the PR and it works.

Reviewed-By: Cezar Craciunoiu <cezar.craciunoiu@gmail.com>

marcrittinghaus:
@craciunoiuc, @dragosargint
Please ensure that the commit messages contain a clear description of the problem and the solution.

Then I am not sure I understand this correctly. Is there a PR for `lib-newlib` and `nolibc`? According to [signal.h:L63](https://github.com/unikraft/lib-newlib/blob/286d5b8ae7c4b9d14e9bf269389e2d37ef60bcc5/musl-imported/include/signal.h#L63), our version of `lib-newlib` defines `sigset_t` as `unsigned long`. Obviously, the new macros won't work with this, but only with the definition from `musl`. The same is true for `nolibc` [signal.h:L63](https://github.com/unikraft/unikraft/blob/ec8882b8e34346fd890c04d15257316895b7de21/lib/nolibc/musl-imported/include/signal.h#L63).

craciunoiuc:
Hey @marcrittinghaus,

Thanks for the comment suggestion. @dragosargint please also add the explanation from this discussion to the commit description.

Yes, there is a separate PR for updating the structure redefinition (https://github.com/unikraft/unikraft/pull/327).
I'm not 100% sure how the macros will behave with `newlib` as I only looked at them from the `musl`/`nolibc` perspective.

Let's see the perspective of @dragosargint.

marcrittinghaus:
This won't work with `newlib`, I think. 

@dragosargint Please have a look and publish a PR similar to `nolibc` for `lib-newlib`, if necessary. Otherwise, we'll break builds with `lib-newlib` with this PR.

dragosargint:
Hi @craciunoiuc, @marcrittinghaus. First of all, I would like to mention that these PRs [327](https://github.com/unikraft/unikraft/pull/327) and [326](https://github.com/unikraft/unikraft/pull/326) were made initially by Bernard and I took them over. So there are things that I missed too. In my work with musl, I only used [326](https://github.com/unikraft/unikraft/pull/326). The reason why this PR is necessary is that when you try to compile an application that uses both uksignal and musl some compilation errors are encountered. These can be seen in the photo attached [1]. This is due to the `sigset_t` structure which is defined in musl as follows: 
```
typedef struct __sigset_t { unsigned long __bits[128/sizeof(long)]; } sigset_t;
```
However, in `uksignal` the macros that are used are not suitable for this structure. Hence the changes from this PR.
On the other hand, `newlib` does indeed define `sigset_t` as an `unsigned long` and this PR will break compilation for it (I didn't know that last week when I made the changes to this PR). I tried for a few hours to make a change like PR [327](https://github.com/unikraft/unikraft/pull/327)  for `newlib`, but unfortunately, this is not as straight forward and it still gives some compilation errors[2]. So I will need more time to look into it. 
[1]
image
[2]
image

craciunoiuc (changes requested):
Hey @marcrittinghaus,

I tested the PR and it worked just fine with the others (newlib/nolibc/musl)

@dragosargint please reformat the commit message adding the required `selector` and some additional description

dragosargint:
Done!

craciunoiuc (approved):
Hey @marcrittinghaus @dragosargint,

I think the PR looks good now. I hope I didn't miss anything.

lib/uksignal/include/uk/uk_signal.h
```diff
@@ -45,6 +45,16 @@
...
+#ifndef is_sig_dfl
+#define is_sig_dfl(ptr) \
+	(!((ptr)->sa_flags & SA_SIGINFO) && (ptr)->sa_handler == SIG_DFL)
+#endif
+
+#ifndef is_sig_ign
+#define is_sig_ign(ptr) \
+	(!((ptr)->sa_flags & SA_SIGINFO) && (ptr)->sa_handler == SIG_IGN)
+#endif
+
```
marcrittinghaus, line 57: Hi @dragosargint

These macros are only used within `lib/uksignal` and are not part of the official libc headers. I thus agree that it makes sense to define them here. Could you then also remove the definitions in `lib-newlib` and `nolibc` so we have the same source of definition no matter what libc we are using? Then you can also remove the `#ifndef` barriers.

marcrittinghaus (changes requested):
Hi @dragosargint,

looks good so far. I have one point just to keep definitions clean and to not create error-prone and confusing duplicates.

PR #174, opened 2021-03-29
More system calls

skuenzer:
Registers a number of system calls. Work was done towards EuroSys'21 submission by Constantin Raducanu and myself. This series depends on PR #172.

Issue #199, opened 2021-05-03
Clean up pre-rigorous use of checkpatch errors and warnings


nderjung:
Before the switch to Github, all code which was upstreamed was manually checked with [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl).  This meant that some patches/commits may have been run against checkpatch, but we cannot be sure for certain as there is no explicit record if it was checked or not.  In contrast, PRs coming into Github now are run against `checkpatch.pl` and the PR ID is referenced in the commit series, so we can now check.

Running `checkpatch.pl` in "file mode" on every file in the `lib/` directory of the Unikraft core yields **331 errors** and 1073 warnings.  Whilst some of these errors and warnings can be ignored, I'm sure there are some which should be fixed which were introduced before we started rigorously checking all new code.

To check all files in the `lib/` directory, I used:
```bash
for F in $(tree --noreport -fi ./lib); do ./support/scripts/checkpatch.pl --no-summary --terse -f $F; done
```

An additional thought is to introduce in-line code which downgrades errors, in some sort of comment syntax, e.g.:
```diff
diff --git a/lib/vfscore/task.c b/lib/vfscore/task.c
index 20c474c..b89f389 100644
--- a/lib/vfscore/task.c
+++ b/lib/vfscore/task.c
@@ -119,6 +119,7 @@ path_conv(char *wd, const char *cpath, char *full)
        }
        *tgt = '\0';

+       /* Checkpatch-WarnOnly: RETURN_PARENTHESES */
        return (0);
 }
 ```
 
This is useful for anyone who inspects the code to understand that the syntax is non-standard.  Additional reasoning inline would also be preferred.

skuenzer:
Please note that we have to exclude some items since they are imported and will be updated from time to time. Out of my head, I remember the following:
* lib/fdt
* Official Xen API headers in plat/xen/include/xen

Commits should also be done on a per library basis so that fixes can be individually integrated. This would be useful for the case where we have too many outstanding PRs that we would like to include before. Otherwise we risk that a single syntax commit would break too many older PRs because they would not apply anymore.

PR #329, opened 2021-11-10
lib/vfscore: Fix `pread64` and `pwrite64` syscall implementations

skuenzer:
### Description of changes

PR #174 caused a compilation error for vfscore because of undefined symbols `uk_syscall_r_pwritev` and `uk_syscall_r_preadv`. This PR fixes this issues and closes two other compiler warnings caused by `lib/vfscore`.

### DIFF
--- a/lib/vfscore/mount.c
+++ b/lib/vfscore/mount.c
@@ -377,12 +377,14 @@ UK_LLSYSCALL_R_DEFINE(int, sync)
 	bio_sync();
 #endif
 	uk_mutex_unlock(&mount_lock);
+
+	return 0;
 }
 
 #if UK_LIBC_SYSCALLS
 void sync(void)
 {
-   uk_syscall_e_sync();
+	uk_syscall_e_sync();
 }
 #endif
 
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -782,7 +782,7 @@ int ioctl(int fd, unsigned long int request, ...)
 	arg = va_arg(ap, void*);
 	va_end(ap);
 
-	return uk_syscall_e_ioctl(fd, request, arg);
+	return uk_syscall_e_ioctl((long) fd, (long) request, (long) arg);
 }
 #endif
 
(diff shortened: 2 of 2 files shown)


lib/vfscore/mount.c
```diff
@@ -377,12 +377,14 @@ UK_LLSYSCALL_R_DEFINE(int, sync)
...
 	uk_mutex_unlock(&mount_lock);
+
+	return 0;
 }
 
 #if UK_LIBC_SYSCALLS
 void sync(void)
 {
-   uk_syscall_e_sync();
+	uk_syscall_e_sync();
```
nderjung, line 387: Hi @skuenzer, thanks for sorting this fix! It all looks good, but can you please remove this indentation change in `sync` from commit bab546d.

nderjung (changes requested):

lib/vfscore/mount.c
skuenzer: Hey, why :thinking:? Indentation should be tabs, this was originally wrong. This one liner is also such a minimal change so that I think it is okay to have it as part of the `sync` fix, especially since it is related. 😄

marcrittinghaus: Agree with @skuenzer. Don't think this change would justify a dedicated PR, but it should be corrected. So this seems to be a good opportunity, although I understand @nderjung that it is not strictly part of the originally addressed problem.

nderjung: I am very [keen on fixing stylistic inconsistencies](https://github.com/unikraft/unikraft/issues/199) 😀 However, in my opinion, the approach should not be to simultaneously fix a separate stylistic error in a file where the commit occurs, as this undocumented and separate change will disrupt `git-blame`.  Unless the commit message also indicates the change or the main diff happens to touch that line and fix the style, then i think it should be a separate commit entirely. 😄

skuenzer: We should probably start deciding and document how we want to handle minor style fixes within a related patch. As compromise, I added a comment in the patch description about this style fix.

nderjung (approved):
Great, thanks a lot @skuenzer :)

PR #333, opened 2021-11-14
[DRAFT] plat/drivers/virtio-blk: Remove ukalloc usage for virtio_blkdev_requests

jschlumpp:
The virtio_blkdev_queue_dequeue function can be called in an interrupt
handler context and calling `free` there can corrupt allocators such as
ukallocbbuddy.

Signed-off-by: Marco Schlumpp <marco.schlumpp@gmail.com>

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This allocates a chunk of memory on the queue initialization for a ring buffer. The relevant allocate/free is then replace by a simple queue operation, which does not cause issues if called from an interrupt context.

While atomic operations would be more appropriate in this case to prevent the compiler from reordering/splitting the necessary operations, unikraft does not offer them, yet.

### DIFF
--- a/plat/drivers/virtio/virtio_blk.c
+++ b/plat/drivers/virtio/virtio_blk.c
@@ -102,6 +102,12 @@ struct uk_blkdev_queue {
 	/* The scatter list and its associated fragments */
 	struct uk_sglist sg;
 	struct uk_sglist_seg *sgsegs;
+
+	/* ring buffer for requests */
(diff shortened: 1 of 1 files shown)


jschlumpp:
The completion order of requests can be different from the submission order. Therefore, this approach is broken as it assumes the order matches.
I will create another pull request with a fixed implementation.

PR #338, opened 2021-11-22
Virtual Memory API (x86_64, kvm)

### LINKED ISSUE #225: increase physical address space or available memory for unikernel?
I have an application I'm attempting to port into a Unikraft project. One issue I'm experiencing is that the physical address space is only 1GB (up to 0x40000000 I believe due to that being the limit for QEMU pc emulator), but my application needs to make allocations that exceed this amount. Using the ptsupport patch from the eurosys21 repository, all `MULTIBOOT_MEMORY_AVAILABLE` regions are used as available physical memory for a page table. Due to how physical memory is laid out [0], the RAM that is allocated to the machine from the hypervisor has a hole around the 3GB mark. The patch does not resize `_libkvmplat_cfg.heap` after the first memory area is mapped in the page table, and so allocators are only initialized with the same amount of memory as the first available multiboot memory area which is ~3GB in size.

I assumed that not resizing `_libkvmplat_cfg.heap` was a mistake in the patch and that if I changed it to the same size as the total amount of available memory in the page table then it would allow my allocator to use all the memory I set in KVM. Instead I end up hitting a page fault when writing to an address just past the 3GB mark, which is where the physical page would have been located in the next section in RAM. What's odd is that the page fault only occurs when the kernel is executing, but when I change the same location in GDB it works just fine. GDB normally cannot read or write to unmapped regions in memory, so it's strange that I can change a value directory that would normally trigger a page fault during execution.

```
[    0.623086] CRIT: [libkvmplat] <traps.c @  123> Page fault at linear address 4fffffffc, rip 5c8c4d, regs 0xd3ff50, sp 7fffff00feb0, our_sp 0xd3ff28, code b
```
(at a breakpoint at 0x5c8c4d)
```
gef➤  set {int} 0x4fffffffc = 1
gef➤  x 0x4fffffffc
0x4fffffffc:	0x00000001
```

I'm assuming that there must be some hardware-level configuration missing in the patch here. Any help is appreciated. It should be easy to reproduce this by using a eurosys21 docker container with the ptsupport patch applied and running a unikernel with  more than 4.5GB of RAM and something like this within the app:
``` C
int main() {
    // allocates 4GB of memory
    for (long unsigned int i = 0; i < 0x400; ++i) {
        printf("allocating memory, total used: 0x%lx\n", i*0x400000);
        void* m = malloc(0x400000);
        if (m == NULL) {
            printf("failed allocation, total: 0x%lx\n", i*0x400000);
            exit(1);
        }
    }
    return 0;
}

```

[0] - https://wiki.osdev.org/Detecting_Memory_(x86)

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64` 
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

The PR introduces a configuration option under `Platform Configuration | Platform Interface Options` to enable virtual memory.

 - `CONFIG_PAGING=y`

This will automatically disable all platforms currently not supported (e.g., Xen) and also add a dependency on `ukfalloc` and `ukfallocbuddy`, the physical memory allocator libraries.

### Description of changes

Currently, Unikraft establishes a virtual address space using a fixed boot page table that maps the first 1 GiB of memory. This PR introduces the API for dynamically (un-)mapping virtual to physical addresses, setting virtual memory protections, and dynamically managing physical memory. This allows Unikraft to access all available RAM in the system and enables the implementation of system calls such as `mmap`. The frame allocator is a multi-zone buddy-based allocator that is optimized for power-of-two allocations, but can also handle arbitrary allocations without wasting memory like a conventional buddy allocator. Also allocations and frees do not have to match allowing to dynamically manage physical memory on a page (4K) granularity. See commit messages and documentation in the code for more details.

#### Basic (Un-)Mapping
Establish a read-write mapping of 16 MiB (= 4096 pages) at the virtual address `0x10000000000`:
```C
struct uk_pagetable *pt = ukplat_pt_get_active();
ukplat_page_map(pt, 0x10000000000, __PADDR_ANY, 4096, PAGE_PROT_RW, 0);
```
The call implicitly allocates the physical memory (due to `__PADDR_ANY`) and maps it to the given virtual address. The mapping will automatically use the largest page sizes possible according to the alignment, mapping size, available continuous physical memory, and page sizes supported by the architecture. On x86_64, for example, the call will create a mapping of 8 large pages (2 MiB).

Force a mapping of a particular page size:
```C
ukplat_page_map(pt, 0x10000000000, __PADDR_ANY, 8, PAGE_PROT_RW, PAGE_FLAG_SIZE(PAGE_LARGE_LEVEL) | PAGE_FLAG_FORCE_SIZE);
```

Unmap a range of memory (here 10 4K pages):
```C
ukplat_page_unmap(pt, 0x10000000000, 10, 0);
```
This will split the first large page created in the previous calls and unmap the first 10 small pages. This will also free the frames unless `PAGE_FLAG_KEEP_FRAMES` is specified. If page tables can be released during the operation this is done transparently (except `PAGE_FLAG_KEEP_PTES` is specified). Note that it is no problem to unmap physical address ranges previously not allocated by the underlying frame allocator (for example kernel code and data), even if `PAGE_FLAG_KEEP_FRAMES` is not specified.

#### Specific Physical Memory Mapping
Mapping a specific physical address can be done by using the physical address instead of `__PADDR_ANY`. However, this requires that the physical memory has been allocated manually by calling the frame allocator first:
```C
__paddr_t paddr = 0x40000000; /* Physical memory at 1 GiB, __PADDR_ANY for any physical address */
pt->fa->falloc(pt->fa, &paddr, 10); /* paddr receives allocated address if __PADDR_ANY */
ukplat_page_map(pt, 0x10000000000, paddr, 10, PAGE_PROT_RW, 0);
```

If the physical memory allocation should be restricted to a certain range (e.g., for DMA), this can be done like so:
```C
__paddr_t paddr;
pt->fa->falloc_from_range(pt->fa, &paddr, 10, 0, MIN, MAX);
```

#### Page Table Walking
Sometimes direct access to the page table entries (PTEs) is needed. For example to store custom information in user-available bits or set advanced caching flags.
```C
unsigned int level = PAGE_LEVEL; /* perform complete walk down to page level */
__vaddr_t pt_vaddr; /* OPTIONAL, receives virtual address where the page table is mapped that contains the PTE */
__pte_t pte; /* OPTIONAL, receives the PTE */
ukplat_pt_walk(pt, 0x10000000000, &level, &pt_vaddr, &pte);

/* Set some extended PTE flag */
pte |= MY_EXTENDED_FLAG;

/* Update the PTE in the page table */
ukarch_pte_write(pt_vaddr, level, PT_Lx_IDX(0x10000000000, level), pte); 
ukarch_flush_tlb_entry(0x10000000000);
```

If for some reason, the mapping should be removed without loosing these settings, this can be done by:
```C
ukplat_page_unmap(pt, 0x10000000000, 1, PAGE_FLAG_KEEP_PTES);
```
The `PAGE_FLAG_KEEP_PTES` flag preserves the PTEs (and page tables) so that the settings are still there when establishing a new mapping at the same virtual address (page protection is taken from the mapping call).

#### Creating/Destroying Page Tables
New pages tables are created by cloning an existing page table.
```C
struct uk_pagetable new_pt;
ukplat_pt_clone(&new_pt, pt, 0);

ukplat_pt_set_active(&new_pt); /* switch to new page table */
```
This will create a deep copy of the page table hierarchy including any mapped kernel code and data sections and set it as active page table. Note, the page table can be freely used with the other paging functions even if not active.

An empty address space (no kernel code and data!) can be created like so:
```C
struct uk_pagetable new_pt;
ukplat_pt_clone(&new_pt, pt, PAGE_FLAG_CLONE_NEW);
```

Freeing a page table will unmap all memory and free the entire page table hierarchy. Depending on if `PAGE_FLAG_KEEP_FRAMES` is specified, allocated physical memory for the mapped regions is preserved. Note that there is no reference counting in the frame allocator! ⚠️ 
```C
ukplat_pt_free(&new_pt, PAGE_FLAG_KEEP_FRAMES);
``` 

### Extending Architecture Support
Most of the code is already architecture and platform independent, only assuming a n-level forward page table. To extend the support, the following files have to be provided:

`arch/X/include/uk/asm/paging.h`
`plat/common/include/X/paging.h`


mogasergiu (approved):
Looks good to me!

marcrittinghaus:
Hey @michpappas @rene,

if you have time, I would be grateful if you could quickly look over the API and tell me if there is anything problematic or limiting for your purposes. I already incorporated the direct physical memory allocation as suggested by you! Thanks 😄

adinasm (approved):
Seems ok.

Reviewed-by: Adina Smeu <adina.smeu@gmail.com>

include/uk/plat/paging.h
```diff
@@ -0,0 +1,279 @@
...
+ * @return 0 on success, a non-zero value otherwise.
+ */
+int ukplat_pt_init(struct uk_pagetable *pt, __paddr_t start, __sz len);
+
+/**
+ * Adds a physical memory range to the frame allocator of the given page table.
+ * The physical memory is available to all page tables sharing the same frame
+ * allocator.
+ *
+ * @param start start of the physical address range that will be available
```
rene, line 131: @marcrittinghaus , If I understood correctly, I see some potential problem here. If the function can reserve (use) some of the memory area declared in the arguments to put its metadata, this will forbid us to declare specific chunks for DMA. For instance, in some platforms we might have not only one contiguous physical memory chunk to use, but we might need to declare several chunks because some memory in the middle might be reserved, or for example some specific chunks might be used only for DMA. For instance, let's say we have the 3 memory chunks (AB, BC, CD) illustrated below, and BC is used exclusively for DMA, which means that devices will access this chunk directly (unless we use SMMU for translation). When we declare BC it's suppose that the device should be able to use the whole BC physical range, so we cannot have any metadata inside. 

```
|    -----C1---| ---- reserved ---- | ----C2 DMA area ----- | ----------C3-------------- |
A              A1                   B                       C                            D
```

I don't have a better solution to propose, but you could consider the possibility to put metadata into a specific reserved memory inside kernel virtual address space.

Edit: I've added some reserved memory to illustrate better my explanation

rene:
> Hey @michpappas @rene,
> 
> if you have time, I would be grateful if you could quickly look over the API and tell me if there is anything problematic or limiting for your purposes. I already incorporated the direct physical memory allocation as suggested by you! Thanks smile

Thanks @marcrittinghaus , I will need more time to have a more deeply look, but I will post my comments on demand....

include/uk/plat/paging.h
marcrittinghaus: I see what you mean. Generally, the `addmem` functionality is not supposed to be used to exclude "short" holes. In fact, internally zones are searched linearly in certain circumstances, so there should be as few as possible. The scenario that you described could be solved by defining a zone from A all the way up to D and then immediately do an allocation for C2. This can be done like so:
```C
__paddr_t paddr = START_OF_C2;
pt->fa->falloc(pt->fa, &paddr, LENGTH_OF_C2_IN_FRAMES);
```
Afterwards the memory is never touched again and won't contain metadata. There are no restrictions on the alignment and length of C2.

The allocator itself accepts a pointer to an arbitrary virtual address range to put it's metadata into. It is the x86 helper code that decides to use some of the address range provided to `addmem` to give the allocator for metadata. So you could do it differently if the platform requires it. Only exception is that if the platform provides direct-mapped physical memory (i.e., physical memory is always mapped somewhere in the address space to a fixed offset). Then some of the metadata is placed in the free frames itself to save memory for metadata. For example, to have the range BC in the free list, the allocator might put a list entry at the beginning of BC. Of course, only as long as the memory range is not allocated.

So I think when you allocate the range ahead of time and just "forget" it before initializing the area for DMA it is pretty save to do it this way. It would only be a problem, if there must be a guarantee that there is no memory access between `addmem` and `falloc`. Nevertheless, for this case, I could easily add a config parameter that forbids to put metadata into the frames and always use the user-supplied buffer space, irrespective of the availability of direct-mapped physical memory.

marcrittinghaus:
Just done a bit polishing:
* Fixed missing parentheses around `vaddr` macro variable in `X86_VADDR_CANONICALIZE`
* Fixed header inclusion that was automatically done by IDE in `paging.c`
* Changed `PT_Lx_PTES` and `PAGE_Lx_SHIFT` so that they have to be compile-time resolvable macros
* Added handy `PT_PAGES()` macro in `include/uk/arch/paging.h` that calculates number of page table pages needed to map a certain number of data pages. Works at compile time.

michpappas:
Hi @marcrittinghaus, haven't finished looking at this PR, but I'll start with posting some comments.

Overall this is awesome work :) On the API, I only have one comment. In Arm, there is a distinction between Normal Memory and Device Memory. This information is stored in the corresponding PTE, along with other attributes that control shareability, caching, etc. The way currently provided by the API for setting this attribute is through `ukplat_pt_walk()` -> `ukarch_pte_write()` -> `ukarch_flush_tlb_entry()`. This needs to be done for every device declared in the device-tree. I'm not sure if it is worth to consider this sequence expensive, as I don't have numbers to back such claim, and usually there is a limited number of devices. But it would be nice to have an easier way to map device memory.

`pgarch_page_map()` provides the `template` parameter. From the x86 implementation it looks like that this parameter is intended to be used for the same purpose, ie setting attributes. Yet this parameter is not exposed to the API, but only used internally by the implementation.

Based on the above, I can see the following possibilities:
1. Leave it as is, considering that the current way is adequate, and not expensive. This is okay, as we can provide a macro in arm64 like `MREMAP_DEVICE()` that implements the `ukplat_pt_walk()` -> `ukarch_pte_write()` -> `ukarch_flush_tlb_entry()` sequence.
2. Generalize the protections parameter to contain more attributes, eg `PAGE_ATTR_*`. Protections will be a subset of this parameter (eg `PAGE_ATTR_PROT_*`), along with additional architecture-specific protections (like `PAGE_ATTR_PROT_BTI` for AArch64), and other architecture-specific attributes (like mapping as device memory, setting the global bit, and what not).
3. Provide a remap function in the API that would receive an architecture-specific `attributes` parameter and effectively implement the `ukplat_pt_walk()` -> `ukarch_pte_write()` -> `ukarch_flush_tlb_entry()` sequence.

What do you think?

michpappas:
Also, as discussed before, this is a quite big PR :)  Breaking it down into parts that can be reviewed individually, like the frame allocator, the API, and the implementation for x86, would  probably result in better review quality. Tests on the frame allocator would be great, especially given that newly introduced libraries are expected to come with tests. Tests on the paging   API would be even better. 

While working on the Arm implementation I will be looking at the API, and the x86 implementation, but unfortunately I won't have time to look at the frame allocator at all :( I understand that the PR has been sitting there for a while now, and these comments come last moment, so if time does not permit any of this, no complaints :) It would still be worth looking at these later.

marcrittinghaus:
Hi @michpappas,

thank you very much for looking at the code in such a great detail 😃 You are perfectly right and I also think you spotted a weakness in the API. We should in fact change this as I find the proposed sequence too expensive and cumbersome.

I very much like your idea of generalizing the protection flags and we could also think about exposing the template parameter. I will take a look at it and propose concrete changes.

We have asked a community member to write tests for the frame allocator. Indeed, it would be nice to have tests covering all of the code in one way or the other. But I guess it won't happen for the release. Concerning the frame allocator. Do you think there could be the possibility that you can take a look at `lib/ukfalloc/include/uk/falloc.h` and comment on the general API? If there are bugs or design issues in the buddy-based frame allocator itself I think this is not critical as we can fix this easily without breaking things.

At this point, I would not split the PR because that introduces implicit PR dependencies again. But you are right. It can make reviewing a bit easier, although you then have to manually fetch all dependencies etc. to actually test the code. We hopefully find a more elegant solution to this in the future.

include/uk/plat/paging.h
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * @param pt the page table instance on which to operate.
+ * @param vaddr the virtual address of the first page in the new mapping.
+ * @param paddr the physical address of the memory which the virtual region is
+ *   mapped to. This parameter can be __PADDR_ANY to dynamically allocate
+ *   physical memory as needed.
+ * @param pages the number of pages in requested page size to map.
+ * @param prot the protections to set for the new mapping (PAGE_PROT_* flags).
+ * @param flags page flags (PAGE_FLAG_* flags). The page size can be specified
+ *   with PAGE_FLAGS_SIZE(). If PAGE_FLAG_FORCE_SIZE is not specified, the
```
michpappas: Small typo here, s/PAGE_FLAGS_SIZE/PAGE_FLAG_SIZE/ or update the definitions.

Also, how does the caller know know what level value to pass to the `PAGE_FLAGS_SIZE()` macro? This is architecture-specific, and at least in Arm, the page granularity is configurable. You can have: 
* 4KiB Granule: 4KiB, 2MiB, 1GiB
* 16KiB Granule: 16KiB, 32MiB, 
* 64KiB Granule: 64KiB, 512MiB

On the other hand, I don't expect we will need to support anything other than the 4KiB granule (famous last words).

Anyway, if this API was not public as this function would be called by low-level code, where we can assume it can determine somehow the available sizes. Alternatively, we can introduce a function that returns the currently supported page sizes per level. Or instead of level have the user pass definitions like `PAGE_SIZE_4KB` etc. Sometimes you forget which level corresponds to which size, especially given that the indexing convention between AArch64 and x86 for levels is different (L0->L1->L2->L3 vs L3->L2->L1-L0).

Last but not least, should this header be part of the public API? I would expect that include/uk/ would contain higher level functions that use this API, like `mmap()` and friends.

include/uk/arch/paging.h
```diff
@@ -0,0 +1,436 @@
...
+ * @param lvl a page table level [0..PT_LEVELS - 1]
+ *
+ * @return the physical address to which the given page table entry points
+ */
+#ifndef PT_Lx_PTE_PADDR
+__paddr_t PT_Lx_PTE_PADDR(__pte_t pte, unsigned int lvl);
+#endif
+
+/**
+ * PAGE_Lx_SHIFT(lvl)
```
michpappas, line 159: I found the name of this macro confusing at first, as the values returned are also the number of bits to shift obtain the level+1 bits in an address. Eg the order of L0 is 12, and the shift for L1 bits is 12 too.

plat/common/paging.c
```diff
@@ -0,0 +1,1320 @@
...
+#endif /* CONFIG_PAGING_STATS */
+
+/* The largest level for which PAGE_Lx_HAS() returns a non-zero value, that is
+ * the largest level which can map a page.
+ */
+static unsigned int pg_page_largest_level = PAGE_LEVEL;
+
+/*
+ * Pointer to currently active page table.
+ * TODO: With SMP support, this should move to a CPU-local variable or we
```
michpappas, line 80: How about this? Since this is planned to be merged before SMP support, maybe it would worth to create an issue in the backlog?

plat/common/include/x86/paging.h
```diff
@@ -0,0 +1,326 @@
...
+ * for the frame allocators through this area.
+ */
+#define DIRECTMAP_AREA_START	0xffffff8000000000 /* -512 GiB */
+#define DIRECTMAP_AREA_END	0xffffffffffffffff
+#define DIRECTMAP_AREA_SIZE	(DIRECTMAP_AREA_END - DIRECTMAP_AREA_START + 1)
+
+static inline __vaddr_t
+x86_directmap_paddr_to_vaddr(__paddr_t paddr)
+{
+	UK_ASSERT(paddr < DIRECTMAP_AREA_SIZE);
```
michpappas, line 67: Should these be conditional to `CONFIG_UK_DEBUG`?

plat/common/paging.c
```diff
@@ -0,0 +1,1320 @@
...
+						 lvl - 1);
+				if (unlikely(rc))
+					return rc;
+
+				if (!(flags & PAGE_FLAG_KEEP_PTES))
+					pte = template;
+				else
+					template_level = lvl;
+
+				pte = pgarch_pt_pte_create(pt, pt_paddr, lvl,
```
michpappas, line 581: Maybe I'm missing something here, but how does `pgarch_pte_create()` know that this PTE points to a table rather than an actual mapping? The x86 implementation sets the PSE bit conditionally to `level  > PAGE_LEVEL`, which implies that `pgarch_pte_create` only works PTEs that describe mappings.

michpappas:
> I very much like your idea of generalizing the protection flags and we could also think about exposing the template parameter. I will take a look at it and propose concrete changes.

On second thought, it would perhaps be preferable to use attribute definitions instead of exposing the template. That abstracts away low level details. For example in VMSAv8-64 the memory type is defined as an index in the PTEs, that references a set of values encoded in `MAIR_EL1`. The values in `MAIR_EL1` are set by the platform, so it is not possible for the user to know which bit corresponds to which memory type. Using attributes would eliminate this issue, as one would only need to use something like `PAGE_ATTR_DEVICE_nGnRnE`, and the implementation will do the rest.

> We have asked a community member to write tests for the frame allocator.

That's good to know thanks.

> Indeed, it would be nice to have tests covering all of the code in one way or the other.

Yes, test coverage is important 👍🏼

> But I guess it won't happen for the release.

It's okay :)

> Concerning the frame allocator. Do you think there could be the possibility that you can take a look at `lib/ukfalloc/include/uk/falloc.h` and comment on the general API? If there are bugs or design issues in the buddy-based frame allocator itself I think this is not critical as we can fix this easily without breaking things.

Yes, I can do that 👍🏼

> At this point, I would not split the PR because that introduces implicit PR dependencies again. But you are right. It can make reviewing a bit easier, although you then have to manually fetch all dependencies etc. to actually test the code. We hopefully find a more elegant solution to this in the future.

No worries, I should had brought that up much earlier anyway :)

include/uk/plat/paging.h
marcrittinghaus: > Small typo here, s/PAGE_FLAGS_SIZE/PAGE_FLAG_SIZE/ or update the definitions.

I'll fix the typo 👍🏻 

> Also, how does the caller know know what level value to pass to the `PAGE_FLAGS_SIZE()` macro? This is architecture-specific, and at least in Arm, the page granularity is configurable. You can have:
> 
> * 4KiB Granule: 4KiB, 2MiB, 1GiB
> * 16KiB Granule: 16KiB, 32MiB,
> * 64KiB Granule: 64KiB, 512MiB
> 
> On the other hand, I don't expect we will need to support anything other than the 4KiB granule (famous last words).
>
> Anyway, if this API was not public as this function would be called by low-level code, where we can assume it can determine somehow the available sizes. Alternatively, we can introduce a function that returns the currently supported page sizes per level. Or instead of level have the user pass definitions like #PAGE_SIZE_4KB. Sometimes you forget which level corresponds to which size, especially given that the indexing convention between AArch64 and x86 for levels is different (L0->L1->L2->L3 vs L3->L2->L1-L0).

Yes, this is somewhat suboptimal and in fact I decided for the more abstract usage of levels here because I am aware of this configurable page size on ARM. So I did not want to globally define something like `PAGE_SIZE_4KB`. Instead, on x86 I have defined `PAGE_LEVEL`, `PAGE_LARGE_LEVEL`, and `PAGE_HUGE_LEVEL`. I thought that if you need to go for a certain page size you probably have a very good reason for this and that this will also probably happen in architecture-dependent code. So you would be able to use these more human readable `*_LEVEL` macros in that case. Otherwise, you can just leave it to the paging code to choose the largest possible size on the architecture. The only exception is that `PAGE_LEVEL` should always relate to the smallest possible page size. But of course, we could also use the terms `LARGE`, `HUGE` and perhaps at some point in the future `GIANT` in a more architecture-independent way and define these also on ARM (with different meanings for different granule configurations).

Nevertheless, you might still need all these level-based macros like `PAGE_Lx_SIZE` etc. also in generic code. But my assumption is that this will be for the most part code that uses the page table walking function (`ukplat_pt_walk`), where you get the level from the API. So you do not need to know in advance what level to select and you can use one of the `PAGE_Lx_*` macros to find out, for example, what size the page has that is mapped at the walked address.

If you would want to determine available page sizes and levels in a generic way, you would do something like this:
```C
for (int lvl = PAGE_LEVEL; lvl < PT_LEVELS; lvl++) {
    if (PAGE_Lx_HAS(lvl))
       printf("System supports page level %d with page size %d\n", lvl, PAGE_Lx_SIZE(lvl));
}
```
 
> Last but not least, should this header be part of the public API? I would expect that include/uk/ would contain higher level functions that use this API, like `mmap()` and friends.

Yes, this is intended to be UKs public paging API. Everything you build on top of it is your own business. So POSIX `mmap` will be part of the `posix-mmap` library that implements all of the POSIX memory mapping related syscalls (e.g., also `mprotect`). However, I would like to have a generic UK library that manages the virtual address space and which has an overview of existing virtual memory areas. And the `posix-mmap` library would use this library underneath. But I see `mmap`-type functionality as an optional feature because as we discussed in the meeting you might want to go with a more fixed layout, in which case you only have this paging API or not even that if you go with a fully static page table.

include/uk/arch/paging.h
marcrittinghaus: Don't quite get it 🙈 On x86 you have:
```C
PAGE_Lx_SHIFT(0) = 12;  /* 0 = PAGE_LEVEL */
PAGE_Lx_SHIFT(1) = 21;  /* 1 = PAGE_LARGE_LEVEL */
PAGE_Lx_SIZE(0) = 4 KiB;
PAGE_Lx_SIZE(1) = 2 MiB;
```

plat/common/paging.c
marcrittinghaus: Yup, created a corresponding issue #426 👍🏻

plat/common/include/x86/paging.h
marcrittinghaus: Good question. But in fact, this would mean that every `UK_ASSERT` in any platform code would need to be enclosed, which I dislike. I think we are better off with moving the assert to a public header and find a transparent way to have `ukdebug` hook into it.

plat/common/paging.c
marcrittinghaus: Not sure what you mean. I differentiate in the code between `pgarch_pt_pte_create` which is for page tables and `pgarch_pte_create` which is for data pages. The code that you commented creates a page table PTE so it uses the `pt_pte` variant.

michpappas: Argh, I didn't notice that there are actually two functions, the names are too similar 😖 You're right, there are no issues there, sorry about the confusion.

plat/common/include/x86/paging.h
michpappas: Exactly, that's the way 👍🏼 And with the opportunity remove the conditionals from the arch headers in this PR :)

marcrittinghaus:
Ok, good point. Then we should go with the attributes. Could you elaborate a bit more on how you envision device mapping? What do you mean with the `_nGnRnE` suffix?

As I understand you, you would have to create a fixed device-to-index mapping at compile time and set corresponding values for the respective macros. It then becomes transparent for the user, which index the platform internally selects.

If you already have an idea of how some attribute macros (also for caching for example) would look like, I would like to see this. Maybe we can also find some attribute classes that allow for more generic definitions like the protection flags which map to architecture-dependent flags if available. But in any case, such generic definitions could also be added later.

include/uk/plat/paging.h
michpappas: > Yes, this is somewhat suboptimal and in fact I decided for the more abstract usage of levels here because I am aware of this configurable page size on ARM. So I did not want to globally define something like PAGE_SIZE_4KB. Instead, on x86 I have defined PAGE_LEVEL, PAGE_LARGE_LEVEL, and PAGE_HUGE_LEVEL. I thought that if you need to go for a certain page size you probably have a very good reason for this and that this will also probably happen in architecture-dependent code. So you would be able to use these more human readable *_LEVEL macros in that case. Otherwise, you can just leave it to the paging code to choose the largest possible size on the architecture. The only exception is that PAGE_LEVEL should always relate to the smallest possible page size. But of course, we could also use the terms LARGE, HUGE and perhaps at some point in the future GIANT in a more architecture-independent way and define these also on ARM (with different meanings for different granule configurations).
>
> Nevertheless, you might still need all these level-based macros like PAGE_Lx_SIZE etc. also in generic code. But my assumption is that this will be for the most part code that uses the page table walking function (ukplat_pt_walk), where you get the level from the API. So you do not need to know in advance what level to select and you can use one of the PAGE_Lx_* macros to find out, for example, what size the page has that is mapped at the walked address.

I agree for the most part, although something like GIANT would still break if higher granules only provide two possible sizes. In any case, as granularity is controlled by the platform, and may be chosen either based on a KConfig option on compile time, or based on processor capabilities on runtime, one would need to write some boilerplate if they need to support different granularities, whether we express this information on levels or numbers, so it's okay. If different granularities are ever required to be supported in Unikraft, that is. So let's keep it as it is then :)

> Yes, this is intended to be UKs public paging API. Everything you build on top of it is your own business. So POSIX mmap will be part of the posix-mmap library that implements all of the POSIX memory mapping related syscalls (e.g., also mprotect). However, I would like to have a generic UK library that manages the virtual address space and which has an overview of existing virtual memory areas. And the posix-mmap library would use this library underneath. But I see mmap-type functionality as an optional feature because as we discussed in the meeting you might want to go with a more fixed layout, in which case you only have this paging API or not even that if you go with a fully static page table.

Okay, fair enough 👍🏼

michpappas:
> Ok, good point. Then we should go with the attributes. Could you elaborate a bit more on how you envision device mapping? What do you mean with the `_nGnRnE` suffix?

These are type attributes. Each memory type (Normal / Device) has properties that can be controlled by defining attribute-based sub-types. The attributes for each type are:

**Normal Memory:**
* Inner / Outer Shareable
* Write-back / Write-through
* Transient

where inner / outer shareable controls coherency, write-back / write-through controls caching, and transient is a cache allocation hint that indicates that an entry is meant to be cached for a short period. Actually the above list is a bit simplified as you can have types like "Tagged Normal Inner Write-Back, Outer Write-Back, Read-Allocate, Write-Allocate Non-transient" 😅

**Device Memory:**
* Gather (G)
* Re-order (R)
* Early Acknowledgement (E)

where Gather controls whether multiple accesses can be merged, Re-order controls whether accesses can be reordered, and Early Acknowledgement controls whether intermediate memory between the processor and a device can send write acknowledgments before the write is complete. Device_nGnRnE is therefore the most restrictive type, and Device_GRE is the least restrictive type.

> As I understand you, you would have to create a fixed device-to-index mapping at compile time and set corresponding values for the respective macros. It then becomes transparent for the user, which index the platform internally selects.

Exactly. Long answer below 😄 

Stage 1 Block PTEs (blocks are what x86 calls huge pages) and Page PTEs provide the AttrIndx field (bits [4:2]), that contain an index to the contents of MAIR_EL1 for that entry.

The MAIR_EL1 (Memory Attribute Indirection Register) is essentially a table that contains 8 memory types, and is indexed by the AttrIndx bits in the Block and Page descriptors (ie the PTEs). 

Notice that the memory types contained in MAIR_EL1 entries are not fixed. In fact, the architecture defines more than 8 different types, and it's likely more get introduced on time (eg FEAT_MTE introduced a tagged type). So the responsibility of populating MAIR_EL1 with a suitable subset of types is left to the OS.

For example, `plat/common/include/arm/arm64/cpu_defs.h` defines the following types (the encodings are defined by the architecture):
```
/* Device-nGnRnE memory */    
#define MAIR_DEVICE_nGnRnE      0x00    
/* Device-nGnRE memory */    
#define MAIR_DEVICE_nGnRE       0x04    
/* Device-GRE memory */    
#define MAIR_DEVICE_GRE         0x0C    
/* Outer Non-cacheable + Inner Non-cacheable */    
#define MAIR_NORMAL_NC          0x44    
/* Outer + Inner Write-through non-transient */    
#define MAIR_NORMAL_WT          0xbb    
/* Outer + Inner Write-back non-transient */    
#define MAIR_NORMAL_WB          0xff    
```

Then it defines an arbitrary ordering in MAIR_EL1:
```
/*    
 * Memory types, these values are the indexs of the attributes    
 * that defined in MAIR_EL1.    
 */
#define DEVICE_nGnRnE   0    
#define DEVICE_nGnRE    1    
#define DEVICE_GRE      2    
#define NORMAL_NC       3    
#define NORMAL_WT       4    
#define NORMAL_WB       5    
```
And finally it puts these together to define the contents of MAIR_EL1:
```
#define MAIR_INIT_ATTR  \    
                (MAIR_ATTR(MAIR_DEVICE_nGnRnE, DEVICE_nGnRnE) | \    
                MAIR_ATTR(MAIR_DEVICE_nGnRE, DEVICE_nGnRE) |   \    
                MAIR_ATTR(MAIR_DEVICE_GRE, DEVICE_GRE) |       \    
                MAIR_ATTR(MAIR_NORMAL_NC, NORMAL_NC) |         \    
                MAIR_ATTR(MAIR_NORMAL_WT, NORMAL_WT) |         \    
                MAIR_ATTR(MAIR_NORMAL_WB, NORMAL_WB))
```

> If you already have an idea of how some attribute macros (also for caching for example) would look like, I would like to see this. Maybe we can also find some attribute classes that allow for more generic definitions like the protection flags which map to architecture-dependent flags if available. But in any case, such generic definitions could also be added later.

So one way to move forward is to define all attributes as architecture specific. The downside of this is that all calling code will need to use conditionals (ugly).

The other possibility is, as you suggested, to try to generalize as much as possible based on attributes common to both - and hopefully other - architectures (eg RWX protections, Global Flag etc). Inevitably, there will be still some architecture-specific attributes left. This scheme could look like:
```
/* Generic Attributes */
#define PAGE_ATTR_PROT_NONE
#define PAGE_ATTR_PROT_READ
...
#define PAGE_ATTR_GLOBAL
...

/* Architecture-specific Attirbutes */
#ifdef CONFIG_ARCH_ARM_64
#define PAGE_ATTR_TYPE_DEVICE_nGnRnE
...
#endif

#ifdef CONFIG_ARCH_x86_64
...
#endif

```

PS: Attributes are documented in D5.3.3 of the [Arm Architecture Reference Manual (aka the Arm ARM)](https://developer.arm.com/documentation/ddi0487/latest), and specifically in the sub-section "Attribute fields in stage 1 VMSAv8-64 Block and Page descriptors". Notice however that there is a different set of attributes for Table Descriptors described in D5.3.3.

marcrittinghaus:
@michpappas,

thanks for this detailed explanation. I think I have a rough understanding now!

I think it will be ok for the release to just make the protection the only attribute for now but change the API so that it is expecting attributes everywhere. This would also change `ukplat_page_set_prot` to something like `ukplat_page_set_attr`. We then have some more time to figure out what attributes should be generic and how to call them.

At the moment the generic `PAGE_PROT_*` values  in `/include/plat/paging.h` have no direct mapping to the bits in the page table and are completely architecture-independent. Instead, you have to do the mapping in the architecture-specific platform internal header file like so:
```C
pgarch_pte_create(..., unsigned long prot, ...)
{
...
	if (prot & PAGE_PROT_WRITE)
		pte |= X86_PTE_RW;

	if (!(prot & PAGE_PROT_EXEC))
		pte |= X86_PTE_NX;
...
```
The question is how we map the attributes to bits in the PTE. I see three solutions:
1. We could leave it this way and just assign some "random" number to the generic attributes and manually map like in the code shown. However, that means that architecture-specific attributes (defined in `/arch/*/include/uk/asm/paging.h`) must not collide with ranges we use for the generic attributes. Otherwise, you get into trouble when you combine them. I would not `#ifdef` attribute definitions depending on architecture in `/include/plat/paging.h`.

2. We let the generic ones in `/include/plat/paging.h` just map to architecture-specific attributes (the name is then fixed) like so.
And `ARCH_PAGE_ATTR_PROT_WRITE` would have to be defined in `/arch/*/include/uk/asm/paging.h`:
```C
#define PAGE_ATTR_PROT_WRITE  ARCH_PAGE_ATTR_PROT_WRITE
```

3. Just do not define attributes in `/include/plat/paging.h` altogether but let them all be defined in `/arch/*/include/uk/asm/paging.h` with a common naming scheme like `PAGE_ATTR_PROT_WRITE` for an attribute that denotes write access. Then the "common" ones need to be present on all architectures and if not supported could always just be defined as `0`.

I think solution 3 would be the best and most flexible, although I am not 100% happy that it adds some redundant definitions if "common" attributes are defined the same on multiple architectures. What do you think?

For something like the index fields in the PTE we should in any case have a generic macro that would accept an integer and return the bitmask. Like:
```C
#define PAGE_ATTR_TYPE(idx)  /* shift & mask to AttrIndx field bits */
```
However, I am thinking about if it makes sense to be so explicit concerning the exact memory type. I agree that it would make life easier. So this is a definitive plus 😃 But it would also mean that the architecture-specific paging headers implicitly assume a certain setup by the platform. Perhaps we should think about this some more. For example the platform code could use the architecture-specific paging header to derive indices and not the other way round.

Please see the patch file as a first suggestion:
[attr.zip](https://github.com/unikraft/unikraft/files/8318109/attr.zip)

michpappas:
Hi @marcrittinghaus, thanks for that :) So during the weekend I went through the same thought process, and yes Option 3 sounds like the best approach. It also provides enough flexibility to provide generic protections such as `PAGE_ATTR_PROT_EXEC`, along with more fine grained ones like `PAGE_ATTR_PROT_PXN` or `PAGE_ATTR_PROT_UXN` in case one is not happy with the default. It moreover allows architectures to optionally define their attributes to map directly to the bits in the PTE, and avoid the mapping part used in Option 1. Perhaps not applicable to Armv8, but maybe useful on x86 or some other architecture.

For the index fields I've been thinking of all sorts of things, from defining some basic types and leaving out the less usual ones, or using some generic names and make the actual type they map to configuratble, all the way to defining all valid combinations. I have not made up my mind yet, I would like to think a bit more about it. But that's okay, as our approach provides us with flexibility. So let's go for Option 3 then 👍🏼

PS: The patch makes sense 😎

marcrittinghaus:
> Hi @marcrittinghaus, thanks for that :) So during the weekend I went through the same thought process, and yes Option 3 sounds like the best approach. It also provides enough flexibility to provide generic protections such as `PAGE_ATTR_PROT_EXEC`, along with more fine grained ones like `PAGE_ATTR_PROT_PXN` or `PAGE_ATTR_PROT_UXN` in case one is not happy with the default. It moreover allows architectures to optionally define their attributes to map directly to the bits in the PTE, and avoid the mapping part used in Option 1. Perhaps not applicable to Armv8, but maybe useful on x86 or some other architecture.

Ok, great 😃 I have updated the PR to reflect the proposed changes in the patch.

> For the index fields I've been thinking of all sorts of things, from defining some basic types and leaving out the less usual ones, or using some generic names and make the actual type they map to configuratble, all the way to defining all valid combinations. I have not made up my mind yet, I would like to think a bit more about it. But that's okay, as our approach provides us with flexibility. So let's go for Option 3 then 👍🏼

Yes, this is a difficult decision. Maybe it makes sense to have a macro/inline function that can combine all the different characteristics to a final attribute value as an alternative to defining all the possible combinations.

plat/common/include/x86/paging.h
marcrittinghaus: We will address this in a separate PR, I would say. There are also some other headers that have conditionals depending on availability of `UK_ASSERT` (e.g., the new `event.h`). Until then, we perhaps just go with the "broken" standard of doing things 😏

razvand:
@marcrittinghaus , @michpappas , @rene , this has been a lengthy discussion. @michpappas , I'm adding you as the approver for this PR; I wouldn't be fit for it. Please see the final updates and then, if all is OK, approve it. We will then work on getting the VM API on ARM.

michpappas:
@marcrittinghaus, @razvand, @rene if I don't hear from anyone I'll merge tomorrow morning

plat/common/include/x86/paging.h
skuenzer: I think for this header `UK_ASSERT` is fine to have here as is. So far, we handled dependencies of platform-internal headers with the platforms that use them. So, each platform using this `paging.h` should express a dependency to `lib/ukdebug`. Obviously, this gets a bit messy to retrieve all dependencies for shared platform code in the future but we were thinking of a re-organization of the platform code anyways since a while. We should come up with a scheme that - ideally - automatically creates platform dependencies.... I also had another discussion with Marc in PR #227. As @marcrittinghaus also mentioned, an idea might be to move runtime and compiletime assertions to `/include/uk`. We mainly just need to remove the hard-dependency to `uk_pr_crit` printing and print only if `ukdebug`'s print system is there. The option to enable/disable assertions needs to be moved to `Build options`. I agree to address this with a separate PR.

marcrittinghaus:
> @marcrittinghaus, @razvand, @rene if I don't hear from anyone I'll merge tomorrow morning

Hey @michpappas,

I think its fine to approve. There are some minor things open like the possibility to change the frame allocator as we discussed in the meeting but we can address these later.

If against any odds the merge breaks Unikraft, it was your fault of course😄😏

michpappas:
@marcrittinghaus, thanks for the updates. Let's merge now, but we shall not forget about the topics discussed earlier. We'll also give it more scrutiny while completing the Arm part, I've been through two iterations already :)

marcrittinghaus:
> @marcrittinghaus, thanks for the updates. Let's merge now, but we shall not forget about the topics discussed earlier. We'll also give it more scrutiny while completing the Arm part, I've been through two iterations already :)

Thanks @michpappas for your extensive feedback on this PR! I now it was a huge one 🙈  I am looking forward to you ARM patch 😃

razvand:
@michpappas , I think you also have to mark the approve button.

PR #342, opened 2021-11-26
build/grub: Allow building GRUB based ISO images.

### LINKED ISSUE #265: Build Unikraft with grub support
### Feature request summary

It would be nice to add to the build system the option to automatically create an image compatible with grub (or other external bootloaders). This would aid those that want to use an external bootloader with Unikraft.

### Related architectures

x86_64

### Related platforms

kvm

aadhi0319:
### Prerequisite checklist

 - [ ] Updated relevant documentation. --> Not sure if there is any documentation I need to update.

### Base target

 - Architecture(s): x86_64
 - Platform(s): KVM
 - Application(s): N/A

### Additional configuration
- `CONFIG_KVM_GRUB=y`

Building an iso image requires grub-mkrescue, which can be installed with `apt-get install grub-common` on Ubuntu and Debian Linux.

### Description of changes

The traditional KVM images built by Unikraft are intended to be run with
`qemu` using the `-kernel` option. This is because they are not
formatted in a way that most external bootloaders can load. This PR
will allow users to select a GRUB option in menuconfig under KVM
platform specific options. Upon selection, Unikraft will build an ISO
image in addition to the regular KVM image such that any multiboot
complient bootloader (including GRUB) can load and run the image. This
commit is intended to close #265.

### DIFF
--- a/support/scripts/mkgrubimg
+++ b/support/scripts/mkgrubimg
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+
+# help menu
+usage()
+{
+	echo "Usage: $0 [options] [output file]"
+	echo "Creates bootable GRUB images from Unikraft kernel images."
+	echo ""
+	echo "  -h                         Display help and exit"
+	echo "  -f                         Output format (supported: iso)"
+	echo "  -k                         Path to the Unikraft kernel image"
+	echo "  -a                         Kernel command-line parameters (optional)"
+	echo "  -i                         Path to initrd cpio file (optional)"
+	exit 1
+}
+
+# default options
+OPTFORMAT="iso"
+OPTCMDLINE=""
+
+# cleanup
+BUILDDIR=
+sighandler_exit()
+{
+	if [ ! -z "${BUILDDIR}" -a -d "${BUILDDIR}" ]; then
+		rm -rf "${BUILDDIR}"
+	fi
+}
+
+# process options
+while getopts "hk:i:f:a:" OPT; do
+	case "${OPT}" in
+		h)
+			usage
+			;;
+		f)
+			OPTFORMAT="${OPTARG}"
+			;;
+		k)
+			OPTKERNELIMG="${OPTARG}"
+			;;
+		a)
+			OPTCMDLINE="${OPTARG}"
+			;;
+		i)
+			OPTINITRD="${OPTARG}"
+			;;
+		*)
+			usage
+			;;
+	esac
+done
+shift $((OPTIND-1))
+
+# validate arguments
+if [[ -z "${OPTKERNELIMG}" ]]; then
+	echo "Missing path to kernel image ('-k')" 1>&2
+	usage
+fi
+if [[ -z "${1}" ]]; then
+	echo "Missing path to output image" 1>&2
+	usage
+fi
+if [ ! -f "${OPTKERNELIMG}" ]; then
+	echo "${OPTKERNELIMG} does not exist or is not a file" 1>&2
+	exit 1
+fi
+if [ ! -z "${OPTINITRD}" -a ! -f "${OPTINITRD}" ]; then
+	echo "${OPTINITRD} does not exist or is not a file" 1>&2
+	exit 1
+fi
+
+# Register exit handler and create BUILDDIR
+trap sighandler_exit exit
+BUILDDIR="$( mktemp -d )"
+if [ $? -ne 0 -o -z "${BUILDDIR}" -o ! -d "${BUILDDIR}" ]; then
+	echo "Failed to create temporary directory" 1>&2
+	exit 1
+fi
+
+OPTKERNELNAME="${OPTKERNELIMG##*/}"
+OPTINITRDNAME="${OPTINITRD##*/}"
+
+case "${OPTFORMAT}" in
+	# generate grub iso image
+	iso)
+		# configure grub
+		mkdir -p "${BUILDDIR}/boot/grub"
+
+		if [ ! -z "${OPTINITRD}" ]; then
+			GRUB_CFG_INITRD="module /boot/${OPTINITRDNAME}"
+			cp "${OPTINITRD}" "${BUILDDIR}/boot/${OPTINITRDNAME}" || exit 1
+		fi
+		cp "${OPTKERNELIMG}" "${BUILDDIR}/boot/" || exit 1
+
+		cat >"${BUILDDIR}/boot/grub/grub.cfg" <<EOF
+set default=0
+set timeout=0
+
+menuentry "${OPTKERNELNAME}" {
+	multiboot /boot/${OPTKERNELNAME} /${OPTKERNELNAME} ${OPTCMDLINE}
+	${GRUB_CFG_INITRD}
+	boot
+}
+EOF
+
+		# build grub image
+		GRUB_INSTALL_MSGS=$( grub-mkrescue -o "$1" --install-modules="multiboot" --fonts="" --themes="" --locales="" "${BUILDDIR}/" 2>&1 )
+		if [ $? != 0 ]; then
+			printf '%s\n' "$GRUB_INSTALL_MSGS" 1>&2
+			exit 1
+		fi
+		;;
+	*)
+		echo -e "\"${OPTFORMAT}\" not a supported format." 1>&2
+		usage
+		;;
+esac


mandrei12 (approved):
Everything is okay.

Reviewed-by: Andrei Mutu <andrei.mutu@stud.acs.upb.ro>

nderjung:
Hi thanks again for taking the time to look into this @aadhi0319!

As discussed offline, this PR will be re-worked to make the creation of ISO file more abstract.  Highlights from our conversation include:
 - Determining if a CPIO file was generated, and if so, embedding this as as well as the option `initrd` (requires separate PR for introducing building the CPIO file as part of the build-system after #262 is merged) into GRUB (part of this PR is to allow the user to provide the path to the folder to be archives, which should default as the application's root folder);
  - Placing the `grub.cfg` file into the `build/` directory after the prepare step, such that it can be modified before a build in case the user wishes to introduce additional, custom logic and functionality;
   - Finding out and variabilising options of `grub-mkrescue` such that the final ISO archive is lean and minimal;
   - Investigating abstractions that appear both on grub and syslinux to determine whether a build target to create the ISO file can use one of either.

Thanks again! 🐒

nderjung (changes requested):
Hi thanks again for taking the time to look into this @aadhi0319!

As discussed offline, this PR will be re-worked to make the creation of ISO file more abstract such that it appears in the "Build Options" part of the menu and targets whatever arch-plat the user has selected.  Highlights from our conversation include:
 - Determining if a CPIO file was generated and the kernel will recognise it, and if so, embedding this as as well as the option `initrd` (requires separate PR for introducing building the CPIO file as part of the build-system after #262 is merged) into GRUB (part of this PR is to allow the user to provide the path to the folder to be archives, which should default as the application's root folder);
  - Placing the `grub.cfg` file into the `build/` directory after the prepare step, such that it can be modified before a build in case the user wishes to introduce additional, custom logic and functionality;
   - Finding out and variabilising options of `grub-mkrescue` such that the final ISO archive is lean and minimal;
   - Investigating abstractions that appear both on grub and syslinux to determine whether a build target to create the ISO file can use one of either.

Thanks again! 🐒

plat/kvm/Config.uk
```diff
@@ -1,7 +1,7 @@
 menuconfig PLAT_KVM
        bool "KVM guest"
        default n
-       depends on (ARCH_X86_64 || (ARCH_ARM_64 && !HAVE_SYSCALL))
+       depends on ((ARCH_X86_64 && !HAVE_SMP) || (ARCH_ARM_64 && !HAVE_SMP && !HAVE_SYSCALL))
```
aadhi0319: I guess my branch might be a little outdated. Not sure why this changed but I did not edit this line and it is irrelevant to the GRUB ISO commit.

cristian-vijelie: That's from the SMP common API commits. Make sure that your branch is up-to-date with the staging branch.

aadhi0319: Okay fixed it...thanks!

aadhi0319:
Please note that this commit relies on `mkcpio` from the currently open pull request #359 in order to work.

cristian-vijelie (approved):

skuenzer:
Hey @aadhi0319,
I really like having the possibility to create ISO images, this is something that simplifies people's life. However, I think it should be probably done as a support script instead of being part of the build system because of the following reasons:
* Not all platforms support booting from GRUB ISOs (e.g., `plat-linuxu`, `plaot-solo5`, etc.). It is first of all bound to specific platforms that support this option (e.g., bare-metal, KVM). Normally, I would say this option should be in the submenus of the platforms that support ISOs, however...
* I can imagine that some steps during creation require `root` privileges; especially when we enhance it later with more output formats other than ISO, for instance, pen drive, virtual disk image. Requiring `root` privileges while we compile is bad practice and we should avoid this as being part of the build system.
* We would currently end up in duplicates for creating images when splitting across supported platforms.

I am suggesting to create a (bash) script within `support/scripts`, maybe called `mkgrubimg`. A user would call this script after a Unikraft kernel was built. `kraft` would be able to use this script to automatically package a built kernel into an ISO (or similar) when configured via `kraft.yaml`. This means that for `kraft` users the ISO building step would be done automatically. Furthermore, this script would still allow non-`kraft` and enhanced users to manually create images.
I can imagine a command line interface similar to the following:
```
> mkgrubimg -k [path-to-unikraft-kernel] -i [path-to-initrd] -f [output-format] [output]
```
where
* `-k [path-to-unkraft-kernel]` is the path to the unikraft kernel to package, like `build/unikraft_kvm_x86_64`
* `-i [path-to-initrd]` is an optional argument to handover an initrd. The user would separately create it with `mkcpio` or `kraft` is taking care of it when calling `mkgrubimg`
* `-f [output-format]` is a way to specify different image formats. For now I can imagine something like `iso`, `qcow`, `raw`, and `pvgrub`.
  The reason is that they slightly need different partition layouts, filesystems, and grub module installations. The target should be to create a minimal grub installation and configuration (as minimal as possible) for the selected target.
* `[output]` is the output file/image that is created. Ideally the script automatically sizes the output disk image as small as possible to fit only the necessary parts, like: partition table, filesystem, boot sector, grub files, the kernel, and the initrd (if given).
In worst case, one could use `sudo` in front of the command if `root` privileges are necessary (e.g., to setup loop devices) or the script could ask for privilege increase by itself when necessary. For now, I would expect that the script returns an error if the output file exists already.

What do you think?
You can use `getopt` in bash scripts for parsing arguments. I recommend having a look at [qemu-guest](https://github.com/unikraft/kraft/blob/staging/scripts/qemu-guest#L483) in our `kraft` repository.

Simon

aadhi0319:
Hello @skuenzer,

* At the moment, root privileges are not required, but you are right in that future output formats could require root privileges. I agree with you in that the build process should not require root privileges. It is dangerous and likely a security risk too.
* The duplicates issue is not clean and I was looking for a solution to that. Moving this commit to a separate script would certainly help with that.
* As per the points above, I think a script with the API you described would work well. Currently, I have set the configuration to generate a very minimal GRUB image. That said, the current layout allows for users to select more modules should they require it. I think we should allow users to build a less optimal GRUB image if they choose to do so (but also not have to deal with that complexity by default). At that point, we start getting into how we allow users to specify those parameters (the menuconfig selection tool worked well for this).
* The bulk of this update is actually just allowing users to select GRUB modules and other GRUB configuration options from menuconfig. The actual GRUB ISO image building code is only like 6 lines of code (`393-399 plat/kvm/Linker.uk`). The main component of this build process is `grub-mkrescue -o $@ --install-modules="$(GRUB_INSTALL_MODULES-y)" --fonts="" --themes="" --locales="$(GRUB_INSTALL_LOCALES-y)" $(BUILD_DIR)/kvm_iso/`. One thing we could do is accept the modules and locales as whitespace separated parameters after `--modules` and `--locales` parameters respectively. We could alternatively generate the GRUB config file with the correct directory structure and provide users with a default `grub-mkrescue` command that they can edit if they please. Let me know how you think we should move forward on this front.

Thanks,
Aadhithya

skuenzer:
> 

Hi @aadhi0319,

these are good points. In the beginning, I would focus with the script on the absolute minimum required setup and modules for creating a bootable image (ISO, qcow, etc.). For instance, it should include grub's ext module only if you format the disk with ext. In my opinion, the case where you want to add more modules is only important if you start customizing the generated image after the script created it. I think step-wise we can add more advanced options, like people can override the image size, change the fstype, somehow specify a different partitioning layout, and selecting more grub modules. As long as these options are all optional, I think it is fine. However, I could even imagine that one that wants a custom setup may likely create an own script for the image creation. 🤔  This is why I would add these customization functionality later.

Thanks,

Simon

aadhi0319:
Okay sounds good I'll start working on transitioning the updates to a bash script.

razvand:
@skuenzer , please see the updates from @aadhi0319

skuenzer:
> @skuenzer , please see the updates from @aadhi0319

@razvand: I am waiting for @aadhi0319 providing an update with using a script. He mentioned that he works on it at the moment.

aadhi0319:
@skuenzer, sorry I thought GitHub would notify you so I didn't follow up with you directly. The script can be found https://github.com/aadhi0319/unikraft/commit/f17b589bc7d210fd9c18fae6288ce84e92f0b6f5. It is also linked by GitHub under my "Okay sounds good I'll start working on transitioning the updates to a bash script" comment.

skuenzer:
> @skuenzer, sorry I thought GitHub would notify you so I didn't follow up with you directly. The script can be found [aadhi0319@f17b589](https://github.com/aadhi0319/unikraft/commit/f17b589bc7d210fd9c18fae6288ce84e92f0b6f5). It is also linked by GitHub under my "Okay sounds good I'll start working on transitioning the updates to a bash script" comment.

Hey, cool, thanks. You should force push the new commit history to this branch here so that this PR is updated.

Simon

aadhi0319:
> > @skuenzer, sorry I thought GitHub would notify you so I didn't follow up with you directly. The script can be found [aadhi0319@f17b589](https://github.com/aadhi0319/unikraft/commit/f17b589bc7d210fd9c18fae6288ce84e92f0b6f5). It is also linked by GitHub under my "Okay sounds good I'll start working on transitioning the updates to a bash script" comment.
> 
> Hey, cool, thanks. You should force push the new commit history to this branch here so that this PR is updated.
> 
> Simon

~~How do I do that?~~
I think I did it, but you might want to make sure I did it properly.

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+
+# help menu
+usage()
+{
+	echo "Usage: $0 [options] [output file]"
+	echo "Creates bootable GRUB images from Unikraft kernel images."
+	echo ""
+	echo "  -h                         Display help and exit"
+	echo "  -f                         Output format (supported: iso)"
+	echo "  -i                         (optional) Path to initrd directory"
```
skuenzer: This is just style preference: I would list the `-i` option after `-k` and write `(optional)` after the Parameter description: `Path to init ramdisk (optional)`

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+	echo "  -k                         Path to the Unikraft kernel image"
```
skuenzer, line 11: Could you also add an option tho allow passing of kernel command-line parameters (typically `-a` for `append`)

skuenzer: This parameter is also optional

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+}
+
+# process options
+while getopts "hk:i:f:" options; do
+	case "${options}" in
+		h)
+			usage
+			;;
+		f)
+			OPTFORMAT=${OPTARG}
```
skuenzer: I recommend to quote: `OPTFORMAT="${OPTARG}"`

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+			;;
+		i)
+			OPTINITRD=${OPTARG}
```
skuenzer: Quote

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+			;;
+		k)
+			OPTKERNELIMG=${OPTARG}
```
skuenzer: I would quote here, too

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+case "${OPTFORMAT}" in
+	# generate grub iso image
+	iso)
+		# generate configuration file
+		GRUB_CFG_TXT="set default=0 \n"
+		GRUB_CFG_TXT+="set timeout=0 \n"
+		GRUB_CFG_TXT+="menuentry \"${OPTKERNELIMG##*/}\" { \n\t"
+		GRUB_CFG_TXT+="multiboot /boot/${OPTKERNELIMG##*/} /${OPTKERNELIMG##*/} console=ttyS0 \n\t"
```
skuenzer: instead of `console=ttyS0` (which is Linux specific) put here the command line arguments given from the user with `-a` (or leave it out if not available)

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+		if [ ! -z ${OPTINITRD} ]; then
+			GRUB_CFG_TXT+="module /boot/initrd_archive.cpio \n\t"
+		fi
+		GRUB_CFG_TXT+="boot \n"
+		GRUB_CFG_TXT+="}"
+
+		# configure grub
+		mkdir -p "${BUILDDIR}/isofiles/boot/grub"
+		echo -e ${GRUB_CFG_TXT} > "${BUILDDIR}/isofiles/boot/grub/grub.cfg"
```
skuenzer: please quote the variable, typically I also use printf in such scenarios in order to be on the safe side, but you can actually also use cat for the file template:
```bash
cat > "${BUILDDIR}/isofiles/boot/grub/grub.cfg" <<EOF
set default=0
set timeout=0

${GRUB_CFG_ENTRY}

boot
EOF

Now you just need to generate `GRUB_CFG_ENTRY` only with the `menuentry` and the initrd. No quoting for variables in here needed
```

support/scripts/mkgrubimg
skuenzer: Also, I would not call another tool, just expect the finished initrd file (no directory). This way, this script could even be used for Unikraft images that do not read the CPIO format and expect a different file type than cpio.

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+			;;
+		f)
+			OPTFORMAT=${OPTARG}
+			;;
+		i)
+			OPTINITRD=${OPTARG}
+			;;
+		k)
+			OPTKERNELIMG=${OPTARG}
+			BUILDDIR=${OPTKERNELIMG%/*}
```
skuenzer: I would use temp directory for the build instead (e.g., use `mktemp -d` for instance) and you can delete it with a trap on script exit/cancellation:
```bash
# Place this part at the beginning of the script
BUILDDIR=
sighandler_exit()
{
   if [ ! -z "${BUILDDIR}" -a -d "${BUILDDIR}" ]; then
      rm -rf "${BUILDDIR}"
   fi
}

# Register exit handler and create BUILDDIR
# You should place this snippet directly _after_ finishing argument parsing
trap sighandler_exit exit
BUILDDIR="$( mktemp -d )"
if [ $? -ne 0 -o -z "${BUILDDIR}" -o ! -d "${BUILDDIR}" ]; then
   echo "Failed to create temporary directory" 1>&2
   exit 1
fi
```

Please place this snippet directly after argument parsing (and not part of the switch statement).

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+			${MKCPIO} -q "${BUILDDIR}/isofiles/boot/initrd_archive.cpio" "${OPTINITRD}"
+		fi
+		cp "${OPTKERNELIMG}" "${BUILDDIR}/isofiles/boot/"
+
+		# build grub image
+		GRUB_MODULES="part_acorn part_amiga part_apple part_bsd part_dfly part_dvh part_gpt part_plan part_sun part_sunpc normal multiboot"
+		grub-mkrescue -o $1 --install-modules="${GRUB_MODULES}" --fonts="" --themes="" --locales="" "${BUILDDIR}/isofiles/" &>/dev/null
+
+		# cleanup
+		rm -rf "${BUILDDIR}/isofiles"
```
skuenzer: With the trap, this is not needed anymore

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+	usage
+fi
+
+case "${OPTFORMAT}" in
+	# generate grub iso image
+	iso)
+		# generate configuration file
+		GRUB_CFG_TXT="set default=0 \n"
+		GRUB_CFG_TXT+="set timeout=0 \n"
+		GRUB_CFG_TXT+="menuentry \"${OPTKERNELIMG##*/}\" { \n\t"
```
skuenzer: Hum 🤔 ... either keep the original filename for both: kenrle and initrd or - since the image will always have only one kernel and only one initrd - call the files just: `kernel` and `initrd`. At least I would retrieve the menuentry title from the original filename. As good practice,  I would propose to "compute" the file name just once and store it in a variable, like:
```bash
OPTKERNELNAME="$( basename "${OPTKERNEL}" )"
OPTINITRDNAME="$( basename "${OPTINITRD}" )"
```
or
```bash
OPTKERNELNAME="${OPTKERNEL##*/}"
OPTINITRDNAME="${OPTINITRD##*/}"
```

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+		mkdir -p "${BUILDDIR}/isofiles/boot/grub"
+		echo -e ${GRUB_CFG_TXT} > "${BUILDDIR}/isofiles/boot/grub/grub.cfg"
+		if [ ! -z ${OPTINITRD} ]; then
+			MKCPIO="${BASH_SOURCE[0]%/*}/mkcpio"
+			${MKCPIO} -q "${BUILDDIR}/isofiles/boot/initrd_archive.cpio" "${OPTINITRD}"
+		fi
+		cp "${OPTKERNELIMG}" "${BUILDDIR}/isofiles/boot/"
+
+		# build grub image
+		GRUB_MODULES="part_acorn part_amiga part_apple part_bsd part_dfly part_dvh part_gpt part_plan part_sun part_sunpc normal multiboot"
```
skuenzer: Do we really need all these partition table modules? We will not boot from a GPT or Apple formatted harddisk, we boot from an ISO. I think the only required is in fact `multiboot`. Maybe `normal` is also needed but I don't know what is in there. Somewhere I saw that also `iso9660` and `biosdisk` might be needed, please double check.

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+		grub-mkrescue -o $1 --install-modules="${GRUB_MODULES}" --fonts="" --themes="" --locales="" "${BUILDDIR}/isofiles/" &>/dev/null
+
+		# cleanup
+		rm -rf "${BUILDDIR}/isofiles"
+		;;
+	*)
+		echo -e "${RED}\"${OPTFORMAT}\" not a supported format.${NC}" 1>&2
```
skuenzer: Colors are nice but for this simple script, I don't think we need them. No color sequences make it easier to read when output is redirected, e.g., to a file. This code forces the escape sequence to be written to the file, too... normally, programs check the the output device and turn colors dynamically on or off.

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+	echo ""
+	echo "  -h                         Display help and exit"
+	echo "  -f                         Output format (supported: iso)"
+	echo "  -i                         (optional) Path to initrd directory"
+	echo "  -k                         Path to the Unikraft kernel image"
+	exit 1
+}
+
+# process options
+while getopts "hk:i:f:" options; do
```
skuenzer: I prefer to keep one style for variable names. If you name them all with capital letters, do this also for options, like:
```bash
while getopts "hk:i:f:" OPT; do
   case "$OPT" in
```

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+RED="\033[0;31m"
+
+# help menu
+usage()
+{
+	echo "Usage: $0 [options] [output file]"
+	echo "Creates bootable GRUB images from Unikraft kernel images."
+	echo ""
+	echo "  -h                         Display help and exit"
+	echo "  -f                         Output format (supported: iso)"
```
skuenzer, line 10: I would set `iso` even as the default

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,82 @@
...
+	echo "  -i                         (optional) Path to initrd directory"
+	echo "  -k                         Path to the Unikraft kernel image"
+	exit 1
+}
+
+# process options
```
skuenzer, line 30: You can set here some defaults. like
```bash
OPTFORMAT=iso
```

skuenzer (changes requested):
Hey @aadhi0319 ,

many thanks for this update. I gave you some comments. In general, I think this script is what we should do 👍🏻 

Simon

support/scripts/mkgrubimg
aadhi0319: You were right, only multiboot was needed. I didn't realize ISOs didn't have partitions and I guess I must have mistaken an error fixed by something else as an error fixed by adding the partition modules.

aadhi0319:
@skuenzer I incorporated all of the changes you suggested and tested it on helloworld and sqlite. Let me know how it looks on your end.

skuenzer:
> @skuenzer I incorporated all of the changes you suggested and tested it on helloworld and sqlite. Let me know how it looks on your end.

I replied to your fork. Can you update this PR after integrating the small changes? It looks really good now, btw.
Simon

aadhi0319:
@skuenzer Amended the old commit with the minor quoting fixes.

skuenzer:
Weird, somehow the commit in this PR is not updated. Can you check this?

aadhi0319:
@skuenzer This issue has been around long enough that it's all sorts of weird. See if the PR is updated now; I did a force push from the aadhi0319:grub_script branch to the aadhi0319:grub_support branch.

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,101 @@
...
+menuentry "${OPTKERNELNAME}" {
+	multiboot /boot/${OPTKERNELNAME} /${OPTKERNELNAME} ${OPTCMDLINE}
+	${GRUB_CFG_INITRD}
+	boot
+}
+EOF
+
+		# build grub image
+		grub-mkrescue -o "$1" --install-modules="multiboot" --fonts="" --themes="" --locales="" "${BUILDDIR}/" &>/dev/null
+		;;
```
skuenzer, line 113:
```bash
   grub-mkrescue -o "$1" --install-modules="multiboot" --fonts="" --themes="" --locales="" "${BUILDDIR}/" &>/dev/null || exit 1
   ;;
```

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,101 @@
...
+	# generate grub iso image
+	iso)
+		# configure grub
+		mkdir -p "${BUILDDIR}/boot/grub"
+
+		if [ ! -z "${OPTINITRD}" ]; then
+			GRUB_CFG_INITRD="module /boot/${OPTINITRDNAME}"
+			cp "${OPTINITRD}" "${BUILDDIR}/boot/${OPTINITRDNAME}"
+		fi
+		cp "${OPTKERNELIMG}" "${BUILDDIR}/boot/"
```
skuenzer:
```bash
cp "${OPTKERNELIMG}" "${BUILDDIR}/boot/" || exit 1
```

support/scripts/mkgrubimg
```diff
@@ -0,0 +1,101 @@
...
+
+case "${OPTFORMAT}" in
+	# generate grub iso image
+	iso)
+		# configure grub
+		mkdir -p "${BUILDDIR}/boot/grub"
+
+		if [ ! -z "${OPTINITRD}" ]; then
+			GRUB_CFG_INITRD="module /boot/${OPTINITRDNAME}"
+			cp "${OPTINITRD}" "${BUILDDIR}/boot/${OPTINITRDNAME}"
```
skuenzer: Sorry, I did not see this before: We should probably check return code of the commands and return from the script with an error as well. Otherwise we have no way to detect if the iso was successfully created if we include the script in a Makefile, for example.
```bash
cp "${OPTINITRD}" "${BUILDDIR}/boot/${OPTINITRDNAME}" || exit 1
```

skuenzer (changes requested):

skuenzer:
> @skuenzer This issue has been around long enough that it's all sorts of weird. See if the PR is updated now; I did a force push from the aadhi0319:grub_script branch to the aadhi0319:grub_support branch.

In fact, this is totally weird. But as you can see in the timeline, there wasn't a push before 🤷. Your last force push did the trick.

aadhi0319:
@skuenzer Good point...new push should have added the error catching changes.

skuenzer (changes requested):
> @skuenzer Good point...new push should have added the error catching changes.

great. Yes it works and I could boot from the ISO! This is good. However, I run into some problems and the `/dev/null` redirection wasn't helpful actually. On my system I had to install `xorriso` and uninstall `grub-xen`. The script as of now, returns correctly with an error but you don't see the grub error output - it just silently fails. If I remove the redirection, I could see the errors, fix my system to make it work but then, in the success case the script is too verbose... What do you think? I am kind of tending to store the grub output into a variable and print it only in the error case and also redirect it then to stderr:
To get it working, I tested the grub install command invocation with this:
```bash
GRUB_INSTALL_MSGS=$( grub-mkrescue -o "$1" --install-modules="multiboot" --fonts="" --themes="" --locales="" "${BUILDDIR}/" 2>&1 )
if [ $? != 0 ]; then
   printf '%s\n' "$GRUB_INSTALL_MSGS" 1>&2
   exit 1
fi
```

Additionally, for prettifying we should print a meaningful error message if initrd and kernel image are actually wrong paths and `cp` would fail. I would change yourr argument validation block to the following:

```bash
# validate arguments
if [[ -z "${OPTKERNELIMG}" || -z "$1" ]]; then
	echo "Missing path to kernel image ('-k')" 1>&2
	usage
fi
if [[ -z "${1}" ]]; then
	echo "Missing path to output image" 1>&2
	usage
fi
if [ ! -f "${OPTKERNELIMG}" ]; then
	echo "${OPTKERNELIMG} does not exist or is not a file" 1>&2
	exit 1
fi
if [ ! -z "${OPTINITRD}" -a -f "${OPTINITRD}" ]; then
	echo "${OPTINITRD} does not exist or is not a file" 1>&2
	exit 1
fi
```
You can take the snippet into the script and add the messages and then I would approve the PR.

For the `grub-xen` problem, I don't know if this could have been fixed with another parameter to `grub-mkrescue`. The problem was that the tool was failing while looking up for the multiboot.mod in the grub-xen installation instead of grub-i386. Do you have an idea? I think we can add a fix later if there is actually one that we could apply.

Simon

aadhi0319:
Glad to see you could boot from the ISO. The script not producing any output is a consequence of originally using the script in the unikraft build process. I still agree with you that upon success the script output should be minimal, if anything, but the user should still be able to see relevant error messages. I approve the change, but I'm currently traveling right now. If you can push the change please do, otherwise I will push it tonight. As for xorriso, I'm not sure I've never seen that error. You could try using the old script with extra modules and see if that fixes your issue. If it does, then we could look into adding more modules.

skuenzer:
Sounds great. In my case, xorriso was just missing on my system. The fix was to install it because grub-mkrescue depends on it to create the actual iso format. The thing was that with the silent exit, I wasn't able to quickly fix this issue. This is why I suggest capturing the output in a variable but keep everything silent in success cases for build system integrations.
I can do the approval within the next 12h when you pushed the changes.

Thanks,

Simon

aadhi0319:
@skuenzer integrated your suggestions.

skuenzer:
@aadhi0319 Great, thanks! It also looked to me that you have now multiple commit in the PR history and you need to combine them into one commit..., can you do this, then we are done 😄

skuenzer:
Hey,
I just merged your two commits into one and fixed two little typos that I found while testing:

```diff
@@ -53,7 +53,7 @@ done
 shift $((OPTIND-1))

 # validate arguments
-if [[ -z "${OPTKERNELIMG}" || -z "$1" ]]; then
+if [[ -z "${OPTKERNELIMG}" ]]; then
 	echo "Missing path to kernel image ('-k')" 1>&2
 	usage
 fi
(1/2) Stage this hunk [y,n,q,a,d,j,J,g,/,e,?]? y
@@ -65,7 +65,7 @@ if [ ! -f "${OPTKERNELIMG}" ]; then
 	echo "${OPTKERNELIMG} does not exist or is not a file" 1>&2
 	exit 1
 fi
-if [ ! -z "${OPTINITRD}" -a -f "${OPTINITRD}" ]; then
+if [ ! -z "${OPTINITRD}" -a ! -f "${OPTINITRD}" ]; then
 	echo "${OPTINITRD} does not exist or is not a file" 1>&2
 	exit 1
 fi
```

Now it works :+1:, I am going to approve this PR so that it can become part of the 0.9 release. Thanks for your work!

Simon

skuenzer (approved):

aadhi0319:
@skuenzer Thanks for fixing those typos!

PR #348, opened 2021-11-29
lib/ukrust: Add rust internal library

vladandrew:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Application(s): 

### Additional configuration
 - `CONFIG_LIBUKRUST=y`
 - Rust insalled: `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
 - `bindgen` installed: `sudo apt install -y bindgen` or `cargo install --locked --version 0.56.0 bindgen`
 - Nightly: `rustup override set nightly && rustup update && rustup component add rust-src`

### Description of changes
This patch adds  the kernel crate and links the alloc/core and compiler_builtins crates. 

The kernel crate implements a global allocator and the bindings for the Unikraft C API to Rust.
 All the functions, variables and defines exported by the headers imported in `bindings_helper.h`
are available to Rust libraries. To use these bindings, one has to import the kernel crate:
```
use ukrust_sys::bindings;
bindings::CONFIG_KVM_PCI;
```
We use the `bindgen` tool to generate the bindings. Since we are also linking the core and alloc
crate, we are able to use any function from them e.g.:
```
use alloc::vec::Vec;
let mut v: Vec<u32> = Vec::new();
v.push(31);
```
Our implementation calls `UK_CRASH` through `__ukrust_sys_crash` in all panics and oom scenarios. Both
core and alloc are built from the rust sources. We are currently targeting `x86_64` but will soon push a
patch for `arm64`.


vladandrew:
Some tests are failing because `bindgen` is not installed in the CI/CD image.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,62 @@
...
+		--out-dir $(LIBUKRUST_BUILD) -L $(LIBUKRUST_BUILD) -A dead_code \
+		-A unused_doc_comments --crate-name core $(RUST_LIB_SRC)/core/src/lib.rs)
+
+$(LIBUKRUST_BUILD)/compiler_builtins.o: $(LIBUKRUST_BASE)/compiler_builtins.rs $(LIBUKRUST_BUILD)/core.o
+	$(call build_cmd, BUILD, libukrust, $(notdir $@), \
+	$(RUSTC) $(RUST_UK_FLAGS) \
+		--out-dir $(LIBUKRUST_BUILD) -L $(LIBUKRUST_BUILD) -A dead_code \
+		-A unused_doc_comments --crate-name compiler_builtins $(LIBUKRUST_BASE)/compiler_builtins.rs)
+
+UK_PREPARE += $(LIBUKRUST_BUILD)/bindings_generated.rs $(LIBUKRUST_BUILD)/kernel.o \
```
cffs: The *prepare* step should only be done when ukrust is selected, so that bindings generation is done only when necessary. You can do this by replacing `UK_PREPARE` by `UK_PREPARE-$(CONFIG_LIBUKRUST)`.

lib/ukrust/Config.uk
```diff
@@ -0,0 +1,6 @@
+config LIBUKRUST
+	bool "ukrust: Rust main crate for Unikraft"
```
cffs: Maybe make this more explicit, something like "Expose Unikraft internals as a rust crate"?

lib/ukrust/kernel/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+#![no_std]
+#![no_std]
```
cffs: There is one redundant `#![no_std]`.

lib/ukrust/kernel/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+#![feature(
+    allocator_api,
+    alloc_error_handler,
+    associated_type_defaults,
+    const_fn_trait_bound,
+    const_mut_refs,
+    const_raw_ptr_deref,
```
cffs: Feature `const_raw_ptr_deref` is not needed anymore, it has been stable since rust 1.58.0.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,62 @@
...
+	-Cpanic=abort -Cembed-bitcode=n -Clto=n -Crpath=n \
+	-Cforce-unwind-tables=n -Ccodegen-units=1 \
+	-Zbinary_dep_depinfo=y -Zsymbol-mangling-version=v0 \
+	-Dunsafe_op_in_unsafe_fn -Drust_2018_idioms \
+	--crate-type rlib
+
+# Generate rust bindings for the Unikraft API
+$(LIBUKRUST_BUILD)/bindings_generated.rs: $(LIBUKRUST_BASE)/kernel/bindings_helper.h
+	$(call build_cmd, GEN, libukrust, $(notdir $@), \
+		bindgen -o $(LIBUKRUST_BUILD)/bindings_generated.rs --use-core --ctypes-prefix crate::c_types \
```
cffs, line 24: This `bindgen` call fails when [libcxx](https://github.com/unikraft/lib-libcxx) is also selected, making it impossible to use modules written in rust and modules/apps written in C++ at the same time. It also means you cannot mix modules written in low-level rust with modules using cargo-based rust, because the later depends on [libunwind](https://github.com/unikraft/lib-libunwind) which itself depends on [libcxx](https://github.com/unikraft/lib-libcxx).

This is not crippling however, we can already proceed as-is and fix this later on.

lib/ukrust/kernel/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    receiver_trait,
+)]
+use core;
+
+pub mod allocator;
```
cffs: Is it necessary to include the allocator for rust bindings to Unikraft to work? Is this preventing one to use another allocator in its rust code?

lib/ukrust/kernel/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+pub mod bindings;
+pub mod c_types;
+
+#[alloc_error_handler]
+pub fn alloc_error(_layout: core::alloc::Layout) -> ! {
+        unsafe {bindings::rust_uk_crash();};
```
cffs: Can't we use the definition from `compilers_builtins` instead, that is marked as being diverging? Or `panic!` maybe?

lib/ukrust/compiler_builtins.rs
```diff
@@ -0,0 +1,158 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#![feature(compiler_builtins)]
+#![compiler_builtins]
```
cffs, line 35: The rust documentation says this feature is internal to rustc and should not be used. What does it do in this context?

lib/ukrust/compiler_builtins.rs
```diff
@@ -0,0 +1,158 @@
...
+            #[doc(hidden)]
+            #[no_mangle]
+            pub extern "C" fn $ident() {
+                panic!($reason);
+            }
+        )*
+    }
+);
+
+define_panicking_intrinsics!("`f32` should not be used", {
```
cffs, line 57: Many of those intrinsics are defined by [compiler-rt](https://github.com/unikraft/lib-compiler-rt), which makes this lib incompatible with C++ and/or cargo-based rust code. We can work on that later on, but why is it necessary to define those intrinsics here?
It is the same for the 64-bit ones.

lib/ukrust/kernel/bindings_helper.h
```diff
@@ -0,0 +1,57 @@
...
+#include <uk/ctors.h>
+#include <uk/page.h>
+#include <uk/argparse.h>
+#include <uk/plat/console.h>
+
+void *rust_sys_malloc(size_t size);
+void rust_sys_free(void *ptr);
+void *rust_sys_realloc(void *ptr, size_t size);
+void *rust_sys_calloc(int num, size_t size);
+void rust_uk_crash(void);
```
cffs: Maybe mark this one with `__attribute__((noreturn))`? This is not yet supported by bindgen, but it is expected at some point in the future.

lib/ukrust/target.json
```diff
@@ -0,0 +1,37 @@
+
+{
+  "arch": "x86_64",
```
cffs, line 3: Would the use of this target cause an explicit error if one tries to use it on a different arch? How to support other architectures (e.g. `arm64`)?

cffs (changes requested):
Good job, it works!
However, I would like to see at least one change before I can approve it (see `UK_PREPARE` in `Makefile.uk`), and I have some questions or comments that warrant further discussion.
We can also discuss it on Discord and post a summary here afterward if that is easier.
Thanks!

lib/ukrust/Config.uk
cffs: By the way, how come the crate is actually named `kernel`? Shouldn't we either name it `ukrust`, or make the name explicit in the doc?

lib/ukrust/kernel/lib.rs
TrueDoctor: These are just bindings for the unikraft allocator, you can than in your rust code decide which allocator you actually want to use.

cffs: I was afraid of the global allocator defined with `#[global_allocator]` in that module, but after checking the documentation, it seems one can have multiple global allocators, as long there is at most one per crate. So this should be fine. Thanks!

TrueDoctor: Ah sorry overlooked the actual macro, we should definitely test this, I had issues with multiple global alloc defenitions in the past

TrueDoctor: But tbh, this is most likely the desired behavior anyway, if you want to use a different allocator you would use it as the kernel allocator as well right?

lib/ukrust/compiler_builtins.rs
vladandrew: Without these we cannot link the `core` crate. In the future, I think that we could restrict the usage of these case if either compilert or cargo is used.

lib/ukrust/kernel/bindings_helper.h
vladandrew: Sounds good.

lib/ukrust/target.json
vladandrew: Right now this is x86_64 only. This will be fixed in future patches.

lib/ukrust/kernel/lib.rs
vladandrew: Yep, that's the plan with the allocators.

lib/ukrust/kernel/lib.rs
vladandrew: Nice find. I'll remove this feature.

lib/ukrust/Config.uk
vladandrew: `kernel` sounded more intuitive, but naming `ukrust` makes more sense in this context. I'll update accordingly.

lib/ukrust/kernel/lib.rs
vladandrew: Right, will remove the extra no_std

lib/ukrust/Makefile.uk
vladandrew: Thanks, good idea!

lib/ukrust/Makefile.uk
vladandrew: Yeah, I think it's best to leave this for a future patch series.

lib/ukrust/target.json
vladandrew: I've added a `depends on ARCH_X86_64` until we integrate it with arm.

lib/ukrust/kernel/lib.rs
vladandrew: Yeah, panic! should do the job.

lib/ukrust/kernel/lib.rs
vladandrew: Actually, I've changed it to `mod allocator`.

lib/ukrust/compiler_builtins.rs
vladandrew: I think this feature tells rust to no longer check for any compiler_builtins crate since all the symbols are defined in this file.

vladandrew:
Thanks a Cyril for the review. I've updated the pull request with the requested changes, the remaining comments which I haven't marked as solved(e.g. arm target) are good further work tasks will be solved in further

cffs (approved):
Seems good to me. Deferred for future patches:
- Compatibility with `compilers-rt` (avoid intrinsics clashes).
- Rename to `ukrust-sys` or similar, following rust convention.
- Import other functionalities already developed, but not yet included.

cffs:
Reviewed-by: Cyril Soldani <cyril.soldani@uliege.be>

TrueDoctor (approved):
Looks good to me

cffs (approved):
The rename looks fine to me.

lib/ukrust/target.json
```diff
@@ -0,0 +1,37 @@
...
+  "code-model": "kernel",
+  "cpu": "x86-64",
+  "data-layout": "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128",
+  "disable-redzone": true,
+  "emit-debug-gdb-scripts": false,
+  "env": "gnu",
+  "features": "-mmx,-sse,-sse2,-sse3,-ssse3,-sse4.1,-sse4.2,-3dnow,-3dnowa,-avx,-avx2,+soft-float",
```
skuenzer, line 10: I think the CPU features should be selected depending on the selcted target CPU. We risk that we compile code always with AVX which would us prevent to run on non-AVX CPUs - is this true?

lib/ukrust/Config.uk
```diff
@@ -0,0 +1,9 @@
+config LIBUKRUST
+	bool "ukrust: Rust support for Unikraft's internals"
+	depends on ARCH_X86_64
+	default n
```
skuenzer, line 5: At least this line needs to be added because libc headers are used in the glue code (like stddef)
```
```
Better would be even to compile with unikraft headers only - then this dependency can go away

skuenzer: We have Unikraft-only types, like `__64`, `__sz`, `__uptr`

lib/ukrust/target.json
```diff
@@ -0,0 +1,37 @@
...
+  "linker-flavor": "gcc",
+  "linker-is-gnu": true,
+  "llvm-target": "x86_64-elf",
+  "max-atomic-width": 64,
+  "needs-plt": true,
+  "os": "none",
+  "panic-strategy": "abort",
+  "position-independent-executables": true,
+  "pre-link-args": {
+    "gcc": [
```
skuenzer, line 22: Are there flags that we should take form the build system and feed into here?

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,64 @@
+$(eval $(call addlib_s,libukrust,$(CONFIG_LIBUKRUST)))
+
+CINCLUDES-$(CONFIG_LIBUKRUST) += -I$(LIBUKRUST_BASE)/ukrust_sys
+
+RUSTCFLAGS-y += --extern ukrust_sys -L $(LIBUKRUST_BUILD)/ --extern alloc \
```
skuenzer: Maybe we want to set the rust flags also only when this lib is included. Imagine someone bringing in an own replacement and doing similar stuff.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,64 @@
...
+# Build the ukrust_sys crate, contains the bindings and other useful functions
+$(LIBUKRUST_BUILD)/ukrust_sys.o: $(LIBUKRUST_BASE)/ukrust_sys/lib.rs $(LIBUKRUST_BUILD)/core.o $(LIBUKRUST_BUILD)/compiler_builtins.o
+	$(call build_cmd, BUILD, libukrust, $(notdir $@), \
+	$(RUSTC) $(RUST_UK_FLAGS) --extern alloc  \
+		--out-dir $(LIBUKRUST_BUILD) -L $(LIBUKRUST_BUILD) -A dead_code \
+		-A unused_doc_comments --crate-name ukrust_sys $(LIBUKRUST_BASE)/ukrust_sys/lib.rs)
+
+
+$(LIBUKRUST_BUILD)/alloc.o: $(RUST_LIB_SRC)/alloc/src/lib.rs
+	$(call build_cmd, BUILD, libukrust, $(notdir $@), \
```
skuenzer: Can you take `RUSTC` for the label instead of `BUILD`

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,64 @@
...
+		-A unused_doc_comments --crate-name core $(RUST_LIB_SRC)/core/src/lib.rs)
+
+$(LIBUKRUST_BUILD)/compiler_builtins.o: $(LIBUKRUST_BASE)/compiler_builtins.rs $(LIBUKRUST_BUILD)/core.o
+	$(call build_cmd, BUILD, libukrust, $(notdir $@), \
+	$(RUSTC) $(RUST_UK_FLAGS) \
+		--out-dir $(LIBUKRUST_BUILD) -L $(LIBUKRUST_BUILD) -A dead_code \
+		-A unused_doc_comments --crate-name compiler_builtins $(LIBUKRUST_BASE)/compiler_builtins.rs)
+
+UK_PREPARE-$(CONFIG_LIBUKRUST) += $(LIBUKRUST_BUILD)/bindings_generated.rs $(LIBUKRUST_BUILD)/ukrust_sys.o \
+		$(LIBUKRUST_BUILD)/alloc.o $(LIBUKRUST_BUILD)/core.o $(LIBUKRUST_BUILD)/compiler_builtins.o
```
skuenzer: Do these objects need to be compiled before compiling anything else in rust? If so, can you put a short comment explaining this. Otherwise, we could register the object files to `UK_OLIBS-$(CONFIG_LIBUKRUST)`

lib/ukrust/compiler_builtins.rs
skuenzer: Sounds like a little `/* TODO */` or `/* FIXME */` comment 😄

lib/ukrust/rust_glue.c
```diff
@@ -0,0 +1,62 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <stddef.h>
```
skuenzer: You can avoid stddef and a dependency to libc by using Unirkaft types:
`#include <uk/arch/types.h>`
and
`size_t` --> `__sz`

lib/ukrust/Config.uk
skuenzer: It would be great to treat dependency to `LIBUKALLOC` optional. Either return `ENOMEM` on library absence or exclude it from the build if possible.

skuenzer (changes requested):
Hey, I have some comments regarding the PR. I was basically looking into achieving minimal dependencies and avoiding conflicts with other libraries. But overall, this looks good! Thanks, Simon

lib/ukrust/target.json
skuenzer: btw, this file could be generated with `M4` or using the gcc preprocessor only

vladandrew: Actually, we instruct rustc not to use any SIMD related instructions hence the `-`, we only enable soft floats. I'm not sure if, at least right now, we want to generate this file. I'd leave it for a future pull request If the needs arise. Most project use it as this.

lib/ukrust/Makefile.uk
vladandrew: Makes sense. Will update it.

lib/ukrust/Makefile.uk
vladandrew: Yes, anything else uses the symbols defined in these objects. They're standalone since they depend on no crate.

lib/ukrust/rust_glue.c
vladandrew: Thanks, good idea!

lib/ukrust/Makefile.uk
vladandrew: Fair point, I'll update the pull request accordingly.

lib/ukrust/Config.uk
vladandrew: I don't think it makes sense to use this without LIBUKALLOC since you won't be able to use neither the core nor the alloc or bindings crate. In this case you'd have to use barebones rust.(what we have right now) I've removed the stddef include, we no longer need any libc.

lib/ukrust/target.json
vladandrew: Nothing I can think of.

lib/ukrust/Config.uk
TrueDoctor: Wait, why you need LIBUKALLOC for all that stuff?
Rust and core should work fine without an allocator. Everything that needs an allocator is in `alloc` afaik. Also not sure why that would have an impact on bindgen if we `#ifdef` the header inclusion

vladandrew:
Hi Simon,

Thank you for your review, I've solved most of the comments. Let me know if you spot anything else.

lib/ukrust/Config.uk
vladandrew: Oh, you're right, #ifdef should do the trick for bindgen, let me fix it quickly.

lib/ukrust/Config.uk
```diff
@@ -0,0 +1,9 @@
+config LIBUKRUST
+	bool "ukrust: Rust support for Unikraft's internals"
+	depends on ARCH_X86_64
```
skuenzer, line 3: You should also add the dependency to libukdebug because `uk/print.h` is not available without it.
`select LIBUKDEBUG`

lib/ukrust/ukrust_sys/bindings_helper.h
```diff
@@ -0,0 +1,53 @@
...
+#endif
+#ifdef CONFIG_LIBUKSCHED
+#include <uk/sched.h>
+#endif
+#include <uk/print.h>
+#include <uk/init.h>
+#include <uk/list.h>
+#include <uk/ctors.h>
+#include <uk/page.h>
+#include <uk/argparse.h>
```
skuenzer, line 61: Can you check that all headers that are provided by libraries are only included to the bindings if there are selected to be compiled, like:
```c
#if CONFIG_LIBUKARGPARSE
#include <uk/argparse.h>
#endif
```

lib/ukrust/ukrust_sys/bindings_helper.h
```diff
@@ -0,0 +1,53 @@
...
+#include <uk/plat/console.h>
+
+void *rust_sys_malloc(__sz size);
```
skuenzer: You may want to include `<uk/arch/types.h>` to be sure that the Unikraft types, like `__sz` are available and are not populated by another header that by chance includes that one indirectly.

lib/ukrust/rust_glue.c
```diff
@@ -0,0 +1,62 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <bindings_helper.h>
+#include <uk/alloc.h>
+#include <uk/essentials.h>
+ #include <uk/arch/types.h>
```
skuenzer: I guess this leading space came here by mistake ;-)

lib/ukrust/rust_glue.c
```diff
@@ -0,0 +1,62 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <bindings_helper.h>
+#include <uk/alloc.h>
```
skuenzer: This will fail if `libukalloc` is not included. However, `bindings_helper.h` should include `alloc.h` if it is enabled.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,67 @@
...
+		--target $(LIBUKRUST_BASE)/target.json \
+		-Zallow-features=allocator_api,bench_black_box,concat_idents,global_asm,try_reserve
+
+
+export RUST_BINDINGS_FILE=$(LIBUKRUST_BUILD)/bindings_generated.rs
+RUSTC_SYSROOT = $(shell $(RUSTC)  --print sysroot)
+RUST_LIB_SRC ?= $(RUSTC_SYSROOT)/lib/rustlib/src/rust/library
+
+# Common flags
+RUST_UK_FLAGS = --target=$(LIBUKRUST_BASE)/target.json --emit=dep-info,obj,metadata --edition=2018 \
```
skuenzer: Shall `RUST_UK_FLAGS` function as a global variable? If not, can you namespace it to a library-specific variable:
```Makefile
LIBUKRUST_COMMON_FLAGS = --target=<...>
```
For the global scope we started intorducing `RUSTC<..>` variable names with `support/build/Makefile.rules`. Unfortunately, our `Makefile.uk` files are operating on a single and global varaible space, this is why we need to be careful with variable naming to avoid clashes. Issues coming from this are very hard to notice.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,67 @@
...
+	-Cpanic=abort -Cembed-bitcode=n -Clto=n -Crpath=n \
+	-Cforce-unwind-tables=n -Ccodegen-units=1 \
+	-Zbinary_dep_depinfo=y -Zsymbol-mangling-version=v0 \
+	-Dunsafe_op_in_unsafe_fn -Drust_2018_idioms \
+	--crate-type rlib
+
+# Generate rust bindings for the Unikraft API
+$(LIBUKRUST_BUILD)/bindings_generated.rs: $(LIBUKRUST_BASE)/ukrust_sys/bindings_helper.h
+	$(call build_cmd, GEN, libukrust, $(notdir $@), \
```
skuenzer: Please call it `BINDGEN` instead of `GEN` because of the tool that you are calling.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,67 @@
...
+		bindgen -o $(LIBUKRUST_BUILD)/bindings_generated.rs --use-core --ctypes-prefix crate::c_types \
+		--with-derive-default --no-debug '.*' --size_t-is-usize \
+		$(LIBUKRUST_BASE)/ukrust_sys/bindings_helper.h -- $(CINCLUDES) \
+		$(CINCLUDES-y) -nobuiltininc -DMODULE)
+
+# Build the ukrust_sys crate, contains the bindings and other useful functions
+$(LIBUKRUST_BUILD)/ukrust_sys.o: $(LIBUKRUST_BASE)/ukrust_sys/lib.rs $(LIBUKRUST_BUILD)/core.o $(LIBUKRUST_BUILD)/compiler_builtins.o
+	$(call build_cmd, RUSTC, libukrust, $(notdir $@), \
+	$(RUSTC) $(RUST_UK_FLAGS) --extern alloc  \
+		--out-dir $(LIBUKRUST_BUILD) -L $(LIBUKRUST_BUILD) -A dead_code \
```
skuenzer, line 33: One question to this `-A dead_code` option. Could it make sense to bind its presence to `CONFIG_OPTIMIZE_DEADELIM` or is it someting different? When using GCC with C code, the linker ignores issues for dead code. For development reasons you sometimes want to enforce complete coverage for the linker - even if you have just a simple Hello world. This is why we have this as an option.

lib/ukrust/ukrust_sys/bindings_helper.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+/* Unikraft API */
+#ifdef CONFIG_LIBUKALLOC
+#include <uk/alloc.h>
+#endif
+#ifdef CONFIG_LIBUKSCHED
+#include <uk/sched.h>
+#endif
+#include <uk/print.h>
```
skuenzer, line 44: You may want to add `uk/hexdump.h`, `uk/trace.h` and `uk/assert.h` as well?
More unikraft-native features from LIBUKDEBUG would then be exported.

skuenzer: For the reason of consistency, I recommend to add these header includes only if ukdebug is available:
```c
#if CONFIG_LIBUKARGPARSE
#include <uk/assert.h>
#include <uk/print.h>
#include <uk/hexdump.h>
#include <uk/trace.h>
#endif
```

lib/ukrust/ukrust_sys/bindings_helper.h
```diff
@@ -0,0 +1,53 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
```
skuenzer, line 35: You should include `<uk/config.h>` first, otherwise, the switches don't work as expected.

skuenzer: Don't you want to add an include guard for this header file (style reason)? ...
```c
#ifndef __UKRUST_BINDINGS_HELPER
#define __UKRUST_BINDINGS_HELPER
/**/
#endif /* __UKRUST_BINDINGS_HELPER */
```

lib/ukrust/rust_glue.c
```diff
@@ -0,0 +1,62 @@
...
+#include <uk/essentials.h>
+ #include <uk/arch/types.h>
+
+void *rust_sys_malloc(__sz size)
```
skuenzer: I think it would be clean to return `NULL` if there is no allocator that we could ask for memory. If we have a LIBC that does not use ukalloc, we could forward the calls to libc. I suggest to implement the compile-time check if `LIBUKALLOC` is there, I leave it to you if you want to include the libC case.
```c
#if CONFIG_LIBUKALLOC
#include <uk/alloc.h>
#elif CONFIG_HAVE_LIBC
#include <stdlib.h>
#else /* !CONFIG_LIBUKALLOC && !CONFIG_HAVE_LIBC */
#include <uk/arch/types.h>
#endif

#if CONFIG_LIBUKALLOC
void __ukrust_sys_malloc(__sz size)
{
   return uk_malloc(uk_alloc_get_default(), size);
}

/* <..> */
#elif CONFIG_HAVE_LIBC
void __ukrust_sys_malloc(__sz size)
{
   return malloc(size);
}

/* <..> */
#else /* !CONFIG_LIBUKALLOC && !CONFIG_HAVE_LIBC */
void __ukrust_sys_malloc(__sz size)
{
   return __NULL;
}

/* <..> */
#endif
```
I also suggest to prefix all C symbols of your glue code with `__ukrust_` or `__ukrust_sys_`. Since this library does not have an `exportsyms.uk` file and you do not use `static`, all these symbols are on global scope. This is to avoid potential name clashes.

lib/ukrust/target.json
skuenzer: Okay, this is fine. I assume, as soon more archs are added to ukrust, we have multiple files, one per arch.

lib/ukrust/Makefile.uk
```diff
@@ -0,0 +1,67 @@
+$(eval $(call addlib_s,libukrust,$(CONFIG_LIBUKRUST)))
+
+CINCLUDES-$(CONFIG_LIBUKRUST) += -I$(LIBUKRUST_BASE)/ukrust_sys
+
+RUSTCFLAGS-$(CONFIG_LIBUKRUST) += --extern ukrust_sys -L $(LIBUKRUST_BUILD)/ --extern alloc \
+		--target $(LIBUKRUST_BASE)/target.json \
+		-Zallow-features=allocator_api,bench_black_box,concat_idents,global_asm,try_reserve
+
+
+export RUST_BINDINGS_FILE=$(LIBUKRUST_BUILD)/bindings_generated.rs
```
skuenzer: `export LIBUKRUST_BINDINGS=<...>`? ... see my next comment why I am suggesting this.

lib/ukrust/Config.uk
```diff
@@ -0,0 +1,9 @@
...
+	default n
+	help
+	  This library provides all the necessary functionalities
+	  for writing Rust internal libraries. It exports Rust
```
skuenzer: How about this change `for writing Rust internal libraries` --> `for Rust libraries that utilize Unikraft's native API`? In our terminology, an internal library is directly located in the Unikraft source tree and I think you use this also for external libraries that are written in Rust and utilize Unikraft APIs.

lib/ukrust/ukrust_sys/bindings_helper.h
```diff
@@ -0,0 +1,53 @@
...
+/* Unikraft API */
+#ifdef CONFIG_LIBUKALLOC
```
skuenzer: `#if CONFIG_UKALLOC` is in general safer. Imagine the case `CONFIG_UKALLOC` is defined but set to `0`.

skuenzer (changes requested):
Hi Vlad,
I have added some more comments. Most of them are concerning namespacing and ukalloc. Sorry for being picky with some of my comments but I think this library should be in a very good shape so that it becomes the example for future projects for other languages. Btw, is the library really working as expected if you compile without `lib/ukalloc`? The reason why I am asking is that I still saw some references to `uk/alloc.h` that are always included although we compile without an allocator.
Thanks,
Simon

lib/ukrust/Makefile.uk
vladandrew: Hi Simon,
`-A dead_code` enables dead code warnings, without this we would get compiler error for dead code. So we shouldn't bind it to `CONFIG_OPTIMIZE_DEADELIM`. Thanks!

vladandrew:
Hi Simon,

Thank you for reviewing this pull request so thoroughly. It's really important that we have a strong starting point for Rust. I have updated the pull request with the requested changes. Let me know if anything else is required.

skuenzer (changes requested):
Hey Vlad,
thanks for this update. Looks good! Can you also add a little description to the `TODO` comment in `compiler_builtin.rs`? ...then I think we are good to go.
Thanks,
Simon

vladandrew:
Hi Simon,

Thank! I've added the comment.

nderjung:
Hey, just chipping in here with regarding linting.

At the moment, `checkpatch.pl` does not yet support linting Rust code (and so this current PR won't be checked).  As far as  I can tell from the Rust-for-Linux work (of which `checkpatch.pl` derives) I have found some initial comments:

https://lore.kernel.org/lkml/CANiq72khBa2GcB6-PHM3A44Y90d6vzYAS=BVpk3nT4B6u+NVDw@mail.gmail.com/T/
```
>  # All the preparing..
>  prepare: prepare0 prepare-objtool prepare-resolve_btfids
> @@ -1648,6 +1727,13 @@ help:
>         @echo  '  kselftest-merge   - Merge all the config dependencies of'
>         @echo  '                      kselftest to existing .config.'
>         @echo  ''
> +       @echo  'Rust targets:'
> +       @echo  '  rustfmt         - Reformat all the Rust code in the kernel'

Seems like a good way for drive by commits where someone reformatted
the whole kernel.

> +       @echo  '  rustfmtcheck    - Checks if all the Rust code in the kernel'
> +       @echo  '                    is formatted, printing a diff otherwise.'

Might be nice to invoke this somehow from checkpatch.pl somehow for
changes to rust source files. Not necessary in the RFC, but perhaps
one day.
```

I am not sure 100% the current stance with the Linux work, but it from a Unikraft perspective, it makes sense to include the `rustfmt` program within the `checkpatch.pl` program when it detects `.rs` files.  For this PR, however, it is fine to simply run this locally and ensure the code style and formatting.

Suggestions for future integration are welcome, and the creation of a separate issue (and then PR) to integrate this work.  Not sure if I am missing anything, however, at this point I just need to check that the PR is in a good working format regarding the consensus of rust style (since it acts as reference code for all future integrations).

skuenzer (approved):
Thanks for the updates! Looks good to me, now.

Approved-by: Simon Kuenzer <simon@unikraft.org>

PR #359, opened 2021-12-05
support/scripts: Allow `mkcpio` to quietly gzip compress CPIO images

aadhi0319:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Users may sometimes require compressed CPIO images. With the current
`mkcpio` script, the only way to do this is to create a regular CPIO
image and then gzip it separately. This is inefficient and unnecessary,
as `bsdcpio` natively allows for gzip compression. Furthermore, some
users may wish to suppress the output generated by the `mkcpio` script.
This commit offers new command line arguments that allow the user to
satisfy both of these requirements.

### DIFF
--- a/support/scripts/mkcpio
+++ b/support/scripts/mkcpio
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
 
 usage()
 {
-	echo "$0 [output.cpio] [CPIO root path]" 1>&2
+	echo "$0 [options] [output.cpio] [CPIO root path]" 1>&2
+	echo "-h	display this help screen" 1>&2
+	echo "-q	suppress non-critical output" 1>&2
 	exit 1
 }
 
+# Process extra options
+CPIOFLAGS=()
+while getopts ":hq" option; do
+	case $option in
+		h)
+			usage
+			;;
+		q)
+			OPT_QUIET=true
+			;;
+		?)
+			usage
+			;;
+	esac
+done
+
+if [ "$OPT_QUIET" = true ]; then
+	CPIOFLAGS+=("--quiet")
+else
+	CPIOFLAGS+=("-v")
+fi
+
+shift $((OPTIND - 1))
+
+# Process primary arguments
 [ -z "$1" -o -z "$2" ] && usage
 if [ -e "$1" -a ! -f "$1" ]; then
 	echo "Output '$1' does already exist and is not a file" 1>&2
@@ -29,8 +56,9 @@ if [ -z "${CPIOTOOL}" ]; then
 	exit 1
 fi
 
+# Generate CPIO file
 cd "$2"
 find . -depth \( ! -iname ".gitkeep" \) -print \
 	| tac \
-	| "${CPIOTOOL}" -v -o --format newc > "${OUT_BASE}/${OUT_FILE}"
+	| "${CPIOTOOL}" "${CPIOFLAGS[@]}" -o --format newc > "${OUT_BASE}/${OUT_FILE}"
 exit $?


craciunoiuc:
Hey @aadhi0319,

Thanks for your contribution! Before accepting it I have a couple of questions:
 * You mention the "quiet" flag (`-q`), but it's not actually shown inside the "help" prompt. Is this intentional?
 * The default usage uses the verbose flag. Should this `-v` be added inside the "help" prompt?
 I'm thinking of the case where a user would assume that it's not on by default.

craciunoiuc (changes requested):
Hey @aadhi0319,

Thanks for your contribution! Before accepting it I have a couple of questions:
 * You mention the "quiet" flag (`-q`), but it's not actually shown inside the "help" prompt. Is this intentional?
 * The default usage uses the verbose flag. Should this `-v` be added inside the "help" prompt?
 I'm thinking of the case where a user would assume that it's not on by default.

aadhi0319:
Hello @craciunoiuc!
* The quiet flag was a mistake on my part...I forgot to add it into the help prompt. I've added it in and you should see it with the next commit to this branch.
* I'm not sure I understand what you mean, so please correct me if I misunderstood. For bsdcpio, -v outputs the filenames that are being compressed. The number of blocks that have been written to the archive are also outputted regardless of the -v flag. --quiet suppresses writing the number of blocks written to the archive to stdout. For my use case, I wanted -q to output nothing (so -v was not set and --quiet was set). Not using -q sets -v and does not set --quiet, so the user will see the filenames being archived and the number of blocks written to the archive. Do you want me to split this into a separate -v and -q flags so that -v controls filenames being outputted and -q controls blocks written being outputted? Further, currently this script outputs the filenames and blocks written by default. As such, if we change the default behavior to be silent, this may require some programs that currently use this script to be altered.

craciunoiuc:
> Hello @craciunoiuc!
> 
>     * I'm not sure I understand what you mean, so please correct me if I misunderstood. For bsdcpio, -v outputs the filenames that are being compressed. The number of blocks that have been written to the archive are also outputted regardless of the -v flag. --quiet suppresses writing the number of blocks written to the archive to stdout. For my use case, I wanted -q to output nothing (so -v was not set and --quiet was set). Not using -q sets -v and does not set --quiet, so the user will see the filenames being archived and the number of blocks written to the archive. Do you want me to split this into a separate -v and -q flags so that -v controls filenames being outputted and -q controls blocks written being outputted? Further, currently this script outputs the filenames and blocks written by default. As such, if we change the default behavior to be silent, this may require some programs that currently use this script to be altered.

Hey,

Thanks for the verbose description :wink: . I was only curious about the default behaviour of `bsdcpio`.
There is no need to change anything.

I will add the `reviewed` as soon as you add the change :smiley:

craciunoiuc:
> Hello @craciunoiuc!
> 
>     * I'm not sure I understand what you mean, so please correct me if I misunderstood. For bsdcpio, -v outputs the filenames that are being compressed. The number of blocks that have been written to the archive are also outputted regardless of the -v flag. --quiet suppresses writing the number of blocks written to the archive to stdout. For my use case, I wanted -q to output nothing (so -v was not set and --quiet was set). Not using -q sets -v and does not set --quiet, so the user will see the filenames being archived and the number of blocks written to the archive. Do you want me to split this into a separate -v and -q flags so that -v controls filenames being outputted and -q controls blocks written being outputted? Further, currently this script outputs the filenames and blocks written by default. As such, if we change the default behavior to be silent, this may require some programs that currently use this script to be altered.

Hey,

Thanks for the verbose description :wink: . I was only curious about the default behaviour of `bsdcpio`.
There is no need to change anything.

craciunoiuc:
Hey @aadhi0319 @skuenzer,

The PR looks good now.

skuenzer:
:thinking: The `quiet` option is fine - I would approve this. However, compressed (gzip) cpio archives aren't supported by Unikraft at the moment. I would delay that option until Unikraft supports it. Alternatively, we would need to document it (docs or tool help). I am wondering what you guys think because the intention of the script was to prepare a CPIO in the special format as Unikraft's `ukcpio` library currently needs...

support/scripts/mkcpio
```diff
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
 
 usage()
 {
-	echo "$0 [output.cpio] [CPIO root path]" 1>&2
+	echo "$0 [options] [output.cpio] [CPIO root path]" 1>&2
+	echo "-h	display this help screen" 1>&2
+	echo "-q	suppress non-critical output"
```
skuenzer: I guess a 1>&2 is missing here

support/scripts/mkcpio
```diff
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
...
+	echo "-z	compress [output.cpio] using gzip to produce [output.cpio].gz" 1>&2
 	exit 1
 }
 
+# Process extra options
+FLAGS=""
```
skuenzer: From a style perspective, I would prefer building an Bash array:
```bash
CPIOFLAGS=()
```

skuenzer: This is to get better control of argument splitting...

support/scripts/mkcpio
```diff
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
...
+while getopts ":hqz" option; do
+	case $option in
+		h)
+			usage
+			;;
+		q)
+			FLAGS+="--quiet "
```
skuenzer: You add an argument then with
```bash
CPIOFLAGS+=("--quiet")
```

support/scripts/mkcpio
```diff
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
...
+			;;
+		z)
+			FLAGS+="-z "
```
skuenzer:
```bash
CPIOFLAGS+=("-z")
```

support/scripts/mkcpio
```diff
@@ -29,8 +56,9 @@ if [ -z "${CPIOTOOL}" ]; then
 	exit 1
 fi
 
+# Generate CPIO file
 cd "$2"
 find . -depth \( ! -iname ".gitkeep" \) -print \
 	| tac \
-	| "${CPIOTOOL}" -v -o --format newc > "${OUT_BASE}/${OUT_FILE}"
+	| "${CPIOTOOL}" ${FLAGS} -o --format newc > "${OUT_BASE}/${OUT_FILE}"
```
skuenzer: To keep proper quoting of the individual arguments, you can then let bash expand the list variable with the `@` paramter:
```bash
"${CPIOFLAGS[@]}"
```

support/scripts/mkcpio
```diff
@@ -3,10 +3,37 @@ CPIOTOOL="$( which bsdcpio )"
...
+			;;
+		z)
+			FLAGS+="-z "
+			;;
+		?)
+			usage
+			;;
+	esac
+done
+if [[ $FLAGS != *"--quiet"* ]]; then
```
skuenzer: Maybe it is more efficient to work with an boolean like variable that you set within the case:
```bash
OPT_QUIET=1
# [...]

while getopts ":hqz" OPT; do
	case $OPT in
		# [...]
		q)
			OPT_QUIET=0
			;;
		# [...]
	esac
done

if [ $OPT_QUIET -eq 0 ]; then
	CPIOFLAGS+=("--quiet")
else
	CPIOFLAGS+=("-v")	
fi
```

skuenzer (commented):
Hi Aadhithya,
while checking the commit I found only one missing redirection to stderr (`1>&2`) on line 8. The other comments are just style suggestions from my side, nothing critical and not mandatory to change. I am also wondering if we should rather be in quiet mode as default (as like the default behavior for every Unix command) and provide a verbose option `-v`. But this is probably also just a matter of preference - what do you think?
Thanks,
Simon

support/scripts/mkcpio
aadhi0319: The changes to this script were made to facilitate the grub build script. I required a quiet mode and I thought a compressed cpio image would be better. As for now, I have not figured out how to get a compressed cpio image to work with multiboot anyways, so the compression option is never used. For that reason, I think it would be fine to delay the compressed option until later. As for a a default quiet behavior, I didn't want to break the script for anyone that assumed a list of files would be outputted by default. I wasn't sure who used the script or how they used it, but if it doesn't break anything I think quiet by default would be preferred to maintain consistency with other Unix commands.

support/scripts/mkcpio
aadhi0319: I didn't know bash could expand arrays like that thanks for the tip! I agree with all the other style suggestions and will get a commit in with those fixes today or tomorrow.

craciunoiuc:
Hey @aadhi0319,

The PR looks good now. One mention: you have 2 commits with the same name that do different things. I suppose you wanted to 'squash' them together into a single one.

If yes, you can just:
```
git reset --soft HEAD~2
git commit -s -m "..."

**Review by craciunoiuc — CHANGES_REQUESTED** [review]
Hey @aadhi0319,

The PR looks good now. One mention: you have 2 commits with the same name that do different things. I suppose you wanted to 'squash' them together into a single one.

If yes, you can just:
```
git reset --soft HEAD~2
git commit -s -m "..."
git push --force
```

**Comment by aadhi0319** [comment]
@craciunoiuc Thanks! Just squashed the commits.

**Review by craciunoiuc — APPROVED** [review]
All good now!

**Comment by skuenzer** [comment]
The PR looks fine so far. Probably you should update your commit message and title to be more to the point. How about something like:
```
support/scripts: mkcpio: Option for quietly creating archives

This commit introduces a `quiet` option to `mkcpio`. This option
suppresses standard output while creating an archive. It can be
useful when `mkcpio` is called from other scripts or Makefiles
(e.g., grub tools).
```

aadhi0319:
@skuenzer I pushed again with an updated commit message.

skuenzer (approved):
Thanks a lot for your work!

PR #367, opened 2021-12-11
lib/all reduce checkpatch errors and warnings on lib directory

snemana232:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Clean up checkpatch warnings/errors in lib directory, resolve issue #199

### DIFF
--- a/lib/9pfs/9pfs_vnops.c
+++ b/lib/9pfs/9pfs_vnops.c
@@ -186,8 +186,7 @@ static int uk_9pfs_open(struct vfscore_file *file)
 
 	fd->fid = openedfid;
 	file->f_data = fd;
- 	UK_9PFS_ND(file->f_dentry->d_vnode)->nb_open_files++;
-
+	UK_9PFS_ND(file->f_dentry->d_vnode)->nb_open_files++;
 	return 0;
 
 out_err:
--- a/lib/fdt/fdt_ro.c
+++ b/lib/fdt/fdt_ro.c
@@ -177,7 +177,8 @@ int fdt_subnode_offset_namelen(const void *fdt, int offset,
 int fdt_subnode_offset(const void *fdt, int parentoffset,
 		       const char *name)
 {
-	return fdt_subnode_offset_namelen(fdt, parentoffset, name, strlen(name));
+	return fdt_subnode_offset_namelen(fdt, parentoffset, name,
+	strlen(name));
 }
 
 int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
@@ -212,7 +213,7 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
 				return offset;
 		}
 		q = memchr(p, '/', end - p);
-		if (! q)
+		if (!q)
 			q = end;
 
 		offset = fdt_subnode_offset_namelen(fdt, offset, p, q-p);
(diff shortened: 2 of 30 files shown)


flpostolache:
The code looks good. I have also tested some apps with this PR and everything works ok. I don't really know if in this case all the commits should be squashed into one, like it says in the [contribution guidelines](https://docs.unikraft.org/contribute.html) or it should be one commit per modified library, like @skuenzer mentions in the issue #199 . @nderjung can you give your opinion, please?

michpappas:
Hi @snemana232, lib/fdt should be probably be excluded from checks (and the corresponding fixes), as the idea is to use libfdt unmodified from [its home repository](https://git.kernel.org/pub/scm/utils/dtc/dtc.git/tree/libfdt). To be able to periodically fetch updates with minimal maintenance overhead, the sources under lib/fdt must be modified as little as possible. #378 does the required clean up and separation of UK changes from the upstream code. Changes in libfdt should be done via the upstream path.

skuenzer:
Hey @snemana232 ,

thanks a lot for your work. I really like clean-up code but in general, we need to be a bit careful with integrating this, because:
* Many outstanding PRs might break. So we would need to upstream something like this slowly and partially.
* Some incorrect styles where sometimes introduced on purpose, especially macro definitions. `git blame` can give hints if we should really modify it.
* As @michpappas was pointing out, with `libfdt` we have some externally copied code with an own style convention. Those should not be adopted at all because we want to be able to sync them with upstream from time to time.

I would suggest the following for moving forward:
* Can you split your changes into per-library commits?
* Have a second look to macros. They can be tricky: Missing or added white spaces can be sometimes there on purpose. Some other times checkpatch complains because macros need to be encapsulated in `do { } while(0)` expressions to avoid problems with short if statements (overall it always depends on the purpose of the macro). Have a look to this example:
```c
#define dprint(x) if (DEBUG) uk_pr_debug(x)

/* Problematic usage of this macro: */
if (error)
   dprint("Error happened\n");
else
   dprint("Everything is fine\n");

/* Where does `else` belong to? */
```
* Can you provide a separate PR for your checkpatch adoptions? I really like them. The thing we should change is that using `Reviewed-by:` is a separate case. We should print a warning (instead of an error) that asks to confirm if an offline review really happened. @nderjung, @razvand : What do you think?

Thanks,

Simon

razvand:
@snemana232 , can you still work on this PR and integrate @skuenzer 's comments? Or do you want someone else to pick up the work?

nderjung:
Hi,

I am going to close this PR.  Sorry that we've left it lingering for so long.  Ultimately, the size of the PR and the amount of changes necessary to get this upstream is too big and time consuming.  My suggestion is to break this task up into a lot of smaller PRs so I'm closing in favour of solving the original issue like this, where each PR also has a short description describing the fix itself rather than some large "blanket" commit which can be less informative.  Not only does the PR not solve the issue that it's addressing but the PR itself has issues (the size of the checkpatch itself is an indicator here) but a). each "fix" may have some side-effect and we'll be able to quickly identify any problems that may occur when an individual error is fixed with individual PRs (aka if you fix one of the errors/warnings stated in the original issue, it will have reprocussions and these need to be solved with regard to that fix) and b). by splitting it up into many PRs, we can use the CI/CD system to partially check that for us by testing all supported archs/platforms.

PR #368, opened 2021-12-12
arch/arm64: Do not pass -march with -mcpu

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://docs.unikraft.org/contribute.html) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): arm/arm64
 - Platform(s): kvm
 - Application(s): N/A

### Additional configuration

Support for several new CPUs added for arm/arm64:
* CONFIG_MCPU_ARM64_CORTEX_A34
* CONFIG_MCPU_ARM64_CORTEX_A35
* CONFIG_MCPU_ARM64_CORTEX_A53
* CONFIG_MCPU_ARM64_CORTEX_A57_A53
* CONFIG_MCPU_ARM64_CORTEX_A65
* CONFIG_MCPU_ARM64_CORTEX_A65AE
* CONFIG_MCPU_ARM64_CORTEX_A72
* CONFIG_MCPU_ARM64_CORTEX_A72_A53
* CONFIG_MCPU_ARM64_CORTEX_A73
* CONFIG_MCPU_ARM64_CORTEX_A73_A35
* CONFIG_MCPU_ARM64_CORTEX_A73_A53
* CONFIG_MCPU_ARM64_CORTEX_A75
* CONFIG_MCPU_ARM64_CORTEX_A76
* CONFIG_MCPU_ARM64_CORTEX_A76AE
* CONFIG_MCPU_ARM64_CORTEX_A75_A55
* CONFIG_MCPU_ARM64_CORTEX_A76_A55
* CONFIG_MCPU_ARM64_CORTEX_A77
* CONFIG_MCPU_ARM64_NEOVERSE_E1
* CONFIG_MCPU_ARM64_NEOVERSE_N1
* CONFIG_MCPU_ARM64_NEOVERSE_N2
* CONFIG_MCPU_ARM64_NEOVERSE_V1

### Description of changes

Currently declarations of ARCHFLAGS tie architecture revisions to CPU models. That is `-march` and `-mcpu` are declard together, based on the processor type selected by the user:
```    
    ifeq ($(CONFIG_MCPU_ARM64_CORTEXA55),y)
    ...
[...]

PR #369, opened 2021-12-12
arch/arm/arm64: Introduce Pointer Authentication support

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): common
 - Application(s): N/A

### Additional configuration

This PR introduces `CONFIG_ARM64_FEAT_PAUTH` to enable Pointer Authentication support.

### Description of changes

Pointer Authentication (PAuth) allows signing and authenticating pointers to harden the system against classes of attacks that rely on the manipulation of pointers, such as ROP. Pointer Authentication Codes (PACs) are created by signing a pointer and a 64-bit modifier with an 128-bit key. The modifier is a value that is normally used to restrict the PAC to a specific context (eg the SP). Keys are stored in registerers.

The PAC is stored in the unused upper bits of the memory address, and can be later verified to ensure that the pointer has not been tampered.

The reference algorithm is [the QARMA block cipher](https://eprint.iacr.org/2016/444.pdf), but it is possible for architectures to use an IMPLEMENTATION DEFINED algorithm instead.
 
**Basic Instructions:**
* PAC*: Sign address
* AUT*: Authenticate address
* XPA*: Strip PAC from address (useful for things like unwinding)

**Combined Instructions:**
* RETA*:  Auth and return
* BRA*:   Auth and branch
* BLRA*:  Auth and branch-link
* ERETA*: Auth and return from exception
* LDRA*:  Auth and load

**Keys:**
* APIA_Key: Key A for signing instructions
* APIB_Key: Key B for signing instructions
* APDA_Key: Key A for signing data
* APDB_Key: Key B for signing data
* APGA_Key: Key A for generic use

For more background see:
* [The Arm Architecture Reference Manual Armv8, for Armv8-A architecture profile](https://developer.arm.com/documentation/ddi0487/ga)
* [Learn the architecture: Providing protection for complex software](https://developer.arm.com/documentation/102433/latest/)
* [Pointer Authentication on ARMv8.3](https://www.qualcomm.com/media/documents/files/whitepaper-pointer-authentication-on-armv8-3.pdf)

## Architecture support

**Armv8.3:**
* Introduce FEAT_PAuth as a mandatory feature.
* Introduce FEAT_PAuth2 as an optional feature in Armv8.3 and mandatory in
  Armv8.6. FEAT_PAuth2 changes the way the PAC is applied to the address
  (XOR vs replace).
* Introduce FEAT_FPAC, which allows to control the faulting behavior when authentication
  fails. FEAT_FPAC is an optional feature and depends on FEAT_PAuth2.

## GCC support

**GCC 7:**
* Introduce support for armv8.3-a
* Introduce `--msign-return-address=[none|non-leaf|all]`

The parameters passed to `--msign-return-address` are interpreted as:
* `none`: Do not sign return addresses
* `non-leaf`: Sign/auth the return address of non-leaf functions.
* `all`: Sign/auth the return address of non-leaf and leaf functions.

GCC implements the Basic Set of PAuth instructions using the HINT instruction, in the so-called NOP space. This allows executing protected binaries on older platforms that do not implement Armv8.3-a. In these platforms PAuth instructions
execute as NOP.

When `sign-return-address` is enabled without setting `-march=armv8.3-a`, GCC generates PACIASP / AUTIASP sequences that sign and authenticate the LR upon function entry and exit. In the following snippet notice the opcode of the
HINT instruction (0xd5032300) that GCC generates for PACIASP and AUTIASP.
```
0000000000400564 <main>:
  400564:       d503233f        paciasp
  400568:       a9bf7bfd        stp     x29, x30, [sp, #-16]!
  40056c:       910003fd        mov     x29, sp
  400570:       90000000        adrp    x0, 400000 <_init-0x3e8>
  400574:       9118c000        add     x0, x0, #0x630
  400578:       97ffffb6        bl      400450 <puts@plt>
  40057c:       52800000        mov     w0, #0x0
  400580:       a8c17bfd        ldp     x29, x30, [sp], #16
  400584:       d50323bf        autiasp
  400588:       d65f03c0        ret
```

When `-msign-return-address` is used along with `-march=armv8.3-a`, GCC
generates PACIASP / RETAA sequences instead. In the snippet below notice the
opcode of RETAA (0xd65f0bff) which is in the fused space (instructions in the
Combined Set do not have HINT implementations). Clearly, this code cannot be
executed in architectures earlier than Armv8.3-a.
```
0000000000400564 <main>:
  400564:       d503233f        paciasp
  400568:       a9bf7bfd        stp     x29, x30, [sp, #-16]!
  40056c:       910003fd        mov     x29, sp
  400570:       90000000        adrp    x0, 400000 <_init-0x3e8>
  400574:       9118a000        add     x0, x0, #0x628
  400578:       97ffffb6        bl      400450 <puts@plt>
  40057c:       52800000        mov     w0, #0x0
  400580:       a8c17bfd        ldp     x29, x30, [sp], #16
  400584:       d65f0bff        retaa
```

**GCC 9:**
* Introduce support for armv8.5-a
* Deprecate `-msign-return-address`
* Introduce `-mbranch-protection=[none|pac-ret{+leaf}|bti|standard]`
* Introduce `--enable-standard-branch-protection` as a short to `mbranch-protection=standard`

The parameters passed to `-mbranch-protection` are interpreted as:
* `none`: Disables all protections
* `pac-ret`: Enables PAuth for function returns on non-leaf functions. The
           `+leaf` modifier enables protection for leaf functions.
* `bti`: Enables Branch Target Identification
* `standard`: Enables all protections

**GCC 10:**
* Introduce the b-key addendum to pac-ret in `--mbranch-protection`.
  This allows using the APIB_Key instead of the APIA_Key when signing
  return pointers.

## Changes introduced in this series

This PR provides platforms with an API to initialize PAuth keys, and enable PAuth.

A platform can initialize PAuth as:
```
#ifdef CONFIG_ARM64_FEAT_PAUTH
	if (ukplat_pauth_enable())
		UK_CRASH("Pointer Authentication is not available");
#endif
```

To generate PAyth keys, it is required that platforms provide an implementation of the key generation function that uses an adequate source of randomness.
```
void ukplat_pauth_gen_key(__u64 *key_hi, __u64 *key_lo);
```

Platforms that support Armv8.5 can use the random number generation instructions introduced into the architecture (`RNDR`/`RNDRSS`). Others can initialize drivers of an HWRNG/TRNG, or request randomness from the TEE.

Since the above functions are used for the initialization of PAuth, it is required that GCC excludes them when generating PACIASP/RETAA sequences. Otherwise, once PAuth is enabled, authentication will fail upon return. A macro is provided to help overriding the global gcc settings per function:
```
#define __no_pauth __attribute__((target("branch-protection=none")))
```

This macro is used by `pauth_enable()`, and should be used by all functions generated by GCC in the boot chain, up to - and including - the caller of `pauth_init()`.

Additionally to the above, the following choices are made:

**Do not enable backwards compatibility mode.**
If `-march=armv8.3-a` is not set, it is not possible to initialize the keys.
GCC exits with the following error:
```
bash$ ~/toolchains/gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf/bin/aarch64-none-elf-gcc -mbranch-protection=standard /tmp/test.c
/tmp/ccx7K1nD.s: Assembler messages:
/tmp/ccx7K1nD.s:17: Error: selected processor does not support system register name 'apiakeyhi_el1'
```
I believe that Arm's intention behind compatibility mode was mostly to provide backwards compatibility on userspace libraries. Clearly on Unikraft this feature cannot be utilized, except from the case of linking with prebuilt libraries.
Because of that this PR sets `-march=armv8.3-a` when enabling `CONFIG_ARM64_FEAT_PAUTH`.

**Enable leaf protection by default**

Although it's not strictly required as leaf functions do not save the LR, it doesn't hurt to be there as an additional protection.

**Initialize only the PACIA_Key**

The motivation for that is minimizing the impact on boot time. Additional keys can be enabled later, if needed.

**Do not introduce additional APIs for sign / auth / strip**

The intention of this PR is to provide initial PAuth support. As such, its scope is limited to the signing and verification of the function return address. Nevertheless, Pointer Authentication provides the possibility to harden the system further by signing more pointers. Additional functionality can be introduced into the Unkraft API later, on demand.

### DIFF
--- a/arch/arm/arm/include/uk/asm/compiler.h
+++ b/arch/arm/arm/include/uk/asm/compiler.h
@@ -1,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) YYYY, Copyright Holder. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UK_ESSENTIALS_H__
+#error Do not include this header directly
+#endif
+
+/* Architecture-specific compiler definitions */
+
--- a/arch/arm/arm64/Config.uk
+++ b/arch/arm/arm64/Config.uk
@@ -145,6 +145,18 @@ choice
 			Compile and optimize for Neoverse V1 CPUs.
 endchoice
 
+menu "Armv8-A Extensions"
+	depends on MCPU_ARM64_NONE
+
+	config ARM64_FEAT_PAUTH
+	bool "Armv8.3 Pointer Authentication"
+	help
+		Enable signing and authentication of pointers. This
+		provides protections against classes of attacks that
+		rely on memory corruption, such as stack smashing and
+		ROP.
+endmenu
+
 config ARM64_ERRATUM_858921
 	bool "Workaround for Cortex-A73 erratum 858921"
 	default y
--- a/arch/x86/x86_64/include/uk/asm/compiler.h
+++ b/arch/x86/x86_64/include/uk/asm/compiler.h
@@ -1,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) YYYY, Copyright Holder. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UK_ESSENTIALS_H__
+#error Do not include this header directly
+#endif
+
+/* Architecture-specific compiler definitions */
+
--- a/plat/common/include/arm/arm64/pauth.h
+++ b/plat/common/include/arm/arm64/pauth.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Copyright (c) 2022, Michalis Pappas <mpappas@fastmail.fm>.
+ * Copyright (c) 2021, Michalis Pappas <mpappas@fastmail.fm>.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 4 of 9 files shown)


michpappas:
This PR depends on #368

craciunoiuc:
Hey @michpappas,

I did an initial pass of everything and the PR seems good!

I will come back to it to test it once the other PR is merged.

Cezar

michpappas:
Rebased and fixed a bug in the `__no_branch_prediction` macro in `plat/common/include/arm/arm64/pauth.h`.

plat/common/include/arm/arm64/pauth.h
```diff
@@ -0,0 +1,72 @@
...
+
+/* Macro to disable the global branch-protection settings for a specific
+ * function. This needs to be declared on any functions in the boot-chain
+ * up to the caller of ukplat_pauth_init() to avoid PAuth errors upon return.
+ */
+#if CONFIG_ARM64_FEAT_PAUTH
+#define __no_branch_protection __attribute__((target("branch-protection=none")))
+#else
+#define __no_branch_protection
+#endif
```
craciunoiuc: I don't really get the `#if` brace here.

1. There might be other cases where branch protection needs to be disabled apart from the `PAUTH` case

2. This macro is in the `pauth.h` file. It should be used only in the context of `pauth` anyway.

craciunoiuc (commented):
Hey @michpappas,

This PR looks good. I have just a small comment. After that I can approve it 🙂

plat/common/include/arm/arm64/pauth.h
michpappas: Hi @craciunoiuc, 

> 1. There might be other cases where branch protection needs to be disabled apart from the PAUTH case

Yes in general the `branch-protection` option can also be used for BTI, so my on-off logic was in the current scope of BTI not supported yet. But that should change soon, as I have a PR on BTI almost ready.
With BTI enabled the macro will have to include additional logic to take into account whether `CONFIG_ARM64_FEAT_BTI` is enabled or not, and change the value of `branch-protection` accordingly. Saying that, a more suitable name for that macro would be something like `_no_pauth_prot` or something similar.

> 2. This macro is in the pauth.h file. It should be used only in the context of pauth anyway.

Well, it depends, one could include the header unconditionally,  but overall I agree. Especially as GCC uses the same parameter for PAuth and BTI, this should probably be moved to a generic header.

I'll think about it a bit more as I work on the BTI PR. Will push an update soon.

michpappas:
## Changes in this update

**Config.uk:**
* Switch to a flat structure
* Update help message to a more generic description

**Makefile.uk:**
* Add comment about minimum GCC version
* Rework in anticipation of additional extensions (see #421)

**compiler.h:**
* Introduce header for architecture-specific compiler definitions.
* Move the `__no_branch_protection` macro to compiler.h.
* Rename the `__no_branch_protection` macro to `__no_pauth`.

**pauth.c:**
* Reduce API to a single function, `ukpauth_init()`

**pauth.h:**
* Add include guards
* Change the return type of `ukplat_gen_key()` to void
* Add note about using `__no_pauth` when declaring `ukplat_gen_key()`

**lcpu.h:**
* Use the _AC macro where needed

michpappas:
PR description and commit message updated based on latest changes.

arch/arm/arm64/Config.uk
```diff
@@ -145,6 +145,18 @@ choice
...
 
+menu "Armv8-A Extensions"
+	depends on MCPU_ARM64_NONE
+
+	config ARM64_FEAT_PAUTH
+	bool "Armv8.3 Pointer Authentication"
+	help
+		Enable signing and authentication of pointers. This
+		provides protections against classes of attacks that
+		rely on memory corruption, such as smash stacking and
```
craciunoiuc:
```suggestion
		rely on memory corruption, such as stack smashing and
```

craciunoiuc (changes requested):
Hello again @michpappas,

The PR looks good now, but before adding my tag, here are some typos from the second commit message 😅:

"stored in `registerers`. The PAC" -> "stored in registers. The PAC"
"been tampered." -> "been tampered with."
"Pointer `Authenticaiton`" ->"Pointer Authentication"
"`equired` to" -> "required to"

After this, we can finally close this PR 😆.

arch/arm/arm64/Config.uk
```diff
@@ -145,6 +145,18 @@ choice
 			Compile and optimize for Neoverse V1 CPUs.
 endchoice
 
+menu "Armv8-A Extensions"
+	depends on MCPU_ARM64_NONE
```
craciunoiuc, line 149: Another thing:

Why does this feature depend on `MCPU_ARM64_NONE`?
As far as I understand, doesn't this mean "no specific processor" or something like that?

michpappas: Hi @craciunoiuc, that is because if you choose a specific core, GCC enforces the architecture revision (eg Armv8-3) of that IP. On the other hand, Arm has not yet released IP that implements newer architectural revisions. For example, you won't find a Cortex-A that supports Armv8.3. There are however implementations of licensees, such as Apple, yet these cores are not known to GCC. Consequently, to be able to free-select your own configuration, either for an existing core that is not known to GCC, or an emulated core such as QEMU's virt, you need to use `-mcpu=none`. The changes in 1fadd78ae2b46f03c0b3eb807311ccd6ca4dd743 were made to address the very same issue.

michpappas:
@craciunoiuc, typos are fixed now

arch/arm/arm64/Config.uk
craciunoiuc: Understood 👍

craciunoiuc (approved):
All good!

@razvand your turn now 😉 

arch/arm/arm/include/uk/asm/compiler.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) YYYY, Copyright Holder. All rights reserved.
```
razvand, line 3: @michpappas , please add the date and your name here.

arch/x86/x86_64/include/uk/asm/compiler.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) YYYY, Copyright Holder. All rights reserved.
```
razvand, line 3: Proper date + name

razvand (commented):
@michpappas , this looks good. Before approving, please check the copyright information.

arch/x86/x86_64/include/uk/asm/compiler.h
craciunoiuc: I think this is like this because the files are empty and the person that will fill in code will write themselves there.

arch/arm/arm/include/uk/asm/compiler.h
craciunoiuc: @razvand same as the other comment

arch/x86/x86_64/include/uk/asm/compiler.h
michpappas: @razvand, what Cezar said. These are empty, so I'm not sure if it makes any sense to copyright to myself.

razvand (approved):

PR #370, opened 2021-12-13
include/essentials: Fix and introduce more range macros

skuenzer:
### Description of changes

Introduces two new range test macros:
 - `RANGE_ISEQUAL`: ...to test if two ranges are equal.
 - `RANGE_CONTAINS`: ...to test if a range is a superset of another one.

This PR also fixes also a corner case in the `RANGE_OVERLAP` macro that wasn't handled properly: Two non-overlapping but neighboring ranges were incorrectly count as overlapping.

### DIFF
--- a/include/uk/essentials.h
+++ b/include/uk/essentials.h
@@ -225,10 +225,13 @@ extern "C" {
  *  - The end of range 1 is within range 0
  *  - The start of range 1 is smaller than the start of range 0 while
  *    the end of range 1 is bigger than the end of range 0
+ *  NOTE: The expressions take into account that `base + len` points to
+ *        the first value that is outside of a range.
  */
 #define RANGE_OVERLAP(base0, len0, base1, len1)				\
 	(IN_RANGE((base1), (base0), (len0))				\
-	 || IN_RANGE((base1) + (len1), (base0), (len0))			\
+	 || ((((base1) + (len1)) > (base0))				\
+	     && (((base1) + (len1)) <= (base0) + (len0)))		\
 	 || (((base1) <= (base0))					\
 	     && (((base1) + (len1)) >= ((base0) + (len0)))))
 
--- a/include/uk/essentials.h
+++ b/include/uk/essentials.h
@@ -219,6 +219,19 @@ extern "C" {
 	(((val) >= (base)) && ((val) < (base) + (len)))
 
 /**
+ * Tests if range 0 contains range 1
+ * This is the case when both of the following conditions are true:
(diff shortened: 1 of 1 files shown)


include/uk/essentials.h
```diff
@@ -218,17 +218,40 @@ extern "C" {
...
+ * This is the case when both of the following conditions are true:
+ *  - The start of range 1 is within range 0
+ *  - The end of range 1 is within range 0
+ *  NOTE: The expressions take into account that `base + len` points to
+ *        the first value that is outside of a range.
+ */
+#define RANGE_CONTAIN(base0, len0, base1, len1)				\
+	(IN_RANGE((base1), (base0), (len0))				\
+	 && (((base1) + (len1)) > (base0))				\
+	 && (((base1) + (len1)) <= ((base0) + (len0))))
```
razvand, line 239: Wouldn't it be more expressive if these were replaced with the construct below?
```
IN_RANGE((base1) + (len1) - 1, (base0), (len0)))
```

include/uk/essentials.h
```diff
@@ -218,17 +218,40 @@ extern "C" {
...
  *  - The start of range 1 is smaller than the start of range 0 while
  *    the end of range 1 is bigger than the end of range 0
+ *  NOTE: The expressions take into account that `base + len` points to
+ *        the first value that is outside of a range.
  */
 #define RANGE_OVERLAP(base0, len0, base1, len1)				\
 	(IN_RANGE((base1), (base0), (len0))				\
-	 || IN_RANGE((base1) + (len1), (base0), (len0))			\
+	 || ((((base1) + (len1)) > (base0))				\
+	     && (((base1) + (len1)) <= (base0) + (len0)))		\
```
razvand, line 254: Why not use the construction below?
```
IN_RANGE((base1) + (len1) - 1, (base0), (len0))
```

razvand (commented):
I added some comments with alternatives to macros.

include/uk/essentials.h
skuenzer: I was in fact also thinking about this but I wanted to avoid to imply any data type for the arguments. As soon as I subtract or add an int 1, this would not work anymore for floating point, for instance.

include/uk/essentials.h
skuenzer: ...same here... 😄

razvand (approved):
All good.

nderjung (approved):
Great, thanks! 🐒

Approved-by: Alexander Jung <alex@unikraft.org>

PR #385, opened 2022-01-16
support/scripts: Use more portable bash shebang

WesleyAC:
/bin/bash is not guaranteed to exist, and does not exist, for instance
on NixOS. In contrast, /usr/bin/env is part of the POSIX spec, and
should exist on all systems.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://docs.unikraft.org/contribute.html) regarding submitting new changes to the project; (I read an [archived copy](https://web.archive.org/web/20211204234937/https://docs.unikraft.org/contribute.html), since the main site appears to be down — `curl` reports: `curl: (7) Failed to connect to docs.unikraft.org port 443 after 305 ms: Connection refused`)
 - [x] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR; (I was unable to figure out how to run checkpatch on a patch that's destined for a GitHub PR rather than an email. Perhaps I'm not supposed to use GitHub at all? But it seems like many people do...)
 - [x] Updated relevant documentation. (no documentation updates needed)

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

None

### Description of changes

Use `#!/usr/bin/env bash` instead of `#!/bin/bash`, since it's more portable and allows building on NixOS.

PR #252, opened 2021-07-12
doc: Migrate and update docs

nderjung:
This commit removes the (now very outdated) `docs/` folder from the repository and `CODING_STYLE.md` file.  Towards a greater effort of keeping the docs up-to-date and in-line with the growing Unikraft ecosystem, the location of these documents have migrated to:

  https://github.com/unikraft/docs.git

This standalone repository consolidates documentation across the project, including the core of which docs used to be located.

References to the documentation and how to contribute are updated in this commit to reflect this change.

PR #388, opened 2022-01-24
doc/contributing: Add note about GitHub workflow.

WesleyAC:
I don't know if there's some official stance on this that I'm missing, but this is what the situation seems to be from my experience in #385?


nderjung:
Hi @WesleyAC,

Thanks for the PR!

We are launching a new docs site in 2 weeks to coincide with FOSDEM'22.  The new documentation site has extensive detail on the workflow process.

Thanks again, but I'll close this PR.  In the next two weeks, we'll be merging #252 and open-sourcing a new `docs` repo.

PR #394, opened 2022-02-10
lib/ukmmap: Fix some bugs and add mprotect

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

`CONFIG_LIBUKMMAP=y`

### Description of changes

This PR fixes two bugs:
* Callers of `mmap` expect anonymous memory to be zeroed. The patch adds a `memset` to do so.
* The `munmap` call tries to remap remaining memory to a smaller area. However, the caller won't have a chance to be informed about the changing base address of the area. Changing the base address also violates POSIX. As we are not able to free parts of the memory area, we have to keep it unchanged and silently fail. We only free it, if the caller unmaps the whole area.

In addition, the PR introduces a stub for `mprotect`.

### DIFF
--- a/lib/ukmmap/Makefile.uk
+++ b/lib/ukmmap/Makefile.uk
@@ -3,4 +3,4 @@ $(eval $(call addlib_s,libukmmap,$(CONFIG_LIBUKMMAP)))
 LIBUKMMAP_SRCS-y += $(LIBUKMMAP_BASE)/mmap.c
 
 UK_PROVIDED_SYSCALLS-$(CONFIG_LIBUKMMAP) += mmap-6 munmap-2 madvise-3
-UK_PROVIDED_SYSCALLS-$(CONFIG_LIBUKMMAP) += mremap-5
+UK_PROVIDED_SYSCALLS-$(CONFIG_LIBUKMMAP) += mremap-5 mprotect-3
--- a/lib/ukmmap/exportsyms.uk
+++ b/lib/ukmmap/exportsyms.uk
@@ -4,9 +4,12 @@ uk_syscall_r_mmap
 munmap
 uk_syscall_e_munmap
 uk_syscall_r_munmap
-mremap
 madvise
 uk_syscall_e_madvise
 uk_syscall_r_madvise
+mremap
 uk_syscall_e_mremap
 uk_syscall_r_mremap
+mprotect
+uk_syscall_e_mprotect
+uk_syscall_r_mprotect
--- a/lib/ukmmap/mmap.c
+++ b/lib/ukmmap/mmap.c
@@ -36,6 +36,7 @@
 #include <uk/alloc.h>
 #include <string.h>
 #include <uk/syscall.h>
+#include <uk/page.h>
 
 struct mmap_addr {
 	void *begin;
@@ -93,19 +94,23 @@ UK_SYSCALL_DEFINE(void*, mmap, void*, addr, size_t, len, int, prot,
 		last = tmp;
 		tmp = tmp->next;
 	}
-	void *mem = uk_malloc(uk_alloc_get_default(), len);
 
+	void *mem = uk_memalign(uk_alloc_get_default(), __PAGE_SIZE, len);
 	if (!mem) {
 		errno = ENOMEM;
 		return (void *) -1;
 	}
-	new = uk_malloc(uk_alloc_get_default(), sizeof(struct mmap_addr));
 
+	new = uk_malloc(uk_alloc_get_default(), sizeof(struct mmap_addr));
 	if (!new) {
 		uk_free(uk_alloc_get_default(), mem);
 		errno = ENOMEM;
 		return (void *) -1;
 	}
+
+	/* The caller expects the memory to be zeroed */
+	memset(mem, 0, len);
+
 	new->begin = mem;
 	new->end = mem + len;
 	new->next = NULL;
--- a/lib/ukmmap/mmap.c
+++ b/lib/ukmmap/mmap.c
@@ -121,52 +121,48 @@ UK_SYSCALL_DEFINE(void*, mmap, void*, addr, size_t, len, int, prot,
  * addr needs to be a pointer to a memory block that has been allocated from
  * mmap. If len has the same value with the size of the memory block that has
  * been allocated from mmap the struct mmap_addr counterpart is destroyed.
- * Otherwise the initial memory block is replaced by a smaller one.
+ * Otherwise, we do nothing and unfortunately have to leak the memory. This is
+ * inherent to the way this library works as it implements mmap on top of
+ * ukalloc which does not provide any means to partially free memory or do a
+ * realloc with the guarantee of not moving the existing allocation.
  */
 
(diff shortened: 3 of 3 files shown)


StefanJum (changes requested):
I think the memset that zeroes the memory should be done after we check if the `mem` variable is `NULL` [here](https://github.com/unikraft/unikraft/pull/394/files#diff-9cae6ee04f03a432adc2090da51d2d0dec431feda48f720db65e882db397d2bbR102). If there is not enough space to allocate the new memory, the application crashes since it passes `NULL` to `memset`.

marcrittinghaus:
> I think the memset that zeroes the memory should be done after we check if the `mem` variable is `NULL` [here](https://github.com/unikraft/unikraft/pull/394/files#diff-9cae6ee04f03a432adc2090da51d2d0dec431feda48f720db65e882db397d2bbR102). If there is not enough space to allocate the new memory, the application crashes since it passes `NULL` to `memset`.

Yes, of course 🤦🏻‍♂️ Thanks for catching that! I updated the commit.

StefanJum (approved):
All good now!

hlef:
Thanks for this PR @marcrittinghaus !

I had a look, it looks good overall. If necessary, we could merge it as-is.

That being said, I have a few nits to bring up:
- The comment before the implementation of `munmap` is now outdated. Outdated comments are so confusing. I'd be grateful if you could update it! :slightly_smiling_face: 
- My understanding after taking a quick look at the code is that the new implementation of `munmap` is going to leak memory forever if the caller frees chunks in bits (and not as a whole). This is fine for this PR as this is already a major improvement, but ultimately it'd be nice to address... And at any rate, we must have this clear somewhere, ideally with a NOTE or a FIXME comment in the code.

Do you think that you could address this for the release?

marcrittinghaus:
> Thanks for this PR @marcrittinghaus !
> 
> I had a look, it looks good overall. If necessary, we could merge it as-is.
> 
> That being said, I have a few nits to bring up:
> 
> * The comment before the implementation of `munmap` is now outdated. Outdated comments are so confusing. I'd be grateful if you could update it! 🙂
> * My understanding after taking a quick look at the code is that the new implementation of `munmap` is going to leak memory forever if the caller frees chunks in bits (and not as a whole). This is fine for this PR as this is already a major improvement, but ultimately it'd be nice to address... And at any rate, we must have this clear somewhere, ideally with a NOTE or a FIXME comment in the code.
> 
> Do you think that you could address this for the release?

Hi @hlef,

thanks for looking into this and pointing out the stale comment. I fixed it 👍🏻

Concerning your second point, this is actually something that cannot be fixed as `ukalloc` simply does not allow to emulate `munmap` with partial release. However, I will introduce `posix-mmap` at some point. This will use the paging code to properly implement the `mmap`-type syscalls. I am going to revisit `ukmmap` then and it will probably become obsolete or be replaced with a similar emulation in `posix-mmap` if paging is not available. I will then clearly point out limitations of `ukalloc`-based emulation.

hlef:
Hi Marc,

> Concerning your second point, this is actually something that cannot be fixed as `ukalloc` simply does not allow to emulate `munmap` with partial release. However, I will introduce `posix-mmap` at some point. This will use the paging code to properly implement the `mmap`-type syscalls. I am going to revisit `ukmmap` then and it will probably become obsolete or be replaced with a similar emulation in `posix-mmap` if paging is not available. I will then clearly point out limitations of `ukalloc`-based emulation.

Hum, don't you think that you could keep track of partially freed blocks in your ukalloc-based munmap implementation and free the whole block as soon as you got all the pieces? That'd have a bit of a cost for sure, but it could be optional. Of course, if the user never frees all sub-blocks ultimately, that wouldn't help, but in certain cases, it might. I don't think that it'd be very hard to implement. In another PR of course.

hlef:
Otherwise, thanks for updating!

Approved-by: Hugo Lefeuvre <hugo.lefeuvre@manchester.ac.uk>

marcrittinghaus:
> Hi Marc,
> 
> > Concerning your second point, this is actually something that cannot be fixed as `ukalloc` simply does not allow to emulate `munmap` with partial release. However, I will introduce `posix-mmap` at some point. This will use the paging code to properly implement the `mmap`-type syscalls. I am going to revisit `ukmmap` then and it will probably become obsolete or be replaced with a similar emulation in `posix-mmap` if paging is not available. I will then clearly point out limitations of `ukalloc`-based emulation.
> 
> Hum, don't you think that you could keep track of partially freed blocks in your ukalloc-based munmap implementation and free the whole block as soon as you got all the pieces? That'd have a bit of a cost for sure, but it could be optional. Of course, if the user never frees all sub-blocks ultimately, that wouldn't help, but in certain cases, it might. I don't think that it'd be very hard to implement. In another PR of course.

Yes of course, this would work, but it is not the interface `ukalloc` exposes at the moment. So we would have to implement like a feature flag in `ukalloc` and upgrade our own allocators. Because allocators like mimalloc don't do it and you don't want to adapt these 😏

marcrittinghaus:
> Hi Marc,
> 
> > Concerning your second point, this is actually something that cannot be fixed as `ukalloc` simply does not allow to emulate `munmap` with partial release. However, I will introduce `posix-mmap` at some point. This will use the paging code to properly implement the `mmap`-type syscalls. I am going to revisit `ukmmap` then and it will probably become obsolete or be replaced with a similar emulation in `posix-mmap` if paging is not available. I will then clearly point out limitations of `ukalloc`-based emulation.
> 
> Hum, don't you think that you could keep track of partially freed blocks in your ukalloc-based munmap implementation and free the whole block as soon as you got all the pieces? That'd have a bit of a cost for sure, but it could be optional. Of course, if the user never frees all sub-blocks ultimately, that wouldn't help, but in certain cases, it might. I don't think that it'd be very hard to implement. In another PR of course.

Yes, of course, this can be done and I guess should be part of a more proper emulation in `posix-mmap`. But as you said it won't actually free the memory until the whole region is unmapped.

nderjung:
Hi, when compiling NGINX (aka with `newlib`), I am now receiving:

```
/bin/../lib/gcc/x86_64-linux-gnu/9.2.0/../../../../x86_64-linux-gnu/bin/ld: /usr/src/unikraft/apps/nginx/build/libnewlibglue.o: in function `mprotect':
/usr/src/unikraft/libs/newlib/mem.c:72: multiple definition of `mprotect'; /usr/src/unikraft/apps/nginx/build/libukmmap.o:/usr/src/unikraft/unikraft/lib/ukmmap/mmap.c:194: first defined here
make[1]: *** [/usr/src/unikraft/unikraft/plat/kvm/Linker.uk:22: /usr/src/unikraft/apps/nginx/build/nginx_kvm-x86_64.dbg]
Error 1
make: *** [Makefile:1015: sub-make] Error 2
```

marcrittinghaus:
Ah. damn it. Yes, let me quickly do a PR for lib-newlib.

PR #395, opened 2022-02-10
Enhance arm(64)/x86_64 Bitscan Functions

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm`, `arm64`,  `x86_64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The patches correct wrong function descriptions in the arm versions of the bitscan functions. The patches also change the fls-type of bitscan functions to return an `unsigned int` instead of an `unsigned long` as promotion of the `int` result from the builtin functions is unnecessary. The patches then synchronize the x86_64 versions with these new descriptions and signatures to have the same API on all architectures.

### DIFF
--- a/arch/arm/arm/include/uk/asm/atomic.h
+++ b/arch/arm/arm/include/uk/asm/atomic.h
@@ -38,12 +38,11 @@
 #endif
 
 /**
- * ukarch_ffs - find first (lowest) set bit in word.
- * @word: The word to search
+ * Find first (lowest) set bit in word.
+ * @param x The word to operate on
  *
- * Returns one plus the index of the least significant 1-bit of x, or
- * if x is zero, returns zero.
- * ffs(1)=0, ffs(0x8000000)=31
+ * @return The index of the least significant 1-bit of x, or if x is zero,
+ *   the result is undefined. ffs(1)=0, ffs(3)=0, ffs(0x80000000)=31
  */
 static inline unsigned int ukarch_ffs(unsigned int x)
 {
@@ -51,11 +50,11 @@ static inline unsigned int ukarch_ffs(unsigned int x)
 }
 
 /**
- * ukarch_fls - find last (highest) set bit in word.
- * @word: The word to search
+ * Find last (highest) set bit in word.
+ * @param x The word to operate on
  *
- * Undefined if no bit exists, so code should check against 0 first.
- * fls(1)=0, fls(0x8000000)=31
+ * @return The index of the most significant 1-bit of x, or if x is zero,
+ *   the result is undefined. fls(1)=0, fls(3)=1, fls(0x80000001)=31
  */
 static inline unsigned int ukarch_fls(unsigned int x)
 {
@@ -63,10 +62,11 @@ static inline unsigned int ukarch_fls(unsigned int x)
 }
 
 /**
- * ukarch_ffsl - find first (lowest) set bit in word.
- * @word: The word to search
+ * Find first (lowest) set bit in long word.
+ * @param x The long word to operate on
  *
- * Undefined if no bit exists, so code should check against 0 first.
+ * @return The index of the least significant 1-bit of x, or if x is zero,
+ *    the result is undefined. ffs(1)=0, ffs(3)=0, ffs(0x80000000)=31
  */
 static inline unsigned long ukarch_ffsl(unsigned long x)
 {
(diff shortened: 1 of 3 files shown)


razvanvirtan (approved):
I've checked the code and tested the functions. Everything looks good to me!

michpappas:
Moin @marcrittinghaus, overall it looks good :) I have one nitpick and proposed enhancement:
* I think the intention of tagging the subject line of a commit message is to specify the base component rather than the file changed. So maybe it would be better to change `arm64/atomic.h: xyz` to `arch/arm64: xyz`. This also follows the convention used in previous commits.
* I noticed that `atomic.h` in arm and arm64 are identical. Moreover, arm and arm64 use builtins, while x86 implements bitscan functions using native instructions. If we change x86 to also use builtins, we can use a common header under `plat/common/include/`. What do you think?

marcrittinghaus:
> Moin @marcrittinghaus, overall it looks good :) I have one nitpick and proposed enhancement:

Moin @michpappas 😄 

> * I think the intention of tagging the subject line of a commit message is to specify the base component rather than the file changed. So maybe it would be better to change `arm64/atomic.h: xyz` to `arch/arm64: xyz`. This also follows the convention used in previous commits.

Yep, thank you for pointing it out. I will change that 👍🏻 

> * I noticed that `atomic.h` in arm and arm64 are identical. Moreover, arm and arm64 use builtins, while x86 implements bitscan functions using native instructions. If we change x86 to also use builtins, we can use a common header under `plat/common/include/`. What do you think?

I had the same thought and wondered why x86 does not use the builtins. I also think that it would be nice and would increase portability. However, I have compared the ASM output and found that the inline assembler creates more compact code on x86.
```C
int main(int argc, char *argv[] __unused)
{
	return __builtin_ffsl(argc) - 1; // vs. ukarch_ffsl(argc);
}
```
With `ukarch_ffsl(argc)`:
```ASM
movslq %edi,%rax
bsf    %rax,%rax
retq
```
With `__builtin_ffsl(argc) - 1`:
```ASM
movslq %edi,%rdi
mov    $0xffffffffffffffff,%rax
bsf    %rdi,%rdi
cmovne %rdi,%rax
retq
```
Only when you use a constant, the builtin allows the compiler to derive the result at compile-time and remove the actual bitscan instruction. But I think this will probably be less often the case. Unfortunately, removing the `-1` does not bring the builtin version to just emitting `bsf`. In fact, it makes it even worse by adding an `add    $0x1,%rax` instruction.

So the question is, if we are willing to forgo this optimization for less code and increased portability.

michpappas:
@marcrittinghaus, unless these are used in performance-critical code, I wouldn't care much about saving an instruction or two, as the compiler in general will create more or fewer instructions depending on the optimization level and perhaps depending on the context as these are inline. I don't have a strong opinion though ☺️   If we go the low-level way, it would be better at some point to also update the arm side to use the `clz` instruction and perhaps consider a way of unifying the two identical headers (maybe include one header in the other?).

marcrittinghaus:
If you do a loop like this, it becomes around 33% slower with the builtin versions:
```C
int x = 0;
__nsec start = ukplat_monotonic_clock();
for (int i = 0; i < 2000; i++)
    x += ukarch_ffs(argc + i);
__nsec stop = ukplat_monotonic_clock();
```

image

Although admittingly this is just a microbenchmark and in most practical code (e.g., in the paging code and the bitmap functions), the bitscans are only done once as part of a more complex operation. So this is probably not so much of a hit.

michpappas:
@marcrittinghaus Nice! 😎 So shall we merge this one as is, and at some point I take care of the Arm part?

marcrittinghaus:
> @marcrittinghaus Nice! 😎 So shall we merge this one as is, and at some point I take care of the Arm part?

Yes, for the release next week I would just leave it as is. Although, we can perhaps in the future think about if we are really willing to have the low-level implementations also on the other architectures or if in practice the improvement is not so big that it justifies the change. In that case, I am still open for unification to a single builtin-based header.

razvand:
I agree with @marcrittinghaus , let's get this in the 0.8 release. @michpappas , please so it can be merged.

michpappas:
> Yes, for the release next week I would just leave it as is. Although, we can perhaps in the future think about if we are really willing to have the low-level implementations also on the other architectures or if in practice the improvement is not so big that it justifies the change. In that case, I am still open for unification to a single builtin-based header.

Yup sounds good, thanks.

PR #405, opened 2022-02-14
lib/vfscore, lib/*fs: Use non-prototyped no-ops

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

For example:
- `CONFIG_LIBVFSCORE=y`
- `CONFIG_LIBRAMFS=y`

### Description of changes

VFS operations are implemented with function calls. In order to stub some of the operations a no-op function is assigned to those function pointers. As the function signature between the generic no-op function and the respective VOP does not match, the compiler displays a warning. We currently suppress this warning with a compile flag. However, this also suppresses legitimate warnings and may hide bugs. This commit changes the no-op functions to be non-prototypical so that any function signature is treated as valid. The compile flags can thus be removed.

### DIFF
--- a/lib/vfscore/vnode.c
+++ b/lib/vfscore/vnode.c
@@ -477,25 +477,25 @@ vnode_dump(void)
 #endif
 
 int
-vfscore_vop_nullop(void)
+vfscore_vop_nullop()
 {
 	return 0;
 }
 
 int
-vfscore_vop_einval(void)
+vfscore_vop_einval()
 {
 	return EINVAL;
 }
 
 int
-vfscore_vop_eperm(void)
+vfscore_vop_eperm()
 {
 	return EPERM;
 }
 
 int
-vfscore_vop_erofs(void)
+vfscore_vop_erofs()
 {
 	return EROFS;
 }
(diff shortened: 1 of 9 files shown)


lib/vfscore/vnode.c
```diff
@@ -528,4 +528,3 @@ void vn_del_name(struct vnode *vp __unused, struct dentry *dp)
 	/* UK_ASSERT(uk_mutex_is_locked(&vp->v_lock)); */
 	uk_list_del(&dp->d_names_link);
 }
-
```
jschlumpp, line 531: At least the [editorconfig](https://github.com/unikraft/unikraft/blob/979f60f93275e5e833f6ad7d135a75f88d406133/.editorconfig#L9) specifies that a final newline should be inserted at the end of file.

jschlumpp (changes requested):

lib/vfscore/vnode.c
marcrittinghaus: Hi @mschlumpp,

nice that you spotted this. However, it is fine because the file has two empty lines at the end, so VS removed one
image

jschlumpp: You are right. GitHub does not show the last line...

jschlumpp (approved):
Reviewed-by: Marco Schlumpp <marco.schlumpp@gmail.com>

skuenzer (approved):
Wow, this is a nice improvement. Thanks a lot for this.

PR #411, opened 2022-02-15
lib/vfscore: Fixing various bugs

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

- `CONFIG_LIBVFSCORE=y`

### Description of changes

This PR fixes a couple of bugs in `lib/vfscore` where:
+ `sys_fchdir` sets the wrong current working directory if the FS is not mounted at `/` (root)
+ the `read`/`write`/`stat` syscalls return values according to the C lib but not the raw syscall implementation

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -377,31 +377,6 @@ out_error:
 	return -error;
 }
 
-UK_TRACEPOINT(trace_vfs_read, "%d %p %d", int, void *, int);
-UK_TRACEPOINT(trace_vfs_read_ret, "0x%x", ssize_t);
-UK_TRACEPOINT(trace_vfs_read_err, "%d", int);
-
-UK_SYSCALL_R_DEFINE(ssize_t, read, int, fd, void *, buf, size_t, count)
-{
-	ssize_t bytes;
-
-	UK_ASSERT(buf);
-
-	struct iovec iov = {
-			.iov_base	= buf,
-			.iov_len	= count,
-	};
-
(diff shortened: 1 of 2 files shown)


craciunoiuc (approved):
Hey @marcrittinghaus @skuenzer,

All good here 🙂

marcrittinghaus:
I think there is a problem when the syscall shim is not activated. That is why it is not building. Have to revise

craciunoiuc:
Hmm, strange, I am 99% sure I looked at the checks and they were okay. Also the build worked for me. Maybe I'm going crazy 😅 

I'll also recheck now

Edit: Nevermind, you are right, I forgot to check without the shim

craciunoiuc (changes requested):
I removed the Reviewed-by tag, thanks for catching the error @marcrittinghaus 

Good thing we have a CI/CD pipeline

marcrittinghaus:
@craciunoiuc

I just had to move the definition of the `read` and `write` syscalls so that they are below the `readv` and `writev` syscalls. It works with the shim active because then you include a header with all the syscalls as forward declarations. So it should be good now.

marcrittinghaus:
Forgot to fix the outstanding checkpatch warning. Done.

craciunoiuc (approved):
All good! 🙏

skuenzer (approved):
Thanks a lot for these fixes.

PR #434, opened 2022-03-26
Introduce ukarch_random and add implementation for arm64 (FEAT_RNG)

michpappas:
### Prerequisite checklist

### Base target

 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

The ukarch_random API is unconditionally present. Each architecture's support is conditional and can be checked via the `ukararch_random_avail()`.

In arm64, availability is controlled by a newly introduced parameter `CONFIG_ARM_64_FEAT_RNG`.

#### The ukarch_random API

Processor-supported random number generation can be useful for different purposes on the platform and library level. Some examples include key / seed generation for hardware-based security protections, such as PAuth, MTE in Armv8, or seed generation for the ChaCha20-based PRNG in the ukswrand library.

This PR introduces a minimal API for processor-supported randomness, namely `ukarch_random`. The API consists of merely three functions, that map closely to the functionality provided by the relevant mechanisms in Armv8 and x86_64: `ukarch_random_avail()` checks whether the feature is implemented by the processor, `ukarch_random_get_long()` returns a random long integer, and `ukarch_random_long_reseed()` returns a random long integer after reseeding the RNG.

Given that the trustworthiness of processor-provided randomness is a controversial topic, higher layers shall ideally use `ukarch_random` optionally and when possible provide the user with additional options for other sources of randomness, such as drivers to TRNGs or Crypto Engines external to the processor.

#### Implementation in Armv8

Processor-supported randomness (FEAT_RNG) is introduced as an OPTIONAL extension in Armv8.5-a. The architecture defines the RNG as the output of a Deterministic Random Bit Generator (DRBG) that produces random numbers from a cryptographically secure algorithm, and is seeded by a True Random Number Generator (TRNG). The architecture requires that the DRBG and TRNG conform to a set of standards. Specifically the DRBG should conform to NIST SP800-90A Rev1, while TRNG should conform to NIST SP800-90B, NIST SP800-22, FIPS 140-2, BSI AIS-31.

The RNG is exposed to the user through the RNDR and RNDRSS registers. The RNDR outputs random numbers, and is re-seeded at an IMPLEMENTATION DEFINED rate. RNDRSS re-seeds the DRBG immediately before generating a new random number.

This PR provides an implementation of the `ukarch_random` API via the RNDR / RNDRSS registers, along with a new option `CONFIG_ARM_64_FEAT_RNG`, to enable the feature. As no Armv8 IP provided by Arm implements Armv8.5-A, this can be only tested in QEMU with `-machine virt -cpu max` or third-party SoCs that implement `FEAT_RNG`.

***Notice:*** Since the rate at which TRNGs can generate random bits is limited, RNDR and RNDRSS try to return a random number at a "reasonable period of time". When that fails, the PSTATE.NZCV bits are set to 0b0100, and a zero value is returned. Callers must therefore always check the result of the operation, and act as required on failure.

### DIFF
--- a/arch/Config.uk
+++ b/arch/Config.uk
@@ -33,3 +33,8 @@ config STACK_SIZE_PAGE_ORDER
 		Indirectly configures the stack size by changing the stack size page
 		order. Stack size is equal with 2^order * page size (e.g. 4KB).
 		Only change this if you know what you're doing.
+
+config HAVE_RANDOM
+	bool
+	help
+		Processor-generated randomness support.
--- a/arch/arm/arm64/Config.uk
+++ b/arch/arm/arm64/Config.uk
@@ -161,6 +161,15 @@ menu "Armv8-A Extensions"
 	help
 		BTI protects against JOP-like attacks by placing and
 		verifying landing pads on branch targets.
+
+config ARM64_FEAT_RNG
+	bool "Armv8.5 Random Number Generator"
+	select HAVE_RANDOM
+	help
+	  This enables the hardware RNG functionality provided by the
+	  processor. The architecture specifies that this is implemented
+	  by a Deterministic Random Bit Generator (DRBG) that is seeded
+	  from a TRNG.
 endmenu
 
 config ARM64_ERRATUM_858921
--- a/arch/arm/arm64/include/uk/asm/random.h
+++ b/arch/arm/arm64/include/uk/asm/random.h
@@ -1,0 +1,90 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2022, Michalis Pappas <mpappas@fastmail.fm>.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKARCH_RANDOM_H__
+#error Do not include this header directly
+#endif
+
+#ifdef CONFIG_ARM64_FEAT_RNG
+
+static inline int ukarch_random_init(void)
+{
+	__u64 reg;
+
+	__asm__ __volatile__("mrs %x0, ID_AA64ISAR0_EL1\n" : "=r"(reg));
+
+	return !((reg >> ID_AA64ISAR0_EL1_RNDR_SHIFT) &
+		 ID_AA64ISAR0_EL1_RNDR_MASK);
+}
+
+static inline int ukarch_random_u64(__u64 *val)
+{
+	__u64 res;
+
+	__asm__ __volatile__("	mrs	%x0, RNDR\n" /* Get rand */
+			     "	mrs	%x1, NZCV\n" /* Get result */
+			     : "=r"(*val), "=r"(res));
+	return res;
+}
+
+static inline int ukarch_random_u32(__u32 *val)
+{
+	__u32 res;
+	__u64 val64;
+
+	res = ukarch_random_u64(&val64);
+
+	*val = (__u32)val64;
+
+	return res;
+}
+
+static inline int ukarch_random_seed_u64(__u64 *val)
+{
+	__u64 res;
+
+	__asm__ __volatile__("	mrs	%x0, RNDRRS\n" /* Get rand */
+			     "	mrs	%x1, NZCV\n"   /* Get result */
+			     : "=r"(*val), "=r"(res));
+	return res;
+}
+
+static inline int ukarch_random_seed_u32(__u32 *val)
+{
+	__u32 res;
+	__u64 val64;
+
+	res = ukarch_random_seed_u64(&val64);
+
+	*val = (__u32)val64;
+
+	return res;
+}
+
+#endif /* CONFIG_ARM64_FEAT_RNG */
--- a/include/uk/arch/random.h
+++ b/include/uk/arch/random.h
@@ -1,0 +1,117 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2022, Michalis Pappas <mpappas@fastmail.fm>.
+ *                     All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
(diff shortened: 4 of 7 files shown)


samoilescusebastian (commented):
There are some conflicts that should be resolved. Otherwise, the PR looks fine to me.

michpappas:
@samoilescusebastian thanks for looking into this. I recently put a bit more thinking into it, and just pushed an update with the following changes:
* Rebased to staging.
* Added the `__check_result` macro (see the relevant commit). This is useful for catching bugs, but we should make sure it's not abused, [to avoid similar problem linux folks had before](https://lwn.net/Articles/204682/). I propose that we only use it in security-critical functions. CC: @marcrittinghaus, @skuenzer.
* Replaced `ukarch_random_avail()` with `ukarch_random_init()`. This allows more generic initialization, and the implementation can be reduced to a simple check of availability if no further initialization is needed.
* Renamed functions to use fixed sizes instead of generic types, ie `ukarch_random_get_long()` now becomes `ukarch_random_u64()`. I also added `ukarch_random_u32()`. Macros for generic types can be provided in the platform layer (not part of this PR - we need to put a bit more thinking into it).
* Added `CONFIG_UKARCH_RANDOM` and stubs when not available. This will also be useful later at the platform layer.

samoilescusebastian:
Green light :)
However looks like `concourse-ci/app-helloworld-staging/arm-linuxu/build`   test has failed, but I don't have access to the details.

samoilescusebastian (approved):

michpappas:
Rebased to `staging`

marcrittinghaus:
@michpappas

I had a quick look and it seems that the changes you mentioned (https://github.com/unikraft/unikraft/pull/434#issuecomment-1107448595) are not part of the current PR 👀

michpappas:
@marcrittinghaus who knows what I did back then, sorry 😅 Updated with correct version.

marcrittinghaus:
@michpappas Thanks for updating. Now it has a conflict due to the paging code merged 😆

michpappas:
Rebased to staging

arch/arm/arm64/Config.uk
```diff
@@ -161,6 +161,15 @@ menu "Armv8-A Extensions"
...
 		verifying landing pads on branch targets.
+
+	config ARM64_FEAT_RNG
+	bool "Armv8.5 Random Number Generator"
+	select UKARCH_RANDOM
+	help
+		This enables the hardware RNG functionality provided by the
+		processor. The architecture specifies that this is implemented
+		by a Deterministic Random Bit Generator (DRBG) that is seeded
+		from a TRNG.
```
marcrittinghaus: Same comments as to #458 apply regarding the indentation.

arch/arm/arm64/include/uk/asm/random.h
```diff
@@ -0,0 +1,91 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2022, Michalis Pappas <mpappas@fastmail.fm>.
+ *                     All rights reserved.
```
marcrittinghaus: Same optional note as for #458 regarding `All rights reserved.`. Your choice 😃

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKARCH_RANDOM_H__
+#define __UKARCH_RANDOM_H__
+
+#ifndef __ASSEMBLY__
+#include <uk/arch/types.h>
+#endif /* !__ASSEMBLY__ */
```
marcrittinghaus: Makes generally sense, however, the remainder of the header is not ASM compatible. Since there are no definitions that could be used in ASM anyways, it is fine to just remove the check.
```suggestion
#include <uk/arch/types.h>
```

arch/Config.uk
```diff
@@ -33,3 +33,8 @@ config STACK_SIZE_PAGE_ORDER
 		Indirectly configures the stack size by changing the stack size page
 		order. Stack size is equal with 2^order * page size (e.g. 4KB).
 		Only change this if you know what you're doing.
+
+config UKARCH_RANDOM
```
marcrittinghaus: To be more in line with the feature set config options in `lib/Config.uk` use `UKARCH_HAVE_RANDOM`.
```suggestion
config UKARCH_HAVE_RANDOM
```

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+
+#if CONFIG_UKARCH_RANDOM
+
+#include <uk/asm/random.h>
+
+/* Initialize the RNG
+ *
+ * @return 0 on success, negative value on failure
+ */
+extern int ukarch_random_init(void);
```
marcrittinghaus: Is the extern actually necessary?
```suggestion
int ukarch_random_init(void);
```

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+
+/* Get a 64-bit random integer
+ *
+ * @param [in|out] val A pointer to store the generated value
```
marcrittinghaus: This is only an out variable.
```suggestion
 * @param [out] val A pointer to store the generated value
```

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+ * @return 0 on success, negative value on failure
+ */
+extern __check_result int ukarch_random_u32(__u32 *val);
+
+/* Get a 32-bit random integer
+ *
+ * @param [in|out] val A pointer to store the generated value
```
marcrittinghaus:
```suggestion
 * @param [out] val A pointer to store the generated value
```

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+ * @return 0 on success, negative value on failure
+ */
+extern __check_result int ukarch_random_u64(__u64 *val);
+
+/* Reseed the RNG and get a 64-bit random integer
+ *
+ * @param [in|out] val A pointer to store the generated value
```
marcrittinghaus: Is this really supplying a new seed value via the parameter? In that case, please fix the parameter description to make that clear.

Don't know if this is just my VSCode plugin being uncooperative, but syntax highlighting is not working for me with `|`. Do you have the same issue? I thus always use `,`.

```suggestion
 * @param [in,out] val A pointer to store the generated value
```
image
image

If the reseed does not take a value as input, then just remove the `in`.

include/uk/arch/random.h
```diff
@@ -0,0 +1,113 @@
...
+extern __check_result int ukarch_random_seed_u64(__u64 *val);
+
+#else /* CONFIG_UKARCH_RANDOM */
+
+static inline int __check_result ukarch_random_init(void)
+{
+	return -1;
+}
+
+static inline int __check_result ukarch_random_u32(__u32 *val __unused)
```
marcrittinghaus, line 91: Since you are using `__check_result` and `__unused` you have to include `uk/essentials`.

include/uk/essentials.h
```diff
@@ -96,6 +96,9 @@ extern "C" {
 #ifndef __noinline
 #define __noinline             __attribute__((noinline))
 #endif
+#ifndef __check_result
+#define __check_result	__attribute__((warn_unused_result))
```
marcrittinghaus: See spacing comment from #458.

marcrittinghaus (commented):
Hi @michpappas,

thanks for submitting the PR! 😃 Looks pretty good. I left some minor feedback.

arch/arm/arm64/Config.uk
michpappas: Ok

arch/arm/arm64/include/uk/asm/random.h
michpappas: Will remove, thanks

include/uk/arch/random.h
michpappas: ACK

arch/Config.uk
michpappas: Will change, but to me the `HAVE` part seems redundant (I doubt anyone would thing that `CONFIG_RNG` disables the feature 😄). Added an item in the coding conventions document to discuss.

include/uk/arch/random.h
michpappas: Nope, probably left over while I was working on it. Thanks for spotting it out.

include/uk/arch/random.h
michpappas: ACK

include/uk/arch/random.h
michpappas: ACK

include/uk/arch/random.h
michpappas: No, the seed is not provided by the caller. Reseeding happens by the hardware, so it's only an `[out]` parameter. Fixed.

include/uk/arch/random.h
michpappas: Thanks

include/uk/essentials.h
michpappas: Ok

michpappas:
@marcrittinghaus thanks for the review, pushed a new revision with all comments addressed.

arch/Config.uk
marcrittinghaus: You are right. It is as clear as it gets also without the `HAVE`. Just suggested to be in line with the current convention. Thanks for changing and also for adding this to the document for discussion 😃

marcrittinghaus:
@michpappas Thanks for addressing the comments 😃 It looks good now from my perspective. I am going to test it again.

@samoilescusebastian Could you also do a second pass and then approve as reviewer if everything works for you?

craciunoiuc (approved):
All good here 🎲

marcrittinghaus (approved):
Yeet! 🥏

PR #441, opened 2022-04-04
lib/vfscore: Solve redefining conflicts

dragosargint:
There are many functions that are defined in musl as well as in vfscore. When using
musl we want to use the musl functions. When we use musl the `LIBSYSCALL_SHIM_NOWRAPPER`
macro is set, which leads to the setting of `UK_LIBC_SYSCALLS` to 0. We can use this
to exclude the redefined functions from vfscore.

This PR is part of a series:
https://github.com/unikraft/unikraft/pull/454
https://github.com/unikraft/unikraft/pull/453
https://github.com/unikraft/unikraft/pull/442
https://github.com/unikraft/unikraft/pull/441

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -178,7 +178,7 @@ int open(const char *pathname, int flags, ...)
 #endif
 
 LFS64(open);
-#endif
+#endif /* UK_LIBC_SYSCALLS */
 
 UK_LLSYSCALL_R_DEFINE(int, openat, int, dirfd, const char *, pathname,
 		      int, flags, int, mode)
@@ -234,7 +234,7 @@ int openat(int dirfd, const char *pathname, int flags, ...)
 #endif
 
 LFS64(openat);
-#endif
+#endif /* UK_LIBC_SYSCALLS */
 
 UK_SYSCALL_DEFINE(int, creat, const char*, pathname, mode_t, mode)
 {
@@ -289,7 +289,8 @@ UK_TRACEPOINT(trace_vfs_mknod, "\"%s\" 0%0o 0x%x", const char*, mode_t, dev_t);
 UK_TRACEPOINT(trace_vfs_mknod_ret, "");
 UK_TRACEPOINT(trace_vfs_mknod_err, "%d", int);
 
-int __xmknod(int ver, const char *pathname, mode_t mode, dev_t *dev __unused)
+static int __xmknod_helper(int ver, const char *pathname,
+		mode_t mode, dev_t *dev __unused)
 {
 	UK_ASSERT(ver == 0); // On x86-64 Linux, _MKNOD_VER_LINUX is 0.
 	struct task *t = main_task;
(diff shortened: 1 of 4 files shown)


lib/vfscore/main.c
```diff
@@ -2563,6 +2610,7 @@ int posix_fadvise(int fd __unused, off_t offset __unused, off_t len __unused,
 #endif
 
 LFS64(posix_fadvise);
+#endif // UK_LIBC_SYSCALLS
```
skuenzer: Can you use `/* */` arguments?

lib/vfscore/mount.c
```diff
@@ -308,11 +308,13 @@ UK_SYSCALL_R_DEFINE(int, umount2, const char*, path, int, flags)
 	return error;
 }
 
+#if UK_LIBC_SYSCALLS
 int
 umount(const char *path)
 {
 	return umount2(path, 0);
 }
+#endif // UK_LIBC_SYSCALLS
```
mogasergiu: I think you missed changing some of the comments to the /* */ style.

mogasergiu (commented):
Would it be possible to also squash the commit changing the style of the comments?

mogasergiu (approved):

lib/vfscore/mount.c
```diff
@@ -308,11 +308,13 @@ UK_SYSCALL_R_DEFINE(int, umount2, const char*, path, int, flags)
 	return error;
 }
 
+#if UK_LIBC_SYSCALLS
 int
 umount(const char *path)
```
skuenzer, line 313: Wouldn't make using `UK_SYSCALL_R_DEFINE` instead, more sense? I assume the same for `umount2`

lib/vfscore/pipe.c
```diff
@@ -604,10 +604,12 @@ UK_SYSCALL_R_DEFINE(int, pipe2, int*, pipefd, int, flags)
 	return 0;
 }
 
+#if UK_LIBC_SYSCALLS
 /* TODO maybe find a better place for this when it will be implemented */
 int mkfifo(const char *path __unused, mode_t mode __unused)
```
skuenzer, line 712: Wouldn't make using `UK_SYSCALL_R_DEFINE` instead, more sense?

skuenzer (changes requested):
Hi @dragosargint,

thanks a lot for these fixes. Can you check the few comments that I posted? It is basically concerning `mkfifo` and `umount`. We should use `UK_SYSCALL_R_DEFINE` for those instead. Then they are also properly registered at the syscall shim.

Thanks,

Simon

lib/vfscore/mount.c
dragosargint: `umount` is not actually a syscall. Musl will do:
```
int umount(const char *special)
{
        return syscall(SYS_umount2, special, 0);
}
```
And newlib will use the definition above.

lib/vfscore/pipe.c
dragosargint: `mkfifo` is not a syscall. Musl will do:
```
int mkfifo(const char *path, mode_t mode)
{
        return mknod(path, mode | S_IFIFO, 0);
}
```
And newlib will use the definition above.

lib/vfscore/mount.c
skuenzer: You are right, then this is fine.

lib/vfscore/pipe.c
skuenzer: Fair enough :+1:

skuenzer (approved):

PR #442, opened 2022-04-04
lib/posix-process: Solve redefining conflicts

dragosargint:
There are many functions that are defined in musl as well as in posix-process.
When using musl we want to use the musl functions. When we use musl the
`LIBSYSCALL_SHIM_NOWRAPPER` macro is set, which leads to setting of
`UK_LIBC_SYSCALLS` to 0. We can use this to exclude the redefined functions
from posix-process.

This PR is part of a series:
https://github.com/unikraft/unikraft/pull/454
https://github.com/unikraft/unikraft/pull/453
https://github.com/unikraft/unikraft/pull/442
https://github.com/unikraft/unikraft/pull/441


mogasergiu (approved):
Looks good to me.

lib/posix-process/process.c
```diff
@@ -281,19 +286,30 @@ int tcsetpgrp(int fd __unused, pid_t pgrp)
...
+	/* We don't support priority updates for unikernels */
+	return -EPERM;
+}
 
+#if UK_LIBC_SYSCALLS
 int nice(int inc __unused)
 {
 	/* We don't support priority updates for unikernels */
-	errno = EPERM;
+	errno = uk_syscall_r_nice(inc);
```
skuenzer: Hum, this call is making the errno value negative. Secondly, `UK_SYSCALL_R_DEFINE(int, nice, int, inc)` is already provide `nice` so this wrapper should be removed completely.

skuenzer (changes requested):
Hey Dragos,

can you try to declare all calls with `UK_SYSCALL_R_DEFINE` instead? Then we register them properly to the syscall shim and don't just remove the C symbol if a richt libc is is part of the build.

Thanks,

Simon

lib/posix-process/process.c
dragosargint: I agree on the `nice` syscall, but the other functions are not syscalls. For example `wait3` is defined in musl as follows:
```
pid_t wait3(int *status, int options, struct rusage *usage)
{
        return wait4(-1, status, options, usage);
}
```

skuenzer: Fair enough, this makes sense. In fact `wait`, `waitpid` and `wait3` are all mapped to `wait4`.

skuenzer (approved):

PR #443, opened 2022-04-06
plat/kvm/arm: Zero out the .bss on runtime setup

### LINKED ISSUE #360: kvm on aarch64 does not zero-out the bss
### Describe the bug

The boot-code in plat/kvm/arm/ does not zero-out the bss. This causes values to be retained after a system reset.

### Steps to reproduce

The issue can be reproduced by issuing PSCI_FN_SYSTEM_RESET. After reset Unikraft will exit with
`[    0.030335] CRIT: [libukboot] <boot.c @  259> Could not set the platform memory allocator`
The error is caused by `plat_allocator` retaining its pre-reset value.

### Expected behavior

kvm should zero-out the bss on boot.

arm64

kvm

### Relevant log output

```shell
[    0.019254] Info: [libukallocbbuddy] <bbuddy.c @  515> Initialize binary buddy allocator 4011f000
[    0.020292] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 4012a000 - 4012c000 (order 1)
[    0.021201] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 4012c000 - 40130000 (order 2)
[    0.021980] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40130000 - 40140000 (order 4)
[    0.022326] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40140000 - 40180000 (order 6)
[    0.022667] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40180000 - 40200000 (order 7)
[    0.023008] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40200000 - 40400000 (order 9)
[    0.023346] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40400000 - 40800000 (order 10)
[    0.023687] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 40800000 - 41000000 (order 11)
[    0.024030] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 41000000 - 42000000 (order 12)
[    0.024372] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 42000000 - 44000000 (order 13)
[    0.024720] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 44000000 - 48000000 (order 14)
[    0.025062] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 48000000 - 50000000 (order 15)
[    0.025405] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 50000000 - 60000000 (order 16)
[    0.025748] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 60000000 - 70000000 (order 16)
[    0.026088] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 70000000 - 78000000 (order 15)
[    0.026430] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 78000000 - 7c000000 (order 14)
[    0.026774] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7c000000 - 7e000000 (order 13)
[    0.027118] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7e000000 - 7f000000 (order 12)
[    0.027462] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7f000000 - 7f800000 (order 11)
[    0.027803] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7f800000 - 7fc00000 (order 10)
[    0.028143] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7fc00000 - 7fe00000 (order 9)
[    0.028484] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7fe00000 - 7ff00000 (order 8)
[    0.028824] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7ff00000 - 7ff80000 (order 7)
[    0.029162] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7ff80000 - 7ffc0000 (order 6)
[    0.029500] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7ffc0000 - 7ffe0000 (order 5)
[    0.029838] dbg:  [libukallocbbuddy] <bbuddy.c @  471> 4011f000: Add allocate unit 7ffe0000 - 7fff0000 (order 4)
[    0.030335] CRIT: [libukboot] <boot.c @  259> Could not set the platform memory allocator
```

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Additional configuration

None

### Description of changes

The bootcode in kvm/arm does not zero-out the bss. This causes the
retention of values after a system reset. Update entry64.S to zero-out
the bss when setting up the runtime.

Resolves #360

### DIFF
--- a/plat/kvm/arm/entry64.S
+++ b/plat/kvm/arm/entry64.S
@@ -121,6 +121,14 @@ ENTRY(_libkvmplat_entry)
 
 	mov sp, x27
 
+	/* Zero-out the BSS */
+	ldr x26, =__bss_start
+	ldr x27, =_end
+2:
+	str	xzr, [x26], #8
+	cmp	x26, x27
+	b.lo	2b
+
 	/* Set the context id */
 	msr contextidr_el1, xzr
 


razvanvirtan (approved):
Hi, Michalis! Things look good on my side, thanks!

skuenzer:
Hi @michpappas,
thanks a lot for looking into this and providing a fix. However, my concern with wiping the `.bss` is cost of boot time - obviously the size varies with libraries and application and can become at some point significant. Since `.bss` contains uninitialized data it should be fine to keep it as is. If for security reasons we want to wipe `.bss`, I would actually prefer to do it in generic code, like the first thing to do when entered `lib/ukboot`. This way, `.bss` wiping would be platform-agnostic. Unfortunately, we would get problems with `.bss` variables that are in use by early platform code and would then be resetted later during boot. Not sure how to solve it other than in the entrance assembly code.
Nevertheless, when reading issue #360, couldn't it be solved by moving the `platform_allocator` to `.data`, meaning giving it an actual initialization value? What do you think?
Thanks,

Simon

michpappas:
Hey @skuenzer! 😃 

yes that's very true, zeroing-out the `.bss` definitely impacts boot performance. But then again leaving it contain random data breaks very fundamental assumptions of the user, and makes the system non-deterministic on each boot. From elf(5):
```
.bss      This  section holds uninitialized data that contributes to the program's memory image. By definition,
          the system initializes the data with zeros when the program begins to run.  This section is of type
          SHT_NOBITS.  The attribute types are SHF_ALLOC and SHF_WRITE.
```

Similarly by the C standard (Section 6.7.8 of [C ISO Standard 9899:1999](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)):
```
If an object that has static storage duration is not initialized explicitly,
then:
— if it has pointer type, it is initialized to a null pointer;
— if it has arithmetic type, it is initialized to (positive or unsigned) zero;
— if it is an aggregate, every member is initialized (recursively) according to these rules;
— if it is a union, the first named member is initialized (recursi vely) according to these rules.
```

As you pointed out, if we move this logic to `ukboot` we break the platform code. Besides, does it make any difference if we do it at `ukboot`? To me it sounds like moving the problem from A to B, but eventually boot time would be affected either way. In fact, in `ukboot` it may be even slower if high-level functions such as `memset` are not optimized.

> Nevertheless, when reading issue https://github.com/unikraft/unikraft/issues/360, couldn't it be solved by moving the platform_allocator to .data, meaning giving it an actual initialization value?

But then we don't solve the problem at its root, later something else will turn up and every time this happens will require some debugging effort.

Other options I can think of:
1. Optimize: Admittedly my PR uses the naive approach of multiple stores, but we can probably do better. I had a quick look, and one common approach is using `DC ZVA` to zero-out the cache instead. I can look further into this, and other alternatives for optimization.
2. Have the `.bss` initialized in the image. One way to do that is to move the `.bss` inside `.data`. But of course that will impact image size and load time.
3. Implement zero-out on demand. Also possible, but as it requires more work I would rather if we started the simple way so that the system becomes deterministic, and open an issue so that we can look at on-demand method later.
4. A combination of 1 and 3.

I can also have a look on how they do it on linux etc.

marcrittinghaus:
@michpappas You wrote that `.bss` won't be initialized after a "reset". What situation are you referring to? Do you mean a call to `reset()` in `cpu_native.c`? 

Then as a compromise, we could do the zero-out there, because as you correctly state, any VMM offering a `-kernel` like argument or any ELF bootloader for that matter needs to zero-out this area for the first boot. So for this critical path it would be a waste of precious boot time.

skuenzer:
Hey @michpappas,

all true, yes. I think - as Marc said - our boot time is really precious. The fact that `.bss` is effectively zero initialized for ELF programs on Linux has also to do with avoiding data leakage from other programs. However, since GCC does not throw a warning it seems in fact that this became common understanding that non-initialized .bss variables are zero'ed out - otherwise we would have catched this error much earlier... :thinking: Now, I would love an GCC option that throws such for non-initialized .bss variables...
Is Marc's suggestion a valid and alternative option when we wipe .bss when performing a reset? Or do you think there are still cases that are not covered?

Simon

michpappas:
Hi @marcrittinghaus, @skuenzer the problem goes beyond that particular bug, or the case of reboot - these are just the symptoms. The real problem is that any code that expects that `.bss` is initialized to zero will fail. And that includes pretty much any code written for OSes other than Unikraft, like nginx and so on.

Anyway, if these are all clear I won't insist, so if you think that this should not be merged, please feel free to close 🙂

marcrittinghaus:
> Hi @marcrittinghaus, @skuenzer the problem goes beyond that particular bug, or the case of reboot - these are just the symptoms. The real problem is that any code that expects that `.bss` is initialized to zero will fail. And that includes pretty much any code written for OSes other than Unikraft, like nginx and so on.
> 
> Anyway, if these are all clear I won't insist, so if you think that this should not be merged, please feel free to close 🙂

Hi @michpappas,

I am not quite sure if I got this right. From my perspective, we have four scenarios:

1. The VMM provides the means to boot an ELF file (e.g., via multiboot or Linux boot protocol)
2. We use an existing boot loader like GRUB
3. We use a custom boot loader stage for Unikraft
(4. The binary runs on bare metal)

I think as all first three solutions are aware of `.bss` they MUST ensure zeroed memory for this section anyways for the initial boot. So these are not problematic. For the bare metal scenario I think you can assume to have a boot loader in between, which makes this case 2 or 3.

However, when you reset or reboot the machine you will probably have a problem in case 1. From my understanding this is what you observe? This would also happen, if we had a soft reset function in Unikraft.

So that is why I suggest to add `.bss` zero code to the reset/reboot path and not to the boot path. With Unikraft running into the reset/reboot path being much more unlikely, we save boot time for the general case.

michpappas:
> I am not quite sure if I got this right. From my perspective, we have four scenarios:
> 
>     1. The VMM provides the means to boot an ELF file (e.g., via multiboot or Linux boot protocol)
> 
>     2. We use an existing boot loader like GRUB
> 
>     3. We use a custom boot loader stage for Unikraft
>        (4. The binary runs on bare metal)
> 
> 
> I think as all first three solutions are aware of `.bss` they MUST ensure zeroed memory for this section anyways for the initial boot. So these are not problematic. For the bare metal scenario I think you can assume to have a boot loader in between, which makes this case 2 or 3.

If there is an ELF loader, I think it's reasonable to make the assumption that the runtime is set by the loader, or even make it a requirement.

For baremetal images, we cannot assume much about the entity that loads the image - it can be completely dumb. Moreover, it cannot zero out the `.bss` as it cannot know where it is. So the best-case scenario is to assume that it zeroes-out all memory (IMO we cannot rely on that).

Saying that, for KVM in particular one could easily argue that we always expect an ELF loader, and in fact `plat/kvm/Linker.uk` does not even output binary images, which is fair enough 😎 

> 
> However, when you reset or reboot the machine you will probably have a problem in case 1. From my understanding this is what you observe? This would also happen, if we had a soft reset function in Unikraft.
> 
> So that is why I suggest to add `.bss` zero code to the reset/reboot path and not to the boot path. With Unikraft running into the reset/reboot path being much more unlikely, we save boot time for the general case.

Yes, for KVM in particular, and assuming we agree that we expect an ELF loader to prepare the runtime, zeroing-out on reset only sounds the way to go. I will implement this and update this PR.

michpappas:
Depends on https://github.com/unikraft/unikraft/issues/571

razvand:
@michpappas , @skuenzer , @razvanvirtan , as agreed, this will make use of the recent shutdown feature #1148 and will be part of a rewrite by @michpappas , to free the `.bss` before reboot.

PR #453, opened 2022-04-21
lib/uksignal: Solve redefining conflicts

dragosargint:
There are many functions that are defined in musl as well as in uksignal.
When using musl we want to use the musl functions. When we use musl the
`LIBSYSCALL_SHIM_NOWRAPPER` macro is set, which leads to the setting of
`UK_LIBC_SYSCALLS` to 0. We can use this to exclude the redefined functions
from vfscore.

This PR is part of a series:
https://github.com/unikraft/unikraft/pull/454
https://github.com/unikraft/unikraft/pull/453
https://github.com/unikraft/unikraft/pull/442
https://github.com/unikraft/unikraft/pull/441

Signed-off-by: Dragos Iulian Argint <dragosargint21@gmail.com>

PR #454, opened 2022-04-21
lib/uktime: Solve redefining conflicts

dragosargint:
This PR is part of a series:

GitHub-Depends-On: #453
GitHub-Depends-On: #442
GitHub-Depends-On: #441


mogasergiu (approved):

lib/uktime/Makefile.uk
```diff
@@ -5,11 +5,13 @@ LIBUKTIME_COMMON_INCLUDES-y      += -I$(LIBUKTIME_BASE)/musl-imported/include
 CINCLUDES-$(CONFIG_LIBUKTIME)    += $(LIBUKTIME_COMMON_INCLUDES-y)
 CXXINCLUDES-$(CONFIG_LIBUKTIME)  += $(LIBUKTIME_COMMON_INCLUDES-y)
 
+ifneq ($(CONFIG_LIBMUSL), y)
```
skuenzer: Hum.. We should probably remove these files completely or move them to nolibc and if they are missing for nolibc builds. Can you check this and move or delete them from this library?

gabrielmocanu:
@mogasergiu  Can you review this again?

mogasergiu (approved):

lib/uktime/Makefile.uk
```diff
@@ -1,15 +1,9 @@
 $(eval $(call addlib_s,libuktime,$(CONFIG_LIBUKTIME)))
 
 LIBUKTIME_COMMON_INCLUDES-y      += -I$(LIBUKTIME_BASE)/include
-LIBUKTIME_COMMON_INCLUDES-y      += -I$(LIBUKTIME_BASE)/musl-imported/include
```
razvanvirtan, line 4: @dragosargint I think we have a problem with this one. Removing the include directory here will brake the compilation, since there will be undefined symbols (e.g. `CLOCK_MONOTONIC` from https://github.com/unikraft/unikraft/blob/staging/lib/uktime/musl-imported/include/time.h. I observed this behavior while trying to build `app-micropython` with newlib. I think this path should be replaced with something else, but not completely removed.

dragosargint: I think you mean it breaks the compilation for newlib. Unfortunately, the `unikraft:staging-muslpreview` branch, which is the base for this PR is made precisely to introduce things that could break the compilation for newlib.

lib/posix-futex/Config.uk
```diff
@@ -1,7 +1,7 @@
 menuconfig LIBPOSIX_FUTEX
 	bool "posix-futex: Futex"
 	select NOLIBC if !HAVE_LIBC
```
craciunoiuc: Shouldn't this be

`select NOLIBC if !HAVE_LIBC` -> `select LIBNOLIBC if !HAVE_LIBC`

I know it's not a change in this PR 😄

craciunoiuc (commented):
Hey

One small thing I noticed below

lib/posix-futex/Config.uk
dragosargint: You're right. I added that in the last commit

lib/posix-time/Config.uk
```diff
@@ -0,0 +1,4 @@
+config LIBPOSIX_TIME
+       bool "posix-time: Time syscalls"
+       default n
+       select HAVE_TIME
```
craciunoiuc, line 4: This should change now to `HAVE_POSIX_TIME`, right? + all the subsequent changes

This is something pretty minor so I wouldn't block on it

dragosargint: It is another symbol that is defined. Probably there are sources that depend on the definition of this name. I would not rename it because errors may occur in the replacement process.

craciunoiuc: Ok, will mark as resolved

lib/nolibc/Config.uk
```diff
@@ -2,7 +2,7 @@ menuconfig LIBNOLIBC
 	bool "nolibc: Only neccessary subset of libc functionality"
 	depends on !HAVE_LIBC
 	default y if !HAVE_LIBC
-	imply LIBUKTIME
```
skuenzer, line 5: I think you can completely remove this `imply`. If we can compile without `posix-time`, we should not enable it for the build.

lib/vfscore/Config.uk
```diff
@@ -2,7 +2,7 @@ config LIBVFSCORE
 	bool "vfscore: VFS Core Interface"
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
-	select LIBUKTIME if !HAVE_LIBC
```
skuenzer, line 5: Shouldn't `libposix-time` always be selected if vfscore really needs it (remove !HAVE_LIBC)? and do really all libc's provide its functionality?

skuenzer (changes requested):
Hi @dragosargint,

thanks a lot for restrcturing this code. I found one issue with this PR: When I compile the Unikraft base for x86_64 or Aarch64 without including `posix-time` the build fails at `__secs_to_tm.c`:
```
In file included from /home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/src/time/time_impl.h:1,
                 from /home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/src/time/__secs_to_tm.c:1:
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/include/time.h:45:1: error: unknown type name ‘clock_t’
   45 | clock_t clock (void);
      | ^~~~~~~
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/include/time.h:46:1: error: unknown type name ‘time_t’; did you mean ‘size_t’?
   46 | time_t time (time_t *);
      | ^~~~~~
      | size_t
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/include/time.h:46:14: error: unknown type name ‘time_t’; did you mean ‘size_t’?
   46 | time_t time (time_t *);
      |              ^~~~~~
      |              size_t
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/include/time.h:47:1: warning: parameter names (without types) in function declaration
   47 | double difftime (time_t, time_t);
      | ^~~~~~
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/musl-imported/include/time.h:48:1: error: unknown type name ‘time_t’; did you mean ‘size_t’?
   48 | time_t mktime (struct tm *);
      | ^~~~~~
      | size_t
/
[etc.]
```
I wonder if some basic time-related declaration should be still provided so that the default time functionality provided by nolibc compiles and is usable. `posix-time` should really only provide additional syscalls and should be compilable together with `musl` and `nolibc`. Any left (if any) pure libc function should be moved to `nolibc`.

Also, can you make sure to remove every `#include_next` directive from the headers. I saw that they depend on `nolibc` not being part of the build but since the header is now provided only with nolibc, we should remove the whole blocks:
```
include/time.h
23:#include_next <time.h>

include/sys/time.h
16:#include_next <sys/time.h>
```

I also found a left reference to `LIBUKTIME` in `plat/xen`. Since basic time handling is provided with `nolibc`, this should also be removed from `plat/xen/Config.uk`. If a system call is needed from `posix-time`, the platform should depend on that library.

Thanks,

Simon

lib/uktime/Makefile.uk
dragosargint: This is already removed. I don't know how you found this commit.

lib/vfscore/Config.uk
dragosargint: I removed them from anywhere

lib/nolibc/Config.uk
dragosargint: Removed

craciunoiuc (approved):
<you didn't see anything here>

lib/nolibc/musl-imported/include/time.h
```diff
@@ -42,6 +37,26 @@ struct tm {
 	const char *__tm_zone;
 };
 
+typedef long time_t;
```
skuenzer: Unfortunately, declaring the types directly here in this file break other headers, like `stat.h` that also depend on the declaration. The thing is that libC's `stat.h` should not include `time.h` by itself because this defines more than a user actually expects with inluding `stat.h` (would be bad practice). What you need to do is to move these declarations to nolibc's `shareddef.h` and guard each declaration like `useconds_t`:
```c
#if defined(__NEED_useconds_t) && !defined(__DEFINED_useconds_t)
typedef unsigned useconds_t;
#define __DEFINED_useconds_t
#endif
```
In the `time.h` header you include the type definitons with:
```c
#define __NEED_suseconds_t
#define __NEED_struct_timeval
#define __NEED_struct_timespec
#define __NEED_timer_t
#define __NEED_clockid_t
#define __NEED_clock_t

#include <nolibc-internal/shareddefs.h>
```

lib/nolibc/musl-imported/include/sys/time.h
```diff
@@ -6,16 +6,6 @@ extern "C" {
 
 #include <uk/config.h>
 
-#define __NEED_time_t
-#define __NEED_suseconds_t
-#define __NEED_struct_timeval
```
skuenzer, line 11: I think you should keep these three defines ahead and instead of including `uk/time_types.h`, include `nolibc-internal/shareddefs.h`. I have a longer explanation at the `time.h` header.

skuenzer (changes requested):
Hi @dragosargint,

unfortunately, `vfscore` and some of the other `posix-` libraries (e.g., `posix-process`) do not compile anymore in my case:
```
  CC      libvfscore: file.o
In file included from /home/skuenzer/workspace/unikraft/unikraft/lib/vfscore/include/vfscore/vnode.h:36,
                 from /home/skuenzer/workspace/unikraft/unikraft/lib/vfscore/include/vfscore/mount.h:41,
                 from /home/skuenzer/workspace/unikraft/unikraft/lib/vfscore/vfs.h:37,
                 from /home/skuenzer/workspace/unikraft/unikraft/lib/vfscore/file.c:34:
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/include/sys/stat.h:37:18: error: field ‘st_atim’ has incomplete type
   37 |  struct timespec st_atim;
      |                  ^~~~~~~
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/include/sys/stat.h:38:18: error: field ‘st_mtim’ has incomplete type
   38 |  struct timespec st_mtim;
      |                  ^~~~~~~
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/include/sys/stat.h:39:18: error: field ‘st_ctim’ has incomplete type
   39 |  struct timespec st_ctim;
      |                  ^~~~~~~
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/include/sys/stat.h:108:41: error: array type has incomplete element type ‘struct timespec’
  108 | int futimens(int, const struct timespec [2]);
      |                                         ^
/home/skuenzer/workspace/unikraft/unikraft/lib/nolibc/include/sys/stat.h:109:56: error: array type has incomplete element type ‘struct timespec’
  109 | int utimensat(int, const char *, const struct timespec [2], int);
      |                                                        ^
make[1]: *** [/home/skuenzer/workspace/unikraft/unikraft/support/build/Makefile.build:27: /home/skuenzer/workspace/unikraft/unikraft/build/libvfscore/file.o] Error 1
make: *** [Makefile:1016: sub-make] Error 2
```
This has to do with the time related data type defintions that we probably have to move to nolibc's `shareddef.h` header. Besides `stat.h`, `sys/time.h`, and `time.h` you should probably scan over the nolibc headers if all get the needed time declaration as needed.

I have some more comments at the `time.h` header.

Thanks,

Simon

lib/nolibc/musl-imported/include/time.h
dragosargint: Thank you for the comments. I adapted the code so that the compilation succeeds.

lib/nolibc/musl-imported/include/sys/time.h
dragosargint: You're right these must be kept.

skuenzer (approved):
Great! This version works well for me. Thanks a lot.

PR #191, opened 2021-04-25
Add Kernel Adress Sanitization support

vladandrew:
Kernel Address Sanitizer for Unikraft. Currently working only under KVM x86_64. We use the first part of the heap for the shadow memory(e.g. KASAN_MD_SHADOW_SIZE bytes are used for the shadow memory).  We instrument the memory allocator to mark memory as either valid or invalid. We use the shadow memory to keep track of our marks. 

Kasan instruments all memory accesses such that they are valid, if an memory region marked as invalid is accessed then we call `UK_CRASH~. `uk_malloc_ifpages` is instrumented such that the allocated memory is marked as valid and can be accessed, `kasan_mark_invalid` marks the memory as invalid and causes the kernel to crash with a helpful message. We add a global KASAN flag and disable KASAN for several internal libraries such as the allocator.

Example of KASAN crash output when we do an out of bounds access of an array:
```
[    0.105388] CRIT: [libkasan] <kasan.c @  179> ===========KernelAddressSanitizer===========
[    0.107294] CRIT: [libkasan] <kasan.c @  179> ERROR:
[    0.108431] CRIT: [libkasan] <kasan.c @  179> * invalid access to address 0x25b01b
[    0.110177] CRIT: [libkasan] <kasan.c @  179> * write of size 1
[    0.111537] CRIT: [libkasan] <kasan.c @  179> * redzone code 0x2 (partial redzone)
[    0.113275] CRIT: [libkasan] <kasan.c @  179> ============================================
```

PR #458, opened 2022-05-01
arch/arm64: Armv8 MTE - Memory Tagging Extensions

### LINKED ISSUE #575: Memory Tagging Extension
### Feature request summary

Add support for [MTE](https://source.android.com/docs/security/test/memory-safety/arm-mte) (*Memory Tagging Extension*) to tag memory on ARM64.

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [arm-64]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Memory tagging is a hardware protection against memory safety violations in the scope of both spatial (eg buffer overflows) and temporal (eg use-after-free) safety.
    
The idea behind MTE is to restrict pointer access to predefined blocks of memory. This is done by tagging memory regions and pointers, and check for match upon access. A mismatch generates a Tag Check Fault (TCF). Specifically, the architecture introduces the concepts of:
* Physical Address Tags, which are and are stored into the upper bits of pointers.    
* Allocation Tags, which tag blocks of memory. These are stored into the Tag PA Space, an IMPLEMENTATION    
  DEFINED memory space that is logically different from the Data PA Space.    
    
Combined together, these ensure that only memory within a given block is accessed. For instance, in the case of a buffer overflow, memory access outside the allocated bufer will cause a tag mismatch. Similarly, assuming that freeing memory re-tags the freed region, a use-after-free attempt will also cause a tag mismatch. Other accesses like prefetches, cache maintenance, translation table walks, accesses to the tag PA space, etc are always tag unchecked.
 
MTE requires support from the operating system. For example, addresses returned by malloc() must be tagged. Notice that the tags are only 4-bits wide, ie tags can have 16 possible values. Consequently, the same tag may be used on different regions on a given time, or that a region may have different tag values during time.

## Architecture support    
    
Armv8.5 introduces Memory Tagging as an OPTIONAL feature, through: 
* FEAT_MTE:  Basic tag management instructions, system instructions. Without MTE2 some of these are not usable (UNDEFINED, UNPREDICATBLE).
* FEAT_MTE2: Full MTE support.    
* FEAT_MTE3: Support for asymmetric TCF reporting.

Armv8.7 makes FEAT_MTE3 a mandatory when FEAT_MTE2 is implemented.
    
## Compiler support

**GCC-9**
    
Introduces initial MTE support with march=armv8.5-a+memtag. According to the GCC docs,     
"This option is only to enable the extension at the assembler level and does not affect code generation".
    
**GCC-10**
    
Introduces MTE heap tagging support. The 'memtag' option enables memory tagging intrinsics.    

At the time of writing, MTE stack tagging is not yet supported.

## Hardware Support

~~None. At the time of writing MTE is only known to be implemented in QEMU, and in Arm FVP.~~

Lately vendor support for MTE has started to appear on the news. Notable examples include the [DSU-110 DynamIQ cluster](https://fuse.wikichip.org/news/5270/arm-launches-the-dsu-110-for-new-armv9-cpu-clusters/) from Arm for Armv9, and Samsung's [Exynos 2200](https://news.samsung.com/global/samsung-introduces-game-changing-exynos-2200-processor-with-xclipse-gpu-powered-by-amd-rdna-2-architecture).

## Changes introduced in this series

This PR implements basic MTE support for arm64-based platforms, with the following changes:

* A new configuration option to enable MTE (CONFIG_ARM64_FEAT_MTE).
* A new memory type for tagged memory. This is set as the default type of Normal Memory when MTE is enabled.
* A new header under arch/arm/arm64/ with macros for generating, storing and loading tags. Currently only loads / stores of allocations to granule is suppported (`ldg` / `sdg`). Additional functionality such as store allocation tag with zeroing (`sdgz`), store allocation tag for blocks (`sdgm`), etc is not implemented.
* Platform helpers for MTE initialization, and tagging of contiguous regions.
* A new configuration option to enable memory tagging, and additional options to control tag check fault reporting (`CONFIG_ARM64_FEAT_MTE_TCF_[SYNC,ASYNC,ASYMMETRIC]`).
* A check in the IRQ vector to check for accumulated errors when tag check fault reporting is set to Async or Asymmetric.

An additional commit demonstrates how an allocator can be modified to support MTE, by introducing instrumentation to `ukalloc`. This is a fairly simple implementation that tags memory on `uk_malloc_ifpages()`, `uk_free_ifpages()`, and `uk_posix_memalign_ifpages()`. The region tagged is equal to the size requested by the user aligned to the MTE granule, ie 16 bytes. Any attempt to access memory outside that region, including allocation metadata, will cause a fault.

## Resources

1. [Memory Tagging Extension Whitepaper](https://developer.arm.com/-/media/Arm%20Developer%20Community/PDF/Arm_Memory_Tagging_Extension_Whitepaper.pdf)
2. [Learn the architecture: Providing protection for complex software](https://developer.arm.com/documentation/102433/latest/)
3. [The Arm Architecture Reference Manual Armv8, for Armv8-A architecture profile](https://developer.arm.com/documentation/ddi0487/ga)


michpappas:
@marcrittinghaus, @skuenzer I would like your comments on this one, especially in the topics below.

**The changes in the allocator:** These somewhat relate to the discussion in #191 on whether we should add a wrapper around alloc for these things. My understanding is that this would help with any allocators provided by Unikraft, but not with allocators in external implementations of libc (newlib / musl). The other option is to instrument the allocator's implementation. This has the drawback that some work needs to be duplicated for every allocator, but also benefits in performance, as you can optimize `calloc()`, `malloc()` for large blocks etc using instructions like "store allocation tag with zeroing" (`sdgz`), "store allocation tag for blocks" (`sdgm`), etc. CC: @vladandrew 

**The implementation of asynchronous TCF reporting:** My implementation comes with a penalty of one register check per IRQ. IIRC linux checks the TFSR on task switch, thread switch, and when swicthing between EL0 and EL1. We don't have most of these choices on Unikraft, so IRQ is one of our few choices. If people find this an issue, and would rather check in other places (say in the scheduler), we can make checks in IRQ optional and provide alternatives via KConfig.

michpappas:
This PR depends on #434. As RNG support is a dependency for MTE, I am adding the relevant commits here too. Once #434 is merged I will rebase this PR to `staging`.

plat/common/arm/mte.c
```diff
@@ -0,0 +1,118 @@
...
+		       (MTE_TCF_SYNC << SCTLR_EL1_TCF_SHIFT)));
+#endif
+	/* Enable TBI */
+	SYSREG_WRITE(TCR_EL1, (SYSREG_READ(TCR_EL1) | TCR_TBI0_BIT));
+
+	/* TCR_EL1.TCMA0 must be zero */
+	SYSREG_WRITE64(TCR_EL1, (SYSREG_READ64(TCR_EL1) & ~TCR_TCMA0_BIT));
+
+	/* Use default random tag generation */
+	SYSREG_WRITE64(GCR_EL1, (SYSREG_READ64(GCR_EL1) & GCR_EL1_RRND_BIT));
```
craciunoiuc: any reason why there is a combination of `SYSREG_WRITE64`/`SYSREG_WRITE` and `SYSREG_READ64`/`SYSREG_READ`?

lib/ukalloc/alloc.c
```diff
@@ -41,6 +41,10 @@
 #include <uk/arch/limits.h>
 #include <uk/arch/lcpu.h>
 
+#if CONFIG_ARM64_FEAT_MTE
```
craciunoiuc, line 44: Please change `#if` to `#ifdef` to make things consistent

lib/ukalloc/alloc.c
```diff
@@ -65,9 +69,12 @@ int uk_alloc_register(struct uk_alloc *a)
 }
 
 struct metadata_ifpages {
+#ifdef CONFIG_ARM64_FEAT_MTE
+	__sz		size;
+#endif /* CONFIG_ARM64_FEAT_MTE */
 	unsigned long	num_pages;
 	void		*base;
-};
+} __attribute__ ((aligned(16)));
```
craciunoiuc: I think there is a `__aligned` somewhere in unikraft that does this, you can use that

craciunoiuc (commented):
Hey @michpappas,

I did an initial pass over this PR and it looks pretty ok, I left some comments.

One thing that bothers me a bit is the the overall spam of `ifdef/endif` guards.
Unfortunately I'm not sure if there is a better way to this, as the protected code is pretty small and non-repetitive so I'm not sure if it can be abstracted.

Other than this, the PR looks okay! I'll try it out after the RNG PR is in.

I'll let @skuenzer now give his input on this.

Cezar

plat/common/arm/mte.c
michpappas: No reason, changed all to `SYSREG_READ`/`SYSREG_WRITE`, thanks for spotting this out 👍🏼

lib/ukalloc/alloc.c
michpappas: Fixed, thanks 👍🏼

michpappas:
> One thing that bothers me a bit is the the overall spam of ifdef/endif guards.
Unfortunately I'm not sure if there is a better way to this, as the protected code is pretty small and non-repetitive so I'm not sure if it can be abstracted.

A better way would be adding hooks to ukalloc, as I'm sure they could be useful for other purposes too (other variants of ASAN at minimum). Perhaps @skuenzer has thoughts on this?

michpappas:
Resolved reviewer's feedback, rebased to staging, and updated the "Supported Hardware" section in the cover letter.

gabrielmocanu:
@craciunoiuc  Can you take a look again? Thanks!

lib/ukalloc/alloc.c
```diff
@@ -76,7 +83,11 @@ struct metadata_ifpages {
  * compile time assertion will abort the compilation and this value will have
  * to be increased.
  */
-#define METADATA_IFPAGES_SIZE_POW2 16
+#ifdef CONFIG_ARM64_FEAT_MTE
+#define METADATA_IFPAGES_SIZE_POW2 32
+#else
+#define METADATA_IFPAGES_SIZE_POW2 24
+#endif /* CONFIG_ARM64_FEAT_MTE */
```
craciunoiuc: This was 16 previously, why is it 24 now?

plat/common/arm/mte.c
```diff
@@ -0,0 +1,117 @@
...
+ */
+#include <arm/arm64/cpu.h>
+#include <stddef.h>
+#include <uk/plat/mte.h>
+#include <uk/arch/lcpu.h>
+#include <uk/arch/random.h>
+#include <uk/arch/traps.h>
+#include <uk/event.h>
+
+void *mte_tag_contig(void *ptr, size_t size)
```
craciunoiuc: Stupid question: what does `contig` come from?

craciunoiuc (commented):
Hey @michpappas,

I left two questions for you.

In the meantime I will test the feature a bit more

plat/common/arm/mte.c
michpappas: `contig` stands for "contiguous"

lib/ukalloc/alloc.c
michpappas: Looks like by accident, as it's definitely incorrect (24 is not a power of 2). Reverted.

michpappas:
@craciunoiuc besides the comment above, I updated the align attribute of `struct metadata_ifpages` to use `MTE_TAG_GRANULE` for clarity, and restructured the definition of this struct a bit to reduce `#ifdef` noise, thanks.

plat/common/arm/mte.c
craciunoiuc: ok, just wanted to make sure

craciunoiuc:
I have no more comments on this. It looks pretty good.

It compiles fine on hello-world with different ARM features turned on/off.
I couldn't run it properly, as no cpu I used had this feature. Maybe in newer Qemu versions there's ARMv8.5 CPUs

I'll wait for the signal from @marcrittinghaus

michpappas:
@craciunoiuc if you'd like to do runtime testing you can pass `-machine virt,mte=on` to QEMU. I use qemu-system-aarch64 6.2.0.

craciunoiuc:
Ok, good to know, will test today

craciunoiuc:
Tested all the features + combinations

It works on my side :smile:

arch/arm/arm64/include/uk/asm/mte.h
```diff
@@ -0,0 +1,95 @@
...
+	__asm__ __volatile__ ("msr	tco, %0\n"		\
+			      : : "r"(PSTATE_TCO_BIT));		\
+})
+
+#define PSTATE_TCO_CLEAR() ({					\
+	__asm__ __volatile__ ("msr	tco, %0\n"		\
+			      : : "r"(0));			\
+})
+
+/* Tags address with a random tag.
```
marcrittinghaus:
```suggestion
/**
 * Tags address with a random tag.
```

arch/arm/arm64/include/uk/asm/mte.h
```diff
@@ -0,0 +1,95 @@
...
+ * @param  addr input address
+ * @return tagged address
+ */
+#define mte_insert_random_tag(addr) ({				\
+	__u64 reg;						\
+	__asm__ __volatile__("gmi	%x0, %x1, xzr\n"	\
+			     "irg	%x1, %x1, %x0\n"	\
+			     : "=r"(reg) : "r" (addr));		\
+	addr;							\
+})
```
marcrittinghaus: Is there a specific reason to not implement these macros as `static inline` functions? It would add type safety and simplify debugging

arch/arm/arm64/include/uk/asm/mte.h
```diff
@@ -0,0 +1,95 @@
...
+
+/* Reads tag from allocation memory
+ *
+ * @param  allocation address
```
marcrittinghaus:
```suggestion
 * @param addr allocation address
```

arch/arm/arm64/include/uk/asm/mte.h
```diff
@@ -0,0 +1,95 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2021, Michalis Pappas <mpappas@fastmail.fm>.
+ * All rights reserved.
```
marcrittinghaus:
```suggestion
 * Copyright (c) 2022, Michalis Pappas <mpappas@fastmail.fm>.
```
I recently read an article about the `All rights reserved` clause and it basically said it does not make sense as you say with this that all rights belong to you only to say some lines below with the BSD license that you give up rights.

arch/arm/arm64/Config.uk
```diff
@@ -161,6 +161,52 @@ menu "Armv8-A Extensions"
...
+	endif
+
+	config ARM64_FEAT_RNG
+	bool "Armv8.5 Random Number Generator"
+	select UKARCH_RANDOM
+	help
+		This enables the hardware RNG functionality provided by the
+		processor. The architecture specifies that this is implemented
+		by a Deterministic Random Bit Generator (DRBG) that is seeded
+		from a TRNG.
```
marcrittinghaus: Indentation seems somewhat mixed in this file as there are options that do not indent their fields, whereas others do and then it is also mixed for the same option (e.g., `choice` has indent at the top, but not in the new code). Not critical, but it adds to the already existing style clutter in the kconfig files.

Btw: I have been working on the code conventions during the gathering in Sinaia (feel free to edit/give feedback):
https://docs.google.com/document/d/1_iRkdBjQtnqOj798ZNr23OQTWQcXPdpDsnfBikm6Pcg

plat/common/arm/mte.c
```diff
@@ -0,0 +1,117 @@
...
+{
+	uint64_t addr = (uint64_t)ptr;
+
+	UK_ASSERT(!(size % MTE_TAG_GRANULE));
+
+	addr = mte_insert_random_tag(addr);
+
+	for (unsigned int i = 0; i < size; i += MTE_TAG_GRANULE)
+		mte_store_alloc(addr + i);
```
marcrittinghaus: Using an `unsigned int` is probably not a good idea here. Also, the definition should be outside the loop, and you can use `__vaddr_t` as type for a virtual address.

```suggestion
	__vaddr_t addr = (__vaddr_t)ptr;
	__off off;
	
	UK_ASSERT(!(size % MTE_TAG_GRANULE));

	addr = mte_insert_random_tag(addr);

	for (off = 0; off < size; off += MTE_TAG_GRANULE)
		mte_store_alloc(addr + off);
```
Also add the `#include <uk/arch/types.h>` for `__vaddr_t` and `__off`

arch/arm/arm64/include/uk/asm/arch.h
```diff
@@ -212,6 +254,10 @@
 #define ID_AA64MMFR2_EL1_CNP_SHIFT		_AC(0, U)
 #define ID_AA64MMFR2_EL1_CNP_MASK		_AC(0xf, ULL)
 
+/* RGSR_EL1: Random Allocation Tag Seed Register */
+#define RGSR_EL1_SEED_SHIFT	_AC(8, U)
+#define RGSR_EL1_SEED_MASK	_AC(0xffff, UL)
```
marcrittinghaus: Maybe adjust value alignment with the `SCTLR_` register?

include/uk/plat/mte.h
```diff
@@ -0,0 +1,42 @@
...
+#ifndef __UK_MTE_H__
+#define __UK_MTE_H__
+
+#include <stddef.h>
+#include <uk/asm/mte.h>
+#include <uk/essentials.h>
+
+#define MTE_ALIGN_SIZE(sz) ALIGN_UP(sz, MTE_TAG_GRANULE)
+
+void *mte_tag_contig(void *ptr, size_t size);
```
marcrittinghaus: If you use `__sz` you could avoid including `stddef.h` in a UK global header file.

Since @craciunoiuc asked. I also stumbled over `contig` and first thought it was a typo and should be `config`. Question is: Is there a case where you would tag a non-contiguous buffer?

include/uk/plat/mte.h
```diff
@@ -0,0 +1,42 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UK_MTE_H__
```
marcrittinghaus: The include guard name does not match the position in the source tree.

plat/common/include/arm/mte.h
```diff
@@ -0,0 +1,44 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKPLAT_MTE_H__
```
marcrittinghaus: Include guard name does not fit the naming scheme

plat/kvm/arm/setup.c
```diff
@@ -256,8 +261,9 @@ int _init_paging(void)
 				     _libkvmplat_cfg.heap.start;
 
 	frames = _libkvmplat_cfg.heap.len >> PAGE_SHIFT;
+	attr = PAGE_ATTR_PROT_RW | PAGE_ATTR_TYPE_NORMAL_WB_TAGGED;
```
marcrittinghaus, line 264: Just a question: Is this working to specify `PAGE_ATTR_TYPE_NORMAL_WB_TAGGED` irrespective of support for MTE?

plat/common/arm/traps_arm64.c
```diff
@@ -198,5 +203,12 @@ void trap_el1_irq(void)
 {
 	UK_ASSERT(gic);
 
+#ifdef CONFIG_ARM64_FEAT_MTE
+	if (unlikely(mte_async_fault())) {
+		uk_pr_crit("EL1 async tag check fault\n");
+		ukplat_crash();
+	}
```
marcrittinghaus:
```suggestion
	if (unlikely(mte_async_fault()))
		UK_CRASH("EL1 async tag check fault\n");
```

plat/common/arm/mte.c
```diff
@@ -0,0 +1,117 @@
...
+{
+	__u64 reg;
+	__u64 seed;
+	unsigned int mte_version;
+
+	mte_version = (SYSREG_READ(ID_AA64PFR1_EL1) >>
+			ID_AA64PFR1_EL1_MTE_SHIFT) &
+			ID_AA64PFR1_EL1_MTE_MASK;
+
+	if (mte_version < ARM64_FEAT_MTE2) {
```
marcrittinghaus: You could add `unlikely` for the error cases (although admittingly it is not that performance critical).

include/uk/plat/mte.h
```diff
@@ -0,0 +1,42 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
marcrittinghaus: I am not 100% happy with this header. Memory tagging is more of an architecture thing than a platform thing. So, it should be moved to `uk/arch`. 

But then I also think selling MTE as a broadly available feature on different architectures by placing the header globally in `uk/arch` is not really optimal either. Especially, if at the same time, we add checks for `ARM_FEAT_MTE` everywhere (e.g., `ukalloc`). 

So, if we place this in the global `uk/arch` directory and give it a dedicated header, we should rename identifiers to have the `ukarch_` prefix and then provide empty definitions for all other architectures or if the feature is not enabled. This would also eliminate the numerous checks in everywhere else.

Alternative would be to remove this header and just include the `asm/mte.h`, for example, in the `arch/lcpu.h` header and keep the `ARM_FEAT_MTE` checks.

What was your original motivation for the current layout (i.e., /uk/plat/ + no prefix + ARM checks when used)?

lib/ukalloc/alloc.c
```diff
@@ -64,10 +68,18 @@ int uk_alloc_register(struct uk_alloc *a)
...
+	__sz		size;
+	unsigned long	num_pages;
+	void		*base;
+} __align(MTE_TAG_GRANULE);
+#else
 struct metadata_ifpages {
 	unsigned long	num_pages;
 	void		*base;
 };
+#endif
```
marcrittinghaus:
```suggestion
#endif /* CONFIG_ARM64_FEAT_MTE */
```

plat/common/arm/mte.c
```diff
@@ -0,0 +1,117 @@
...
+	SYSREG_WRITE(TCR_EL1, (SYSREG_READ(TCR_EL1) | TCR_EL1_TBI0_BIT));
+
+	/* TCR_EL1.TCMA0 must be zero */
+	SYSREG_WRITE(TCR_EL1, (SYSREG_READ(TCR_EL1) & ~TCR_EL1_TCMA0_BIT));
+
+	/* Use default random tag generation */
+	SYSREG_WRITE(GCR_EL1, (SYSREG_READ(GCR_EL1) & GCR_EL1_RRND_BIT));
+
+	/* Set seed for RGSR_EL1 */
+	if (ukarch_random_init() == -1)
```
marcrittinghaus: This is not safe according to the function description which might return any negative number to indicate an error.
```suggestion
	if (ukarch_random_init())
```

plat/common/arm/mte.c
```diff
@@ -0,0 +1,117 @@
...
+	__u64 seed;
+	unsigned int mte_version;
+
+	mte_version = (SYSREG_READ(ID_AA64PFR1_EL1) >>
+			ID_AA64PFR1_EL1_MTE_SHIFT) &
+			ID_AA64PFR1_EL1_MTE_MASK;
+
+	if (mte_version < ARM64_FEAT_MTE2) {
+		uk_pr_err("FEAT_MTE2 is not implemented\n");
+		return -1;
```
marcrittinghaus: Should we return an errno number instead? This would make more sense, because during initialization the error code is printed when `mte_init` fails.

plat/common/arm/mte.c
marcrittinghaus: Would it make sense, or it is possible to move the `ukarch_random_init` and `ukarch_random_u64` before the `SYSREG_WRITE` commands? You could then do a graceful exit just like with the feature unsupported errors.

include/uk/essentials.h
```diff
@@ -96,6 +96,9 @@ extern "C" {
 #ifndef __noinline
 #define __noinline             __attribute__((noinline))
 #endif
+#ifndef __check_result
+#define __check_result	__attribute__((warn_unused_result))
```
marcrittinghaus: Please align the `__attribute__` and in this case use spaces to make it consistent with the other definitions (blame @skuenzer 😄).

lib/ukalloc/alloc.c
```diff
@@ -145,23 +164,42 @@ void *uk_malloc_ifpages(struct uk_alloc *a, __sz size)
 	metadata = (struct metadata_ifpages *) intptr;
 	metadata->num_pages = num_pages;
 	metadata->base = (void *) intptr;
+#ifdef CONFIG_ARM64_FEAT_MTE
+	metadata->size = size;
 
+	return mte_tag_contig((void *)(intptr + sizeof(*metadata)), size);
+#endif /* CONFIG_ARM64_FEAT_MTE */
 	return (void *)(intptr + sizeof(*metadata));
```
marcrittinghaus: Doesn't this complain about unreachable code? 🤔

lib/ukalloc/alloc.c
```diff
@@ -64,10 +68,18 @@ int uk_alloc_register(struct uk_alloc *a)
 	return 0;
 }
 
+#ifdef CONFIG_ARM64_FEAT_MTE
+struct metadata_ifpages {
+	__sz		size;
+	unsigned long	num_pages;
```
marcrittinghaus: I am wondering if you could replace `num_pages` with `size` and just do a `DIV_ROUND_UP(size, __PAGE_SIZE)` to get `num_pages` again. Then the struct would keep its size. I guess you do not want to use `num_pages * __PAGE_SIZE` to compute `size` to save tagging unused memory in the page?

marcrittinghaus (commented):
Hi @michpappas,

I have done a first pass and have a couple suggestions and also questions. I have not included the commits of `ukarch_random` in the feedback to not mix up things.

A general question: I see that you modified the `ifpages` interface to tag memory regions. So, the MTE feature is only working with the buddy allocator, right? Other allocators would need to be modified directly I assume.

EDIT: Just seeing in the discussion that you already mentioned the problem of a missing global allocation hook. I personally would say that IFF performance critical optimizations like using the other tagging instructions is not feasible in a generic interface then we should go for the per-allocator modification. It is not that much code to add and from my experience, even large allocators have only few allocation paths where this would need to be added.

arch/arm/arm64/include/uk/asm/mte.h
michpappas: Good point, I had a look at https://opensource.org/licenses/BSD-3-Clause and it's not specified there either. I also did a lookup and found some explanation in a [github issue opened in choosealicense.com](https://github.com/github/choosealicense.com/issues/556). Additional argumentation [here](https://opensource.stackexchange.com/questions/9427/with-bsd-3-do-i-need-to-propagate-an-all-rights-reserved-statement).

Other open source projects [use it nevertheless](https://fedoraproject.org/wiki/Licensing:BSD#New_BSD_(no_advertising,_3_clause)). I also did a grep in Unikraft's sources and it's pretty much everywhere 😬 

So is it safe to assume that from now on new PRs in Unikraft should not use this statement, and existing sources can left as is, given that it's a void statement anyway?

PS: I don't mind removing (or keeping) that clause, but I would rather if this PR would't block on this topic if it turns out that it will take time to come up with a final decision 🙂

arch/arm/arm64/include/uk/asm/mte.h
michpappas: Thanks will fix

arch/arm/arm64/include/uk/asm/mte.h
michpappas: No reason, I'll switch to inlines 👍🏼

arch/arm/arm64/include/uk/asm/mte.h
michpappas: ACK

arch/arm/arm64/Config.uk
michpappas: > Btw: I have been working on the code conventions during the gathering in Sinaia (feel free to edit/give feedback)

Nice! I'll go through it.

I will also update `ARM64_FEAT_MTE` in this PR according to the guidelines, and open an additional PR to fix the rest.

plat/common/arm/mte.c
michpappas: > Using an unsigned int is probably not a good idea here. 

Okay, but then why `off_t` and not `size_t`, which also the type of `size` that the loop compares against?

> Also, the definition should be outside the loop

Actually I think it's a pretty good practice to do that, as it reduces clutter in the variable declarations at the top of the function, so I would like if we considered to permit (or even mandate) this practice in the guidelines. Do you see any downsides?

> and you can use __vaddr_t as type for a virtual address.

The rationale for using `uint64_t` is that this feature is not expected to be portable as it is only available in AArch64. I make this assumption in general for certain features in `arm64`, while for the parts that I want to be backwards compatible,  I tend to use generic types (eg in the implementation of the SMCCC that also applies to Armv7, and both AArch64 and AArch32 in Armv8). Unless we want to be 128-bit ready, that is 😄

arch/arm/arm64/include/uk/asm/arch.h
michpappas: ACK

include/uk/plat/mte.h
michpappas: > If you use __sz you could avoid including stddef.h in a UK global header file.

Ok

> Since @craciunoiuc asked. I also stumbled over contig and first thought it was a typo and should be config. Question is: Is there a case where you would tag a non-contiguous buffer?

Funny, to me it's completely obvious, but maybe it's because it's frequently used in academia in the UK. Anyway, since it's confusing for others I'll rename to `mte_tag_region()`. The intention was to be forwards compatible, in case for example we export a function to tag a single address.

include/uk/plat/mte.h
michpappas: Will fix thanks

plat/common/include/arm/mte.h
michpappas: Will fix. Based on the number of occurrences, I assume the right convention is `__PLAT_CMN_<name>_H_`. It's pretty messed up in there, but we can fix the rest in the scope of the plat-rearch effort.

plat/kvm/arm/setup.c
michpappas: Yes, if MTE is not enabled this is ignored, so I thought I could save us from the extra `#ifdef`.

plat/common/arm/traps_arm64.c
michpappas: Of course 👍🏼

plat/common/arm/mte.c
michpappas: It's a good practice to use it anyway, will fix 👍🏼

lib/ukalloc/alloc.c
michpappas: ACK

plat/common/arm/mte.c
michpappas: Yes, looking at it now I'm not even sure why I did it that way 🤔

plat/common/arm/mte.c
craciunoiuc: Hmm, is there an errno for missing hardware feature?

include/uk/essentials.h
craciunoiuc: I totally missed this 😆

lib/ukalloc/alloc.c
craciunoiuc: nope

include/uk/plat/mte.h
craciunoiuc: `Question is: Is there a case where you would tag a non-contiguous buffer?`

Exact same thing I asked myself too. `contig` suggests there is a case where there will also be a `non_contig`, but maybe I'm just too pedantic

plat/common/arm/mte.c
michpappas: Good point, thanks 👍🏼

plat/common/arm/mte.c
michpappas: In general I'm also in favor of errno, but given that MTE is a security feature perhaps we don't want to handle an initialization error gracefully? Of course we can return values from errno and let the platform choose how to handle it. I will change to `-ENOTSUP`.

Btw, maybe I'm missing something but I see some inconsistency here. On the one hand we don't want `stdint` types in global headers (in my understanding to avoid libc dependency) and on the other hand we use `errno` values sparingly. I know we discussed that before (IIRC in the scope of the SMP API) and the outcome was something along the lines of _"by using errno numbers we don't mandate libc, the platform is free to define these values independently of a standard library"_. Do I remember correctly? Please correct me if I'm wrong. But if we go down that road, what stops someone from arguing that we can then also assume the same for `stdint` given that types are more fundamental than error numbers? 😬 Where is the line? 🤔 I'll add this topic to in my notes for discussion on plat-rearch.

include/uk/essentials.h
michpappas: "Bad @skuenzer! Bad!"
Thanks, blamed @skuenzer , and updated whitespace.

lib/ukalloc/alloc.c
michpappas: I am testing this with `gcc-arm-10.3-2021.07-x86_64-aarch64-none-elf`, and I don't see any warning. But I'll add an `#else` anyway.

lib/ukalloc/alloc.c
michpappas: Yes exactly, the idea was to tag only the memory requested by the user, up to `MTE_TAG_GRANULE_SIZE`. Both for performance and saving tag memory. I also imagine that storing the intended size could be useful for other purposes like stats etc. Do you think we should tag to page granularity instead?

include/uk/plat/mte.h
michpappas: Well, since it should be initialized by the platform I envisioned support is hierarchical, as arch -> plat. Exporting this via `include/uk/plat` would allow applications to tag memory even without libc. But I did that for convenience, rather than having a use-case in mind, so I'm happy to revert and restrict to internal use only. Saying that, in the Unikraft gathering we discussed a bit the possibility to deprecate the requirement of building for multiple platforms on a single build and allow applications to access internal platform / arch functionality (still inconclusive). The prospect of this being supported would also be valid argumentation towards making `include/uk/plat/mte.h` redundant. Last but not least, applications can always manually include internal headers in their makefiles.

Also, now that we will soon have POSIX `mmap()`, we can instead add support for the `PROT_MTE` there. I can do that after #603 is merged.

> But then I also think selling MTE as a broadly available feature on different architectures by placing the header globally in uk/arch is not really optimal either. Especially, if at the same time, we add checks for ARM_FEAT_MTE everywhere (e.g., ukalloc).

Another possibility is to give this feature a more generic name like `ukarch_memtag`, given that most likely such a feature will eventually also land in x86_64 (perhaps it already exists there too?), and for now, as you suggest, stub it out there. Then eliminate the various `#ifdef`s. What do you think?

michpappas:
@marcrittinghaus thanks for the review, I updated the PR addressing most of your comments except from a few that require additional feedback.

arch/arm/arm64/include/uk/asm/mte.h
marcrittinghaus: It was just a hint that this might not be necessary. I used the phrase before, too. Definitely not a blocker. You can keep it or remove it, as you like.

plat/common/arm/mte.c
marcrittinghaus: > Okay, but then why off_t and not size_t, which also the type of size that the loop compares against?

Right, `off_t` would not make sense. But `__off` is defined in `uk/arch/types.h` along with `__u64`, `__sz` etc. and is one of those Unikraft native types.

> Actually I think it's a pretty good practice to do that, as it reduces clutter in the variable declarations at the top of the function, so I would like if we considered to permit (or even mandate) this practice in the guidelines. Do you see any downsides?

I see your point and I also had this feeling a few times that the loop variable felt like clutter in the definition block. However, I am a bit reluctant to add the exception for loop variables to the coding convention as it makes the convention more complex and opens the door for further exceptions. For example, you could argue why not having a general exception for variables used only in a nested/closed scope (which is a valid point, btw.). Making it mandatory is too late, from my perspective, as basically no code in UK would follow this. Nevertheless, we can discuss this topic as part of the coding conventions. For now, I think, since all other code has the definition outside, I would go with that.

> The rationale for using uint64_t is that this feature is not expected to be portable as it is only available in AArch64. I make this assumption in general for certain features in arm64...

Ok, makes sense. Then maybe use `__64` if you want to keep it consistent with `__sz`.

lib/ukalloc/alloc.c
marcrittinghaus: No, makes absolutely sense to restrict tagging to the actual allocation size (from a performance and security perspective). My intention was to make the number of places with ARCH-specific `#ifdefs` as small as possible to keep the code clean and simple. When replacing the `num_pages` field with the `size` field, we could use this with and without MTE. We only need to replace `num_pages` in the code with something like `ALIGN_UP(size, __PAGE_SIZE)`.

plat/common/arm/mte.c
marcrittinghaus: > In general I'm also in favor of errno, but given that MTE is a security feature perhaps we don't want to handle an initialization error gracefully?

Good point. I mean we do crash in the platform anyways, so technically it does not make a difference. 

My argument was that while the caller may not know the details, he generally has a better picture of the context. So, providing the caller enough information so that he can make an informed decision is usually a good idea, I think. And if a security feature is critical to the operational context, the caller will know and terminate anyways.

> Btw, maybe I'm missing something but I see some inconsistency here. On the one hand we don't want stdint types in global headers (in my understanding to avoid libc dependency)...

Yes, you are absolutely right. From my perspective, I am strongly in favor of adding a `uk/errno.h` with Linux compatible values (at least a sensible subset). The header would define the values if not provided by a C library. We could then include `uk/errno.h` everywhere, also in low-level platform code and have a clear cut.

plat/common/arm/mte.c
marcrittinghaus: BTW: I added a section about variable definitions and a comment with your feedback related to loop variables to the code conventions doc so we can discuss.

include/uk/plat/mte.h
marcrittinghaus: > Funny, to me it's completely obvious, but maybe it's because it's frequently used in academia in the UK. Anyway, since it's confusing for others I'll rename to mte_tag_region(). The intention was to be forwards compatible, in case for example we export a function to tag a single address.

I like that solution and we can easily add a single address version if needed.

include/uk/plat/mte.h
marcrittinghaus: It does not really fit into `include/uk/plat` because the header includes an `asm` path, which only architectural headers should do. And it makes sense to have the actual functions defined in `asm` because this is an architectural feature and not a platform feature.

However, all functionality defined in `include/uk/[arch|plat]` should use the `uk_`, `ukarch_` or `ukplat_` prefixes. So, yes. Strictly speaking, this needs to get a name like `ukarch_memtag` with stubs for x86. It is probably the cleaner solution even in the case there won't be an x86 memory tagging feature.

plat/common/arm/mte.c
michpappas: Agreed, thanks 👍🏼 I updated `mte_init()` to return `-ENOTSUP` in the last iteration already, so I think we can resolve this one now. I put the topic of `errno.h` in my list for plat-rearch discussion.

plat/common/arm/mte.c
michpappas: ACK, that shold be `__u64` indeed, will change.

> Right, off_t would not make sense. But __off is defined in uk/arch/types.h along with __u64, __sz etc. and is one of those Unikraft native types.

Sorry I think I made things confusing. I meant `__off` is the native equivalent of `off_t`, but that type doesn't make sense conceptually in this case, as AFAIK it is normally used for files. So the `__sz` type currently used seems fine to me for this case.

include/uk/plat/mte.h
michpappas: Ok, that was updated in the last iteration, we can resolve this one now 👍🏼

plat/common/arm/mte.c
marcrittinghaus: ok, that is fine, too 😃 👍🏻

lib/ukalloc/alloc.c
michpappas: @marcrittinghaus the problem I see with that is that we lose information required for `free()`. On the one hand, for the requirements of tagging we need to preserve the user size padded to MTE tag granule (to re-tag on `free()`). On the other hand, for the requirements of the allocator, we need to preserve the allocated pages. However during free this cannot always be obtained by aligning up `size + sizeof(struct metadata_ifpages)` to page size, because of the additional alignment potentially imposed if the user previously called `memalign_ifpages()`.

One idea to reduce the `#ifdef` noise a little bit is to unconditionally add the element `size` to `struct metadata_ifpages`, and also create a definition for the alignment, eg:
```
#ifdef CONFIG_ARM64_FEAT_MTE    
#define __align_metadata_ifpages __align(MTE_TAG_GRANULE);    
#else    
#define __align_metadata_ifpages    
#endif /* CONFIG_ARM64_FEAT_MTE */    

...
         
struct metadata_ifpages {    
        __sz            size;      /* user size */     
        unsigned long   num_pages; /* alloc pages */    
        void            *base;    
} __align_metadat_ifpages;
```
This will eliminate some noise like:
```
#ifdef CONFIG_ARM64_FEAT_MTE
        __sz size;
#endif /* CONFIG_ARM64_FEAT_MTE */
...
#ifdef CONFIG_ARM64_FEAT_MTE
        metadata->size = size;
#endif
```
but of course others will remain. What do you think?

marcrittinghaus: Ah, you are right. Yes, sounds good. Thanks for looking into this 😃

include/uk/plat/mte.h
michpappas: @marcrittinghaus agreed. Notice that introducing a generic `ukarch_memtag` API requires making some assumptions on how memory tagging will be implemented in other architectures. I am deriving the requirements of the API from what we have needed so far for ukalloc:
1. A way to tag regions
2. A way to align region size to tag granularity
3. A way to strip the tag from a pointer

Point 1 is quite generic and I think it is safe to assume that a function such as `void *ukarch_memtag_region(void *ptr, __sz size);` will do. Notice however that even that function returns a tagged pointer, which makes an assumption about the implementation, that is tags are stored in pointers. This assumption is carried over to the 3rd point. For point 2 we also have an assumption about granularity requirements, but it's easy to overcome if some architecture does not impose any alignment requirements. Perhaps all these are a bit stretched, and in practice it's likely that all implementations will look the same, more or less. But still, before we introduce this interface, perhaps it's worth to be aware of the risks.

So based on the above, I see the following options:
1. Introduce a generic `ukarch_memtag` with the above features, based on the assumption that implementations in other architectures will look similar.
2. Move the public interface of MTE from `include/uk/plat/` to `include/uk/arch/` and reconsider generalizing if other architectures introduce a similar feature.

What do you think?

Edit: It's worth to keep in mind that this PR only introduces basic support in ukalloc, so the API requirements derived from that are also simple.

marcrittinghaus: @michpappas These are valid points and perhaps it is really better to wait until we see other architectures implementing this to come up with a safe generic interface. However, I am not super happy because it would be the first header there that introduces functions without the `ukarch_` prefix and which are only available on a certain architecture. It also means that the changes in the allocator(s) are ARM specific and need `#ifdef` guards everywhere. Also, for compatibility, would we need to keep this ARM-specific interface even if we introduce a generic interface?

So, maybe in the end it does not make much of a difference if we would (change)/extend an existing API vs. introducing one later and (removing)/keeping the old ARM-specific one around.

Could you imagine a different way to tag regions besides encoding it in the pointer?

From my perspective, we should move this to `include/uk/arch` anyways.

michpappas: @marcrittinghaus 
> However, I am not super happy because it would be the first header there that introduces functions without the ukarch_ prefix and which are only available on a certain architecture.

Ah yes, I forgot about that.

> Could you imagine a different way to tag regions besides encoding it in the pointer?

I cannot, that's why I said all this are probably a bit stretched 🙂 

Okay, then I will update the PR and introduce the generic `ukarch_memtag` interface, thanks for the feedback 👍🏼

michpappas:
Hi @marcrittinghaus I pushed another update with a rework to introduce a generic ukarch_memtag and move all MTE implementation there. Please notice that:
1. Now we arch/arm64/include/uk/asm/memtag.h includes `errno.h` for `ukarch_memtag_ini()` to return `-ENOTSUP` as you requested on your previous comment.
2. I left handling of asynchronous reporting out of `ukarch_memtag`, and only conditional to MTE.
3. In theory ukalloc could be changed to have all memory tagging operations unconditional, and have `ukarch_memtag` stub them out if memory tagging is not supported, or not enabled. Not sure if that's what you had in mind. Saying that, all this is subject to change anyway, as we need a common memory tagging library that will use KASAN if a hardware implementation is not available (in linux MTE is also part of the KASAN subsystem as "hardware KASAN"). Also see that the changes in `alloc.c` for KASAN are very close to MTE. All this was dicussed during the security workshop in Sinaia, yet I would rather get this PR merged first and work with @vladandrew towards the `libmemtag` as KASAN gets on the way to be merged on 0.12, rather going over endless iterations of reworks here. CC: @hlef

craciunoiuc:
I'll go ahead and test it again, just to make sure

Edit: Still works like a charm 🥇

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -35,6 +35,10 @@
 #include <uk/asm/arch.h>
 #include <uk/config.h>
 
+#if CONFIG_UK_ARM64_FEAT_MTE
+#include <uk/asm/mte.h>
+#endif
+
```
marcrittinghaus:
```suggestion
```

The config option does not exist and it seems that there is no need to include `mte.h` here, right?

marcrittinghaus (commented):
@michpappas Thanks for putting the extra effort into this!
 
> 1. Now we arch/arm64/include/uk/asm/memtag.h includes `errno.h` for `ukarch_memtag_ini()` to return `-ENOTSUP` as you requested on your previous comment.

Ok, no problem. We discuss the introduction of an `uk/errno.h` also for the code conventions and can adapt later if needed.

> 2. I left handling of asynchronous reporting out of `ukarch_memtag`, and only conditional to MTE.

Yes, makes sense 👍🏻

> 3. In theory ukalloc could be changed to have all memory tagging operations unconditional, and have `ukarch_memtag` stub them out if memory tagging is not supported, or not enabled. Not sure if that's what you had in mind. Saying that, all this is subject to change anyway, as we need a common memory tagging library that will use KASAN if a hardware implementation is not available (in linux MTE is also part of the KASAN subsystem as "hardware KASAN"). Also see that the changes in `alloc.c` for KASAN are very close to MTE. All this was dicussed during the security workshop in Sinaia, yet I would rather get this PR merged first and work with @vladandrew towards the `libmemtag` as KASAN gets on the way to be merged on 0.12, rather going over endless iterations of reworks here. CC: @hlef

Yes, when I wrote the comment, I had in mind to avoid `#ifdefs` where possible to reduce code clutter. I wasn't aware of the planned integration with KASAN to be honest as I am not reviewing that PR and not really wrapped my head around it that much 🙈, but it totally makes sense 👍🏻 Nevertheless, I think it is cleaner now. Just like with the random API you can now use a common `ukarch` API in KASAN to use HW features if available. So at least it will make KASAN cleaner in the end 😃 Thanks for making the changes 🙏🏻

@craciunoiuc Thanks for checking it out!

arch/arm/arm64/include/uk/asm/lcpu.h
michpappas: Yes, you're right I missed that thanks. Pushed an update to fix.

michpappas:
> Yes, when I wrote the comment, I had in mind to avoid `#ifdefs` where possible to reduce code clutter. I wasn't aware of the planned integration with KASAN to be honest as I am not reviewing that PR and not really wrapped my head around it that much see_no_evil, but it totally makes sense 👍🏻 Nevertheless, I think it is cleaner now. Just like with the random API you can now use a common `ukarch` API in KASAN to use HW features if available. So at least it will make KASAN cleaner in the end smiley Thanks for making the changes 🙏🏻

Yes exactly, the moving to ukarch is not wasted, it should had been done anyway, the only part subject to changes is `alloc.c`, and the addition of a higher level library that would probably deploy either software or hardware memory tagging. Btw, I somehow took it also for granted that we were on the same page, then I realized that this was mostly discussed in Sinaia, sorry 😅

craciunoiuc (approved):
I'll go ahead and add the review tag, as we are nearing the release.

marcrittinghaus (approved):
@michpappas: Just saw that you have to remove the commits from #434 (I think they are not the newest, too). We can then merge #434 first and then merge this one.

michpappas:
Rebased from `staging`, and removed RNG patches. @marcrittinghaus if you're happy we can merge.

marcrittinghaus (approved):
Thanks @michpappas!

PR #459, opened 2022-05-07
lib/ukstore: Introduce the `ukstore` library (static)

craciunoiuc:
These commits introduce a new internal library to Unikraft, `ukstore`.
The role of this new library is to provide a standard way of passing
information getters/setters from one place to another.

The function pointers are stored in entries and entries are stored in
folders. These are structured in two different ways, statically and
dynamically. The static ones are used for information that does not
change at runtime.

To generate sections per library the AWK tool is used together with
the build system. This library has a dependency only to nolibc.

This PR does not depend on any other `ukstore` PR.
The old PR was marked as Draft.


craciunoiuc:
@skuenzer we will continue our discussion here. This can/should make it this release

I have 2 things I will comment upon because I'm not sure about them.
Also, please do a full pass over it, because I had to reorder the commits and I'm not 100% I removed everything dynamic-related (even though I checked it and it works)

also tagging: @marcrittinghaus @nderjung @mandrei12

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		char *to_set = malloc(sizeof(*val) * strlen(val));
+
+		if (unlikely(!to_set))
+			return -ENOMEM;
+
+		strcpy(to_set, val);
+
+		ret = e->set.charp(e->cookie, to_set);
+
+		free(to_set);
```
craciunoiuc: This makes sense, but doesn't feel right, we need to discuss how to manage memory inside it.
Most probably we need to pass to it also a buffer, or inform people that their buffer will be saved as a reference, not reallocated.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+
+	UK_ASSERT(e);
+
+	switch (e->type) {
+	case uk_store_get_entry_type(u8): {
+		__u8 val;
+
+		str = calloc(UK_STORE_NR_SIZE8_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(s8): {
+		__s8 val;
+
+		str = calloc(UK_STORE_NR_SIZE8_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(u16): {
+		__u16 val;
+
+		str = calloc(UK_STORE_NR_SIZE16_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(s16): {
+		__s16 val;
+
+		str = calloc(UK_STORE_NR_SIZE16_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(u32): {
+		__u32 val;
+
+		str = calloc(UK_STORE_NR_SIZE32_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(s32): {
+		__s32 val;
+
+		str = calloc(UK_STORE_NR_SIZE32_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(u64): {
+		__u64 val;
+
+		str = calloc(UK_STORE_NR_SIZE64_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(s64): {
+		__s64 val;
+
+		str = calloc(UK_STORE_NR_SIZE64_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

lib/ukstore/store.c
```diff
@@ -0,0 +1,1214 @@
...
+		*out  = str;
+		return ret;
+	}
+
+	case uk_store_get_entry_type(uptr): {
+		__uptr val;
+
+		str = calloc(UK_STORE_NR_SIZE64_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
```
craciunoiuc: Also here.

craciunoiuc (commented):
Hey @skuenzer,

we need to discuss the allocations done here.
Basically, who does the allocation, the caller or the callee.

This is the case also for the `dynamic` part, but I'll do the changes also there when it's the case.

mandrei12 (commented):
I tested the static part and after checking the code I did not find anything wrong with it.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+ * Returns the id of the place it was called from.
+ *
+ * @return this libraries' id
+ */
+#define uk_store_libid_self()	\
+	uk_store_libid(__LIBNAME__)
+
+/* All basic types that exist - use these when interacting with the API */
+enum uk_store_entry_type {
+	_LIB_UK_STORE___undef = 0,
```
skuenzer: You don't need to set a number for each enum if they are anyways linearly increasing. The C enumeration declaration will do this for you automatically. Just define the value for `undef`.

lib/ukstore/exportsyms.uk
```diff
@@ -0,0 +1,23 @@
...
+_uk_store_get_s8
+_uk_store_get_u16
+_uk_store_get_s16
+_uk_store_get_u32
+_uk_store_get_s32
+_uk_store_get_u64
+_uk_store_get_s64
+_uk_store_get_uptr
+_uk_store_get_static_entry
+_uk_store_free_entry
```
skuenzer: Put only the symbols here that you are actually introducing with this commit and the ones we need for static entries support.

lib/ukstore/Makefile.uk
```diff
@@ -29,6 +33,22 @@ $(LIBUKSTORE_BUILD)/uk_libs.in.new:
 		     $(UK_LIBS) $(UK_LIBS-y) \
 		| tr ' ' '\n' > $@)
 
+$(LIBUKSTORE_GEN_INCLUDES_PATH)/uk/_store_libs.h: $(LIBUKSTORE_BASE)/gen_libs.h.awk $(LIBUKSTORE_BUILD)/uk_libs.in
+	$(call build_cmd,AWK,libukstore,$(notdir $@), \
```
skuenzer, line 36: Hey, I think the generation of id lists of the selected libraries should go into an own library: `libs/uklibid`. As it currently looks like, we will need something like this for the changes we work on `lib/ukdebug`. This is for something later: Could you provide a PR after we integrated this one that makes this split? While you do this, could you also adopt the rules to use the built-in `AWK` preprocessing rule that is available in the meantime. For example:
```Makefile
LIBUKSTORE_SRCS-y += $(LIBUKSTORE_BASE)/_store_libs.awk>.h
LIBUKSTORE_GEN_LIBS_AWKFLAGS    += -F '-'
LIBUKSTORE_GEN_LIBS_AWKINCLUDES += $(LIBUKSTORE_BASE)/uk_libs.in
```
...generates:
`(LIBUKSTORE_BUILD)/_store_libs.h`
from:
`$(LIBUKSTORE_BASE)/_store_libs.awk`
...as part of the `preprocess` build target. Note that the target name is derived from the source file name.

The purpose of this new build system extension is to make ther `Makefile.uk` more readable. Also, it can become really handy, for generating `.c` or `lds.S` files, the resulting files will be automatically included to the build process accordingly, e.g.,  `>.c` preprocessing target will cause also generating a C compile rule for this file.

lib/ukstore/gen_libs.h.awk
```diff
@@ -0,0 +1,22 @@
+
```
skuenzer: Empty newline at the beginning of the file? 😄

lib/ukstore/include/uk/store.h
skuenzer: For name the I would also suggest to add `ENTRY_TYPE` for better differentiation : `_LIB_UK_STORE_ENTRY_TYPE_<type>` (as inline with your `ENTRY_FLAG` definition).

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+
+/* Do not call directly */
+#define _UK_STORE_STATIC_ENTRY(entry, lib_str, e_type, e_get, e_set)	\
+	__UK_STORE_STATIC_ENTRY(entry, lib_str, e_type, e_get, e_set)
+
+/**
+ * Adds an entry to the entry section of a library.
+ *
+ * @param entry the entry in the section
+ * @param e_type the type of the function
```
skuenzer: maybe add also example of entry types:
```
@param e_type Type for the entry, e.g., s8, u16, charp
```

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+ * @param e_get getter pointer
```
skuenzer: I would also say that getter/setter are optional. Obviosly settin both to `NULL` at the same time would not make much sense, but it should not break anything if a user does it. So, no need to check arguments, just mention that `NULL` can be set, too as part of the parameter description.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+
+	/* Entry getter/setter type */
+	enum uk_store_entry_type type;
+
+	/* A cookie for extra storage */
+	void *cookie;
+} __align8;
+
+/* Used internally for storage */
+struct uk_store_folder;
```
skuenzer: Introduce this later, when intorducing the dynamic entries

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+	_uk_store_get_entry(uk_store_libid(libname), foldername, entryname)
+
+/**
+ * Frees the memory of an entry
+ *
+ * Guard this with `#if CONFIG_LIBUKSTORE`
+ *
+ * @param entry the entry to free
+ */
+#define uk_store_free_entry(entry)	\
```
skuenzer: I think this is also for later, right... when we have dynamic entries?

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+			const char *e_name)
+{
+	return _uk_store_get_static_entry(libid, e_name);
+}
+
+/**
+ * Searches for an entry in a folder in a library. Increases the refcount.
+ *
+ * @param libname the name of the library to search in
+ * @param foldername the name of the folder to search in
```
skuenzer: Add a comment that foldername is optional and reserved for future use with dynamic entries. Setting it to `NULL` makes sense

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+int
+_uk_store_get_uptr(const struct uk_store_entry *e, __uptr *out);
+int
+_uk_store_get_charp(const struct uk_store_entry *e, char **out);
+
+/**
+ * Calls the getter to get it's value
+ *
+ * @param entry the entry to call the getter from
+ * @param outtype the type of the variable used
```
skuenzer: give an example of type: e.g., `s8, u8, uptr`

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+int
+_uk_store_set_uptr(const struct uk_store_entry *e, __uptr val);
+int
+_uk_store_set_charp(const struct uk_store_entry *e, const char *val);
+
+/**
+ * Calls the set to set a new value value
+ *
+ * @param entry the entry to call the setter from
+ * @param intype the type of the variable used
```
skuenzer: I would also add here an example of types

lib/ukstore/store.c
```diff
@@ -29,3 +29,1186 @@
...
  */
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <uk/arch/limits.h>
+#include <uk/arch/types.h>
+#include <uk/store.h>
+
+struct uk_store_folder {
```
skuenzer: Can you only introduce the relevant structures and funtions for static entries? The others should be added with the dynamic entries patches.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+
+/**
+ * Returns the id of the place it was called from.
+ *
+ * @return this libraries' id
+ */
+#define uk_store_libid_self()	\
+	uk_store_libid(__LIBNAME__)
+
+/* All basic types that exist - use these when interacting with the API */
```
skuenzer: I would add the list of simple types so that people can check what they have to enter when using the interfaces of this library.

lib/ukstore/Config.uk
```diff
@@ -0,0 +1,5 @@
+config LIBUKSTORE
+	bool "ukstore: Run-time configuration and monitoring"
```
skuenzer: ukstore: Run-time configuring and monitoring

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,344 @@
...
+	_LIB_UK_STORE___s32   = 5,
+	_LIB_UK_STORE___u32   = 6,
+	_LIB_UK_STORE___s64   = 7,
+	_LIB_UK_STORE___u64   = 8,
+	_LIB_UK_STORE___uptr  = 9,
+	_LIB_UK_STORE___charp = 10
+};
+
+/* Transforms a simple data type to a ukstore specific one */
+#define uk_store_get_entry_type(simple_type)	\
```
skuenzer: Hum...  🤔 this could be a confusing name for the macro because it is close to `uk_store_entry_get_type()`, what do you think? What if we just name it `UK_STORE_ENTRY_TYPE()` because it is a pure pre-processor macro and not a getter function. 🤔 This would be inline with the macro `UK_STORE_ENTRY_ISSTATIC`, for example.

lib/ukstore/store.c
skuenzer: I see what you mean, yeah... In this particular case I would document that the provided charp setter function of the library needs to take care of duplicating the given string. Here, I would just forward the argument as `const char *`. More tricky and less obvious is the `getter` that returns a charp.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1186 @@
...
+	uk_store_entry_freed_func_t clean;
+
+	/* The linked list head */
+	struct uk_list_head list_head;
+
+	/* Refcount */
+	__atomic refcount;
+} __align8;
+
+#define UK_STORE_NR_MAX_LEN 32
```
skuenzer: I would suggest a different name. It wasn't clear to me what these definitions mean until I saw the code using them. Additionally, they are also internal only, so you do not need the `UK_STORE` prefix.
What would you think of:
```c
#define _U8_STRLEN (3 + 1) /* biggest num (127) has 3 digits plus terminating '\0' */
#define _S8_STRLEN (1 + _U8_STRLEN) /* space for potential sign in front */
#define _U16_STRLEN (5 + 1)
#define _S16_STRLEN (1 + _U16_STRLEN)
#define _U32_STRLEN (10 + 1)
#define _S32_STRLEN (1 + _U32_STRLEN)
#define _U64_STRLEN (19 + 1)
#define _S64_STRLEN (1 + _U64_STRLEN)
#define _UPTR_STRLEN (2 + 16 + 1) /* this is for uptr (64bit) in hex with space for leading `0x` prefix */
```

lib/ukstore/store.c
```diff
@@ -29,3 +29,1186 @@
...
+		return ret;
+	}
+
+	default:
+		return -EINVAL;
+	}
+}
+
+int
+_uk_store_get_charp(const struct uk_store_entry *e, char **out)
```
skuenzer, line 1074: For your question, I think we should document that the caller must free the string and then it is fine. The library provided charp getter function would also allocate a string for an output with a total custom size. I think it would be fine to tell the API user that strings need to be free'd.
We can provide another variant, like `_uk_store_get_ncharp(const struct uk_store_entry *e, char *out, size_t maxlen)`. Here, you handover a buffer to write to and its size. Things would then be truncated. You can use `snprintf()` within the implementation. In the charp-to-charp case you would need to do `snprintf(out, maxlen, "%s", val)`.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1186 @@
...
+		__u8 val;
+
+		str = calloc(UK_STORE_NR_SIZE8_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
+
+		ret = e->get.u8(e->cookie, &val);
+		if (ret < 0)
+			return ret;
+		sprintf(str, "%" __PRIu8, val);
```
skuenzer: if you use `snprintf` instead, the string will be `\0`-terminated in the target buffer - even if truncated but you size them big enough. You will also stay within bounds and you could just `malloc` the string buffer instead of `calloc`.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1186 @@
...
+		__uptr val;
+
+		str = calloc(UK_STORE_NR_SIZE64_LEN, sizeof(char));
+		if (unlikely(!str))
+			return -ENOMEM;
+
+		ret = e->get.uptr(e->cookie, &val);
+		if (ret < 0)
+			return ret;
+		sprintf(str, "%p", (void *) val);
```
skuenzer: I would suggest to use `"0x"__PRIx64` as format. `%p` is sometimes differently implementented across libcs. I have seen it once without leading `0x` prefix and once with.

skuenzer (changes requested):
Hey @craciunoiuc,
thanks for the separation of the PR. I have only a few comments to this PR, mostly style related. Please also move pieces (functions, symbols, etc.) that belong to dynamic entries to the dynamic entry PR.
Let me know when you are ready. In fact I would like to include this PR (static entries) to this release.
Thanks, Simon

lib/ukstore/store.c
skuenzer: Actually, double-check that we use the same expected format when we have a setter and get a charp as input. Ideally, the function should support receiving hex with `0x` prefix and without.

lib/ukstore/exportsyms.uk
craciunoiuc: don't know how I missed them 😅

lib/ukstore/gen_libs.h.awk
craciunoiuc: whoops, no checkpatch for awk 😄

lib/ukstore/include/uk/store.h
craciunoiuc: yup, again, missed ;(

lib/ukstore/include/uk/store.h
craciunoiuc: yup, makes sense

lib/ukstore/include/uk/store.h
craciunoiuc: in the comment right? adding them now

lib/ukstore/store.c
craciunoiuc: I'll change all of them now

lib/ukstore/store.c
craciunoiuc: I'll just do it like this for that case

```c
		if (strncmp(val, "0x", 2) == 0)
			ret = sscanf(val, "0x%" __SCNx64, (void **) &to_set);
		else
			ret = sscanf(val, "%" __SCNx64, (void **) &to_set);
```

lib/ukstore/store.c
craciunoiuc: I'm okay with `_uk_store_get_ncharp`, but, just so you know, that would change the define macro from:

```c
#define uk_store_get_value(entry, outtype, out)	\
	_uk_store_get_ ## outtype((entry), (out))
```

to something like

```c
#define uk_store_get_value(entry, outtype, out, maxlen)	\
	_uk_store_get_ ## outtype((entry), (out), (maxlen))
```

and this would change the signatures too. Is this ok, should I go for it?

lib/ukstore/Makefile.uk
craciunoiuc: Yup, no problem, I can start working on it for the next release. (0.10)

Just so I understood clearly: Nothing to be done for this PR right?

skuenzer: Yes 😄

lib/ukstore/store.c
skuenzer: Hum... I would actually treat the ncharp as an additional variant and keep what you have. maxlen wouldn't be usable for the other outtypes. The thing is also to define maxlen for getters that return charp already. Because our interface does imply any size limitations for charp it is likely that the output will be truncated with the ncharp varaint.

craciunoiuc: ok, I did it separately, like this

```c
#define uk_store_get_value_ncharp(entry, out, maxlen)	\
	_uk_store_get_ncharp((entry), (out), (maxlen))
```

\+ adding one new typedef/elem to the union/enum item of course

mandrei12 (commented):
I retested the static part and I did not find anything wrong with it.

craciunoiuc:
Hey @skuenzer,

Ready for round 2? (more like round N 😄)

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,350 @@
...
+ * The caller must provide a buffer with enough space to store the result.
+ * The max length of the result is given by maxlen.
+ *
+ * @param entry the entry to call the getter from
+ * @param out space to store the result of the call
+ * @param maxlen the maximum length of the result
+ * @return 0 or error on fail
+ */
+#define uk_store_get_value_ncharp(entry, out, maxlen)	\
+	_uk_store_get_ncharp((entry), (out), (maxlen))
```
skuenzer: I wouldnt put the macro in between for the ncharp type. Declare the symbol directly.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,350 @@
...
+/**
+ * Calls the getter to get it's value.
+ * The caller is responsible for freeing the returned value if it is a charp.
+ *
+ * @param entry the entry to call the getter from
+ * @param outtype the type of the variable used (e.g., u8, s8, u16, uptr, ...)
+ * @param out space to store the result of the call
+ * @return 0 or error on fail
+ */
+#define uk_store_get_value(entry, outtype, out)	\
```
skuenzer, line 310: This may fail when you specify ncharp as type because of the missing maxlen argument. I would actually exclude ncharp from the macro resolution. See my comment below....

lib/ukstore/store.c
```diff
@@ -29,3 +29,1267 @@
...
+
+		ret = e->get.uptr(e->cookie, &val);
+		if (ret < 0)
+			return ret;
+		snprintf(str, _UPTR_STRLEN, "0x%" __PRIx64, val);
+		*out  = str;
+		return ret;
+	}
+
+	case UK_STORE_ENTRY_TYPE(charp): {
```
skuenzer, line 1219: If you really add ncharp as entry type you need to handle it here too, but actually - I would exclude ncharp from the entry type and just have this one additional interface symbol.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,350 @@
...
+		uk_store_get_u8_func_t     u8;
+		uk_store_get_s16_func_t    s16;
+		uk_store_get_u16_func_t    u16;
+		uk_store_get_s32_func_t    s32;
+		uk_store_get_u32_func_t    u32;
+		uk_store_get_s64_func_t    s64;
+		uk_store_get_u64_func_t    u64;
+		uk_store_get_uptr_func_t   uptr;
+		uk_store_get_charp_func_t  charp;
+		uk_store_get_ncharp_func_t ncharp;
```
skuenzer: I wouldnt add the extra ncharp case currently because you will need to handle it in all switch cases as well.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1267 @@
...
+		ret = e->get.uptr(e->cookie, &val);
+		if (ret < 0)
+			return ret;
+		snprintf(*out, maxlen, "0x%" __PRIx64, val);
+		return ret;
+	}
+
+	case UK_STORE_ENTRY_TYPE(charp): {
+
+		ret = e->get.ncharp(e->cookie, out, maxlen);
```
skuenzer: Hum :thinking:. The entry is type of `charp` but a `ncharp` callback is called? I was assuming that we take the `charp` interface, do an `snprintf()` and then free the allocated buffer before returning this function. Of course, this is not zero-copy anymore but hopefully the `charp` interface is most rarely used.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1267 @@
...
+}
+
+
+int
+_uk_store_get_ncharp(const struct uk_store_entry *e, char **out, size_t maxlen)
+{
+	int ret;
+
+	UK_ASSERT(e);
+
```
skuenzer, line 1244: In all the getter variabts and setter variants you should make sure that a getter is available. Otherwise return `-ENOENT`. Since get is a union you can do a non-NULL check of just one preferred field. Please add this to all getter and setter API functions.

lib/ukstore/store.c
```diff
@@ -29,3 +29,1267 @@
...
+	}
+
+	default:
+		return -EINVAL;
+	}
+}
+
+
+int
+_uk_store_get_ncharp(const struct uk_store_entry *e, char **out, size_t maxlen)
```
skuenzer: since we do not chnage the allocation of the `out` buffer, a single reference is good enough:
```c
int
uk_store_get_ncharp(const struct uk_store_entry *e, char *out, size_t maxlen)
```

lib/ukstore/store.c
```diff
@@ -29,3 +29,1267 @@
...
+{
+	int ret;
+
+	UK_ASSERT(e);
```
skuenzer, line 1240: assert that out exists, e.g.:
`UK_ASSERT(out)`

skuenzer (changes requested):

lib/ukstore/include/uk/store.h
craciunoiuc: Sure

I wanted to keep it uniform, with the rest

lib/ukstore/store.c
craciunoiuc: makes sense, especially if we are removing `get.ncharp` anyway

lib/ukstore/store.c
craciunoiuc: I can just check if the enum is `_LIB_UK_STORE_ENTRY_TYPE___undef`  -  I will do it like this

craciunoiuc:
```
	if (unlikely(e->type == UK_STORE_ENTRY_TYPE(undef)))
		return -ENOENT;
```

craciunoiuc: But doesn't it already go in the default if the type is something other than the ones used?

```
	default:
		return -EINVAL;
	}
```

lib/ukstore/store.c
craciunoiuc: it will just fall in the default right? (if it's either undefined or something else)

lib/ukstore/store.c
craciunoiuc: I added the check anyway, hope it's okay

lib/ukstore/store.c
skuenzer: We have a Union so just one function type can be registered. If we call it with the wrong casting, it will fail.

lib/ukstore/store.c
skuenzer: It is for the case that we have get-only or set-only entries. Actually I would return -EIO if you set on a get-only entry (and vice versa). I think EIO is more appropriate for this case

skuenzer (commented):
Can you add the get-only/set-only check?

mandrei12 (approved):
I retested the static part and did not find anything wrong with it.

Reviewed-by: Andrei Mutu <andrei.mutu@stud.acs.upb.ro>

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,345 @@
...
+extern "C" {
+#endif
+
+#if !CONFIG_LIBUKSTORE
+#define uk_store_libid(libname) (-1)
+
+enum uk_store_entry_type;
+struct uk_store_entry;
+
+#define __UK_STORE_STATIC_ENTRY(entry, lib_str, e_type, e_get, e_set)	\
```
skuenzer: hum... did you try compiling instrumented code when the library is off. Because there is no definition of the struct and datatypes, I guess this registration will fail. I remember that we added this in order to keep type checking although it is not used. In fact - we may want to fix this by making the struct and the simple types always available.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,345 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+
+#ifndef __UK_STORE_H__
+#define __UK_STORE_H__
+
+#include <string.h>
```
skuenzer, line 38: When the library is not part of the built, this header that is always available should not introduce unnecessary dependencies. Your library can't express any if not part of a build.

skuenzer: So I would move header inclusion to line 97.

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,345 @@
...
+
+	/* Entry getter/setter type */
+	enum uk_store_entry_type type;
+
+	/* A cookie for extra storage */
+	void *cookie;
+} __align8;
+
+/* Flags if an entry is static or dynamic */
+#define UK_STORE_ENTRY_FLAG_STATIC	(1 << 0)
```
skuenzer, line 142: in order to avoid duplication, put common defintions (when library is on or off) before the guard checking if the library is part of the built: line 48

lib/ukstore/include/uk/store.h
```diff
@@ -29,3 +29,345 @@
...
+		e_type, e_get, e_set)
+
+/* Flags if an entry is static or dynamic */
+#define UK_STORE_ENTRY_FLAG_STATIC	(1 << 0)
+
+#define UK_STORE_ENTRY_ISSTATIC(entry)	\
+	((entry)->flags & UK_STORE_ENTRY_FLAG_STATIC)
+
+#define uk_store_get_entry(libname, foldername, entryname)	\
+	do { } while (0)
```
skuenzer: But, actually, the more I think about this, we should not provide any replacements for query calls. If someone wants to retrieve or set something from/to `ukstore`, that code has to make sure that the dependency to `lib/ukstore` is fulfilled. It is just convenience if we can transparently comment out with the preprocessor entry registration from libraries as soon as `ukstore` is not part of a built.

lib/ukstore/include/uk/store.h
skuenzer: When moving, make sure that headers that are required for definitions wheen the library is off, stay here. I hope these are only general unikraft headers (`/include`) and/or libc provided headers.

skuenzer (changes requested):
Hey @craciunoiuc , I just found now that we have some problems with compiling instrumented code when the library is not selected as part of a build. Can you double-check? Sorry that I did not see this earlier.
Simon

lib/ukstore/include/uk/store.h
craciunoiuc: Ok, I'm deleting everything except the entry registration

lib/ukstore/include/uk/store.h
craciunoiuc: I think there is nothing that is common between the two if/else cases. I'll just move the `#endif` lower.

lib/ukstore/include/uk/store.h
craciunoiuc: There are no library dependencies when the library is off, it's fine

lib/ukstore/include/uk/store.h
craciunoiuc: I don't know how this got here 😅 I'm removing it now

mandrei12:
It still works after the changes.

skuenzer (approved):
Thanks a lot @craciunoiuc for the quick changes! This looks good.

PR #461, opened 2022-05-12
Introduce initial riscv64 support

### LINKED ISSUE #60: RISC-V Support
Add support for RISC-V architecture.

**Summary of objectives**

- [x] #461 
- [ ] Add SMP support.
- [ ] Do code restructuring during/after platform rearchitecting.
- [ ] Add any other features that may be currently missing from the RISC-V port and are present in other architectures.

**Depends on**: #586

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [`riscv64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_ARCH_RISCV_64=y`

### Introduction 

The PR specifically targets the KVM platform (the Xen hypervisor hasn't been ported yet to RISC-V). Since there is currently no RISC-V board out there with the Hypervisor extension which adds hardware-assisted virtualization required by KVM, testing can be done using `qemu-system-riscv64` with the `virt` board for the moment.

A new entry called `RISCV64 compatible` has been added in the "Architecture Selection" menu in `menuconfig`. It targets a `RISCV64GC` ISA more precisely. Platform and library configuration process is the same as with the other architectures.

When cross-compiling, the build system invokes the `riscv64-linux-gnu-` toolchain which can already be found in the repositories of popular Linux distributions, so there is no need to compile the toolchain manually from source. A slightly modified `qemu-guest` shell script which can launch an emulated riscv64 instance using the `-W -t riscv64gc` commandline argument is provided with this PR.

### RISC-V privilege modes and the Supervisor Binary Interface (SBI)
There are 3 main RISC-V privilege modes: `machine`, `supervisor` and `user` modes (additionally, the Hypervisor extension introduces the `virtual supervisor` and `virtual user` modes and changes the `supervisor` mode to `hypervisor` mode). This `riscv64` port of Unikraft has been written such that it always executes in `S-mode` (or `VS-mode` for that matter - the RISC-V Spec states that `S-mode` software can run without modfication as `VS-mode` guests, hence we'll use `S`/`VS` and `M`/`HS` modes interchangeably throughout this text). Even the very first instructions found at the entry point `_libkvmplat_entry` are executed in `S/VS-mode`. This is because we expect that Unikraft is running on top of an SBI implementation, which is a firmware commonly present in RISC-V systems; for example, `qemu` by default loads `OpenSBI`, an open-source SBI reference implementation, as a bios when launching an emulated RISC-V instance, while `KVM` provides its own SBI implementation for virtualized instances. 

The SBI provides an abstraction for platform specific functionality through an interface between `S-mode` and `M-mode` (or `VS-mode` and `HS-mode`) , as described in the "RISC-V SBI" spec. So the SBI implementation runs at a higher privilege than our kernel and at boot eventually jumps to `S/VS-mode` to a fixed address where our Unikraft kernel image is loaded (no particular bootloader is necessary for Unikraft to run). This Unikraft port then relies on the SBI for system shutdown, programming timer interrupts, and in the future possibly inter-processor interrupts. All the SBI functions that are used by Unikraft can be found in the `sbi.h` header in `plat/common/include/riscv`

### Similarities with the ARM port
This port tries to follow roughly the same directory structures and bootflow of the other architectures. It shares the most similarity with the ARM port: both are mmio based and do their platform configuration and devices discovery through the device tree blob, for example. As such, some of the code is reused from the ARM port: for example the `plat/kvm/riscv/setup.c` file is very similar to `plat/kvm/arm/setup.c`. 

### Paging
This PR also includes an implementation of the new Paging API for the `Sv39` addressing mode using 3 levels of page tables to setup a virtual address space that is similar to the x86 one. An initial static page table is provided, which is later used as a basis for a dynamic one if CONFIG_PAGING is selected.  There is a "directmap" area starting from 256GiB which is mapped 1:1 directly into the physical memory starting from 0x0, the heap and stack are also mapped roughly in the same areas as in the x86 implementation. The main difference is that we also identity map the first 2GiB of physical address space which we use on RISC-V for MMIO. Maybe a better solution would be to use the directmap area from the virtual address space to do MMIO, but we should be careful to translate the physical addresses parsed from the DTB into the appropriate virtual addresses if we remove the identity mapping.

### Traps
Traps are set up in a straightforward manner: in the `_init_traps` function from `plat/kvm/riscv/traps.c` we set the Direct (not Vectored) trap handling mode which causes the program counter to be set to the address of the assembly routine `__trap_handler` from `exceptions.S` for all traps taken into `S-mode`. By default, all traps are taken into `M-mode` unless they are explicitly delegated by the firmware/SBI implementation to a lower privilege mode. While we can't control exactly which traps are handled by Unikraft, we can safely assume that the `Supervisor timer interrupt`, `Supervisor external interrupt` and `Supervisor software interrupt` described in the RISC-V privileged specification are delegated to `S-mode`, where Unikraft runs. However, if one wishes to implement the binary compatibility layer, they must take into account that `ECALL` instructions are used in calling SBI functions and thus generate `Environment call from S-mode` exceptions which are taken in `M-mode` (or more appropriately, `HS-mode`, if running in a virtual machine). They can't be used for horizontal traps.

### Interrupts
Regarding interrupts: a RISC-V platform can implement various interrupt controllers, such as the PLIC, APLIC, CLIC, CLINT or ACLINT, each with their own separate specification. Some even complement each other, so not all of them are mutually exclusive. This Unikraft port relies on a PLIC + (CLINT or ACLINT) scheme, where the PLIC handles interrupts generated by I/O devices and sends them to a RISC-V hart (hardware thread) by raising a `Supervisor external interrupt` trap. Meanwhile, timer and inter-processor interrupts are handled separately by the CLINT/ACLINT which raises a different trap, `Supervisor timer interrupt` and `Supervisor software interrupt`, respectively. A basic PLIC driver can be found in `plat/common/riscv/plic.c`. The CLINT/ACLINT is programmed indirectly through SBI function calls such as `sbi_set_timer`, which we can use to set time alarms, for example. Using SBI is necessary because both the CLINT and the ACLINT have memory-mapped registers which can mainly be accessed `M-mode` software only.

### PCI and MMIO
PCI and MMIO probing is virtually identical to the existing implementation written for ARM targets. QEMU emulates the exact same PCI host controller for RISC-V platforms (`pci-host-ecam-generic`). Information about the devices (base addresses, IRQs etc.) are parsed from the DTB as well, where the most significant difference from the ARM implementation is that we don't need to translate the IRQs from the DTB to absolute values.

### Clocks
Both monotonic and wall time clocks are supported. QEMU emulates a Goldfish RTC device for RISC-V targets, so a basic driver for that device has been added.

### Newlib and Musl compatibility
Newlibc was also ported in https://github.com/unikraft/lib-newlib/pull/18, and applications such as `Redis`, `NGINX`, `SQLite` and `Python` have been tested and seem to run successfully! Moreover, `musl` was also ported as part of  https://github.com/unikraft/lib-musl/pull/39, and thus applications can be run on the latest version of Unikraft.


eduardvintila:
Rewrote the whole commit history, fixed (_hopefully_) most of the checkpatch issues, adapted to the latest version of Unikraft (0.12) and also to `lib-musl`.

gdace829:
do you finish this issue?

gdace829:
do you finish this issue?

PR #462, opened 2022-05-15
Fix Warning from Redis app

ghost:
This PR consists in fixing warnings that come up when building the redis app. It has the following modifications 

* plat/linuxu:  Verify that SA_SIGINFO is not yet defined before defining it signal.h
* lib/uksignal: replace uk_syscall_r_getpid with uk_syscall_r_stub
* plat/xen/drivers/9p: add #define _GNU_SOURCE to make defined functions visible during compilation.

### DIFF
--- a/lib/uksignal/signal.c
+++ b/lib/uksignal/signal.c
@@ -360,7 +360,7 @@ int killpg(int pgrp, int sig)
 		return -1;
 	}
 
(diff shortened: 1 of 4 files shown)


ouss-ch (commented):
I was not able to reproduce the warnings. So it looks like this is not connected to the issue? 
Reviewed by Oussama "ouss-chayeb@gmail.com"

PR #463, opened 2022-05-15
Update on the Makefile.uk to fix a -Wnonnull-compare warning when compiling the nolibc library.

MoubarakZoure:
I disabled the Wno-null-compare in the Makefile.uk of the nolibc library to mitigate the following warning :
```
/home/ubuntu/workdir/unikraft/lib/nolibc/string.c:695:5: warning: nonnull argument ‘src’ compared to NULL [-Wnonnull-compare]
  695 |  if (src!=NULL) {
```
However, I am wondering if the check "`( 695 |  if (src!=NULL) {`)" in the string.c could not simply be removed to mitigate the warning message. I think we should choose either to remove the rule that constraint that restricts the "src" not to be NULL, or the " `695 |  if (src!=NULL) {"` check control that ensures that src is not NULL.

### Prerequisite checklist

### Base target

### DIFF
--- a/lib/nolibc/Makefile.uk
+++ b/lib/nolibc/Makefile.uk
@@ -14,6 +14,7 @@ LIBNOLIBC_NO_BUILTINS		+= -fno-builtin-sscanf
 LIBNOLIBC_NO_BUILTINS		+= -fno-builtin-vscanf
 LIBNOLIBC_NO_BUILTINS		+= -fno-builtin-vfscanf
 LIBNOLIBC_NO_BUILTINS		+= -fno-builtin-vsscanf
+LIBNOLIBC_NO_BUILTINS		+= -Wno-nonnull-compare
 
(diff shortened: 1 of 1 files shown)


MoubarakZoure:
I disabled the Wno-null-compare in the Makefile.uk of the nolibc library to mitigate the following warning : "/home/ubuntu/workdir/unikraft/lib/nolibc/string.c:695:5: warning: nonnull argument ‘src’ compared to NULL [-Wnonnull-compare]
695 | if (src!=NULL) {"
However, i wondering if the check "( 695 | if (src!=NULL) {)" in the string.c could not simply be removed to mitigate the warning message. I think we should choose either to remove the rule that constraint that restricts the "src" not to be NULL, or the " 695 | if (src!=NULL) {" check control that ensures that src is not NULL.
I tested my patch on the linuxu x86_64 platform.

gabrielmocanu:
@razvand  we need a reviewer here.

razvand:
This is already handled by #546. I will close this PR.

Issue #414, opened 2022-02-18
Security: Code Segment is Writable


lrapp-x41-pub:
### Describe the bug

According to the W^X policy, a memory page should be executable xor writable, but never both. This is a common security hardening technique to hinder an attacker from writing exploit code to a memory segment and execute it afterwards.
When trying to execute code on the stack or heap, unikraft stops with a page fault error as expected. Modifying a function in the code segment works and does not trigger any page fault error.
An attacker able to exploit an arbitrary write vulnerability could use this to place exploit code in the code segment or manipulate existing functions. This is especially problematic  without ASLR.

### Steps to reproduce

Compile and execute the proof of concept code:

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

void __attribute__ ((noinline)) call_target() {
	printf("Called target function!\n");
}


int main(int argc, char *argv[]) {
	printf("----- W^X ------\n");
	// Test code segment write permission
	printf("[info] Print unmodified function:\n");
	printf("func: %lx @ %p\n", *((long*)call_target), &call_target);
	printf("[info] Execute unmodified function:\n");	
	call_target();
	printf("[info] Attempting to overwrite function code with 'ret':\n");
	// Overwrite call_target() function with a simple return statement
	// thus, calling the call_target() function will not print anything
	*((long*)call_target) = 0xccccccc3;
	printf("[info] Written!\n");
	printf("[info] Print modified function:\n");
	printf("func: %lx @ %p\n", *((long*)call_target), &call_target);
	printf("[info] Execute modified function:\n");	
	call_target();
	printf("[info] Done!\n");
}
```

Output:
```
SeaBIOS (version 1.15.0-1.fc35)
Booting from ROM...
Powered by
o.   .o       _ _               __ _
Oo   Oo  ___ (_) | __ __  __ _ ' _) :_
oO   oO ' _ `| | |/ /  _)' _` | |_|  _)
oOo oOO| | | | |   (| | | (_) |  _) :_
 OoOoO ._, ._:_:_,\_._,  .__,_:_, \___)
                    Mimas 0.7.0~da0f1ee
----- W^X ------
[info] Print unmodified function:
func: 2d058b4818ec8348 @ 0x109eb0
[info] Execute unmodified function:
Called target function!
[info] Attempting to overwrite function code with 'ret':
[info] Written!
[info] Print modified function:
func: ccccccc3 @ 0x109eb0
[info] Execute modified function:
[info] Done!
```

### Expected behavior

[...]

PR #465, opened 2022-05-20
Enforce W^X to counter code inject attacks

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [`kvm`, `xen`]
 - Application(s): [N/A]

### Additional configuration

- `CONFIG_LIBUKBOOT_ENFORCE_W_XOR_X=y`

### Description of changes

When all memory regions use default memory protections (RWX) it is easy to inject shellcode. For example, an attack could use a stack buffer overflow to overwrite the return address and hijack execution to run the injected code. Another example would be to use a write primitive to modify the .text segment so that during regular program execution, attacked-controller code is
run.

The standard method to mitigate these simple attacks is to enforce write XOR execute (W^X) permissions - meaning that a memory region can either be writeable or executable but not both. This way, all memory that is executable cannot be changed at runtime and every memory that is writeable is not executable.

The PR at hand implements W^X using the paging API by enforcing the respective page permissions during boot in `ukboot`. Doing it in `ukboot` has the advantage that the solution is agnostic to the platform. A disadvantage is that it causes additional boot time overhead. It should therefore also be investigated, how to build the static page tables with the proper protections already in. This would also enable the feature for unikernels without the paging API.

Issue #414

### DIFF
--- a/lib/ukboot/Config.uk
+++ b/lib/ukboot/Config.uk
@@ -117,4 +117,13 @@ if LIBUKBOOT
 		bool "None"
 
 	endchoice
+
+	config LIBUKBOOT_ENFORCE_W_XOR_X
+		bool "Enforce W^X"
+		default n
+		depends on PAGING
+		help
+			During boot the protection settings of all memory
+			ranges are updated to enforce write XOR execute (W^X)
+			and mitigate code injection attacks.
 endif
--- a/include/uk/plat/memory.h
+++ b/include/uk/plat/memory.h
@@ -45,11 +45,12 @@ extern "C" {
 /* Memory region flags */
 #define UKPLAT_MEMRF_FREE	(0x1)	/* Region with uninitialized memory */
 #define UKPLAT_MEMRF_RESERVED	(0x2)	/* Region is in use by platform */
-#define UKPLAT_MEMRF_EXTRADATA	(0x4)	/* Contains additional loaded data
-					 * (e.g., initramdisk)
+#define UKPLAT_MEMRF_EXTRADATA	(0x4)	/* Contains additional loaded data \
+					 * (e.g., initramdisk) \
 					 */
 #define UKPLAT_MEMRF_READABLE	(0x10)	/* Region is readable */
 #define UKPLAT_MEMRF_WRITABLE	(0x20)	/* Region is writable */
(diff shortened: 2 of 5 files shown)


vladandrew:
I have stated looking over this pull request! It's a really  important addition to Unikraft.

lib/ukboot/Config.uk
```diff
@@ -117,4 +117,13 @@ if LIBUKBOOT
 		bool "None"
 
 	endchoice
+
+	config LIBUKBOOT_ENFORCE_W_XOR_X
```
gabrielmocanu, line 121: Hi @marcrittinghaus.

I started to look at this PR and I saw that this feature depends on PAGING.
Wouldn't be a better idea to use select?
```
select PAGING
```
Because it is a configuration that does not come by default, it is quite difficult to find by activating PAGING first.
What do you think about this?

gabrielmocanu (commented):
I tried to run a simple hello world on KVM with this feature enabled and I got this error:

Screenshot from 2022-05-28 13-32-21

Should I enable something else besides Virtual Memory API?

vladandrew:
Hi @marcrittinghaus ,

Should we upstream this even if the `ukplat_memregion_*` API is not complete yet?

marcrittinghaus:
> Hi @marcrittinghaus ,
> 
> Should we upstream this even if the `ukplat_memregion_*` API is not complete yet?

I would prefer to have the memory regions enumerated properly first. The thing is that you quickly run into issues with permission settings if the regions are not accurately represented (as it probably was the case for @gabrielmocanu). Unfortunately, we don't have any actual plans at the moment on how to do this. I think optimally, you would incorporate:
- information on ELF segments if available
- symbols like `_TEXT` only if no ELF headers are available
- boot information like multiboot memory regions

But I think QEMU does not load the ELF headers into the VM, does it?

lib/ukboot/Config.uk
marcrittinghaus: Yes, makes sense 👍🏻

marcrittinghaus:
This functionality will be part of #506 and eventually be superseded with a compile-time built page table.

PR #468, opened 2022-05-31
lib/syscall_shim: Ignore unnumbered system calls

skuenzer:
### Base target

 - Architecture(s): all
 - Platform(s): all
 - Application(s): all

### Additional configuration

 - `CONFIG_LIBSYSCALL_SHIM=y`

### Description of changes

This PR adds a warning issued by the pre-processor for system calls that do not have a system call number defined. This can happen when a system call has been registered to `UK_PROVIDED_SYSCALLS-y`, but no system call number is assigned in the architecture-dependent `syscall.h.in`. `lib/syscall_shim` will not be linked with such ignored system calls.

This PR is the first step in making the `syscall_shim` compilable on Arm. It is intended to replace PR #422 and PR #336. In a later step, it is currently planned to distinguish between 1) actual system call implementations, 2) legacy system call interfaces and 3) architecture-dependent system calls. The idea is that legacy interfaces can be treated as optional because they are less powerful system call interfaces of a system call that is only available on Linux architectures that have been supported longer, such as `open` as a legacy interface to `openat`. Architecture-dependent system calls are system calls that exist only for a particular architecture, such as `arch_prctl` on x86. The `syscall_shim` will treat actual system calls and architecture-dependent system calls as mandatory; legacy system calls will be treated as optional, and the `syscall_shim` will silently ignores legacy definitions if no number mapping exists. Since this requires more code adoptions, we plan to complete the work with the 0.10 release.

### DIFF
--- a/lib/syscall_shim/gen_provided.awk
+++ b/lib/syscall_shim/gen_provided.awk
@@ -1,13 +1,18 @@
 BEGIN {
 	print "/* Automatically generated file; DO NOT EDIT */"
 	print "#ifndef __LIBSYSCALL_SHIM_PROVIDED_SYSCALLS_H__"
-	print "#define __LIBSYSCALL_SHIM_PROVIDED_SYSCALLS_H__\n"
+	print "#define __LIBSYSCALL_SHIM_PROVIDED_SYSCALLS_H__"
+	print "\n#include <uk/bits/syscall_nrs.h>"
 }
 
 /[a-zA-Z0-9]+-[0-9]+/{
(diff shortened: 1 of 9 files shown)


craciunoiuc (approved):
Hey @skuenzer @razvand,

I think everything is good here and it's safe to add the review tag (the deadline is close too).
I tested the PR and it seems to do what it says and there's no missing spots.

The only thing I don't get is why newlines are at the beginning and not at the end, but probably that's just preference.

razvand:
Thanks, @skuenzer . This works. We'll improve it for release 0.10.

PR #469, opened 2022-06-01
plat/common: Architecture-independent parts of SMP API

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_UKPLAT_LCPU_MAXCOUNT > 1`

### Description of changes

This PR provides an implementation of the architecture-independent parts of the SMP API based on the common code developed in the PRs for x86 (#244) and ARM64 (#373). The architecture implementation must provide the following functions as described in `plat/common/include/uk/plat/common/lcpu.h`:

- `lcpu_arch_mp_init` which is executed on the BSP by `lcpu_mp_init()` during system boot to allow architecture-dependent initialization of the multi-processor functionality. This function should perform CPU enumeration and allocate LCPUs using `lcpu_alloc()`. The first LCPU must be the BSP itself (it does not need to be explicitly allocated).
- `lcpu_arch_init()` which is executed on the a CPU when calling `lcpu_init()`. This function can be used to initialize extended CPU features or setup traps and interrupts for the newly started CPU.
- `lcpu_arch_id()` which queries the hardware ID of the CPU which executes the function using architecture-dependent instructions / facilities.
- `lcpu_arch_start()` which signals a certain CPU to begin with the startup sequence (e.g., by issuing a special interprocessor interrupt (IPI))
- `lcpu_arch_post_start()` [OPTIONAL] if the architecture defines `LCPU_ARCH_MULTI_PHASE_STARTUP` and which is called in `ukplat_lcpu_start()` after all LCPUs specified in the call have received an `lcpu_arch_start()` call. This can be used to speed up starting multiple CPUs at the same time on some architectures like x86.
- `lcpu_arch_run()` which queues a function supplied to `ukplat_lcpu_run()` to the given LCPU and signals the corresponding CPU to wake up and dispatch queued functions. The common code expects this to be triggered by a special IPI whose vector has to be defined via `LCPU_ARCH_IPI_RUN_IRQ`
- `lcpu_arch_wakeup()` which forces the given CPU to wake up from a temporary halt state. The common code expects this to be triggered by a special IPI whose vector has to be defined via `LCPU_ARCH_IPI_WAKEUP_IRQ`. The architecture may decide to implement a very lightweight IRQ handler for this vector which just acknowledges the IRQ and returns. However, in that case, the architecture must ensure that no other handlers will be registered for this vector.

The platform boot code (usually in `setup.c`) should call `lcpu_init()` to setup the current BSP and afterwards call `lcpu_mp_init()` to initialize multi-processor functionality including CPU discovery. The boot code should crash the system if any of these initializations fail. If `lcpu_init()` fails on a secondary CPU, the CPU should be put into a halt state using `lcpu_halt()` with an error code indicating the reason. Secondary CPUs must not call `lcpu_mp_init()`.

### DIFF
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -109,9 +109,9 @@ struct ukplat_lcpu_func {
 	 * Function to execute.
 	 *
 	 * @param regs pointer to a snapshot of the current CPU state. Changes
-	 *   to the state are applied after the function returns
+	 *   to the state are applied after the RUN IRQ handler returns
 	 * @param fn pointer to this structure. The structure is not touched
-	 *   after function invocation and can be freed if necessary during
+	 *   after function invocation and must be freed if necessary during
 	 *   function execution
 	 */
 	void (*fn)(struct __regs *regs, struct ukplat_lcpu_func *fn);
@@ -126,16 +126,18 @@ struct ukplat_lcpu_func {
 __lcpuid ukplat_lcpu_id(void);
 
 /**
- * Returns the number of logical CPUs present on the system
+ * Returns the number of logical CPUs present in the system
  */
 __lcpuid ukplat_lcpu_count(void);
 
 /**
  * Starts multiple logical CPUs and assigns them the given stacks. The logical
  * CPUs execute the entry functions if supplied or enter a low-power wait state
- * otherwise.
+ * otherwise. CPUs that have already been started are ignored.
  *
- * @param lcpuid array with the IDs of the logical CPUs that are to be started
+ * @param lcpuid array with the IDs of the logical CPUs that are to be started.
+ *   Can be NULL to include all logical CPUs except the one executing the
+ *   function.
  * @param sp array of stack pointers, one for each specified logical CPU
  * @param entry optional array of entry function pointers, one for each
  *   specified logical CPU. Provided functions must not return. If the
@@ -174,7 +176,7 @@ int ukplat_lcpu_wait(__lcpuid lcpuid[], unsigned int num, __nsec timeout);
  * @param num number of logical CPU IDs given
  * @param fn the function to be executed
  * @param flags (architecture-dependent) flags that specify how the function
- *   should be executed (see UKPLAT_LCPU_RFLG_* flags)
+ *   should be executed (see UKPLAT_LCPU_RFLG_* flags if available)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
@@ -183,6 +185,11 @@ int ukplat_lcpu_run(__lcpuid lcpuid[], unsigned int num,
 
 /**
  * Wakes up the specified logical CPUs from a halt or low-power sleep state.
+ *
+ * @param lcpuid array with the IDs of the logical CPUs that should be waked up.
+ *   Can be NULL to execute the function on all logical CPUs except the current
+ *   one.
+ * @param num number of logical CPU IDs given
  *
  * @return 0 on success, an errno-type error value otherwise
  */
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -149,8 +149,11 @@ __u32 ukplat_lcpu_count(void);
  *   sequential order of lcpuidx. If the call succeeds, input and output values
  *   are equal. Must be NULL if lcpuidx is NULL
  * @param sp array of stack pointers, one for each logical CPU to start. If
- *   lcpuidx is NULL, must be ukplat_lcpu_count() - 1 stacks
- * @param entry array of entry function pointers, one for logical CPU to start.
+ *   lcpuidx is NULL, must be ukplat_lcpu_count() - 1 stack pointers. The
+ *   stacks may be specifically prepared to contain arguments for the entry
+ *   function (e.g., cdecl calling convention). The platform may use the
+ *   following stack space to execute initialization routines
+ * @param entry array of entry functions, one for each logical CPU to start.
  *   Can be NULL, otherwise if lcpuidx is NULL, must contain
  *   ukplat_lcpu_count() - 1 function pointers. Provided functions must not
  *   return. If the parameter or individual function pointers are NULL the
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -138,17 +138,17 @@ __lcpuid ukplat_lcpu_count(void);
  * @param lcpuid array with the IDs of the logical CPUs that are to be started.
  *   Can be NULL to include all logical CPUs except the one executing the
  *   function.
+ * @param num number of logical CPU IDs given
  * @param sp array of stack pointers, one for each specified logical CPU
  * @param entry optional array of entry function pointers, one for each
  *   specified logical CPU. Provided functions must not return. If the
  *   parameter or individual elements are NULL the respective logical CPUs
  *   enter a low-power wait state after startup
- * @param num number of logical CPU IDs given
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_start(__lcpuid lcpuid[], void *sp[],
-		      ukplat_lcpu_entry_t entry[], unsigned int num);
+int ukplat_lcpu_start(__lcpuid lcpuid[], unsigned int num, void *sp[],
+		      ukplat_lcpu_entry_t entry[]);
 
 /**
  * Waits for the specified logical CPUs to enter idle state, or until the
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -122,7 +122,7 @@ struct ukplat_lcpu_func {
 	 *   after function invocation and must be freed if necessary during
 	 *   function execution
 	 */
-	void (*fn)(struct __regs *regs, struct ukplat_lcpu_func *fn);
+	void (*fn)(struct __regs *regs, const struct ukplat_lcpu_func *fn);
 
 	/* Optional user-supplied pointer. */
 	void *user;
@@ -144,9 +144,13 @@ __u32 ukplat_lcpu_count(void);
  * otherwise. CPUs that have already been started are ignored.
  *
  * @param lcpuidx array with the indices of the logical CPUs that are to be
- *   started. Can be NULL to include all logical CPUs except the one executing
- *   the function
- * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
+ *   started. CPUs are started in the order specified in the array. Can be NULL
+ *   to include all logical CPUs except the one executing the function, in which
+ *   case CPUs are started in sequential order according to their CPU index
+ * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
+ *   elements in lcpuidx, and [OUT] the number of successfully started CPUs in
+ *   sequential order of lcpuidx. If the call succeeds, input and output values
+ *   are equal. Must be NULL if lcpuidx is NULL
  * @param sp array of stack pointers, one for each logical CPU to start. If
  *   lcpuidx is NULL, must be ukplat_lcpu_count() - 1 stacks
  * @param entry array of entry function pointers, one for logical CPU to start.
@@ -159,8 +163,8 @@ __u32 ukplat_lcpu_count(void);
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_start(__lcpuidx lcpuidx[], unsigned int num, void *sp[],
-		      ukplat_lcpu_entry_t entry[], unsigned long flags);
+int ukplat_lcpu_start(const __lcpuidx lcpuidx[], unsigned int *num, void *sp[],
+		      const ukplat_lcpu_entry_t entry[], unsigned long flags);
 
 /**
  * Waits for the specified logical CPUs to enter idle state, or until the
@@ -168,14 +172,18 @@ int ukplat_lcpu_start(__lcpuidx lcpuidx[], unsigned int num, void *sp[],
  *
  * @param lcpuidx array with the indices of the logical CPUs to wait for. Can
  *   be NULL to include all logical CPUs except the one executing the function
- * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
+ * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
+ *   elements in lcpuidx, and [OUT] the number of CPUs in idle state until the
+ *   timeout expired in sequential order of lcpuidx. If the call succeeds,
+ *   input and output values are equal. Must be NULL if lcpuidx is NULL
  * @param timeout timeout in nanoseconds for the wait to be satisfied. Can be
  *   0 to wait indefinitely
  *
  * @return 0 if the wait for all specified logical CPUs has been satisfied,
  *   an errno-type error value otherwise (e.g., timeout)
  */
-int ukplat_lcpu_wait(__lcpuidx lcpuidx[], unsigned int num, __nsec timeout);
+int ukplat_lcpu_wait(const __lcpuidx lcpuidx[], unsigned int *num,
+		     __nsec timeout);
 
 /**
  * Executes a function on the specified logical CPUs. The run function does not
@@ -185,26 +193,36 @@ int ukplat_lcpu_wait(__lcpuidx lcpuidx[], unsigned int num, __nsec timeout);
  * @param lcpuidx array with the indices of the logical CPUs that should
  *   execute the function. Can be NULL to execute the function on all logical
  *   CPUs except the current one
- * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
+ * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
+ *   elements in lcpuidx, and [OUT] the number of CPUs on which the function has
+ *   been successfully queued in sequential order of lcpuidx. If the call
+ *   succeeds, input and output values are equal. Must be NULL if lcpuidx is
+ *   NULL
  * @param fn the function to be executed
  * @param flags (architecture-dependent) flags that specify how the function
- *   should be executed (see UKPLAT_LCPU_RFLG_* flags if available)
+ *   should be executed (see UKPLAT_LCPU_RFLG_* flags)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_run(__lcpuidx lcpuidx[], unsigned int num,
-		    struct ukplat_lcpu_func *fn, unsigned long flags);
+int ukplat_lcpu_run(const __lcpuidx lcpuidx[], unsigned int *num,
+		    const struct ukplat_lcpu_func *fn, unsigned long flags);
+
+/* Do not block while trying to queue the function to the remote core */
+#define UKPLAT_LCPU_RFLG_DONOTBLOCK	0x1
 
 /**
  * Wakes up the specified logical CPUs from a halt or low-power sleep state.
  *
  * @param lcpuidx array with the indices of the logical CPUs that should be
- *   woken up. Can be NULL to wakeup all logical CPUs except the current one.
- * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
+ *   woken up. Can be NULL to wakeup all logical CPUs except the current one
+ * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
+ *   elements in lcpuidx, and [OUT] the number of successfully woken up CPUs in
+ *   sequential order of lcpuidx. If the call succeeds, input and output values
+ *   are equal. Must be NULL if lcpuidx is NULL
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_wakeup(__lcpuidx lcpuidx[], unsigned int num);
+int ukplat_lcpu_wakeup(const __lcpuidx lcpuidx[], unsigned int *num);
 
 #else /* CONFIG_HAVE_SMP */
 #define ukplat_lcpu_idx()	(0)
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -80,7 +80,7 @@ void ukplat_lcpu_irqs_handle_pending(void);
 /**
  * Halts the current logical CPU execution
  */
-void ukplat_lcpu_halt(void);
+void __noreturn ukplat_lcpu_halt(void);
 
 /**
  * Halts the current logical CPU. Execution is resumed when an interrupt/signal
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -144,11 +144,13 @@ __lcpuid ukplat_lcpu_count(void);
  *   specified logical CPU. Provided functions must not return. If the
  *   parameter or individual elements are NULL the respective logical CPUs
  *   enter a low-power wait state after startup
+ * @param flags (architecture-dependent) flags that specify how to start the
+ *   CPUs (see UKPLAT_LCPU_SFLG_* flags if available)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
 int ukplat_lcpu_start(__lcpuid lcpuid[], unsigned int num, void *sp[],
-		      ukplat_lcpu_entry_t entry[]);
+		      ukplat_lcpu_entry_t entry[], unsigned long flags);
 
 /**
  * Waits for the specified logical CPUs to enter idle state, or until the
@@ -181,7 +183,7 @@ int ukplat_lcpu_wait(__lcpuid lcpuid[], unsigned int num, __nsec timeout);
  * @return 0 on success, an errno-type error value otherwise
  */
 int ukplat_lcpu_run(__lcpuid lcpuid[], unsigned int num,
-		    struct ukplat_lcpu_func *fn, int flags);
+		    struct ukplat_lcpu_func *fn, unsigned long flags);
 
 /**
  * Wakes up the specified logical CPUs from a halt or low-power sleep state.
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -128,7 +128,7 @@ __lcpuid ukplat_lcpu_id(void);
 /**
  * Returns the number of logical CPUs present in the system
  */
-__lcpuid ukplat_lcpu_count(void);
+__u32 ukplat_lcpu_count(void);
 
 /**
  * Starts multiple logical CPUs and assigns them the given stacks. The logical
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -96,11 +96,19 @@ void ukplat_lcpu_halt_to(__nsec until);
  */
 void ukplat_lcpu_halt_irq(void);
 
+/* Non-prototyped logical CPU entry function */
+typedef void __noreturn (*ukplat_lcpu_entry_t)();
+
+typedef __u32 __lcpuidx;	/* Sequential index of logical CPU */
+typedef __u64 __lcpuid;		/* Physical ID of logical CPU */
+
+/**
+ * Returns the hardware ID of the current logical CPU
+ */
+__lcpuid ukplat_lcpu_id(void);
+
 #ifdef CONFIG_HAVE_SMP
 #include <uk/list.h>
-
-typedef void (*ukplat_lcpu_entry_t)(void) __noreturn;
-typedef __u32 __lcpuid;
 
 struct ukplat_lcpu_func {
 	struct uk_list_head lentry;
@@ -121,9 +129,9 @@ struct ukplat_lcpu_func {
 };
 
 /**
- * Returns the ID of the current logical CPU
+ * Returns the index of the current logical CPU
  */
-__lcpuid ukplat_lcpu_id(void);
+__lcpuidx ukplat_lcpu_idx(void);
 
 /**
  * Returns the number of logical CPUs present in the system
@@ -135,70 +143,71 @@ __u32 ukplat_lcpu_count(void);
  * CPUs execute the entry functions if supplied or enter a low-power wait state
  * otherwise. CPUs that have already been started are ignored.
  *
- * @param lcpuid array with the IDs of the logical CPUs that are to be started.
- *   Can be NULL to include all logical CPUs except the one executing the
- *   function.
- * @param num number of logical CPU IDs given
- * @param sp array of stack pointers, one for each specified logical CPU
- * @param entry optional array of entry function pointers, one for each
- *   specified logical CPU. Provided functions must not return. If the
- *   parameter or individual elements are NULL the respective logical CPUs
- *   enter a low-power wait state after startup
+ * @param lcpuidx array with the indices of the logical CPUs that are to be
+ *   started. Can be NULL to include all logical CPUs except the one executing
+ *   the function
+ * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
+ * @param sp array of stack pointers, one for each logical CPU to start. If
+ *   lcpuidx is NULL, must be ukplat_lcpu_count() - 1 stacks
+ * @param entry array of entry function pointers, one for logical CPU to start.
+ *   Can be NULL, otherwise if lcpuidx is NULL, must contain
+ *   ukplat_lcpu_count() - 1 function pointers. Provided functions must not
+ *   return. If the parameter or individual function pointers are NULL the
+ *   respective logical CPUs enter a low-power wait state after startup
  * @param flags (architecture-dependent) flags that specify how to start the
  *   CPUs (see UKPLAT_LCPU_SFLG_* flags if available)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_start(__lcpuid lcpuid[], unsigned int num, void *sp[],
+int ukplat_lcpu_start(__lcpuidx lcpuidx[], unsigned int num, void *sp[],
 		      ukplat_lcpu_entry_t entry[], unsigned long flags);
 
 /**
  * Waits for the specified logical CPUs to enter idle state, or until the
  * timeout expires.
  *
- * @param lcpuid array with the IDs of the logical CPUs to wait for. Can be
- *   NULL to include all logical CPUs except the one executing the function
- * @param num number of logical CPU IDs given
+ * @param lcpuidx array with the indices of the logical CPUs to wait for. Can
+ *   be NULL to include all logical CPUs except the one executing the function
+ * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
  * @param timeout timeout in nanoseconds for the wait to be satisfied. Can be
  *   0 to wait indefinitely
  *
  * @return 0 if the wait for all specified logical CPUs has been satisfied,
  *   an errno-type error value otherwise (e.g., timeout)
  */
-int ukplat_lcpu_wait(__lcpuid lcpuid[], unsigned int num, __nsec timeout);
+int ukplat_lcpu_wait(__lcpuidx lcpuidx[], unsigned int num, __nsec timeout);
 
 /**
  * Executes a function on the specified logical CPUs. The run function does not
  * wait for the execution to start or complete. Multiple functions can be run
  * at the same time without having to wait for their completion.
  *
- * @param lcpuid array with the IDs of the logical CPUs that should execute the
- *   function. Can be NULL to execute the function on all logical CPUs except
- *   the current one
- * @param num number of logical CPU IDs given
+ * @param lcpuidx array with the indices of the logical CPUs that should
+ *   execute the function. Can be NULL to execute the function on all logical
+ *   CPUs except the current one
+ * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
  * @param fn the function to be executed
  * @param flags (architecture-dependent) flags that specify how the function
  *   should be executed (see UKPLAT_LCPU_RFLG_* flags if available)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_run(__lcpuid lcpuid[], unsigned int num,
+int ukplat_lcpu_run(__lcpuidx lcpuidx[], unsigned int num,
 		    struct ukplat_lcpu_func *fn, unsigned long flags);
 
 /**
  * Wakes up the specified logical CPUs from a halt or low-power sleep state.
  *
- * @param lcpuid array with the IDs of the logical CPUs that should be waked up.
- *   Can be NULL to execute the function on all logical CPUs except the current
- *   one.
- * @param num number of logical CPU IDs given
+ * @param lcpuidx array with the indices of the logical CPUs that should be
+ *   woken up. Can be NULL to wakeup all logical CPUs except the current one.
+ * @param num number of logical CPUs specified, 0 if lcpuidx is NULL
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_wakeup(__lcpuid lcpuid[], unsigned int num);
+int ukplat_lcpu_wakeup(__lcpuidx lcpuidx[], unsigned int num);
 
 #else /* CONFIG_HAVE_SMP */
-#define ukplat_lcpu_id()	(0)
+#define ukplat_lcpu_idx()	(0)
 #define ukplat_lcpu_count()	(1)
 #endif /* CONFIG_HAVE_SMP */
 
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -108,23 +108,20 @@ typedef __u64 __lcpuid;		/* Physical ID of logical CPU */
 __lcpuid ukplat_lcpu_id(void);
 
 #ifdef CONFIG_HAVE_SMP
-#include <uk/list.h>
 
 struct ukplat_lcpu_func {
-	struct uk_list_head lentry;
-
 	/**
 	 * Function to execute.
 	 *
-	 * @param regs pointer to a snapshot of the current CPU state. Changes
-	 *   to the state are applied after the RUN IRQ handler returns
-	 * @param fn pointer to this structure. The structure is not touched
-	 *   after function invocation and must be freed if necessary during
-	 *   function execution
+	 * @param regs pointer to a snapshot of the current CPU register state.
+	 *    Changes to the registers are applied after the RUN IRQ handler
+	 *    returns. The parameter might be NULL if the platform does not
+	 *    support supplying a register snapshot.
+	 * @param arg user-supplied argument
 	 */
-	void (*fn)(struct __regs *regs, const struct ukplat_lcpu_func *fn);
+	void (*fn)(struct __regs *regs, void *arg);
 
-	/* Optional user-supplied pointer. */
+	/* Optional user-supplied argument. */
 	void *user;
 };
 
--- a/plat/Config.uk
+++ b/plat/Config.uk
@@ -6,14 +6,28 @@ config UKPLAT_MEMRNAME
 		Enable name field in memory region descriptors
 
 config UKPLAT_LCPU_MAXCOUNT
-        int "Maximum number of supported logical CPUs"
-        range 1 256
-        default 1
+	int "Maximum number of supported logical CPUs"
+	range 1 256
+	default 1
 
 config HAVE_SMP
-        bool
-        default y if UKPLAT_LCPU_MAXCOUNT > 1
-        default n
+	bool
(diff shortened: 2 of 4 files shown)


marcrittinghaus:
I noticed there are a couple of synchronization issues still open. For example you could have:
```C
void __noreturn lcpu_entry_default(struct lcpu *this_lcpu)
{
...
	ukplat_lcpu_enable_irq();
	while (1) {
		/* BUG: If we have an interrupt during halt and then a second
		 * one right before halt() but after setting to IDLE, we do
		 * not return state back to idle, but keep it as BUSY.
		 */
		this_lcpu->state = LCPU_STATE_IDLE;
		// <- Second RUN interrupt here (set to BUSY)
		halt(); // <- First RUN interrupt here (set to BUSY)
	}
}
```

Also I noticed this:

```C
int ukplat_lcpu_wait(__lcpuidx lcpuidx[], unsigned int num, __nsec timeout)
{
...
		/* BUG: If we have two waits, the first with a long timeout
		 * and the second with a short timeout, the thread with
		 * the short timeout will be blocked for the long timeout
		 * while waiting for the lock.
		 */

		ukarch_spin_lock(&lcpu->lock);
...
```
I will spent some more time scratching my head to find all of this f**k ups 🤔

plat/common/lcpu.c
```diff
@@ -30,14 +37,141 @@
...
+	UK_ASSERT(this_lcpu->state == LCPU_STATE_INIT);
+
+	UK_INIT_LIST_HEAD(&this_lcpu->fnlist);
+
+	/* Do architecture-specific initialization */
+	rc = lcpu_arch_init(this_lcpu);
+	if (unlikely(rc))
+		return rc;
+
+	barrier();
```
michpappas: This is a compiler barrier, I think we need a data barrier instead.

plat/common/lcpu.c
```diff
@@ -30,14 +37,141 @@
...
+}
+
+int lcpu_init(struct lcpu *this_lcpu)
+{
+	int rc;
+
+	UK_ASSERT(ukplat_lcpu_irqs_disabled());
+	UK_ASSERT(this_lcpu->state == LCPU_STATE_INIT);
+
+	UK_INIT_LIST_HEAD(&this_lcpu->fnlist);
```
michpappas: If we only run this on secondary cores, then you need to initialize the list of the bootcore somewhere.

plat/common/lcpu.c
```diff
@@ -48,3 +182,336 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+
+	/* Initialize architecture-dependent functionality. This will also do
+	 * CPU discovery and allocation
+	 */
+	rc = lcpu_arch_mp_init(bsp_cpu_id, arg);
+	if (unlikely(rc))
+		return rc;
+
+	/* Register the lcpu_run and lcpu_wakeup interrupt handlers */
+	rc = ukplat_irq_register(LCPU_ARCH_IPI_RUN_IRQ,
```
michpappas: The definitions of the IRQ numbers should probably be platform-defined rather than architecture-defined. Also that gives the architecture a chance to make these compile-time configurable, in case it doesn't want to reserve hard-coded IRQ numbers.

plat/common/lcpu.c
```diff
@@ -48,3 +182,336 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+	int old;
+	int rc, err = 0;
+
+	UK_ASSERT(sp);
+
+	lcpu_lcpuidx_list_foreach(lcpuidx, num, i, lcpu) {
+		if (lcpu->id == this_cpu_id) {
+			/* If the caller did not supply an index array, we
+			 * assume that we do not have parameters and a stack
+			 * for the executing CPU. Otherwise, i.e., the caller
```
michpappas: Did you mean: "Otherwise *if* the caller ..., we still ignore ..."

plat/common/lcpu.c
```diff
@@ -30,14 +37,141 @@
...
+	/* We are coming from BUSY0 state and want to transition to IDLE
+	 * state. However, there can already be functions queued so we have
+	 * to use a decrement here.
+	 */
+	ukarch_dec(&this_lcpu->state);
+
+	/* Enable IRQs. If there are functions queued we will immediately
+	 * jump to the IRQ handler.
+	 */
+	ukplat_lcpu_enable_irq();
```
michpappas: My understanding here is that this assumes that the this is the first time interrupts are enabled on that core. This implies an expectation that until now they were disabled. For Armv8 this is okay. I propose we keep the same assumptions as the ones made by linux: https://www.kernel.org/doc/Documentation/arm64/booting.txt (search for "Before jumping into the kernel, the following conditions must be met"). That way what boot linux can also boot Unikraft. It would be worth though to document it somewhere.

CC: @razvanvirtan

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,353 @@
...
+ * @param bsp_cpu_id id of the bootstrap processor
+ * @param arg an optional parameter from the boot code. Can be NULL
+ *
+ * @return 0 on success, -errno otherwise
+ */
+int lcpu_arch_mp_init(__lcpuid bsp_cpu_id, void *arg);
+
+/**
+ * Start the given logical CPU. The CPU should execute the entry function
+ * with the supplied stack. The CPU will be in INIT state.
```
michpappas, line 364: Maybe "Upon return, the CPU will be in INIT state." would be more clear.

michpappas (changes requested):
Hi @marcrittinghaus just a couple of comments for now in case you're already working on an update. I'll keep reviewing.

plat/common/lcpu.c
marcrittinghaus: The bootcore has to run `lcpu_init()` and `lcpu_mp_init()`.

But I am thinking of a plan for getting rid of `lcpu_init()` (at least for the secondary CPUs) by setting up some per-CPU data structures like the GDT and TSS in `lcpu_arch_mp_init()` when discovering the CPUs. Then we only have to activate these when booting a secondary core in the ASM trampoline, which should work. That makes it possible to simply run `main()` functions with prepared stacks as you suggested.

plat/common/lcpu.c
marcrittinghaus: Agree 👍🏻 Will hopefully drop the function (see below).

But @razvanvirtan, that means you have to set the `LCPU_STATE_BUSY0` state in the ASM trampoline and take care of memory fences there!

plat/common/lcpu.c
marcrittinghaus: Hmm. I could add two configuration options so these would no longer be defined in the source?

plat/common/lcpu.c
marcrittinghaus: Yes, fixed 👍🏻

plat/common/lcpu.c
marcrittinghaus: Yes, for x86 I used the Linux Boot Protocol and the Multiboot Specification as basis for the expected system state:

https://github.com/marcrittinghaus/unikraft/blob/mritting/smp/plat/kvm/x86/multiboot.S
https://github.com/marcrittinghaus/unikraft/blob/mritting/smp/plat/kvm/x86/lcpu_start.S

For maximum compatibility I separated the code that is first executed for a specific boot environment (e.g., multiboot) and the CPU ASM boot trampoline. On x86 I can also use the same trampoline for both the BSP and the AP (see the `lcpu_boot_startup_args` variable in `multiboot.S` and how it is used in `lcpu_start.S:265`.

So for Firecracker, we can just replace `multiboot.S` with a Linux Boot Protocol compatible stub and select a different entry function that parses Linux Boot Parameter in `lcpu_boot_startup_args`.

But that is all specific to the architecture/platform. But I agree, I have to put a documentation piece somewhere that states the expected CPU state after leaving the ASM trampoline.

plat/common/lcpu.c
marcrittinghaus: One question:

If you directly run `main`, you enable interrupts in the ASM trampoline?

plat/common/include/uk/plat/common/lcpu.h
marcrittinghaus: Actually, it will be in `INIT` state when **entering** the function. Perhaps here is a good place to put more information on the expected CPU state prior call and before calling the entry function.

plat/common/lcpu.c
michpappas: I think it's safe to assume that everyone expects interrupts to be already enabled when they enter `main()`.

plat/common/lcpu.c
michpappas: The platform can also `#define` config options to these if you don't want to make them universally configurable. But if you think it's a good idea not to make any assumptions, perhaps they should be sub-items of the SMP section in KConfig.

plat/common/include/uk/plat/common/lcpu.h
michpappas: > The bootcore has to run lcpu_init() and lcpu_mp_init().

I think that proves my point about the need for clarification then 😄

plat/common/lcpu.c
michpappas: > The bootcore has to run lcpu_init() and lcpu_mp_init().

Right, I thought the platform was only expected to call `lcpu_mp_init()` 👍🏼

michpappas: > The bootcore has to run lcpu_init() and lcpu_mp_init().

But I think that won't work as `lcpu_init()` asserts that cpu->state is `LCPU_STATE_INIT`, yet that is only set by `ukplat_lcpu_start()` on secondary cores.

plat/common/lcpu.c
```diff
@@ -48,3 +182,336 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+		 * separate counter to index the arguments.
+		 */
+		lcpu->s_args.entry = (entry && entry[n]) ?
+			entry[n] : lcpu_entry_default;
+		lcpu->s_args.stackp = sp[n];
+
+		n++;
+
+		/* Ensure that the startup arguments have been written back
+		 * before issuing the startup IPI
```
michpappas: Not necessarily IPI 🙂 PSCI is a call to the firmware.

marcrittinghaus: 👍🏻

marcrittinghaus:
The latest push contains the following changes:
* Incorporates the resolved comments
* Add IN/OUT semantic for `num`, but only for `lcpuidx != NULL` to keep a consistent semantic for `num`, i.e., [IN] "Here are num lcpuidx elements", [OUT] "This is how many lcpuidx elements I processed without an error". 
The implementation actually also works with `lcpuidx == NULL` and `num != 0` to report the failed CPU also in these cases, but this functionality is undocumented at the moment so we have asserts to block that combination.
* Limits function queue to just one slot (synchronization all via atomic operations). That keeps one `struct lcpu` to 32 bytes, half a cache line, and avoids all the nasty allocation problems. Should be enough for most basic needs.
* Adds `DONOTBLOCK` flag for `ukplat_lcpu_run` so the function returns also on the first `EAGAIN` error. Otherwise, it retries until success. `EAGAIN` is likely to be returned if two `ukplat_lcpu_run` calls to the same CPU are issued directly in succession due to the single function slot.
* Fixes a couple of bugs

Open issues:
* Add commits that change ISRs to pass the trap frame to the registered interrupt handlers in Unikraft so that remote functions are able to read and arbitrarily modify the current CPU state (for example, to force execution onto a different track).
* Incorporate further remarks

plat/common/lcpu.c
marcrittinghaus: :+1:

plat/common/lcpu.c
marcrittinghaus: Changed the code so that initialization for the boot core is done in `lcpu_init()` and not `lcpu_mp_init()` so it is much earlier in the right state. Also adapted the `UK_ASSERT` check for `INIT` state as you suggested 👍🏻

include/uk/plat/lcpu.h
```diff
@@ -96,100 +96,129 @@ void ukplat_lcpu_halt_to(__nsec until);
  */
 void ukplat_lcpu_halt_irq(void);
 
-#ifdef CONFIG_HAVE_SMP
-#include <uk/list.h>
+/* Signature for a CPU entry function */
+typedef void (*ukplat_lcpu_entry_t)() __noreturn;
```
michpappas: Shouldn't that have the same prototype as `lcpu_entry_default`, ie:
```
typedef void (*ukplat_lcpu_entry_t)(struct lcpu *this_lcpu) __noreturn;
```
which also implies we need to move `struct lcpu` to the public header.

include/uk/plat/lcpu.h
```diff
@@ -96,100 +96,129 @@ void ukplat_lcpu_halt_to(__nsec until);
...
+struct ukplat_lcpu_func {
 	/**
 	 * Function to execute.
 	 *
 	 * @param regs pointer to a snapshot of the current CPU state. Changes
-	 *   to the state are applied after the function returns
-	 * @param fn pointer to this structure. The structure is not touched
-	 *   after function invocation and can be freed if necessary during
-	 *   function execution
+	 *   to the state are applied after the RUN IRQ handler returns
```
michpappas: If I understand correctly the term "state" here refers to two different things:
1. "...to a snapshot of the current  CPU state." refers to registers
2. "...to the state applied..." refers to `lcpu->state`

This can be confusing, perhaps it would be better to use that term for a single context or otherwise make it less ambiguous.

include/uk/plat/lcpu.h
```diff
@@ -96,100 +96,129 @@ void ukplat_lcpu_halt_to(__nsec until);
...
+ *   return. If the parameter or individual function pointers are NULL the
+ *   respective logical CPUs enter a low-power wait state after startup
+ * @param flags (architecture-dependent) flags that specify how to start the
+ *   CPUs (see UKPLAT_LCPU_SFLG_* flags if available)
  *
  * @return 0 on success, an errno-type error value otherwise
  */
-int ukplat_lcpu_start(__lcpuid lcpuid[], void *sp[],
-		      ukplat_lcpu_entry_t entry[], unsigned int num);
+int ukplat_lcpu_start(const __lcpuidx lcpuidx[], unsigned int *num, void *sp[],
```
michpappas, line 166: Aren't we moving this to the private header after all?

include/uk/plat/lcpu.h
```diff
@@ -96,100 +96,129 @@ void ukplat_lcpu_halt_to(__nsec until);
...
  *
  * @return 0 if the wait for all specified logical CPUs has been satisfied,
  *   an errno-type error value otherwise (e.g., timeout)
  */
-int ukplat_lcpu_wait(__lcpuid lcpuid[], unsigned int num, __nsec timeout);
+int ukplat_lcpu_wait(const __lcpuidx lcpuidx[], unsigned int *num,
+		     __nsec timeout);
 
 /**
  * Executes a function on the specified logical CPUs. The run function does not
```
michpappas, line 189: This description about the non-blocking behavior is now only partially true.

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+UK_CTASSERT(__offsetof(struct lcpu_sargs, entry)  == LCPU_SARGS_ENTRY_OFFSET);
+UK_CTASSERT(__offsetof(struct lcpu_sargs, stackp) == LCPU_SARGS_STACKP_OFFSET);
+
+UK_CTASSERT(sizeof(struct lcpu_sargs) == LCPU_SARGS_SIZE);
+#endif /* !__ASSEMBLY__ */
+
+/*
+ * Logical CPU (LCPU) Structure
+ */
+#define LCPU_STATE_OFFSET		0x00
```
michpappas, line 82: Why don't we define these directly to `offsetof` and instead we need to maintain hardcoded values and compile-time assertions? Is GCC confused by the union or something?

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+struct lcpu {
+	/* Current CPU state (LCPU_STATE_*).
+	 * Working on it with atomic instructions - must be 8-byte aligned
+	 */
+	volatile int state __align(8);
+
+	__lcpuidx idx;
+	__lcpuid id;
+
+	union {
```
michpappas, line 108: Is there any chance we need to refer back to this information that we now we override using that union?
Also from what I see we don't save that much memory:
* s_args: 16 bytes
* fn: 8 bytes
* error_code: 4 bytes

So 16 vs 28 bytes for a struct that we pad to 64 bytes (cache line size) anyway 🙂

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+		 * Only valid in LCPU_STATE_HALTED
+		 */
+		int error_code;
+	};
+
+	/* Architectural-dependent part */
+	struct lcpu_arch arch;
+
+	/* We want every CPU on a separate cache line */
+	__u8 padding[LCPU_SIZE - LCPU_MEMBERS_SIZE];
```
michpappas: I think you can also declare the structure with
```
__attribute__(aligned(CACHE_LINE_SIZE))
```

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+		struct ukplat_lcpu_func fn;
+#endif /* CONFIG_HAVE_SMP */
+
+		/* Error code indicating the halt reason
+		 * Only valid in LCPU_STATE_HALTED
+		 */
+		int error_code;
+	};
+
+	/* Architectural-dependent part */
```
michpappas: s/Architectural/Architecture/

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+	struct lcpu_arch arch;
+
+	/* We want every CPU on a separate cache line */
+	__u8 padding[LCPU_SIZE - LCPU_MEMBERS_SIZE];
+};
+
+UK_CTASSERT(sizeof(struct ukplat_lcpu_func)        == 0x10);
```
michpappas: Maybe replace 0x10 with a `#define` to be consistent with the check for `lcpu_sargs` at line 73

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -0,0 +1,367 @@
...
+UK_CTASSERT(sizeof(struct lcpu_sargs)              == 0x10);
```
michpappas: This one is redundant, we make the same check in line 73

plat/common/include/uk/plat/common/lcpu.h
michpappas: This one is still pending.

plat/common/lcpu.c
```diff
@@ -30,14 +37,124 @@
...
+
+	return &lcpus[idx];
+}
+
+struct lcpu *lcpu_get_current(void)
+{
+	return lcpu_get(ukplat_lcpu_idx());
+}
+
+int lcpu_init(struct lcpu *this_lcpu)
```
michpappas, line 119: So if we want the bootcore to call `lcpu_init(&lcpus[0])` then we need a declaration of `lcpus` to `lcpu.h` under `plat/common/include/`.

plat/common/lcpu.c
```diff
@@ -30,14 +37,124 @@
...
+
+	/* Initialize the bootstrap CPU */
+	if (this_lcpu->idx == 0) {
+		UK_ASSERT(lcpu_count == 1);
+		UK_ASSERT(this_lcpu == lcpu_get(0));
+
+		this_lcpu->id    = lcpu_arch_id();
+		this_lcpu->state = LCPU_STATE_INIT;
+	}
+
```
michpappas, line 141: Is the `UK_ASSERT(this_lcpu->state == LCPU_STATE_INIT);` here gone intentionally?

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+	wmb();
+
+	return 0;
+}
+
+static void lcpu_fn_dequeue(struct lcpu *this_lcpu, struct ukplat_lcpu_func *fn)
+{
+	*fn = this_lcpu->fn;
+
+	/* Ensure that we have captured the whole function object */
```
michpappas, line 269: I understand the purpose of this barrier, but the "whole function object" statement confused me, as it implies... an object rather than a pointer. Or I think I understand but I'm actually missing something? 🤔

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+
+	/* Initialize architecture-dependent functionality. This will also do
+	 * CPU discovery and allocation
+	 */
+	rc = lcpu_arch_mp_init(arg);
+	if (unlikely(rc))
+		return rc;
+
+	/* Register the lcpu_run and lcpu_wakeup interrupt handlers */
+	rc = ukplat_irq_register(LCPU_ARCH_IPI_RUN_IRQ,
```
michpappas: So we still call use the term `ARCH` in the name of the IRQ definitions? I really think they should not be architecture-defined, but platform defined. Perhaps change the naming to `PLAT` and add a check in the platform `lcpu.h`as:
```
#ifdef CONFIG_HAS_SMP
#if !defined(LCPU_PLAT_IPI_RUN_IRQ) || !defined(LCPU_PLAT_IRI_WAKEUP_IRQ)
#error LCPU_PLAT_IPI_RUN_IRQ and LCPU_PLAT_IPI_WAKEUP_IRQ are not defined by the platform
#endif
#endif /* CONFIG_HAS_SMP */
```
The platform still has the choice to hardcode or provide KConfig options.

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+	if (unlikely(rc)) {
+		uk_pr_crit("Could not register handler for wakeup IRQ %d\n",
+			   LCPU_ARCH_IPI_WAKEUP_IRQ);
+		return rc;
+	}
+
+	return 0;
+}
+
+void __weak __noreturn lcpu_entry_default(struct lcpu *this_lcpu)
```
michpappas, line 354: So we still define this as `weak` even though that is not required for solving the custom entry with unmodified stack problem?

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+{
+	int rc;
+
+	/* Finish initialization. As there is nothing to return to, we
+	 * just enter halted state if an error occurs.
+	 */
+	rc = lcpu_init(this_lcpu);
+	if (unlikely(rc))
+		lcpu_halt(this_lcpu, rc);
```
michpappas, line 366: Do we expect the caller to retry bringing up the core, or halt is halt here? If we need to be able to wake up perhaps we need that be `lcpu_halt_to_irq()`?

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+
+	/* Enable IRQs. If there are functions queued we will immediately
+	 * jump to the IRQ handler.
+	 */
+	ukplat_lcpu_enable_irq();
+	while (1) {
+		/* Besides interrupts in general, the halt can be
+		 * interrupted by calls to ukplat_lcpu_run().
+		 */
+		halt();
```
michpappas: Same here. Should be `lcpu_halt_to_irq()`, and we should fix the arm implementation of `halt()` to be uninterruptible (currently uses `wfi` which is effectively `lcpu_halt_to_irq()`.

plat/common/lcpu.c
```diff
@@ -48,3 +165,424 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+		 */
+		wmb();
+
+		rc = lcpu_arch_start(lcpu, flags);
+		if (unlikely(rc)) {
+			lcpu->state = LCPU_STATE_HALTED;
+			lcpu->error_code = rc;
+
+			/* There is a more serious problem. Stop here. The
+			 * caller can skip the CPU and retry if the likes to
```
michpappas: Something's wrong with that sentence...

michpappas (changes requested):
Hi @marcrittinghaus I just gave it another look. I think it looks good, all comments are minor.

include/uk/plat/lcpu.h
marcrittinghaus: No, this is on purpose. I want to give the platform the possibility to provide arbitrary parameters to the entry function also depending on if this is the boot BSP or not. So I made this a non-prototyped function. For example, on x86 I provide for the BSP the `lcpu` struct and also the boot parameters (e.g., from multiboot or Linux Boot Protocol). We only have the requirement that the first parameter must be an `lcpu` struct for the secondary CPUs as they would end up in the `default` function.

I still have to find a place where to document the required CPU/system state after leaving the ASM trampoline.

But anyways, that also avoids making `lcpu` public at the moment.

marcrittinghaus: Actually, it is probably better to do the following:

Change the entry function to a function object that allows a user argument.
```C
struct ukplat_lcpu_entry_func {
     void (*fn)(void *arg);
     void *arg;
};
```
Then I can use this argument for the boot params on the BSP and you can use it to prepare (or supply) an untouched stack (see below). And the entry function can always use `lcpu_get` in the platform to get the `lcpu` struct.

include/uk/plat/lcpu.h
marcrittinghaus: Actually, I decided against it, because I think it makes sense to give full control of what to do with the CPUs. As we talked on Discord, this would imply that we have to also make `lcpu_init` public. But you can avoid all of this with your suggestion to always start in the `default` entry for secondary CPUs, do the `lcpu_init` there and then jump out to the supplied entry function if there is one.

If you need a specifically prepared stack you can still prepare it in the supplied entry function and then do the actual jump to the target function. So the user would do:
```C
int main(int argc, char* argv[]) {
   Run me on a diffierent core
}

void __noreturn myentry(void *arg) {
   PREPARE stack with info from arg
   jump to main
}

void *stackp = malloc(STACK_SIZE);
ukplat_lcpu_entry_func fn = {
  .fn = myentry;
 .arg = Some info to prepare stack, or actually pointer to prepared stack
} 
ukplat_lcpu_start(othercpu, 1, stackp, fn, 0);
```
But in fact the call chain is like this:
`ASM trampoline` -> `entry_default` -> `myentry`

plat/common/include/uk/plat/common/lcpu.h
marcrittinghaus: We need the values in the ASM trampoline as offsets (at least the stackp). Unfortunately, that does not work with `offsetof` and `sizeof` ☹️ But I think the static asserts are a good addition, which we should also consider for the offsets defined for  `struct regs`.

michpappas: Yes if we cannot use `offsetof` in the definitions, then the compile-time asserts are definitely needed 👍🏻

plat/common/include/uk/plat/common/lcpu.h
marcrittinghaus: I don't think that we need to refer to the information anymore. In fact, this makes the total size of `lcpu` 32-bytes which is the cache line size on ARM (32) 😏 Unfortunately, making the proposed change to the entry function (see above) would breaks this.

But anyways, there is still the architecture part that might need additional space so I think saving memory does make sense. And it is easier to later relax it and use regular members than the other way round.

plat/common/include/uk/plat/common/lcpu.h
marcrittinghaus: Yes, right. Good point!

plat/common/lcpu.c
marcrittinghaus: See above. I think you suggestion to always call `default` and only then jump to the entry function is the best solution.

plat/common/lcpu.c
marcrittinghaus: Yes, because you were right that the state does not have to be `INIT` for all CPUs. The BSP is not initialized by `start` so the `lcpu->state` is `OFFLINE` when entering the function. I just had in x86 a line in the ASM trampoline which set it to `INIT` so it was ok. But I am still working on this and might move things for the BSP around a bit and reintroduce the check if it valid.

plat/common/lcpu.c
marcrittinghaus: This is not only a single pointer, but a struct consisting of a function pointer and a user argument. Thus the term function object. Do you think this is still confusing then?

plat/common/lcpu.c
marcrittinghaus: Good point and we should change it. But it would not work like you suggest. At the moment the common code is in the `common` directory and imports the _common_ architecture-specific header. A certain platform is not mentioned anywhere. Instead, the platform imports the common API and defines the symbols. So what we can do is define two variables `lcpu_plat_ipi_run/wakeup_irq` that are then referenced here. The platform could initialize them using a config option or even use some boot code to dynamically find the right IRQ numbers before the call to `lcpu_arch_mp_init`.

plat/common/lcpu.c
marcrittinghaus: Right, it is not strictly needed. Not even when we now adopt your idea that we always call `default` and then jump to entry only there, so that we have the `lcpu_init` call hidden.

But I like it anyways because this gives you the power to replace the `default` entry function and do something completely different there in the platform if needed. Or for example add intermediate steps before again calling this `default` method.

plat/common/lcpu.c
marcrittinghaus: Actually, at the moment we do not have a state transition back from the `HALTED` state to the `OFFLINE` state. I thought this will probably be also very architecture-specific how to reset a CPU core. We might want to add an extra API call like `ukplat_lcpu_reset` for example in the future. However, we are also missing a public API to recognize that a CPU "crashed" and get the error code. It may also make sense to signal an uk_event or something like that if a CPU crashes.

But this whole topic of error recovery, we do not have thought through enough. For the moment, I think just doing a halt and storing the `rc` is thus enough.

plat/common/lcpu.c
marcrittinghaus: I think this is correct. `halt` should just put the CPU into a low-power state where it waits for interrupts. If you want an uninterruptable halt to have to call `ukplat_lcpu_halt`, which disables IRQs first (although it is buggy in that it should also have a loop around `halt` to also remain in this state if there is a non-maskable interrupt. Only a CPU reset should be able to recover the CPU). But that is fixed with this PR.

include/uk/plat/lcpu.h
michpappas: > Actually, I decided against it, because I think it makes sense to give full control of what to do with the CPUs.

Okay for this PR, but eventually we should really provide a clear definition of that. To me it sounds paradoxical to start the cores from the application and at the same time have no access to drivers or other low-level facilities. Or we stick to the current scheme for consistency, and expect that any application that needs to start the cores has to manually add the internal header into its `CINCLUDES-y` into its Makefile.uk. This is what we do for other private interfaces.

plat/common/include/uk/plat/common/lcpu.h
michpappas: Ah I forgot about 32-bit Arm, and it is actually a use-case. Thanks 👍🏼

marcrittinghaus:
Hey @michpappas,

thank you very much for all of this feedback! 😃 I will work some more on the PR today. Unfortunately, yesterday I did not find any desk time.

plat/common/lcpu.c
michpappas: No it was just me not being aware of the copy struct by pointer trick. In fact the comment is useful. Live and learn 😊

include/uk/plat/lcpu.h
marcrittinghaus: Ok, so you would still make this private? I think this whole flexibility to defined per-cpu custom entry functions looses a bit of power if you keep it in the platform. May @skuenzer could add a third opinion on this?

plat/common/lcpu.c
michpappas: Okay, but then what is the purpose of having an `lcpu_halt_to_irq()` function?

plat/common/lcpu.c
michpappas: Works for me 👍🏼

include/uk/plat/lcpu.h
michpappas: > Ok, so you would still make this private?

I wouldn't block the PR over this one, it's not such a big deal 🙂 My point is more towards agreeing on the necessity to resolve this and other core architectural topics before too much debt is accumulated.

marcrittinghaus: I had a short discussion with Simon about this. He would also prefer this to be public. But based on this discussion I will also keep the entry function typedef as we have it at the moment.

But agree, we have to discuss this more as part of our platform restructure so that we have better guidelines for what APIs to have at what level.

plat/common/lcpu.c
cristian-vijelie: Those interrupts aren't really dependent on the platform. They should be the same for every platform of the same architecture, in my opinion, so it would make sense to be defined at architecture level

marcrittinghaus: Yes, right. I also think that we will probably have the same IRQ vectors for all platforms of a certain architecture. But it is fair to keep it flexible. To not introduce further symbols, I simply made it parameters to `lcpu_mp_init`. Then you can compute the IRQ values at runtime or just pass `CONFIG_` defines.

plat/common/lcpu.c
marcrittinghaus: 👍🏻

plat/common/lcpu.c
marcrittinghaus: The `ukplat_lcpu_halt_irq` is the public interface and `halt` is just the function that implements the actual halt (e.g., `hlt` or `wfi`) like `cpuid` or `readmsr` just call the underlying ASM instructions. But you are right, that it is somewhat confusing and we should call `halt` something like `lcpu_halt_irq`. We can do this as a separate PR as it will touch all platforms. But it makes sense. As `halt` is implemented along `reset` and `system_off`, it is clearly an artifact from designing for uniprocessor systems. It has a pretty global touch to it.

plat/common/lcpu.c
cristian-vijelie: I guess he wanted to say "if he likes to", but I would replace it with "if it wants to"

marcrittinghaus: Right, the sentence got messed up. Actually, at the moment you cannot even retry because the CPU is put into `HALTED` state. As we discussed, we have to think more about how to handle errors at this level. Perhaps have a reset function or so. For now, we at least inform about the error.

plat/common/lcpu.c
michpappas: > Those interrupts aren't really dependent on the platform. They should be the same for every platform of the same architecture, in my opinion, so it would make sense to be defined at architecture level

@cristian-vijelie you cannot guarantee that a given interrupt will not be reserved by some device or the HV on a given platform, so they really need to be configurable. Moreover, the architecture does not impose such restrictions, it only provides you the means, so the choice of IRQ is one level above the architecture.

plat/common/lcpu.c
michpappas: ~~Then maybe in that case it would be better if we were using the `ukplat_lcpu_` interface consistently in this scope, rather than using in some parts the higher level interface and in some others the low level implementation?~~

Actually never mind about that one.

> We can do this as a separate PR as it will touch all platforms

Okay 👍🏼

include/uk/plat/lcpu.h
marcrittinghaus: I am not quite sure what you mean with the "non-blocking behavior" documented wrong here. Please have a look at the new version if this still persists.

marcrittinghaus:
@michpappas I pushed a new version. I am now working on the commits that still need to be added for making the trap frame (`struct regs`) available in the interrupt handler as expected by the API. But otherwise, I think we are getting close to a ready version.

plat/common/lcpu.c
```diff
@@ -30,14 +37,127 @@
...
+	UK_ASSERT(ukplat_lcpu_irqs_disabled());
+
+	/* Initialize the bootstrap CPU */
+	if (lcpu_is_bsp(this_lcpu)) {
+		UK_ASSERT(lcpu_count == 1);
+		UK_ASSERT(this_lcpu->idx == 0);
+
+		this_lcpu->id    = lcpu_arch_id();
+		this_lcpu->state = LCPU_STATE_INIT;
+	} else {
```
razvanvirtan, line 137: @marcrittinghaus I hope I don't repeat a previous discussion here. Shouldn't we initialize the lock associated to this cpu here? (I think we need a lock in the `lcpu` structure regardless of the arch)

marcrittinghaus: No, in the common API I have solved everything without a lock using atomic instructions. So the common part does not need a lock. If you need one for ARM you have to put it in the `lcpu_arch` structure and initialize it in `lcpu_arch_init`.

But actually, I don't think you need a lock on ARM. Please elaborate, where you think you need one.

plat/common/lcpu.c
razvanvirtan: Actually, I think everything in the current version is solved by the common code, with the APs moving from INIT to BUSY0 at the end of `lcpu_init()`.

marcrittinghaus: Yes, this is a stale comment.

plat/Config.uk
```diff
@@ -6,14 +6,38 @@ config UKPLAT_MEMRNAME
...
+	help
+		The hardware assigns each CPU an ID, which typically encodes
+		the CPUs position in the processor topology. Many VMMs, however,
+		just use sequential numbers so internally mapping IDs to
+		indices in Unikraft is not necessary.
+
+config UKPLAT_LCPU_RUN_IRQ
+	int "Remote Function IRQ#"
+	default 13 if ARCH_X86_64
+	depends on ARCH_X86_64
```
razvanvirtan: I think we should remove the `depends on ARCH_X86_64` here.

plat/Config.uk
```diff
@@ -6,14 +6,38 @@ config UKPLAT_MEMRNAME
...
+
+config UKPLAT_LCPU_WAKEUP_IRQ
+	int "Wakeup IRQ#"
+	default 14 if ARCH_X86_64
+	depends on ARCH_X86_64
```
razvanvirtan: I think we should remove the `depends on ARCH_X86_64` here.

plat/common/lcpu.c
```diff
@@ -30,14 +37,150 @@
...
+	 */
+
+	/* Initialize the bootstrap CPU */
+	if (lcpu_is_bsp(this_lcpu)) {
+		if (unlikely(lcpu_count > 1))
+			return -EPERM;
+
+		this_lcpu->idx   = 0;
+		this_lcpu->id    = lcpu_arch_id();
+		this_lcpu->state = LCPU_STATE_INIT;
```
michpappas, line 136: This keeps the bootcore to `INIT` forever, perhaps consider to set instead to `IDLE`?

plat/common/lcpu.c
```diff
@@ -48,3 +191,447 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+
+int ukplat_lcpu_start(const __lcpuidx lcpuidx[], unsigned int *num, void *sp[],
+		      const ukplat_lcpu_entry_t entry[], unsigned long flags)
+{
+	__lcpuid this_cpu_id = ukplat_lcpu_id();
+	struct lcpu *lcpu;
+	unsigned int i, n, argi = 0;
+	const int new = LCPU_STATE_INIT;
+	int old;
+	int rc = 0, rc2 = 0;
```
michpappas: This causes a `warning: unused variable`. Perhaps make conditional to `LCPU_ARCH_MULTI_PHASE_STARTUP`?

plat/common/lcpu.c
```diff
@@ -48,3 +191,447 @@ void ukplat_lcpu_halt_to(__nsec until)
...
+	return 1;
+}
+
+int ukplat_lcpu_run(const __lcpuidx lcpuidx[], unsigned int *num,
+		    const struct ukplat_lcpu_func *fn, unsigned long flags)
+{
+	__lcpuid this_cpu_id = ukplat_lcpu_id();
+	struct lcpu *lcpu;
+	unsigned int n, i;
+	int rc, rc2 __maybe_unused;
```
michpappas: `rc2` is not used in this function and it wasn't caught because of `__maybe_unused`.

michpappas (changes requested):
@marcrittinghaus I left a few final notes for a couple of issues I spotted while testing the arm64 PR. Also I see some commits in this series that update other commits in the same series. These should probably be squashed.

plat/common/lcpu.c
marcrittinghaus: I cannot reproduce this problem. After breaking in one of the APs remote functions and inspecting system state, I get:
```
lcpus[] = {
    state = 0x3, /* BUSY0 */
    idx = 0x0,
    id = 0x0,
    fn = { fn = 0x0, user = 0x0 }
  }, {
    state = 0x4, /* BUSY1 */
    idx = 0x1,
    id = 0x1,
    fn = {fn = 0x401138f0, user = 0x0}
  }, {
    state = 0x4, /* BUSY1 */
    idx = 0x2,
    id = 0x2,
    fn = { fn = 0x401138f0, user = 0x0 }
  }, {
    state = 0x0, /* OFFLINE */
    idx = 0x3,
    id = 0x3,
    s_args = { entry = 0x0, stackp = 0x0
    }
}
```
Also the code is:
```C
	if (lcpu_is_bsp(this_lcpu)) {
		if (unlikely(lcpu_count > 1))
			return -EPERM;

		this_lcpu->idx   = 0;
		this_lcpu->id    = lcpu_arch_id();
		this_lcpu->state = LCPU_STATE_INIT;  <--- set BSP to init
	} else { ... }

	rc = lcpu_arch_init(this_lcpu);              <--- arch init of BSP

        ...

	this_lcpu->state = LCPU_STATE_BUSY0;         <--- sets BSP to BUSY0

	return rc;
```

marcrittinghaus:
> @marcrittinghaus I left a few final notes for a couple of issues I spotted while testing the arm64 PR. Also I see some commits in this series that update other commits in the same series. These should probably be squashed.

Hi @michpappas,
I have fixed the issues you mentioned apart from the `BUSY` state problem, that I cannot reproduce. Maybe you find some time to elaborate on this?

I left the separate commits intact. I agree that some of them could be technically squashed but we do various changes to an API that we already published (although without functional implementation). I thus thought that it is useful to keep the explanation for all the various changes and kind of transition from current to new step-by-step.

michpappas:
Hi @marcrittinghaus ,

> I have fixed the issues you mentioned apart from the `BUSY` state problem, that I cannot reproduce. Maybe you find some time to elaborate on this?

You mean that the bootcore actually leaves the `INIT` state?

> I left the separate commits intact. I agree that some of them could be technically squashed but we do various changes to an API that we already published (although without functional implementation). I thus thought that it is useful to keep the explanation for all the various changes and kind of transition from current to new step-by-step.

Alright, make sense 👍🏼

marcrittinghaus:
> Hi @marcrittinghaus ,
> 
> > I have fixed the issues you mentioned apart from the `BUSY` state problem, that I cannot reproduce. Maybe you find some time to elaborate on this?
> 
> You mean that the bootcore actually leaves the `INIT` state?

Yes, take a look at the additional info I supplied above.

michpappas:
> > Hi @marcrittinghaus ,
> > > I have fixed the issues you mentioned apart from the `BUSY` state problem, that I cannot reproduce. Maybe you find some time to elaborate on this?
> > 
> > 
> > You mean that the bootcore actually leaves the `INIT` state?
> 
> Yes, take a look at the additional info I supplied above.

Ah, I missed that comment. I think you're right, yesterday early evening I was looking `ukplat_lcpu_wait()` and why the core is set to `BUSY0` and not `IDLE`, which we later discussed and concluded that it is okay. Then when I posted that comment it was already past midnight and somehow brain transitioned into "bootcore never leaves `IDLE`", sorry about the confusion, I'm just tired 😓

michpappas (approved):
@marcrittinghaus once again thank you very much for all this. I'll wait for one more RB from @cristian-vijelie or @razvanvirtan's before I merge.

razvanvirtan (approved):
I have tested it again with the ARM64 implementation, all good on my side!

michpappas:

PR #474, opened 2022-06-11
Add Intel SGX support to Unikraft

### LINKED ISSUE #33: Unikraft on TEEs (TrustZone, SGX, etc.)

xymeng16:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

 - `CONFIG_ARCH_X86_64_SGX=y`
 - `CONFIG_KVM_RING3=1`

### Description of changes

This PR is for adding Intel SGX support to Unikraft. It also introduces some tiny modifications on the GDT/TSS part to provide experimental ring-3 support, because SGX ENCLU instruction must be executed under ring-3.

### Introduction

This library `uksgx` provides the SGX driver, SDK and platform software (PSW, aka runtime), to provide a basic SGX functionality to Unikraft, which allows an enclave in the CPU to be executed securely at native speed. The enclave memory is not encrypted and can only be accessed by the enclave itself, guaranteed by the MMU inside the CPU package. Enclave runs at ring-3 only, hence no system calls are allowed. The enclave is linked statically and is self-contained with the pre-compiled SDK libraries.

A device `/dev/sgx` is registered to the devfs with the following IOCTL commands (for compatibility consideration):
- SGX_IOC_ENCLAVE_CREATE: validate user parameters and invoke ECREATE instruction
- SGX_IOC_ENCLAVE_ADD_PAGE: invoke EADD to add REG/TCS pages into the enclave and invoke EEXTEND to generate cryptographic hashes of the contents in the enclave
- SGX_IOC_ENCLAVE_INIT: verify enclave's content against signed SIGSTRUCT and initialize the enclave, then mark it ready to be used via EINIT instruction

After creation/loading/initialization, app/enclave can invoke EENTER/EEXIT/ERESUME to control the in/out of the enclave. Those switches are handled by `sgx_ocall()` and `sgx_ecall()`, which is wrapped in the SDK.

### Timeline
It is only a drafted PR and lots of work is required to be done. At the final time, I will squash the commits and do some cleanup before submitting the formal PR.
- 11 Jun 2022: At the time of the first draft, the probing of SGX is finished, that is to say, Unikraft can be aware of SGX if enabled in `make menuconfig`. And as inspired by [linux-sgx-driver](https://github.com/intel/linux-sgx-driver), the uksgx library registers a device /dev/sgx to the devfs, and some stub ioctl function is added to the library, in order to provide a unified and easy interaction channel between uksgx and the applications.
- 23 Jun 2022: As of today, the experimental ring-3 support dedicated to the SGX driver is finished when paging is enabled. Both paging and ring-3 are the fundamental features for SGX support.
- 7 Jul 2022: The EPC physical area is successfully mapped to the virtual address area given a hardcoded `SGX_VA_OFFSET` as its beginning address. This should be refined when the function for finding a usable VA area is implemented.
- 25 Jul 2022: `sgx_encl_create` is finished. Seems VMA is not available to Unikraft. So we need to interact with the page table using raw APIs for page management.
- 16 Aug 2022: #520 is online. This PR introduces a radix-tree implementation, which is required by SGX driver to manage the EPC pages.
- 8 Sep 2022: As of commit `ea1e61e`, the SGX driver is ported, except for implementations of a. VMA management; b. interruptible sleep and c. synchronized EPC page adding. Those, especially for the VMA management required a thorough review and understanding of the Unikraft's memory implementation, which requires lots of time. Hence I am going to work on the SGX SDK first.
- 14 Sep 2022: 
  1.  SDK is ported to Unikraft. Once `uksgx` is enabled, the SGX SDK will be compiled to some static libraries (`libsgx_tstdc.a`, `libsgx_tcxx.a` and `libsgx_tservice.a`) by default.
  2. Update this PR to add a basic introduction of Intel SGX implementation in Unikraft


lib/uksgx/include/uk/sgx_cpu.h
```diff
@@ -0,0 +1,283 @@
...
+#define __cpuid(x, y)                                                    \
+	asm volatile("cpuid"                                                   \
+		     : "=a"(x[0]), "=b"(x[1]), "=c"(x[2]), "=d"(x[3])          \
+		     : "a"(y))
+
+#define Genu 0x756e6547
+#define ineI 0x49656e69
+#define ntel 0x6c65746e
+
+static inline void rdmsr(unsigned int msr, __u32 *lo, __u32 *hi)
```
cristian-vijelie: You can use the functions already defined in `plat/common/include/x86/cpu.h`

Same for `rdmsrl`, `wrmsr`, `wrmsrl`

lib/uksgx/include/uk/sgx_cpu.h
```diff
@@ -0,0 +1,283 @@
...
+#define MSR_IA32_SGXLEPUBKEYHASH3	0x0000008F
+
+/* __cpuid_count(unsinged int info[4], unsigned int leaf, unsigned int subleaf); */
+#define __cpuid_count(x, y, z)                                                       \
+	asm volatile("cpuid"                                                   \
+		     : "=a"(x[0]), "=b"(x[1]), "=c"(x[2]), "=d"(x[3])          \
+		     : "a"(y), "c"(z))
+
+/* __cpuid(unsinged int info[4], unsigned int leaf); */
+#define __cpuid(x, y)                                                    \
```
cristian-vijelie: You can use the `cpuid` function from `plat/common/include/x86/cpu.h`

lib/uksgx/include/uk/sgx_cpu.h
```diff
@@ -0,0 +1,283 @@
...
+/*
+ * Volatile isn't enough to prevent the compiler from reordering the
+ * read/write functions for the control registers and messing everything up.
+ * A memory clobber would solve the problem, but would prevent reordering of
+ * all loads stores around it, which can hurt performance. Solution is to
+ * use a variable and mimic reads and writes to it to enforce serialization
+ */
+extern unsigned long __force_order;
+
+static inline unsigned long native_read_cr0(void)
```
cristian-vijelie: Maybe it would be a good idea to define the functions related to the x86 architecture in `plat/common/include/x86/cpu.h`, so we don't have duplicates

xymeng16: Yep, I defined them in my own lib because at first I am not familiar with other parts of Unikraft hence do not want to modify it. I will perform a simplification and clarification once we reach a great milestone of this project, say make a "hello world" work from SGX.

xymeng16:
#520 introduce the radix-tree implementation to Unikraft. This is required by SGX driver to manage the enclave pages. Once it passes the tests, I will merge it into the SGX driver branch.

hlef:
Thank you so much for this PR @xymeng16!

A few high-level comments and points that I (we) came to realize since we started this project:
- If we take a step back, what is our threat model here? Using SGX within Unikraft, we essentially do not trust (some of the) "kernel" part of Unikraft. I don't know how much sense this makes; this part is already very thin! Clearly, it does bring a little something, but is it worth the performance, compatibility, and effort?
- More specifically, SGX is probably not the best target mechanism for Unikraft / unikernels in general. It forces you to switch to ring 3, a very inelegant and costly design quirk. Compared to other mechanisms that we can use for compartmentalization within Unikraft (MPK, etc.), it doesn't feel like a good match.
- Finally, SGX is a dying technology. It has been deprecated on laptop CPUs and it feels a bit like its existence on server processors is only a matter of time ([interesting discussion](https://github.com/intel/linux-sgx/issues/760)). Beyond SGX, the industry is moving away from the SGX model for a confidential VM model (AMD SEV, Intel TDX) which is much more suited for Unikraft and unikernels in general.

So, nothing against your work @xymeng16, it's great, we realized a lot of things! :+1: 

That being said, I do not think that we should ultimately merge this PR.

Comments?

dimakuv:
Hello, a Gramine maintainer here. I accidentally stumbled upon this PR.

> It forces you to switch to ring 3, a very inelegant and costly design quirk. Compared to other mechanisms that we can use for compartmentalization within Unikraft (MPK, etc.), it doesn't feel like a good match.

Intel SGX technology has nothing to do with compartmentalization. It's about confidential computing (so that privileged software and hardware adversaries cannot read/modify your application's memory). Comparing SGX vs MPK is wrong.

> Finally, SGX is a dying technology. [...] it feels a bit like its existence on server processors is only a matter of time.

That's incorrect.

PR #481, opened 2022-06-20
lib/posix-procfs: Add support for the proc filesystem

mandrei12:
Add support for the /proc/meminfo file
Add support for the /proc/version file
Add support for the /proc/cmdline file
Add support for the /proc/filesystems file
Add support for the /proc/mounts file
Add static getters for the above mentioned files
Add configuration for the library

Signed-off-by: Andrei Mutu <andrei.mutu@stud.acs.upb.ro>

### Prerequisite checklist

### Base target


lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
+posix-procfs: implementation of the proc fs
+
+Currently the library offers support for 5 files:
+/proc/meminfo
+/proc/version
+/prov/cmdline
+/proc/filesystems
+/proc/mounts
```
nderjung: This might be better as a table, e.g.:

| Path | Description |
|-|-|
| `/proc/meminfo` | Returns the current memory usage. |

Perhaps (since proc is not really standardized (and not actually posix...), even the table could suggest which other kernels it supports (e.g. whether the path is linux-specific, *nix, *bsd etc.)

lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
...
+
+Prerequisites:
```
nderjung:
```diff
- Prerequisites:
+ ## Prerequisites
```

lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
+posix-procfs: implementation of the proc fs
```
nderjung: Please turn this into a markdown file (`.md`)

```diff
- posix-procfs: implementation of the proc fs
+ # The procfs filesystem
```

lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
...
+
+Prerequisites:
+In order to use the library one must include vfscore and ukstore.
+Furthermore in order to mount proc at /proc one must mount a root
+file system.
+
+Of the implemented files the first two have no further prerequisites,
+while the latter three require libnewlib for text formatting.
+
+Implementation:
```
nderjung:
```diff
- Implementation:
+ ## Implementation
```

lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
...
+The library creates the information on demand, when reading from the
+files. This is done with the help of static getters from the ukstore 
+library.
+The filesystem is mounted at /proc, while the containing files are 
+linked to the interior root (/).
+In order to know which files are read and return the proper
+information we parse the input in the read function.
+
+Contributing:
```
nderjung:
```diff
- Contributing:
+ ## Adding new entries into procfs
```

I would then create a list with a more detailed tutorial, including code snippets, files, existing examples, etc.

lib/posix-procfs/README
```diff
@@ -0,0 +1,36 @@
...
+In order to further expand the filesystem and add more entries there
+are a few steps necessary for the current format:
+-> create a new file after mounting the root file in procfs
+-> populate a buffer with the needed information (it would be nice if
+it is especially done with the help of ukstore)
+-> make sure to properly parse the input in the procfs_read function
+
+An example of the a test application can be found here:
+https://github.com/mandrei12/procfs-app
```
nderjung: I would directly include an example as a snippet rather than as a separate external link.

```diff
- An example of the a test application can be found here:
- https://github.com/mandrei12/procfs-app
+ For example, you can read the entries in procfs with:
+ ```c
+ #include <stdio.h>
+ #include <fcntl.h>
+ #include <errno.h>
+ #include <stdlib.h>
+ #include <unistd.h>
+ 
+ int main()
+ {
+ 	char buf[200];
+ 	int fd;
+ 	fd = open("/proc/meminfo", O_RDONLY);
+	if (fd > 0) {
+		printf("/proc/meminfo:\n");
+		read(fd, buf, 200);
+		printf("%s", buf);
+	}
+ ```
```

nderjung (changes requested):
Hi, thanks for this! :)

I have left a few comments more to do with documentation of this new library. I believe @skuenzer and @marcrittinghaus will be able to guide you more on the implementation side.

marcrittinghaus:
Hey @mandrei12,

thank you very much for providing the PR. This is much appreciated😃 As I was not involved in the topic yet, I just have a quick question on the general design. 

The concept is usually to have the procfs "driver" provide the means for other parts in the system to register nodes in the `/proc` tree. So, for example, the memory subsystem is responsible for publishing relevant memory information in procfs. Now I see that you import data from `ukstore` which makes perfectly sense and is the right way to go. But shouldn't the respective subsystems that publish the `ukstore` entries also create the nodes in procfs? Maybe by providing procfs the relevant `ukstore` nodes? Otherwise, just as it is in your version, you will have to add all the different nodes from all of the subsystems in procfs itself. I think this does not scale (from a source code and runtime performance perspective) and is also problematic concerning libraries external to the core repository. Have there been some discussions and decisions made on this previously?

razvand:
Hi, @marcrittinghaus , great comments. This is a very first community discussion on the topic. So every point should be considered. Nothing is set in stone.

The aim is to start from this PoC / draft implementation to push to a "perfect" version with recommendations / feedback / reviews from the community. Keep that coming.

mandrei12:
Hi @nderjung, @marcrittinghaus 
Thank you very much for the feedback!
@marcrittinghaus you are right. The current implementation is not exactly the best, actually, it's more like the best effort of stitching already existing parts of Unikraft in order to create something that works. As @razvand mentioned, this is a PoC so there is still much to be done, but after presenting the bachelor thesis I am excited to create a more scalable version.

marcrittinghaus:
@mandrei12 @razvand 
I marked the PR as draft at this point so it is clear that there are going to be further developments happening.

skuenzer:
Hi @marcrittinghaus, the actual idea behind this was that we do not need extra instrumentation within the libraries for procfs and just need to do it ones and only with `ukstore`. `procfs` becomes just one export format. One idea we had was to automatically populate all ukstore entries automatically under `/proc/uk` or even with a sysfs path (like `/sys/uk`). The subpaths would directly follow the ukstore entry structure (static and dynamic entries). The most common `procfs` needed entries (e.g., `cmdline`, `meminfo`) are in fact tricky because they define a custom format and often provide a combined view over multiple subsystems. Ideally, we have all needed information on ukstore to create a combined view. Thanks to the `ukstore` design, we can save the ukstore entry references so that we avoid entry lookup's over and over again. The dynamic entries will have hooks to notify when an entry disappeared or appeared. Whenever a procfs entry iks accessed we should only directly call the registered callbacks stored in each entry to gather the needed information and to create the combined output. However, it is true that these combined entries are manually implemented and I think the best place is procfs itself (because of the format definition and creating views over multiple subsystems (e.g., allocators, block caches, etc.).

lib/posix-procfs/procfs_vnops.c
```diff
@@ -0,0 +1,632 @@
...
+#endif /* CONFIG_LIBPROCFS_MOUNTS */
+
+static int
+procfs_read(struct vnode *vp, struct vfscore_file *fp __unused,
+	   struct uio *uio, int ioflag __unused)
+{
+	struct procfs_node *np =  vp->v_data;
+	size_t len;
+	char *return_me;
+	char *value = malloc(sizeof(char) * 100);
```
marcrittinghaus, line 400: On error, the buffer will be leaked. Also the size seems a bit random. Probably, it would be better to let the `read_` functions allocate the buffer. They know how large their output will be. You can use a double pointer for `value`.

razvand:
This will be reworked by @flpostolache (under supervision from @andreittr and @skuenzer ). We'll keep the PR open for reference, but it will be closed sometime in the future, after @flpostolache has the rework done.

sid-agrawal:
Hi,
I was wondering how this initial list of things in `/proc `to support is determined. Was it based on a common set of util commands used inside a Unikraft shell that use these paths in `/procfs`? 

```bash
Add support for the /proc/meminfo file
Add support for the /proc/version file
Add support for the /proc/cmdline file
Add support for the /proc/filesystems file
Add support for the /proc/mounts file
```

I have not yet spun up a Unikfrat VM, but when I spin up a docker container and enter the shell, I type `ps` due to muscle memory which iterates through `/proc/pid` directory. I understand that there is only 1 process (i.e. address space) inside a Unikraft VM (the combined app and kernel), so maybe looking at `/proc/pid` is moot.

razvand:
> Hi, I was wondering how this initial list of things in `/proc `to support is determined. Was it based on a common set of util commands used inside a Unikraft shell that use these paths in `/procfs`?
> 
> ```shell
> Add support for the /proc/meminfo file
> Add support for the /proc/version file
> Add support for the /proc/cmdline file
> Add support for the /proc/filesystems file
> Add support for the /proc/mounts file
> ```
> 
> I have not yet spun up a Unikfrat VM, but when I spin up a docker container and enter the shell, I type `ps` due to muscle memory which iterates through `/proc/pid` directory. I understand that there is only 1 process (i.e. address space) inside a Unikraft VM (the combined app and kernel), so maybe looking at `/proc/pid` is moot.

Hi, @sid-agrawal

The initial list was to have some demo files, no other reason.

PR #482, opened 2022-06-20
build: Consider obj-only libs for linking

kubanrob:
**This patch enables linking libraries consisting only of prebuilt object files, but no source files.**
Currently, libraries without any source files are ingored by the build system. The patch also fixes a related problem with libraries consisting only of device tree source (DTS) files.

Both types of libraries might by desirably:
- distribution of libraries without source code
- libraries that provide a shared device tree for multiple applications

### Prerequisite checklist

### Base target

 - Architecture(s): all
 - Platform(s): all
 - Application(s): all

### Additional configuration

There are 2 scenarios that do not build originally, but build after applying this patch.

#### DTS-only libraries

Libraries consisting only of DTS files break the build. For testing purposes, you can add [this example library](https://github.com/kubanrob/lib-nosource) to any application and select _only_ "Provide DTS source" in the configuration.

#### Object-only libraries

Libraries consisting only of object files will be ignored. For testing purposes, you can add [this example library](https://github.com/kubanrob/lib-nosource) to the app-helloworld application

1. Use (for example print) `extern const char* libnosource_export` in the application.
2. Build with _only_ "Provide C source" in the configuration (should succeed)
2. Build with _only_ "Provide prebuild object file" in the configuration (should fail)

### Description of changes

Generating the build rules, libraries without any object files need to be filtered out in order to avoid calling the linker without any input files. This patch changes the requirement for including an olib in a build from checking for source files to checking for object files. As `buildrule_libobj_multitarget` adds the object files corresponding to the source files into `LIBNAME_OBJS-y`, "normal" libraries consisting of source files still build properly.

### DIFF
--- a/support/build/Makefile.build
+++ b/support/build/Makefile.build
@@ -25,16 +25,15 @@
 
 ifneq ($(call qstrip,$(UK_LIBS) $(UK_LIBS-y)),)
 $(foreach L,$(UK_LIBS) $(UK_LIBS-y), \
-$(if $(call qstrip,$($(call uc,$(L))_SRCS) $($(call uc,$(L))_SRCS-y) \
-	    $(EACHOLIB_SRCS) $(EACHOLIB_SRCS-y)), \
 $(foreach S,$($(call uc,$(L))_SRCS) $($(call uc,$(L))_SRCS-y) \
 	    $(EACHOLIB_SRCS) $(EACHOLIB_SRCS-y), \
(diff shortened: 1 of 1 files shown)


marcrittinghaus (approved):
Hey @kubanrob,

I actually also observed this and once had to add a dummy source file when testing to link against pre-built `*.o` files. This change makes sense and looks good. I also tested it with the provided `libnosource`.

Thanks!

skuenzer (approved):
Thanks @kubanrob, this looks great.

PR #484, opened 2022-06-21
VFS Eventpoll API

marcrittinghaus:
### Prerequisite checklist

 - [X] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [X] Tested your changes against relevant architectures and platforms;
 - [X] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [X] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR adds the eventpoll API to VFScore (`lib/vfscore/eventpoll.c`). The existing file systems and special files are adapted to have stubs for the new `poll` vnode operation. The PR comes with an implementation for `pipe`. 

At the moment, the eventpoll API is included in VFScore and tightly based on how the `epoll` system calls work and which data structures they use (e.g., `epoll_event`). Hence the dependency on the `epoll.h` header. With the future introduction of the file descriptor library, the eventpoll API should be moved to this new library and perhaps be in part redesigned to be more Unikraft-specific.

This PR is part of a larger group of dependent PRs which must be merged in the following order:
1. https://github.com/unikraft/lib-newlib/pull/23
2. https://github.com/unikraft/unikraft/pull/484 (this PR)
3. https://github.com/unikraft/unikraft/pull/65
4. https://github.com/unikraft/lib-lwip/pull/17
5. https://github.com/unikraft/unikraft/pull/485

⚠️ Without having merged all of these PRs, network is no longer functional. All PRs should thus be checked and reviewed first and then merged in quick succession⚠️

PR #485, opened 2022-06-21
Introduce lib/posix-event

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

`CONFIG_LIBPOSIX_EVENT=y`

### Description of changes

This PR introduces `posix-event` which contains an implementation for the following system calls based on the VFS eventpoll API (#484):
- `poll()`, `ppoll()`
- `select()`, `pselect`, `pselect6()`
- `epoll_create(), epoll_create1()`
- `epoll_ctl()`
- `epoll_wait()`, `epoll_pwait()`
- `eventfd()`, `eventfd2()`

In contrast to the current implementation of `poll` and `select` in `lib-lwip`, file descriptors can be of any type for this implementation as long as they implement the `poll` VFS operation.

This PR is part of a larger group of dependent PRs which must be merged in the following order:
1. https://github.com/unikraft/lib-newlib/pull/23
2. https://github.com/unikraft/unikraft/pull/484
3. https://github.com/unikraft/unikraft/pull/65
4. https://github.com/unikraft/lib-lwip/pull/17
5. https://github.com/unikraft/unikraft/pull/485 (this PR)

⚠️ Without having merged all of these PRs, network is no longer functional. All PRs should thus be checked and reviewed first and then merged in quick succession⚠️ 

#### Limitations & Future Work
1. The `p`-variants are not actually working on the signal mask, yet. Idea is to add this functionality after signals in Unikraft have been revisited as part of or after the re-architecture of the scheduling API by @skuenzer.
2. Non-blocking mode in `eventfd` does not work properly (see TODO in `eventfd.c`) as the VFS write operation does not provide the file description object (´vfscore_file`) which is needed to check for fp flags. Due to an inconsistency in the VFS ops, read provides the necessary argument (screams for a fix!)
3. The `epoll_*` class of system calls does not support recursive use. So it is not possible to reference an `epoll`-fd in another `epoll`, although this is permitted by the specification. The current implementation will report an `EBADF` error.


lib/vfscore/vnode.c
```diff
@@ -450,7 +450,11 @@ vnode_dump(void)
 	struct vnode *vp;
 	struct mount *mp;
 	char type[][6] = { "VNON ", "VREG ", "VDIR ", "VBLK ", "VCHR ",
-			   "VLNK ", "VSOCK", "VFIFO" };
+			   "VLNK ", "VSOCK", "VFIFO",
+#ifdef CONFIG_LIBPOSIX_EVENT
+			   "VEPOLL", "VEVENT",
+#endif /* CONFIG_LIBPOSIX_EVENT */
+			 };
```
craciunoiuc: `char type[][6]` -> `char type[][7]` right? (and an additional space everywhere)

in the current version, for the new 2 file types, there is no space for `\0`

lib/posix-event/select.c
```diff
@@ -0,0 +1,237 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Hugo Lefeuvre <hugo.lefeuvre@manchester.ac.uk>
+ *
+ * Copyright (c) 2021, Karlsruhe Insitute of Technology (KIT)
+ * Copyright (c) 2020, NEC Laboratories Europe GmbH, NEC Corporation,
+ *                     All rights reserved.
```
craciunoiuc, line 8: I think these should be reversed

lib/posix-event/select.c
```diff
@@ -0,0 +1,237 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Hugo Lefeuvre <hugo.lefeuvre@manchester.ac.uk>
```
craciunoiuc: I think here @hlef should have his `NEC` address, not the Manchester one. Not sure 🤷

lib/posix-event/eventfd.c
```diff
@@ -0,0 +1,423 @@
...
+	if ((efd->val >= (uint64_t)-1 - val) && (fp->f_flags & O_NONBLOCK)) {
+		uk_mutex_unlock(&efd->lock);
+		return EAGAIN;
+	}
+	*/
+
+	uk_waitq_wait_event_mutex(&efd->w_wq, efd->val < (uint64_t)-1 - val,
+				  &efd->lock);
+
+	UK_ASSERT(efd->val < (uint64_t)-1 - val);
```
craciunoiuc: I get this, still, it feels a bit wrong. Maybe there is a macro somewhere for this already.

lib/posix-event/eventfd.c
```diff
@@ -0,0 +1,423 @@
...
+static unsigned int eventfd_events(struct eventfd *efd)
+{
+	unsigned int events = 0;
+
+	if (efd->val > 0)
+		events |= EPOLLIN;
+	if (efd->val < (uint64_t)-1)
+		events |= EPOLLOUT;
+	if (efd->val == (uint64_t)-1)
+		events |= EPOLLERR;
```
craciunoiuc: Here too. The must be another, more elegant, way to write this.

craciunoiuc (changes requested):
For this PR I actually found an error 😅 

Check the comment about the string length. The rest are less important.

lib/vfscore/vnode.c
marcrittinghaus: Oh, yeah! Good catch. The fun fact is that I can remember counting the number of chars to make sure but forgot the `\0` 🤦🏻‍♂️ I changed the strings to `VEPLL` and `VEVNT` to avoid breaking the formatting of the debug output.

lib/posix-event/select.c
marcrittinghaus: Why?

lib/posix-event/select.c
marcrittinghaus: Yes, changed it 👍🏻

lib/posix-event/select.c
craciunoiuc: I was talking about the order not the commit 😅 

They usually are ordered by the year of the copyright from older to newer

marcrittinghaus: I see. This implementation is loosely based on a `select` that Hugo once did but this is a new file and implementation based on the eventpoll API. So I ordered the authors according to their contribution.

craciunoiuc: Oh okay, we can leave it like this then. It was just a nitpick anyway.

lib/posix-event/eventfd.c
marcrittinghaus: Replaced the `(uint64_t)-1` with `UINT64_MAX`, ok?

lib/posix-event/eventfd.c
marcrittinghaus: Introduced the following macro:
```C
#define eventfd_no_overflow(cur, val)				\
	((cur) < UINT64_MAX - (val))
```

lib/posix-event/eventfd.c
craciunoiuc: Perfect, thanks a lot :D

lib/posix-event/eventfd.c
craciunoiuc: Perfect

lib/posix-event/poll.c
```diff
@@ -0,0 +1,190 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
razvand, line 31:
```suggestion
 */
 
```

lib/posix-event/select.c
```diff
@@ -0,0 +1,237 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
razvand, line 34:
```suggestion
 */
 
```

lib/vfscore/include/vfscore/vnode.h
```diff
@@ -59,6 +60,10 @@ enum vtype {
 	VLNK,	    /* symbolic link */
 	VSOCK,	    /* socks */
 	VFIFO,	    /* FIFO */
+#ifdef CONFIG_LIBPOSIX_EVENT
+	VEPOLL,	    /* Epoll */
```
razvand:
```suggestion
	VEPOLL,	    /* epoll */
```

lib/vfscore/include/vfscore/vnode.h
```diff
@@ -59,6 +60,10 @@ enum vtype {
...
+	VEVENT,	    /* Eventfd */
```
razvand:
```suggestion
	VEVENT,	    /* eventfd */
```

razvand (changes requested):
Hi, @marcrittinghaus . I did a first "static" pass (just browsing the code). Everything seems OK, I need to test it.

Apart from minor cosmetic suggestions and some notes I made about conventions we should use, here are some other ideas.

I would use dot (`.`) at the end of each commit message / sentence, for consistency.

[This commit](https://github.com/unikraft/unikraft/pull/485/commits/86c9724ba1d5742fd88e043795c3703193683f54) could have improved wording by replacing `and eventpoll file descriptions added` with `and eventpoll file descriptors can be added`.

This is not important now, but in the future we should add Doxygen comments to function declarations and structures in header files.

Why is `sys/eventfd.h` placed in `musl-imported/include/`? What is the criterion used to place a header file in `include/` or in `musl-imported/include/`?

marcrittinghaus:
> Apart from minor cosmetic suggestions and some notes I made about conventions we should use, here are some other ideas.

Fixed 👍🏻

> I would use dot (`.`) at the end of each commit message / sentence, for consistency.

Fixed 👍🏻 
 
> [This commit](https://github.com/unikraft/unikraft/pull/485/commits/86c9724ba1d5742fd88e043795c3703193683f54) could have improved wording by replacing `and eventpoll file descriptions added` with `and eventpoll file descriptors can be added`.

Fixed 👍🏻 

> This is not important now, but in the future we should add Doxygen comments to function declarations and structures in header files.
>
> Why is `sys/eventfd.h` placed in `musl-imported/include/`? What is the criterion used to place a header file in `include/` or in `musl-imported/include/`?

`posix-event` has no public headers except the POSIX headers, where I either copied the missing pieces or the whole header. For `eventfd.h` I just copied the header file from `musl` so it makes sense to put it in the `musl-imported` directory. And these headers should not be modified to keep them in sync with `musl`. At least that is my understanding of the reason for the distinction. I also don't think that it makes sense to write documentation for POSIX types and functions. These are publicly documented enough.

razvand:
> I also don't think that it makes sense to write documentation for POSIX types and functions. These are publicly documented enough.

Agreed. The POSIX types / functions don't need to be documented. I was overzealous when suggesting that.

razvand:
@marcrittinghaus , all is OK. @craciunoiuc , please approve this PR and then I'll approve it as well. I'll be careful to do the approval in the order signaled by @marcrittinghaus .

craciunoiuc (approved):
All good on my side!

razvand (approved):
Thanks, @marcrittinghaus . This works.

PR #502, opened 2022-07-19
plat/kvm/x86: Add SMP Support

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

This PR adds an SMP implementation for x86/KVM based on the original work by @cristian-vijelie (#244). For a detailed explanation of the changes, see the respective commit messages.

> **Warning** This PR previously depended on a couple of other PRs. However, as the same persons are assigned for review to these PRs and feedback also for commits not belonging to this PR have been given here, I closed the other PRs. So this PR now includes all of the changes previously split into independent PRs.

### Testing
For testing you may set `CONFIG_UKPLAT_LCPU_MAXCOUNT` to something larger than 3 and use the following code as `main`:
```C
#include <stdio.h>
#include <errno.h>
#include <uk/essentials.h>
#include <uk/plat/lcpu.h>
#include <uk/alloc.h>

void myfunc(struct __regs *regs __unused, void *arg __unused) {
	printf("Hello from %ld %d\n", ukplat_lcpu_id(), ukplat_lcpu_idx());
}

void myfunc2(struct __regs *regs __unused, void *arg __unused) {
	printf("Super! %ld %d\n", ukplat_lcpu_id(), ukplat_lcpu_idx());
}

int main(int argc __unused, char *argv[] __unused)
{
	struct uk_alloc *a = uk_alloc_get_default();
	__lcpuidx idx[] = {1, 2};
	void *stack[] = {a->palloc(a, 8) + 8 * __PAGE_SIZE,
			 a->palloc(a, 8) + 8 * __PAGE_SIZE};
	struct ukplat_lcpu_func fn = {myfunc, NULL};
	struct ukplat_lcpu_func fn2 = {myfunc2, NULL};
	unsigned int num = sizeof(idx) / sizeof(__lcpuidx);
	int rc;

	rc = ukplat_lcpu_start(idx, &num, stack,
			       NULL, 0);
	if (unlikely(rc))
		uk_pr_err("Failed start: %d\n", rc);

	rc = ukplat_lcpu_wait(NULL, 0, 0);
	if (unlikely(rc))
		uk_pr_err("Failed wait: %d\n", rc);

	rc = ukplat_lcpu_run(NULL, 0, &fn, 0);
	if (unlikely(rc))
		uk_pr_err("Failed run: %d\n", rc);

	rc = ukplat_lcpu_run(NULL, 0, &fn2, 0);
	if (unlikely(rc))
		uk_pr_err("Failed run: %d\n", rc);

	rc = ukplat_lcpu_wait(NULL, 0, 0);
	if (unlikely(rc))
		uk_pr_err("Failed wait: %d\n", rc);

	return 0;
}
```
`
sudo qemu-system-x86_64 -k de -m 1024 --kernel build/unikraft_kvm-x86_64 -nodefaults -nographic -vga none --device sga -s -serial stdio --enable-kvm -smp cores=3
`
NOTE: Output will be totally out of order as the cores output characters in parallel.

### Future Work

On x86, we currently use the legacy i8259 Intel PICs to process IRQs from timers and I/O devices. However, for SMP using the APIC is required. This PR keeps both the PICs and the APIC active at the same time, only routing SMP-related IRQs through the APIC. In a future PR, we have to transition the x86 platform to a modular design comparable to the GICv2 / GICv3 drivers used on ARM and transparently use PICs **OR** APIC. However, this will require more extensive changes also for pure APIC operation (e.g., 256 trap vectors, support for I/O APIC, etc.).


plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,305 @@
...
+	movl	$7, %eax
+	xorl	%ecx, %ecx
+	cpuid
+
+	/* Enable FS and GS base */
+	LCPU_ENABLE_FSGSBASE no_fsgsbase
+no_fsgsbase:
+#if CONFIG_HAVE_X86PKU
+	/* Enable memory protection keys (PKU) */
+	LCPU_ENABLE_PKU fail
```
Sairajkodilkar: Instead of halting the CPU, you can simply proceed without memory protection keys since it is an optional feature.

plat/kvm/x86/traps.c
```diff
@@ -27,135 +27,195 @@
...
-	FILL_TRAP_GATE(stack_error,     2);
-	FILL_TRAP_GATE(gp_fault,        2);
-	FILL_TRAP_GATE(page_fault,      2);
-
-	FILL_TRAP_GATE(coproc_error,    2);
-	FILL_TRAP_GATE(alignment_check, 2);
-	FILL_TRAP_GATE(machine_check,   2);
-	FILL_TRAP_GATE(simd_error,      2);
-	FILL_TRAP_GATE(virt_error,      2);
+
```
Sairajkodilkar, line 184: The trap function assignment and pointer assignment should be separated, because `traps_init()` is called by `lcpu_arch_init()` and `lcpu_arch_init()` is called separately for each CPUs. This will result in unnecessary calls to `FILL_TRAP_GATE()` and `FILL_IRQ_GATE()`. 
You can write two separate functions :
`init_trap_vectors()` which will be called only once.
`init_idtptr()` which will initialize idtptr for each CPU.

plat/kvm/x86/lcpu_start.S
marcrittinghaus: Actually, this is also the behavior in the current code and should be kept. Thanks for pointing out! I fixed this in PR #501. Thanks 👍🏻

plat/kvm/x86/traps.c
marcrittinghaus: This is not the case, because there is a check in lines [169 ff](https://github.com/unikraft/unikraft/blob/8096a7885862a52ce721affa9097ee363392419e/plat/kvm/x86/traps.c#L169) that is true for all secondary cores and which leads to an early exit that only sets the IDT based on the existing table.

Sairajkodilkar: Yeah, you are right
But wouldn't it be better to separate these two ? 
because we need to initialize the gates only once.

marcrittinghaus: Actually, I had the same thought when I was updating the functions but I wanted to keep all three functions looking and working the same (i.e., do all necessary init and activate). Also I had the following thoughts. If we separate initialization and activation, we have three options to do this:

1. Do a similar check in `traps_init()` based on the `lcpu` and call initialization only for BSP
2. Export initialization function in addition to `traps_init()` and call it in the common x86 code in `lcpu_arch_init()`, and do the check for the BSP there
3. Export initialization function and call it in `setup.c` before `lcpu_init()`

The first option would not bring any real benefit because we are still doing the check and now this function is the only one of the three which exposes the necessity for `traps_init` to differentiate between BSP and APs. The second option is the same only uglier. 

The third option is the only one which would actually remove the check altogether and make sense from an optimization and design point of view. The somewhat ugly aspect here is, however, that the `lcpu` is not yet initialized (not that bad because we don't need it) and I felt that doing a separate `trap_vectors_init()` before the `lcpu_init()` unnecessarily exposes a part of the CPU initialization (i.e., what `lcpu_init()` is supposed to do).

There is also an alternative solution:
4. Initialize the IDT statically at compile-time, thereby removing the run-time initialization altogether. This would avoid all the aforementioned problems and improve boot time! I did a quick trial to do this with assembler. The problem here is that you cannot do mask and shift operations on symbols and it is always a pain to build a loop with an iterator variable because you need to nest macros for this. So this only works with a more powerful python script that is invoked during the build and I thought, I leave this for the future, because it also affects the ability to do ASLR (symbol addresses).

What are your thoughts on all of this?

marcrittinghaus: @Sairajkodilkar ping

Sairajkodilkar: Hi @marcrittinghaus, you should go with method 3, where you call the traps initialisation function in setup.c and idtptr initialisation in `lcpu_init()`. 

I think this does does not expose the part of CPU init because we are just initialising trap gates in setup.c. 

`lcpu_init()` does the actual initialisation by assigning idtptr

marcrittinghaus: @Sairajkodilkar Ok, I will update the PR accordingly.

marcrittinghaus:
@Sairajkodilkar I have updated the PR to have the trap vector table initialization separated from the lcpu initialization.

Do you have any other comments?

plat/kvm/x86/traps.c
```diff
@@ -27,135 +27,198 @@
...
-	desc->type = 14; /* == 0b1110 */
-	desc->dpl = 0;
-	desc->p = 1;
+	desc->offset_hi	= (__u64) fun >> 16;
+	desc->offset_lo	= (__u64) fun & 0xffff;
+	desc->selector	= GDT_DESC_OFFSET(GDT_DESC_CODE);
+	desc->ist	= ist;
+	desc->type	= 14; /* == 0b1110 */
+	desc->dpl	= 0;
+	desc->p		= 1;
```
Sairajkodilkar, line 156: Define the macros for descriptor.

Sairajkodilkar (commented):
Hey @marcrittinghaus thanks for adding x86 SMP support, I have added just one suggestion, apart from that the PR looks really good

plat/kvm/x86/traps.c
marcrittinghaus: Hi @Sairajkodilkar thanks for looking into this! Your help is very much appreciated! 😃

What do you mean with:
> Define the macros for descriptor.

Sairajkodilkar: you can define the macros for descriptor type, privilege level.

marcrittinghaus: > you can define the macros for descriptor type, privilege level.

Can you provide a suggestion on how you would like this to look?

Sairajkodilkar: For example the IDT gate types can be defined as:
`#define IDT_GATE_TYPE_TASK 0x5`
`#define IDT_GATE_TYPE_INTR  0x6`
`#define IDT_GATE_TYPE_TRAP 0x7`

plat/kvm/x86/pagetable64.S
```diff
@@ -121,6 +121,7 @@
  * 0x0000000000000000 - 0x00000000ffffffff Mapping of first 4GB
  *
  * However, the first 1M is inaccessible, except for:
```
razvanvirtan, line 124: I now this is not part of this PR, but I am a bit confused by this statement. Shouldn't it be "The first 1G is inaccessible"? Because it seems like we have only 4M accessible in the first GB of the 4GB area.

cristian-vijelie: the first 1MB may give a GPE when accessing. The others can be read, and possibly written

razvanvirtan: I got it now, my bad

razvanvirtan (commented):
Hey @marcrittinghaus, this looks good to me.
There is a typo in the https://github.com/unikraft/unikraft/pull/502/commits/121eaaa15257626e9a1932644edb9769ce184ab8 commit message (`(e.g., by extending the current stack`).
Other than that, I also tested with more than 2 cores and by providing an id array to the `ukplat_lcpu_run()` function and everything works as expected.
The only observation is that the test described in the PR text only works with paging support disabled, because 1GiB pages are not supported. However, I don't think this has anything to do at all with the SMP code, but with the underlying system provided by qemu.

marcrittinghaus:
@razvand:
> **Warning** This PR previously depended on a couple of other PRs. However, as the same persons are assigned for review to these PRs and feedback also for commits not belonging to this PR have been given here, I closed the other PRs. So this PR now includes all of the changes previously split into independent PRs.

@Sairajkodilkar I added defines for the types. Please have a look if you are fine with the changes.

@razvanvirtan I fixed the typo. The error you mention has nothing to do with SMP. You have to supply `-cpu host` for 1GB pages to work. However, I tested paging and paging+smp again and found that both modes did not work (anymore) 😬 For paging alone there was an `include <uk/config.h>` missing and for paging+smp I had to adapt the paging initialization in `setup.c` to map memory regions marked as `reserved` by multiboot. These contain APIC and ACPI tables/registers.

Everything should work now.

plat/common/include/x86/apic.h
```diff
@@ -0,0 +1,160 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/common/include/x86/apic_defs.h
```diff
@@ -0,0 +1,111 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/common/include/x86/lcpu_defs.h
```diff
@@ -0,0 +1,43 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/common/x86/lcpu.c
```diff
@@ -0,0 +1,240 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/kvm/x86/lcpu_helpers.S
```diff
@@ -0,0 +1,149 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/kvm/x86/lcpu_start.S
```diff
@@ -0,0 +1,306 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

plat/kvm/x86/multiboot.S
```diff
@@ -0,0 +1,95 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Cristian Vijelie <cristianvijelie@gmail.com>
+ *
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University Politehnica of Bucharest.
```
razvand:
```suggestion
 * Copyright (c) 2022, University POLITEHNICA of Bucharest.
```

razvand (commented):
Hi, @marcrittinghaus . This seems OK for the first pass. I'll need to test it.

See my minor comments regarding the full name of UPB.

Two other items, related to commit messages:
* [This commit](https://github.com/unikraft/unikraft/pull/502/commits/b49e5195c83049725ad19dda0027a21d262be7d2) uses `xAPI`. I think it should be `xAPIC`.
* [This commit](https://github.com/unikraft/unikraft/pull/502/commits/e8246b28cb7dc9325a2cde963987f98f121b03a8) uses `plat/[kvm]:` as prefix. Shouldn't it be just `plat/kvm:` (withouth the square brackets)?

marcrittinghaus:
@razvand I updated the header files to fix the spelling of UPB and fixed the typos in the commit messages.

razvand:
@marcrittinghaus , the moment I add `newlib` to an x86 SMP build, I get the following error:

```
In file included from /home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/ukdebug/include/uk/assert.h:40,
                 from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/x86/lcpu.c:38:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/include/x86/apic.h: In function ‘x2apic_send_sipi’:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/include/x86/apic.h:118:45: error: ‘__PAGE_SHIFT’ undeclared (first use in this function)
  UK_ASSERT((addr & (APIC_ICR_VECTOR_MASK << __PAGE_SHIFT)) == addr);
```

This is fixed by adding `#include <uk/arch/limits.h>` to `plat/common/include/x86/apic.h`.

Then another issue appears:

```
In file included from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/include/uk/plat/common/lcpu.h:42,
                 from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/common/include/x86/traps.h:39,
                 from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/kvm/x86/cpu_vectors_x86_64.S:27:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/include/uk/essentials.h:46:10: fatal error: sys/cdefs.h: No such file or directory
 #include <sys/cdefs.h>
```

marcrittinghaus:
> @marcrittinghaus , the moment I add `newlib` to an x86 SMP build, I get the following error:
> ...

Hi @razvand, thank you for testing this so rigorously 🕵🏻🐛

I have fixed the build issues as you suggested for `uk/arch/limits.h` and removed the dependency on `uk/essentials.h` to fix the other issue 😃

razvand:
> > @marcrittinghaus , the moment I add `newlib` to an x86 SMP build, I get the following error:
> > ...
> 
> Hi @razvand, thank you for testing this so rigorously 🕵🏻bug
> 
> I have fixed the build issues as you suggested for `uk/arch/limits.h` and removed the dependency on `uk/essentials.h` to fix the other issue smiley

Will do a new test run. BRB.

razvand:
@marcrittinghaus , I've checked this PR and tested it to the best of my abilities. It looks good to me.

@razvanvirtan , @Sairajkodilkar , please approve the PR and leave the `...-by` message, as mentioned in the [review guidelines](https://unikraft.org/docs/contributing/review-process/#finishing-a-review). I will then do the final approval step.

Sairajkodilkar (approved):
Reviewed-by: Sairaj Kodilkar skodilkar7@gmail.com

razvanvirtan:
All good, great addition!

razvand (approved):
Thanks, @marcrittinghaus . All is OK.

Issue #496, opened 2022-07-16
syscall clock_getres not implemented


StefanJum:
### Describe the bug

The `clock_getres`system call is not implemented, at the moment it [just returns 0](https://github.com/unikraft/unikraft/blob/staging/lib/uktime/time.c#L142).
This becomes an issue with binary compatibility since it is used by some run-time environments.
For example, the dlang helloworld app crashes because it does a division by the number of nanoseconds returned by `clock_getres` in the (struct timespec *res) parameter (see the #495 issue for more detailed information).

marcrittinghaus:
Solved with PR #503

Issue #495, opened 2022-07-16
binary-compat: dlang: Unhandled Trap 0 error


StefanJum:
### Describe the bug

When running dlang static PIE binaries under binary compatibility mode, using [run-app-elfloader](https://github.com/unikraft/run-app-elfloader), it crashes with this message:
```
[    0.561820] CRIT: [libkvmplat] <trace.c @   41> RIP: 000000003f8701ae CS: 0008
[    0.563785] CRIT: [libkvmplat] <trace.c @   42> RSP: 000000003ffafb80 SS: 0010 EFLAGS: 00010293
[    0.566175] CRIT: [libkvmplat] <trace.c @   44> RAX: 000000003b9aca00 RBX: 000000003fa54f88 RCX: 0000000000000000
[    0.568947] CRIT: [libkvmplat] <trace.c @   46> RDX: 0000000000000000 RSI: 000000003ffafb80 RDI: 0000000000000001
[    0.571702] CRIT: [libkvmplat] <trace.c @   48> RBP: 0000000000000003 R08: 000000003f8700f0 R09: 0000000040020001
[    0.574433] CRIT: [libkvmplat] <trace.c @   50> R10: 000000003fa52f20 R11: 0000000000000246 R12: 000000003fc0dea0
[    0.577181] CRIT: [libkvmplat] <trace.c @   52> R13: 000000003fc017c0 R14: 0000000000000000 R15: 0000000000000001
[    0.580019] CRIT: [libkvmplat] <traps.c @   80> 3f8701ae <+0>: idiv %rcx
[    0.581823] CRIT: [libkvmplat] <traps.c @   80> 3f8701b1 <+3>: mov %rax, (%rbx)
[    0.583776] CRIT: [libkvmplat] <traps.c @   80> 3f8701b4 <+6>: test %rax, %rax
[    0.585699] CRIT: [libkvmplat] <traps.c @   80> 3f8701b7 <+9>: jnz 0x000000003F870144
[    0.587851] CRIT: [libkvmplat] <traps.c @   80> 3f8701b9 <+11>: add $0x20, %rsp
[    0.589852] CRIT: [libkvmplat] <traps.c @   80> 3f8701bd <+15>: pop %rbx
[    0.591690] CRIT: [libkvmplat] <traps.c @   80> 3f8701be <+16>: ret
[    0.593343] CRIT: [libkvmplat] <traps.c @   80> 3f8701bf <+17>: nop
[...]

PR #503, opened 2022-07-20
lib/uktime: Add clock_getres syscall

StefanJum:
### Prerequisite checklist

### Description of changes

The clock_getres syscall returns the clock resolution,
stored in the `struct timespec *tp` argument.
The resolution is saved in `UKPLAT_TIME_TICK_NSEC`,
defined [here](https://github.com/unikraft/unikraft/blob/staging/include/uk/plat/time.h#L53) as `UKARCH_NSEC_PER_SEC / CONFIG_HZ`
(i.e. nanoseconds passed between consecutive clock ticks).

Solves issues #495 #496.

Signed-off-by: Stefan Jumarea <stefanjumarea02@gmail.com>

### DIFF
--- a/lib/uktime/time.c
+++ b/lib/uktime/time.c
@@ -140,10 +140,31 @@ UK_SYSCALL_R_DEFINE(int, gettimeofday, struct timeval *, tv, void *, tz)
 }
 
 UK_SYSCALL_R_DEFINE(int, clock_getres, clockid_t, clk_id,
-		    struct timespec *, res)
+		    struct timespec *, tp)
 {
-	UK_WARN_STUBBED();
+	int error;
+
+	if (!tp) {
+		error = EFAULT;
+		goto out_error;
+	}
+
+	switch (clk_id) {
+	case CLOCK_MONOTONIC:
+	case CLOCK_MONOTONIC_COARSE:
+	case CLOCK_REALTIME:
+		tp->tv_sec = 0;
+		tp->tv_nsec = UKPLAT_TIME_TICK_NSEC;
+		break;
+	default:
+		error = EINVAL;
+		goto out_error;
+	}
+
 	return 0;
+
+out_error:
+	return -error;
 }
 
 UK_SYSCALL_R_DEFINE(int, clock_gettime, clockid_t, clk_id, struct timespec*, tp)


marcrittinghaus:
Actually, I am not an expert on time-related syscalls. @StefanJum While this make sense at the first sight, have you done any tests to verify that you return the correct value?

StefanJum:
> Actually, I am not an expert on time-related syscalls. @StefanJum While this make sense at the first sight, have you done any tests to verify that you return the correct value?

Hi @marcrittinghaus. I've tested it with a few custom values for the timer frequency and it seems to return the right thing.
The syscall is implemented similarly in both [linux](https://elixir.bootlin.com/linux/latest/source/kernel/time/posix-stubs.c#L108) and [openBSD](https://github.com/openbsd/src/blob/master/sys/kern/kern_time.c#L243), doing some initial checks and returning a value previously defined as `10^9 (aka UKARCH_NSEC_PER_SEC) / clk_hz`, which is the number of nsecs between clock ticks.

razvand:
@StefanJum , @marcrittinghaus , I think it's a simple PR and it makes sense to have it part of 0.10. @dragosargint , could you please do a review?

marcrittinghaus:
> Hi @marcrittinghaus. I've tested it with a few custom values for the timer frequency and it seems to return the right thing. The syscall is implemented similarly in both [linux](https://elixir.bootlin.com/linux/latest/source/kernel/time/posix-stubs.c#L108) and [openBSD](https://github.com/openbsd/src/blob/master/sys/kern/kern_time.c#L243), doing some initial checks and returning a value previously defined as `10^9 (aka UKARCH_NSEC_PER_SEC) / clk_hz`, which is the number of nsecs between clock ticks.

Thanks for putting together this additional information. Looking at the references, I think the patch makes sense and is fine 😃 

I am only wondering why this does not need to incorporate the resolution of the TSC. What I also recognized is that Linux rounds the number of nanoseconds up (https://elixir.bootlin.com/linux/latest/source/include/vdso/jiffies.h#L9). But the latter is not relevant to this PR.

dragosargint (approved):
Looks good to me
Reviewed-by: Dragos Iulian Argint <dragosargint21@gmail.com>

StefanJum:
> Thanks for putting together this additional information. Looking at the references, I think the patch makes sense and is fine smiley
> 
> I am only wondering why this does not need to incorporate the resolution of the TSC. What I also recognized is that Linux rounds the number of nanoseconds up (https://elixir.bootlin.com/linux/latest/source/include/vdso/jiffies.h#L9). But the latter is not relevant to this PR.

@marcrittinghaus Hm, I don't know if I'm right, but I think POSIX does not offer a way to interact with the TSC [for reliability reasons](https://en.wikipedia.org/wiki/Time_Stamp_Counter#Use), the closest clock to it being [`CLOCK_MONOTONIC_RAW`](https://manpath.be/f34/2/clock_getres#L107):
 ```a  raw hardware-based  time  that  is not subject to NTP adjustments or the incremental adjustments performed  by  adjtime(3)```.
 
 I'm not sure how you would get the resolution on the TSC, but I don't think you can do that via a syscall.

marcrittinghaus (approved):
> @marcrittinghaus Hm, I don't know if I'm right, but I think POSIX does not offer a way to interact with the TSC [for reliability reasons](https://en.wikipedia.org/wiki/Time_Stamp_Counter#Use), the closest clock to it being [`CLOCK_MONOTONIC_RAW`](https://manpath.be/f34/2/clock_getres#L107): `a raw hardware-based time that is not subject to NTP adjustments or the incremental adjustments performed by adjtime(3)`.
> 
> I'm not sure how you would get the resolution on the TSC, but I don't think you can do that via a syscall.

I was just thinking about this because the monotonic clock works with the TSC. Anyways. I think the PR is fine 👍🏻

PR #504, opened 2022-07-20
lib/posix-user: New functions and cleanup

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The current `posix-user` implementation uses a lot of stubs and needs some cleanup as noted by the original authors:
image

This PR performs the cleanup by
1. separating user/passwd and group related functions
2. ensuring consistent implementation schemes (e.g., using static global variable instead of iterator)
3. fixing minor formatting issues
4. harmonizing the order in which functions are implemented and referenced

From a functional perspective the PR does the following changes:
1. adds the following functions or their implementation: `login_r`, `getpwnam_r`, `getpwuid_r`, `getgrgid_r`, `getgrouplist`
2. fixes a bug in `getgrnam_r` where not enough memory is allocated to accommodate the array of group name pointers.
3. returns proper default values (e.g., `UK_DEFAULT_GID=0`) instead of just `0`. Although practically this does not make a difference
4. adds checks to all functions to accept only valid user/group IDs as it was already done for `setresgid`

Additional note: Whereas all functions make directly use of the fact that there is only a single passwd/group entry with fixed values, we do not assume the same in the `getgr_r` and `getpw_r` functions. This is due to the fact that these two functions are more complex to implement and we want to make sure these do not need to be changed IFF we would have multiple entries at some point (e.g., impersonation).

### Future Work

There are the `capget` and `capset` system calls which are currently implemented in `posix-user` although they are probably better moved to `posix-process` as they are related to the capabilities of threads. For now, we keep them here so that we do not collide with changes to `posix-process` that are about to be published by @skuenzer for musl compatibility.

### Dependencies

Before this PR can be merged, PR https://github.com/unikraft/lib-newlib/pull/22 needs to be merged ⚠️


lib/posix-user/group.c
```diff
@@ -0,0 +1,309 @@
...
+	/* Check if the grouplist is large enough */
+	if (*ngroups < ngrps) {
+		*ngroups = ngrps;
+		return -1;
+	}
+
+	groups[0] = pwd->pw_gid;
+	if (group != pwd->pw_gid)
+		groups[1] = group;
+
```
StefanJum, line 329:
```suggestion
*ngroups = ngrps;
```
Should we update the value stored in `ngroups` to match the actual number of groups found, even if it is greater or equal to `ngrps`?
From the [getgrouplist manual](https://linux.die.net/man/3/getgrouplist):
```text
The ngroups argument is a value-result argument: on return it always contains the number of groups
found for user, including group; this value may be greater than the number of groups stored in groups.
```

lib/posix-user/group.c
```diff
@@ -0,0 +1,309 @@
...
+/* not a syscall */
+int initgroups(const char *user __unused, gid_t group __unused)
+{
+	return 0;
+}
+
+UK_SYSCALL_R_DEFINE(int, getgroups, int, size, gid_t *, list)
+{
+	return 0;
```
StefanJum, line 191:
```suggestion
	if (size <= 0)
	    return 0;
	
	UK_ASSERT(list);
	list[0] = UK_DEFAULT_GID;
	
	return 1;
```

Should this return a list containing `UK_DEFAULT_GID`?

StefanJum (commented):
Looks good besides the few comments.

lib/posix-user/group.c
marcrittinghaus: Yes, makes sense. The part of the man page describing the result value is a bit unclear but I assume you are right. I updated the code. Very good catch 😃

lib/posix-user/group.c
marcrittinghaus: Actually, `initgroups`, `getgroups`, and `setgroups` are not implemented but only stubbed. But anyways, these work on supplementary groups so this would not include the default GID, but only additional ones. Since the specification does not state if the effective group ID is returned, it is fine to just return an empty list. However, I added a commit that "implements" `setgroups` so it returns `EINVAL` if an invalid GID is attempted to be set by the user. We then assume that even if the user specified a list containing only the default (valid) GID, we deduplicate this list against the effective GID. This way the list completely collapses and it is fine to return an empty list in `getgroups`.

StefanJum: Ok, that makes a lot of sense, thanks.

StefanJum (approved):
All looks good now.

StefanJum (changes requested):
Sorry I missed that earlier, but because of the moving of `user.h`, the build on some apps fail.
For example, when building `nginx`, we get the error `uk/user.h: No such file or directory`.

I think the `user.h` file should be moved into `posix-user/include/uk` and then the guards from `group.c` and `passwd.c` changed from `#include "user.h"` into `#include <uk/user.h>`.

marcrittinghaus:
Hi @StefanJum, thanks for pointing that out! 👍🏻 I must have missed this or fixed it for me and forgot about it. Could you state where the offending include is done?

I am against keeping `uk/user.h`, however, because it is introducing the signatures of glibc APIs and that has nothing to do in a Unikraft header. For `libnewlib` the declarations are missing, for `musl` they are in `unistd.h` with `_GNU_SOURCE` defined (as it should be).

StefanJum:
> Hi @StefanJum, thanks for pointed that out! 👍🏻 I must have missed this or fixed it for me and forgot about it. Could you state where the offending include is done?
> 
> I am against keeping `uk/user.h`, however, because it is introducing the signatures of glibc APIs and that has nothing to do in a Unikraft header. For `libnewlib` the declarations are missing, for `musl` they are in `unistd.h` with `_GNU_SOURCE` defined (as it should be).

Hi @marcrittinghaus, [this is the line](https://github.com/unikraft/lib-newlib/blob/staging/include/sys/unistd.h#L40) that causes the crash. It is the only one I've found.
It makes a lot of sense to not keep `uk/user.h`, you're right.

marcrittinghaus:
> Hi @marcrittinghaus, [this is the line](https://github.com/unikraft/lib-newlib/blob/staging/include/sys/unistd.h#L40) that causes the crash. It is the only one I've found. It makes a lot of sense to not keep `uk/user.h`, you're right.

Great! Thanks for looking this up. Actually, I can remember that I removed it locally. As the upcoming release will not be 100% musl yet, I will add a PR for `newlib` that fixes this. I have to create a `newlib` PR also for a different compatibility issue with the new eventpoll stuff anyways. I will reference the PR here.

StefanJum:
> 
> Great! Thanks for looking this up. Actually, I can remember that I removed it locally. As the upcoming release will not be 100% musl yet, I will add a PR for `newlib` that fixes this. I have to create a `newlib` PR also for a different compatibility issue with the new eventpoll stuff anyways. I will reference the PR here.

That's great, you can add me as a reviewer there then we can merge both PRs at the same time.

marcrittinghaus:
> That's great, you can add me as a reviewer there then we can merge both PRs at the same time.

@StefanJum I created the PR for the `uk/user.h` header. See the PR description for the reference. However, I am not able to assign you as reviewer in `lib-newlib`. Could you just test if it now works on your side, anyways?

StefanJum (approved):
@marcrittinghaus done. It all works fine now.

marcrittinghaus:
@razvand Should someone else be assigned to this so this makes it into the release?

razvand:
> @razvand Should someone else be assigned to this so this makes it into the release?

@marcrittinghaus , I'm assigning myself. Will check it out.

lib/posix-user/user.h
```diff
@@ -1,8 +1,12 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Authors: Costin Lupu <costin.lupu@cs.pub.ro>
```
razvand, line 3: Any reason for Costin Lupu to be removed as author? His contributions have been removed from the code?

lib/posix-user/passwd.c
```diff
@@ -0,0 +1,290 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "user.h"
```
razvand, line 37: Why is this `#include "user.h"` and not `#include <uk/user.h>`?

lib/posix-user/group.c
```diff
@@ -0,0 +1,333 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "user.h"
```
razvand, line 37: Why is this `#include "user.h"` and not `#include <uk/user.h>`?

lib/posix-user/exportsyms.uk
```diff
@@ -1,101 +1,76 @@
-getpid
```
razvand, line 1: I'm not sure how these came to be here (`getpid` and `getppid`) but they are also not defined in any other place.

razvand (changes requested):
Hi, @marcrittinghaus . Looks good, thanks for this work.

See my comments inline.

Other than those, there is an inconsistency in [this commit](https://github.com/unikraft/unikraft/pull/504/commits/1ef45afd873fc3fc24379de7b2b50e5166fe5497). The commit message says the function name is `login_r` but the actual function defined is `getlogin_r`.

Also, there is an issue with newlib, caused by removing the `include/user.h` file and moving it under `user.h` (local use only in the library):

```
In file included from /home/razvan/projects/unicore/unikraft-my/apps/app-redis.git/build/libnewlibc/origin/newlib-2.5.0.20170922/newlib/libc/include/unistd.h:4,
                 from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/drivers/virtio/virtio_9p.c:34:
/home/razvan/projects/unicore/unikraft-my/libs/newlib/include/sys/unistd.h:40:10: fatal error: uk/user.h: No such file or directory
 #include <uk/user.h>
```

Also, the build command still includes the `lib/posix-user/include` folder. If it's no longer used it should be removed.

lib/posix-user/user.h
marcrittinghaus: This is actually a new file and I just copied over the header. I removed Felipe and Simon as authors.

lib/posix-user/exportsyms.uk
marcrittinghaus: Yes, these do not belong here. They are defined in `posix-process` and exported there as no `exportsyms.uk` means all global symbols are exported.

lib/posix-user/group.c
marcrittinghaus: Because the header is no longer in the `include/uk` folder. Actually the old header no longer exists as well as the directory, as you pointed out in a different comment. 

This new header is private and I use `"` so the preprocessor looks in the same directory as the source file and not in the include directories.

lib/posix-user/passwd.c
marcrittinghaus: See other comment.

marcrittinghaus:
@razvand 
> See my comments inline.

Fixed and answered 👍🏻

> Other than those, there is an inconsistency in [this commit](https://github.com/unikraft/unikraft/pull/504/commits/1ef45afd873fc3fc24379de7b2b50e5166fe5497). The commit message says the function name is `login_r` but the actual function defined is `getlogin_r`.

Fixed 👍🏻

> Also, there is an issue with newlib, caused by removing the `include/user.h` file and moving it under `user.h` (local use only in the library):
> 
> ```
> In file included from /home/razvan/projects/unicore/unikraft-my/apps/app-redis.git/build/libnewlibc/origin/newlib-2.5.0.20170922/newlib/libc/include/unistd.h:4,
>                  from /home/razvan/projects/unicore/unikraft-my/unikraft.git/plat/drivers/virtio/virtio_9p.c:34:
> /home/razvan/projects/unicore/unikraft-my/libs/newlib/include/sys/unistd.h:40:10: fatal error: uk/user.h: No such file or directory
>  #include <uk/user.h>
> ```

This issue has already been mentioned by @StefanJum and has been fixed by making this PR dependent on:

1. https://github.com/unikraft/lib-newlib/pull/22

> Also, the build command still includes the `lib/posix-user/include` folder. If it's no longer used it should be removed.

Removed 👍🏻

lib/posix-user/user.h
razvand: roger

razvand:
> This issue has already been mentioned by @StefanJum and has been fixed by making this PR dependent on:
> https://github.com/unikraft/lib-newlib/pull/22

Gotcha. I reviewed that, all is OK.

@StefanJum , please mark the `...-by` message for this PR and then I can finalize it to upstream it.

StefanJum (approved):
All good.

razvand (approved):
Works. Thanks, @marcrittinghaus 

PR #506, opened 2022-07-22
Paging arm64

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): `kvm`

### Description of changes

This PR implements paging support on arm64 according to VMSAv8-64, and updates KVM with the following changes:
* Simplify the page-table generation process, by replacing runtime generation using complex code, with compile-time generation using assembler macros. This is based on the implementation of kvm in x86.
* Initialize paging. Since there is no support for virtual address space management and on-demand paging yet, assign all available memory to the heap.
* Enforce W^X. Initializing the page tables at compile-time uses RWX permissions on the entire kernel image, as we cannot know the boundaries of each section before compile. Update the PTEs of different sections with the correct protections at runtime.
 
In Unikraft we support the lower 256TiB VA range, controlled by TTBR0. This should provide more than enough memory for unikernel use-cases and allows Unikraft to run on any Exception Level.

Of the above, the region between `0x0000_ff80_0000_0000` and `0x0000_ffff_ffff_ffff` is reserved, and is used to direct map the PA range between `0x0000_0000_0000_0000` - `0x0000_007f_ffff_ffff`, i.e. the first 512GiB. This is used internally by the paging subsystem to provide fast translation of the page tables. Any memory controlled by TTBR1, that is VA[55] == 1, is considered invalid.

We choose 48-bit addressing with a 4KiB granule. This is the most commonly used configuration, and should cover most unikernel use-cases.


michpappas:
Hi @marcrittinghaus, I have included adapted commits of #465. If you are happy with that approach perhaps you can update #465 and I can remove these commits from my PR?

A few things that could be improved:
* The `FIXME` in `arch/arm/arm64/include/uk/asm/paging.h` is because register definitions are in `plat/common/include/arm/arm64/cpu_defs.h`. That file contains both architectural definitions (i.e. things in the Architecture Reference Manual), and policy (Unikraft-specific configuration). Eventually I would like to separate these, and perhaps the former should be in a header defined under arch (has to be accessible by plat too). What do you think?
* `pgarch_pte_change_attr()` is not implemented. It seems that this function is not used so far. Not sure what the intention is.
* The initialization of paging in kvm wastes one page, that is the boot stack. I can add this memory to the frame allocator once we switch to the regular stack.

This PR works on QEMU, I also did a diff of the page tables and everything looks good. Next week I will test this PR on real hardware.

michpappas:
Changes in this update:
- Added a fix for architectures that support < 48-bit PAs.
- Removed hardware enforcement of W^X via `STCLR_EL1.WXN`. While it works on QEMU, for some reason it causes an abort when running on hardware. Could be due to several reasons, so further investigation is required.
- Minor improvements.

This is now also confirmed to work on real hardware.

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,182 @@ struct ukarch_pagetable {
...
+#define __PADDR_INV					\
+	PAGE_Lx_ALIGN_DOWN(ARM64_INVALID_ADDR, PT_MAP_LEVEL_MAX)
+
+/* Any PTE with bit[0] == 0 is invalid */
+#define PT_Lx_PTE_INVALID(lvl)		0x0UL
+
+#define ARM64_PADDR_MAX ((1ULL << ARM64_PADDR_BITS) - 1)
+#define ARM64_VADDR_MAX ((1ULL << ARM64_VADDR_BITS) - 1)
+
+#define ARM64_PADDR_VALID(paddr) (paddr <= (__paddr_t)ARM64_VADDR_MAX)
```
razvanvirtan: `ARM64_PADDR_MAX` instead of `ARM64_VADDR_MAX`

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,182 @@ struct ukarch_pagetable {
...
+	*((__pte_t *)pt_vaddr + idx) = pte;
+
+	return 0;
+}
+
+static inline __paddr_t ukarch_pt_read_base(void)
+{
+	__paddr_t reg;
+
+	__asm__ __volatile__("mrs %x0, ttbr0_el1\n" : "=r" (reg));
```
razvanvirtan, line 222: According to the docs [1], the base address is stored in bits 47:1 of the TTBR0_EL1 register, the 63:48 bits are supposed to represent the ASID. Do we need the entire register or shall we apply a mask?

[1] https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/TTBR0-EL1--Translation-Table-Base-Register-0--EL1-

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,182 @@ struct ukarch_pagetable {
...
+			     :: "r" (pt_paddr));
+	return 0;
+}
+
+static inline void ukarch_tlb_flush_entry(__vaddr_t vaddr)
+{
+	__asm__ __volatile__(
+		"	dsb ishst\n"       /* wait for writes to complete */
+		"	tlbi vaae1, %x0\n" /* invalidate by vaddr, any ASID */
+		"	dsb ish\n"         /* wait for invalidate completes */
```
razvanvirtan: `for invalidate to complete`, I think it still doesn't go over 80 characters

razvanvirtan: The docs specify that `VAAE1` means that "The invalidation applies to the PE that executes this System instruction". Do we want to propagate this to the inner / outer shareable domain or it is intended to invalidate only on the current core?

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,306 @@
...
+ * Page tables
+ */
+static inline __vaddr_t
+pgarch_pt_pte_to_vaddr(struct uk_pagetable *pt __unused, __pte_t pte,
+		       unsigned int level __maybe_unused)
+{
+	(void)pt;
+
+	__vaddr_t ret = arm64_directmap_paddr_to_vaddr(PT_Lx_PTE_PADDR(pte, level));
+
```
razvanvirtan: Maybe `return pgarch_pt_map(PT_Lx_PTE_PADDR(pte, level)) ;`? Or just remove the `ret` variable, since we don't have any check on it.

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,306 @@
...
+		  ID_AA64MMFR0_EL1_TGRAN4_MASK;
+
+	if (tgran4 != ID_AA64MMFR0_EL1_TGRAN4_SUPPORTED)
+		UK_CRASH("4KiB granule not supported\n");
+#endif /* CONFIG_LIBUKDEBUG */
+	return 0;
+}
+
+static inline int
+pgarch_pt_add_mem(struct uk_pagetable *pt, __paddr_t start, __sz len)
```
razvanvirtan, line 319: It seems like this function is almost identic to the x86 one, excepting the mapping calls. Maybe we can think about a way of placing it in a common file? (Now or in a future update)

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,306 @@
...
+	UK_ASSERT(pt->fa->addmem);
+	rc = pt->fa->addmem(pt->fa, fa_meta, start, pages, dm_off);
+	if (unlikely(rc))
+		return rc;
+
+	return 0;
+}
+
+static inline int
+pgarch_pt_init(struct uk_pagetable *pt, __paddr_t start, __sz len)
```
razvanvirtan, line 354: Same comment as for `pgarch_pt_add_mem()`, maybe it makes sense to think about a way of moving this in a common file.

plat/kvm/arm/bpt64.S
```diff
@@ -0,0 +1,170 @@
...
+.section .data
+
+.global arm64_bpt_l3_pt0
+
+/* L3: 0 - 2TiB (512GiB / entry)
+ *
+ * 0x0000000000000000 - 0x0000007fffffffff	Table descriptor to l2_pt0
+ * 0x0000008000000000 - 0x000000ffffffffff	Table descriptor to l2_pt1
+ * 0x0000010000000000 - 0x0000ff7fffffffff 	Unmapped
+ * 0x0000ff8000000000 - 0x0000ffffffffffff	Table descriptor to l2_pt511
```
razvanvirtan, line 120: This page table is only added in the next commit, should we remove it from here? (Maybe I am too pedantic, sorry)

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,182 @@ struct ukarch_pagetable {
...
+		"	isb\n"             /* sync context */
+		:: "r" (vaddr) : "memory");
+}
+
+static inline void ukarch_tlb_flush(void)
+{
+	__asm__ __volatile__(
+		"	dsb ishst\n"      /* wait for writes to complete */
+		"	tlbi vmalle1is\n" /* invalidate all inner-shareable */
+		"	dsb ish\n"        /* wait for invalidate complete */
```
razvanvirtan: `wait for invalidate to complete`

razvanvirtan (commented):
Hi, Michalis! Thanks a lot, this looks great!

Besides the on-code comments, some other remarks:

- typo in https://github.com/unikraft/unikraft/pull/506/commits/1eab843c0c91d81a7df3bbaaa02ef0e3fc3a53be title `arm64` instead of `arm64`
- typo in https://github.com/unikraft/unikraft/pull/506/commits/dc2a94a78f6e5b4fbd895b5111146f6019dcd963 description: `Replace page table` (space missing)
- I see that https://github.com/unikraft/unikraft/pull/506/commits/9304dbfba93cc2a655ee10fea665a58cb3a1f9ff removes a comment that was changed before by https://github.com/unikraft/unikraft/pull/506/commits/dc2a94a78f6e5b4fbd895b5111146f6019dcd963. Maybe we can move https://github.com/unikraft/unikraft/pull/506/commits/9304dbfba93cc2a655ee10fea665a58cb3a1f9ff earlier? (I am not sure if this would break something if the state is reverted before https://github.com/unikraft/unikraft/pull/506/commits/dc2a94a78f6e5b4fbd895b5111146f6019dcd963)
- There are some functions that are architecture dependent only because of `arm64_directmap_paddr_to_vaddr() / arm64_directmap_vaddr_to_paddr()`. I've pointed 2 of those in the comments too. Maybe we should think about making these functions common and only have a different implementation for the mapping functions? (I think this could be part of a future PR, just wanted to open the discussion)

plat/common/include/arm/arm64/paging.h
michpappas: Maybe in this case, but doesn't mean it will apply to any processor or any implementation (eg not every implementation has to use direct mapped memory), so maybe we should skip the temptation to overfit.

plat/common/include/arm/arm64/paging.h
michpappas: See above

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Good point. Now we get away with it because we don't use ASIDs, but better safe than sorry 👍🏼

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Yes we do. Similarly to what we do with full invalidation. Good catch. I'll replace with `vaae1is`. Btw, that shifts the alignment  of the comments by two characters, so I have to change back the "invalidate to complete" to the more terse "invalidate complete". I'll also make the corresponding comment in `ukarch_tlb_flush()` similar, for consistency.

plat/kvm/arm/bpt64.S
michpappas: Not pedantic 🙂  I was doing some rework and that accidentally ended up there, thanks for pointing it out.

michpappas:
Hi @razvanvirtan thanks for reviewing. I pushed an update that I think addresses all your comments. I also added the ability to pass memory type and shareability attributes. Btw, these could had been much cleaner if we mapped them directly to the PTE attributes. Right now it's not possible because some properties use "reverse biflags" (is there such a thing?!). For example, the W protection clears a bit in the PTE. So things like `PROT_READ | PROT_WRITE` break this. One possibility is to remove definitions of single protections, like `PAGE_ATTR_PROT_READ` etc, and mandate that `PAGE_ATTR_PROT_RW` and `PAGE_ATTR_PROT_RWX` (and also `PAGE_ATTR_PROT_RO`, `PAGE_ATTR_PROT_RX`) are defined by the platform. @marcrittinghaus any opinions?

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -1,9 +1,11 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
  * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *          Michalis Pappas <michalis.pappas@opensynergy.com>
  *
  * Copyright (c) 2021, Karlsruhe Institute of Technology (KIT).
  *                     All rights reserved.
+ * Copyright (c) 2022, OpenSynergy GmbH. All rights reserved.
  *
```
marcrittinghaus, line 9: You may want to move your name to the top to underline that this is primarily your work 😃 I just provided the empty skeleton.

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,213 @@ struct ukarch_pagetable {
...
+
+#define DEVICE_nGnRnE	0
+#define DEVICE_nGnRE	1
+#define DEVICE_GRE	2
+#define NORMAL_NC	3
+#define NORMAL_WT	4
+#define NORMAL_WB	5
+
+#define TTBR0_EL1_BADDR_MASK _AC(0x0000fffffffffffe, UL)
+/* ---------------- FIXME ----------------------- */
```
marcrittinghaus: Just a quick question. What is the reason to have these definitions duplicated in `cpu_defs.h`?

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,213 @@ struct ukarch_pagetable {
...
+
+#define PT_Lx_PTE_PRESENT(pte, lvl) \
+	((pte & PTE_Lx_VALID_BIT))
+
+#define PT_Lx_PTE_CLEAR_PRESENT(pte, lvl) \
+	(pte & ~PTE_Lx_VALID_BIT)
+
+#define PT_MAP_LEVEL_MAX	(PT_LEVELS - 2)
+
+#define PAGE_Lx_HAS(lvl) ((lvl) <= PT_MAP_LEVEL_MAX)
```
marcrittinghaus: If I wanted to be a nitpicker, I'd say indentation is not 100% consistent here for the `\` and the single line definitions 😏😄

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,213 @@ struct ukarch_pagetable {
...
+#define ARM64_VADDR_MAX ((1ULL << ARM64_VADDR_BITS) - 1)
+
+#define ARM64_PADDR_VALID(paddr) (paddr <= (__paddr_t)ARM64_PADDR_MAX)
+#define ARM64_VADDR_VALID(vaddr) (vaddr <= (__vaddr_t)ARM64_VADDR_MAX)
+
+#ifndef __ASSEMBLY__
+static inline __paddr_t PT_Lx_PTE_PADDR(__pte_t pte, unsigned int lvl)
+{
+	__paddr_t paddr;
+	__u64 pte_lx_map_paddr_mask[] = {
```
marcrittinghaus: You can achieve much better optimization of the function if you define this as `static`. This will also increase the chance of inlining 😃
```suggestion
	static __u64 pte_lx_map_paddr_mask[] = {
```
Not static:
image

Static:
image

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,213 @@ struct ukarch_pagetable {
...
+	return (ARM64_PADDR_VALID(end)) && (ARM64_PADDR_VALID(start));
+}
+
+static inline int ukarch_vaddr_range_isvalid(__vaddr_t start, __vaddr_t end)
+{
+#ifdef CONFIG_LIBUKDEBUG
+	UK_ASSERT(start <= end);
+#endif /* CONFIG_LIBUKDEBUG */
+
+	return (ARM64_VADDR_VALID(end)) && (ARM64_VADDR_VALID(start));
```
marcrittinghaus, line 187: This is not quite right, I think. From my understanding you also have addresses in canonical and non-canonical form on ARM64, just like on x86, don't you? What you can have now is that the given address range overlaps somehow with the non-canonical address range. This will result in at least some of the virtual addresses being invalid.

If this is the case, you may have a look at the X86 header for a check that ensures canonical virtual addresses.

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -47,19 +65,213 @@ struct ukarch_pagetable {
...
+	return 0;
+}
+
+static inline void ukarch_tlb_flush_entry(__vaddr_t vaddr)
+{
+	__asm__ __volatile__(
+		"	dsb ishst\n"         /* wait for writes to complete */
+		"	tlbi vaae1is, %x0\n" /* invalidate by vaddr, any ASID */
+		"	dsb ish\n"           /* wait for invalidate complete */
+		"	isb\n"               /* sync context */
```
marcrittinghaus: I don't quite get the indentation of the ASM instructions here (and in `ukarch_tlb_flush`). Did you mean this?
```suggestion
		"dsb	ishst\n"         /* wait for writes to complete */
		"tlbi	vaae1is, %x0\n"  /* invalidate by vaddr, any ASID */
		"dsb	ish\n"           /* wait for invalidate complete */
		"isb\n"                  /* sync context */
```

plat/common/include/arm/arm64/paging.h
marcrittinghaus: I agree, good point @razvanvirtan, but I would revisit in the future and we should indeed take care to not overfit as @michpappas said. We also have a RISC-V implementation from @eduardvintila and can take that into consideration later, too.

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,82 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ *
```
marcrittinghaus, line 5: Please also add your name to the files (source + header) since you were also involved.

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,82 @@
...
+
+		pages = DIV_ROUND_UP(d.len, PAGE_SIZE);
+		prot = PAGE_ATTR_PROT_READ;
+
+		if (d.flags & UKPLAT_MEMRF_EXECUTABLE)
+			prot |= PAGE_ATTR_PROT_EXEC;
+		else if (d.flags & UKPLAT_MEMRF_WRITABLE)
+			prot |= PAGE_ATTR_PROT_WRITE;
+
+		uk_pr_info("Setting protections for %s: %"
```
marcrittinghaus: I had this as `uk_pr_debug()` and the error below as `uk_pr_err()`. I guess you changed this for testing?

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,82 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <uk/config.h>
+#include <uk/plat/memory.h>
+#include <uk/plat/paging.h>
+
```
marcrittinghaus, line 43: Perhaps add `#include <uk/print.h>`

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,82 @@
...
+#endif /* CONFIG_UKPLAT_MEMRNAME */
+
+void __weak enforce_w_xor_x(void)
+{
+	struct ukplat_memregion_desc d;
+	unsigned long prot, pages;
+	int rc, i = 0;
+
+	while ((rc = ukplat_memregion_get(i++, &d)) >= 0) {
+
```
marcrittinghaus, line 74: Open issue as mentioned by @mariasfiraiala @eduardvintila
The base address might not be page-aligned.

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,342 @@
...
+	return (__paddr_t)(vaddr - DIRECTMAP_AREA_START);
+}
+
+static inline __pte_t
+pgarch_pte_create(__paddr_t paddr, unsigned long attr, unsigned int level,
+		  __pte_t template, unsigned int template_level __unused)
+{
+	__pte_t pte;
+
+	unsigned long pte_lx_map_paddr_mask[] = {
```
marcrittinghaus: I suspect that defining this as `static` will lead similar optimizations as in `PT_Lx_PTE_PADDR`.

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,342 @@
...
+		pte |= ATTR_IDX(DEVICE_nGnRnE);
+		break;
+	case PAGE_ATTR_TYPE_DEVICE_nGnRE:
+		pte |= ATTR_IDX(DEVICE_nGnRE);
+		break;
+	case PAGE_ATTR_TYPE_DEVICE_GRE:
+		pte |= ATTR_IDX(DEVICE_GRE);
+		break;
+	default:
+		UK_CRASH("Invalid memory type\n");
```
marcrittinghaus: Same as above

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,342 @@
...
+		pte |= ATTR_SH(ATTR_SH_IS);
+		break;
+	case PAGE_ATTR_SHAREABLE_OS:
+		pte |= ATTR_SH(ATTR_SH_OS);
+		break;
+	case PAGE_ATTR_SHAREABLE_NS:
+		pte |= ATTR_SH(ATTR_SH_NS);
+		break;
+	default:
+		UK_CRASH("Invalid shareability type\n");
```
marcrittinghaus: This is pretty drastic but also justified as it can be expected that only valid flags should be supplied. However, considering this, I would do a `UK_ASSERT(0)` instead to save the code and data in release builds.

You can also try:
```suggestion
		UK_ASSERT(0 && "Invalid shareability type");
```

plat/kvm/arm/entry64.S
```diff
@@ -142,4 +93,3 @@ ENTRY(_libkvmplat_entry)
 	ldr x0, =_dtb
 	b _libkvmplat_start
 END(_libkvmplat_entry)
-
```
marcrittinghaus: Is this removing the empty line at the end?

plat/kvm/include/kvm-arm/arm64/mm.h
```diff
@@ -76,9 +76,6 @@
 
 #ifndef __ASSEMBLY__
 
-/* Total memory size that will be used by pagetable */
-extern uint64_t page_table_size;
-
 #endif /*__ASSEMBLY__ */
```
marcrittinghaus: I would remove the whole `__ASSEMBLY__` block as it is not used anymore and can be easily reintroduced when necessary 🧼

plat/kvm/arm/bpt64.S
```diff
@@ -0,0 +1,183 @@
...
+ */
+.align 12
+arm64_bpt_l1_pt0:
+	pte_zero        0x0000000000000000, 64
+	pte_fill        0x0000000008000000, 448, 1, SECT_ATTR_DEVICE_nGnRE
+
+/* L1: 1GiB - 2GiB (2MiB per entry)
+ *
+ * 0x0000000000000000 - 0x00000000001fffff	Table descriptor to l0_pt0
+ * 0x0000000000200000 - 0x0000000000000000	RAM	@ 1.2GiB
```
marcrittinghaus: The end address seems missing here.

plat/kvm/arm/setup.c
```diff
@@ -205,6 +207,76 @@ static void _dtb_get_cmdline(char *cmdline, size_t maxlen)
...
+	uint64_t start;
+	uint64_t len;
+	unsigned long frames;
+	__sz free_memory, res_memory;
+
+	/* Assign all available memory beyond the boot stack
+	 * to the frame allocator.
+	 */
+	start = ALIGN_UP(__END + __STACK_SIZE, PAGE_SIZE);
+	len   = _libkvmplat_cfg.bstack.end - start;
```
marcrittinghaus, line 223: There seems to be something wrong here, or I just don't get it 🤔 In lines 160+ the boot stack is positioned at the end of the available memory, but here you set things up for locating it right behind the image. The latter is also in line with the comment before the configuration of `heap.start`. But then in lines 266+ you configure the boot stack to be behind the heap again.

plat/kvm/arm/bpt64.S
```diff
@@ -0,0 +1,183 @@
...
+	pte_fill      0x0000000040200000, 511, 1, SECT_ATTR_NORMAL
+
+/* L0: 1GiB - 1.2GiB (4KiB / entry)
+ *
+ * 0x0000000040000000 - 0x00000000401fffff	Kernel	@ 1GiB
+ */
+.align 12
+arm64_bpt_l0_pt0:
+	pte_fill        0x0000000040000000, 512, 0, SECT_ATTR_NORMAL_RWX
+
```
marcrittinghaus: Open issue as mentioned by @mariasfiraiala @eduardvintila
Executable code for sqlite reaches into above `0x40200000`. Open for discussion

marcrittinghaus (commented):
Hey @michpappas,

great work with the implementation for ARM64 🎉 Here just a few things I noticed when reading the code

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Will do. I previously put yours first based on alphabetical order. Btw, the author statement is maybe obsolete these days as all information is in git. And as more people change a file, the list of authors will grow. Even if we consider only adding the original author(s), that information may also become obsolete as the code evolves, as the majority of the code may even no longer be from the original author. That is also partially the reason I created `bpt64.S` and didn't use the existing `pagetable64.S`, as it has the following note:
```
  1 /* SPDX-License-Identifier: BSD-2-Clause */
  2 /*-
  3  * This code is based on:
  4  * https://github.com/freebsd/freebsd/blob/master/sys/arm64/arm64/locore.S
  5  *
  6  * Authors: Andrew Turner
  7  *          Wei Chen <Wei.Chen@arm.com>
  8  *
  9  * Copyright (c) 2012-2014 Andrew Turner. All rights reserved.
 10  * Copyright (c) 2018 Arm Ltd. All rights reserved.
 11  *
```
After this PR we remove the majority of that code. If we added the contents of `bpt64.S` there, the statement in the header would just no longer be true.

PS: I hope you're happy with the attributions in general, as I used the x86 implementation as a reference. In `plat/common/include/arm/arm64/paging.h` I state:
```
  4  *
  5  * Based on plat/common/include/x86/paging.h by Marc Rittinghaus.
  6  *
```

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Okay, let me try to explain the problem: Currently `cpu_defs.h` contains two types of information:
1. Architectural definitions, ie information provided in the Arm ARM.
2. Definitions defined by the platform. For example the indices we choose to use in MAIR_EL1 for different memory types, or the default PTE attributes we choose to use when we populate the pagetables. I will refer to this below as "policy".

This is problematic because:
1. Register definitions cannot be used by headers under `arch/` (hence the duplication you see above)
2. The semantics of `cpu_defs.h` are not clear. It's a mixed bag.
3. Policy is still required in `arch`, see the definitions of device-type in lines 79-84.

To solve 1 and 2, I would rather have all architectural definitions in a single header under `arch/arm/arm64/`, with a rule that says "If it's not defined in the Arm ARM, it doesn't go here". I think that's much cleaner that spreading architectural definitions across different headers, and it's also a common practice in general.

For 3, I don't know what's the right way to go 😅

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Okay, what is the rule? That's also a good candidate for the coding style document.

arch/arm/arm64/include/uk/asm/paging.h
michpappas: That's a cool optimization, thanks! 👍🏼

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Yes there are in Arm too, although not termed as "canonical". The idea is the same though, you have an upper and a lower region, separated by a hole of invalid addresses. The size of each is configurable, and the PTs of each are managed by a different register, in EL1 that is `TTBR0_EL1` and `TTBR1_EL1`. An OS can then use the lower range / `TTBR0_EL1` for userspace, and the upper range / `TTBR1_EL1` for its own mappings. When address translation happens, bit 55 is used for selecting the TTBRx register. The remaining bits must be equal to bit[55]. If `TBI` is enabled, the upper byte is ignored and can be used for various purposes. If PAuth and / or MTE are enabled, the upper bits store the PAC and memory tags.

In our implementation we set `sizeof(vaddr) == sizeof(paddr) == 48 bits` and use TTBR0_EL1, that is our address space is between `0x0000_0000_0000_0000` and `0x0000_ffff_ffff_ffff`. The high range is invalid. What I see I'm missing here is masking out the top bits when doing the comparison. From what I understand that's what you also do in `X86_VADDR_CANONOCALZE` Is that what you had in mind? If not could you please provide an example?

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Another candidate for the coding style document 🙂 Previously we used syntax with prepended tabs, eg https://github.com/unikraft/unikraft/blob/staging/arch/x86/x86_64/include/uk/asm/spinlock.h#L76 so I assumed that this is the convention. Or does it only apply if we use labels? Btw, there are places where we use `asm volatile` and `__asm__ __volatile__`.  Also, if we choose not to prepend with tabs, we need to append them like `\n\t`.

plat/common/w_xor_x.c
michpappas: Will do thanks 👍🏼

plat/common/include/arm/arm64/paging.h
michpappas: For sure, thanks 👍🏼

plat/common/w_xor_x.c
michpappas: You're right, I'm pretty sure I changed them for testing, will revert.

plat/common/w_xor_x.c
michpappas: ACK. With that fix `SCTLR_EL1.WXN` also works. Will add it back.

plat/kvm/arm/entry64.S
michpappas: Looks like it is 😸 Will revert.

plat/kvm/include/kvm-arm/arm64/mm.h
michpappas: Oooops, that wasn't meant to be there, thanks

plat/kvm/arm/bpt64.S
michpappas: ACK, as discussed. Since the kernel image starts at 1GiB, will update the remaining memory of `arm64_bpt_l1_pt1` to `NORMAL_RWX`, ie to the 2GiB boundary.

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: 🤦🏻‍♂️ Oh yes sure, now I get it! There is no convention, if this has to be indented without labels. I would probably only do it if there are labels, but you are right with the `\n\t`. I never thought about that this way. So it makes sense to do it always 👍🏻 

I think the tab between instr. and operands makes also sense for multiline ASM listings to align operands as we also do in the `.S` files and the ASM code in the spinlock header. What do you think?

For the `__asm__` vs `asm`, I always go with `__asm__` because I have stored in the back of my head that it is more compatible. But I have not researched if this is actually true.

(agree, something that should go into the convention doc. I've put a note there so we don't forget)

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: It is probably only what I mostly do and what I meant with consistent in this document, that you have `\` always indented farther to the right to the same column so it is easier to spot that this is a multi-line definition.

For the single-line one you have aligned them all to the same column. See `PT_Lx_PTES` and above. I usually tend to keep this alignment if possible, but that might also just be me. What is your thought about/preference for this?

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: Yes, I totally agree that the authors statement is obsolete and actually, I would be fine with removing it but I am not sure if this is compatible with the license text because it says `the above copyright notice`. So something has to go there. I usually keep the authors up to a felt personal threshold where I think that so much code has changed that there is effectively no connection to the original authors. Of course, this is subjective and suboptimal.

Regarding the attribution: yes, yes of course. Not needed even from my perspective, so thanks 😃

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: Ah, I see. Thanks for explaining it. Makes sense now 😃 So you do not have valid addresses in the high range. For x86, I had to for example consider the case where I have a valid high address and a valid low address, but these would then span the invalid address range. For ARM, then it is totally fine to just do a `<=` comparison (when masked to low addresses as you mentioned) 😎

arch/arm/arm64/include/uk/asm/paging.h
michpappas: Okay I'll align them similarly to yours then. Looks like single line tabs are aligned at column 45, and in multi-line tabs the backslash is aligned at column 65. Generally I don't have a strong opinion, but establishing a convention would be useful for having a consistent style, and hopefully decrease the discussions on styling issues in code reviews 🙂

arch/arm/arm64/include/uk/asm/paging.h
michpappas: As before, no strong opinion. I only noticed now that prepending with tab + tab between arguments (makes sense) + comments after each instruction (also good practice) brings us often to beyond 80 chars.That requires to be creative with comments. Will update and use some newspeak to reduce comment size.

plat/kvm/arm/setup.c
michpappas: Okay, so during entry, the stack is set up to follow the kernel image (entry64.S:80). Later, after the device tree is parsed and the amount of available memory is determined, the stack is positioned to the end of the image (setup.c:160). Without paging, we would later switch to the new stack (setup.c:341).

When paging is enabled, `_init_paging()` is called before we migrate to the new stack. The calculation above, is therefore based on the current one, ie the early stack. I think the confusion is caused by KVM using the term "boot stack" ambiguously. The way I see it is that the boot stack is the early stack. The one we later switch to is just "the stack", as it is used for all runtime - not just boot. If we agree on that, I can also add a patch to rename `_libkvmplat_cfg.bstack` to `_libkvmplat_cfg.stack`, along with other references. I wanted to change that for a while now 😈

marcrittinghaus:
> Hi @razvanvirtan thanks for reviewing. I pushed an update that I think addresses all your comments. I also added the ability to pass memory type and shareability attributes. Btw, these could had been much cleaner if we mapped them directly to the PTE attributes. Right now it's not possible because some properties use "reverse biflags" (is there such a thing?!). For example, the W protection clears a bit in the PTE. So things like `PROT_READ | PROT_WRITE` break this. One possibility is to remove definitions of single protections, like `PAGE_ATTR_PROT_READ` etc, and mandate that `PAGE_ATTR_PROT_RW` and `PAGE_ATTR_PROT_RWX` (and also `PAGE_ATTR_PROT_RO`, `PAGE_ATTR_PROT_RX`) are defined by the platform. @marcrittinghaus any opinions?

These "reverse bitflags" I also face with the executable bit on x86, where you set the exact opposite in the PTE. I also thought that this seems like a waste of processor cycles because you have to add an explicit check. Removing single protections would thus be a tempting solution. It would potentially shorten some statements, too. 

But there are also some drawbacks. You cannot easily add or remove a protection obviously, which would make such code really complex. Also translating from POSIX protections to ATTR would fall into this category. At the moment we can just simply do:
```C
if prot & PROT_READ
    p |= PAGE_ATTR_PROT_READ;
if prot & PROT_WRITE
   p |= PAGE_ATTR_PROT_WRITE;
```
Without single protections it would look something like this:
```C
if (prot & PROT_READ)
  if (prot & PROT_WRITE)
     p = PAGE_ATTR_PROT_RW;
  else
     p = PAGE_ATTR_PROT_R;
```
where the if statements nest deeper with every added bit, probably making this so dirty that a lookup table would be the better solution.

What we could also do is use singe-bit protections but add helper functions that are architecture-defined:
```C
static inline unsigned long ukarch_page_attr_add(unsigned long existing, unsigned long attr);
static inline unsigned long ukarch_page_attr_rem(unsigned long existing, unsigned long attr);
```
For x86 these would just have this single if clause to handle the Non-Executable bit. Otherwise it would just do an OR/XOR.

Any thoughts on this?

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: Probably comments will get so encoded and short that only true ASM hackers understand 😎

arch/arm/arm64/include/uk/asm/paging.h
marcrittinghaus: I agree with the solution to 1 and 2 to move all strict architecture-specific definitions like registers etc. to `arch/ARCH/` and actually never understood why this has not been done right from the start 🤷🏻‍♂️ I think it is ok to have them grouped semantically though, like having PTE bit definitions still be in `paging.h`. As a compromise to a single header, this would also allow them to be mixed with "policy" definitions and still feel right. But at least all arch definitions would be in one directory and not some only available in `plat`, some globally available and (even worse) other ones duplicated.

michpappas:
Hi @marcrittinghaus I pushed a new revision that addresses all your feedback except from:
- Register definitions in arch
- Access permissions
- Your question on setup.c (not sure if my reply there answers the question)

I am planning to address these topics tomorrow.

arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -1,9 +1,11 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Authors: Marc Rittinghaus <marc.rittinghaus@kit.edu>
+ * Authors: Michalis Pappas <michalis.pappas@opensynergy.com>
+ *          Marc Rittinghaus <marc.rittinghaus@kit.edu>
  *
  * Copyright (c) 2021, Karlsruhe Institute of Technology (KIT).
  *                     All rights reserved.
+ * Copyright (c) 2022, OpenSynergy GmbH. All rights reserved.
```
marcrittinghaus: Better also switch copyrights to match author list

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,122 @@
...
+#endif /* CONFIG_ARCH_ARM64 */
+
+		/* We expect platforms to provide sane sections,
+		 * that is, if multiple sections reside in the same
+		 * page, they should have consistent protections.
+		 */
+		base = (void *)ALIGN_DOWN((unsigned long)d.base, PAGE_SIZE);
+		end  = (void *)ALIGN_UP((unsigned long)d.base + d.len,
+			PAGE_SIZE);
+		pages = DIV_ROUND_UP(end - base, PAGE_SIZE);
```
marcrittinghaus: To be on the safe side of the C standard, it might be better to use `__uptr` for `base` and `end`, as subtracting pointers is only defined for elements in the same array 🤓
```suggestion
		base = ALIGN_DOWN((__uptr)d.base, PAGE_SIZE);
		end  = ALIGN_UP((__uptr)d.base + d.len,
			PAGE_SIZE);
		pages = (end - base) / PAGE_SIZE;
```

marcrittinghaus (commented):
Hey @michpappas,

thanks for updating the PR 😃🙏🏻  There are two small things that caught my eye. Concerning the `setup.c` thing, thanks for explaining! I need a quite minute to follow. Will look into this tomorrow.

michpappas:
Hi @marcrittinghaus, here's yet another update. I introduced `arch.h` under `arch/arm/arm64` for architectural definitions, and moved register default initialization to `arch/arm/arm64/include/uk/asm/lcpu.h`. It was quite laborious but also something I wanted to do for a while now. With this we get rid of `cpu_defs.h` among everything else we remove from `plat/` in this PR. We also eliminate the duplicate definitions in `paging.h`. One thing you might notice is that we are still inconsistent in the formatting of preprocessor definitions. I am still not sure what the rule is, maybe we need to come up with a definition 🙂 

I also addressed your two earlier comments.

marcrittinghaus:
@michpappas,

This looks awesome! I like it very much that you moved the definitions up. The changes in the remaining source due to this even not seem very intrusive, which is very cool. 

From the log, I see that checkpatch did not run again 🤔Maybe @razvanvirtan or @StefanJum: Do you have the time and possibility to give it another quick local check, and compile and run it to ensure that everything works?

StefanJum:
> @michpappas,
> 
> This looks awesome! I like it very much that you moved the definitions up. The changes in the remaining source due to this even not seem very intrusive, which is very cool. 
> 
> From the log, I see that checkpatch did not run again 🤔Maybe @razvanvirtan or @StefanJum: Do you have the time and possibility to give it another quick local check, and compile and run it to ensure that everything works?

Hi. I tested the previous changes (pushed on Friday, 5 Aug) and they worked perfectly, but I didn't get to look at the last pushes. I'm on vacation now, but I may be able to take a look in the next few days.

razvanvirtan:
I can test it today in the evening.

michpappas:
Changes in this update:
- Fix bug when calculating shareability PTE attributes from function attributes
- Add memory type and shareability attributes to `pgarch_attr_from_pte()`

@marcrittinghaus did some more testing and everything looks good. Unless I discover something else tomorrow, I will only make further updates based on your feedback 🙂

razvanvirtan:
I compiled and run again a simple test and things look good. Also applied checkpatch manually for the commits and there is no error.
Only a small remark on my side, in the description of https://github.com/unikraft/unikraft/pull/506/commits/b9b23adf2f9b8b62dd112639c9b281f5ec740835, I think there is a missing file name ("Migrate
default register configuration from to lcpu.h").

marcrittinghaus:
@michpappas Great 😃 I will go through everything a last time, but it looks good I think 👍🏻

@razvanvirtan Thanks for testing and well done spotting the mistake in the commit message🕵🏻

arch/arm/arm64/include/uk/asm/arch.h
```diff
@@ -0,0 +1,368 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
```
marcrittinghaus, line 2: From a git perspective you are introducing this as a new file, so I would use the opportunity to fix spacing issues in this file. I have prepared a patch that you can use if you like
[arch.patch.gz](https://github.com/unikraft/unikraft/files/9314860/arch.patch.gz)

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+}
+
+static inline int pgarch_init(void)
+{
+	/* Sanity checks to make sure that the PE supports the minimum
+	 * requirements and the platform has set up a sane environment.
+	 * This is expected to be useful during bring-up, so keep it
+	 * conditional to UKDEBUG, to improve boot performance.
+	 */
+#ifdef CONFIG_LIBUKDEBUG
```
marcrittinghaus: Actually, what makes more sense here is the following because `ukdebug` is effectively part of every build. Ensure that you match the `endif`.

```suggestion
#ifdef CONFIG_UKDEBUG
```

plat/common/w_xor_x.c
```diff
@@ -0,0 +1,120 @@
...
+void __weak enforce_w_xor_x(void)
+{
+	struct ukplat_memregion_desc d;
+	void *base;
+	void *end;
+	unsigned long prot, pages;
+	int rc, i = 0;
+
+	while ((rc = ukplat_memregion_get(i++, &d)) >= 0) {
+ #ifdef CONFIG_ARCH_ARM_64
```
marcrittinghaus: A single space has sneaked in here

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+
+#include <errno.h>
+
+/* The implementation for arm64 supports a 48-bit virtual address space
+ * with 4KiB translation granule, as defined by VMSAv8.
+ *
+ * Notice:
+ * The Unikraft paging API uses the x86_64 convention, where pages are
+ * defined at L0, and the top level table is at L3. This is the opposite
+ * of the convention use by VMSAv8-A, where pages are defined at L3 and
```
marcrittinghaus:
```suggestion
 * of the convention used by VMSAv8-A, where pages are defined at L3 and
```

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+	return pte;
+}
+
+static inline __pte_t
+pgarch_pte_change_attr(__pte_t pte, unsigned long new_attr,
+		       unsigned int level __unused)
+{
+	(void)pte;
+	(void)new_attr;
+
```
marcrittinghaus:
```suggestion
pgarch_pte_change_attr(__pte_t pte __unused, unsigned long new_attr __unused,
		       unsigned int level __unused)
{
```

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/*
+ * Page tables
+ */
+static inline __vaddr_t
+pgarch_pt_pte_to_vaddr(struct uk_pagetable *pt __unused, __pte_t pte,
+		       unsigned int level __maybe_unused)
+{
+	(void)pt;
```
marcrittinghaus: The `(void)pt` is not necessary with the `__unused` attribute also in place.
```suggestion
```

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/* Create a table descriptor */
+static inline __pte_t
+pgarch_pt_pte_create(struct uk_pagetable *pt __unused, __paddr_t pt_paddr,
+		     unsigned int level __unused, __pte_t template,
+		     unsigned int template_level __unused)
+{
+	(void)template;
+
```
marcrittinghaus: You are not incorporating the template PTE here. I did this on x86 to make sure I do not remove attributes set by the user by a manual `pt_walk`. I assume this is not necessary on ARM? In that case you could remove the `(void)template` and add the `__unused` flag instead.

plat/common/include/arm/arm64/paging.h
```diff
@@ -0,0 +1,381 @@
...
+	 */
+#ifdef CONFIG_LIBUKDEBUG
+	unsigned int ia_size;
+	unsigned int oa_size;
+	unsigned int tgran4;
+
+	__u64 reg = SYSREG_READ64(TCR_EL1);
+
+	/* Check 48-bit addressing mode configured */
+	if (reg & TCR_EL1_DS_BIT)
```
marcrittinghaus: This is just a totally optional thing as this code is there only if the DEBUG macro is defined anyways. But an error condition can nevertheless be considered unlikely.
```suggestion
	if (unlikely(reg & TCR_EL1_DS_BIT))
```

plat/kvm/arm/bpt64.S
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * 0x0000000000000000 - 0x0000000007ffffff	Hole:   @ 0
+ * 0x0000000008000000 - 0x000000003fffffff	Devices @ 128MiB
+ */
+.align 12
+arm64_bpt_l1_pt0:
+	pte_zero        0x0000000000000000, 64
+	pte_fill        0x0000000008000000, 448, 1, PTE_BLOCK_DEVICE_nGnRE
+
+/* L1: 1GiB - 2GiB (2MiB per entry)
```
marcrittinghaus: Consistency nitpick. Also spaces and tabs mixed 😛 I prepared a patch again to save you some time: [bpt64.patch.gz](https://github.com/unikraft/unikraft/files/9315122/bpt64.patch.gz)
```suggestion
/* L1: 1GiB - 2GiB (2MiB / entry)
```

plat/kvm/arm/setup.c
marcrittinghaus: This makes sense now 😃 We should have the following layout over time then:
image

So we are "wasting" the true boot stack that is accounted for via `STACK_SIZE`, right?

marcrittinghaus (commented):
Hi @michpappas,

I have done a last complete review and I have only small polishing left.

plat/common/include/arm/arm64/paging.h
michpappas: Correct. That is because I haven't implemented yet support for next-level attributes of table descriptors. For now just replacing `(void)templat` with `__unused`.

plat/common/include/arm/arm64/paging.h
michpappas: Why not, I'll add it

plat/common/include/arm/arm64/paging.h
marcrittinghaus: Maybe add a `TODO` comment?

michpappas:
Yes, that's correct. I'll provide a diagram next time, they are easier to understand than words. In theory we could assign that wasted memory to the frame allocator for runtime mappings, eg if an allocation results into splitting a L3 mapping the frame allocator will need to create additional tables.

marcrittinghaus:
> Yes, that's correct. I'll provide a diagram next time, they are easier to understand than words. In theory we could assign that wasted memory to the frame allocator for runtime mappings, eg if an allocation results into splitting a L3 mapping the frame allocator will need to create additional tables.

Ok, thanks. I'd say for now it is fine to waste the memory. The frame allocator is not very fast ATM to work with multiple separate physical memory regions because it does an O(n) linear search in some operations to map physical addresses to regions.

michpappas:
@marcrittinghaus thanks. I updated the PR with all your comments.

marcrittinghaus:
Then I think this is good to go 👍🏻

It seems the CI/CD system is broken again, though...

razvand:
@marcrittinghaus , @michpappas , @razvanvirtan , the CI/CD system is now up.

razvand:
@razvanvirtan , @marcrittinghaus , I re-requested your review on this PR. If all is set, please approve it and add the `...-by` messages.

razvanvirtan:
All good on my side.

razvanvirtan (approved):
All good on my side.

marcrittinghaus (approved):
Thanks!

PR #522, opened 2022-08-18
lib/vfscore: Solve undefined type `loff_t`

dragosargint:
The following commit https://github.com/unikraft/unikraft/commit/e6eaeabf3c0f18989ca7a274ebab68006753b728
introduces an error when we compile an application with musl: undefined type `loff_t`.
This commit fixes this bug.
image

### DIFF
--- a/lib/vfscore/file.c
+++ b/lib/vfscore/file.c
@@ -31,14 +31,15 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "vfs.h"
+
 #include <unistd.h>
 #include <stdlib.h>
 #include <errno.h>
 #include <uk/print.h>
+#include <uk/assert.h>
 #include <vfscore/file.h>
 #include <vfscore/eventpoll.h>
-#include <uk/assert.h>
-#include "vfs.h"
 
 int fdrop(struct vfscore_file *fp)
 {


lib/vfscore/file.c
```diff
@@ -36,9 +36,9 @@
 #include <errno.h>
 #include <uk/print.h>
 #include <vfscore/file.h>
-#include <vfscore/eventpoll.h>
 #include <uk/assert.h>
 #include "vfs.h"
+#include <vfscore/eventpoll.h>
```
marcrittinghaus, line 41: Could try out:
```C
#include "vfs.h"

#include <unistd.h>
#include <stdlib.h>
#include <errno.h>
#include <uk/print.h>
#include <uk/assert.h>
#include <vfscore/file.h>
#include <vfscore/eventpoll.h>
```
This is cleaner as it moves the internal header up instead of some (random) dependency down.

marcrittinghaus, line 42: Could try out:
```C
#include "vfs.h"

#include <unistd.h>
#include <stdlib.h>
#include <errno.h>
#include <uk/print.h>
#include <uk/assert.h>
#include <vfscore/file.h>
#include <vfscore/eventpoll.h>
```
This is cleaner as it moves the internal header up instead of some (random) dependency down.

marcrittinghaus (commented):
@dragosargint Thanks for fixing this!

I am not sure how to configure the build ATM with musl so I could not reproduce the issue. I left a comment.

Also please fix the commit message and make checkpatch happy. It should also complain about the commit reference not being detailed enough and when you fix this it complains that the line is too long 🤦🏻‍♂️ At least I had this once. In that case, go for the proper reference.

lib/vfscore/file.c
dragosargint: You're right, it's cleaner this way.

marcrittinghaus (approved):
I actually did not test this with musl. But for newlib and nolibc this change works. So all fine!

PR #523, opened 2022-08-19
plat/kvm/arm: Preserve x18 in exception vectors

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): kvm
 - Application(s): N/A

### Additional configuration

None

### Description of changes

This fixes a bug in the exception vectors, where x18 is not preserved. According to the APCS:

> The role of register r18 is platform specific. If a platform ABI has need of a dedicated general-purpose register to carry inter-procedural state (for example, the thread context) then it should use this register for that purpose. If the platform ABI has no such requirements, then it should use r18 as an additional temporary register. The platform ABI specification must document the usage for this register.

This means that x18 needs to be restored either way.

[1] https://github.com/ARM-software/abi-aa/blob/60a8eb8c55e999d74dac5e368fc9d7e36e38dda4/aapcs64/aapcs64.rst

### DIFF
--- a/plat/kvm/arm/exceptions.S
+++ b/plat/kvm/arm/exceptions.S
@@ -127,7 +127,7 @@
 	ldp x24, x25, [sp, #16 * 12]
 	ldp x22, x23, [sp, #16 * 11]
 	ldp x20, x21, [sp, #16 * 10]
-	ldp x18, x19, [sp, #16 * 9]
+	/* Skip x18, x19 */
 	ldp x16, x17, [sp, #16 * 8]
 	ldp x14, x15, [sp, #16 * 7]
 	ldp x12, x13, [sp, #16 * 6]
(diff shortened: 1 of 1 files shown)


marcrittinghaus (approved):
Hi @michpappas,

I have checked and tested the code and everything is fine! Thanks 😃👍🏻

I saw that you did not update `sp` in the entry code as I did in the original patch and instead decided to always do the `add sp, sp, #__TRAP_STACK_SIZE`. This is fine for now, however, mid term we should change this because it means that the stack pointer in the trap frame is not correct (whereas all other registers are correct). For example, if you want to modify the stack pointer in a trap frame this is not possible at the moment without offsetting it by `__TRAP_STACK_SIZE`. This does not make sense and of course also confuses debuggers. But we can address this later when doing the remaining cleanups.

@razvand This is ready for approval.

michpappas:
> I saw that you did not update `sp` in the entry code as I did in the original patch and instead decided to always do the `add sp, sp, #__TRAP_STACK_SIZE`. This is fine for now, however, mid term we should change this because it means that the stack pointer in the trap frame is not correct (whereas all other registers are correct). For example, if you want to modify the stack pointer in a trap frame this is not possible at the moment without offsetting it by `__TRAP_STACK_SIZE`. This does not make sense and of course also confuses debuggers. But we can address this later when doing the remaining cleanups.

Alright, I can add it to my list for the next release. Thanks! 👍🏼

razvand (approved):
This works. Thanks, @michpappas 

PR #527, opened 2022-08-20
lib/vfscore: stdio: Include missing syscall.h

skuenzer:
The header `<uk/syscall.h>` was not included in `stdio.c`. This caused a build warning because the prototype definition of `uk_syscall_r_dup2()` was missing.


razvand (approved):
This works. Thanks for the fix, @skuenzer 

PR #535, opened 2022-08-31
`arch/libcontext`: Re-factoring context, extended context, and TLS/TCB handling

skuenzer:
This PR is a preparation for refactoring the `uksched` API and for `clone` system call support (needed by `libmusl`). It revises the API for handling context, extended context (floating point, vector units, ...) and TLS/TCB. All of these interfaces are provided by a new architecture library called `libcontext`.

The following highlights are implemented by this PR: 
- Support for lightweight contexts (threads) with and without TLS/TCB
- Support for lightweight contexts (threads) that run only ISR-safe code
- Support for user-defined stack sizes
- Support for embedding custom TCBs by third-party libraries (e.g., libc, pthread-embedded)

Please note that this PR breaks `pthread-embedded`. It needs a reintegration later (after musl integration completed).

Special thanks to @kubanrob for the Aarch64 implementation of `libcontext`.

### Base target

 - Architecture(s): [`x86_64`] [`aarch64`]
 - Platform(s): any
 - Application(s): any


dragosargint (commented):
Hi, Simon

I went through this PR, things seem in order. But the problem is that it breaks `pthread_embedded`. So either a fix for `pthread-embedded` must be taken into account, or the base of the PR must be changed, or we just accept the fact that `pthread-embedded` is broken.

lib/uksched/thread.c
```diff
@@ -41,27 +41,21 @@
...
-static void stack_push(unsigned long *sp, unsigned long value)
-{
-	*sp -= sizeof(unsigned long);
-	*((unsigned long *) *sp) = value;
-}
+typedef void (*capsule_fn)(void *);
 
-static void init_sp(unsigned long *sp, char *stack,
-		void (*function)(void *), void *data)
+static __noreturn void thread_capsule(long funcp, long argp)
```
dragosargint, line 46: Now here is what I found, this is how a thread is created for `pthread-embedded`
```
struct pte_entry_capsule {
        pte_osThreadEntryPoint entry_point;
        void *argv;
};
...
pte_osThreadCreate() { 
  ...
  capsule.entry_point = entry_point;
  capsule.argv        = argv;
  ...
  th = uk_thread_create_attr(NULL, NULL, PTE_CAPSULE_MAGIC, &capsule);
  ...
}
```

When the execution reaches this function (i.e. `thread_capsule()`), the function pointer will point to `PTE_CAPSULE_MAGIC`, which is just garbage in this case.

skuenzer: You are right, in fact, we currently decided to break `newlib` and `pthread-embedded`. The idea is to come back to it later, this is why we have currently `staging-muslpreview` as branch for these PRs. `newlib` and especially `pthread-embedded` need an updated integration. However, this would be much cleaner with the latest APIs.
Additionally, this capsule is a temporary solution for implementing libcontext with the current `uksched` API. The upcoming scheduler PR with the new API will properly integrate `libcontext` and will make this capsule unnecessary.

dragosargint (approved):
This looks great!
Reviewed-by: Dragos Iulian Argint <dragosargint21@gmail.com>

arch/arm/arm64/ctx.S
```diff
@@ -0,0 +1,161 @@
...
+ * Do a local context switch taken from thread_start64.S
+ */
+ENTRY(ukarch_ctx_switch)
+
+	/*
+	 * Store the x18 register on the stack.
+	 *
+	 * The role of register x18 is platform specific,
+	 * but saving and restoring it later is definitely
+	 * safer then leaving random values in it.
```
razvand:
```suggestion
	 * safer than leaving random values in it.
```

arch/arm/arm64/ctx.S
```diff
@@ -0,0 +1,161 @@
...
+         */
+	str x18, [sp, #-16]!
+
+	/* Save callee-saved registers to prevctx's stack */
```
razvand, line 122:
```suggestion
	/* Save callee-saved registers to prevctx's stack. */
```

arch/arm/arm64/ctx.S
```diff
@@ -0,0 +1,161 @@
...
+	mov x29, xzr /* reset frame pointer */
+#endif /* !__OMIT_FRAMEPOINTER__ */
+	ldr x1, [sp], #16
+	ldr x0, [sp], #16
+	/* jump to entrance function left on stack */
+	ldr x30, [sp], #16
+	ret
+
+/*
+ * Do a local context switch taken from thread_start64.S
```
razvand: I don't understand the wording _taken from thread_start64.S_

arch/arm/arm64/include/uk/asm/ctx.h
```diff
@@ -0,0 +1,65 @@
...
+ * SUCH DAMAGE.
+ */
+
+#include <uk/essentials.h>
+
+#ifndef __UKARCH_CTX_H__
+#error Do not include this header directly
+#endif
+
+/* stack needs to be quadword (16 byte) aligned */
```
razvand:
```suggestion
/* Stack needs to be quadword (16 bytes) aligned. */
```

razvand: Isn't quadword 8 bytes?

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+#include <uk/assert.h>
+#include <uk/print.h>
+#else /* !CONFIG_LIBUKDEBUG */
+#define UK_ASSERT(..) do {} while (0)
+#define uk_pr_debug(..) do {} while (0)
+#endif /* !CONFIG_LIBUKDEBUG */
+
+/*
+ * aarch64 ABI requires 16 bytes of TCB
+ * after the TLS arch pointer
```
razvand:
```suggestion
 * aarch64 ABI requires 16 bytes of TCB after the TLS arch pointer.
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * - in musl that included in the size of the TCB (ptread)
+ * - other libc might need to add 16 byte after the TCB
```
razvand: I'm not sure I understand these two sentences. @kubanrob , could you assist?

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+#endif /* !__UKARCH_TLS_HAVE_TCB__ */
+
+UK_CTASSERT(TCB_SIZE >= AARCH64_RESERVED);
+
+extern char _tls_start[], _etdata[], _tls_end[];
+
+/*
+ * This file implements a static exec TLS layout
+ * (variant 2 with one static TLS block).
+ * The calls create a TLS containing space for a TCB.
```
razvand: What are "The calls"? Why plural?

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+ *             |                         | |   > Static TLS block
+ *             + - - - - - - - - - - - - + |  |
+ *             |                         | |  |
+ *             | .tbss                   | |  |
+ *             |                         | |  |
+ *             +-------------------------+ /  /
+ *
+ * Note: ukarch_tlsp_get() points to the last 16 bytes of TCB of the thread.
+ *       AARCH64 ABI requieres the tlsp to point to 2 pointers (16 bytes),
+ *       the pointer to the DTV and the an implementation specific pointer.
```
razvand:
```suggestion
 *       AARCH64 ABI requires the tlsp to point to 2 pointers (16 bytes):
 *       the pointer to the DTV and the an implementation specific pointer.
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+ *
+ *       In reality, this 16 byte area is never used by the compiler
```
razvand:
```suggestion
 *       In reality, this 16 bytes area is never used by the compiler
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+ *        "Re: TLS storage offsets for TLS_ABOVE_TP", Fri, 9 Feb 2018)
+ */
+
+/*
+ * this is usually read from the elf header
+ * (p_align of the TLS section)
+ *
+ * this NEEDS to match, as aarch64 uses
+ * this at link time to calculate the offset
+ * from the tlsp
```
razvand:
```suggestion
 * This is usually read from the ELF header (p_align of the TLS section).
 *
 * This NEEDS to match, as aarch64 uses this at link time to calculate the offset
 * from the tlsp.
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+ */
+#define TLS_AREA_ALIGN 16
+
+static __sz tls_data_offset(void)
+{
+	/*
+	 * the distance between tlsp and the start of
+	 * the initial TLS segment is max(16, p_align)
+	 * (source:
+	 * https://fuchsia.dev/fuchsia-src/development/kernel/threads/tls)
```
razvand:
```suggestion
	 * The distance between tlsp and the start of the initial TLS segment is max(16, p_align).
	 * (source: https://fuchsia.dev/fuchsia-src/development/kernel/threads/tls)
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+	return (AARCH64_RESERVED > TLS_AREA_ALIGN)
+		? AARCH64_RESERVED
+		: TLS_AREA_ALIGN;
+}
+
+static __sz tls_padding_2_size(void)
+{
+	/*
+	 * The tlsp points to the mandatore 16 byte block,
+	 * padding 2 fills the rest of it.
```
razvand:
```suggestion
	 * The tlsp points to the mandatory 16 bytes block, padding 2 fills the rest of it.
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+	 *
+	 * The size of this padding is therefore determined
+	 * by the ABI.
```
razvand:
```suggestion
	 * The size of this padding is therefore determined by the ABI.
```

arch/arm/arm64/tls.c
```diff
@@ -0,0 +1,281 @@
...
+}
+
+static __sz tls_padding_1_size(void)
+{
+	/* We do not control:
+	 * - TCB_SIZE
+	 * - TLS_AREA_ALIGN
+	 * - and therefore size of padding 2
+	 * but adding data_offset don't break alignment,
+	 * so we can just align tlsp.
```
razvand:
```suggestion
	 * but adding data_offset doesn't break alignment, so we can just align tlsp.
```
What is `data_offset`?

arch/arm/ctx.c
```diff
@@ -0,0 +1,143 @@
...
+
+void ukarch_ctx_init(struct ukarch_ctx *ctx,
+		     __uptr sp, int keep_regs,
+		     __uptr ip)
+{
+	__uptr _sp;
+
+	UK_ASSERT(ctx);
+	UK_ASSERT(sp);			/* a stack is needed */
+	UK_ASSERT(ip);			/* NULL as IP will cause a crash */
```
razvand:
```suggestion
	UK_ASSERT(ip);			/* IP == NULL will cause a crash */
```

arch/arm/ctx.c
```diff
@@ -0,0 +1,143 @@
...
+	UK_ASSERT(!keep_regs && sp);	/* a stack is needed when clearing */
```
razvand: Clearing what?

arch/arm/ctx.c
```diff
@@ -0,0 +1,143 @@
...
+
+void ukarch_ctx_init_entry0(struct ukarch_ctx *ctx,
+			    __uptr sp, int keep_regs,
+			    ukarch_ctx_entry0 entry)
+{
+	__uptr _sp;
+
+	UK_ASSERT(ctx);
+	UK_ASSERT(sp);			/* a stack is needed */
+	UK_ASSERT(entry);		/* NULL as func will cause a crash */
```
razvand:
```suggestion
	UK_ASSERT(entry);		/* entry == NULL will cause a crash */
```

include/uk/arch/ctx.h
```diff
@@ -0,0 +1,234 @@
...
+#endif
+
+#ifndef __ASSEMBLY__
+struct ukarch_ctx {
+	__uptr ip;	/**< instruction pointer */
+	__uptr sp;	/**< stack pointer */
+} __packed;
+
+/*
+ * Context functions are not allowed to return
```
razvand, line 65:
```suggestion
 * Context functions are not allowed to return.
```

razvand (commented):
Hi, @skuenzer . I did a first pass and made some cosmetic comments.

I'm not that fond of very limited comments. I thin they are often limited to 50 columns, which is pretty darn ugly and hard to follow. I adhere to a more lax coding style with respect to column limits, such as [Linus Torvalds]() pointed out.

Also, I'm assuming the internal functions don't need Doxygen-style comments right?

Do you want me to be pedantic about Doxygen-style comments or we can come to that at some point in the future? (to hasten this PR being merged)

I will do a more in-depth pass + testing in the next few days.

arch/arm/arm64/tls.c
kubanrob: It is explained here very well: https://www.openwall.com/lists/musl/2018/02/09/7

Basically: aarch64 ABI demands a block of 16 bytes after the tlsp, which demands values to be stored there. However, do to the structure of the TLS these values contain not enough to access the TLS and so the compiler generated code never accesses this block. Knowing that, musl just reuses that space for other members in the TCB.

Basically these 16bytes are part of the TCB in this layout. If you don't want them to be for some libc, you need to add 16bytes to the actual size of the TCB.

arch/arm/arm64/ctx.S
kubanrob: This code was adapted from the code in `plat/common/arm/thread_start64.S`. The author / copyright information in the header also reflect that.

Probably there is a better way to annotate that.

arch/arm/arm64/include/uk/asm/ctx.h
kubanrob: Wordsize is depended on the architecture on ARM, its 32bits (see: https://stackoverflow.com/a/55430777). However, I'm not opposed to drop the "quadword" completely in favor of "16 bytes".

arch/arm/arm64/tls.c
kubanrob: This introduction was copied from `arch/x86/x86_64/tls.c`. I would not mind dropping the sentence.

arch/arm/ctx.c
kubanrob: This file is a copy of `arch/x86/ctx.c`, I forgot to mentioned that up to now ... maybe we can put this file in a common directory and fix problems in one place @skuenzer.

arch/arm/arm64/tls.c
kubanrob: It is `tls_data_offset`, which is also in the shown in the diagram above.

gabrielmocanu:
@skuenzer 
Can you please provide an update on this PR? Thanks!

arch/arm/arm64/ctx.S
skuenzer: True, fixed.

arch/arm/arm64/ctx.S
skuenzer: I think the dot at the end of one-liner comment is really only personal preference. Since the file has more one-liner comments without ending dot, I'd leave it as is.

arch/arm/arm64/ctx.S
skuenzer: It is confusing, I updated the comment to
```
/*
 * Switch context on the current LCPU.
 */
```

arch/arm/arm64/include/uk/asm/ctx.h
skuenzer: I am removing the word `quadword` to avoid confusion.

arch/arm/arm64/tls.c
skuenzer: Fixed

arch/arm/arm64/tls.c
skuenzer: Removed

arch/arm/arm64/tls.c
skuenzer: I changed the whole block to:
```c
/*
 * Aarch64 ABI requires 16 bytes of space after the tlsp as part of the
 * TCB.
 * NOTE: We assume that libraries (like musl) reserve this space when using
 *       this TLS API. You can find more information here:
 *       https://www.openwall.com/lists/musl/2018/02/09/7
 */
```

arch/arm/arm64/tls.c
skuenzer: Fixed

arch/arm/arm64/tls.c
skuenzer: Fixed

arch/arm/arm64/tls.c
skuenzer: Fixed

arch/arm/arm64/tls.c
skuenzer: Fixed

arch/arm/arm64/tls.c
skuenzer: Updated

arch/arm/arm64/tls.c
skuenzer: Updated

arch/arm/arm64/tls.c
skuenzer: Updated

arch/arm/ctx.c
skuenzer: I added a comment regarding the origin of this file

arch/arm/ctx.c
skuenzer: Standard registers... I added this to the comment.

arch/arm/ctx.c
skuenzer: Hum... 🤔 I assume this is more personal preference. I am actually fine with both ways (original and your suggestion). I updated it to reflect that we speak about the entry parameter here.

include/uk/arch/ctx.h
skuenzer: Updated

skuenzer:
> @skuenzer Can you please provide an update on this PR? Thanks!

I just pushed the update. Thanks!

include/uk/arch/ctx.h
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Similar to `ukarch_ctx_init_entry0()` but with an entry function accepting
+ * two arguments.
+ */
+void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
+			    __uptr sp, int keep_regs,
+			    ukarch_ctx_entry2 entry, long arg0, long arg1);
+
+/**
+ * Pushes a value to the stack of an remote context that should
```
adinasm:
```suggestion
 * Pushes a value to the stack of a remote context that should
```

skuenzer: Thanks a lot for spotting this 👍🏻, I fixed it.

lib/uksched/sched.c
```diff
@@ -47,6 +45,9 @@
 
 struct uk_sched *uk_sched_head;
 
+/* TODO: Define per CPU */
```
razvand: When should this TODO item be fixed, i.e define `__uk_sched_thread_current` as a per-CPU variable?

arch/Makefile.rules
```diff
@@ -0,0 +1,14 @@
+_UKARCH_TLS_HAVE_TCB=
+
+## Make sure that the reserved space for a TCB is set only once.
+## A second call of this function will make compiling stop.
+## Usage:
+##  $(eval $(call ukarch_tls_tcb_reserve,<tcb_size_in_bytes>))
```
razvand: Any reason to use double comment signs (`##`) instead of a single one? Is this something to insist on?

plat/common/arm/thread_start64.S
```diff
@@ -29,7 +29,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-#include <uk/plat/common/sw_ctx.h>
+//#include <uk/plat/common/ctx.h>
```
razvand: Shouldn't this simply be removed?

plat/common/include/arm/arm64/tls.h
```diff
@@ -30,6 +31,13 @@
 #ifndef __PLAT_CMN_ARM64_TLS_H__
 #define __PLAT_CMN_ARM64_TLS_H__
 
-#define set_tls_pointer(ptr) do { } while (0)
+#include <arm/arm64/cpu.h> /* rdmsrl, wrmsrl */
```
razvand:
```suggestion
#include <arm/arm64/cpu.h> /* SYSREG_READ, SYSREG_WRITE */
```

plat/linuxu/Config.uk
```diff
@@ -1,7 +1,7 @@
 menuconfig PLAT_LINUXU
        bool "Linux user space"
        default n
-       depends on !HAVE_SYSCALL && !HAVE_SMP && ((!ARCH_ARM_32 && !ARCH_ARM_64) || !HAVE_SCHED)
+       depends on !HAVE_SYSCALL && !HAVE_SMP && (!ARCH_ARM_32 || !HAVE_SCHED)
```
razvand, line 4: Why is `ARCH_ARM64` removed?

razvand (changes requested):
Hi, @skuenzer , see my comments.

Any idea why there are [failing checks](https://github.com/unikraft/unikraft/pull/535/checks) from Codacy? I couldn't figure it out.

Also, please fix the wording of the message for [commit `a06dc8`](https://github.com/unikraft/unikraft/pull/535/commits/a06dc8e91d0c78ff89b51d2050af070f6db9c387). I would use:

```
This change avoids reordering when setting the TLS register.
```

lib/uksched/sched.c
skuenzer: This is needs to done as soon as we have a per-CPU variable declaration. It will be required to support scheduling on multiple cores.

arch/Makefile.rules
skuenzer: Good question. It seems that I took it over from uknetdev but it is obviously not needed. I'll make it to single `#`.

plat/common/arm/thread_start64.S
skuenzer: Yes, it should 😉

plat/common/include/arm/arm64/tls.h
skuenzer: :+1:

plat/linuxu/Config.uk
skuenzer: Because with the changes, the linux user space target finally works also for Aarch64. It is making `linuxu` available when Arm is selected under architecture configuration.

lib/uksched/sched.c
skuenzer: I added some more comments to the commit message and to the comment in the source.

skuenzer:
@razvand I just updated the PR. Thanks a lot for your review.

include/uk/plat/tls.h
```diff
@@ -73,4 +80,43 @@ static inline void ukplat_tls_set(void *tls_area)
 	ukplat_tlsp_set(ukarch_tls_tlsp(tls_area));
 }
 
+/**
+ * Executes a function with a different TLS pointer activated.
+ * This wrapper function will apply a given TLS pointer before executing
+ * a given function. After the execution, the original TLS pointer is
+ * restored.
+ * NOTE: This function is decalred as `__noinline`  so that we keep
```
kubanrob: `declared`

include/uk/arch/ctx.h
```diff
@@ -35,10 +35,161 @@
...
+ * @param value
+ *   Value to push on the stack of `ctx`.
+ */
+#define ukarch_rctx_stackpush(ctx, value)				\
+	({								\
+		(ctx)->sp = ukarch_rstack_push((ctx)->sp, (value));	\
+	})
+
+/**
+ * Switch the currrent logical CPU to context `load`. The current context
```
kubanrob: `current`

include/uk/arch/ctx.h
```diff
@@ -35,10 +35,161 @@
...
+ * two arguments.
+ */
+void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
+			    __uptr sp, int keep_regs,
+			    ukarch_ctx_entry2 entry, long arg0, long arg1);
+
+/**
+ * Pushes a value to the stack of a remote context that should
+ * not be executed at the moment. This macro reserves exactly
+ * the size of `value`'s datatype on the stack.
```
kubanrob: This comment is not true for aarch64, where the stack is aligned to 16 bytes in `ukarch_rstack_push` (7ecc8534ce30823e0efceca5bb0be8df0cef94c9).
This stack alignment is mandated by the architecture.

kubanrob (commented):
I found some spelling errors (mostly my own).

- commit message of 0b2a31da7953682bfa038ff680e96ad9cd2c06d5: adapt TLS layout to ~to~ aarch64
- commit message of 89349ff6f5c831f62c36e93a3c0074a948b7343c: `adopt` -> `adapt`, `adoption` -> `adaption`
I think it would be better to rephrase it completely:  ```This commit adds a compiler barrier to `set_tls_pointer()`, which sets the TLS register. This avoids re-ordering.```

skuenzer:
Hi @kubanrob , thanks a lot for going over it as well. I implemented the changes. :+1:

kubanrob (approved):
Thank you @skuenzer!

razvand (approved):
Thanks, @skuenzer 

PR #546, opened 2022-09-10
issue #543 fix

### LINKED ISSUE #543: Build warning in `nolibc/string.c`
### Describe the bug

A build warning related to `strncat` appears when using `nolibc/string.c`:

```
/home/razvan/usoc/unikraft/lib/nolibc/string.c: In function ‘strncat’:
/home/razvan/usoc/unikraft/lib/nolibc/string.c:696:5: warning: nonnull argument ‘src’ compared to NULL [-Wnonnull-compare]
  if (src != NULL) {   
```

**Hackathon points**: 15

### Steps to reproduce

Build an application that uses `nolibc`.

### Expected behavior

There should be no build warnings.

x86_64

kvm

### Relevant log output

_No response_

mariamosn:
GCC <= 9 gives a build warning when comparing builtin function arguments with NULL. This is likely the case because compiler builtins use `__attribute(nonnull)__`.

This commit removes the NULL comparison in `strncat` that caused the build warning.

Signed-off-by: Maria Moșneag <<mariamosneag@gmail.com>>

### DIFF
--- a/lib/nolibc/string.c
+++ b/lib/nolibc/string.c
@@ -693,11 +693,9 @@ char *strncat(char *dest, const char *src, size_t n)
 
 	dest = dest + strlen(dest);
 
-	if (src != NULL) {
-		while (n && *src) {
-			n--;
-			*dest++ = *src++;
-		}
+	while (n && *src) {
+		n--;
+		*dest++ = *src++;
 	}
 
 	*dest++ = 0;


lib/nolibc/string.c
```diff
@@ -693,11 +693,9 @@ char *strncat(char *dest, const char *src, size_t n)
...
 	dest = dest + strlen(dest);
 
-	if (src != NULL) {
-		while (n && *src) {
-			n--;
-			*dest++ = *src++;
-		}
+	while (n && *src) {
+		n--;
+		*dest++ = *src++;
```
paulvlase, line 698: Shouldn't we check for buffer overflow when copying to src?

lib/nolibc/string.c
```diff
@@ -693,11 +693,9 @@ char *strncat(char *dest, const char *src, size_t n)
 
 	dest = dest + strlen(dest);
 
-	if (src != NULL) {
-		while (n && *src) {
-			n--;
-			*dest++ = *src++;
-		}
+	while (n && *src) {
```
paulvlase, line 696: Maybe we could check if n is greater then strlen(dest), use max(strlen(dest), n).

StefanJum:
@mariamosn please also update your commit message and sign it off.
You can find some guidelines [here](https://unikraft.org/docs/contributing/submitting-changes/#commits).

lib/nolibc/string.c
mariamosn: Hi! Could you please elaborate a little on this to help me understand better? Do you mean to do this instead of using __attribute__((nonnull)) and getting rid of the (src != NULL) check? Or is this an idea for further improvement by replacing the repeated decrementing of n with an iterator that goes from 0 to max(strlen(src), n) - 1?

paulvlase: Yes for my initial wrong suggestion. It is only an ideea to check that we don't copy from src array more then length of the string + the null terminator.

StefanJum:
@mariamosn we also have declared a symbol for `__attribute__((nonnull))` [here](https://github.com/unikraft/unikraft/blob/staging/include/uk/essentials.h#L82). Please use that instead of the `__attribute__` construction.

lib/nolibc/include/string.h
```diff
@@ -66,7 +66,7 @@ char *strtok(char *restrict s, const char *restrict sep);
 char *strtok_r(char *restrict s, const char *restrict sep, char **restrict p);
 char *strndup(const char *str, size_t len);
 char *strdup(const char *str);
-char *strncat(char *dest, const char *src, size_t n);
+char *strncat(char *dest, const char *src, size_t n)__nonnull;
```
StefanJum:
```suggestion
char *strncat(char *dest, const char *src, size_t n) __nonnull;
```

StefanJum (changes requested):
Besides the one comment, please squash and sign off your commits.
Other than that it works file.

lib/nolibc/include/string.h
razvand: Remove the `__nonnull` suffix here. It will be confusing why we're only using it for `strncat`. This requires a larger PR to add the `__nonnull` suffix to many more other functions.

razvand (changes requested):
Remove the `__nonnull` keyword. It's confusing as to why only some functions are using it. So just remove it and we will start a lengthier discussion on using it for other functions.

razvand:
Remove the extra commit from the PR.

razvand (changes requested):
Add a more descriptive commit message.

Configure your name and e-mail address properly: https://gitimmersion.com/lab_01.html

Then amend the author of your commit by using ` git commit --amend --no-edit --reset-author`

razvand:
@mariamosn , you have added some extra items in the commit, from `posix-socket`. Remove them.

mariamosn:
So in the end the only change should be removing the check for src? This: if (src != NULL)?

StefanJum:
The commit looks better now, but please use a message that just describes the changes, without referring an issue or any external links.
You can look at the `unikraft` commits log for reference and [here](https://dev.to/i5han3/git-commit-message-convention-that-you-can-follow-1709) for best practices.

razvand:
> So in the end the only change should be removing the check for src? This: if (src != NULL)?

Yes.

razvand (changes requested):
Hi, @mariamosn . Nice work so far.

Please do the following:
* Use `< ... >` (angular brackets) for your e-mail as part of the `Signed-off-by:` message. Use `git commit --amend` to edit that.
* Leave a blank line between the commit message and the `Signed-off-by:` message.
* Improve the commit message by following [this guideline](https://cbea.ms/git-commit/). What is suggest is:

```
GCC <= 9 gives a build warning when comparing builtin function arguments with NULL. This is likely the case because compiler builtins use `__attribute(nonnull)__`.

This commit removes the NULL comparison in `strncat` that caused the build warning.
```

mariamosn:
Thank you for the help, @razvand!

razvand (changes requested):
Hi, @mariamosn , please make the following fixes to the commit message:

Use action verbs for commit messages, i.e. instead of "Warning fix:" use "Fix build warning on nonnull argument".

Prefix the commit message with the item it refers to. So the first line should be:

```
lib/nolibc: Fix build warning on nonnull argument
```

Use the same name for author and signed-off-by. So that means removing the "ș" and replacing with "s" for consistency.

Wrap commit description message to 72 characters. The current two lines in the description are too long. I would join the two lines and wrap the resulting paragraph to 72 characters per line.

skuenzer:
> Remove the `__nonnull` keyword. It's confusing as to why only some functions are using it. So just remove it and we will start a lengthier discussion on using it for other functions.

In fact it looks confusing but if the check in the function is removed, you loose any potential compiler support to avoid an unexpected crash. I think the header `string.h` should reflect the new function behavior with the `__nonnull` attribute and maybe we need to do the same later with other function prototypes. Do you know how the builtin is actually hitting here? It looks to me as if GCC uses the builtin's just for the prototype defintion. I wonder if GCC is also replacing the function itself when compiling - and never uses our definition. Another potential solution could be to disable GCC's builtin and force it to use our function although I'd always prefer using GCC's optimized versions.

razvand:
@skuenzer , I totally support the use of `__nonnull`. But what I support more is consistency. If we use `__nonnull` on one function and not on the others, it will confuse developers.

My proposal is to:
- Remove the `_nonnull` keyword from `strncat`.
- Create an issue about adding `__nonnull` on all functions requiring it. Or at least all functions in a given compilation unit, for consistency.
- Create a PR (or multiple ones) solving that issue.

We can disable the use of the GCC builtin but that will add a compile flag `-fno-builtin-strncat`. And we will have to use one for each function. It would be really ugly and convoluted. Instead we should let the build process use the `__nonnull` tag for the declaration of `strncat`, because that's useful. In the future we would add `__nonnull` to all functions requiring it, thus doubling down on what GCC builtins are already doing (or doing it if the builtins are not doing it).

skuenzer:
> @skuenzer , I totally support the use of `__nonnull`. But what I support more is consistency. If we use `__nonnull` on one function and not on the others, it will confuse developers.
> 
> My proposal is to:
> 
>     * Remove the `_nonnull` keyword from `strncat`.
> 
>     * Create an issue about adding `__nonnull` on all functions requiring it. Or at least all functions in a given compilation unit, for consistency.
> 
>     * Create a PR (or multiple ones) solving that issue.
> 
> 
> We can disable the use of the GCC builtin but that will add a compile flag `-fno-builtin-strncat`. And we will have to use one for each function. It would be really ugly and convoluted. Instead we should let the build process use the `__nonnull` tag for the declaration of `strncat`, because that's useful. In the future we would add `__nonnull` to all functions requiring it, thus doubling down on what GCC builtins are already doing (or doing it if the builtins are not doing it).

Okay, this is fine with me. We remove the NULL argument check as the commit is proposing and open an issue for the `__nonnull` attributes that actually does not exist yet. :+1:

razvand:
@mariamosn, this PR is good. We will merge it. @skuenzer agrees with using a separate PR to add `__nonnull` keyword to functions requiring them. I created issue #730 for this.

@StefanJum, please approve the PR on your side, and then I'll do the final approval part.

StefanJum (commented):
This works as expected, removes the warning, but the app will break with a segfault if the given source is `NULL`.
@razvand should we merge it like this and fix the issue later, with a pr that will refer to #730? If so, I'll leave the review tag.

razvand:
@StefanJum, getting a segmentation fault is expected. When PR #730 is provided, a compiler warning will be issued when passing a `NULL` pointer (detectable at compile time). For runtime checks, this requires a deeper discussion on the cost-benefit of doing checks (overhead vs security).

StefanJum (dismissed):
Agreed @razvand, all looks good then.

razvand (dismissed):

razvand (approved):

StefanJum (approved):

PR #553, opened 2022-09-12
build: Add clone rule to the build system

flpostolache:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -600,6 +600,7 @@ PATCH		:= patch
 GZIP		:= gzip
 TAR		:= tar
 UNZIP		:= unzip -qq -u
+GIT		:= git
 WGET		:= wget
 SHA1SUM		:= sha1sum -b
 SHA256SUM	:= sha256sum -b
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -373,12 +373,12 @@ endef
 # A variable called LIB[LIBNAME]_ORIGIN is initialized with
 # the path to the extracted files
 #
-# unarchive $libname,$archive_fname,$dir_with_patches(optional),$extra_deps(optional)
+# unarchive $libname,$archive_fname,$extra_deps(optional)
 define unarchive =
-$(if $(filter %.tar.gz %.tgz,$(2)),$(call _extracttoorigin_tgz,$(1),$(2),$(4)),\
-$(if $(filter %.tar.xz %.txz,$(2)),$(call _extracttoorigin_tgz_xz,$(1),$(2),$(4)),\
-$(if $(filter %.tar.bz2 %.tbz2,$(2)),$(call _extracttoorigin_tgz_bz2,$(1),$(2),$(4)),\
-$(if $(filter %.zip,$(2)),$(call          _extracttoorigin_zip,$(1),$(2),$(4)),\
+$(if $(filter %.tar.gz %.tgz,$(2)),$(call _extracttoorigin_tgz,$(1),$(2),$(3)),\
+$(if $(filter %.tar.xz %.txz,$(2)),$(call _extracttoorigin_tgz_xz,$(1),$(2),$(3)),\
+$(if $(filter %.tar.bz2 %.tbz2,$(2)),$(call _extracttoorigin_tgz_bz2,$(1),$(2),$(3)),\
+$(if $(filter %.zip,$(2)),$(call          _extracttoorigin_zip,$(1),$(2),$(3)),\
 $(error $(2): missing extraction rule for archive type)\
 ))))
 UK_FETCH-y += $(BUILD_DIR)/$(1)/.origin \
@@ -419,7 +419,7 @@ endef
 
 # Downloads an archive from remote and invokes unarchive. The target filename
 # (without path!) is specified
-# fetchas $libname,$url,$target_fname(no_path!),$dir_with_patches(optional)
+# fetchas $libname,$url,$target_fname(no_path!)
 define fetchas =
 $(BUILD_DIR)/$(1)/$(3):
 	$(call verbose_cmd,WGET,$(1)':' $(2), \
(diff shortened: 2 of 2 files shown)


flpostolache:
@skuenzer This is how the clone command should look like in order to build and patch. I don't really know if the patch rule should be changed. Please have a look and tell me. Sadly, the `git clone` command can't accept cloning specific commits, only specific branches or tags. If you want to clone to a specific commit, you need to clone the entire branch and then go to that commit.(which can't be done if you tell the clone command to make a shallow clone of the repo/branch). Another way that should work better is fetching the commit/branch/tag that we are interested in instead of cloning. What do you think?

vladandrew:
@maniatro111 Let me know if you managed to speak with Simon. I

flpostolache:
@vladandrew Yes. He will take a look at it.

support/build/Makefile.rules
```diff
@@ -467,7 +487,6 @@ endef
 #################################################
 # Generates a build rule for an object of a library
 # Adds library-specific (AS/C/CXX)FLAGS to the build
-#
```
skuenzer: Please keep your changes minimal. This line change seems not to be relevant for support git clone.

support/build/Makefile.rules
```diff
@@ -459,6 +459,26 @@ define fetch2 =
 $(call fetchas2,$(1),$(2),$(3),$(notdir $(2)))
 endef
 
+# Downloads a repo from remote
```
skuenzer: Rather name it that it "Clones a git repo"

support/build/Makefile.rules
```diff
@@ -459,6 +459,26 @@ define fetch2 =
...
+# This command can not clone a specific commit from a
```
skuenzer: Add the following prefix to the comment:
```Makefile
# NOTE: This command <...>
```

support/build/Makefile.rules
```diff
@@ -355,12 +355,12 @@ endef
 # Takes all files from $patchdir, sorts them, and applies each
 # of these patches to the source tree in $source_target_dir
 #
-# patch $libname,$patchdir,$source_target_dir
+# patch $libname,$patchdir,$source_target_dir(optional)
 define patch =
 $(BUILD_DIR)/$(1)/.patched: $(BUILD_DIR)/$(1)/.origin
 	@$(foreach P,$(sort $(wildcard $(2)/*)), \
 	  $(call verbose_cmd_inner,PATCH,$(1)':' $(notdir $(P)), \
-	  $(PATCH) -sd $(BUILD_DIR)/$(1)/origin/$(3)/ -p1 < $(P)) &&)\
```
skuenzer: Is there a reason why you remove the trailing slash? It could lead to errors if people forget to add a trailing slash when using this argument. On the other hand, ending up with double-slashes should not be an issue at all.

support/build/Makefile.rules
```diff
@@ -459,6 +459,26 @@ define fetch2 =
 $(call fetchas2,$(1),$(2),$(3),$(notdir $(2)))
 endef
 
+# Downloads a repo from remote
+# This command can not clone a specific commit from a
+# specific branch of a repo
+# clone $libname $giturl $branch/tag $dir_with_patches(optional)
+define clone =
+$(BUILD_DIR)/$(1)/.origin:
+	$(call verbose_cmd,RM,$(1),$(RM) -r $(BUILD_DIR)/$(1)/origin)
```
skuenzer: Hum... :thinking: I think this is more a personal preference, but I would prefer if you run this particular RM command as part of the git cmd - otherwise the RM message is always shown even if there is nothing to delete:
```Makefile
$(call verbose_cmd,GIT,$(1),$(RM) -r $(BUILD_DIR)/$(1)/origin && $(GIT) clone -q --progress --depth 1 --branch $(3) --no-tags $(2) $(BUILD_DIR)/$(1)/origin/ && $(TOUCH) $(BUILD_DIR)/$(1)/.origin)
```

skuenzer (changes requested):
Hi @maniatro111,
thanks a lot. It took for me a while to remember why we said to remove the patch option tfrom the fetch. Maybe you could add in the decription that it is about the base folder that is part of the archive (if exist). In general, I think the motiviation should be to make `Makefile.uk` easier to read. I wonder if we shouldn't in general remove the patching option from unarchiv and also clone - just to avoid bad practise. IOf somebody wants to apply patches, the `patch` command should be used directly.
What do you think?
Thanks,
Simon

support/build/Makefile.rules
skuenzer: Please, also add `git` for `git clone` in the commit message

Makefile
```diff
@@ -598,8 +598,9 @@ YACC		:= bison
 LEX     	:= flex
 PATCH		:= patch
 GZIP		:= gzip
-TAR		:= tar
+TAR			:= tar
```
vladandrew: Here's a typo, there's an extra tab.

Makefile
```diff
@@ -598,8 +598,9 @@ YACC		:= bison
...
 UNZIP		:= unzip -qq -u
+GIT			:= git
```
vladandrew: Could you keep the same indentation level?

support/build/Makefile.rules
```diff
@@ -459,6 +459,26 @@ define fetch2 =
...
+	$(call verbose_cmd,GIT,$(1),$(GIT) clone -q --progress --depth 1 --branch $(3) --no-tags $(2) $(BUILD_DIR)/$(1)/origin/ && $(TOUCH) $(BUILD_DIR)/$(1)/.origin)
```
vladandrew: Would it make sense to also clone the submodules? `--recurse-submodules`

vladandrew (changes requested):
Hi,

I have a few more comments :D. @skuenzer do you know why we had `dir_with_patches` in the first place?

support/build/Makefile.rules
skuenzer: This is an interesting idea. Is the parameter `--recurse-submodules` harmful when no submodiules exist (e.g., in terms of exit code of `git`)? If not we could add this parameter.

flpostolache: No. It does not. Will add it to the command. Thanks, @vladandrew.

Makefile
flpostolache: Modified.

Makefile
flpostolache: Yes.

support/build/Makefile.rules
flpostolache: Added.

support/build/Makefile.rules
flpostolache: Added.

support/build/Makefile.rules
flpostolache: Yes.

support/build/Makefile.rules
flpostolache: Added.

support/build/Makefile.rules
flpostolache: No. My mistake. I added it back.

support/build/Makefile.rules
flpostolache: In the command example, the RM message will never be shown. Is it ok?

skuenzer:
> Hi,
> 
> I have a few more comments :D. @skuenzer do you know why we had `dir_with_patches` in the first place?

It is actually my fault, when implementing originally, I thought it is better to have less commands but it turned out that it was even harder to use so that nobody used it in the end. The main issue was that patches weren't plain re-usable on library version updates because many archives included a top-level directory with the same name as the archive (which often includes a version number). Because that path needs to be included into your patch files, a version change required touching them. With the separate commnd you can choose a different base folder for the patches, so that this problem can be avoided.

vladandrew (approved):
Thanks Florin! All looks good to me now.

skuenzer (approved):
Hi @maniatro111, it works for me, I tested the changes, thanks a lot!

PR #603, opened 2022-10-18
lib/ukvmem and lib/posix-mmap

### LINKED ISSUE #589: POSIX mmap
### Feature request summary

The current implementation of the mmap-group of functions is based on `malloc`. Unmapping does not work reliably with this. No support for paging code.

**Summary of objectives**

- [ ] Create data structure for efficient VA management (interval tree).
- [ ] Create uk-style API/library for VA management.
- [ ] Implement `posix-mmap` based on this API/library.
- [ ] Keep the options to switch back to "poor man's version" of `mmap` (such as when paging is not enabled).

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

- `CONFIG_LIBUKVMEM`
- `CONFIG_LIBPOSIX_MMAP`
- `CONFIG_PAGING`

### Description of changes

This is the first version of the PR that will introduce virtual address space management to Unikraft via a new library called `lib/ukvmem` (virtual memory). The functionality will be exposed in a POSIX compatible way via `lib/posix-mmap`.

However, the API introduced by `lib/ukvmem` via header `uk/vmem.h` (see `lib/ukvmem/include/uk/vmem.h`) is ready for first feedback by the community and defines the following core functions. See the header file and the examples below for usage:
```C
int uk_vma_map(struct uk_vas *vas, __vaddr_t *vaddr, __sz len,
	       unsigned long attr, unsigned long flags, const char *name,
	       const struct uk_vma_ops *ops, void *args);

int uk_vma_unmap(struct uk_vas *vas, __vaddr_t vaddr, __sz len,
		 unsigned long flags);

int uk_vma_set_attr(struct uk_vas *vas, __vaddr_t vaddr, __sz len,
		    unsigned long attr);

const struct uk_vma *uk_vma_find(struct uk_vas *vas, __vaddr_t vaddr);
```

The core concepts are:
- Every virtual address space (VAS) is represented by a `struct uk_vas`.
- Every virtual memory area (VMA) in an address space is represented by a `struct uk_vma`.
- The core library does **only** generic management of VMAs (e.g., add, modify, remove etc.) without implementing the actual different VMA types such as anonymous memory or file mappings.
- When a VMA is created, the user specifies a set of VMA operations that determine the behavior of the VMA and thereby allowing different types of VMAs (e.g., anon, file mappings) to be implemented.

The library will come with an implementation for the following VMA types:
- Simple anonymous memory
- Direct-mapped physical memory
- File-backed memory - no global cache, simple pre-initialization with file contents
- Reserved VMAs - to reserve an address range in the VAS and to represent static/fixed mappings already existing in the page table (e.g., kernel `.text`, `.rodata` etc.).
- Stack VMAs - simple anonymous memory with a guard page to signal stack overflows

Just like Linux, adjacent VMAs are merged, if possible. Also, just like Linux, changing the properties of a certain address range in an existing VMA leads to the VMA transparently being split into separate VMAs as needed to reflect the different ranges.

The library will implement **demand-paging** as well as pre-population.

## API Usage Examples
The functions `uk_vma_map_*()` are wrappers to `uk_vma_map()` referring to the corresponding VMA operations and necessary parameters for easier use.

## Example 1
Create an anonymous writable memory mapping with demand paging. Physical
memory is dynamically allocated on access.
```C
__vaddr_t vaddr = __VADDR_ANY;
uk_vma_map_anon(uk_vas_get_active(), &vaddr, <SIZE>, PAGE_ATTR_PROT_RW,
                0, NULL);
```
## Example 2
Create a writable memory mapping of an existing physical memory range
(e.g., to map device memory). UK_VMA_MAP_POPULATE is optional and just
ensures that the mappings in the page table are already established so they
are present in interrupt context.
```C
__vaddr_t vaddr = __VADDR_ANY;
uk_vma_map_dma(uk_vas_get_active(), &vaddr, <SIZE>, PAGE_ATTR_PROT_RW,
               UK_VMA_MAP_POPULATE, NULL, <PADDR>);
```

## Example 3
Create a writable memory mapping of a newly allocated physical memory
area that uses 2MB pages. The size must be a multiple of 2MB.
```C
__vaddr_t vaddr = __VADDR_ANY;
uk_vma_map_anon(uk_vas_get_active(), &vaddr, <SIZE>, PAGE_ATTR_PROT_RW,
                UK_VMA_MAP_SIZE_2MB, NULL);
```

## Example 4
Create a fixed VMA for a mapping already existing in the pagetable (e.g.,
for an initrd). Note that the attributes specified must match the configuration
in the page table. Otherwise, behavior is undefined.
```C
uk_vma_reserve_ex(uk_vas_get_active(), <VADDR>, <SIZE>,
                  PAGE_ATTR_PROT_READ, 0, ".initrd");
```

## Example 5
Reserve a 1GB address range, create an initial 1MB mapping, extend the
mapping by 2MB, and shrink the mapping by 1MB again. The reserved range
acts as a placeholder which can be only be used for other mappings by
explicitly providing a virtual address of the range and specifying the
UK_VMA_MAP_REPLACE flag. The flag is also used to replace part of the
heap mapping with a reservation again. The figures to the right depict
the VMAs after each operation.
```C
__vaddr_t vaddr = HEAP_BASE;                          ┌────────┐0MB
uk_vma_reserve(uk_vas_get_active(),                   │        │
               &vaddr, 0x40000000);                   │  RSVD  │
                                                      │        │
                                                      │        │
                                                      └ ─ ─ ─ ─┘
// Initial 1MB mapping                                ┌────────┐0MB
vaddr = HEAP_BASE;                                    │  HEAP  │
uk_vma_map_anon(uk_vas_get_active(),                  ├────────┤1MB
           &vaddr, 0x100000, PAGE_ATTR_PROT_RW,       │  RSVD  │
           UK_VMA_MAP_REPLACE, "HEAP");               │        │
                                                      └ ─ ─ ─ ─┘
// Grow mapping by 2MB                                ┌────────┐0MB
vaddr += 0x100000;                                    │/  /  / │
uk_vma_map_anon(uk_vas_get_active(),                  │  HEAP  │
           &vaddr, 0x200000, PAGE_ATTR_PROT_RW,       │ /  /  /│
           UK_VMA_MAP_REPLACE, "HEAP");               ├────────┤3MB
                                                      └ ─ ─ ─ ─┘
// Shrink mapping by 1MB                              ┌────────┐0MB
vaddr += 0x100000;                                    │/  /  / │
uk_vma_reserve_ex(uk_vas_get_active(),                │  HEAP  │
             &vaddr, 0x100000, 0,                     ├────────┤2MB
             UK_VMA_MAP_REPLACE, NULL);               │  RSVD  │
                                                      └ ─ ─ ─ ─┘
```

## Example 6
It is also possible to just create a large virtual memory area that uses
demand paging and only release the physical memory when not needed anymore.
Thus keeping the VMA intact and accessible.
```C
vaddr = HEAP_BASE;
uk_vma_map_anon(uk_vas_get_active(), &vaddr, 0x300000, PAGE_ATTR_PROT_RW, 0,
                "HEAP");

/* ... heap manager decides to release some physical memory in an unused area */
vaddr = HEAP_BASE + 0x200000;
uk_vma_advise(uk_vas_get_active(), vaddr, 0x100000, UK_VMA_ADV_DONTNEED, 0);
```

## Example 7
Create a linear ring buffer that mirrors the buffer at the end to avoid
copying (see https://en.wikipedia.org/wiki/Circular_buffer#Optimization).
Note: Precede an address reservation for the whole 2 * PAGE_SIZE * <PAGES>
range to ensure race free allocation of the virtual addresses.
Note: Physical memory mapped with uk_vma_map_dma() will not be freed
automatically when the VMA is unmapped.
```C
struct uk_vas *vas = uk_vas_get_active();
__paddr_t paddr = uk_falloc(vas->pt->fa, <PAGES>);
__vaddr_t vaddr = __VADDR_ANY;
uk_vma_map_dma(vas, &vaddr, PAGE_SIZE * <PAGES>, PAGE_ATTR_PROT_RW,
               UK_VMA_MAP_POPULATE, NULL, paddr);
vaddr += PAGE_SIZE * <PAGES>;
uk_vma_map_dma(vas, &vaddr, PAGE_SIZE * <PAGES>, PAGE_ATTR_PROT_RW,
               UK_VMA_MAP_POPULATE, NULL, paddr);
```

## Changes to Paging API
Prototyping some bits and pieces of the implementation showed that some changes to the paging code are needed:
- Make `ukplat_page_set_attr()` ignore ranges that are not mapped to allow a simple call for a whole VMA irrespective of what ranges of a VMA are actually represented in the page table due to demand-paging
- Introduce `ukplat_page_kmap()` and `ukplat_page_kunmap()` to perform a quick temporary mapping of an arbitrary physical frame for initialization of a page in the page fault handlers of `lib/ukvmem`. These mapping will not actually use new mappings if a direct-mapped physical memory exists. In this case, only the corresponding virtual address can be returned with the `unmap` being a noop.
- Replace `ukplat_page_map()` with `ukplat_page_mapx()` that allows to execute a function in before the new PTE is written. This way, it is possible to change the PTE or initialize the physical memory. It also allows allocating the physical memory in a different way (e.g., using a frame from a cache). For compatibility there is a ukplat_page_map() wrapper.


marcrittinghaus:
PR is ready for review 🏎️

include/uk/arch/paging.h
```diff
@@ -160,6 +160,19 @@ __paddr_t PT_Lx_PTE_PADDR(__pte_t pte, unsigned int lvl);
...
+/**
+ * PT_Lx_PTE_SET_PADDR(pte, lvl, paddr)
+ *
+ * @param pte a page table entry from a page table at the given level
+ * @param lvl a page table level [0..PT_LEVELS - 1]
+ * @param paddr the physical address which to set in the PTE
+ *
+ * @return the PTE with the updated physical address
+ */
+#ifndef PT_Lx_PTE_SET_PADDR
```
michpappas, line 171: This has to be inside `!__ASSEMBLY__`  as `__pte_t` causes an undefined mnemonic compiler error when `paging.h` is included by `kvm/arm/bpt64.s`.

plat/kvm/Makefile.uk
```diff
@@ -45,7 +45,7 @@ endif
 ## Architecture library definitions for x86_64
 ##
 ifeq ($(CONFIG_PAGING),y)
-LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(UK_PLAT_COMMON_BASE)/paging.c|common
+LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(UK_PLAT_COMMON_BASE)/paging.c|isr
```
jschlumpp, line 48: Should this also happen for ARM?

arch/x86/x86_64/include/uk/asm/paging.h
```diff
@@ -187,6 +194,16 @@ static inline int ukarch_vaddr_range_isvalid(__vaddr_t start, __vaddr_t end)
...
 #define PAGE_ATTR_PROT_EXEC		0x04 /* Page is executable */
 
+/* Page fault error code bits */
+#define X86_PF_EC_P			0x0001UL /* 0=non-present, 1=prot */
+#define X86_PF_EC_WR			0x0002UL /* 0=read, 1=write */
+#define X86_PF_EC_US			0x0004UL /* 0=kernel, 1=user */
+#define X86_PF_EC_RSVD			0x0008UL /* reserved bit in PTE */
+#define X86_PF_EC_ID			0x0010UL /* instruction fetch */
+#define X86_PF_EC_PK			0x0020UL /* protection key violation */
+#define X86_PF_EC_SS			0x0040UL /* shadow stack access */
```
jschlumpp, line 204:
```suggestion
#define X86_PF_EC_SS			0x0040UL /* shadow stack access */
#define X86_PF_EC_HLAT			0x0080UL /* no translation using HLAT paging */
```
Or is there a reason to that one bit?

include/uk/plat/paging.h
```diff
@@ -161,112 +172,280 @@ int ukplat_pt_clone(struct uk_pagetable *pt, struct uk_pagetable *pt_src,
...
- *   - if __PADDR_ANY flag is set and there are no more free frames;
+ * @param pt
+ *   The page table instance on which to operate
+ * @param vaddr
+ *   The virtual address of the first page in the new mapping
+ * @param paddr
+ *   The physical address of the memory which the virtual region is mapped to.
+ *   This parameter can be __PADDR_ANY to dynamically allocate physical memory
+ *   as needed. Note that, the physical memory might not be contiguous.
+
```
jschlumpp:
```suggestion
 *
```

lib/ukvmem/vmem.c
```diff
@@ -0,0 +1,1106 @@
...
+{
+	UK_ASSERT(vas);
+	UK_ASSERT(pt);
+	UK_ASSERT(a);
+
+	vas->a = a;
+
+#ifdef CONFIG_HAVE_PAGING
+	vas->pt = pt;
+	UK_ASSERT(vas->pt);
```
jschlumpp: There is already a `UK_ASSERT` for `pt`. This is one is superfluous.

lib/ukvmem/vmem.c
```diff
@@ -0,0 +1,1106 @@
...
+{
+	return -EPERM;
+}
+
+int vma_op_unmap(struct uk_vma *vma, __vaddr_t vaddr, __sz len)
+{
+	UK_ASSERT(vaddr >= vma->start);
+	UK_ASSERT(vaddr + len <= vma->end);
+
+	return ukplat_page_unmap(vma->vas->pt, vaddr, len / PAGE_SIZE, 0);
```
jschlumpp, line 444: Just a question: `len % PAGE_SIZE == 0` is always true? Otherwise, the round down behavior might be wrong.

lib/ukvmem/vmem.c
```diff
@@ -0,0 +1,1106 @@
...
+static __vaddr_t vmem_first_fit(struct uk_vas *vas, __vaddr_t base, __sz align,
+				__sz len)
+{
+	__vaddr_t vaddr = base;
+	struct uk_vma *cur;
+
+	UK_ASSERT(vas);
+
+	/* Since we are scanning the VAS for an empty address range, we need
+	 * to be careful not to overflow. Checks are thus active and not just
```
jschlumpp:
```suggestion
	 * to be careful not to overflow. Checks are thus always active and not just
```

lib/ukvmem/vma_stack.c
```diff
@@ -0,0 +1,92 @@
...
+
+		memset_isr((void *)vaddr, 0, PAGE_SIZE);
+		ukplat_page_kunmap(pt, vaddr, 1, 0);
+	}
+
+	fault->paddr = paddr;
+	return 0;
+}
+
+/* We deny splits and merges so that the guard page is preserved */
```
jschlumpp, line 77: Would it be possible to have the guard page as a separate VMA and allow merging of the data pages?

lib/ukvmem/vma_file.c
```diff
@@ -0,0 +1,234 @@
...
+	UK_ASSERT(PAGE_ALIGNED(args->offset));
+
+	/* Shared mappings are not supported */
+	if (*flags & UK_VMA_FILE_SHARED)
+		return -ENOTSUP;
+
+	/* Since we cannot do ISR-safe file accesses in the fault handler,
+	 * we enforce full load at mapping time for now.
+	 *
+	 * TODO: Remove this restriction if possible.
```
jschlumpp, line 53: We may have to implement some kind of uninterruptible sleep for threads and queue a softirq-like task, but that should be possible. (Okay and we have to have some way to prevent going back to the interrupted thread)

lib/ukvmem/vma_file.c
```diff
@@ -0,0 +1,234 @@
...
+	struct uio uio = {
+		.uio_iov = &iovec,
+		.uio_iovcnt = 1,
+		.uio_offset = offset,
+		.uio_resid = len,
+		.uio_rw = UIO_READ,
+	};
+	int rc;
+
+	vn_lock(vp); {
```
jschlumpp: This looks like someone wants to use RAII but has to use C. I would say the style does not match the one of Unikraft.

plat/drivers/virtio/virtio_ring.c
```diff
@@ -411,12 +417,31 @@ struct virtqueue *virtqueue_create(__u16 queue_id, __u16 nr_descs, __u16 align,
...
+
+	rc = uk_vma_map_dma(uk_vas_get_active(), &vaddr, ring_size,
+			    PAGE_ATTR_PROT_RW, UK_VMA_MAP_POPULATE,
+			    "virtqueue", paddr);
+	if (unlikely(rc))
+		goto err_freevq;
+
+	vrq->vring_mem = (void *)vaddr;
+#else /* CONFIG_LIBUKVMEM */
 	if (uk_posix_memalign(a, &vrq->vring_mem,
```
jschlumpp, line 439: I think this is already not safe for `HAVE_PAGING && !LIBUKVMEM`. See the main review comment.

jschlumpp (changes requested):
Some notes:
* `plat/common: Introduce ukplat_page_mapx()`: Typo in the message "[...] ~~does~~ aborts with an error. [...]". 
   Also I'm a bit surprised that the thing is not called UkPlatPageMapEx /j.
* `plat/common: Introduce ukplat_page_k(un)map()`: `In contrast to ukplat_page_mapx(), these functions are safe
to call from an interrupt handler. However, they are not guaranteed
to succeed if for some reason no kernel mapping can be established.` I'm a bit confused by this message. Is `k(un)map` allowed to fail or not? I'm assuming you talk about `mapx`, but the text structure seems wrong in that case.
* `drivers/virtio: Use DMA mapping with ukvmem`: Demand paging makes it easier to break the driver. But paging should be already enough to break it. An `ukalloc` allocation might already return a non-continuous physical region IIRC.

lib/ukvmem/vmem.c
marcrittinghaus: This function should only be used as default operation in which case it is always called by `vmem_vma_unmap()`. This function has extensive length checks. But as a good measure, I added another `UK_ASSERT(PAGE_ALIGNED(len));`.

lib/ukvmem/vma_file.c
marcrittinghaus: You got me 😆 Removed.

jschlumpp (approved):
Amazing work :tada: 

plat/drivers/virtio/virtio_ring.c
marcrittinghaus: Yes, this is correct. This has to be addressed in a future PR as this comes with a set of difficulties:
- We cannot simply allocate memory from the frame allocator because physical memory is allocated (almost) entirely by the heap if we have paging but no ukvmem.
- Without ukvmem we have to select a virtual address somehow for the DMA mapping. However, this can easily lead to conflicts and the solution has to cope with multiple mappings.

Since the chance is very high for the queue being allocated from a contiguous 2MB huge heap page with paging enabled, and since we already have this problem in the current version, I will not address this in this PR. I will open an issue to track this.

lib/ukvmem/vma_file.c
marcrittinghaus: Yes, in the end we should have a solution for this. Maybe, for this simple file mapping implementation, it is the easiest solution to just save the registers in case we need to call into the file system. Or we keep it as-is and do not support on-demand paging with the simple file mapping. I would favor this one.

lib/ukvmem/vma_stack.c
marcrittinghaus: I don't get that. Can you elaborate in which case you want to do this? The stack does not grow indefinitely, so the location of the guard page is well-known.

michpappas:

lib/ukvmem/vma_stack.c
jschlumpp: I misunderstood the merge operation.

PR #616, opened 2022-10-28
Move guest scripts to Unikraft core repository

skuenzer:
This PR moves the scripts `qemu-guest` and `xen-guest` back to the Unikraft core repository. Those scripts are handy for some of the Unikraft kernel development tasks and were borrowed by `kraft`.

Production deployments should not use these scripts. It is highly recommend to use `kraftkit` instead which comes with a native guest management integration: https://github.com/unikraft/kraftkit/

### DIFF
--- a/support/scripts/qemu-guest
+++ b/support/scripts/qemu-guest
@@ -1,0 +1,1041 @@
+#!/bin/bash
+#
+# Authors: Simon Kuenzer <simon.kuenzer@neclab.eu>
+#
+# Copyright (c) 2019, NEC Laboratories Europe GmbH,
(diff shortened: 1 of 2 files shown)


craciunoiuc (commented):
Can you maybe also fix the typo reported by checkpatch? It's good that it was caught. 🙏

craciunoiuc (approved):
Strange, it's gone now. All good here. Thanks!

nderjung (approved):
The scripts we know and love! Back where they belong ;) Thanks!

Approved-by: Alexander Jung <alex@unikraft.io>

PR #618, opened 2022-11-01
plat/drivers: Fix GICv3 redistributor access

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [arm64]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The implementation of the GICv3 driver assumes that the GICR_ registers access each core's local redistributor. This is not true, as the redistributor registers of all are mapped in 64KiB adjascent regions, with GICR_BASE being the offset to the register of the first core. This results into a bug where all cores update the registributor registers of the first core.

Update the GIC_RDIST_REG macro to access the right region for a given core, using the current lcpu index.

### DIFF
--- a/plat/drivers/gic/gic-v3.c
+++ b/plat/drivers/gic/gic-v3.c
@@ -51,7 +51,9 @@
 #include <gic/gic-v3.h>
 #include <ofw/fdt.h>
 
-#define GIC_RDIST_REG(gdev, r) ((void *)(gdev.rdist_mem_addr + (r)))
+#define GIC_RDIST_REG(gdev, r)					\
+	((void *)(gdev.rdist_mem_addr + (r) +			\
+	lcpu_get_current()->idx * GICR_STRIDE))
 
 #define GIC_AFF_TO_ROUTER(aff, mode)				\
 	((((uint64_t)(aff) << 8) & MPIDR_AFF3_MASK) | ((aff) & 0xffffff) | \
--- a/plat/drivers/include/gic/gic-v3.h
+++ b/plat/drivers/include/gic/gic-v3.h
@@ -117,6 +117,7 @@
 #define GICR_TYPER_PROC_NUM_MASK	(0xffff << GICR_TYPER_PROC_NUM_SHIFT)
 
 /* GICR frames offset */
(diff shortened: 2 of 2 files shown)


michpappas:
@razvanvirtan this is the smallest set of changes to get this to work. Notice that, since all redistributor registers are available to all cores, another possibility is to configure the redistributor for all cores from the bootcore, and eliminate the initialization of the redistributor from secondary cores. Yet, that could theoretically impact the boot performance of the bootcore on setups with many cores, so I chose to leave redistributor initialization to each core as it currently is. CC: @marcrittinghaus.

marcrittinghaus:
> @razvanvirtan this is the smallest set of changes to get this to work. Notice that, since all redistributor registers are available to all cores, another possibility is to configure the redistributor for all cores from the bootcore, and eliminate the initialization of the redistributor from secondary cores. Yet, that could theoretically impact the boot performance of the bootcore on setups with many cores, so I chose to leave redistributor initialization to each core as it currently is. CC: @marcrittinghaus.

I like the approach and it makes totally sense. One question: you use a single global variable to store the cpu index. Doesn't this lead to problems with multiple CPUs as the value is overwritten?

As a quick fix you could just call `ukplat_lcpu_idx()` in the macro.

michpappas:
> One question: you use a single global variable to store the cpu index. Doesn't this lead to problems with multiple CPUs as the value is overwritten?
> 
> As a quick fix you could just call ukplat_lcpu_idx() in the macro.

Ooof you're right, looks like I have developed bad habits from working with un-conventional SMP implementations for too long 😅  
Fixed.

marcrittinghaus:
> Ooof you're right, looks like I have developed bad habits from working with un-conventional SMP implementations for too long 😅 Fixed.

Thanks! Looks good now 😄 Maybe @razvanvirtan you have time to give it a shot?

razvanvirtan (approved):
I checked it and also tested with my [GIC-v3 PR](https://github.com/unikraft/unikraft/pull/471), everything looks good.
Thanks a lot for this @michpappas!

marcrittinghaus (approved):
Thanks @michpappas, @razvanvirtan for fixing and testing!

PR #619, opened 2022-11-01
build: Add support for long application directory paths

kubanrob:
Linking of libraries with huge amount of object files can lead to problems if paired with a long path to the application directory:
```
make[2]: execvp: /bin/bash: Argument list too long
```
This PR makes changes that enable reading command line arguments from build tools that support it.

### Base target

 - Architecture(s): 
 - Platform(s):
 - Application(s):

### Additional configuration

- Create a directory with a very long name, for example: `long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-name`
- In this directory, checkout the helloworld app, lib-musl, and unikraft.
- Configure app-helloworld to build with musl
- BUILD: fails with `make[2]: execvp: /bin/bash: Argument list too long`
- Reconfigure to enable the config build option `BUILD_ARGS_FILE` by checking `Build Options  ---> Use file for long build arguments`
- REBUILD: application is build as expected
  
### Description of changes

This PR adds a macro that writes potentially long build arguments into a file in order to avoid running into argument size limits. This behavior can be controlled by the config option `BUILD_ARGS_FILE`.
The build tool can then read the arguments from the file, which is an option for most tools in the gcc toolchain.

The PR includes a commit to implement the basic functionallity for `build_cmd`, and a second one enabling it for `build_cmd_fixdep` and making changes to the build rules.

### DIFF
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -252,21 +252,30 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 verbose_cmd_inner = printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 endif
 
+# Writes arguments into a build_args file
+#
+# write_build_args $target,$longargs(optional)
+ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
+write_build_args = $(if $(2),$(file > $(addsuffix .build_args,$(1)),$(2))@$(addsuffix .build_args,$(1)),)
+else
+write_build_args = $(2)
+endif
+
 # Calls a command that creates an object
 #
-# build_cmd $quietlabel,$libname(optional),$target,$command
+# build_cmd $quietlabel,$libname(optional),$target,$command,$longargs(optional)
 ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(4))
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(4) $(call write_build_args,$(3),$(5)))
 endef
 else
 ifeq ($(CONFIG_RECORD_BUILDTIME_LIFTOFF),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(LIFTOFF) $(LITFOFFFLAGS) -o $(addsuffix .liftoff,$(3)) -- $(4))
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(LIFTOFF) $(LITFOFFFLAGS) -o $(addsuffix .liftoff,$(3)) -- $(4) $(call write_build_args,$(3),$(5)))
 endef
 else
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(4))
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(4) $(call write_build_args,$(3),$(5)))
 endef
 endif
 endif
@@ -289,14 +298,15 @@ endef
 # Returns a list of files to be cleaned when build_cmd was used
 #
 # build_clean $target
-ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
-build_clean = $(1) $(addsuffix .time,$(1))
-else
-ifeq ($(CONFIG_RECORD_BUILDTIME_LIFTOFF),y)
-build_clean = $(1) $(addsuffix .liftoff,$(1))
-else
 build_clean = $(1)
+ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
+build_clean += $(addsuffix .time,$(1))
 endif
+ifeq ($(CONFIG_RECORD_BUILDTIME_LIFTOFF),y)
+build_clean += $(addsuffix .liftoff,$(1))
+endif
+ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
+build_clean += $(addsuffix .build_args,$(1))
 endif
 
 # Helper that generates a command for validating a checksum for
@@ -857,7 +867,8 @@ $(call libname2preolib,$(1)): $($(call vprefix_lib,$(1),OBJS)) \
 			      $($(call vprefix_lib,$(1),DTB)) \
 			      $($(call vprefix_lib,$(1),DTB-y))
 	$(call build_cmd,LD,,$(call libname2preolib,$(1)),\
-		$(LD) $(LIBLDFLAGS) $(LIBLDFLAGS-y) \
+		$(LD),\
+		      $(LIBLDFLAGS) $(LIBLDFLAGS-y) \
 		      $($(call vprefix_lib,$(1),LDFLAGS)) \
 		      $($(call vprefix_lib,$(1),LDFLAGS-y)) \
 		      $($(call vprefix_lib,$(1),OBJS)) \
--- a/Config.uk
+++ b/Config.uk
@@ -180,6 +180,12 @@ if RECORD_BUILDTIME
 	endchoice
 endif
 
+config BUILD_ARGS_FILE
+	bool "Use file for long build arguments"
+	default n
+	help
+		Uses a file to pass long arguments to the build commands.
+
 config CROSS_COMPILE
 	string "Custom cross-compiler tool prefix (optional)"
 	help


craciunoiuc (commented):
Hey @kubanrob 

I tried it out and it works just fine.

Personally, I would prefer `.args` instead of `.build_args`, but that's up to you.

Let's see what others say.

support/build/Makefile.rules
```diff
@@ -857,7 +867,8 @@ $(call libname2preolib,$(1)): $($(call vprefix_lib,$(1),OBJS)) \
 			      $($(call vprefix_lib,$(1),DTB)) \
 			      $($(call vprefix_lib,$(1),DTB-y))
 	$(call build_cmd,LD,,$(call libname2preolib,$(1)),\
-		$(LD) $(LIBLDFLAGS) $(LIBLDFLAGS-y) \
+		$(LD),\
```
vladandrew, line 870: Hi,
There seems to be a typo here, what's with the extra `,` ?

support/build/Makefile.rules
```diff
@@ -252,21 +252,30 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 verbose_cmd_inner = printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 endif
 
+# Writes arguments into a build_args file
+#
+# write_build_args $target,$longargs(optional)
+ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
+write_build_args = $(if $(2),$(file > $(addsuffix .build_args,$(1)),$(2))@$(addsuffix .build_args,$(1)),)
+else
+write_build_args = $(2)
```
vladandrew, line 261: Why do we use $(2) here?

support/build/Makefile.rules
```diff
@@ -483,19 +483,19 @@ endef
...
-		       $$(ASINCLUDES) $$(ASINCLUDES-y) \
-		       $$($(call vprefix_lib,$(1),ASINCLUDES)) $$($(call vprefix_lib,$(1),ASINCLUDES-y)) \
-		       $$($(call vprefix_src,$(1),$(2),$(3),INCLUDES)) $$($(call vprefix_src,$(1),$(2),$(3),INCLUDES-y)) \
-		       $$($(call vprefix_glb,$(3),ARCHFLAGS)) $$($(call vprefix_glb,$(3),ARCHFLAGS-y)) \
-		       $$(ASFLAGS) $$(ASFLAGS-y) $$(ASFLAGS_EXTRA) \
-		       $$($(call vprefix_lib,$(1),ASFLAGS)) $$($(call vprefix_lib,$(1),ASFLAGS-y)) \
-		       $$($(call vprefix_src,$(1),$(2),$(3),FLAGS)) $$($(call vprefix_src,$(1),$(2),$(3),FLAGS-y)) \
-		       $(5) \
-		       $$(DBGFLAGS) $$(DBGFLAGS-y) \
+		$(AS)  $(5) \
```
vladandrew: Is there an advantage to moving this at the beginning and not leave it where it previously was?

support/build/Makefile.rules
```diff
@@ -510,19 +510,19 @@ buildrule_sx = $(call buildrule_S_,$(1),$(2),$(3),$(4))
...
```
vladandrew: Same question as before.

support/build/Makefile.rules
```diff
@@ -483,19 +483,19 @@ endef
...
 		       -D__LIBNAME__=$(1) -D__BASENAME__=$(notdir $(2)) $(if $(3),-D__VARIANT__=$(3)) \
-		       -c $(2) -o $(4) $(call depflags,$(4))
+		       -c $(2) -o $(4) $(call depflags,$(4)), \
```
vladandrew: There is an extra `,` at the end here, is it necessary?

support/build/Makefile.rules
```diff
@@ -618,16 +618,17 @@ endef
...
-		       $$($(call vprefix_lib,$(1),ASINCLUDES)) $$($(call vprefix_lib,$(1),ASINCLUDES-y)) \
-		       $$($(call vprefix_src,$(1),$(2),$(3),INCLUDES)) $$($(call vprefix_src,$(1),$(2),$(3),INCLUDES-y)) \
-		       $$(ARCHFLAGS) $$(ARCHFLAGS-y) \
-		       $$(ASFLAGS) $$(ASFLAGS-y) $$(ASFLAGS_EXTRA) \
-		       $$($(call vprefix_lib,$(1),ASFLAGS)) $$($(call vprefix_lib,$(1),ASFLAGS-y)) \
-		       $$($(call vprefix_src,$(1),$(2),$(3),FLAGS)) $$($(call vprefix_src,$(1),$(2),$(3),FLAGS-y)) \
+		$(AS)  -E -P -x assembler-with-cpp \
 		       $(5) \
-		       $(2) -o $(4) $(call depflags,$(4))
+		       $(2) -o $(4) $(call depflags,$(4)), \
```
vladandrew: Same question about the `,`. I don't think we need to reformat all of this.

vladandrew (changes requested):
Hi Robert,

Thank you for this work! There are a few comments from me and I'm a bit concerned about the extra formatting changes, are they really necessary? I'd keep this patch as small as possible since it's touching the build system.

support/build/Makefile.rules
kubanrob: The comma seperates the `$command` from the new optional `$longargs` argument. Everything in longargs will be written into the built_args file if the the config option `BUILD_ARGS_FILE` is enabled and replaced with a reference to that file.

support/build/Makefile.rules
kubanrob: If `BUILD_ARGS_FILE` is disabled, the `$longargs` will not be written into the file but directly appended to the argument. So `write_build_args` expands to either the name of the file where `$longargs` is written too (`BUILD_ARGS_FILE`) or directly the `$longargs` (`!BUILD_ARGS_FILE`).

support/build/Makefile.rules
kubanrob: See https://github.com/unikraft/unikraft/pull/619/files#r1031575279

support/build/Makefile.rules
kubanrob: See https://github.com/unikraft/unikraft/pull/619/files#r1031575279

support/build/Makefile.rules
kubanrob: See https://github.com/unikraft/unikraft/pull/619/files#r1031575279

support/build/Makefile.rules
kubanrob: So the arguments in these have 2 variants in the original: single-dollar (`$(5)`) and double-dollar (`$$(COMPFLAGS)`). The single-dollar ones will be expanded by make~~and their contend will also passed to fixdep~~. The double-dollar will not be expanded and only the name (`$(COMPFLAGS)`) will be passed to `build_cmd_fixdep` ~~and eventually fixdepth~~.

In the original code, these variables will be resolved before invoking bash (or by bash, I never exactly figured it out). They will not by resolved before writing them into the build_args file.
EDIT: Running `make -n` confirmed to me that the double-dollar variable will be expanded before execution. This means the second commit (a4a6def9ddf11a6d631d76e50eb9a7a47a4be5e6) may break fixdep (because now the `$longargs` are omitted). On the other hand, that means that there is the same error waiting to happen for these rules if, for example, the number of include directories increases.

These  double-dollar variables are exactly the one we want in the build_args file, so we need them to be resolved and put to the new `$longarg` argument of `build_cmd[_fixdepth]`.

kubanrob:
As a note: The first commit (554044695719e43ac55cb70d9268e06872920b6c) is, as far as I know, enough to fix the current problems, as it shows when linking the lib-musl preolib.

I can also put the change to `build_cmd_fixdep` into a separate commit so we can drop the changes to the buildrules if these are to risky.

support/build/Makefile.rules
vladandrew: Makes sense now, thanks!!

support/build/Makefile.rules
kubanrob: I found this, which introduced the secondary expansion: https://github.com/unikraft/unikraft/commit/e48f9653922679cb6ae230faa90c22f989a0bbbb

I am not quite sure why this is needed.

kubanrob:
After talk with @skuenzer, removing the the double secondary expansion might lead to problems in some cases. This needs to be investigated.
I will remove the second commit here and will open a new draft pr for it.

The first commit should be enough to resolve the issue for now.

vladandrew:
Sounds good to me! Thanks!

support/build/Makefile.rules
vladandrew: I remember there was some talk about the need for secondary expansions, Simon should know more.

support/build/Makefile.rules
vladandrew: Makes sense. Thanks!

razvand:
@vladandrew , @craciunoiuc , please see if all is OK to approve this PR.

support/build/Makefile.rules
kubanrob: I removed the second commit now and will open a draft PR for that. The first one is enough for now.

craciunoiuc (changes requested):
@kubanrob some small things from me and then we can merge

"The tool can then read the `argumants` from the file" -> `arguments` in the commit message
"It should be possible to `extent`" -> `extend` in the commit messaged

kubanrob:
@craciunoiuc Thank you for finding these typos.

craciunoiuc (approved):
All good here. Thanks!

vladandrew:
Looks good to me.

vladandrew (approved):
Looks good to me. Thanks!

razvand (approved):

PR #622, opened 2022-11-03
lib/nolibc: Expose rmdir/rename from nolibc

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBVFSCORE=y`

### Description of changes

These two functions were not exposed by the nolibc library but are implemented by vfscore.

### DIFF
--- a/lib/nolibc/include/stdio.h
+++ b/lib/nolibc/include/stdio.h
@@ -86,6 +86,10 @@ void psignal(int sig, const char *s);
 int fputc(int _c, FILE *fp);
 int putchar(int c);
 
+#if CONFIG_LIBVFSCORE
+int rename(const char *oldpath, const char *newpath);
+#endif
+
 #ifdef __STDIO_H_DEFINED_va_list
 #undef va_list
 #endif
--- a/lib/nolibc/include/unistd.h
+++ b/lib/nolibc/include/unistd.h
@@ -91,6 +91,7 @@ int dup2(int oldfd, int newfd);
 int dup3(int oldfd, int newfd, int flags);
 int unlink(const char *pathname);
 off_t lseek(int fd, off_t offset, int whence);
+int rmdir(const char *pathname);
 #endif
 
 #if CONFIG_LIBUKSIGNAL


razvand (commented):
Hi, @mschlumpp . Thanks. What was your use case for this? What did you use to test it?

jschlumpp:
The use case was to test the ported file system code and these functions weren't provided by the modified headers.

mogasergiu:
Hi, @mschlumpp . Did you also enable syscall_shim's libc wrappers? These are supposed to be syscalls and, AFAICT, they are registered and their symbols should have already been visible.

jschlumpp:
@mogasergiu Isn't that inconsistent with functions like `close`, `unlink`, ...? What are the cases that should be only handled by syscall_shim and what are the cases declaring them in the header?

mogasergiu:
> @mogasergiu Isn't that inconsistent with functions like `close`, `unlink`, ...? What are the cases that should be only handled by syscall_shim and what are the cases declaring them in the header?

I guess I misunderstood what you were trying to do, I thought the system was seeing it as not implemented. So this is just to solve the warning for the implicit declaration? In that case, why is `CONFIG_LIBVFSCORE` needed for `rename`?

jschlumpp:
I don't want to introduce an additional systemcall_shim dependency for an internal project. 
And I only want to declare the function when it's actually available (= vfscore being enabled)

mogasergiu:
> I don't want to introduce an additional systemcall_shim dependency for an internal project. And I only want to declare the function when it's actually available (= vfscore being enabled)

Ah, right. I did not notice there is a `#if CONFIG_LIBVFSCORE` in `unistd.h` as well.

mogasergiu (approved):

razvand:
@mschlumpp, this makes sense.

I've two questions:

1. At some point, we would need to do a deep search through all the internal libraries and make sure that all libc-style functions (such as `rmdir()` / `rename()`) ar part of corresponding headers in `lib/nolibc/`, right? If so, I would open an issue with this.
1. I'm wondering whether it made sense for this PR to have two commits, one for `rmdir()` and one for `rename()`, given they are located in different header files in `lib/nolibc/`. I think it made sense to have a single commit if all missing function declarations were part of it.

jschlumpp:
@razvand 
Regarding 1: Sure I can create a issue for that.
Regarding 2: I don't see the benefit of investing the time to split the commit.

razvand (approved):

PR #623, opened 2022-11-03
lib/vfscore: Add additional flags for compatibility

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBVFSCORE=y`

### Description of changes

This adds some additional flags which can be used for filesystem implementations.

### DIFF
--- a/lib/vfscore/include/vfscore/vnode.h
+++ b/lib/vfscore/include/vfscore/vnode.h
@@ -144,6 +144,8 @@ struct vattr {
 
 #define IO_APPEND	0x0001
 #define IO_SYNC		0x0002
+#define IO_ASYNC	0x0004	/* start the storage I/O asynchronously */
+#define IO_DIRECT	0x0008	/* try to bypass the buffer cache */
 
 /*
  * ARC actions
--- a/lib/vfscore/include/vfscore/mount.h
+++ b/lib/vfscore/include/vfscore/mount.h
@@ -82,6 +82,9 @@ struct mount {
 #ifndef	MNT_ASYNC
 #define	MNT_ASYNC	0x00000040	/* file system written asynchronously */
 #endif
+#ifndef MNT_NOATIME
+#define MNT_NOATIME	0x00000080	/* do not update access times */
+#endif
 
 /*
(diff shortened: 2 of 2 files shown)


razvand (commented):
Hi, @mschlumpp . What was the use case for these flags? I didn't find `IO_ASYNC` nor `IO_DIRECT` in Linux. Where did you adapt the values for them?

jschlumpp:
They are from FreeBSD and are necessary for the ext filesystem driver. The values don’t matter as much because only the symbols are used.

lib/vfscore/include/vfscore/vnode.h
```diff
@@ -144,6 +144,8 @@ struct vattr {
 
 #define IO_APPEND	0x0001
 #define IO_SYNC		0x0002
+#define IO_ASYNC	0x0004
+#define IO_DIRECT	0x0008
```
StefanJum: Please add some short comments on these macros, like the one on the `MNT_NOATIME`.

StefanJum (changes requested):
Only one comment.

Other than that, does it make sense to add the flags if we are not using them at the moment, or should we add them when they will be used.

razvand:
@mschlumpp, the reason I'm concerned about the actual values is in case of a binary compatibility issue. If the symbols / values are only going to be used internally in Unikraft, then it's good.

I understand @StefanJum's comment on keeping this PR open until we have an actual use (i.e. a PR that makes use of these macros). I think this is something that makes sense long term. The addition of a feature, function, macro, should be supported by a use case.

jschlumpp:
The `IO_` constants are only relevant for "kernel"-level interactions. Therefore, ABI compatiblity for userspace applications is not a concern.

razvand:
@StefanJum, we can approve this PR. Please review / approve it, and then I'll do it.

StefanJum (approved):
All good, thanks.

razvand (approved):

PR #625, opened 2022-11-03
support/build: Prefer local include dirs

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

For include directories the preprocessor searches the paths in left-to-right order. In case of conflicting "system" headers, the previous order caused a surprising behaviour: the system headers would be preferred to the library local headers. Using `#include "header.h"` would be the proper solution, but this can be inconvenient when packaging external libraries for unikraft.


vladandrew (approved):
Makes sense that we should prefer the library local headers over the system ones. Thanks Marco!

skuenzer (approved):

PR #627, opened 2022-11-03
lib/nolibc: Improve nolibc compatibility

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBUKSWRAND=y`

### Description of changes

These patches improve the compatibility of nolibc with other libraries and applications by providing additional functions. Most of the newly implemented functions were taken from the musl library.


jschlumpp:
@skuenzer I removed the errno defines.

razvand:
@mschlumpp, could you please rebase this PR?

razvand (approved):

lib/nolibc/stdio.c
```diff
@@ -504,3 +504,26 @@ int putchar(int c)
...
+
+	len = strlen(s);
+
+	if (stream == stdout)
+		ret = ukplat_coutk(s, len);
+	else if (stream == stderr)
+		ret = ukplat_coutd(s, len);
+
+	if ((size_t)ret == len)
+		return fputc('\n', stream);
```
marcrittinghaus: This is not according to the standard. `puts` will print a newline, but `fputs` won't do it (nice! 🥳 ). To avoid two calls to `strlen()` I suggest to not implement `puts` as a call to `fputs()` ☹️

lib/nolibc/syslog.c
```diff
@@ -0,0 +1,139 @@
...
+	if (facility == 0)
+		facility = log_facility;
+	priority &= LOG_PRIMASK;
+
+	/* Forward call */
+	uk_printk(level_map[priority], "[%s] ", facility_to_str(facility));
+	if (priority == LOG_DEBUG)
+		uk_printd(format, ap);
+	else {
+		priority = MIN(priority,
```
marcrittinghaus: The `uk_printk()` must only be used for the `!= LOG_DEBUG` case. Otherwise, it is `uk_printd()`, too.

Also according to the coding conventions, please use `{` `}` for both branches of an `if` in case any of the branches is more than one line.

lib/nolibc/syslog.c
```diff
@@ -0,0 +1,139 @@
...
+			       (int) (sizeof(level_map) / sizeof(*level_map)));
+		priority = MAX(priority, 0);
+		uk_printk(level_map[priority], format, ap);
```
marcrittinghaus, line 138: As a small safety net you could add a `UK_ASSERT()` to make sure the priority value is in the bounds of the `level_map` table.

lib/ukswrand/Config.uk
```diff
@@ -4,6 +4,12 @@ menuconfig LIBUKSWRAND
 	default n
 
 if LIBUKSWRAND
+config LIBUKSWRAND_RANDOM
+	bool "Enable random()/srandom() functions"
+	depends on LIBNOLIBC
+	help
+		Enable POSIX pseudo-random number generators.
+
```
marcrittinghaus: Does it really make sense to put the stdlib `random()` functions into `ukswrand()`? I am not sure. Maybe these should rather belong to `nolibc`, because `ukswrand` merely implements the "kernel PRNG", whereas the `libc` implementation is also something completely separate on Linux. You will also get a different implementation with `musl`/`newlib`.

🤔

lib/nolibc/include/syslog.h
```diff
@@ -0,0 +1,96 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __ASSERT_H__
+#define __ASSERT_H__
```
marcrittinghaus: 👀

lib/nolibc/include/syslog.h
```diff
@@ -0,0 +1,96 @@
...
+#endif
+
+#include "stdarg.h"
+
+#define LOG_CONS   0x01
+#define LOG_NDELAY 0x02
+#define LOG_NOWAIT 0x04
+#define LOG_ODELAY 0x08
+#define LOG_PERROR 0x10
+#define LOG_PID    0x20
```
marcrittinghaus: For optimal pleasure of my eyes, I would use `\t` and align with the rest of the values.

lib/nolibc/include/strings.h
```diff
@@ -0,0 +1,48 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Authors: Marco Schlumpp <marco.schlumpp@gmail.com>
+ *
+ *
```
marcrittinghaus: Remove excess line. Maybe even use the new short header?

lib/nolibc/string.c
```diff
@@ -799,6 +817,18 @@ char *strerror(int errnum)
...
+{
+	char *out = dest;
+
+	out = out + strlen(dest);
+	while (*src != 0)
+		*out++ = *src++;
+
+	*out = 0;
+	return dest;
+}
```
marcrittinghaus: In musl this is implemented as

```C
strcpy(dest + strlen(dest), src);
```
I was just wondering if you are gaining any performance benefit from one or the other. If we would implement an optimized version of strcpy for example, we would benefit here also. But maybe until then, we have an additional function call 🤔

lib/nolibc/include/ctype.h
```diff
@@ -65,6 +65,8 @@ extern const unsigned char _nolibc_ctype[];
 #define isascii(c)	(((unsigned char)(c)) <= 0x7f)
 #define toascii(c)	((int)(((unsigned char)(c)) & 0x7f))
 
+#define isblank(c)	((c) == ' ' || (c) == '\t')
```
marcrittinghaus, line 68: The commit message says `isspace` vs. the definition is `isblank`. Actually, I only know `isspace`.

lib/nolibc/include/nolibc-internal/shareddefs.h
```diff
@@ -203,3 +203,10 @@ typedef unsigned short __sa_family_t;
 typedef __sa_family_t sa_family_t;
 #define __DEFINED_sa_family_t
 #endif
+
+#if defined(__NEED_BSD_TYPES) && !defined(__DEFINED_BSD_TYPES)
+typedef unsigned char u_char;
+typedef unsigned short u_short;
+typedef unsigned int u_int;
+typedef unsigned long u_long;
+#endif
```
marcrittinghaus: This block is missing
```C
#define __DEFINED_BSD_TYPES
```

lib/nolibc/Config.uk
```diff
@@ -12,4 +12,23 @@ if LIBNOLIBC
...
+	menuconfig LIBNOLIBC_TEST
+		bool "Test nolibc"
+		select LIBNOLIBC_TEST_STRSEP if LIBUKTEST_ALL
+		default n
+
+	if LIBNOLIBC_TEST
+	config LIBNOLIBC_TEST_STRSEP
+		bool "test: strsep()"
+		default n
+	endif
```
marcrittinghaus: For libc tests we have lib-libc-test (https://github.com/unikraft/lib-libc-test). So I would only include tests here that challenge functionality that is internal and easier to test directly instead of via the POSIX functions.

marcrittinghaus (changes requested):
Hey @mschlumpp,

thanks for the PR. Looks good already. I have a couple minor things and questions. Please see the individual comments 😃

lib/nolibc/syslog.c
jschlumpp: I'm changing the clamp procedure to be more readable. Then IMO the assert should be unnecessary (It'll just repeat the expression)

lib/ukswrand/Config.uk
jschlumpp: :100: 
I'm not sure why I put it there, maybe it was related to newlibc.

lib/nolibc/string.c
jschlumpp: Let's go with the musl variant. It's more maintainable and I would assume that a performance-conscious person would use a well optimized libc such as musl anyway.

lib/nolibc/stdlib.c
```diff
@@ -431,3 +431,9 @@ void abort(void)
 	uk_pr_crit("Abnormal termination!\n");
 	ukplat_crash();
 }
+
+void exit(int status)
+{
+	uk_pr_info("exit called with status %d, halting system\n", status);
+	ukplat_terminate(status);
+}
```
marcrittinghaus, line 440: This function is also defined as part of the syscall definition of `exit` in `posix-process`. Either we guard it depending on the presence of `posix-process` or we remove it.

marcrittinghaus (changes requested):
Hi @mschlumpp,

looks good now. I just found a small issue with `exit` being redefined if `posix-process` is enabled.

marcrittinghaus (approved):

PR #636, opened 2022-11-04
lib/vfscore: Replace `access` and `stat` with unikraft syscalls

mariasfiraiala:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): `AArch64`
 - Platform(s): `kvm`
 - Application(s): `SQLite`, `redis`

### Description of changes

Current `access` and `stat` calls from `vfscore` generate infinite loops when used with `musl` due to these calls being interpreted as library function calls.

Consequently, we shouldn't be relying on calling wrappers when inside `unikraft` core, but on our own "syscalls". For the `access` function case, that would be `uk_syscall_r_access`, for the `stat` family function that would be `uk_syscall_r_stat`, `uk_syscall_r_fstat`, `uk_syscall_r_lstat` and this PR changes them accordingly.

Closes: #635 

Signed-off-by: Maria Sfiraiala <maria.sfiraiala@gmail.com>

PR #638, opened 2022-11-05
lib/vfscore: Change `uk_syscall_e_open()` calls to `uk_syscall_r_open()`

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [x86_64, arm]
 - Platform(s): [N/A]
 - Application(s): [app-redis]

### Description of changes

Raw syscalls implementations should always return the raw value. This PR fixes an incorrect calls to `open` in `openat` by replacing the `uk_syscall_e` variant with the `uk_syscall_r` raw implementation.

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -184,15 +184,13 @@ UK_LLSYSCALL_R_DEFINE(int, openat, int, dirfd, const char *, pathname,
 		      int, flags, int, mode)
 {
 	if (pathname[0] == '/' || dirfd == AT_FDCWD) {
-		return uk_syscall_e_open((long int)pathname, flags, mode);
+		return uk_syscall_r_open((long int)pathname, flags, mode);
 	}
 
 	struct vfscore_file *fp;
 	int error = fget(dirfd, &fp);
-	if (error) {
-		errno = error;
-		return -1;
-	}
+	if (error)
+		return -error;
 
 	struct vnode *vp = fp->f_dentry->d_vnode;
 	vn_lock(vp);
@@ -205,12 +203,12 @@ UK_LLSYSCALL_R_DEFINE(int, openat, int, dirfd, const char *, pathname,
 	strlcat(p, "/", PATH_MAX);
 	strlcat(p, pathname, PATH_MAX);
 
-	error = uk_syscall_e_open((long int)p, flags, mode);
+	error = uk_syscall_r_open((long int)p, flags, mode);
 
 	vn_unlock(vp);
 	fdrop(fp);
 
-	return -error;
+	return error;
 }
 
 #if UK_LIBC_SYSCALLS


StefanJum:
Thank you for this @eduardvintila.
Only one comment on the `creat` syscall. It uses the [`UK_SYSCALL_DEFINE` macro](https://github.com/unikraft/unikraft/blob/staging/lib/syscall_shim/include/uk/syscall.h#L194), that is based on [`UK_LLSYSCALL_DEFINE`](https://github.com/unikraft/unikraft/blob/staging/lib/syscall_shim/include/uk/syscall.h#L150), that returns `-1` on error and sets `errno`.
Should we keep using the `_e_` syscall variant there?

eduardvintila:
Hi, @StefanJum. You're totally right, my bad for not catching this. Thanks a lot!

StefanJum (changes requested):
One more thing @eduardvintila, we set `error = uk_syscall_r_open...` [here](https://github.com/unikraft/unikraft/pull/638/files#diff-50257766dd062f03b631cf9789a11b13c936be36b9423bca03ec8277b79264b0R208), but then the return value is `-error` [here](https://github.com/unikraft/unikraft/blob/73c294ee7c7d9707f23e8f03161e6542d1e8bb72/lib/vfscore/main.c#L213), so in case an error occurs when calling `syscall_r_open`, a positive value will be returned, and if the call succeeds then it will return a negative value instead of the new fd.
We should change the return value to `error`.

eduardvintila:
You make a very good point, @StefanJum. It seems that the `open` syscall implementation already returns a negative value on error, which means that the negation applied at `return -error` [here](https://github.com/unikraft/unikraft/blob/73c294ee7c7d9707f23e8f03161e6542d1e8bb72/lib/vfscore/main.c#L213) is indeed redundant and will actually end up returning a negative value for a valid `fd`.

Thanks a lot, I don't know how this has surfaced just now, since I don't think this is tied to whether the call is made to the `syscall_r_open` or the `syscall_e_open` variant in the first place.

StefanJum:
> Thanks a lot, I don't know how this has surfaced just now, since I don't think this is tied to whether the call is made to the `syscall_r_open` or the `syscall_e_open` variant in the first place.

@eduardvintila Yes it's definitely an old error and it's very strange we did not encounter it before, my guess is that since the libc wrapper uses the `uk_syscall_e_openat` variant [here](https://github.com/unikraft/unikraft/blob/56b71f484f49f1231cee708b1a37f27c2ea127f0/lib/vfscore/main.c#L229), `uk_syscall_r_openat` was just (almost) never used. It's still weird since the `_e_` version uses the `_r_` return value.

Taking another look, the implementation of the syscall still looks quite broken, since in case `fget` returns an error it behaves as the `uk_syscall_e_openat` should, it sets `errno` and returns `-1` [here](https://github.com/unikraft/unikraft/blob/56b71f484f49f1231cee708b1a37f27c2ea127f0/lib/vfscore/main.c#L192-L195), which should not be done, it should return the negative error code.

Sorry for the long chain of issues, hopefully this is the last one :)

eduardvintila:
Yes, @StefanJum, I completely agree. It should not even set `errno` altogether. I've noticed a few other syscalls with this pattern, such as `fstatat`. I'll include a fix for `openat`, but fixes for other syscalls in `vfscore` I think should be part of a larger PR (which will probably include changes in calling the raw implementation of syscalls similar to this PR and #636).

StefanJum (approved):
Looks good now @eduardvintila, thank you,
Can you please create an issue stating that there are raw syscall implementations that set `errno` and give some examples like `fstatat`, so we can track the progress on that one?

eduardvintila:
> Looks good now @eduardvintila, thank you, Can you please create an issue stating that there are raw syscall implementations that set `errno` and give some examples like `fstatat`, so we can track the progress on that one?

Will do, @StefanJum. Thank you for your **really** good observations!

razvand (approved):
This works. Thanks, @eduardvintila 

PR #639, opened 2022-11-05
lib/9pfs: Fix 9pfs scatter/gather I/O

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`, `aarch64`]
 - Platform(s): [N/A]
 - Application(s): [`app-sqlite`, `app-redis`]

### Description of changes

This PR properly implements scatter/gather I/O for 9pfs by actually iterating through all the buffers available in the `uio_iov` list, and not just read/write to a single non-empty buffer as the previous implementation did.

This is especially important for gettings apps to work properly with `Musl`, since functions such as `__stdio_read` make use of scatter/gather I/O through `readv`. 

### DIFF
--- a/lib/9pfs/9pfs_vnops.c
+++ b/lib/9pfs/9pfs_vnops.c
@@ -433,7 +433,8 @@ static int uk_9pfs_read(struct vnode *vp, struct vfscore_file *fp,
 	struct uk_9pdev *dev = UK_9PFS_MD(vp->v_mount)->dev;
 	struct uk_9pfid *fid = UK_9PFS_FD(fp)->fid;
 	struct iovec *iov;
-	int rc;
+	int64_t bytes;
+	int i = 0;
 
 	if (vp->v_type == VDIR)
 		return EISDIR;
@@ -443,25 +444,32 @@ static int uk_9pfs_read(struct vnode *vp, struct vfscore_file *fp,
 		return EINVAL;
 	if (uio->uio_offset >= (off_t) vp->v_size)
 		return 0;
-
 	if (!uio->uio_resid)
 		return 0;
 
-	iov = uio->uio_iov;
-	while (!iov->iov_len) {
-		uio->uio_iov++;
-		uio->uio_iovcnt--;
+	while (i < uio->uio_iovcnt) {
+		iov = &uio->uio_iov[i];
+		if (!iov->iov_len) {
+			i++;
+			continue;
+		}
+
+		bytes = uk_9p_read(dev, fid, uio->uio_offset,
+				   iov->iov_len, iov->iov_base);
+		if (unlikely(bytes < 0))
+			return -(int)bytes;
+		if (!bytes)
+			break;
+
+		UK_ASSERT(uio->uio_offset <= __OFF_MAX - bytes);
+		UK_ASSERT(uio->uio_resid >= bytes);
+		UK_ASSERT(iov->iov_len >= (uint64_t)bytes);
+
+		iov->iov_base = (char *)iov->iov_base + bytes;
+		iov->iov_len -= bytes;
+		uio->uio_offset += bytes;
+		uio->uio_resid -= bytes;
 	}
-
-	rc = uk_9p_read(dev, fid, uio->uio_offset,
-			   iov->iov_len, iov->iov_base);
-	if (rc < 0)
-		return -rc;
-
-	iov->iov_base = (char *)iov->iov_base + rc;
-	iov->iov_len -= rc;
-	uio->uio_resid -= rc;
-	uio->uio_offset += rc;
 
 	return 0;
 }
@@ -471,7 +479,8 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
 	struct uk_9pdev *dev = UK_9PFS_MD(vp->v_mount)->dev;
 	struct uk_9pfid *fid;
 	struct iovec *iov;
-	int rc;
+	int64_t bytes;
+	int rc, i = 0;
 
 	if (vp->v_type == VDIR)
 		return EISDIR;
@@ -481,7 +490,7 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
 		return EINVAL;
 	if (uio->uio_offset >= LONG_MAX)
 		return EFBIG;
-	if (uio->uio_resid == 0)
+	if (!uio->uio_resid)
 		return 0;
 
 	if (ioflag & IO_APPEND)
@@ -496,29 +505,31 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
 	if (rc < 0)
 		goto out;
 
-	if (!uio->uio_resid)
-		return 0;
+	while (i < uio->uio_iovcnt) {
+		iov = &uio->uio_iov[i];
+		if (!iov->iov_len) {
+			i++;
+			continue;
+		}
 
-	iov = uio->uio_iov;
-	while (!iov->iov_len) {
-		uio->uio_iov++;
-		uio->uio_iovcnt--;
+		bytes = uk_9p_write(dev, fid, uio->uio_offset,
+				    iov->iov_len, iov->iov_base);
+		if (unlikely(bytes < 0)) {
+			rc = (int)bytes;
+			break;
+		}
+		if (!bytes)
+			break;
+
+		UK_ASSERT(uio->uio_offset <= __OFF_MAX - bytes);
+		UK_ASSERT(uio->uio_resid >= bytes);
+		UK_ASSERT(iov->iov_len >= (uint64_t)bytes);
+
+		iov->iov_base = (char *)iov->iov_base + bytes;
+		iov->iov_len -= bytes;
+		uio->uio_offset += bytes;
+		uio->uio_resid -= bytes;
 	}
-
-	rc = uk_9p_write(dev, fid, uio->uio_offset,
-			    iov->iov_len, iov->iov_base);
-	if (rc < 0)
-		return -rc;
-
-	iov->iov_base = (char *)iov->iov_base + rc;
-	iov->iov_len -= rc;
-	uio->uio_resid -= rc;
-	uio->uio_offset += rc;
-
-	if (rc < 0)
-		goto out;
-
-	rc = 0;
 
 	/*
 	 * If the uio offset after completion of the write requests is bigger


StefanJum (approved):
Thank you @eduardvintila .The commit makes a lot of sense, but something still seems to be wrong with the scatter/gather list.
I can't get it to write more that 2 `iovec` buffers using `writev / pwritev`.

Again, this commit is fine, the issue also appears when not using this pr, will keep digging.

marcrittinghaus:
@eduardvintila @StefanJum From my understanding `uk_9pfs_read` simply does not process the scatter/gather list so it is not a surprise it does not work. There should be a loop around the `uk_9p_read` that iterates over the `uio->uio_iov` vectors and calls `uk_9p_read` for each one.

Something along the lines of:
```C
int i;

for (i = 0; i < uio->uio_iovcnt; i++) {
	iov = uio->uio_iov[i];
	if (!iov->iov_len)
		continue;
	
	rc = uk_9p_read(dev, fid, uio->uio_offset,
			iov->iov_len, iov->iov_base);
	if (unlikely(rc < 0))
		return -rc;
	
	iov->iov_base = (char *)iov->iov_base + rc;
	iov->iov_len -= rc;
	uio->uio_offset += rc;
	uio->uio_resid -= rc;
}
```
There would also need to be a check that `uio->uio_resid` does not become negative from my perspective as it could be less than the total of all `iov->iov_len` :warning:

marcrittinghaus:
BTW: The code is also problematic because it should make sure that `uio->uio_offset` does not overflow. At least with a `UK_ASSERT(uio->uio_offset <= OFF_MAX - rc);` so we would recognize in a debug build.

eduardvintila:
Hi, @StefanJum and @marcrittinghaus! Indeed, scatter/gather I/O is not fully implemented and currently we just do I/O on only the first non-empty buffer that we find. Thanks a lot for your great suggestion, Marc! Will update my PR soon based on what you have wrote.

eduardvintila:
@StefanJum, I have modified the PR to include @marcrittinghaus's suggestions. Please take another look when you have the time. Thanks!!

StefanJum (approved):
Works fine now, thank you @eduardvintila.

marcrittinghaus (commented):
@eduardvintila, @StefanJum thanks for updating the PR and testing 😃 

Generally, this is much better than what is currently upstream. Although, my suggestion was not meant as an exact solution (...along the lines of). This is because it does not consider the case that `uk_9p_read()`/`uk_9p_write()` do not read/write a full `iov` element. So, actually you would need to check what number of bytes are returned, increment the position on the current `iov` element and proceed to the next one if you reach the end. For `rc == 0` you can probably `break` the loop.

Nevertheless, in case there are other things for the release more pressing, it is ok for now. In that case, the commit message with a small `/* TODO */` and a `UK_ASSERT(rc == 0 || rc == iov->iov_len);` in the code should be added to indicate this.

eduardvintila:
Hey, @marcrittinghaus. Thanks a lot for helping with this! Good point on breaking the loop when `rc == 0`, since that means EOF was reached and there's no need to iterate through the next buffers at that point. 

Also, I have completely missed the fact that, indeed, a `uk_9p_read()`/`uk_9p_write()` call may not process a full `iov` element by returning a number less than the number of bytes requested, which doesn't necessarily guarantee that EOF was actually reached (when doing a `read()`, at least). So we should definitely proceed to the next `iov` element only when the current one was completely processed. Thank you for pointing this out! 🙏🏻 

I have updated my PR accordingly. Really appreciate the help!

lib/9pfs/9pfs_vnops.c
```diff
@@ -447,21 +447,31 @@ static int uk_9pfs_read(struct vnode *vp, struct vfscore_file *fp,
...
+		iov->iov_base = (char *)iov->iov_base + rc;
+		iov->iov_len -= rc;
+		uio->uio_offset += rc;
+		uio->uio_resid -= rc;
+
+		/* Test for file offset overflow */
+		UK_ASSERT(uio->uio_offset <= __OFF_MAX - rc);
+
+		/* Test for negative remaining number of bytes to process */
+		UK_ASSERT(uio->uio_resid >= 0);
```
marcrittinghaus: Asserts testing for overflow/underflow conditions should always test before actually performing the operation. Especially, the first `UK_ASSERT` does not make sense after the fact. Comments are not necessary, from my point of view, as these are obvious checks.

```suggestion
		UK_ASSERT(uio->uio_offset <= __OFF_MAX - rc);
		UK_ASSERT(uio->uio_resid >= rc);
		UK_ASSERT(uio->iov_len >= rc);

		iov->iov_base = (char *)iov->iov_base + rc;
		iov->iov_len -= rc;
		uio->uio_offset += rc;
		uio->uio_resid -= rc;
```

lib/9pfs/9pfs_vnops.c
```diff
@@ -499,25 +509,31 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
...
+		uio->uio_offset += rc;
+		uio->uio_resid -= rc;
 
-	if (rc < 0)
-		goto out;
+		/* Test for file offset overflow */
+		UK_ASSERT(uio->uio_offset <= __OFF_MAX - rc);
 
+		/* Test for negative remaining number of bytes to process */
+		UK_ASSERT(uio->uio_resid >= 0);
```
marcrittinghaus: Same as for read concerning the asserts.

lib/9pfs/9pfs_vnops.c
```diff
@@ -433,7 +433,7 @@ static int uk_9pfs_read(struct vnode *vp, struct vfscore_file *fp,
 	struct uk_9pdev *dev = UK_9PFS_MD(vp->v_mount)->dev;
 	struct uk_9pfid *fid = UK_9PFS_FD(fp)->fid;
 	struct iovec *iov;
-	int rc;
+	int rc, i = 0;
```
marcrittinghaus: Since we are already fixing this, let us also fix the return variable type for `uk_9p_read/write`, which must be an `int64_t`. Since `rc` is also used with `uk_9p_open` in the write variant and there also returned, I would define a different variable `bytes` and use that in the code below. Then we have a matching variable pattern in read+write.
```suggestion
	int64_t bytes;
	int i = 0;
```
In the `<0` case you would do a `return = -(int)bytes;`

lib/9pfs/9pfs_vnops.c
```diff
@@ -471,7 +481,7 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
 	struct uk_9pdev *dev = UK_9PFS_MD(vp->v_mount)->dev;
 	struct uk_9pfid *fid;
 	struct iovec *iov;
-	int rc;
+	int rc, i = 0;
```
marcrittinghaus, line 483:
```suggestion
        int64_t bytes;
	int rc, i = 0;
```

lib/9pfs/9pfs_vnops.c
```diff
@@ -499,25 +509,31 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
 	if (!uio->uio_resid)
 		return 0;
```
marcrittinghaus: This will leak `fid` and must be moved above the `uk_9p_walk` ⚠️

lib/9pfs/9pfs_vnops.c
```diff
@@ -499,25 +509,31 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
...
+		}
 
-	rc = uk_9p_write(dev, fid, uio->uio_offset,
-			    iov->iov_len, iov->iov_base);
-	if (rc < 0)
-		return -rc;
+		rc = uk_9p_write(dev, fid, uio->uio_offset,
+				iov->iov_len, iov->iov_base);
+		if (unlikely(rc < 0))
+			goto out;
```
marcrittinghaus: The file size might have changed already so we need to execute the check before the `out` label. Just set `rc` here and remove the `rc=0` after the loop.
```suggestion
		if (unlikely(bytes < 0)) {
			rc = -(int)bytes;
			break;
		}
```

lib/9pfs/9pfs_vnops.c
```diff
@@ -499,25 +509,31 @@ static int uk_9pfs_write(struct vnode *vp, struct uio *uio, int ioflag)
...
 
-	if (rc < 0)
-		goto out;
+		/* Test for file offset overflow */
+		UK_ASSERT(uio->uio_offset <= __OFF_MAX - rc);
 
+		/* Test for negative remaining number of bytes to process */
+		UK_ASSERT(uio->uio_resid >= 0);
+	}
 	rc = 0;
```
marcrittinghaus: `rc` will be 0 when `uk_9p_open` is successful.
```suggestion
```

marcrittinghaus (changes requested):
Hi @eduardvintila,

Thanks for updating 😃 This looks really good and is much more complete 💪🏻 Looking at the code there are still some small 🐛 left, we should fix.

lib/9pfs/9pfs_vnops.c
eduardvintila: It actually seems that there already is an identical check before the `uk_9p_walk`, at line 494. I will remove this one altogether, since it's redundant.

lib/9pfs/9pfs_vnops.c
eduardvintila: You're totally right, my bad for this. Especially for the first assert.

lib/9pfs/9pfs_vnops.c
eduardvintila: Shouldn't we set `rc` to `(int) bytes` here, since at the end of the function we already negate `rc` with `return -rc`?

lib/9pfs/9pfs_vnops.c
eduardvintila: Good idea, completely agree!

eduardvintila:
Hey, @marcrittinghaus! I've just updated the PR - really great suggestions as always, and my apologies for missing the inconsistencies you have pointed out.

Please let me know if I have missed anything else. Thanks a lot!!

marcrittinghaus (commented):
@eduardvintila Thanks for the quick update and finding the extra `-` 👀 😃 From a style perspective I would align the second line of arguments of `uk_9p_read/write`. Other than that, I think this is ready!

@StefanJum Any more suggestions from your side?

eduardvintila:
> From a style perspective I would align the second line of arguments of `uk_9p_read/write`.

Alignment done :) @marcrittinghaus 


Also, I've casted the `bytes` variable to `uint64_t` in the last assert: `UK_ASSERT(iov->iov_len >= (uint64_t)bytes);`. `bytes` is of type `int64_t` and the compiler would have emitted a warning since the signedness of the operators in the comparison did not match. And I think we can safely cast since we previously check that `bytes` is non-negative

StefanJum:
I'm testing it now, looks great so far :rocket:.

StefanJum (approved):
Works good @eduardvintila, thanks a lot!

marcrittinghaus:
> Works good @eduardvintila, thanks a lot!

Thanks for testing. If you add your tag, I will approve it 👍🏻

StefanJum:
done @marcrittinghaus

marcrittinghaus (approved):
Thanks everyone!

PR #566, opened 2022-10-04
`lib/posix-process`: Implementation of `clone` system call

skuenzer:
This PR implements the `clone` system call which is used by [`lib/musl`](https://github.com/unikraft/lib-musl) for thread creation and which is also an important addition for supporting a Linux binary compatibility mode. The introduced architecture moves handling of particular clone flags to corresponding subsystem libraries so that our separation and configuration flexibility is kept.

This PR depends on PR #565.

PR #565, opened 2022-10-04
`lib/syscall-shim`: Support for clone implementation

skuenzer:
This PR introduces support for implementing system calls that 1) require to know the system call return address, like `clone`, 2) support caller TLSes that are different to Unikraft-internal managed TLS, 3) improve the debug printing options, and 4) implement missed features to map system calls statically with `musl`.

GitHub-Depends-On: #564

PR #642, opened 2022-11-06
lib/posix-process: Add tests

adinasm:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR adds to the `posix-process` library the [tests](https://github.com/dragosargint/test) created by @dragosargint as `uktest` test cases.

This PR depends on PR #566. When testing the functionalities implemented in this PR, `libmusl` must be used with @dragosargint 's [version](https://github.com/unikraft/lib-musl/pull/9).

### Problems
- The test for PR #565 exports  `uk_syscall_r_fork` and `uk_syscall_e_fork`.


lib/posix-process/Makefile.uk
```diff
@@ -16,6 +16,22 @@ LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signals.c
 LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_CLONE) += $(LIBPOSIX_PROCESS_BASE)/clone.c
 LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_CLONE) += $(LIBPOSIX_PROCESS_BASE)/clonetab.ld
 
+ifneq ($(filter y,$(CONFIG_LIBPOSIX_PROCESS_TEST) $(CONFIG_LIBUKTEST_ALL)),)
+	LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_tls.c
+
+	ifeq (x86_64,$(CONFIG_UK_ARCH))
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim_helper.s
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim.c
+		UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS) += fork-0
```
dragosargint, line 27: Although this is guarded, I would still add a comment saying that this is not a real fork

lib/posix-process/exportsyms.uk
```diff
@@ -81,3 +81,5 @@ uk_syscall_r_clone
 uk_syscall_e_clone
 uk_syscall_r_clone3
 uk_syscall_e_clone3
+uk_syscall_r_fork
```
dragosargint, line 86: I wouldn't export these. Placing them here is misleading since someone might think is a real fork. Removing them shouldn't affect the compilation

lib/posix-process/tests/test_pr_566.h
```diff
@@ -0,0 +1,81 @@
...
+#include <uk/print.h>
+#include <uk/syscall.h>
+#include <uk/sched.h>
+
+/*
+ * The red lines with "GOOD:" are not errors
+ * The red lines with "BAD:" are from the devil himself
+ */
+
+#define PRINT_GOOD(...) \
```
dragosargint: You can take these macros out `PRINT_GOOD` and `PRINT_BAD`...the duality of life and code.

dragosargint (commented):
Hi, Adina! Thank you for integrating the tests with uk_test. I left just a few comments.

lib/posix-process/exportsyms.uk
adinasm: It does actually affect the compilation process. When adding `fork` to the provided syscalls in [`Makefile.uk`](https://github.com/unikraft/unikraft/pull/642/files#diff-bef8ca6eca25e3c6b4fd6c3c03a0a968d7cb46d2e57f572fc576b91bdf75656cR27), it also ends up in `build/libsyscall_shim/include/uk/bits/provided_syscalls.h`, where `HAVE_uk_syscall_fork` is defined.

When `uk_syscall_e_fork` and `uk_syscall_r_fork` are not exported, then linking fails. One such error consists of the fact that these 2 cannot be found in [`uk_syscall6`](https://github.com/unikraft/unikraft/blob/staging/lib/syscall_shim/gen_uk_syscall6.awk)  although `HAVE_uk_syscall_fork` is defined.

I also feel that they should not be exported, but right now I cannot think of a better way of implementing this test. Do you have any ideas?

lib/posix-process/tests/test_clone.c
```diff
@@ -0,0 +1,140 @@
...
+	unsigned int no_of_tries = 5;
+	while (no_of_tries--) {
+		if (!__check_if_child_exited(__clone_child_uk_thread_ptr))
+			sleep(1);
+		else
+			break;
+	}
+
+	if (!__check_if_child_exited(__clone_child_uk_thread_ptr)) {
+		PRINT_BAD("VERY BAD the child didn't exit, let's commit suicide");
```
dragosargint: Please take this out as well, because is no longer defined and the compilation will fail.

lib/posix-process/exportsyms.uk
dragosargint: You're right, the compilation fails if we don't export them. In that case we can add a comment like "PLEASE NOTE THIS IS NOT A REAL FORK". You can add them with `#` in exportsyms.uk. Can you please check if this compiles for you as well?

dragosargint (changes requested):

razvand:
@adinasm , please also check the `checkpatch`, Codacy and CodeFactor warnings and fix them.

lib/posix-process/Makefile.uk
```diff
@@ -11,7 +11,26 @@ LIBPOSIX_PROCESS_CFLAGS-y           += $(LIBPOSIX_PROCESS_SUPPRESS_FLAGS-y)
...
+
+	ifeq (x86_64,$(CONFIG_UK_ARCH))
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim_helper.s
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim.c
+
+		# This is not a real fork, it is used only for testing purposes
+		UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS) += fork-0
+	endif
+
+	ifeq(y,$(CONFIG_LIBMUSL))
```
dragosargint: Can you add a space between `ifeq` and `(` here? Otherwise, the make doesn't work. (sorry I didn't add a change suggestion but for some reason, I couldn't)

lib/posix-process/Config.uk
```diff
@@ -2,3 +2,35 @@ config LIBPOSIX_PROCESS
...
+	select LIBPOSIX_PROCESS_PIDS
+
+	config LIBPOSIX_PROCESS_DEBUG
+	bool "Enable debug messages"
+	default n
+
+	config LIBPOSIX_PROCESS_TEST
+	bool "Enable tests"
+	default n
+	select LIBUKTEST
```
dragosargint, line 35: To make the build work for the 2 tests that don't need musl, you should add the following select dependecies:
```
        select LIBSYSCALL_SHIM
        select LIBSYSCALL_SHIM_HANDLER
        select LIBSYSCALL_SHIM_HANDLER_ULTLS
        select LIBPOSIX_TIME
```

lib/posix-process/exportsyms.uk
```diff
@@ -0,0 +1,86 @@
...
+uk_syscall_r_exit_group
+uk_syscall_e_exit_group
+uk_posix_process_create
+uk_posix_process_kill
+clone
+uk_syscall_r_clone
+uk_syscall_e_clone
+uk_syscall_r_clone3
+uk_syscall_e_clone3
+# PLEASE NOTE THIS IS NOT A REAL FORK
```
dragosargint: Can you also say that they are used for testing and refer to the test that uses them?

lib/posix-process/tests/test_pr_566.h
```diff
@@ -0,0 +1,61 @@
+#include <uk/test.h>
```
dragosargint: Can you please rename this header? I'm referring to `test_pr_566.h`

lib/posix-process/tests/test_pthread.c
```diff
@@ -0,0 +1,40 @@
+#include "test_pr_566.h"
```
dragosargint: Also rename here

lib/posix-process/tests/test_syscall_shim.c
```diff
@@ -0,0 +1,79 @@
...
+	ukplat_tlsp_set((unsigned long) TLS_REGION);
+
+	UK_TEST_ASSERTF(((void *) ukplat_tlsp_get() == TLS_REGION),
+		"expected ULTLS 0x%lx to be set to %p", ukplat_tlsp_get(),
+		TLS_REGION);
+
+	is_binary = 1;
+	call_binary_fork();
+
+	UK_TEST_ASSERTF((uk_thread_current()->uktlsp == ktls),
```
dragosargint, line 55: When using musl this will make the program crash, because we don't have a valid TLS. I think we still need to do the checking (lines 49-67) but move all the printing after line 69, when the real TLS is restored.

dragosargint (changes requested):

razvand:
@adinasm , please rebase this PR given the changes introduced by `posix-process`.

tiberiudeak:
Hey! I would make a small addition in Makefile.uk, where for the tests "test_clone.c", "test_detached.c" and "test_pthread.c" I would also check if LIBUKMMAP is enabled, not only if LIBMUSL is (it's true that when LIBMUSL is selected, LIBUKMMAP will also be enabled, but it can be disabled afterwards and the tests won't work).

Also, not specifically related with this PR, when I tested on the ARM64 architecture with the hello-world application, some tests from POSIX-FUTEX failed. Maybe the problem is on my side, but I'll leave a list of the tests that failed here:

- posix_futex_testsuite->test_cmp_requeue_different_val: expected `errno` to be 11 but was -38
- posix_futex_testsuite->test_cmp_requeue_two_waiters_no_requeue: expected `rets[2][0]` to be 1 but was -1
									expected `rets[0][0]` to be 0 but was -1
									expected `var_to_change` to be 1 but was 0
- posix_futex_testsuite->test_cmp_requeue_two_waiters_one_waker: all
- posix_futex_testsuite->test_one_waiter_one_waker: all except three tests
- posix_futex_testsuite->test_two_waiters_one_waker: all except four tests
- posix_futex_testsuite->test_wait_different_value: expected `errno` to be 11 but was -38
- posix_futex_testsuite->test_wait_timeout: expected `errno` to be 110 but was -38
- posix_futex_testsuite->test_wake_no_waiters: expected `ret` to be 0 but was -1

lib/posix-process/Makefile.uk
```diff
@@ -16,6 +16,24 @@ LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_CLONE) += $(LIBPOSIX_PROCESS_BAS
...
+
+	ifeq (x86_64,$(CONFIG_UK_ARCH))
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim_helper.s
+		LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/tests/test_syscall_shim.c
+
+		# This is not a real fork, it is used only for testing purposes
+		UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS) += fork-0
+	endif
+
+	ifeq (y,$(CONFIG_LIBMUSL))
```
tiberiudeak, line 30:
```suggestion
	ifeq (yy,$(CONFIG_LIBMUSL)$(CONFIG_LIBUKMMAP))
```

PR #651, opened 2022-11-19
build/checkpatch: Updated checkpatch configuration and helper script

skuenzer:
This PR updates `.checkpatch.conf` to make it inline with our CI/CD settings. A new helper script `checkpatch.uk` makes it easier to call `checkpatch` from any base directory with the Unikraft settings enabled.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -1,3 +1,5 @@
 --no-tree
 --ignore ASSIGN_IN_IF
 --ignore NEW_TYPEDEFS
+--ignore FILE_PATH_CHANGES
+--ignore OBSOLETE
--- a/support/scripts/checkpatch.uk
+++ b/support/scripts/checkpatch.uk
@@ -1,0 +1,25 @@
+#!/bin/bash
+_SELF=$( readlink -f "$0" )
+_SELF_BASE=$( dirname "${_SELF}" )
+
+if [ -z "${_SELF}" ]; then
+	echo "Failed to detect fully qualified path of '$0'" 1>&2
+	exit 2
+fi
+if [ -z "${_SELF}" -a ! -d "${_SELF_BASE}" ]; then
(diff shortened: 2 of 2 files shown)


craciunoiuc (approved):
I find the `.uk` extension to the script a bit confusing, but it's fine.

All good here. Thanks!

nderjung:
Hi, I think this is an unnecessary abstraction of the `checkpatch.pl` script + as @craciunoiuc says, the `.uk` file is not synonymous with its existing use -- which is Unikraft-centric files.  This is in fact a shell script, so `.sh` makes more sense.  Regardless, I think either the `checkpatch.pl` script itself should be updated to better indicate its use or automatically find the `.checkpatch.conf` file in the context of the Unikraft core repo (which this additional script does).

The additions to `.checkpatch.conf` are, however, fine.

skuenzer:
> Hi, I think this is an unnecessary abstraction of the `checkpatch.pl` script + as @craciunoiuc says, the `.uk` file is not synonymous with its existing use -- which is Unikraft-centric files. This is in fact a shell script, so `.sh` makes more sense. Regardless, I think either the `checkpatch.pl` script itself should be updated to better indicate its use or automatically find the `.checkpatch.conf` file in the context of the Unikraft core repo (which this additional script does).
> 
> The additions to `.checkpatch.conf` are, however, fine.

Normally I'd say the same but the problem with checkpatch is that we want modification as minimal as possible (similar to `libfdt`). Updating the script with the latest upstream version from the Linux tree and keeping our changes is becoming just a bigger headache as more modifications we have. It is a big limitation that checkpatch does not provide any way to load a config file via a parameter because a command alias on your shell could solve the problem. The script that I am introducing makes it not just possible to call it form a subdirectory within the Unikraft core repo, it is also usable on external libraries repos.
The `.uk` ending is indicating that this is the checkpatch version for Unikraft (like `Makefile.uk`).

razvand (approved):

PR #662, opened 2022-11-25
lib/uktest: Change the printing function from printf to _uk_printk

flpostolache:
This commit changes the printing function of uktest from printf _uk_printk. This is needed because with the new musl changes the printf will be dependent of the tcb intialisation. This causes a problem when trying to run uktests before the tcb initialisation, like the sanity tests of uktest.

Signed-off-by: Florin Postolache <florin.postolache80@gmail.com>

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### DIFF
--- a/lib/uktest/include/uk/test.h
+++ b/lib/uktest/include/uk/test.h
@@ -200,10 +200,8 @@
 
 /* Custom print method. */
 #define uk_test_printf(fmt, ...)					\
-	printf("    "							\
-	       LVLC_TESTNAME ":"					\
-	       UK_ANSI_MOD_RESET "\t"					\
-	       fmt, ##__VA_ARGS__)
+	_uk_printk(KLVL_INFO, __NULL, __NULL, 0x0,			\
+	"\t" fmt, ##__VA_ARGS__)
 
 
 /**
--- a/lib/uktest/test.c
+++ b/lib/uktest/test.c
@@ -178,14 +178,14 @@ uk_testsuite_run(struct uk_testsuite *suite)
 #endif /* CONFIG_LIBUKTEST_LOG_STATS */
 
 #ifdef CONFIG_LIBUKTEST_LOG_TESTS
-		printf(LVLC_TESTNAME "test:" UK_ANSI_MOD_RESET
-		       " %s->%s",
-		       suite->name,
-		       esac->name);
+		_uk_printk(KLVL_INFO, __NULL, __NULL, 0x0,
+			(LVLC_TESTNAME "test:" UK_ANSI_MOD_RESET
+			" %s->%s"), suite->name, esac->name);
 		if (esac->desc != NULL)
-			printf(": %s\n", esac->desc);
+			_uk_printk(KLVL_INFO, __NULL, __NULL, 0x0, ": %s\n",
+			esac->desc);
 		else
-			printf("\n");
+			_uk_printk(KLVL_INFO, __NULL, __NULL, 0x0, "\n");
 #endif /* CONFIG_LIBUKTEST_LOG_TESTS */
 
 		esac->func(esac);
--- a/lib/uktest/Config.uk
+++ b/lib/uktest/Config.uk
@@ -3,6 +3,7 @@ menuconfig LIBUKTEST
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
 	select LIBUKDEBUG
+	select LIBUKDEBUG_PRINTK
 
 if LIBUKTEST
 


craciunoiuc (commented):
I see the message format changed a bit and is more verbose now. I'm fine with this. Let's see what @skuenzer says.

Before:
`    :   expected '((void *) 0)' to be 0 and was 0 ....................................... [ PASSED ]`

After:
`[    0.000000] Info: [libuktest] <myself.c @  643> expected '((void *) 0)' to be 0 and was 0 ....................................... [ PASSED ]`

nderjung:
Hi, the original purpose of using `printf` was to make the output of the test appear neater across all test suites and cases.

Before:

image

After:

image

Is there a way to mitigate against the issue described whilst simultaneously allowing us to reduce the noise which provided via the use of `_uk_printk`?  There is no need to print the prefixed information coming from `_uk_printk` since this is already incorporated in the test name. Furthermore, the prefixed information is invalid: notice that the line numbers are incorrect.

flpostolache:
Hi, @nderjung. I've updated the PR. The lib name, source file and line number won't be printed anymore. The timestamp and header can be also removed, but then I'll need to do some changes to the `_vprint` function. ( add a new level just for tests with its own header ) What do you think? Also, @skuenzer what is your opinion on this?

skuenzer:
> Hi, the original purpose of using `printf` was to make the output of the test appear neater across all test suites and cases.
> 
> Before:
> 
> image
> 
> After:
> 
> image
> 
> Is there a way to mitigate against the issue described whilst simultaneously allowing us to reduce the noise which provided via the use of `_uk_printk`? There is no need to print the prefixed information coming from `_uk_printk` since this is already incorporated in the test name. Furthermore, the prefixed information is invalid: notice that the line numbers are incorrect.

Personally, I am fine with this change with the agreement that we are going to fix the print layout with a newer version of `ukdebug` that we are currently anywys working on but it won't make it into this release. The thing is that printing with `printf` can't stay today because it requires a TLS for some proper libc's, like musl. Without TLS, `printf` crashes and this is serious in terms of user experience. `nolibc` wasn't requiring it because it does not implement the while feature set. Because the selftest and other unit tests can run before a TLS is allocated, we need a quick fix which I think this does the job.

nderjung:
Since we are on a tight deadline for the release and after speaking offline about a more general solve to this problem, I will approve this PR.  It should be noted that this is a temporarily solution. The ideal solution involves abstracting the `outc`, `outk` and a very generic "print" function into a new library, `ukprint` (see #665).  The idea is that `ukdebug` and `uktest` should be able to leverage it such that it can print whatever it wants without the restrictions that `_uk_printf` imply (namely, prepending information about the library, file, line no. etc.).

skuenzer:
@maniatro111 : Can you double check that the test titles are also printed with `uk_pr_info`

craciunoiuc (approved):
So, from what I understood, it's fine like this for now. At least on my side it looks good.
image


All good here. Thanks!

flpostolache:
@skuenzer, @craciunoiuc, @skuenzer Please take another look. I also modified the printing of the test titles. Thank you, Simon!

lib/uktest/include/uk/test.h
```diff
@@ -200,10 +200,8 @@
 
 /* Custom print method. */
 #define uk_test_printf(fmt, ...)					\
-	printf("    "							\
-	       LVLC_TESTNAME ":"					\
-	       UK_ANSI_MOD_RESET "\t"					\
-	       fmt, ##__VA_ARGS__)
+	_uk_printk(KLVL_INFO, NULL, NULL,	\
```
skuenzer: maybe keep some of the indention whitespaces for readability purposes. The library name could and should be kept, I think `__STR_LIBNAME__`. If we do not want to show the source filename, you can handover a yero insead of `__LINE__`, otherwise the filename to use is `__STR_BASENAME__`. Preferably you should also use `__NULL` instead of `NULL` to be on the safe side. `NULL` is provided by a libc, `__NULL` is provided by Unikraft types.

lib/uktest/test.c
```diff
@@ -178,14 +178,15 @@ uk_testsuite_run(struct uk_testsuite *suite)
...
-		       " %s->%s",
-		       suite->name,
-		       esac->name);
+		_uk_printk(KLVL_INFO, NULL, NULL, __LINE__,
+			(LVLC_TESTNAME "test:" UK_ANSI_MOD_RESET " %s->%s"),
+			suite->name,
+			esac->name);
 		if (esac->desc != NULL)
-			printf(": %s\n", esac->desc);
+			_uk_printk(KLVL_INFO, NULL, NULL, __LINE__, ": %s\n",
```
skuenzer: Same here:
`NULL` -> `__NULL`
and if you change it at the previous message, also here: `__STR_LIBNAME__`, `__STR_BASENAME__`

skuenzer (changes requested):
Hi @maniatro111, I have just minor cosmetic suggestions. Other from this, I am fine with th efix.

lib/uktest/include/uk/test.h
flpostolache: I don't really understand. The libname and basename should be kept. The line number should return an error. I understood the `__NULL` part. Did I understand correctly?

skuenzer: > I don't really understand. The libname and basename should be kept. The line number should return an error. I understood the `__NULL` part. Did I understand correctly?

I mean, do either this:
```c
#define uk_test_printf(fmt, ...)						\
	_uk_printk(KLVL_INFO, __STR_LIBNAME__, __STR_BASENAME__, __LINE__,	\
			"\t" fmt, ##__VA_ARGS__)
```
or this:
```c
#define uk_test_printf(fmt, ...)					\
	_uk_printk(KLVL_INFO, __NULL, __NULL, 0x0,			\
			"\t" fmt, ##__VA_ARGS__)
```

skuenzer: Sorry for my confusing reply 😉 , it has many options for your preference in case you have a strong preference or point for something.

skuenzer (approved):
Thanks a lot for the changes!

lib/uktest/include/uk/test.h
flpostolache: I think the second option is the better one. This way, all test cases and test suites are aligned. ( Otherwise the libname and source file would mess the alignment and the tests would be harder to read. )

PR #663, opened 2022-11-25
build: Enable build argument file for all build rules

kubanrob:
This PR is a continuation of https://github.com/unikraft/unikraft/pull/619

An big amount of includes directories files can **potentially** lead to problems if paired with a long path to the application directory:
```
make[2]: execvp: /bin/bash: Argument list too long
```
This PR makes changes that enable reading command line arguments from build tools that support it.

### Prerequisite checklist

### Base target

 - Architecture(s):
 - Platform(s):
 - Application(s):

### Additional configuration

- Reconfigure to enable the config build option `BUILD_ARGS_FILE` by checking `Build Options  ---> Use file for long build arguments`

### Description of changes

This PR:
- adds the longargs parameter to `build_cmd_fixdep`
- uses the longargs command to in all relevant buildrules and removes the secondary expansion (e48f9653922679cb6ae230faa90c22f989a0bbbb) as arguments need to by known when they are written into the file

Known problems with possible solutions:
- the longargs parameter is not part of the fixdep command line, which could potentially break fixdepth
   - needs register all changes to the commandline -> maybe add hash of the `build_args` file
- secondary expansion (e48f9653922679cb6ae230faa90c22f989a0bbbb) is removed, which may cause issues
    - look if secondary expansion is necessary
    - write arguments only during secondary expansion if that is possible

Alternatives and their problems:
- use relative paths
   - absolute paths can come from any library, not a central point in the build system
   - libraries could potentially be outside the "common" root directory and still introduce long paths

### DIFF
--- a/Config.uk
+++ b/Config.uk
@@ -180,12 +180,6 @@ if RECORD_BUILDTIME
 	endchoice
 endif
 
-config BUILD_ARGS_FILE
-	bool "Use file for long build arguments"
-	default n
-	help
-		Uses a file to pass long arguments to the build commands.
-
 config CROSS_COMPILE
 	string "Custom cross-compiler tool prefix (optional)"
 	help
--- a/Makefile
+++ b/Makefile
@@ -1077,3 +1077,9 @@ help:
 	@echo ''
 
 endif #umask
+
+# This is used to detect if a makefile macro expansion is immediate or deferred.
+# As the last statement in the main makefile, this is read in the end of the
+# immediate expansion phase. Therefore, UK_DEFERRED_EXPANSION is expanded to a
+# empty string in immediate expansion and to 'y' in a deferred expansion.
+UK_DEFERRED_EXPANSION := y
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -252,30 +252,29 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 verbose_cmd_inner = printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 endif
 
-# Writes arguments into a build_args file
+# Writes to a file during deferred expansion
 #
-# write_build_args $target,$longargs(optional)
-ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
-write_build_args = $(if $(2),$(file > $(addsuffix .build_args,$(1)),$(2))@$(addsuffix .build_args,$(1)),)
-else
-write_build_args = $(2)
-endif
+# write_file_deferred $filepath,$content
+write_file_deferred = $(if $(UK_DEFERRED_EXPANSION),$(file > $(1),$(2)),$$(file > $(1),$(2)))
 
 # Calls a command that creates an object
 #
-# build_cmd $quietlabel,$libname(optional),$target,$command,$longargs(optional)
+# build_cmd $quietlabel,$libname(optional),$target,$command
 ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(4) $(call write_build_args,$(3),$(5)))
+$(call write_file_deferred,$(addsuffix .cmd,$(3)),$(strip $(4)))\
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(SHELL) $(addsuffix .cmd,$(3)))
 endef
 else
 ifeq ($(CONFIG_RECORD_BUILDTIME_LIFTOFF),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(LIFTOFF) $(LITFOFFFLAGS) -o $(addsuffix .liftoff,$(3)) -- $(4) $(call write_build_args,$(3),$(5)))
+$(call write_file_deferred,$(addsuffix .cmd,$(3)),$(strip $(4)))\
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(LIFTOFF) $(LITFOFFFLAGS) -o $(addsuffix .liftoff,$(3)) -- $(SHELL) $(addsuffix .cmd,$(3)))
 endef
 else
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(4) $(call write_build_args,$(3),$(5)))
+$(call write_file_deferred,$(addsuffix .cmd,$(3)),$(strip $(4)))\
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)), $(SHELL) $(addsuffix .cmd,$(3)))
 endef
 endif
 endif
@@ -290,7 +289,7 @@ tmp_depfile = $(dir $1).$(notdir $1).d
 depflags = -Wp$(comma)-MD$(comma)$(call tmp_depfile,$(1))
 define build_cmd_fixdep =
 	$(call build_cmd,$1,$2,$3,$4)
-	$Q $(UK_FIXDEP) $(call tmp_depfile,$3) $3 '$(call strip,$4)' \
+	$Q $(UK_FIXDEP) $(call tmp_depfile,$3) $3 '$(SHELL) $(addsuffix .cmd,$(3))' \
 		> $(call out2dep,$3) && \
 		$(RM) $(call tmp_depfile,$3)
 endef
@@ -298,15 +297,12 @@ endef
 # Returns a list of files to be cleaned when build_cmd was used
 #
 # build_clean $target
-build_clean = $(1)
+build_clean = $(1) $(addsuffix .cmd,$(1))
 ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
 build_clean += $(addsuffix .time,$(1))
 endif
 ifeq ($(CONFIG_RECORD_BUILDTIME_LIFTOFF),y)
 build_clean += $(addsuffix .liftoff,$(1))
-endif
-ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
-build_clean += $(addsuffix .build_args,$(1))
 endif
 
 # Helper that generates a command for validating a checksum for
@@ -886,8 +882,7 @@ $(call libname2preolib,$(1)): $($(call vprefix_lib,$(1),OBJS)) \
 			      $($(call vprefix_lib,$(1),DTB)) \
 			      $($(call vprefix_lib,$(1),DTB-y))
 	$(call build_cmd,LD,,$(call libname2preolib,$(1)),\
-		$(LD),\
-		      $(LIBLDFLAGS) $(LIBLDFLAGS-y) \
+		$(LD) $(LIBLDFLAGS) $(LIBLDFLAGS-y) \
 		      $($(call vprefix_lib,$(1),LDFLAGS)) \
 		      $($(call vprefix_lib,$(1),LDFLAGS-y)) \
 		      $($(call vprefix_lib,$(1),OBJS)) \
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -869,44 +869,44 @@ endef
 
 # buildrule_olib $libname
 define buildrule_olib =
-$(call libname2preolib,$(1)): $($(call vprefix_lib,$(1),OBJS)) \
-			      $($(call vprefix_lib,$(1),OBJS-y)) \
-			      $(EACHOLIB_OBJS) \
-			      $(EACHOLIB_OBJS-y) \
-			      $($(call vprefix_lib,$(1),ALIBS)) \
-			      $($(call vprefix_lib,$(1),ALIBS-y)) \
-			      $(EACHOLIB_ALIBS) \
-			      $(EACHOLIB_ALIBS-y) \
-			      $($(call vprefix_lib,$(1),LDS)) \
-			      $($(call vprefix_lib,$(1),LDS-y)) \
-			      $($(call vprefix_lib,$(1),DTB)) \
-			      $($(call vprefix_lib,$(1),DTB-y))
+$(call libname2preolib,$(1)): $$($(call vprefix_lib,$(1),OBJS)) \
+			      $$($(call vprefix_lib,$(1),OBJS-y)) \
+			      $$(EACHOLIB_OBJS) \
+			      $$(EACHOLIB_OBJS-y) \
+			      $$($(call vprefix_lib,$(1),ALIBS)) \
+			      $$($(call vprefix_lib,$(1),ALIBS-y)) \
+			      $$(EACHOLIB_ALIBS) \
+			      $$(EACHOLIB_ALIBS-y) \
+			      $$($(call vprefix_lib,$(1),LDS)) \
+			      $$($(call vprefix_lib,$(1),LDS-y)) \
+			      $$($(call vprefix_lib,$(1),DTB)) \
+			      $$($(call vprefix_lib,$(1),DTB-y))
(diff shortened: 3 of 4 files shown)


Config.uk
```diff
@@ -180,6 +180,12 @@ if RECORD_BUILDTIME
 	endchoice
 endif
 
+config BUILD_ARGS_FILE
```
skuenzer: I like the idea to write each build cmd in to a `.cmd` file that is then executed. I can imagine that this is also an interesting way to debug things, like checking that includes were correct etc. Because we have the argument list too long problem with `execvp` coming up over and over it is probably better to always create such a cmd file and do not provide an option to switch this behavior on or off.

support/build/Makefile.rules
```diff
@@ -252,21 +252,30 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 verbose_cmd_inner = printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
 endif
 
+# Writes arguments into a build_args file
+#
+# write_build_args $target,$longargs(optional)
+ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
+write_build_args = $(if $(2),$(file > $(addsuffix .build_args,$(1)),$(2))@$(addsuffix .build_args,$(1)),)
```
skuenzer: Do you think it is possible to write down the whole command string (including the command to be executed)? Would this work? If so, I personally would prefer creating a `.cmd` file.

Config.uk
kubanrob: To be honest I wanted this option to be able to turn it of if there are problems I did not consider with this PR, for example if some toolchain does not support a file for passing the arguments.

support/build/Makefile.rules
kubanrob: That should be possible, I thought about it but did not test it. The limit we run in is the one for [a single argument](https://stackoverflow.com/q/64201512), the limit for the complete argument is much bigger, but there is still a limit.

The problem for fixdep and the second expansion are independent from that.

Config.uk
skuenzer: Hum... but since the buildrule is modified, if something would break, it would have broke already or should be actually tested. Every C/C++ etc. file is going through the rules that you modiifed. What I want to say is thhat we would jave notice or would would noitce quickly if something got broken.

support/build/Makefile.rules
skuenzer: Hum, my goal would be to fix this transparently behind the scenes with `build_cmd` (no new argument) plus having a nice way to check later which arguments and command was called to create a file. This could work agnostic to the chosen compiler (so no need to change any buildrule, avoids compatibility problems). If possible, I would keep the prettyfying `echo` command within within the build system but the actual build command in a `.cmd` file. Probably `fixdep` would also go there. Would this fix or avoid the fixdep problem that you mentioned?

Config.uk
kubanrob: The buildrule I modified in PR #619 is only for the linking of the preolibs, it could (in theory) break if somebody uses a linker that does not support reading the arguments from a file. Otherwise I don't see a problem, and even this will go away if we just write out the whole command :+1:
The build rules for C/C++ in this PR are more complicated, this is why I split this from the PR #619.

So in short: I have no problem with removing config option.

support/build/Makefile.rules
kubanrob: > my goal would be to fix this transparently behind the scenes with build_cmd (no new argument) plus having a nice way to check later which arguments and command was called to create a file.

Yes, that should work as long as the commands stay in the (much bigger) limit of the overall argument size.

> Would this fix or avoid the fixdep problem that you mentioned?

Unfortunately not.
`fixdep` gets passed the command to figure out if it changes: `$Q $(UK_FIXDEP) $(call tmp_depfile,$3) $3 '$(call strip,$4)' ...` The 3rd parameter is the whole command as one argument, which will cause the original issue to reappear.
Omitting it will make `fixdep` ignore changes to the commandline. I think passing it a hash of the command instead of the actual command should work, but it is hard to tell if that is a good solution.

kubanrob: To be clear, I haven't run into an actual problem with `fixdep` that I know of.

On the other hand, this argument is there for a reason, and I don't want to break the build system in subtle ways.

Config.uk
skuenzer: Okay, I think we should remove the config option, I am concerned that this option is confusing and a weird workaround to fix some cases (if it stays).

support/build/Makefile.rules
skuenzer: Ah, okay, we are not using that part of fixdep yet - so I assume the problem you mention is currently irrelevant. In truth, we still need to find a way to force recompilation if some compiler arguments change (and not the file itself). This parameter of fixdep could have been helpful for this but maybe `.cmd` files would even be the better way for us.
Do you agree to continue with `cmd` file that contains command and arguments (and thus having it independent of used compiler and tools)?

kubanrob: Yes, I will switch to writing the whole command.

Then what is left if the issue with the double expansion. I struggling to write to the command file only in the second expansion.

support/build/Makefile.rules
```diff
@@ -252,30 +252,24 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
...
-endif
-
 # Calls a command that creates an object
 #
-# build_cmd $quietlabel,$libname(optional),$target,$command,$longargs(optional)
+# build_cmd $quietlabel,$libname(optional),$target,$command
 ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(4) $(call write_build_args,$(3),$(5)))
+$(file > $(addsuffix .cmd,$(3)),$(4))\
```
kubanrob: The following will _not work_, as the file function will not be expanded.
```suggestion
$$(file > $(addsuffix .cmd,$(3)),$(4))\
```

skuenzer: hum, but it is a question now if the double `$` would even make sense here. The thing is that the variables `$(3)` and `$(4)` are derived from `$$()` variables. What is actually the problem that you face with `$$(file`?

skuenzer: Did you see any problems with the second expansion?

kubanrob: This is exactly the problem, the $$() variables will be written into the file unexpanded.

You can try it with the current state of the PR.

support/build/Makefile.rules
skuenzer: :+1:

support/build/Makefile.rules
skuenzer: Okay, I just found this: https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html . Is it possible that we need to do something along these lines?:
```Make
$$(file > $$(addsuffix .cmd,$$(3)),$$(4))\
```

kubanrob: That will leads to outputting `$(file > $(3).cmd,$(4))`. But even if this would be expanded later, outside the context of `build_cmd`, `$(3)` has a different value then.

I think we are missing something:

- If the $$arguments would expanded before expanding `build_cmd`, `$(file` would work.
- If the $$arguments would be expanded after expanding `build_cmd` `$$(file` should work
- If the $$arguments would not be expanded a second time, the the build system would not work.

Running `make -n` I noticed $$ variables (and file) written into `build_cmd` are expanded in some rules and  not in others. I will look into that.

kubanrob: It seem like the second expansion is only happening for rules that are generated by an expansion (`buildrule_*`), but not for (non-generated) rules were `build_cmd` is called only in the recipe.

```
app-helloworld_kvm-arm64
app-helloworld_kvm-arm64.dbg
app-helloworld_kvm-arm64.gz
app-helloworld_kvm-arm64.ld.o
app-helloworld_kvm-arm64.o
provided_syscalls.h
provided_syscalls.h.in.new
syscall_map.h.new
syscall_nrs2.h.new
syscall_nrs.h.new
syscall_r_static.h
syscall_static.h
syscall_stubs.h.new
```

It makes more sense for me now. In generated build rules, the variables should be expanded only when the recipe is expanded, not when the rule itself is generated. This is probably why the $$ variables are there.
I would suspect this is not even a secondary expansion as described in the [manual](https://www.gnu.org/software/make/manual/html_node/Secondary-Expansion.html), but the rule generation is an additional prior extension the $$ variables need to be escaped from.

kubanrob:
@skuenzer I pushed a potential solution.

There are a few open questions:
- As dicussed above, I now removed the potentially to long argument from fixdep, should that be handled differently?
- The buildrule for preolibs does not use $$ variables, but gives the above explanation it should. Should I add an commit to update this buildrule, ignore or leave that for a later PR?
- I needed to add a call to strip on the commandline to remove whitespaces that would break the generated rule. Do you see problems with that?

support/build/Makefile.rules
skuenzer: Hum, it is possible that `$(eval` operations that we have for generating the rules may be the source of confusion. POssible that we need to handle a path with `eval` differently from cases where `build_cmd` is directly invoked.

skuenzer:
> @skuenzer I pushed a potential solution.
> 
> There are a few open questions:
> 
>     * As dicussed above, I now removed the potentially to long argument from fixdep, should that be handled differently?

What do you mean with remove, will you replace it with some placehoilder? fixdep is used for correcting hte header dependency tracking.

> 
>     * The buildrule for preolibs does not use $$ variables, but gives the above explanation it should. Should I add an commit to update this buildrule, ignore or leave that for a later PR?

You can add the `$$` to the preoblibs, I think this is fine. In general is a good idea to have any generated rule using secondary expansion. 👍🏻 

> 
>     * I needed to add a call to strip on the commandline to remove whitespaces that would break the generated rule. Do you see problems with that?

No, I think this is fine. Add it under `Common rules` (`support/build/Makefile.rules`).

kubanrob:
#### fixdep
> What do you mean with remove, will you replace it with some placehoilder? fixdep is used for correcting hte header dependency tracking.

I replaced it with an static string. I think it would be possible to make it depended on the content of the .cmd file (the hashing proposal above). I did not do that yet because of this comment:

> Ah, okay, we are not using that part of fixdep yet

#### preolibs

> You can add the $$ to the preoblibs, I think this is fine. In general is a good idea to have any generated rule using secondary expansion.

Great, will do :+1: 

#### Common rules

> Add it under `Common rules` (`support/build/Makefile.rules`).

I am not sure what you mean, what exactly needs to be moved there?

support/build/Makefile.rules
```diff
@@ -290,24 +289,21 @@ tmp_depfile = $(dir $1).$(notdir $1).d
 depflags = -Wp$(comma)-MD$(comma)$(call tmp_depfile,$(1))
 define build_cmd_fixdep =
 	$(call build_cmd,$1,$2,$3,$4)
-	$Q $(UK_FIXDEP) $(call tmp_depfile,$3) $3 '$(call strip,$4)' \
+	$Q $(UK_FIXDEP) $(call tmp_depfile,$3) $3 'command' \
```
kubanrob: This is the relevant line for the fixdep issue. `$(call strip,$4)` can be to large for a single argument, but for correctness the argument need to change whenever `$(4)` changes.

skuenzer:
> #### fixdep
> 
> > What do you mean with remove, will you replace it with some placehoilder? fixdep is used for correcting hte header dependency tracking.
> 
> 
> 
> I replaced it with an static string. I think it would be possible to make it depended on the content of the .cmd file (the hashing proposal above). I did not do that yet because of this comment:
> 
> 

Sounds like a good idea. You may also use the path to the cmd file

> 
> > Ah, okay, we are not using that part of fixdep yet
> 
> 
> 
> #### preolibs
> 
> 
> 
> > You can add the $$ to the preoblibs, I think this is fine. In general is a good idea to have any generated rule using secondary expansion.
> 
> 
> 
> Great, will do :+1: 
> 
> 
> 
> #### Common rules
> 
> 
> 
> > Add it under `Common rules` (`support/build/Makefile.rules`).
> 
> 
> 
> I am not sure what you mean, what exactly needs to be moved there?

I only meant to place the function for stripping whitespaces at the beginning of Makefile.rules

support/build/Makefile.rules
kubanrob: > POssible that we need to handle a path with eval differently from cases where build_cmd is directly invoked.

Yes, this is exactly what I have done in the  current version (4cea54722fea9e44bb2235383570b4b87ccd1dde).

kubanrob:
> You may also use the path to the cmd file

I will do that.

> I only meant to place the function for stripping whitespaces at the beginning of Makefile.rules

Okay. It's the [buildin strip function](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html). So there is nothing to do there.

support/build/Makefile.rules
```diff
@@ -252,30 +252,29 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
...
 
 # Calls a command that creates an object
 #
-# build_cmd $quietlabel,$libname(optional),$target,$command,$longargs(optional)
+# build_cmd $quietlabel,$libname(optional),$target,$command
 ifeq ($(CONFIG_RECORD_BUILDTIME_TIME),y)
 define build_cmd =
-$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) $(4) $(call write_build_args,$(3),$(5)))
+$(call write_deferred,$(addsuffix .cmd,$(3)),$(strip $(4)))\
+$(call verbose_cmd,$(1),$(if $(2),$(2)':' ,)$(notdir $(3)),$(TIME) $(TIMEFLAGS) -o $(addsuffix .time,$(3)) bash $(addsuffix .cmd,$(3)))
```
skuenzer: can you use $(SHELL) instead of bash? It is defined to be bash in the main Makefile

skuenzer (changes requested):

support/build/Makefile.rules
skuenzer: I run some more test and the PR looks good. If you use $(SHELL) and update the commit message, I can approve it 👍🏻

Makefile
```diff
@@ -1077,3 +1077,7 @@ help:
 	@echo ''
 
 endif #umask
+
+# should be last statement to be expanded immediatly, following should be only
+# deferred expansion of recipies and secondary expansion of dependecies
```
skuenzer: Maybe say that you use this variable in order to detect later in write_deferred if we are called from a secondary expansion or not.

support/build/Makefile.rules
```diff
@@ -252,30 +252,29 @@ verbose_cmd = @printf '  %-7s %s\n' '$(1)'  '$(2)' && $(3)
...
+# writes to a file during deferred expansion
 #
-# write_build_args $target,$longargs(optional)
-ifeq ($(CONFIG_BUILD_ARGS_FILE),y)
-write_build_args = $(if $(2),$(file > $(addsuffix .build_args,$(1)),$(2))@$(addsuffix .build_args,$(1)),)
-else
-write_build_args = $(2)
-endif
+# write_deferred $filepath,$content
+write_deferred = $(if $(UK_DEFERRED_EXPANSION),$(file > $(1),$(2)),$$(file > $(1),$(2)))
```
skuenzer: Just a minor thing: I‘d call it „file_deferred“. What do you think?

skuenzer (commented):
Thanks a lot, I have just minor suggestions. You could maybe also rename the commit title of the one that also reverts the previous PR: Remove the revert part and name it so that always the cmd file is always used: for example:
`build: Always write build command to file`

razvand:
As a contribution from my side, I tested this with [the Musl test scripts](https://github.com/unikraft-upb/scripts/tree/main/musl) and all works OK. I approved the PR. When @skuenzer 's suggestions are integrated, he can approve it and that would be it.

razvand (approved):

support/build/Makefile.rules
kubanrob: The reason why this is called `write_deferred` is that the `file` function can also used to read from a file.  How about `write_file_deferred`?

I don't have a strong opinion on that though, so `file_deferred` would be okay for me.

skuenzer (approved):
Thanks a lot for the update!

PR #666, opened 2022-11-28
lib/posix-socket: Fix returning error codes

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBPOSIX_SOCKET=y`

### Description of changes

For ioctl the error code was not returned at all. Poll had two error checks and the second check was dead code.

### DIFF
--- a/lib/posix-socket/socket_vnops.c
+++ b/lib/posix-socket/socket_vnops.c
@@ -287,7 +287,7 @@ posix_socket_vfscore_ioctl(struct vnode *vnode,
 	if (unlikely(ret < 0)) {
 		PSOCKET_ERR("ioctl on socket %d failed: %d\n", fp->fd,
 			    (int)ret);
-		ret = -ret;
+		return -ret;
 	}
 
 	return 0;
@@ -305,12 +305,10 @@ static int posix_socket_vfscore_poll(struct vnode *vnode, unsigned int *revents,
 	sock = (struct posix_socket_file *)vnode->v_data;
 
 	ret = posix_socket_poll(sock, revents, ecb);
-	if (ret < 0)
-		return -ret;
 	if (unlikely(ret < 0)) {
 		PSOCKET_ERR("poll on socket %d failed: %d\n",
 			    sock->vfs_file->fd, (int)ret);
-		ret = -ret;
+		return -ret;
 	}
 
 	return 0;


craciunoiuc (approved):
All good here. Thanks!

lib/posix-socket/socket_vnops.c
```diff
@@ -290,7 +290,7 @@ posix_socket_vfscore_ioctl(struct vnode *vnode,
 		ret = -ret;
 	}
 
-	return 0;
+	return ret;
```
marcrittinghaus: It only makes sense to allow returning a positive non-zero value from `posix_socket_ioctl` (no error case) if you want to conform to the `ioctl` system call, which says that an ioctl may return a positive value. This would be nice to have, but since vfscore is an absolute mess, it expects positive error values. This is why the return value is negated on error. So, returning a positive non-zero value from `posix_socket_ioctl` is not possible due to the interpretation by vfscore as error value. The function thus simply returns 0. I don't see a reason to change that behavior. Especially, as it also does not comply with the documented behavior of the callback.

However, I agree, that there is a bug in that line `290` should be `return -ret`.

lib/posix-socket/socket_vnops.c
```diff
@@ -305,15 +305,13 @@ static int posix_socket_vfscore_poll(struct vnode *vnode, unsigned int *revents,
...
-	if (ret < 0)
-		return -ret;
 	if (unlikely(ret < 0)) {
 		PSOCKET_ERR("poll on socket %d failed: %d\n",
 			    sock->vfs_file->fd, (int)ret);
 		ret = -ret;
 	}
 
-	return 0;
+	return ret;
```
marcrittinghaus: Same as for `ioctl`. In addition, returning a positive number here also does not bring any benefit because currently any non-zero value is interpreted as error condition in `eventpoll.c`.

lib/posix-socket/socket_vnops.c
```diff
@@ -305,15 +305,13 @@ static int posix_socket_vfscore_poll(struct vnode *vnode, unsigned int *revents,
 	sock = (struct posix_socket_file *)vnode->v_data;
 
 	ret = posix_socket_poll(sock, revents, ecb);
-	if (ret < 0)
-		return -ret;
```
marcrittinghaus, line 309: How could that happen 😱

marcrittinghaus (changes requested):
Hi @mschlumpp,

thanks for pointing out the bugs 🕵🏻‍♂️! Shame on me. However, I do not quite agree with the fix. Please see the comments. Thanks 😃

lib/posix-socket/socket_vnops.c
jschlumpp: Okay, you are right. To prevent unnecessary merge conflicts I'll fix that as part of the fdtab refactoring.

lib/posix-socket/socket_vnops.c
jschlumpp: I think I marked this as resolved a bit too fast. There is still a return in line 311 missing, right? Currently, this ignores all errors.

marcrittinghaus: Yes, for sure. Same bug as in ioctl 😃

lib/posix-socket/socket_vnops.c
marcrittinghaus: Merge conflicts?

jschlumpp: I did quite some extensive changes to vfscore. If I fix the error handling now, I'll have to update the changes I did for fdtab.

marcrittinghaus: Hmm, you probably have to solve merge conflicts / rebase anyways, when you prepare your fdtab work for upstream. Since we have the discussion here, I think it would be better to have it fixed with this PR.

jschlumpp: Okay, fine by me :+1:

marcrittinghaus (approved):
@mschlumpp Thanks! Looks good now 😃

razvand (commented):
@marcrittinghaus , @mschlumpp , I see that `posix_socket_vfscore_read/write` uses the construct `return (int)-ret;`

I would use that for `ioctl` and `poll` as well, for consistency.

jschlumpp:
> I would use that for `ioctl` and `poll` as well, for consistency.

In `read` and `write` the type of `ret` is `ssize_t`, which makes the cast necessary. For ioctl, I don't see a good reason to add an useless cast.

craciunoiuc (approved):
All good here. Thanks!

PR #669, opened 2022-11-28
virtio-net fixes

dinhngtu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

N/A

### Description of changes

This PR proposes a couple small fixes to the `virtio-net` driver:

- A fix for the `VIRTIO_NET_F_MTU` feature flag check;
- A fix for notifying the virtio-net device of available RX buffers upon driver start.

### DIFF
--- a/plat/drivers/virtio/virtio_net.c
+++ b/plat/drivers/virtio/virtio_net.c
@@ -896,14 +896,12 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 	VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_MAC);
 
 	/**
-	 * MTU information (needed)
+	 * MTU information
 	 */
-	if (!VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_STATUS)) {
-		uk_pr_err("%p: Host system does not offer MTU feature\n", n);
-		rc = -EINVAL;
-		goto err_negotiate_feature;
-	}
-	VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_STATUS);
+	if (!VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_MTU))
+		uk_pr_debug("%p: Host system does not offer MTU feature\n", n);
+	else
+		VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_MTU);
 
 	/**
 	 * Gratuitous ARP
@@ -941,6 +939,21 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 				   __offsetof(struct virtio_net_config, mac),
 				   &vndev->hw_addr.addr_bytes[0],
 				   UK_NETDEV_HWADDR_LEN, 1);
+
+	if (VIRTIO_FEATURE_HAS(drv_features, VIRTIO_NET_F_MTU)) {
+		virtio_config_get(vndev->vdev,
+				  __offsetof(struct virtio_net_config, mac),
+				  &vndev->mtu, sizeof(vndev->mtu), 1);
+		vndev->max_mtu = vndev->mtu;
+	} else {
+		/**
+		 * Report some reasonable defaults. This breaks down in cases
+		 * where MTU is <UK_ETH_PAYLOAD_MAXLEN (e.g. encapsulation),
+		 * but there's no way to signal the lack of MTU reporting in
+		 * mtu_get as yet, so we're stuck with this for now.
+		 */
+		vndev->max_mtu = vndev->mtu = UK_ETH_PAYLOAD_MAXLEN;
+	}
 
 	return 0;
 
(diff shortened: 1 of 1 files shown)


jschlumpp (approved):
Thanks! Looks good to me.

eduardvintila (changes requested):
Hello, @dinhngtu. Thanks a lot for publishing this PR. 

Your fix for the `MTU` flag makes a lot of sense given the context, but unfortunately it seems to introduce a dependency on passing the `host_mtu=` parameter to QEMU's `virtio-net` device, for example. Without specifying this, the whole feature negotiation seems to fail and `lwip`  will simply skip this device, resulting in network-based applications such as `httpreply` or `redis` to no longer work properly:
```
ERR:  [libkvmvirtionet] <virtio_net.c @  902> 0x199028: Host system does not offer MTU feature
ERR:  [liblwip] <init.c @  188> Failed to probe features of network device 0: -22; skipping device...
```

However, it seems weird to me that this MTU feature is mandatory (it is even marked as needed in the comment at line 899), even though I've seen no occurrences of it elsewhere in the code that actually uses it. Maybe @skuenzer can help us with this and give us more details?

plat/drivers/virtio/virtio_net.c
```diff
@@ -1139,6 +1139,9 @@ static int virtio_net_start(struct uk_netdev *n)
 	virtio_dev_drv_up(d->vdev);
 	uk_pr_info(DRIVER_NAME": %"__PRIu16" started\n", d->uid);
 
+	for (i = 0; i < d->rx_vqueue_cnt; i++)
+		virtqueue_host_notify(d->rxqs[i].vq);
+
```
Sairajkodilkar, line 1157: Hi @dinhngtu, We notify the device of receive queues when we call the `virtio_netdev_rx_fillup()`, which is called during `virtio_netdev_rx_queue_setup()`. Since this function is always called before `virtio_net_start()`, we don't need to notify the host again.

so you should remove this commit

plat/drivers/virtio/virtio_net.c
```diff
@@ -898,12 +898,12 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
...
 	 * MTU information (needed)
 	 */
-	if (!VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_STATUS)) {
+	if (!VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_MTU)) {
 		uk_pr_err("%p: Host system does not offer MTU feature\n", n);
 		rc = -EINVAL;
 		goto err_negotiate_feature;
 	}
-	VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_STATUS);
+	VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_MTU);
```
Sairajkodilkar: Hi @dinhngtu, I think you should also add the check for `VIRTIO_NET_F_STATUS` . that way, you can check the link status during initialization.
_reference virtio v1.1 specs, section 5.1.4.2_

@eduardvintila I think it's better to have MTU as an optional feature, What do you think?

eduardvintila: Hey, @Sairajkodilkar. Yeah, I agree, I think that should make sense.

plat/drivers/virtio/virtio_net.c
dinhngtu: `virtio_netdev_rx_queue_setup` calls `virtio_netdev_rx_fillup` with `notify=0` which does not send the necessary notifications. In any case, these notifications must be sent later (after `DRIVER_OK`) which is why I've done it here.

plat/drivers/virtio/virtio_net.c
dinhngtu: `VIRTIO_NET_F_MTU` seems necessary since `uk_netdev_ops` has the `mtu_get` operation (but the wiring for MTU reporting doesn't seem to exist ATM). I'll change `virtio_net_mtu_get` to report the actual device MTU.

plat/drivers/virtio/virtio_net.c
Sairajkodilkar: oh, you are right, sorry about that

plat/drivers/virtio/virtio_net.c
Sairajkodilkar: During feature negotiation, you can see whether the MTU is available. If it is available, then read the MTU value; otherwise, assign the default MTU value

eduardvintila (approved):
Hey, @dinhngtu. The fixes look really good to me now, thanks for this! 

razvand (approved):

Sairajkodilkar (approved):
Hi @dinhngtu, Thanks for the fixes. I have tested this PR, and it works without any issues.

Reviewed-by: Sairaj Kodilkar [skodilkar7@gmail.com](mailto:skodilkar7@mail.com)

Issue #596, opened 2022-10-08
`ukrust` does not compile


fabianpatras:
### Describe the bug

Lib `ukrust` makes unikraft not compile.

### Steps to reproduce

# Part I

Steps to reproduce:
1. `git pull git@github.com:unikraft/unikraft.git`
(so we are on `staging` branch)
2. `make menuconfig`
    -> Platform Configuration -> [\*] Linux user space
     -> Library configuration -> [\*] ukrust
     -> Save and exit
3. `make`

Output:
```
$ make
  CP      config
make[2]: *** No rule to make target '/core/src/lib.rs', needed by '/home/fabian/hacktoberfest/my-unikernel/unikraft/build/libukrust/core.o'.  Stop.
make[1]: *** [Makefile:1016: sub-make] Error 2
make: *** [Makefile:32: _all] Error 2
```

# Part 2:
See expected behavior Part I.
Then: 
```
rustup default nightly
rustup component add rust-src
make
```
 

### Expected behavior

# Part I

I expected it to compile.

It seems that `lib/ukrust/Makefile.uk:11` contains a `CONFIG_LIBRUST` which might be a typo. Changing that to `CONFIG_LIBUKRUST` takes us problem number 2. (see steps to reproduce Part 2)

# Part II

Now I expected it to compile, but it seems that the source code could not be compiled. 
Locally, `rustup show` prints:
```
installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)
1.53-x86_64-unknown-linux-gnu

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.66.0-nightly (8b0c05d9a 2022-10-07)
``` 

Is it necessary to build `ukrust` by compiling the Rust standard library ourselves?

### Relevant log output

```shell
# Part I

$ make
  CP      config
make[2]: *** No rule to make target '/core/src/lib.rs', needed by '/home/fabian/hacktoberfest/my-unikernel/unikraft/build/libukrust/core.o'.  Stop.
make[1]: *** [Makefile:1016: sub-make] Error 2
make: *** [Makefile:32: _all] Error 2


# Part II

$ make
   RUSTC   libukrust: core.o
warning: `-Z symbol-mangling-version` is deprecated; use `-C symbol-mangling-version`

warning: unused variable: `index`
[...]

PR #680, opened 2022-12-01
lib/ukrust: Update ukrust to work alongside Cargo

vladandrew:
### Prerequisite checklist

### Base target

### DIFF
--- a/lib/nolibc/string.c
+++ b/lib/nolibc/string.c
@@ -815,3 +815,8 @@ char *strncat(char *dest, const char *src, size_t n)
 	*dest++ = 0;
 	return a;
 }
+
+int bcmp(const void *s1, const void *s2, size_t count)
+{
+	return memcmp(s1, s2, count);
+}
--- a/lib/ukrust/Cargo.toml
+++ b/lib/ukrust/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+license-file = "../../COPYING.md"
+name = "ukrust"
+version = "0.1.0"
+edition = "2018"
+
+[profile.dev]
+panic = "abort"
+lto = "no"
+
+[profile.release]
+panic = "abort"
+lto = "no"
--- a/lib/ukrust/ukrust_sys/lib.rs
+++ b/lib/ukrust/ukrust_sys/lib.rs
@@ -49,3 +49,14 @@ pub mod c_types;
 pub fn alloc_error(_layout: core::alloc::Layout) -> ! {
         panic!("Alloc error");
 }
+
+extern "C" {
+    fn __ukrust_sys_crash() -> !;
+}
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
+    unsafe {
+        __ukrust_sys_crash();
+    }
+}
--- a/lib/nolibc/exportsyms.uk
+++ b/lib/nolibc/exportsyms.uk
@@ -51,6 +51,7 @@ qsort
 
 # string
 memcpy
+bcmp
 memset
 memchr
 memrchr
--- a/lib/ukrust/Makefile.uk
+++ b/lib/ukrust/Makefile.uk
@@ -8,7 +8,7 @@ RUSTCFLAGS-$(CONFIG_LIBUKRUST) += --extern ukrust_sys -L $(LIBUKRUST_BUILD)/ --e
 
 export LIBUKRUST_BINDINGS_FILE=$(LIBUKRUST_BUILD)/bindings_generated.rs
 
-ifeq ($(CONFIG_LIBRUST),y)
+ifeq ($(CONFIG_LIBUKRUST),y)
 # Do not perform these if CONFIG_RUST is not set, as it is not
 # guaranteed that rustc will be available in the system.
 RUSTC_SYSROOT := $(shell $(RUSTC)  --print sysroot)
--- a/lib/ukrust/compiler_builtins.rs
+++ b/lib/ukrust/compiler_builtins.rs
@@ -151,14 +151,3 @@ define_panicking_intrinsics!("`u64` division/modulo should not be used", {
     __aeabi_uldivmod,
     __mulodi4,
 });
-
-extern "C" {
-    fn __ukrust_sys_crash() -> !;
-}
-
-#[panic_handler]
-fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
-    unsafe {
-        __ukrust_sys_crash();
-    }
-}
--- a/lib/ukrust/src
+++ b/lib/ukrust/src
@@ -1,0 +1,1 @@
+ukrust_sys/


Makefile
```diff
@@ -325,6 +325,8 @@ GOCFLAGS :=
 GOCFLAGS-y :=
 RUSTCFLAGS :=
 RUSTCFLAGS-y :=
+CARGOFLAGS :=
+CARGOFLAGS-y :=
```
skuenzer: Do we really need cargo in the core repository? I'd prefer to have it part of an external language support library.

vladandrew: Hi Simon, I'll actually move all the build rules to the external library.  This PR is a late stage draft.

lib/ukrust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "ukrust"
+version = "0.1.0"
+edition = "2018"
```
TrueDoctor, line 5: We should probably also add a license here

fabianpatras:
This PR fixes the typo found in #596.

lib/ukrust/Cargo.toml
vladandrew: Makes sense!

lib/nolibc/string.c
```diff
@@ -815,3 +815,8 @@ char *strncat(char *dest, const char *src, size_t n)
 	*dest++ = 0;
 	return a;
 }
+
+int bcmp(const void *s1, const void *s2, size_t count)
```
skuenzer, line 819: As far as I saw, `bcmp` is an obsolete libc function and `memcmp` should be used instead. Where is `bcmp` still used?

support/build/Makefile.build
```diff
@@ -53,9 +55,11 @@ $(foreach S,$($(call uc,$(L))_SRCS) $($(call uc,$(L))_SRCS-y) \
 	    $(EACHOLIB_SRCS) $(EACHOLIB_SRCS-y), \
 $(eval $(call buildrule_libobj_multitarget,$(L),$(S))) \
 ); \
-$(if $(call qstrip,$($(call uc,$(L))_OBJS) $($(call uc,$(L))_OBJS-y)), \
+$(if $(call qstrip,$($(call uc,$(L))_OBJS) $($(call uc,$(L))_OBJS-y) \
+                   $($(call uc,$(L))_ALIBS) $($(call uc,$(L))_ALIBS-y)), \
 $(eval $(call buildrule_olib,$(L))); \
 $(eval $(call uc,$(P))_OLIBS-y    += $(call libname2olib,$(L))); \
+$(eval UK_ALIBS-y += $($(call uc,$(L))_ALIBS) $($(call uc,$(L))_ALIBS-y)); \
```
skuenzer: Hum... I am not sure if this makes so much sense. If you want to link with a static library at the last step, you would register such a lib to `UK_ALIBS-y` in your `Makefile.uk` (like with `UK_ALIBS-$(CONFIG_MYLIB)+=mylib.a`). Similar to other scopes, it was intended that static libraries that are registered to a library focus are linked before the `exportsyms.uk` is applied. How useful this feature is, i a different question because linking with static libraries work differently (simple form of DCE).

Makefile
```diff
@@ -588,6 +588,11 @@ RUSTC		:= rustc --target=$(CONFIG_LLVM_TARGET_ARCH)
 else
 RUSTC		:= rustc
 endif
+ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
+CARGO		:= cargo --target=$(CONFIG_LLVM_TARGET_ARCH)
```
skuenzer: Hum 🤔  Where is this used? ...and for which reasons do we need cargo in the core repo?

vladandrew: The user should be able to set his cargo path this way. Something like `make CARGO=xcargo or /path/to/cargo`.

support/build/Makefile.build
vladandrew: `UK_ALIBS-$(CONFIG_MYLIB)+=mylib.a)` doing this right now does not work as expected. The alib is imidiatly link to an object file.

lib/nolibc/string.c
vladandrew: rustc generate code that uses bcmp. LLVM 9.0 still emits calls to `bcmp`[1]. I was getting undefined symbol errors when building with cargo.

[1] https://lists.llvm.org/pipermail/llvm-bugs/2019-March/072889.html

support/build/Makefile.build
vladandrew: Actually you're right, this exact line doesn't make sense. I've removed these changes from the pull request. The only thing remaining is to link alibs at the end, not before the exportsyms.uk is applied.

Makefile
skuenzer: I mean, I do not see a build rule using `$(CARGO)`. The thing is, if cargo is used by a external support lib, you should implement this with this library's `Makefile.rules`. We could also add support to load external `Makefile.build`. @marcrittinghaus looked into this.

lib/nolibc/string.c
skuenzer: Hum, okay. Then I wonder two points:
1) Could we introduce an `alias` to `memcmp` instead of defining a new function?
2) I am not yet sure what the best location for the alias definition would be. Do you know if `musl` provides `bcmp`? If yes, it has to be `nolibc`. If not, we may want to place the alias with the rust support library that uses `cargo`.

Ah... and maybe there is a third option where we could handover a built-in option for `bcmp` to LLVM calls?

support/build/Makefile.rules
```diff
@@ -873,10 +873,6 @@ $(call libname2preolib,$(1)): $$($(call vprefix_lib,$(1),OBJS)) \
 			      $$($(call vprefix_lib,$(1),OBJS-y)) \
 			      $$(EACHOLIB_OBJS) \
 			      $$(EACHOLIB_OBJS-y) \
-			      $$($(call vprefix_lib,$(1),ALIBS)) \
-			      $$($(call vprefix_lib,$(1),ALIBS-y)) \
-			      $$(EACHOLIB_ALIBS) \
-			      $$(EACHOLIB_ALIBS-y) \
```
skuenzer: Hum... 🤔 either this commits needs a different description why we remove the ALIB option for libraries or the patch should not be part of the PR?

Makefile
vladandrew: I've moved it to the library's `Makefile.rules`[1] :confetti_ball: .

[1] https://github.com/unikraft/lib-rust/blob/staging/Makefile.rules#L39

lib/nolibc/string.c
vladandrew: Yes, musl 1.1.19 (the version ported on unikraft) defines bcmp at `src/string/bcmp.c`. They use the same call to `memcpy`, so I'd keep this as a function and not an alias to match musl's approach. I wouldn't complicate things with the handover.

support/build/Makefile.rules
vladandrew: Got it, yeah, I've removed this commit entirely

lib/ukrust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+license-file = "../../LICENSE.txt"
```
mkroening: Which license file does this reference? Preferably, we would specify `license` instead of `license-file`: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

lib/ukrust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "ukrust"
+version = "0.1.0"
+edition = "2018"
```
TrueDoctor, line 5: This should be updated to 2021 right?

lib/ukrust/ukrust_sys/lib.rs
```diff
@@ -49,3 +49,15 @@ pub mod c_types;
...
+extern "C" {
+    fn __ukrust_sys_crash() -> !;
+}
+
+#[panic_handler]
+fn panic(_info: &core::panic::PanicInfo<'_>) -> ! {
+    unsafe {
+        __ukrust_sys_crash();
+    }
+}
```
fabianpatras, line 62: line 54: ```warning: `__ukrust_sys_crash` redeclared with a different signature``` 
It is saying about `__ukrust_sys_crash` exported by `bindings`.

Can we remove the explicit ` -> !` to make the warning go away and insert a `loop {}` in the panic handler?

Like the examples from https://doc.rust-lang.org/nomicon/panic-handler.html

lib/ukrust/Cargo.toml
vladandrew: Good catch! I think unikraft is under the BSD 3 license, but not sure if there are changes. This is the main license file https://github.com/unikraft/unikraft/blob/staging/COPYING.md , I'll update the path to it.

lib/ukrust/Cargo.toml
vladandrew: Yeah, I did try that initially, but it doesn't build because the core crate is built using the 2018 edition[1].  Though, in the future we should definitely update it. Thanks Dennis!

[1] https://github.com/unikraft/unikraft/blob/staging/lib/ukrust/Makefile.uk#L19

lib/ukrust/ukrust_sys/lib.rs
vladandrew: Hi Fabian, I would keep it like this right now, since this would be a fix, and this pull request is related only to moving. Maybe add it in a separate pull request, but I remember they didn't use the `loop {}` in Linux neither, so I wonder why.

lib/ukrust/Cargo.toml
TrueDoctor: Ah, that makes sense

lib/ukrust/ukrust_sys/lib.rs
fabianpatras: Hi Vlad,

I see Linux using `loop {}` this way inside the `panic_handler`. [1]
This is, however, a workaround for a reason mentioned there ([1]) in the comments.

I understand your reasoning. Let's add it next time then.

[1] - https://github.com/torvalds/linux/blob/041fae9c105ae342a4245cf1e0dc56a23fbb9d3c/rust/kernel/lib.rs#L78-L87

vladandrew: Cool, they updated it! Yeah, we should do an upgrade in the future.

fabianpatras (commented):
Looks good to me (this I can only say about the Rust part). :ok_hand: 
There are things to modify, but as you mentioned (in the comments), they are not the scope of this PR and they should be done in a separate PR.

I don't have "Approve" or "Request changes" rights so I can only "Comment".

Reviewed-by: Fabian Patras <patras.fabi@gmail.com>

razvand:
@fabianpatras , you can hit the approve button now, I added you as a reviewer.

fabianpatras (approved):

lib/ukrust/ukrust_sys/lib.rs
```diff
@@ -49,3 +49,15 @@ pub mod c_types;
...
+
```
razvand: I would remove the extra blank line.

razvand (changes requested):
Thanks, @vladandrew , this works great. I would update the commit message to:

```
    To make life easier for Cargo, Rust crates should add `ukrust`
    as a dependency. We update `ukrust` to include a `Cargo.toml` file and
    to match the expected structure of a Cargo project. We use `src/` as a symbolic link directory
    pointing to `ukrust_sys/`.
```

lib/ukrust/ukrust_sys/lib.rs
vladandrew: Done

razvand (approved):
Thanks, @vladandrew 

PR #685, opened 2022-12-09
Makefile: Add flags for cross-compiler `LLVM` target

### LINKED ISSUE #921: Support cross-compiling with clang
### Feature request summary

This issue tracks sub-tasks for introducing support for cross-compiling with clang.

We will use arm64 as the reference use-case, with the clang target set to `aarch64-none-elf`, but the principles should also apply to future architectures. Also consider the possibility of cross-compiling on Arm hosts for x86_64 guests.

- [ ] Fix dependencies to GCC-specific directives. One known issue is Arm Errata when compiling with clang.
- [ ] Test that multithreading works clang's `-fno-emulated-tls` when the target is set to `aarch64-none-elf`. For background see https://github.com/unikraft/docs/issues/153.
- [ ] Implement missing items in the build system: Some work already covered by https://github.com/unikraft/unikraft/blob/staging/Makefile#L580-L599. Known remaining items are to deprecate `CONFIG_LLVM_TARGET_ARCH` and have the makefile set the target (or relevant flags) to `aarch64-none-elf` upon selecting arch/arm64 in Kconfig.
- [ ] Update documentation.

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

_No response_

mariasfiraiala:
### Prerequisite checklist

### Base target

 - Architecture(s): `AArch64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

- Download the cross-compiling toolchain. Get it from [the official ARM developer website](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads); look out for the `AArch64 bare-metal target (aarch64-none-elf)`.

- Disable all erratum options from `Architecture Selection` -> `Arm8 Compatible` -> `Workaround for [...] erratum` (for now, until #949 gets upstreamed)

- Build your app:
```console
make CC=clang CROSS_COMPILE=~/toolchains/gcc-arm-11.2-2022.02-x86_64-aarch64-none-elf/bin/aarch64-none-elf-
```

### Description of changes

If one wishes to compile an app using `clang`, for instance, they would have to choose the  `Custom cross-compiler LLVM target` option from the make build system, and manually set the flags.

This commit adds flags needed for cross-compiling with `clang` for `AArch64`.
Instead of the extra step of filling in the cross-compiler in the `Custom cross-compiler LLVM target` field, we harcode it to be `aarch64-none-elf`, which with `clang`, works like a charm (it doesn't get a trap due to the `emutls` issue - https://github.com/unikraft/docs/issues/153).

Fixes: #921

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -634,6 +634,13 @@ CC_VER_MAJOR   := $(word 2,$(subst ., ,$(CC_INFO)))
 CC_VER_MINOR   := $(word 3,$(subst ., ,$(CC_INFO)))
 CC_VERSION     := $(CC_VER_MAJOR).$(CC_VER_MINOR)
 
+ifeq ($(call have_clang),y)
+ifeq ("$(ARCH)", "arm64")
+ARCHFLAGS             += --target=aarch64-none-elf
+ISR_ARCHFLAGS         += --target=aarch64-none-elf
+endif
+endif
+
 ASFLAGS		+= -DCC_VERSION=$(CC_VERSION)
 CFLAGS		+= -DCC_VERSION=$(CC_VERSION)
 CXXFLAGS	+= -DCC_VERSION=$(CC_VERSION)


michpappas (changes requested):
@mariasfiraiala I think we should be striving for better clang integration, if possible one that does not depend on a GCC toolchain, but rather on clang, ldd, and binutils. If that is not possible, we should still update the build system so that the user is not required to manually set so many env vars.

If you decide to pursue this, here are some useful resources:
* https://static.linaro.org/connect/bkk19/presentations/bkk19-210.pdf
* https://archive.fosdem.org/2018/schedule/event/crosscompile/attachments/slides/2107/export/events/attachments/crosscompile/slides/2107/How_to_cross_compile_with_LLVM_based_tools.pdf

StefanJum:
@mariasfiraiala @michpappas @razvand, should we go ahead with some variation of this PR until we get a better `clang` integration? It would be nice to have `clang` support on `arm64` without needing to pull PRs, and I think better `clang` integration will take quite some time.

I think we can do something very transparent to the user, like check if the user selected any `LLVM target` in the menuconfig, if not (and the used CC is clang) just add a default one (I've been testing everything with `--targtet=aarch64-linux-none-gnu` so far).

The only env variable that should be set is the `CROSS_COMPILE` that should point to the toolchain path, [everything else is set based on that](https://github.com/unikraft/unikraft/blob/staging/Makefile#L581-L599), but adding the `--target` flag works for me without even setting the `CROSS_COMPILE` variable.

StefanJum:
@mariasfiraiala following the discussion [here](https://discord.com/channels/762976922531528725/1082548752612593694/1113024585911259236) we should update this PR to just add the `aarch64-none-elf` target to `CFLAGS`, `CXXFLAGS`, `ASFLAGS` etc if `arch/arm64` is selected in the configuration menu and `CC==clang`.

We should eventually just remove the whole `CONFIG_LLVM_TARGET_ARCH`, but it looks like Rust support might need it.

Makefile
```diff
@@ -634,6 +634,14 @@ CC_VER_MAJOR   := $(word 2,$(subst ., ,$(CC_INFO)))
 CC_VER_MINOR   := $(word 3,$(subst ., ,$(CC_INFO)))
 CC_VERSION     := $(CC_VER_MAJOR).$(CC_VER_MINOR)
 
+ifeq ($(call have_clang),y)
+ifeq ("$(ARCH)", "arm64")
+CFLAGS          += --target=aarch64-none-elf
+ASFLAGS         += --target=aarch64-none-elf
+CXXFLAGS        += --target=aarch64-none-elf
```
StefanJum:
```suggestion
ARCHFLAGS             += --target=aarch64-none-elf
ISR_ARCHFLAGS         += --target=aarch64-none-elf
```
Would it make more sense to add those to `ARCHFLAGS`?

StefanJum (changes requested):
Thank you @mariasfiraiala. Besides the comment, I would replace `flags` with something more specific (like `target flags`?) in the commit message / description.

mariasfiraiala:
Done @StefanJum, please take another look!

mariasfiraiala:
Just to have it here for documentation purposes, there is a small issue with  the TLS phdrs, that rarely shows up, a better description of it being on the [discord channel](https://discord.com/channels/762976922531528725/1082548752612593694/1131276221183967293).

StefanJum (approved):
Thank you @mariasfiraiala, all good.

michpappas:
Thanks @mariasfiraiala 🚀 

PR #686, opened 2022-12-11
Use the ukarch_random API to generate PAuth keys and enable Pointer Authentication on KVM

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [arm64]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

With the addition of the ukarch_random() API it is now possible to generate PAuth keys without delegating the implementation of key generation to the platform.

Remove ukplat_pauth_gen_key() from the platform API and generate PAuth keys using the ukarch_random() API.

Enable PAuth on KVM.


plat/kvm/arm/setup.c
```diff
@@ -311,6 +315,12 @@ void __no_pauth _libkvmplat_start(void *dtb_pointer)
 #endif /* CONFIG_ENFORCE_W_XOR_X */
 #endif /* CONFIG_PAGING */
 
+#ifdef CONFIG_ARM64_FEAT_PAUTH
```
mariasfiraiala, line 318: The enabling of the Pointer Authentication happens **after** a few other functions are called (`pl011_console_init()`, for instance) which in asm have instructions such as `paciasp`, `autiasp` (due to the `mbranch-protection` flag). These cannot be executed properly due to the feature not being initialized yet.

Is this how it's supposed to be? If not, suggest moving the initialization block for Pointer Authentication at the beginning of `_libkvmplat_start()`.

mariasfiraiala (changes requested):
Hi @michpappas! I've tested this PR against `qemu 4.2` and `qemu 7` with CPUs that support and do not support the Pointer Authentication feature and I have a few observations:

* when the feature is not available for that CPU, it hangs due to asm instructions not being able to get decoded. This happens due to the  check being done way to far in `_libkvmplat_start()`

* same goes for when the feature is supported, you'll find a better explanation below

michpappas:
Hi @mariasfiraiala

> when the feature is not available for that CPU, it hangs due to asm instructions not being able to get decoded. This happens due to the check being done way to far in _libkvmplat_start()

Yes, this is the same behavior as with any other unsupported instruction, so it's normal. What do you mean it happens due to the check being done way too far in `_libkvmplat_start()`?

> The enabling of the Pointer Authentication happens after a few other functions are called (pl011_console_init(), for instance) which in asm have instructions such as paciasp, autiasp (due to the mbranch-protection flag). These cannot be executed properly due to the feature not being initialized yet.
>
> Is this how it's supposed to be? If not, suggest moving the initialization block for Pointer Authentication at the beginning of _libkvmplat_start().

This shouldn't be happening, PAuth is applied only after the feature is enabled in the processor via `SCTLR_EL1.EnIA`. In fact, I cannot reproduce that behavior either. Could you please provide more info on your setup?

mariasfiraiala:
> Could you please provide more info on your setup?

@michpappas I was using a compiler with a version lower than 10 I was actually getting the issue documented here #728 . I am now testing this with `gcc-arm-11.2-2022.02-x86_64-aarch64-none-linux-gnu` and `qemu 4.2` (using the `max` CPU for the virt platform). `helloworld` looks good for now. I'll test `sqlite` too and that should be it, hopefully.

mariasfiraiala (approved):
I've tested `SQLite` too (this time with `qemu 6`). Everything looks ok. Thank you @michpappas for your support!

razvanvirtan (approved):
All good on my side too, thank you!

razvand (approved):

Issue #688, opened 2022-12-14
memory leak in buddy allocator


noureddine-taleb:
### Describe the bug

1. in the end of `uk_allocbbuddy_init` when we try to add the region to the allocator, we remove metalen byte from the region, which is already subtracted before, which lead to a waste of a couple pages:

```c
       min -= metalen

      ....

	if (max > min + metalen) {
		/* add left memory - ignore return value */
		bbuddy_addmem(a, (void *)(min + metalen),
				 (size_t)(max - min - metalen));
	}
```

2. in `bbuddy_addmem` when we try to calculate the number of pages in the region, we include the bitmap size in it, which is wrong because the first page eligible to allocation resides after the bitmap. this could crash the system in certain circumstances (overwrite system data):

```c
       memr->nr_pages =
               BITS_PER_BYTE * (range - sizeof(*memr)) /
               (BITS_PER_BYTE * __PAGE_SIZE + 1);
```

the size of the bitmap is not subtracted from `memr->nr_pages`.

PR #689, opened 2022-12-14
lib/ukallocbbuddy: fix a memory leak

noureddine-taleb:
@unikraft/maintainers-mem

### Prerequisite checklist

### Base target

### Description of changes

relevant information could be found in this github issue: https://github.com/unikraft/unikraft/issues/688

### DIFF
--- a/lib/ukallocbbuddy/bbuddy.c
+++ b/lib/ukallocbbuddy/bbuddy.c
@@ -11,8 +11,9 @@
  *      Author: Rolf Neugebauer (neugebar@dcs.gla.ac.uk)
  *     Changes: Grzegorz Milos
  *     Changes: Simon Kuenzer <simon.kuenzer@neclab.eu>
+ *     Changes: Nour-eddine Taleb <contact@noureddine.xyz>
  *
- *        Date: Aug 2003, changes Aug 2005, changes Oct 2017
+ *        Date: Aug 2003, changes Aug 2005, changes Oct 2017, changes Dec 2022
  *
  * Environment: Unikraft
  * Description: buddy page allocator from Xen.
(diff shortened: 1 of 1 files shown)


nderjung:
Hi @noureddine-taleb! Thank you for this contribution! Please could you add the following `GitHub-Fixes` reference in your commit message (and force-pushing again?) Thanks!

```
GitHub-Fixes: #688
Signed-off-by: ...
```

Also, cc'ing @unikraft/maintainers-mem to have a look at this fix

noureddine-taleb:
done

@unikraft/maintainers-mem

flpostolache:
Hello, @noureddine-taleb . Thank you for submitting this PR. I was looking through the issue points and the second point is a little bit unclear to me. Please correct me if I'm wrong. From my understanding the `memr->nr_pages` should store the number of pages that we have free and can play with. If that is so then the number of pages that we can have in that range would be equal with the formula given [above](https://github.com/unikraft/unikraft/blob/4ef7571e9f4bd7fbfde9bf5e43e5716e2015b386/lib/ukallocbbuddy/bbuddy.c#L428-L435). By directly taking the range and dividing it with the page size, the resulting number would also include the pages needed for the bitmap and memory region structure. This would then add more unnecessary bits to the bitmap and waste one page in some cases when it is not needed. The first point makes sense to me.

noureddine-taleb:
Hi @maniatro111 . so it's a classical checken or egg situation, because to allocate the bitmap we need first to know the amount of available free pages, and the to know the amount of available free pages we need first to allocate the bitmap, so we compromise and allocate the bitmap first using the whole range, as you said that may lead to an extra page wasted in the bitmap in some situations. but this at least gave us the correct number of free pages we can use. in contrast to the old implementation `memr->nr_pages =
		BITS_PER_BYTE * (range - sizeof(*memr)) /
		(BITS_PER_BYTE * __PAGE_SIZE + 1);
`  which correctly calculates the the number of free pages **except it doesn't take into consideration  the following rounding**: `memr_size = round_pgup(sizeof(*memr) + DIV_ROUND_UP(memr->nr_pages, BITS_PER_BYTE));` which reduces the number of free pages even more, in this case relying the calculated `memr->nr_pages` may lead to overflowing into unauthorized memory which is an undefined behavior. in this patch I count the whole range as free, and use it to calculate the bitmap size, and after doing the rounding I subtract the pages taken by the bitmap from the free range: `range -= memr_size;` and ended up with the correct `nr_pages`, plus the new implementation is more readable.

flpostolache (approved):
Thank you for the explanation. Makes sense now.

Reviewed-by: Florin Postolache <florin.postolache80@gmail.com>

marcrittinghaus (approved):
Thanks for the fixes @noureddine-taleb!

PR #690, opened 2022-12-16
lib/ukdebug: Use correct variable name for sect-strip.py tool

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBUKDEBUG=y`

### Description of changes

The variable used didn't match the one defined in the platform part and therefore the specified sections were never stripped.


RaduNichita (approved):
I tested it and everything works as expected. Thanks, @mschlumpp!

marcrittinghaus:
Hey @mschlumpp,

good catch! After applying the patch I get the following warning from `strip`:
`strip: /unikraft/build/unikraft_linuxu-x86_64.dbg: warning: empty loadable segment detected at vaddr=0x400000, is this intentional?`

The problem is gone for XEN after applying
- https://github.com/unikraft/unikraft/pull/698

Irrespective of this patch, I noticed that the `linuxu` target crashes with a segfault when tracing is enabled.

Any idea?

jschlumpp:
@marcrittinghaus This new warning is the result of the sect-strip.py actually doing something.
The linuxu crash is likely related to the `--omagic` flag, which sometimes produces invalid ELF files.

marcrittinghaus:
> @marcrittinghaus This new warning is the result of the sect-strip.py actually doing something. The linuxu crash is likely related to the `--omagic` flag, which sometimes produces invalid ELF files.

I know that it is due to it now actually doing something. I was wondering, if we can avoid the warning. But it seems at least with #698 to be only a problem for `linuxu` and tracing enabled.

jschlumpp:
Due to the placement of the `.uk_trace_keyvals` the linker seems to be forced to emit a new segment just for that section. However, `strip` doesn't seem to remove that leftover empty segment... I don't think we can do anything to fix that except moving the section for example behind `.rodata`. But doing so will probably leave a gap between `.rodata` and `.data`.

marcrittinghaus (approved):
Ok, I see. Too bad 😭

PR #691, opened 2022-12-16
include/event: Handle error code returns from event handlers

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

According to the comments the handlers should be able to return error codes to signal a failure. This fixes the event handler loop to handle these negative return values. We stop the handler iteration when a handler returns an error.


include/uk/event.h
```diff
@@ -267,8 +267,8 @@ static inline int uk_raise_event_ptr(struct uk_event *e, void *data)
 
 	uk_event_handler_foreach(itr, e) {
 		__uk_event_assert(*itr);
-		ret |= ((*itr)(data));
-		if (ret == UK_EVENT_HANDLED)
+		ret = ((*itr)(data));
+		if (ret < 0 || ret == UK_EVENT_HANDLED)
```
marcrittinghaus: A successful return must be sticky, which is not the case in your solution. E.g., you could have `HANDLED_CONT`->`NOT_HANDLED` and it would return `NOT_HANDLED` but should return `HANDLED(_CONT)`.

The code should look like this:
```C
		...
		rc = ((*itr)(data));
		if (rc < 0 || rc == UK_EVENT_HANDLED)
			return rc;
		ret |= rc;
	}
```

marcrittinghaus (changes requested):
Hey @mschlumpp,

just wanted to test this while debugging another issue and noticed it does not work. Please see my comment.

Another issue very closely related to this is that the calling sites of `uk_raise_event(_ptr)()` only check for a non-zero value to accept it as handled. But since the function might also return an `errno` the fault should be considered not handled for negative return values. Means only `> 0` means success.

Could you fix these as part of the PR?

include/uk/event.h
jschlumpp: Good point. However, were the enum values intended to be able to represent a bit-mask? In your proposed change the `raise_event` could also return `UK_EVENT_HANDLED_CONT | UK_EVENT_HANDLED == 3`.

marcrittinghaus: Right, I also had a variant with this instead of the OR.
```C
if (rc == UK_EVENT_HANDLED_CONT)
   ret = UK_EVENT_HANDLED;
```
It is just an additional branch that I wanted to avoid. But you are right, taking the function description into account the OR variant is not ok.

craciunoiuc (commented):
Hey @mschlumpp 

I'm not 100% sure how I should test this, but I tried some crashes and I guess things worked.

include/uk/event.h
```diff
@@ -251,25 +251,30 @@ struct uk_event {
...
 	int ret = UK_EVENT_NOT_HANDLED;
 
 	uk_event_handler_foreach(itr, e) {
 		__uk_event_assert(*itr);
-		ret |= ((*itr)(data));
-		if (ret == UK_EVENT_HANDLED)
+		rc = ((*itr)(data));
+		if (rc < 0 || rc == UK_EVENT_HANDLED)
 			break;
+		if (rc == UK_EVENT_HANDLED)
```
gabrielmocanu: Is this correct? For me, it seems that this code won't be executed.

gabrielmocanu (changes requested):

include/uk/event.h
jschlumpp: Oops. You are right. Good catch.

jschlumpp:
@craciunoiuc You could write some event handlers returning the various return values to check whether it works.

craciunoiuc:
> @craciunoiuc You could write some event handlers returning the various return values to check whether it works.

I feel stupid, I can't seem to get it to work. The [code comments](https://github.com/unikraft/unikraft/pull/691/files#diff-ef64eda0b3f99a203d7e108026575ac468a50a498579f2881cbb5a07c932e267L106)  (around Line 106 in `include/uk/event.h`) might be out of date.

With the comments there I think I created an event successfully, but I can't register the handlers, cause the call is incorrect (missing 1 argument).

Edit: Spoke too soon, got it to work, but the comment example is still wrong :D

jschlumpp:
Yes, the documentation seems to be wrong.
The macro is defined as:
```c
UK_EVENT_HANDLER_PRIO(event, fn, prio)
```
but the documentation uses it as:
```c
UK_EVENT_HANDLER_PRIO(handler1, UK_PRIO_EARLIEST);
```

This is what you are talking about, right?

craciunoiuc:
Yup, I was testing it now, almost done, it works btw ;)

include/uk/event.h
```diff
@@ -105,9 +105,9 @@
  *       return UK_EVENT_HANDLED;
  * }
  *
- * UK_EVENT_HANDLER_PRIO(handler1, UK_PRIO_EARLIEST);
- * UK_EVENT_HANDLER_PRIO(handler2, UK_PRIO_LATEST);
- * UK_EVENT_HANDLER_PRIO(handler3, UK_PRIO_LATEST);
+ * UK_EVENT_HANDLER_PRIO(event, handler1, UK_PRIO_EARLIEST);
+ * UK_EVENT_HANDLER_PRIO(event, handler2, UK_PRIO_LATEST);
+ * UK_EVENT_HANDLER_PRIO(event, handler3, UK_PRIO_LATEST);
```
craciunoiuc, line 110: Might as well go ahead and also mention in the comments how to trigger an event, I think this is also missing

```c
        struct myevent_data data;
        // ...
        int ret = uk_raise_event(myevent, &data);
```

craciunoiuc (changes requested):
@mschlumpp 

Only one last comment from my side.

Also, check the codacy errors, maybe they are fixable.

marcrittinghaus:
Thanks guys for fixing the glitches in the doc and the code. Crazy how that slipped through although I can remember that we had a fairly long discussion about the events.

include/uk/event.h
jschlumpp: That's already there https://github.com/unikraft/unikraft/blob/ac15f671186430848629313836715616687ff4e5/include/uk/event.h#L79

craciunoiuc: whoops, missed it, closing this

include/uk/event.h
```diff
@@ -251,24 +253,32 @@ struct uk_event {
...
+	int rc;
 	int ret = UK_EVENT_NOT_HANDLED;
 
 	uk_event_handler_foreach(itr, e) {
 		__uk_event_assert(*itr);
-		ret |= ((*itr)(data));
-		if (ret == UK_EVENT_HANDLED)
+		rc = ((*itr)(data));
+		if (rc < 0)
+			return rc;
```
marcrittinghaus: I won't get tired saying this. Can you please add the `unlikely` for error cases 😛

craciunoiuc (commented):
All good here. Thanks!

I'll wait with the review tag for @marcrittinghaus

marcrittinghaus (approved):
Hi @mschlumpp,

thanks 😃 Looks good!

craciunoiuc (approved):
All good here. Thanks!

PR #694, opened 2022-12-18
lib/uksched: Fix race condition during context switch

### LINKED ISSUE #643: uksched: Interrupt during booting process causes switch to idle thread which gets stuck
### Describe the bug

Functions from the `__wq_wait_event_deadline` macro (defined in [`lib/uksched/include/uk/wait.h`](https://github.com/unikraft/unikraft/blob/staging/lib/uksched/include/uk/wait.h#L96)) are heavily influenced by the presence or absence of interrupts.

The problem lies in what might happen between lines [116](https://github.com/unikraft/unikraft/blob/staging/lib/uksched/include/uk/wait.h#L116) and [119](https://github.com/unikraft/unikraft/blob/staging/lib/uksched/include/uk/wait.h#L119). Starting with line 116, interrupts are enabled. If there is an interrupt fired before the `yield` call at line 119, the IRQ handler `_ukplat_irq_handler()` kicks the scheduler out of the halting loop by clearing the `sched_have_pending_events` bit, which marks that no more work is left to be done. 

This is a problem when `yield()` is invoked in the early process of starting the unikernel (for boot threads, `init` and `idle`). Control is passed to the `idle` thread after the `yield()` call and due to the `sched_have_pending_events` bit being 0, the scheduler never passes control back to the `init` thread which does all the work.

This is how execution stops and gets stuck at certain points in the boot process when an interrupt fires right before the `yield()` call. It happens on both `redis` and `SQLite`, right after mounting the `9pfs` filesystem or other I/O operations.

The problem resurfaces now, with current work being done on refactoring the scheduling API and replacing `newlib` with `musl`.

### Steps to reproduce

Use an app that relies on "multithreading", `redis` or `SQLite`, with the new scheduling API.

### Expected behavior

I expected the unikernel to not get stuck during early boot stages.

arm64

kvm

### Relevant log output

_No response_

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`, `x86_64`
 - Platform(s): `kvm`
 - Application(s): `app-python3`, `app-redis`, `app-sqlite`

### Description of changes

The sched refactor effort introduced a race condition: if the current thread blocks and yields, and an interrupt is triggered right before the context switch, the interrupt handler tries to wake up the current thread.
The thread will remain blocked (potentially forever) until the next interrupt fires, because the current thread cannot be added to the run_queue list. The reason is that the condition `uk_thread_current() != t` is way too restrictive and does not cover the corner case where the current thread is just about to be evacuated.

This PR sets a flag on the current thread as soon as the next thread is scheduled. The interrupt will queue the current thread if it's also marked as queueable, in addition to it being runnable.

There was a similar commit (606f5f5) that was disregarded once the sched-refactoring work was upstreamed. This PR brings those changes up-to-date with the new API.

Github-Fixes: https://github.com/unikraft/unikraft/issues/643

### DIFF
--- a/lib/uksched/include/uk/thread.h
+++ b/lib/uksched/include/uk/thread.h
@@ -141,6 +141,14 @@ struct uk_thread *uk_thread_current(void)
 #define UK_THREADF_UKTLS      (0x002)	/**< Unikraft allocated TLS */
 #define UK_THREADF_RUNNABLE   (0x004)
 #define UK_THREADF_EXITED     (0x008)
+/*
+ *  A flag used for marking that a thread could potentially
+ *  be added to the run queue. A thread marked as such must
+ *  neither be running (except when it's about to be scheduled
+ *  out from the CPU during a context switch), nor be already
+ *  present in the run queue.
+ */
+#define UK_THREADF_QUEUEABLE  (0x010)
 
 #define uk_thread_is_exited(t)   ((t)->flags & UK_THREADF_EXITED)
 #define uk_thread_is_runnable(t) (!uk_thread_is_exited(t) \
@@ -148,11 +156,16 @@ struct uk_thread *uk_thread_current(void)
 #define uk_thread_is_blocked(t)  ((t)->flags &				\
 				  (UK_THREADF_EXITED | UK_THREADF_RUNNABLE) == \
 				  0x0)
+#define uk_thread_is_queueable(t) ((t)->flags & UK_THREADF_QUEUEABLE)
 
 #define uk_thread_set_runnable(t) \
 	do { (t)->flags |= UK_THREADF_RUNNABLE; } while (0)
 #define uk_thread_set_blocked(t) \
 	do { (t)->flags &= ~UK_THREADF_RUNNABLE; } while (0)
+#define uk_thread_set_queueable(t) \
+	do { (t)->flags |= UK_THREADF_QUEUEABLE; } while (0)
+#define uk_thread_clear_queueable(t) \
+	do { (t)->flags &= ~UK_THREADF_QUEUEABLE; } while (0)
 /* NOTE: Setting a thread as EXITED cannot be undone. */
 /* NOTE: Never change the EXIT flag manually. Trnasition to exit state reqiures
  * the terminate funcrtiomns to be called.
--- a/lib/ukschedcoop/schedcoop.c
+++ b/lib/ukschedcoop/schedcoop.c
@@ -115,6 +115,16 @@ static void schedcoop_schedule(struct uk_sched *s)
 		next = &c->idle;
 	}
 
+	if (next != prev) {
+		/*
+		 * Queueable is used to cover the case when during a
+		 * context switch, the thread that is about to be
+		 * evacuated is interrupted and woken up.
+		 */
+		uk_thread_set_queueable(prev);
+		uk_thread_clear_queueable(next);
+	}
+
 	ukplat_lcpu_restore_irqf(flags);
 
 	/* Interrupting the switch is equivalent to having the next thread
@@ -168,8 +178,10 @@ static void schedcoop_thread_woken(struct uk_sched *s, struct uk_thread *t)
 
 	if (t->wakeup_time > 0)
 		UK_TAILQ_REMOVE(&c->sleep_queue, t, queue);
-	if (t != uk_thread_current() && uk_thread_is_runnable(t))
+	if (uk_thread_is_queueable(t) && uk_thread_is_runnable(t)) {
 		UK_TAILQ_INSERT_TAIL(&c->run_queue, t, queue);
+		uk_thread_clear_queueable(t);
+	}
 }
 
 static __noreturn void idle_thread_fn(void *argp)


lib/ukschedcoop/schedcoop.c
```diff
@@ -168,8 +171,11 @@ static void schedcoop_thread_woken(struct uk_sched *s, struct uk_thread *t)
 
 	if (t->wakeup_time > 0)
 		UK_TAILQ_REMOVE(&c->sleep_queue, t, queue);
-	if (t != uk_thread_current() && uk_thread_is_runnable(t))
+	if ((t != uk_thread_current() || uk_thread_is_queueable(t))
```
jschlumpp: Under the assumption that the comments are above are really issues:
I wonder if there is a case when `uk_thread_is_queuable(t)` does not imply `t != uk_thread_current`. (Except when the thread is initially created). Otherwise, the condition could be simply `uk_thread_is_queueable(t) && uk_thread_is_runnable(t)`

lib/ukschedcoop/schedcoop.c
```diff
@@ -97,11 +97,13 @@ static void schedcoop_schedule(struct uk_sched *s)
...
-		    && uk_thread_is_runnable(prev)
-		    && !uk_thread_is_exited(prev))
-			UK_TAILQ_INSERT_TAIL(&c->run_queue, prev,
-					     queue);
+		if ((prev != &c->idle) && !uk_thread_is_exited(prev)) {
+			if (uk_thread_is_runnable(prev))
+				UK_TAILQ_INSERT_TAIL(&c->run_queue, prev,
+									queue);
+
+			uk_thread_clear_queueable(next);
```
jschlumpp: Shouldn't be the queueable flag of `next` be cleared in all cases (at least it shouldn't depend on the previous thread)?

lib/ukschedcoop/schedcoop.c
```diff
@@ -115,6 +117,7 @@ static void schedcoop_schedule(struct uk_sched *s)
 		next = &c->idle;
 	}
 
+	uk_thread_set_queueable(prev);
```
jschlumpp: This can set the queueable flag for `next` if `next` == `prev` (see line 109 in the updated file). Not sure if that's intentional.

lib/uksched/include/uk/thread.h
```diff
@@ -141,18 +141,24 @@ struct uk_thread *uk_thread_current(void)
 #define UK_THREADF_UKTLS      (0x002)	/**< Unikraft allocated TLS */
 #define UK_THREADF_RUNNABLE   (0x004)
 #define UK_THREADF_EXITED     (0x008)
+#define UK_THREADF_QUEUEABLE  (0x010)
```
jschlumpp, line 151: For the following comments, I assumed that the meaning of the flag is "not running and not in the run-queue". It could be useful to incorporate the meaning in the documentation comment above.

jschlumpp (commented):
I went through the code and added some comments.

lib/ukschedcoop/schedcoop.c
```diff
@@ -168,8 +171,11 @@ static void schedcoop_thread_woken(struct uk_sched *s, struct uk_thread *t)
 
 	if (t->wakeup_time > 0)
 		UK_TAILQ_REMOVE(&c->sleep_queue, t, queue);
-	if (t != uk_thread_current() && uk_thread_is_runnable(t))
```
Sairajkodilkar, line 171: Hey @mariasfiraiala and @eduardvintila, Why don't you remove the check `t != uk_thread_current()` since if the thread is not runnable, that means wait has already called uk_thread_set_block() and  uk_sched_thread_block() . This means we can safely remove current thread from sleep queue and add it to run queue. 
(also @skuenzer is there a specific reason, because of which you have added this check ?)

So Even if we receive the interrupt `ukplat_lcpu_restore_irqf()` and `uk_sched_yield()`, we are sure that the current thread is on sleep queue and we can remove it. 

ideally if we do this, we do not need queuable flag.

eduardvintila:
Hey, @mschlumpp. Really great suggestions. I've incorporated virtually all of them in the PR now. Please take another look when you have the time. Thanks a lot!

lib/ukschedcoop/schedcoop.c
eduardvintila: Hello, @Sairajkodilkar. By eliminating the `t != uk_thread_current()` condition, we risk placing the current thread on the `run_queue`  before the execution of `yield()`. In our case, consider the `__wq_wait_event_deadline` macro from [lib/uksched/include/uk/wait.h,](https://github.com/unikraft/unikraft/blob/staging/lib/uksched/include/uk/wait.h#L119) which is used for blocking the main thread, for example.

If an interrupt fires at this point, right before the `yield()`:
```C
            ukplat_lcpu_restore_irqf(flags); \
            if (lock) \
                unlock_fn(lock); \
            uk_sched_yield(); \
```
... and wakes up the thread by adding it to the `run_queue`, then during the yield, this assertion will fail: 
```C
    next = UK_TAILQ_FIRST(&c->run_queue);
    if (next) {
        UK_ASSERT(next != prev);
 ```
which can be seen during booting:
```console
[    0.156197] CRIT: [libukschedcoop] <schedcoop.c @   93> Assertion failure: next != prev
[    0.156494] Info: [libkvmplat] <shutdown.c @   35> Unikraft halted
```

However, thank you nevertheless for your suggestion! Eventually I have replaced the check you have mentioned with `uk_thread_is_queueable`, which is more general in our case.

lib/uksched/include/uk/thread.h
```diff
@@ -141,18 +141,29 @@ struct uk_thread *uk_thread_current(void)
 #define UK_THREADF_UKTLS      (0x002)	/**< Unikraft allocated TLS */
 #define UK_THREADF_RUNNABLE   (0x004)
 #define UK_THREADF_EXITED     (0x008)
+/*
+ *  A flag used for marking that a thread could potentially
+ *  be added to the run queue. A thread marked as such
+ *  must not be running and already present in run queue.
```
jschlumpp: When I first read the comment I understood it as: `A thread marked as such (must not be running) and (already present in run queue).`
Might be less ambiguous using "neither .. nor ...".

eduardvintila:
I've updated the PR, @mschlumpp. Good catch as always, thanks!

jschlumpp (approved):
Thanks!

Sairajkodilkar (commented):
Thank @eduardvintila, I have tested this PR

Reviewed by: Sairaj Kodilkar skodilkar7@gmail.com

Sairajkodilkar (approved):

razvand (approved):

PR #695, opened 2022-12-19
arch/x86_64: Implement ukarch_random

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

Introduce a "Processor Features" menu under "Architecture Selection" when x86 is selected.
The `CONFIG_X86_64_HAVE_RANDOM` option enables the processor-generated randomness.

### Description of changes

This PR introduces an implement the ukarch_random API for x86_64. The implementation uses the `RDRAND` / `RDSEED` instructions provided by Intel's DRNG technology [1]. Each of these instructions has been introduced at different stages of the architecture's timeline:
* **RDRAND:** Intel Atom Silvermont, Intel Core 3rd Generation,  Intel Xeon E3-1200 v2, E5 v2, E7 v2.
* **RDSEED:** Intel Atom Goldmont, Intel Core 5th Generation, Intel Core M.

This implementation requires that both instructions are supported by the processor.


arch/x86/x86_64/include/uk/asm/random.h
```diff
@@ -0,0 +1,113 @@
...
+		return -ENOTSUP;
+	}
+
+	return 0;
+}
+
+static inline int ukarch_random_u64(__u64 *val)
+{
+	unsigned int i;
+	unsigned char res = 0;
```
razvanvirtan, line 60: `res` seems a bit confusing to me, since it suggests that this variable stores the actual `rdrand` result. However, it only stores the flag that informs us whether a random number was available or not. Maybe `ok` or `rand_success` would be a more appropriate name?

razvanvirtan, line 57: `res` seems a bit confusing to me, since it suggests that this variable stores the actual `rdrand` result. However, it only stores the flag that informs us whether a random number was available or not. Maybe `ok` or `rand_success` would be a more appropriate name?

arch/x86/x86_64/include/uk/asm/random.h
```diff
@@ -0,0 +1,113 @@
...
+
+	for (i = 0; i < RDRAND_RETRY_LIMIT; i++) {
+		__asm__ __volatile__(
+			"	rdrand	%0\n" /* get rand */
+			"	setc	%1\n" /* get result */
```
razvanvirtan, line 62: Same as above, I would suggest something like `get success flag`.

arch/x86/x86_64/include/uk/asm/random.h
```diff
@@ -0,0 +1,113 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef __UKARCH_RANDOM_H__
+#error Do not include this header directly
+#endif
+
+#include <errno.h>
+
```
razvanvirtan, line 36: I tested the PR by generating some random numbers in `app-helloworld`. The functions work, but I can't build without adding here these headers:
```
#include <uk/arch/lcpu.h>
#include <uk/print.h>
```

razvanvirtan (commented):
Hi, @michpappas!
I tested the functions and they seem to work just as expected! I will let here a couple of comments.

arch/x86/x86_64/include/uk/asm/random.h
michpappas: I don't see a problem with this. Intel does it this way in their document 
https://www.intel.com/content/www/us/en/developer/articles/guide/intel-digital-random-number-generator-drng-software-implementation-guide.html and so does linux https://elixir.bootlin.com/linux/latest/source/arch/x86/include/asm/archrandom.h#L20. The return value is also documented in the API, and we even protected it with `__check_result` as it is critical to check whether a number was generated successfully.

arch/x86/x86_64/include/uk/asm/random.h
michpappas: Yes you're right, somehow that slipped. I'll push an update.

arch/x86/x86_64/include/uk/asm/random.h
razvanvirtan: I do not suggest removing the variable, just renaming it in `ok` or something similar (I first took the idea from the Intel docs where they use `ok`, I see also Linux does the same). However, I agree it doesn't have any functional relevance, so we can also keep it as it is :)

razvanvirtan (approved):
Thank you for this!

mariasfiraiala (approved):
Looks good to me. Thank you @michpappas !

arch/x86/x86_64/include/uk/asm/random.h
```diff
@@ -0,0 +1,110 @@
...
+		);
+		if (likely(res))
+			break;
+	}
+	return !res;
+}
+
+static inline int ukarch_random_seed_u64(__u64 *val)
+{
+	unsigned char res = 0;
```
marcrittinghaus: Just a nitpick. The initialization is not used due to `=qm`. For the other ones you can argument it is good style because there is a loop whose body conceptionally does not need to be executed.

I will approve afterwards.

michpappas: Thanks, done.

marcrittinghaus (approved):
Thanks! 🚀

PR #698, opened 2022-12-20
build: Fix segments in ELF output images

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): yes
 - Platform(s): also yes
 - Application(s): N/A

### Description of changes

This PR includes the following things:
 - removal of the haunted :ghost: `omagic` flag, which has been the root cause for issues like broken ELF binaries, non-mmapable unikernel images, and other atrocities.
 - a hand-written section-to-segment mapping. This fixes the segments to not have overly lax permissions of segments such as the `.text` section being writable by being in the same segment as `.data`.
 - the removal of the sect-strip.py tool which has only been used to remove a warning from the build process. The script is replaced by a simpler grep call.

Due to the lack of toolchain/platforms I was only able to test it on {KVM,linuxu}/x86. It would be a great help if somebody can test the other platform/arch combinations.

- [x] KVM x86
- [x] KVM ARM
- [ ] Xen x86
- [ ] Xen ARM (?)
- [x] linuxu x86
- [ ] linuxu ARM


lib/uktest/tests.ld
```diff
@@ -9,4 +9,4 @@ SECTIONS
 		uk_asserttab_end = .;
 	}
 }
-INSERT AFTER .data;
+INSERT AFTER .rodata;
```
marcrittinghaus: The test sections have to be RW so they still have to be located after `.data` 😃

marcrittinghaus:
The PR is missing `posix-socket`:
```C
SECTIONS
{
	. = ALIGN(0x8);
	.posix_socket_driver_list : {
		PROVIDE(posix_socket_driver_list_start = .);
		KEEP(*(SORT_BY_NAME(.posix_socket_driver_*)));
		PROVIDE(posix_socket_driver_list_end = .);
	}
}
INSERT AFTER .data;
```

plat/kvm/arm/link64.lds.S
```diff
@@ -42,6 +42,16 @@ OUTPUT_FORMAT("elf64-littleaarch64")
...
 
+PHDRS
+{
+	headers PT_PHDR PHDRS;
+	text PT_LOAD FILEHDR PHDRS FLAGS(5);
+	rodata PT_LOAD FLAGS(4);
+	data PT_LOAD;
+	tls PT_TLS;
+	stack PT_GNU_STACK FLAGS(6);
+}
```
marcrittinghaus: Just an idea: maybe we could put these definitions in `common.lds.h`:
```C
#define PHDRS_PF_X 0x1 /* Executable */
#define PHDRS_PF_W 0x2 /* Writeable */
#define PHDRS_PF_R 0x4 /* Readable */

#define PHDRS_PF_RX (PHDRS_PF_R | PHDRS_PF_X)
#define PHDRS_PF_RW (PHDRS_PF_R | PHDRS_PF_W)
```

```suggestion
PHDRS
{
	headers PT_PHDR PHDRS;
	text PT_LOAD FILEHDR PHDRS FLAGS(PHDRS_PF_RX);
	rodata PT_LOAD FLAGS(PHDRS_PF_R);
	data PT_LOAD;
	tls PT_TLS;
	stack PT_GNU_STACK FLAGS(PHDRS_PF_RW);
}
```

plat/kvm/arm/link64.lds.S
```diff
@@ -42,6 +42,16 @@ OUTPUT_FORMAT("elf64-littleaarch64")
 OUTPUT_ARCH(aarch64)
 ENTRY(_libkvmplat_entry)
 
+PHDRS
+{
+	headers PT_PHDR PHDRS;
+	text PT_LOAD FILEHDR PHDRS FLAGS(5);
```
marcrittinghaus: I have a question here. When you build a unikernel you get something like this:
```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x00000000000ff040 0x00000000000ff040
                 0x0000000000000150 0x0000000000000150  R      0x8
  LOAD           0x0000000000000000 0x00000000000ff000 0x00000000000ff000
                 0x00000000002467ce 0x00000000002467ce  R E    0x1000
  LOAD           0x0000000000247000 0x0000000000346000 0x0000000000346000
                 0x00000000000760c8 0x00000000000760c8  R      0x1000
  LOAD           0x00000000002be000 0x00000000003bd000 0x00000000003bd000
                 0x00000000001ad3a8 0x00000000001f37a8  RW     0x1000
  TLS            0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000         0x8
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x8
```
The headers are located in the first page (@`0xff000`) of the first `LOAD` segment. So they share the **executable** permissions of the `.text` segment. Means you could execute whatever data comprises the ELF header 😱 

Now it is alright with multiboot because we do not load that part. However, this might be different for the ASLR case, where we would use the ELF headers to load the kernel.

lib/uktest/tests.ld
jschlumpp: Ah. I forget that the test results are stored in the section...

plat/kvm/arm/link64.lds.S
jschlumpp: I moved the headers into their own segment.

jschlumpp:
I replaced the objcopy step by making the ELF loadable using multiboot directly. Thanks @mkroening, for the hints regarding multiboot ELF loading.

plat/kvm/arm/link64.lds.S
marcrittinghaus: > I replaced the objcopy step by making the ELF loadable using multiboot directly. Thanks @mkroening, for the hints regarding multiboot ELF loading.

I just saw this change. Interesting! You state that it is loadable by grub in the commit message. It is still loadable by QEMU directly, too?

marcrittinghaus:
I just saw this change. Interesting! You state that it is loadable by grub in the commit message. It is still loadable by QEMU directly, too?

Also since you specify the architecture to be ELF32. Is this working properly with GDB without having to manually switch the 64-bit architecture in GDB?

jschlumpp:
> I just saw this change. Interesting! You state that it is loadable by grub in the commit message. It is still loadable by QEMU directly, too?

Oh, I meant QEMU but it's just a normal multiboot header. So it's not wrong either, just untested for GRUB at the moment.

> Also since you specify the architecture to be ELF32. Is this working properly with GDB without having to manually switch the 64-bit architecture in GDB?

It is set to ELF32 in the strip step, therefore the `.dbg` file is still ELF64. However, I have to figure out how to do that without breaking ARM (the CI fails at the moment).

marcrittinghaus:
I just tested this together with #722 and it works good so far 😃 Very cool, to get rid of omagic this way and also have proper ELF segments! However, I still have to check the PR with an actual application  and also do a second pass over the individual changes.

plat/kvm/x86/multiboot.S
```diff
@@ -49,11 +49,11 @@ multiboot_header:
...
-	.long multiboot_header				/* header addr */
-	.long 0x100000					/* load addr */
-	.long __bss_start				/* load end addr */
-	.long _end					/* bss end addr */
-	.long multiboot_start32				/* entry addr */
+	.long 0						/* header addr */
+	.long 0						/* load addr */
+	.long 0						/* load end addr */
+	.long 0						/* bss end addr */
+	.long 0						/* entry addr */
```
marcrittinghaus, line 56: These fields can be completely removed. When `AOUT_KLUDGE` is not specified they don't need to be present at all.

marcrittinghaus: These fields can be removed. If `AOUT_KLUDGE` is not defined the fields do not need to be present.

marcrittinghaus (changes requested):
Hi @mschlumpp,

I just did a second pass. I have two additional comments:

1. Could you extend the commit message that is changing the binary to `elf32` format to give a short explanation why this is not a problem for an x64 build. Also explicitly note that you can no longer load the `*.dbg` version of the binary into QEMU and that this has to remain x64 so you can easily load it into gdb for debugging.

2. I noticed that the `.bss` segment will be placed together with the `.data` segment. That leads to the ELF file growing notably. I guess it would be possible to add a program header for the `.bss` segment?

jschlumpp:
Hey @marcrittinghaus, the problem with `.bss` seems to be that the next section `.intrstack` is marked as `PROGBITS`, forcing the area occupied by `.bss` to be allocated. I don't think we need this separate section anymore as we can now properly handle alignment gaps. Therefore, I would remove it completely, which also solves the size increase.

marcrittinghaus (approved):
Thanks very much @mschlumpp! Much appreciated. Looks good.

skuenzer:
Short question from my side: Did you guys double-test if the tracepointing code and scripts still work with these changes? Otherwise, looks really good. Thanks @mschlumpp !

jschlumpp:
I did a quick test and the tracepoints still work.

skuenzer (approved):
Thanks a lot @mschlumpp!

PR #699, opened 2022-12-20
plat/*: Add IRQ events

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`, `arm`
 - Application(s): N/A

### Description of changes

This allows external libraries to hook into any IRQ by using the ukevent functionality. An event handler can also decide whether to pass the control flow to other event handlers and normal IRQ handlers.

### DIFF
--- a/include/uk/plat/irq.h
+++ b/include/uk/plat/irq.h
@@ -32,6 +32,8 @@
 #ifndef __UKPLAT_IRQ_H__
 #define __UKPLAT_IRQ_H__
 
+#include <uk/arch/lcpu.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -55,6 +57,22 @@ typedef int (*irq_handler_func_t)(void *);
  * @return 0 on success, a negative errno value on errors
  */
 int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg);
+
+/** The event payload for the #UKPLAT_EVENT_IRQ event */
+struct ukplat_event_irq_data {
+	/** The registers of the interrupted code */
+	struct __regs *regs;
+	/** The platform specific interrupt vector number */
+	unsigned long irq;
+};
+
+/**
+ * This event is raised before the platform code handles an IRQ. The normal
+ * IRQ handling will continue or stop according to the returned `UK_EVENT_*`
+ * value.
+ * Note: this event is usually raised in an interrupt context.
+ */
+#define UKPLAT_EVENT_IRQ ukplat_event_irq
 
 #ifdef __cplusplus
 }
--- a/plat/common/arm/traps_arm64.c
+++ b/plat/common/arm/traps_arm64.c
@@ -199,7 +199,7 @@ void trap_el1_sync(struct __regs *regs, __u64 far)
 	ukplat_crash();
 }
 
-void trap_el1_irq(void)
+void trap_el1_irq(struct __regs *regs)
 {
 	UK_ASSERT(gic);
 
@@ -208,5 +208,5 @@ void trap_el1_irq(void)
 		UK_CRASH("EL1 async tag check fault\n");
 #endif /* CONFIG_ARM64_FEAT_MTE */
 
-	gic->ops.handle_irq();
+	gic->ops.handle_irq(regs);
 }
--- a/plat/common/include/uk/plat/common/irq.h
+++ b/plat/common/include/uk/plat/common/irq.h
@@ -63,8 +63,9 @@ enum uk_irq_polarity {
  * Calls the registered IRQ handlers for the given IRQ vector. This will
  * acknowledge the IRQ.
  *
+ * @param regs the registers of the interrupted code
  * @param irq IRQ vector
  */
-void _ukplat_irq_handle(unsigned long irq);
+void _ukplat_irq_handle(struct __regs *regs, unsigned long irq);
 
 #endif /* __PLAT_CMN_IRQ_H__ */
--- a/plat/drivers/gic/gic-v3.c
+++ b/plat/drivers/gic/gic-v3.c
@@ -499,7 +499,7 @@ static void gicv3_init_dist(void)
 	uk_pr_info("GICv3 distributor initialized.\n");
 }
 
-static void gicv3_handle_irq(void)
+static void gicv3_handle_irq(struct __regs *regs)
 {
 	uint32_t stat, irq;
 
@@ -518,7 +518,7 @@ static void gicv3_handle_irq(void)
 		isb();
 
 		if (irq < GIC_MAX_IRQ) {
-			_ukplat_irq_handle((unsigned long)irq);
+			_ukplat_irq_handle(regs, (unsigned long)irq);
 			gicv3_eoi_irq(stat);
 
 			continue;
--- a/plat/kvm/irq.c
+++ b/plat/kvm/irq.c
@@ -86,7 +86,7 @@ int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
  */
 extern unsigned long sched_have_pending_events;
 
-void _ukplat_irq_handle(unsigned long irq)
+void _ukplat_irq_handle(struct __regs *regs, unsigned long irq)
 {
 	struct irq_handler *h;
 	int i;
--- a/plat/kvm/irq.c
+++ b/plat/kvm/irq.c
@@ -33,9 +33,12 @@
 #include <uk/plat/common/irq.h>
 #include <kvm/intctrl.h>
 #include <uk/assert.h>
+#include <uk/event.h>
 #include <uk/print.h>
 #include <errno.h>
 #include <uk/bitops.h>
+
+UK_EVENT(UKPLAT_EVENT_IRQ);
 
 /* IRQ handlers declarations */
 struct irq_handler {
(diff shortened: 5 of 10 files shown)


plat/kvm/irq.c
```diff
@@ -86,13 +89,24 @@ int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
...
 	int i;
+	int rc;
+	struct ukplat_event_irq_data ctx;
 
 	UK_ASSERT(irq < __MAX_IRQ);
 
+	ctx.regs = regs;
+	ctx.irq = irq;
+	rc = uk_raise_event(UKARCH_EVENT_IRQ, &ctx);
+	if (rc == UK_EVENT_HANDLED)
```
mogasergiu: Since you're not using `rc` anywhere else, why not remove it and just
```
if (uk_raise_event(UKARCH_EVENT_IRQ, &ctx) == UK_EVENT_HANDLED)
```

plat/linuxu/irq.c
```diff
@@ -150,9 +153,17 @@ static void _irq_handle(int irq)
 {
 	struct irq_handler *h;
 	int i;
+	int rc;
```
mogasergiu, line 156: ditto

plat/xen/events.c
```diff
@@ -85,6 +89,8 @@ void unbind_all_ports(void)
 int do_event(evtchn_port_t port, struct __regs *regs)
 {
 	ev_action_t *action;
+	int rc;
```
mogasergiu, line 92: ditto

plat/kvm/irq.c
marcrittinghaus: Personally, I prefer the separate `rc` version just for consistency with the rest of the code, but   add a `likely` for good practice. And with the `likely` also the line becomes a bit long and hard to read.

plat/drivers/gic/gic-v3.c
```diff
@@ -499,7 +499,7 @@ static void gicv3_init_dist(void)
 	uk_pr_info("GICv3 distributor initialized.\n");
 }
 
-static void gicv3_handle_irq(void)
+static void gicv3_handle_irq(struct __regs *regs)
```
marcrittinghaus, line 502: Don't forget to add `__unused` attributes if necessary.

plat/kvm/irq.c
jschlumpp: I don't see a reason to use either `likely` or `unlikely` here. The return values really depends on the event handlers and the parameters. (But there is still the error case `rc < 0` missing, which is introduced in the other PR, for that unlikely makes sense).

For the comments on the `rc` variable I agree with Marc.

plat/drivers/gic/gic-v3.c
jschlumpp: Is this a more general comment? `regs` is used in the function.

marcrittinghaus: Ah, yes. My bad 👍🏻

jschlumpp:
I just wanted to adjust this PR to #691 regarding error from event handlers. I only see two options to handle these errors in the interrupt handler: 

1. Call `UK_CRASH`. That means any error returned from an event handler is fatal.
2. Ignore the error and insert a tracepoint for debugging.

(we can't print in the context of the interrupt handler, as the print functionality is not ISR-safe)

I'm currently leaning towards implementing option (1). Are there any opinions or better ideas about this?

marcrittinghaus:
As unhandled traps also cause a `UK_CRASH`, it makes sense to go for option 1.

plat/common/arm/traps_arm64.c
```diff
@@ -208,5 +208,5 @@ void trap_el1_irq(void)
 		UK_CRASH("EL1 async tag check fault\n");
 #endif /* CONFIG_ARM64_FEAT_MTE */
 
-	gic->ops.handle_irq();
+	gic->ops.handle_irq(regs);
```
mogasergiu, line 211: I believe you wanted to pass `regs` as an argument to `trap_el1_irq` as well?

include/uk/plat/irq.h
```diff
@@ -56,6 +58,13 @@ typedef int (*irq_handler_func_t)(void *);
  */
 int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg);
 
+struct ukplat_event_irq_data {
+	struct __regs *regs;
+	unsigned long irq;
+};
+
+#define UKARCH_EVENT_IRQ event_irq
```
marcrittinghaus: We should call it `UKPLAT_` since the ctx structure is also `ukplat` and it is accessible via the `plat` header. Putting in it the `arch/x86/x86_64/include/uk/asm/lcpu.h` header (and duplicate for other architectures) would also work with `ukarch` prefix. However, we are currently discussing that the `arch` folder should only contain things actually defined by the architectural manual. And the events are more of a platform construct anyways.

In any case also put the prefix in the actual event name. We should have done this also for the trap events. I'll probably change it during re-arch. 

Also add a remark when this event is raised.
```suggestion
/** Bla */
#define UKPLAT_EVENT_IRQ ukplat_event_irq
```

include/uk/plat/irq.h
```diff
@@ -56,6 +58,13 @@ typedef int (*irq_handler_func_t)(void *);
  */
 int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg);
 
+struct ukplat_event_irq_data {
+	struct __regs *regs;
+	unsigned long irq;
+};
```
marcrittinghaus: Please add comments to the structure.
```suggestion
/** Bla */
struct ukplat_event_irq_data {
	/** Bla */
	struct __regs *regs;
	/** Bla */
	unsigned long irq;
};
```

plat/kvm/irq.c
```diff
@@ -86,13 +89,26 @@ int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
...
+	ctx.regs = regs;
+	ctx.irq = irq;
+	rc = uk_raise_event(UKARCH_EVENT_IRQ, &ctx);
+	if (unlikely(rc < 0))
+		UK_CRASH("IRQ event handler returned error: %d\n", rc);
+	if (rc == UK_EVENT_HANDLED)
+		/* Skip all normal handlers if an event handler handled the
+		 * event
+		 */
+		goto exit_ack;
```
marcrittinghaus: Put `{}` around the block due to the comment. Checkpatch will not complain.

plat/common/include/uk/plat/common/irq.h
```diff
@@ -65,6 +65,6 @@ enum uk_irq_polarity {
  *
  * @param irq IRQ vector
```
marcrittinghaus, line 67:
```suggestion
 * @param regs ...
 * @param irq IRQ vector
```

plat/common/arm/traps_arm64.c
marcrittinghaus: +1. This will also require changes to the ARM assembly for trap handling just like with x86.

marcrittinghaus (changes requested):
Hey @mschlumpp,

many thanks for this PR! Please see the comments.

plat/xen/events.c
```diff
@@ -38,11 +38,15 @@
...
 #include <xen/xen.h>
 #include <uk/print.h>
+#include <uk/plat/irq.h>
+#include <uk/event.h>
 #include <uk/assert.h>
 #include <uk/bitops.h>
 
 #define NR_EVS 1024
 
+UK_EVENT(UKARCH_EVENT_IRQ);
```
marcrittinghaus:
```suggestion
UK_EVENT(UKPLAT_EVENT_IRQ);
```

plat/xen/events.c
```diff
@@ -93,6 +99,14 @@ int do_event(evtchn_port_t port, struct __regs *regs)
 		return 1;
 	}
 
+	ctx.regs = regs;
+	ctx.irq = port;
+	rc = uk_raise_event(UKARCH_EVENT_IRQ, &ctx);
```
marcrittinghaus:
```suggestion
	rc = uk_raise_event(UKPLAT_EVENT_IRQ, &ctx);
```

plat/linuxu/irq.c
```diff
@@ -150,9 +153,19 @@ static void _irq_handle(int irq)
...
+	struct ukplat_event_irq_data ctx;
 
 	UK_ASSERT(irq >= 0 && irq < __MAX_IRQ);
 
+	ctx.regs = NULL;
+	ctx.irq = irq;
+	rc = uk_raise_event(UKARCH_EVENT_IRQ, &ctx);
```
marcrittinghaus:
```suggestion
	rc = uk_raise_event(UKPLAT_EVENT_IRQ, &ctx);
```

plat/linuxu/irq.c
```diff
@@ -49,6 +50,8 @@
 #define __MAX_IRQ 16
 #define HANDLER_RESERVED ((void *) 0x1)
 
+UK_EVENT(UKARCH_EVENT_IRQ);
```
marcrittinghaus:
```suggestion
UK_EVENT(UKPLAT_EVENT_IRQ);
```

marcrittinghaus (changes requested):
One last thing 😄

mogasergiu (approved):

marcrittinghaus (approved):
Thanks @mschlumpp, @mogasergiu!

PR #703, opened 2022-12-25
Update documentation of ukarch_random

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Fix summary on ukarch_random_u32/u64. Remove implementation-specific
information from ukarch_random_seed_u32/u64.

### DIFF
--- a/include/uk/arch/random.h
+++ b/include/uk/arch/random.h
@@ -50,7 +50,7 @@ extern "C" {
 int ukarch_random_init(void);
 
 /**
- * Get a 64-bit random integer
+ * Get a 32-bit random integer
  *
  * @param [out] val Pointer to store the generated value
  * @return 0 on success, negative value on failure
@@ -58,7 +58,7 @@ int ukarch_random_init(void);
 int __check_result ukarch_random_u32(__u32 *val);
 
 /**
- * Get a 32-bit random integer
+ * Get a 64-bit random integer
  *
  * @param [out] val Pointer to store the generated value
  * @return 0 on success, negative value on failure
@@ -66,7 +66,7 @@ int __check_result ukarch_random_u32(__u32 *val);
 int __check_result ukarch_random_u64(__u64 *val);
 
 /**
- * Reseed the RNG and get a 64-bit random integer
+ * Get a 32-bit random integer suitable for seeding
  *
  * @param [out] val Pointer to store the generated value
  * @return 0 on success, negative value on failure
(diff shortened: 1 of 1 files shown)


vladandrew (approved):
Looks good to me. Thanks Michalis!

include/uk/arch/random.h
```diff
@@ -50,31 +50,31 @@ extern "C" {
...
+ * Get a 64-bit random integer
  *
  * @param [out] val Pointer to store the generated value
  * @return 0 on success, negative value on failure
  */
 int __check_result ukarch_random_u64(__u64 *val);
 
 /**
- * Reseed the RNG and get a 64-bit random integer
+ * Get a 32-bit random integer suitable for seeding
```
marcrittinghaus, line 69: Hi @michpappas,

I am wondering if the changed documentation is actually correct. Since the implementation on ARM will not only return a random number suitable for seeding, but will also reseed the RNG. Important is that the latter is ok because we otherwise do not have a way to reseed as there is simply no API to use with a new seed.

Maybe we should call the API `ukarch_random_reseed_u64()` instead?

michpappas: Hi @marcrittinghaus, actually the point of this PR is exactly that, ie to avoid exposing implementation-specific detail to the documentation of the generic API. With the risk of covering some information already known, I explain:

The API is expected to provide two functions.
* `ukarch_random`: Used for the generation of random numbers for generic use, at a high throughput.
* `ukarch_ranbdom_seed`:  Used for generation of random numbers that are suitable for seeding a CSPRNG. This is not expected to provide the throughput of the generic interface, yet the generated numbers exhibit higher brute-force resistance than numbers generated by the interface for generic randomness.

Intel describes the architecture of RDRAND / RDSEED in some detail [1] (in an attempt to convince people that RDRAND is not backdoored? 😆):

Both interfaces are based on the output of a thermal-noise entropy source that is fed to an AES-CBC-MAC based conditioner.

RDRAND: The output of the conditioner seeds a Deterministic Random Bit Generator (DRBG), which can provide random numbers at a high throughput. The DRBG is periodically re-seeded through the conditioner.

RDSEED: The output of the conditioner is fed to what Intel refers to as an Enanced Non-deterministic Random Number Generator (ENRNG). Some [2] speculate that the ENRNG module is just a direct interface to the output of the conditioner.

Similarly, in [3] Arm states that:

RNDR: "The output random number is from a Deterministic Random Bit Generator (DRBG), which is seeded from a True
Random Number Generator (TRNG)."

RNDRSS: "Returns a 64-bit random number which is reseeded from the True Random Number source immediately before the read of the random number."

To me both sound very similar, but in any case that shouldn't matter, as these are differences between the implementations of Arm and Intel. Also notice that in general one doesn't need to re-seed the generic RNG, as the implementation takes care of that.

[2] https://darkfield.dev/2021/09/25/rdseed-vs-rdrand-a-question-of-entropy/
[3] https://developer.arm.com/documentation/ddi0487/latest

marcrittinghaus (approved):
Thanks @michpappas for the detailed explanation! Makes sense 😄

PR #711, opened 2023-01-08
lib/vfscore: Fix `lstat()` infinite loop

### LINKED ISSUE #652: lib/vfscore: Calls made to libc wrapper functions instead of unikraft core implementations lead to infinite loops
### Describe the bug

Inside `vfscore`, there are multiple occurrences of function calls which end up invoking the libc wrapper, not the actual implementation provided by `vfscore`. The libc wrappers themselves may actually invoke (either directly or indirectly through other wrappers) the initial syscall from `vfscore` which called the libc wrapper, thus ending up in an infinite loop.

For example, consider the `lstat` syscall from [`vfscore`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/main.c#L1603), which calls the `__lxstat` function providing the actual implementation:

```C
#if UK_LIBC_SYSCALLS
int __lxstat(int ver __unused, const char *pathname, struct stat *st)
{
    [actual implementation]
}

#ifdef __lxstat64
#undef __lxstat64
#endif

LFS64(__lxstat);
#endif /* UK_LIBC_SYSCALLS */


UK_SYSCALL_R_DEFINE(int, lstat, const char*, pathname, struct stat*, st)
{
    return __lxstat(1, pathname, st);
}
```

When building with Musl, `__lxstat` is included with it, so this libc implementation ends up being called (not the `vfscore` one from above):

```C
int lstat(const char *restrict path, struct stat *restrict buf)
{
#ifdef SYS_lstat
    return syscall(SYS_lstat, path, buf);
#else
    return syscall(SYS_fstatat, AT_FDCWD, path, buf, AT_SYMLINK_NOFOLLOW);
#endif
}


int __lxstat(int ver, const char *path, struct stat *buf)
{
    return lstat(path, buf);
}
```

In Musl, `__lxstat` is a wrapper for `lstat`. The latter libc wrapper then ends up calling vfscore's `lstat` syscall implementation on x86 systems (where SYS_lstat is defined), which results in calling __lxstat from Musl, ad infinitum...

### Steps to reproduce

Build any application with Musl that uses the `lstat` syscall on x86 systems.

`app-sqlite` is an example, where `lstat` is used when reading the journal file of the database. This means that the issue arises when launching the app after previously creating a database.

### Expected behavior

I expected the application to not get stuck in an infinite loop.

x86_64, arm64

kvm

### Relevant log output

_No response_

mariasfiraiala:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

Build a simple app (such as `helloworld`) on `x86` and call `lstat()`.

### Description of changes

Including `musl` with even simple apps and attempting to call `lstat()` on `x86` results in an infinite loop due to `__lxstat()`, a helper function, existing in `musl` as well.

This PR changes the function called by `lstat()` in order to stop referencing `musl` by adding an extra wrapper, used by `__lxstat()` too. The change was made in order to be consistent to other implementations found in `vfscore`, such as [this one](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/main.c#L315).

Fixes: #652

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -1604,8 +1604,7 @@ UK_TRACEPOINT(trace_vfs_lstat, "pathname=%s, stat=%p", const char*,
 UK_TRACEPOINT(trace_vfs_lstat_ret, "");
 UK_TRACEPOINT(trace_vfs_lstat_err, "errno=%d", int);
 
-#if UK_LIBC_SYSCALLS
-int __lxstat(int ver __unused, const char *pathname, struct stat *st)
+int __lxstat_helper(int ver __unused, const char *pathname, struct stat *st)
 {
 	struct task *t = main_task;
 	char path[PATH_MAX];
@@ -1631,6 +1630,12 @@ int __lxstat(int ver __unused, const char *pathname, struct stat *st)
 	return -error;
 }
 
+#if UK_LIBC_SYSCALLS
+int __lxstat(int ver __unused, const char *pathname, struct stat *st)
+{
+	return __lxstat_helper(1, pathname, st);
+}
+
 #ifdef __lxstat64
 #undef __lxstat64
 #endif
@@ -1642,7 +1647,7 @@ int __lxstat(int ver, const char *pathname, struct stat *st);
 
 UK_SYSCALL_R_DEFINE(int, lstat, const char*, pathname, struct stat*, st)
 {
-	return __lxstat(1, pathname, st);
+	return __lxstat_helper(1, pathname, st);
 }
 
 static int __fxstatat_helper(int ver __unused, int dirfd, const char *pathname,


StefanJum (commented):
This works file @mariasfiraiala, thank you.

I think the `__lxstat` function is used since the actual implementation of `lstat` may differ on different systems, and the `__lxstat` function will be overwritten to provide the desired implementation.
However, I don't see any use for it in the unikraft `lstat` syscall (or in the unikraft core), since we want the syscall implementation to be well defined within the core, not dependant on any kind of external libraries (like `musl`).

This pr will make the `__lxstat` function useless anyway since it will never be called, no matter what library implements it. so I would go for just moving the code in the `UK_SYSCALL_R_DEFINE lstat`, since using another helper function makes things messy and makes the original `__lxstat` useless.

Any suggestions on this @razvand @eduardvintila?

mariasfiraiala:
> This pr will make the `__lxstat` function useless anyway since it will never be called, no matter what library implements it. so I would go for just moving the code in the `UK_SYSCALL_R_DEFINE lstat`, since using another helper function makes things messy and makes the original `__lxstat` useless.

@StefanJum Indeed that would be the natural approach, however, there are other instances in `vfscore` (that I mentioned in the message above) that follow the same path: have a another helper function, used by both the syscall and its helper: [`__xmknod_helper`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/main.c#L315).

I tried to follow this implemetation in order to keep the consistency.

StefanJum:
> @StefanJum Indeed that would be the natural approach, however, there are other instances in `vfscore` (that I mentioned in the message above) that follow the same path: have a another helper function, used by both the syscall and its helper: [`__xmknod_helper`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/main.c#L315).
> 
> I tried to follow this implemetation in order to keep the consistency.

Yes, those instances also don't make sense to me, I would change them too, but there could be something I'm missing.

eduardvintila:
Hi @mariasfiraiala and @StefanJum. The PR is looking good from my side: it finally fixes that pesky infinite loop issue with `app-sqlite` on x86 and also brings `lstat` to be more consistent with other implementations in `vfscore`.

I agree with @StefanJum in that we should eventually consider  to move the code from the helpers straight to the syscall implementations. However, I strongly propose that we postpone this refactoring effort for a seperate PR. There are at least four other syscalls currently present in `vfscore` that I know of which follow a similar pattern:

* `mknod` -> `__xmknod`, `__xmknod_helper`
* `fstat` -> `__fxstat`, `__fxstat_helper`
* `fstatat` -> `__fxstatat`, `__fxstatat_helper`
* `stat` ->`__xstat`, `__xstat_helper`

There may be even more.  I find that the refactoring of all of these functions is outside the scope of this PR, which is to fix the infinite loop issue. I think it's good that this PR brings `lstat` to the existing pattern mentioned above, since it should make an eventual cleanup of `vfscore` more straightforward by having the functions consistent with each other.

StefanJum:
Makes sense @eduardvintila. @mariasfiraiala if we go with this approach, please open an issue on refactoring these `vfscore` functions so we won't forget about them.

razvand:
@mariasfiraiala , @StefanJum , @eduardvintila , as discussed on Discord, I agree we can merge this PR as it is for consistency.

Then, a new PR would be required that updates all these functions (there's 5 of them: `stat`, `lstat`, `fstat`, `fstatat`, `mknod`) with these steps:

- Move the contents of the `__*xstat_helper()` functions to `*stat()` functions.
- Remove the `__*xstat_helper()` functions.
- Have the `__*xstat()` functions call `*stat()` functions.
- Guard the `__*xstat()` functions with `#if UK_LIBC_SYSCALLS`. `*stat()` functions are already guarded via `UK_SYSCALL_R_DEFINE`.

eduardvintila (approved):
Alright, this should go in. Thanks, @mariasfiraiala.

razvand:
@StefanJum, if this is OK with you, please approve this PR and add your message tag. I will then approve it as an assignee.

StefanJum (approved):
Thanks @mariasfiraiala.

razvand (approved):

PR #715, opened 2023-01-14
build: Add target for building a static library image + linker scripts

mkroening:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`, `linuxu`
 - Application(s): N/A

### Additional configuration

 - N/A

### Description of changes

This adds an `alib_images` target to the build system. The `alib_images` target packs the final object files before the final linking step into a static library and generates default and non-default linker script files.

The static library together with the linker scripts can then be used to link applications against Unikraft images with build systems that are not integrated into the Unikraft build system.

I want to use this for unikraft-rs which currently builds a static library from Unikraft's objects manually and searches for and assembles the linker scripts. This approach is very fragile and error-prone, because it has to be adapted manually to support all kinds of different configurations regarding the following points:
1. The selection of object files to put in the static library
2. The selection of linker scripts for the current configuration
3. Properly cleaning and removing the correct object files

This PR would move all of that complexity out of the unikraft-rs build script and would enable linking against Unikraft from external toolchains much more easily.
Inside the Unikraft build system, it is much easier to build these files correctly.

Previous work:
 - [`unikraft-upb/scripts/static-lib`](https://github.com/unikraft-upb/scripts/tree/ba2693e59a595beae95dd31eb7270464bcac81a1/static-lib)
 - [`unikraft-upb/app-helloworld#static-lib`](https://github.com/unikraft-upb/app-helloworld/tree/a40ae59952c9a767f1a50aa615134da3179a03b3)
 - [`mkroening/unikraft-static`](https://github.com/mkroening/unikraft-static)
 - [`mkroening/unikraft-rs/build.rs`](https://github.com/mkroening/unikraft-rs/blob/515eab83acf732f5a8b8234fe8ace8128c52e596/build.rs) (currently archived in favor of the soon-to-be released `unikraft/unikraft-rs`)

Related:
 - https://github.com/unikraft/kraftkit/issues/175

#### Changes

This PR is best reviewed on a commit by commit basis.

1. Replaces `$(<PLATFORM>_IMAGE).dbg` with `$(<PLATFORM>_DEBUG_IMAGE)` for KVM and linuxu, which is already defined. This is already the case for xen.
2. Introduces an intermediate final object file target for kvm and linuxu. This was already existed on kvm, though not as an explicit intermediate target. This allows us to put this single object file into the static library. Alternatively, we could also put the individual object files into the static library, though that would require the user to force linking of the `--whole-archive` for some library objects to not get completely garbage-collected.
3. Adds the top-level `alib_images` target, which produces a static library from the final object file instead of linking it to a proper image.
4. Changes the platform linker script for linuxu to be passed as default linker script to align the behavior to that of the kvm platform.
5. Adds default and normal linker scripts to the `alib_images` output by copying or concatenating respectively.

#### KVM

```console
$ # Configure an image with the default name "unikraft" first
$ make alib_images
$ gcc \
    -nostdlib \
    -Wl,--omagic \
    -Wl,--build-id=none \
    -nostdinc \
    -no-pie \
    -Wl,-e,_libkvmplat_entry \
    -Wl,-dT,build/unikraft_kvm-x86_64.def.lds \
    -Wl,-T,build/unikraft_kvm-x86_64.lds \
    build/unikraft_kvm-x86_64.a \
    -o unikraft_kvm-x86_64
$ qemu-system-x86_64 -nographic -kernel unikraft_kvm-x86_64
```

#### linuxu

```console
$ # Configure an image with the default name "unikraft" first
$ make alib_images
$ gcc \
    -nostdlib \
    -Wl,--omagic \
    -Wl,--build-id=none \
    -nostdinc \
    -no-pie \
    -Wl,-e,_liblinuxuplat_start \
    -Wl,-dT,build/unikraft_linuxu-x86_64.def.lds \
    -Wl,-T,build/unikraft_linuxu-x86_64.lds \
    build/unikraft_linuxu-x86_64.a \
    -o unikraft_linuxu-x86_64
$ ./unikraft_linuxu-x86_64
```

Note that the only difference between the linker invocations is the entry symbol name.


mkroening:
Hmmm, I don't think the CI failure on arm-linuxu is caused by this PR. Looking around, other PRs seem to have the same issue.

vladandrew:
Yep, no worry Martin, I think these CI failures are common right now.

mkroening:
Open question: How should the static library be called? I currently went for `<IMAGE_NAME>.a`. That way, it is lexicographically near to the other final output images.

`lib<IMAGE_NAME>.a` might make sense as well, because of convention for static libraries. That way, it is one `lib*` between the other `lib*.o`s, though.

What do you think?

marcrittinghaus:
A problem with the approach of collecting linker scripts is that it will not work beginning with 0.12 because we will have post-processing steps. The final executable receives an update to the boot information section that is generated from the layout of the executable. This step is executed after linking. We are also likely to have additional such steps, for example, to generate the static boot page table at link-time.

mkroening:
> we will have post-processing steps

I see. I would still like to pursue being able to link a Unikraft image externally, even if it requires additional steps after linking. This is necessary to support Rust executables.

We would still need to figure out how to best perform these additional steps once they are there, but that's separate from this PR, I think.

mkroening:
Closing in favor of a different approach, repurposing the Unikraft build system as a linker. The first step is

* https://github.com/unikraft/unikraft/pull/957

PR #720, opened 2023-01-18
arch/libcontext: arm64: Make TCB overlap configurable

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [`aarch64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

The Aarch64 ABI requires 16 bytes of space after the `tlsp` pointer as part of the TCB. However, this space is not expected by newer versions of the `musl` library.

As such, this PR introduces a new Makefile rule by which a library can register whether this reserved space is not necessary. The rule sets a macro by which we can then decide to assume a 16-byte block space is reserved at the end of the TCB or not.

This maintains backwards compatibility with previous versions of `musl`, since if the rule is not invoked, it is assumed that the library expects the `AARCH64_RESERVED` block of 16 bytes described by the ABI.

### DIFF
--- a/arch/Makefile.rules
+++ b/arch/Makefile.rules
@@ -12,3 +12,10 @@ $(if $(strip $(_UKARCH_TLS_HAVE_TCB)),
 		$(eval CXXFLAGS-y+=-DCONFIG_UKARCH_TLS_HAVE_TCB=1) \
 		$(eval _UKARCH_TLS_HAVE_TCB:=1))
 endef
+
+# Indicate that no block of bytes is reserved at the end of the TCB
+# Usage:
+#  $(eval $(call aarch64_no_reserved_tcb_overlap))
+define aarch64_no_reserved_tcb_overlap =
+	$(eval LIBCONTEXT_CFLAGS-y+=-DCONFIG_AARCH64_NO_TCB_OVERLAP=1)
+endef
--- a/arch/arm/arm64/tls.c
+++ b/arch/arm/arm64/tls.c
@@ -59,11 +59,27 @@
 
 /*
  * Aarch64 ABI requires 16 bytes of space after the tlsp as part of the TCB.
- * NOTE: We assume that libraries (like musl) reserve this space when using
- *       this TLS API. You can find more information here:
+ * NOTE: Some libraries, such as older versions of musl, reserve this space
+ *       when using this TLS API. You can find more information here:
  *       https://www.openwall.com/lists/musl/2018/02/09/7
+ *
+ * Newer versions of musl, however, simply ignore this block of 16 bytes
+ * described by the ABI and expect the tlsp to point just at the end of
+ * the TCB. More details can be found here:
+ * https://www.openwall.com/lists/musl/2018/06/01/14
+ *
+ * Therefore, when such libraries are used, they can call the makefile rule
+ * `aarch64_no_reserved_tcb_overlap` which sets a macro if the library
+ * **does not** expect a 16 bytes reserved block as part of the TCB,
+ *  as is the case with newer versions of musl.
  */
 #define AARCH64_RESERVED 16
+
+#ifdef CONFIG_AARCH64_NO_TCB_OVERLAP
+#define AARCH64_TCB_OVERLAP 0
+#else
+#define AARCH64_TCB_OVERLAP AARCH64_RESERVED
+#endif
 
 #ifndef __UKARCH_TLS_HAVE_TCB__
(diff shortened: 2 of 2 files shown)


arch/arm/arm64/tls.c
```diff
@@ -59,11 +59,25 @@
...
+ * Therefore, when such libraries are used, they can call the makefile rule
+ * `aarch64_no_reserved_tcb_overlap` which sets a macro if the library
+ * **does not** expect a 16 bytes reserved block as part of the TCB,
+ *  as is the case with newer versions of musl.
  */
+#ifdef CONFIG_AARCH64_NO_RESERVED_TCB_OVERLAP
+#define AARCH64_RESERVED 0
+#else
 #define AARCH64_RESERVED 16
+#endif
```
kubanrob, line 82: `AARCH64_RESERVED` was originally indented to be the size of the 16byte reserved block, not the overlap with the TLS.
So I would prefer to introduce a separate define for that, since the reserved block is still there, it just does not overlap the TLS.

Setting it to 0 works because the only point it is used as the size of that block (and not as the overlap) is in `tls_data_offset`, and `AARCH64_RESERVED > TLS_AREA_ALIGN` is not true.

All other occurrences of `AARCH64_RESERVED` actually mean the overlap, not the size of the reserved block.

arch/Makefile.rules
```diff
@@ -12,3 +12,10 @@ $(if $(strip $(_UKARCH_TLS_HAVE_TCB)),
 		$(eval CXXFLAGS-y+=-DCONFIG_UKARCH_TLS_HAVE_TCB=1) \
 		$(eval _UKARCH_TLS_HAVE_TCB:=1))
 endef
+
+# Indicate that no block of bytes is reserved at the end of the TCB
+# Usage:
+#  $(eval $(call aarch64_no_reserved_tcb_overlap))
+define aarch64_no_reserved_tcb_overlap =
+	$(eval LIBCONTEXT_CFLAGS-y+=-DCONFIG_AARCH64_NO_RESERVED_TCB_OVERLAP=1)
```
kubanrob: I find it a bit dangerous to manually make defines in the `CONFIG_` namespace. I am not sure if this should be a proper config option.

EDIT: This is the same naming as in https://github.com/unikraft/unikraft/blob/staging/arch/Makefile.rules, so nevermind.

kubanrob (changes requested):

arch/arm/arm64/tls.c
eduardvintila: Good point, makes sense. I have kept the `AARCH64_RESERVED` macro for the ABI reserved 16 byte block and introduced a new macro `AARCH64_TCB_OVERLAP` which is set to 0 when there is no TCB overlap. Otherwise, it is set to `AARCH64_RESERVED`.

arch/Makefile.rules
eduardvintila: 👍🏻

eduardvintila:
Hey, @kubanrob!  Thanks a lot for pointing this out! I have updated my PR accordingly.

arch/arm/arm64/tls.c
```diff
@@ -86,7 +102,7 @@ extern char _tls_start[], _etdata[], _tls_end[];
  *             | Custom TCB format       | |   > Thread Control Block (TCB)
  *             | (might be used          | |  |  (length: TCB_SIZE)
  *             |  by a libC)             | |  |
- *       tlsp -+-> AARCH64_RESERVED      | |  |
+ *       tlsp -+-> AARCH64_TCB_OVERLAP   | |  |
```
kubanrob, line 105: There is another note/reference to the overlap under this diagram that needs to be updated / deleted. Maybe the explanation next to `AARCH64_RESERVED` is enough.

kubanrob (commented):
I stumbled over one more comment that might needs to be updated, otherwise I think the PR looks good!

eduardvintila:
Hey, @kubanrob. Great catch, thanks! I've updated that comment, as well.

arch/arm/arm64/tls.c
eduardvintila: I've tried to come up with an update. I didn't want to delete it because I found your original comment to be very useful in understanding the TLS layout. Please let me know if you find it okay :)

kubanrob (approved):
Looks good to me, thank you @eduardvintila!

razvand (approved):

PR #722, opened 2023-01-18
Boot code refactoring

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

Currently, the x86-64 KVM platform defines a `multiboot.S` file that provides the early multiboot1 entry point. This entry point eventually communicates the multiboot-specific boot information structures to the `_libkvmplat_entry()` function, which then works on the multiboot structures to setup the platform.

This approach has the disadvantage that for every other boot protocol like Linux Boot Protocol (Firecracker), or EFI, we have to replicate the whole platform initialization. Also initialization steps are hard to share across architectures. 

This PR introduces a generic Unikraft boot information (`ukplat_bootinfo`) structure that the platform initialization is based on. Every individual boot protocol then supplies a custom entry point that extracts all relevant data from the protocol-specific boot information and sets up the Unikraft boot information before jumping into the generic platform initialization.

#### Memory regions
A central piece of information during boot is the current memory setup. This PR extends the existing `ukplat_memregion_desc` so that it is able to store the physical and virtual address of memory regions (see the commit description for further changes). It thereby allows to use the descriptor in the boot information to describe the memory layout also for more complicated cases where the kernel image might be loaded at an arbitrary virtual address (ASLR). 

Previously, iterating over the `ukplat_memregion_desc`s was implemented by using the section symbols (e.g., `__TEXT`, `__ETEXT`). With the memory regions being part of the boot information, this code is now gone.

#### MAP/UNMAP flags (paging only)
The memory regions can be marked with a `MAP` and `UNMAP` flag to instruct the kernel to map/unmap the region at boot (when paging is enabled). This is useful, for example, to unmap code/data of a chainloader after jumping into the final unikernel. Also mapping a dynamically discovered region like the `initrd` can be expressed in a generic way, without having to pollute the platform code with special `initrd` handling.

#### mkbootinfo.py
The PR includes a script `mkbootinfo.py` which is executed at link-time to provide a prepared boot information structure that already contains the layout (with permissions) of the kernel image. For multiboot this information is not supplied by the protocol. Instead at runtime the multiboot information on the location of the `initrd`, device memory and free physical memory is used to supplement the already existing kernel layout information.

#### Plat Re-arch
The code is structured in a way so that it will be easy during platform re-architecture to move each boot protocol entry (multiboot1, multiboot2, EFI, Linux Boot Protocol, ...) into its own library.

### Limitations

- Currently only KVM x86-64. All other platforms just receive compatibility updates.
- Only multibootv1 boot code at the moment.


lib/ukboot/boot.c
```diff
@@ -134,43 +149,48 @@ void ukplat_entry(int argc, char *argv[])
...
+			  alloc_pages << PAGE_SHIFT);
+#else /* CONFIG_HAVE_PAGING */
+	/* Paging is disabled so we still have the static boot page table set
+	 * that maps (some of) the physical memory to virtual memory. The
+	 * FREE memory regions reflect this. Try to use these memory regions to
+	 * initialize the allocator. If it fails, we will try again with the
+	 * next region. As soon we have an allocator, we simply add every
+	 * subsequent region to it.
+	 */
+	ukplat_memregion_foreach(&md, UKPLAT_MEMRT_FREE, 0, 0) {
```
mogasergiu: Probably something to keep in mind: sometimes the memory map descriptors passed on by the firmware may be out of order, or may contain separate descriptors of allocatable memory ranges that may otherwise could have been coalesced into one bigger memory range. I do not know if the memory allocators take this into account, but perhaps it would be a good idea to sort the memory descriptors and possibly coalesce them if suitable.

Your function `ukplat_memregion_list_insert` seems to assume they are ordered, for example.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -33,6 +33,108 @@
...
+ *
+ * @return
+ *   The index of the element added on success, an negative errno otherwise
+ */
+static inline int
+ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
+			     const struct ukplat_memregion_desc *mrd)
+{
+	struct ukplat_memregion_desc *p;
+	__u32 i;
```
mogasergiu, line 90: Just a nitpick, but you return an `__u32`, yet the function's signature indicates returning an `int`.

plat/kvm/x86/setup.c
```diff
@@ -1,456 +1,295 @@
...
-	offset = mr[0]->start - PAGE_LARGE_ALIGN_UP(mr[0]->start);
-	start  = PAGE_LARGE_ALIGN_UP(mr[0]->start);
-	len    = PAGE_LARGE_ALIGN_DOWN(mr[0]->len - offset);
-
-	rc = ukplat_page_unmap(&kernel_pt, start,
-			       (PLATFORM_MAX_MEM_ADDR - start) >> PAGE_SHIFT,
-			       PAGE_FLAG_KEEP_FRAMES);
-	if (unlikely(rc))
-		goto EXIT_FATAL;
+	rc = ukplat_memregion_list_insert(&bi->mrds,
```
mogasergiu, line 213: You are inserting here a descriptor whose start address is 0x0, maybe I am wrong but wouldn't it will always end up the first in the list. Would it not be better to just directly call the unmap function here without inserting a dummy descriptor?

marcrittinghaus: With the plat-rearch we will move the paging API into a separate library. For that reason, I also designed the `paging_init()` to be self-contained, so it does not need additional data from the platform. Now, since I cannot do the unmap before (not initialized) or after (clashes with existing mappings) `paging_init()`, and I don't want to put the unmap in there explicitly (that would introduce platform knowledge about the 1:1 boot page table), the way to go is to use an UNMAP entry.

Actually, it is not really a dummy entry in the sense that it would be a hack here. In fact, I expect for chainloading we will have unmapping entries in there to unmap the first-stage loader after jumping into the (relocated) unikernel. What I might add is to remove unmap entries from the list after having done the unmap. But since they are usually not listed anyways, I did not include this for now.

plat/kvm/x86/setup.c
```diff
@@ -1,456 +1,295 @@
...
+				ukplat_memregion_list_delete(&bi->mrds, i);
+		} else if (mrdp->vbase + mrdp->len > PLATFORM_MAX_MEM_ADDR) {
+			/* Region overlaps with unmapped area */
+			uk_pr_info("Memory %012lx-%012lx outside mapped area\n",
+				   PLATFORM_MAX_MEM_ADDR,
+				   mrdp->vbase + mrdp->len);
+
+			if (mrdp->type == UKPLAT_MEMRT_FREE)
+				mrdp->len -= (mrdp->vbase + mrdp->len) -
+						PLATFORM_MAX_MEM_ADDR;
```
mogasergiu, line 254: Perhaps another niptick, but since you are assuming the list of memory descriptors is already ordered descendingly you could `break;` here as well, or drop the `else if` entirely:
 ```
         } else {
			if (mrdp->vbase + mrdp->len > PLATFORM_MAX_MEM_ADDR) {
				/* Region overlaps with unmapped area */
				uk_pr_info("Memory %012lx-%012lx outside mapped area\n",
					   PLATFORM_MAX_MEM_ADDR,
				           mrdp->vbase + mrdp->len);

				if (mrdp->type == UKPLAT_MEMRT_FREE)
					mrdp->len -= (mrdp->vbase + mrdp->len) -
						     PLATFORM_MAX_MEM_ADDR;
			}

			break;
		}
```
though that would add another level of indentation.

File: plat/common/include/uk/plat/common/memory.h [hunk 5]
```diff
@@ -33,6 +33,108 @@
...
+/**
+ * Delete the specified region from the memory list.
+ *
+ * @param list
+ *   The memory region list from which to delete the region
+ * @param idx
+ *   The index of the region to delete
+ */
+static inline void
+ukplat_memregion_list_delete(struct ukplat_memregion_list *list, __u32 idx)
```
mogasergiu, line 124: Would it be better to have this function receive a `ukplat_memregion_desc  *` as a second argument instead of an index? I believe this would make it more uniform with the insert function as well as easier to use inside a `foreach` block, since you do not have an index there. If the index is required for sanity checks perhaps something as easy as
```
static inline __u32
ukplat_memregion_list_get_idx(struct ukplat_memregion_list *list,
			      const struct ukplat_memregion_desc *mrd)
{
	return mrd - list->mrds;
}
```
could be used inside the delete function.

plat/kvm/include/kvm-x86/multiboot.h
```diff
@@ -20,9 +20,10 @@
  */
 /* multiboot.h - Multiboot header file.  */
 
-#ifndef MULTIBOOT_HEADER
-#define MULTIBOOT_HEADER 1
```
mogasergiu, line 24: Not related to this PR but, maybe in the future, would it make sense to move the multiboot headers to a platform common directory? I do not think it would make sense to have it in a KVM only directory since, for example, Xen also supports it. Perhaps even architecture indepedent (especially for Multiboot2) as at one point, iirc, there were some Xen patches to make Multiboot more generic and accomodate for ARM.

plat/common/include/uk/plat/common/memory.h
mogasergiu: Actually, nevermind about the `foreach` block, this might mess up the next `ukplat_memregion_find_next` since it iterates ascendingly.

plat/xen/memory.c
```diff
@@ -47,97 +48,104 @@
...
 	UK_ASSERT(m);
 
 	switch (i) {
 	case 0: /* text */
-		m->base  = (void *) __TEXT;
-		m->len   = (size_t) __ETEXT - (size_t) __TEXT;
-		m->flags = (UKPLAT_MEMRF_RESERVED
-			    | UKPLAT_MEMRF_READABLE
-			    | UKPLAT_MEMRF_EXECUTABLE);
+		mrd[i].pbase = __TEXT;
```
jschlumpp, line 66: Could this be done in a `UK_CTOR` once at boot or even statically instead of every call?

plat/linuxu/memory.c
```diff
@@ -149,39 +150,39 @@ int ukplat_memregion_count(void)
...
+	static struct ukplat_memregion_desc mrd[2];
 	int ret;
 
 	UK_ASSERT(m);
 
 	if (i == 0 && _liblinuxuplat_opts.heap.base) {
-		m->base  = _liblinuxuplat_opts.heap.base;
-		m->len   = _liblinuxuplat_opts.heap.len;
-		m->flags = UKPLAT_MEMRF_ALLOCATABLE;
+		mrd[0].pbase = (__paddr_t)_liblinuxuplat_opts.heap.base;
```
jschlumpp, line 161: See comment in plat/xen/memory.c

plat/kvm/include/kvm-x86/multiboot.h
```diff
@@ -193,4 +207,21 @@ struct multiboot_apm_info {
...
+#define MULTIBOOT_BOOTLOADER_MAGIC		0x2BADB002
+
+/* Align all boot modules on i386 page (4KB) boundaries. */
+#define MULTIBOOT_PAGE_ALIGN			0x00000001
+
+/* Must pass memory information to OS. */
+#define MULTIBOOT_MEMORY_INFO			0x00000002
+
+/* This flag indicates the use of the address fields in the header. */
+#define MULTIBOOT_AOUT_KLUDGE			0x00010000
```
jschlumpp: Note: Delete this define after rebased onto the newest segment fix PR.

plat/common/include/uk/plat/common/bootinfo.h
```diff
@@ -0,0 +1,80 @@
...
+#define __PLAT_CMN_BOOTINFO_H__
+
+#include <uk/arch/types.h>
+#include <uk/arch/limits.h>
+#include <uk/plat/common/memory.h>
+
+/** Unikraft boot info */
+struct ukplat_bootinfo {
+	/** Magic value to recognize Unikraft boot info */
+#define UKPLAT_BOOTINFO_MAGIC			0xb001b0b0 /* Boot Bobo */
```
jschlumpp: F41l
```suggestion
#define UKPLAT_BOOTINFO_MAGIC			0xb007b0b0 /* Boot Bobo */
```

plat/common/bootinfo.lds.S
```diff
@@ -0,0 +1,11 @@
+/* Define .uk_bootinfo section that depending on the boot protocol can
+ * be populated by with boot information
+ */
+SECTIONS {
+	. = ALIGN(8);
+	.uk_bootinfo :
+	{
+		KEEP(*(.uk_bootinfo))
+	} :data
```
jschlumpp: You can omit `:data` because it will be inherited from the preceding sections (`.data`)

support/scripts/mkbootinfo.py
```diff
@@ -0,0 +1,145 @@
...
+UKPLAT_MEMRF_READ       = 0x0001   # Region is readable
+UKPLAT_MEMRF_WRITE      = 0x0002   # Region is writable
+UKPLAT_MEMRF_EXECUTE    = 0x0004   # Region is executable
+
+UKPLAT_MEMR_NAME_LEN    = 36
+
+# Boot info structure (see include/uk/plat/common/bootinfo.h)
+UKPLAT_BOOTINFO_SIZE    = 56
+
+UKPLAT_BOOTINFO_MAGIC   = 0xB001B0B0 # Boot Bobo
```
jschlumpp: If you change the magic value then change this instance one, too.

plat/xen/memory.c
marcrittinghaus: Yes, you could definitely improve that. It is just a temporary solution until we have incorporated the boot info also for Xen and linuxu. We'll keep it this way for now and update the platforms for 0.13.

plat/kvm/include/kvm-x86/multiboot.h
marcrittinghaus: Should we not just keep the definition? Optimally, the multiboot header would contain the definitions of the whole spec.

plat/common/include/uk/plat/common/bootinfo.h
marcrittinghaus: ups 🤣

plat/common/bootinfo.lds.S
marcrittinghaus: 👍🏻

plat/kvm/include/kvm-x86/multiboot.h
marcrittinghaus: Yes, absolutely. We are currently in a platform re-arch. I already laid out the files and most of the code so that we can easily move this into a dedicated `boot-multiboot` library. This would exist together with, for example, `boot-efi` etc. For the PR, I'll keep the header where it is and we move it out when we create the boot library.

Updates regarding multibootv2 we would then either incorporate in the same library and headers or have another one specific for v2.

plat/common/include/uk/plat/common/memory.h
marcrittinghaus: I will keep the `__u32` type since `i` is used multiple times with `list->count` which is `__u32`. But I added a cast `(int)i` for the return.

plat/common/include/uk/plat/common/memory.h
marcrittinghaus: Using the pointer would suggest to also change `ukplat_memregion_get()` which uses an index, too. And changing that leads to changes in quite a lot of places. Furthermore, using a ptr in `ukplat_memregion_list_delete()` etc. would also make it possible to supply memregion objects from a different list, which is hard to check. So I'll keep the index, if you don't mind.

plat/kvm/Linker.uk
```diff
@@ -1,5 +1,6 @@
 ifeq (x86_64,$(CONFIG_UK_ARCH))
 KVM_LDFLAGS-y += -Wl,-m,elf_x86_64
+KVM_LDFLAGS-y += -Wl,--entry _multiboot_entry
```
michpappas: Hi @marcrittinghaus, while I was looking at #603 I spotted this one.
```
LD      app-helloworld_kvm-x86_64.o
gcc: error: _multiboot_entry: No such file or directory
```
Seems that the right syntax is `--entry=_multiboot_entry`.
I'm using GCC 9.4.0.

marcrittinghaus: Thanks very much! This does not seems to be a problem with GCC 12. I updated the offending code.

plat/kvm/include/kvm-x86/multiboot.h
jschlumpp: Realistically, nobody will insert all definitions if you don't do it in this PR ;)
I'll just let you decide what to do.

marcrittinghaus: Fair point. I'll check out the specification and add missing pieces. It should not be that much.

plat/kvm/x86/setup.c
marcrittinghaus: Yes, makes sense. I added the `break` to avoid the additional level of indentation. It is perhaps also a bit easier to grasp.

lib/ukboot/boot.c
marcrittinghaus: Fair point, concerning merging of regions. I also thought about that but decided against it for now because it complicates the insert function. But I agree, if this actually happens, it is better to do merging. It would speed up physical memory management later on. As it will be a simple extension  I would keep it as-is for now and add it when we see this actually happening.

However, `ukplat_memregion_list_insert()` will not assume that the regions are ordered.  It just assumes ranges to not overlap but otherwise, it will sort them ascendingly by physical address. Did I miss something?

mogasergiu: Yes, my bad, I guess whatever comes before `_ukplat_entry` will ensure the mrd list is already sorted since it builds it using `ukplat_memregion_list_insert()` from the very beginning (like ``_multiboot_entry`` does). So, then, whatever support for other boot protocols that may come I guess will do the same thing as `_multiboot_entry` in this case, which would be building the mrd list ascendingly before calling `_ukplat_entry` (I guess the efi-stub should sort/coalesce them as well on its own before calling `_ukplat_entry`).

Regarding the coalescing part, perhaps there would be no need to add it in the insert function and instead do it only once in `_ukplat_entry` somewhere in the beginning before initializing the allocator, after the mrd list it has already been built by whatever called  `_ukplat_entry`. I guess it would be safe to assume those memory regions to be static after boot (unless an event like a memory hot-plug or something similar occurs).

But yeah, you are right, I guess it would be better to be left as it is now for the moment.

jschlumpp (changes requested):
I think the comment warnings by checkpath could be fixed.

lib/ukboot/boot.c
marcrittinghaus: Right, any other entry point (e.g. EFI, Linux Boot Protocol, Multiboot v2 etc.) would prepare the memory region list. We can think about merging once in `_ukplat_entry` as you said. Although, if we do it in `ukplat_memregion_list_insert()`, we only have to modify an existing entry instead of removing one (which means a ´memmove`). So probably it is not really faster. Especially, since you basically have to do the same checks.

We add this, when we add further protocols to the mix 👍🏻

plat/kvm/arm/memory.c
```diff
@@ -0,0 +1,183 @@
...
+#include <string.h>
+#include <uk/plat/common/sections.h>
+#include <sys/types.h>
+#include <uk/plat/memory.h>
+#include <uk/assert.h>
+#include <kvm/config.h>
+
+int ukplat_memregion_count(void)
+{
+	return (9
```
mogasergiu, line 31: Is it okay having these magic values here (same for `static struct ukplat_memregion_desc mrd[11];` below) instead of macro-definitions?

jschlumpp (approved):
Thanks!

mogasergiu (approved):

plat/kvm/arm/memory.c
marcrittinghaus: Yes, that is ugly. Totally agree. I just took the existing code and made compatibility fixes until we have updated the platforms to use boot info, too. So this is going to be replaced soon.

PR #723, opened 2023-01-19
support/qemu-guest: Enable X2APIC for TCG mode

skuenzer:
Enables X2APIC when TCG for x86 mode is selected. This enables running a recent x86_64 version of Unikraft on host devices without x86 hardware virtualization support (e.g., Arm).

### DIFF
--- a/support/scripts/qemu-guest
+++ b/support/scripts/qemu-guest
@@ -1,9 +1,10 @@
 #!/bin/bash
 #
-# Authors: Simon Kuenzer <simon.kuenzer@neclab.eu>
+# Authors: Simon Kuenzer <simon@unikraft.io>
 #
 # Copyright (c) 2019, NEC Laboratories Europe GmbH,
 #                     NEC Corporation All rights reserved.
+# Copyright (c) 2022, Unikraft GmbH
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


Sairajkodilkar:
Hi @skuenzer,
I am trying to test this PR on x86 by setting the `OPT_HWACCEL` flag to 1  so that qemu-guest will use TCG mode
but unikraft is unable to find x2apic and is crashing.

mogasergiu:
> Hi @skuenzer, I am trying to test this PR on x86 by setting the `OPT_HWACCEL` flag to 1 so that qemu-guest will use TCG mode but unikraft is unable to find x2apic and is crashing.

The same happens on my end, the issue seems to originate from the `cpuid` checking for `x2APIC` support and there is none to be found. If I am looking at this correctly, [qemu64](https://github.com/libvirt/libvirt/blob/master/src/cpu_map/x86_qemu64.xml) does not seem to support `x2APIC` by default. @skuenzer did it work on your end?

Edit: There also seems to be an opened [issue](https://gitlab.com/qemu-project/qemu/-/issues/330).

skuenzer:
Hum... 🤔 , sorry, I thought this made it working but it did in fact not. This means that our current SMP code does not run with TCG:
```
[    0.000000] CRIT: <<n/a>> {r:0x10fdb6,f:0x217ff0} [libkvmplat] <setup.c @  467> Failed to init bootstrap processor
```
I will close this PR since, in fact, QEMU does not emulate x2apic in TCG mode.

PR #725, opened 2023-01-20
lib/uknofault: Fault-safe memory access

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`, `arm64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This library provides support for performing fault-safe memory accesses and
`memcpy`s.

Normally, when the CPU attempts to access a memory address that is not
accessible or if the access would violate the protection settings of the
corresponding page, a page fault occurs. Similarly, on x86 the CPU raises a
general protection fault when a non-canonicalized virtual address is used.
Writing code that can potentially be given invalid addresses or which might
attempt accesses to inaccessible memory thus becomes very complicated.

This library provides functions for performing fault-safe memory accesses.
Faults will just cause the functions to indicate an error instead of crashing.

Generally, all functions are compiled as ISR-safe. `uknofault` also provides
variants that will in addition disable on-demand paging for duration of the
access or continue on failure (e.g., to make sure a whole region is not
accessible). These variants can be selected via the flags parameter.

## Example 1)
The following code checks if a given address range [0x100000; 0x100000+42) can
be read:
```C
#include <uk/nofault.h>

if (uk_nofault_probe_r(0x100000, 42, 0) != 42)
        uk_pr_err("Cannot read memory range\n");
```

## Example 2)
The following code performs a fault-safe `memcpy`. Both the `src` and `dst` are
allowed to cause faults:
```C
#include <uk/nofault.h>

char *dst = ...;
const char *src = ...;
__sz l, len = ...;

l = uk_nofault_memcpy(dst, src, len, 0);
if (l != len)
        uk_pr_err("Could only copy %ld bytes\n", l);
```

## Example 3)
The following code disables on-demand paging while performing the probe:
```C
#include <uk/nofault.h>

if (uk_nofault_probe_r(0x100000, 42, UK_NOFAULTF_NOPAGING) != 42)
        uk_pr_err("Cannot read memory range\n");
```


arch/x86/x86_64/include/uk/asm/traps.h
```diff
@@ -63,3 +63,5 @@ struct ukarch_trap_ctx {
 #define UKARCH_TRAP_MATH		trap_math
 
 #define UKARCH_TRAP_SECURITY		trap_security
+
+#define UKARCH_TRAP_X86_GP		trap_gp_fault
```
jschlumpp, line 67: Is there a reason why the name contains `X86`? The other defines do not contain that and the header is x86 specific anyways.

lib/uknofault/arch/arm64/include/arch/maccess.h
```diff
@@ -0,0 +1,68 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UKNOFAULT_ARCH_MACCESS_H__
+#define __UKNOFAULT_ARCH_MACCESS_H__
+
+#include "../../../../excpttab.h"
```
jschlumpp: It would be better to adjust the library-local include directories instead of navigating the directory structure in the `#include`.

lib/uknofault/arch/x86_64/include/arch/maccess.h
```diff
@@ -0,0 +1,68 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UKNOFAULT_ARCH_MACCESS_H__
+#define __UKNOFAULT_ARCH_MACCESS_H__
+
+#include "../../../../excpttab.h"
```
jschlumpp: Same as in the ARM header.

lib/uknofault/excpttab.h
```diff
@@ -0,0 +1,103 @@
...
+ * @param itr
+ *   Iterator variable (nf_excpt_handler *) which points to the individual
+ *   table entries during iteration
+ * @param ctortab_start
+ *   Start address of table (type: const nf_excpttab_entry[])
+ * @param ctortab_end
+ *   End address of table (type: const nf_excpttab_entry)
+ */
+#define nf_excpttab_foreach(itr, excpttab_start, excpttab_end)		\
+	for ((itr) = DECONST(struct nf_excpttab_entry*, excpttab_start);\
```
jschlumpp: What is the reason to remove the `const` here? The other code has no problem dealing with the `const` version.

lib/uknofault/excpttab.h
```diff
@@ -0,0 +1,103 @@
...
+}
+
+_NF_DECLARE_EXCPTTAB_ENTRY_GETTER(ip, __vaddr_t);
+_NF_DECLARE_EXCPTTAB_ENTRY_GETTER(handler, nf_excpt_handler);
+_NF_DECLARE_EXCPTTAB_ENTRY_GETTER(cont_ip, __vaddr_t);
+#undef _NF_DECLARE_EXCPTTAB_ENTRY_GETTER
+
+/* Exception table; provided by the linker script */
+extern const struct nf_excpttab_entry uk_excpttab_start[];
+extern const struct nf_excpttab_entry uk_excpttab_end;
```
jschlumpp: It would be more consistent to refer to the end using a incomplete array type.

lib/uknofault/maccess.c
```diff
@@ -0,0 +1,125 @@
...
+	nf_probe_r_loop(addr, l, __u8, FAULT_EXIT);
+FAULT_EXIT:
+	return len - l;
+}
+
+__sz uk_nofault_probe_r_nopage(__vaddr_t addr, __sz len)
+{
+	__sz r;
+
+	nf_disable_paging();
```
jschlumpp: Depending on how ukvmem works, it might be important to disable interrupts or alternatively ensure that paging is enabled for interrupt handlers.

lib/uknofault/tests/test_nofault.c
```diff
@@ -0,0 +1,135 @@
...
+ */
+UK_TESTCASE(uknofault, test_nofault_probe_r)
+{
+	struct uk_pagetable *pt = ukplat_pt_get_active();
+	__sz len;
+	int rc;
+
+	rc = ukplat_page_map(pt, TEST_MAP_BASE, __PADDR_ANY, 1,
+			     PAGE_ATTR_PROT_READ, 0);
+	UK_ASSERT(rc == 0);
```
jschlumpp: This is why I am not happy with the current uktest API: `UK_ASSERT` isn't optimal because you can disable assertions and then this check will just disappear, ignoring any failures silently.
But `UK_TEST_EXPECT_ZERO` also won't do the job, because that won't fail the test immediately. (google test had two macro variants `EXPECT` and `ASSERT` to distinguish between fatal and non-fatal checks...)

Maybe convert this to a `if (cond) { UK_CRASH(...); }` or `UK_TEST_EXPECT_ZERO(...); if (...} { UK_CRASH(...); }`

jschlumpp (commented):
I added some comments.

marcrittinghaus:
Hey @mschlumpp!

Thanks for the feedback. Very good suggestions 😃 I'll look into these probably tomorrow.

arch/x86/x86_64/include/uk/asm/traps.h
marcrittinghaus: The name contains the `X86` because the event is not available on any other architecture. The other events are meant to be generic and available on all architectures.

lib/uknofault/arch/arm64/include/arch/maccess.h
marcrittinghaus: Right! 👍🏻

lib/uknofault/arch/x86_64/include/arch/maccess.h
marcrittinghaus: Right 👍🏻

lib/uknofault/excpttab.h
marcrittinghaus: Right 👍🏻 This was a leftover from copying from [here](https://github.com/unikraft/unikraft/blob/979432ef14dba9e5235548bbdb0cf5ef00c6081e/include/uk/ctors.h#L101). Removed the `DECONST`.

lib/uknofault/excpttab.h
marcrittinghaus: Right 👍🏻 Don't know why that was the case, actually. Probably, from some early dev. Thanks for spotting 😃

lib/uknofault/maccess.c
marcrittinghaus: Disabling interrupts is not necessary. At least at the moment. Nevertheless, I also realized that it would be better to use a _save flags_ / _restore flags_ approach as it might be that paging is already disabled and you don't want to inadvertently enable it. So I changed the code so that we save and restore state. If you would need to also disable interrupts, either the VM implementation will take care or you have the flexibility here to do it, too, now.

Besides a `nopage` version, it is useful to have a version that continues to probe after the first fault. So you can easily ensure that a whole address range is **NOT** readable/writable (useful in tests, for example). 

But adding all the suffixes to the function names is not very practical. So I added a generic flags parameter. Especially, this allows having all the combinations.

lib/uknofault/tests/test_nofault.c
marcrittinghaus: Yep, makes sense. Until we add something like you suggested, I am using the `if (cond) { UK_CRASH(); }` solution.

marcrittinghaus:
Hey @mschlumpp,

I updated the library according to your suggestions + what I described in one of the comments above regarding the `flags` parameter.

lib/uknofault/tests/test_nofault.c
nderjung: @mschlumpp could you put together an issue describing the the limitations you see with `EXPECT` and `ASSERT` so we can keep track of this?

jschlumpp (approved):
Looks good, thanks!

lib/uknofault/tests/test_nofault.c
jschlumpp: Yes, will do

marcrittinghaus:
Just a minor documentation improvement. PR is ready from my side. Thanks @mschlumpp.

@nderjung Will you approve, or should we assign someone else?

lib/uknofault/tests/test_nofault.c
```diff
@@ -0,0 +1,170 @@
...
+	rc = uk_nofault_try_read(v2, __VADDR_INV);
+	UK_TEST_EXPECT_SNUM_EQ(rc, 0);
+
+	rc = uk_nofault_try_read(v2, &v);
+	UK_TEST_EXPECT_SNUM_EQ(rc, 1);
+	UK_TEST_EXPECT_SNUM_EQ(v2, v);
+}
+
+/**
+ * Tests the uk_nofault_try_write function
```
razvand, line 154:
```suggestion
 * Tests the uk_nofault_try_write function.
```

lib/uknofault/tests/test_nofault.c
```diff
@@ -0,0 +1,170 @@
...
+	/* Check sanity value */
+	UK_TEST_EXPECT_SNUM_EQ(*((unsigned long *)TEST_MAP_BASE), 0xdeadb0b0);
+
+	/* Clean up */
+	rc = ukplat_page_unmap(pt, TEST_MAP_BASE, 3, 0);
+	nf_bug_on(rc != 0);
+}
+
+/**
+ * Tests the uk_nofault_try_read function
```
razvand, line 137:
```suggestion
 * Tests the uk_nofault_try_read function.
```

lib/uknofault/excpttab.h
```diff
@@ -0,0 +1,101 @@
...
+_NF_DECLARE_EXCPTTAB_ENTRY_GETTER(cont_ip, __vaddr_t);
+#undef _NF_DECLARE_EXCPTTAB_ENTRY_GETTER
+
+/* Exception table; provided by the linker script */
+extern const struct nf_excpttab_entry uk_excpttab_start[];
+extern const struct nf_excpttab_entry uk_excpttab_end[];
+
+/**
+ * Helper macro to add an entry to the exception handler table in an inline
+ * asm block.
```
razvand:
```suggestion
 * asm block
```

razvand (commented):
Hi, @marcrittinghaus. See some minor comments from my part.

The issue is that it doesn't build for `arm64`:

```
  CC      libuknofault: maccess.isr.o
In file included from /home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c:15:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c: In function ‘nf_mf_handler’:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/arch/arm64/include/arch/maccess.h:66:33: error: ‘struct __regs’ has no member named ‘pc’
   66 | #define nf_regs_ip(regs) ((regs)->pc)
      |                                 ^~
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c:51:9: note: in expansion of macro ‘nf_regs_ip’
   51 |         nf_regs_ip(ctx->regs) = nf_excpttab_get_cont_ip(e);
      |         ^~~~~~~~~~
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c: In function ‘nf_mem_fault_handler’:
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/arch/arm64/include/arch/maccess.h:66:33: error: ‘struct __regs’ has no member named ‘pc’
   66 | #define nf_regs_ip(regs) ((regs)->pc)
      |                                 ^~
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c:60:31: note: in expansion of macro ‘nf_regs_ip’
   60 |         return nf_handle_trap(nf_regs_ip(ctx->regs), ctx);
      |                               ^~~~~~~~~~
/home/razvan/projects/unicore/unikraft-my/unikraft.git/lib/uknofault/maccess.c:61:1: warning: control reaches end of non-void function [-Wreturn-type]
   61 | }
      | ^
```

This is because the [`struct __regs` structure for aarch64](https://github.com/unikraft/unikraft/blob/staging/arch/arm/arm64/include/uk/asm/lcpu.h#L196) doesn't have a `pc` member. Did I miss something?

marcrittinghaus:
@razvand,

Thanks for pointing out these issues. I have updated the PR. It now also works on ARM.

razvand (approved):

PR #726, opened 2023-01-23
arm64/xen: Introduce xen support for aarch64 architecture

### LINKED ISSUE #62: Xen arm64 (aarch64) support

oleksiimoisieiev:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [`xen`]
 - Application(s): [N/A]

### Additional configuration

- `CONFIG_ARCH_ARM_64=y`
- `CONFIG_PLAT_XEN=y`
- `CONFIG_XEN_GNTTAB=y`                                                                                                                                                                                                
- `CONFIG_LIBXEN_XENBUS=y`  
- `CONFIG_XEN_KERNEL_HV_CONSOLE=y`
- `CONFIG_XEN_KERNEL_EMG_CONSOLE=y`
- `CONFIG_XEN_DEBUG_HV_CONSOLE=y`
- `CONFIG_XEN_DEBUG_EMG_CONSOLE=y`
- `CONFIG_LIBXEN_NETFRONT=y`
- `CONFIG_LIBXEN_BLKFRONT=y` 
- `CONFIG_LIBXEN_BLKFRONT_GREFPOOL=y`

### Description of changes

This is the implementation of the arm64 support for XEN hypervisor. Includes the following features:
- Implemented of the loaders and image header for arm64 architecture;
- Implemented all basic drivers, that are essential for work on XEN, such as: memory management, gnttab, xenbus and console drivers;
- Updated and tested frontend drivers for the devices: 9p, blkfront and netfront;
- Ported pvcalls-frontend driver from the Linux Kernel implementation,

GitHub-DependsOn: #1269


razvand:
Thanks, @oleksiimoisieiev. I've added tags, reviewers and assignees. We'll aim to get the work done for release 0.13 in late April.

oleksiimoisieiev:
Hello @razvand. Is there a way to run Codacy on my local repo? So I don't have to make many force-push to this MR?

oleksiimoisieiev:
The CI failure on arm-linuxu don't looks to be caused by this PR. Looking around, other PRs seem to have the same issue. When are you planning to start the review process?

razvand:
@oleksiimoisieiev, it's a known issue with the `arm-linuxu` build. We'll take care of it.

We'll start the review process these days.

Regarding Codacy running locally, I don't know, I didn't try. @nderjung, @StefanJum, did you try running Codacy locally or do you know how that should be done?

StefanJum:
@oleksiimoisieiev you can run codacy locally using the `codacy-cli` tool. Follow the steps [here](https://docs.codacy.com/related-tools/local-analysis/running-local-analysis/).

oleksiimoisieiev:
@StefanJum Thank you for the tip.

oleksiimoisieiev:
Hello @razvand. Do you have any updates with the review? Please feel free to contact me if you need any help with the review process or environment setup.

razvand:
@oleksiimoisieiev , we are looking for am ARM64 baremetal setup. Could you help out with that, so we can test it? Myself, @michpappas , @gabrielmocanu , are currently on vacation, we will start looking into the PR starting from next week.

oleksiimoisieiev:
Hello @razvand. I've used the following [Script](https://pastebin.com/SNgmaPAR) to create qemu environment on xen. I've downloaded it somewhere, but can't find the link. The only difference I made is I've used prebuilt xen binary. But you can use xen which is prepared by this script. The result will be quemuxen.img which is storage for Qemu.
This script I'm using to start testing vm: [Run](https://pastebin.com/qS5BjDXG)
Virtual machine can be accessed via ssh: ssh -p 2222 debian@localhost
To init xen tools I use the following commands:
`/etc/init.d/xen-watchdog start`
`/etc/init.d/xencommons start`
`/etc/init.d/xendomains start`
`/etc/init.d/xendriverdomain start`


To start helloworld application you need to create config hello.cfg:

`name          = 'hello'`
`vcpus         = '1'`
`memory        = '4'`
`kernel        = '/home/debian/helloworld_xen-arm64'`
`gic_version   = 'default'`
`boot          = 'c'`

And then start it with command:
`
xl create -c hello.cfg
`

razvand:
Thanks for pointing out the script, @oleksiimoisieiev. Did you also test it on an actual ARM64 physical platform?

oleksiimoisieiev:
@razvand Yes, we did. We have tested on 2 setups: R-Car Salvaltor-X H3 (RTP0RC7795SIPB0010S) and R-Car Gen4 Spyder(RTP8A779F0ASKB0SP2SA080)

razvand:
> @razvand Yes, we did. We have tested on 2 setups: R-Car Salvaltor-X H3 (RTP0RC7795SIPB0010S) and R-Car Gen4 Spyder(RTP8A779F0ASKB0SP2SA080)

Gotcha. We'll also aim to test it on a physical system.

oleksiimoisieiev:
Updated with the minor fix. Missed memory mapping parameter when rebased on the latest revision.

razvand:
@oleksiimoisieiev, we're looking into it, but expect some lag. It's a rather big contribution, and it will take a bit of ping-pong.

razvand (changes requested):
@oleksiimoisieiev, some minor items regarding commit messages.

In [this commit](https://github.com/unikraft/unikraft/pull/726/commits/35ea0d37f206282789dab78afbd2b78aedac799d), you should say STREAM, instead of STEAM.

Start the message description for this commits with capital letter:

https://github.com/unikraft/unikraft/pull/726/commits/9d566f41006e9912bee78129497558c38ce553f0
https://github.com/unikraft/unikraft/pull/726/commits/4ac1685c21f1f00c068aae349dfe675fb8f37959

razvand:
In [this commit](https://github.com/unikraft/unikraft/pull/726/commits/18dd725e05ef0f3f13817f53b325e93a63db2319) you have some inconsistencies:

* Use AArch64 throughout the commit message. You sometimes use arm64.
* `for the Unikrafs` -> `for Unikraft`
* `were moved` -> `were extracted`
* `and implements` -> `and implement`
* In the bullet list, use consistently present tense, remove semicolon symbols (`;`)
* `both elf and zboot` -> `both for ELF and ZBOOT formats`

oleksiimoisieiev:
@razvand Rebased and addressed yor comments

plat/xen/Config.uk
```diff
@@ -1,16 +1,16 @@
...
        depends on !HAVE_PAGING
        imply XEN_PV_BUILD_P2M
        select LIBUKDEBUG
        select LIBNOLIBC if !HAVE_LIBC
-       select LIBFDT if ARCH_ARM_32
+       select LIBFDT if (ARCH_ARM_32 || ARCH_ARM_64)
        select XEN_DBGEMERGENCY if ARCH_ARM_32
        help
-                Create a Unikraft image that runs as a Xen guest
+                Create a Unikraft image thans as a Xen guest
```
razvanvirtan: *that runs

plat/xen/Linker.uk
```diff
@@ -58,7 +64,9 @@ $(XEN_IMAGE).sym: $(XEN_DEBUG_IMAGE)
 ifeq (arm,$(CONFIG_UK_ARCH))
 $(XEN_RAW_IMAGE).gz: $(XEN_RAW_IMAGE)
 	$(call build_cmd,GZ,,$@, $(GZIP) -f -9 -c $< >$@)
-
+else ifeq (arm64,$(CONFIG_UK_ARCH))
```
razvanvirtan, line 77: Do we need an `ifeq` here? To me it seems like all the branches do the same thing (maybe I missed something, sorry)

plat/xen/arm/arch_time.c
```diff
@@ -106,16 +115,25 @@ __nsec ukplat_wall_clock(void)
...
 }
 
 void set_vtimer_compare(uint64_t value)
 {
 	uk_pr_debug("New CompareValue : %llx\n", value);
 
+#if defined(__arm__)
 	__asm__ __volatile__("mcrr p15, 3, %0, %H0, c14"
 			     ::"r"(value));
+#elif defined(__aarch64__)
```
razvanvirtan, line 137: Should this be empty?

plat/xen/arm/entry64.S
```diff
@@ -0,0 +1,708 @@
...
+	dsb     sy
+	mrs     x2, sctlr_el1
+	bic     x2, x2, #0x1
+	msr     sctlr_el1, x2
+	isb
+
+	/* Calculate where we are */
+	bl      _calc_offset
+
+	PRINT("- Mini-OS booting -\n")
```
razvanvirtan: I think we should replace `Mini OS` here with `Unikraft`

plat/xen/arm/link64.lds.S
```diff
@@ -0,0 +1,135 @@
...
+  INITTAB_SECTION
+
+  EVENTTAB_SECTION
+
+  _rodata = .;
+  .rodata : { *(.rodata) *(.rodata.*) }
+  . = ALIGN(4096);
+  _erodata = .;
+
+  /* newlib initialization functions */
```
razvanvirtan: Do we plan to un-comment this in the future? If yes. I suggest adding a TODO, otherwise I think we can simply remove this section.

plat/xen/Config.uk
```diff
@@ -75,6 +75,10 @@ config XEN_GNTTAB
 	help
 		Support grant table operations
 
+config UKPLAT_ALLOW_GIC
```
razvanvirtan: Isn't this a duplicate of [this](https://github.com/unikraft/unikraft/pull/726/commits/20c8dc8b96350c610a47a8563c5c8072b6f989a1#diff-fef3f72a280b4cfd00934570995e0ee66baca83b488d2ac2bc4ad16c0fbded0fR157)?

plat/xen/arm/mm.c
```diff
@@ -0,0 +1,139 @@
+#include <xen-arm/os.h>
```
razvanvirtan, line 27: I think this file has a missing SPDX header

plat/xen/include/xen/intctrl.h
```diff
@@ -0,0 +1,37 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
razvanvirtan: This seems like a duplicate of [this source file](https://github.com/unikraft/unikraft/blob/staging/plat/kvm/include/kvm/intctrl.h). Maybe @michpappas has a stronger view if we should duplicate or move the file on an upper level

plat/xen/arm/irq.c
```diff
@@ -0,0 +1,165 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+/* Moved from Mini-OS */
```
razvanvirtan: Is this from Mini-OS?

razvanvirtan (commented):
Hi @oleksiimoisieiev! A first round of comments based on the first commits, I will complete the review tomorrow.

Besides the inline comments:
- in https://github.com/unikraft/unikraft/pull/726/commits/20c8dc8b96350c610a47a8563c5c8072b6f989a1 there is a typo in the commit message `wait forewer`

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -371,6 +371,13 @@ static inline unsigned long ukarch_read_sp(void)
 	return sp;
 }
 
+#if defined(__aarch64__)
```
mariasfiraiala: Is the guard needed? The header is already meant to be used for `AArch64` usecases.

lib/ukring/include/uk/ring.h
```diff
@@ -321,10 +321,10 @@ uk_ring_peek_clear_sc(struct uk_ring *br)
 	if (br->br_cons_head == br->br_prod_tail)
 		return NULL;
 
-#if defined(CONFIG_ARCH_ARM_32) || defined(CONFIG_ARCH_ARM_64)
+#if defined(CONFIG_ARCH_ARM_32)
 	/*
 	 * The barrier is required there on ARM and ARM64 to ensure, that
-	 * br->br_ring[br->br_cons_head] will not be fetched before the above
+	 * br->br_risung[br->br_cons_head] will not be fetched before the above
```
mariasfiraiala: As far as I can tell `br_risung` doesn't constitute a member of the `uk_ring` struct. Is this intended?

plat/xen/arm/gnttab.c
```diff
@@ -56,7 +56,7 @@ static paddr_t get_gnttab_base(void)
 		BUG();
 	}
 
-	gnttab_base = fdt64_to_cpu(regs[0]);
+	gnttab_base = fdt64_ld(regs);
```
mariasfiraiala, line 59: It isn't clear enough from the [commit message](49e9a59dc44a7ef352eeb772431af4afdc0364e9) the need to replace `fdt64_to_cpu()` with `fdt64_ld()`. Due to the fact that they are so similar, I think we should add a bit of description to explain the rationale behind the choice.

plat/common/arm/traps_arm64.c
```diff
@@ -161,8 +161,8 @@ static void dump_registers(struct __regs *regs, __u64 far)
 	int idx;
 
 	uk_pr_crit("\t SP       : 0x%016lx\n", regs->sp);
-	uk_pr_crit("\t ESR_EL1  : 0x%016lx\n", regs->esr_el1);
-	uk_pr_crit("\t ELR_EL1  : 0x%016lx\n", regs->elr_el1);
+	uk_pr_crit("\t SPSR_EL1  : 0x%016lx\n", regs->esr_el1);
```
mariasfiraiala: It is unclear to me why you decided to rename the `ESR_EL1` value to `SPSR_EL1`. AFAIK they represent different things:

`ESR_EL1` indicates the reason the trap was caught and we use it on a wide scale when debugging.

`SPSR_EL1` holds the saved process state when an exception is taken. I can see why you chose to print its value instead of the 0 saved in `ELR_EL1`. I don't see the point of renaming `ESR_EL1`.

plat/xen/arm/entry64.S
```diff
@@ -317,12 +317,11 @@ ENTRY(_libxenplat_setup_cpu)
 	 *   NORMAL_WB               11111111 (0xff)
 	 *   NORMAL_WT               10111011 (0xbb)
 	 */
-	ldr     x0, =(MAIR_DEVICE_nGnRnE | \
-	             MAIR_DEVICE_nGnRE | \
-	             MAIR_DEVICE_GRE | \
-	             MAIR_NORMAL_NC | \
-	             MAIR_NORMAL_WB | \
-	             MAIR_NORMAL_WT)
+	ldr     x0, =(MAIR(MAIR_DEVICE_nGnRnE, MT_DEVICE_nGnRnE) | \
```
mariasfiraiala, line 320: Please check the [commit message](3bb361d9d35c682ab53ccf92afd1d0a80dde1aa0) and replace "which is made" with "which **are** made". Also, use capital letter after a full stop.

plat/xen/Config.uk
```diff
@@ -102,6 +102,12 @@ config XEN_NETFRONT
 	help
 		Driver for netfront devices
 
+config XEN_PVCALLS
```
mariasfiraiala: In the [commit message](0f945326f5711ff00c5b460d0c82d4ab0c5ef3d5) remove the slash before `plat` and reword it  to sound a bit better:
```
The pvcalls driver is the communication mechanism for the domain,
when no ethernet connection is supported. It replaces the AF_INET
socket type and can be accessed through socket controlling syscalls.
```

include/uk/arch/types.h
```diff
@@ -251,6 +251,8 @@ typedef __u64 __paddr_t;
 #define __NULL ((void *) 0)
 #endif
 
+typedef char *caddr_t; /* core address */
```
mariasfiraiala, line 254: Is this `typedef` really needed? I see we only have [one instance](https://github.com/unikraft/unikraft/blob/staging/lib/ukring/ring.c#L49) where it is used and I'd rather have it changed there with `char *` than have a completely new type defined.

mariasfiraiala (changes requested):
Thanks @oleksiimoisieiev, I've taken a first look at the changes. They look good, you'll find some suggestions from me below.

One thing I wish to have amended is the wording for the commit messages. You have some examples from me below, but pretty please go through them again and modify them a bit.

I'll also take another look at the PR this night or tomorrow morning. Great work so far :muscle:

plat/xen/arm/arch_events.c
```diff
@@ -39,12 +44,28 @@ static void virq_debug(evtchn_port_t port __unused,
...
 void arch_init_events(void)
 {
+	int rc;
+
 	debug_port = bind_virq(VIRQ_DEBUG, (evtchn_handler_t)virq_debug, 0);
 	if ((int) debug_port == -1)
 		UK_CRASH("Failed to initialize events\n");
 	unmask_evtchn(debug_port);
+
+	/* Register the lcpu_run and lcpu_wakeup interrupt handlers */
```
razvanvirtan: This comment seems wrong, maybe `Register the the events interrupt handler`?

plat/xen/drivers/pvcalls/pvcalls-front.c
```diff
@@ -0,0 +1,1602 @@
...
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUD
```
razvanvirtan: Remove the blank space here

plat/xen/arm/entry64.S
razvanvirtan: I would also add always capitalizing `MAIR`

plat/common/arm/traps_arm64.c
```diff
@@ -161,8 +161,8 @@ static void dump_registers(struct __regs *regs, __u64 far)
...
+	uk_pr_crit("\t ELR_EL1  : 0x%016lx\n", regs->spsr_el1);
```
razvanvirtan: The commit message format is not followed here, please check [the unikraft guidelines](https://unikraft.org/docs/contributing/submitting-changes/)

plat/xen/drivers/9p/9pfront.c
```diff
@@ -46,8 +46,15 @@
 #include <uk/9preq.h>
 #include <uk/9pdev_trans.h>
 #include <uk/plat/spinlock.h>
+#if defined(__i386__) || defined(__x86_64__)
```
razvanvirtan, line 49: Please add `.` at the end of the sentence (after `aarch64`) in the commit message

plat/xen/arm/setup64.c
```diff
@@ -300,6 +300,13 @@ void _libxenplat_armentry(void *dtb_pointer, paddr_t physical_offset)
 	if (unlikely(r))
 		UK_CRASH("Failed to initialize bootstrapping CPU: %d\n", r);
 
+#ifdef CONFIG_HAVE_SMP
+	ret = lcpu_mp_init(CONFIG_UKPLAT_LCPU_RUN_IRQ,
```
razvanvirtan, line 353: I think the `-` in the commit message should be replaced with a comma

plat/xen/arm/gnttab.c
```diff
@@ -34,6 +34,8 @@
 #include <xen-arm/os.h>
 #include <xen-arm/mm.h>
 
+extern lpae_t fixmap_pgtable[512];
```
razvanvirtan, line 37: Please capitalize at the start of the sentence in the commit message (`That's why..`)

include/uk/essentials.h
```diff
@@ -104,6 +104,11 @@ extern "C" {
 #define __alias(old, new) \
 	extern __typeof(old) new __attribute__((alias(#old)))
 #endif
+
```
razvanvirtan: Missing `.` at the end of the commit message

razvanvirtan (changes requested):

plat/xen/arm/irq.c
oleksiimoisieiev: Yes. I took Mini-OS as base for my work. Original files were moved from there, although they were heavily modified. If this comment is wrong I can simply remove it.

plat/xen/Config.uk
oleksiimoisieiev: fixed

plat/xen/Linker.uk
oleksiimoisieiev: for arm and arm64 it uses XEN_RAW_IMAGE, that's why this if was added in the first place.

plat/xen/arm/arch_time.c
oleksiimoisieiev: Just double-checked. This is empty in mini-OS implementation, which I took as a base. DDI0487H says MCRR is for aarch32 architecture.

plat/xen/arm/entry64.S
oleksiimoisieiev: fixed

plat/xen/arm/link64.lds.S
oleksiimoisieiev: removed

plat/xen/Config.uk
oleksiimoisieiev: fixed

plat/xen/arm/mm.c
oleksiimoisieiev: fixed

plat/xen/include/xen/intctrl.h
oleksiimoisieiev: Hm. It is. But moving this to the upper level will require reworking kvm code as well. that's why I've added my implementation. Let's ask @michpappas how it can be handled.

arch/arm/arm64/include/uk/asm/lcpu.h
oleksiimoisieiev: Removed

lib/ukring/include/uk/ring.h
oleksiimoisieiev: removed

plat/xen/arm/gnttab.c
oleksiimoisieiev: It's correct call to generate memory-aligned value. fdt_to_cpu accepts value while fdt64_ld accepts pointer, which potentially can be an array. For example libfdt using fdt64_ld_ function to parse addresses (although it's a wrapper around fdt64_to_cpu. I won't say it wouldn't work but I think this is safer.

plat/common/arm/traps_arm64.c
oleksiimoisieiev: I've met this during debug sessions when spsr and elr were swapped. Maybe I've just interpreted it wrong. I would drop this commit, but this situation worth a deeper investigation.

plat/xen/arm/entry64.S
oleksiimoisieiev: fixed

plat/xen/Config.uk
oleksiimoisieiev: Thanks. Should improve my English. Fixed

plat/xen/arm/gnttab.c
mariasfiraiala: Nice thing, please add this explanation to the commit message.

plat/common/arm/traps_arm64.c
mariasfiraiala: I've also attached `gdb` just to confirm this, and from my past experience debugging things on `AArch64`, `ELR_EL1` is usually simply 0:

```
pwndbg> p *regs
$2 = {
  x = {0, 1207451680, 1076246672, 0, 0, 0, 0, 0, 0, 33554432, 21, 25, 10, 150994944, 32, 4, 0, 0, 1207566360, 1076246928, 1076246944, 0, 0, 0, 0, 0, 0, 0, 0, 1076246896},
  lr = 0,
  elr_el1 = 0,
  spsr_el1 = 837,
  esr_el1 = 2248146950,
  sp = 1076246608,
  sp_el0 = 1075219168
}

pwndbg> i r ESR_EL1
ESR_EL1        0x86000006          2248146950
pwndbg> i r ELR_EL1
ELR_EL1        0x0                 0
pwndbg> i r SPSR_EL1
SPSR_EL1       0x345               837
```

I don't have a problem with printing for `ELR_EL1` the value stored in `SPSR_EL1`. Just don't name the `ESR_EL1` value as `SPSR_EL1`.

plat/xen/include/xen/intctrl.h
michpappas: Let's leave it as is for `0.14`, we'll be introducing `ukintctlr` in `0.15` (hopefully) and we can consider that then.

oleksiimoisieiev: ok

plat/xen/arm/gnttab.c
oleksiimoisieiev: added

plat/common/arm/traps_arm64.c
oleksiimoisieiev: Yeah. I agree. This is wrong - i would drop this commit.

include/uk/arch/types.h
oleksiimoisieiev: This type is used in ukring and was added in commit: 7f212b4df81d54fca4b647e09f2c52082875f707 and I've met a compilation error due to the missing definition. I assumed this type to be architecture dependent that's why I've added type definition in arch/types.h, rather then changed ukring library with unknown (for me at least) side effects on other platforms.

plat/xen/arm/arch_events.c
oleksiimoisieiev: fixed

plat/common/arm/traps_arm64.c
oleksiimoisieiev: this commit was removed

plat/xen/drivers/pvcalls/pvcalls-front.c
oleksiimoisieiev: fixed

plat/xen/drivers/9p/9pfront.c
oleksiimoisieiev: fixed

plat/xen/arm/setup64.c
oleksiimoisieiev: fixed

plat/xen/arm/gnttab.c
oleksiimoisieiev: fixed

include/uk/essentials.h
oleksiimoisieiev: fixed

plat/common/arm/time.c
```diff
@@ -51,6 +51,13 @@ static const char * const arch_timer_list[] = {
 	NULL
 };
 
+#if defined(CONFIG_PLAT_XEN)
+extern void *HYPERVISOR_dtb;
+#define _DTB_PTR HYPERVISOR_dtb
+#else
+#define _DTB_PTR _dtb
```
mariasfiraiala: I am not sure about the value for the device tree when the platform isn't Xen. How is this `_dtb` value initialized?

mariasfiraiala (changes requested):
Thanks @oleksiimoisieiev  for the changes, see my one question from below and also some more suggestions regarding commit messages:

* 40eba5591f2bc804e6714fc76838c51059e6aace:

```
plat/xen/arm: Use fdt64_ld function instead of fdt64_to_cpu

The `fdt64_ld()` function generates correct memory-aligned values.
This call is safer than `fdt_to_cpu()` because it accepts a pointer,
which potentially can be an array. For example `libfdt` is
using the `fdt64_ld()` function to parse addresses (although it's
a wrapper around `fdt64_to_cpu()`).
```

* 4fd7ac4a1330b581a3e2bb616a33253e45d8ce61:

```
plat/xen/arm: Fix `MAIR` configuration

`MAIR` values should be programed to different slots, which are designated
by the `MAIR` macro introduced in commit 699a985.
The macro moves the MAIR_* mask left on MT_* * 8 bits so the `MAIR`
masks won't overlap.
```

* 3922578600c0282682d8f668386515fc60bc3de6:

Still didn't remove the trailling slash before `plat`

* f0341d0914231c4043b119afcbb2283bbd314488:

```
uk/ring: Implement `ukarch_thread_fence_acq()`

Add implementation of the `thread_fence_acq()` function, needed for
the `AArch64` ring buffer.
```

* 40d991644d1f954ee1e8774898923c73749283e2:

```
uk/types.h: Add `caddr_t` declaration
```

* 45f8ab75f130b1830309f3bf7d90003e0fad85ef:

```
uk/essentials.h: Add `__aligned` implementation

Implement the `__aligned` attribute.
```

* 4f3b05a90bd6ea0a62e04b0855fb096292e99968:

```
plat/xen/arm: Map grant table region and remove setup_table call

When reading from the DTB, the grant table has an address outside of
the memory region allocated for the domain. That's why the grant table
region should be mapped to the page table.

We also remove the setup_table call because it shouldn't be called with
XENMEM_add_to_physmap.
```

* 30420c786daf003e17e289ee4f6e488d80c784eb:

```
plat/xen/arm: Move `set_gnt_entry()` function to header

Move the `set_gnt_entry()` function to common place in order to prevent
duplicating code.
```

* 499965dfa234d8e6b58c035c1a1aa1d0c5b64764:

```
plat/xen: Update netfront and 9p drivers for AArch64

Adjust the netfront and 9p drivers to compile and run on `AArch64`.
```

* 61b045049df88bcfcb57d25ec101428480375684:

```
plat/xen: Fix issue when `__XEN_INTERFACE_VERSION__` is empty

Added check to `__XEN_INTERFACE_VERSION__` being empty. In this case we
should set a default value.
```

* 6add1a6f5de43aee4e3379eeb1728669ac7d8e23:

```
plat/xen: Add generic timer support for AArch64

`arch_timer` implemented for `AArch32` is not suitable for `AArch64`
because it doesn't use interrupts and, as a result, the system waits
forever during sleep call.

We add support for generic interrupt based timer.
```

* 5f059bfcae1fd78890b5b5ff2408dc1ef1638fb8:

```
plat/xen/arm: Interrupts and grant tables support for AArch64

Add implementation of the grant table support and interrupt support.

Partly moved from Mini-OS and from plat/xen/kvm. Interrupt support is
adapted to use gicv2 and gicv3.
```

* 699a98550cc80ba617559ffd41f6a560b57410b8:

```
plat/xen/arm: Implementation of the AArch64 loader support

Changes in this commit were extracted from Mini-OS and implement
AArch64 loader support for Unikraft.

The features we add are the following:

 * create unikernel binary both in ELF and ZBOOT format
 (include ZBOOT entry)
 * initialization of the basic hardware
 * enable emergency console to show program output
 * modify configuration to allow selecting XEN guest for `AArch64`
 * event channels implementation
 * hypervisor console implementation
```

plat/common/arm/time.c
```diff
@@ -51,6 +51,13 @@ static const char * const arch_timer_list[] = {
 	NULL
 };
 
+#if defined(CONFIG_PLAT_XEN)
```
razvanvirtan: I suggest a bit of the re-phrasing in the commit message:
```
The arch_timer implemented for arm32 is not suitable for aarch64 because
it doesn't use interrupts, so it makes system to wait forever during sleep
call.
Add support for generic interrupt based timer.
```

plat/xen/include/xen/xen-compat.h
```diff
@@ -29,11 +29,16 @@
 
 #define __XEN_LATEST_INTERFACE_VERSION__ 0x00040900
 
+#define DO_EXPAND(VAL) (VAL##1)
```
razvanvirtan, line 32:
```
In this case a default value should also be set.
```

plat/xen/drivers/9p/9pfront.c
```diff
@@ -46,8 +46,15 @@
...
 #include <xen-x86/mm.h>
 #include <xen-x86/irq.h>
+#elif defined(__aarch64__)
```
razvanvirtan, line 52: One typo: `Adaptatioqn`

plat/xen/arm/gnttab.c
razvanvirtan, line 37: To me, this sentence is a bit unclear, could you please rephrase it?
```
grant_table region, reading from the DTB has an address, which is
outside of the memory region, allocated for the domain.
```

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -371,6 +371,11 @@ static inline unsigned long ukarch_read_sp(void)
 	return sp;
 }
 
+static inline void ukarch_thread_fence_acq(void)
```
razvanvirtan: Missing `.` at the end of commit message.

plat/xen/Makefile.uk
```diff
@@ -183,6 +183,7 @@ LIBXENPVCALLS_ASINCLUDES-y    += $(LIBXENPLAT_ASINCLUDES-y)
 LIBXENPVCALLS_CFLAGS-y        += $(LIBXENPLAT_CFLAGS-y)
 LIBXENPVCALLS_CINCLUDES-y     += $(LIBXENPLAT_CINCLUDES-y)
 LIBXENPVCALLS_SRCS-y          += $(LIBXENPLAT_BASE)/drivers/pvcalls/pvcalls-front.c
+LIBXENPVCALLS_SRCS-y          += $(LIBXENPLAT_BASE)/drivers/pvcalls/pvcalls-socket.c
```
razvanvirtan: Typo in last sentence: `Pvcalls`

plat/xen/arm/entry64.S
```diff
@@ -317,12 +317,11 @@ ENTRY(_libxenplat_setup_cpu)
...
+	             MAIR(MAIR_DEVICE_nGnRE, MT_DEVICE_nGnRE) | \
```
razvanvirtan, line 321: I think the last sentence can be re-phrased to be more clear:
```
This commit moves MAIR_* mask left on MT_* * 8 bits so that MAIR masks won't overlap.
```

razvanvirtan (changes requested):
@oleksiimoisieiev Besides suggestions regarding some commit messages that I added inline, please also make sure that all the commit titles follow the Unikraft commit selector convention:
https://unikraft.org/docs/contributing/submitting-changes/#commit-message-format

plat/common/arm/time.c
oleksiimoisieiev: Refactored this place. Looks like it was changed after my changes.

klogg:
@razvand @nderjung What do we need to do to push this? Somehow this is stuck for way too long...

razvand:
> @razvand @nderjung What do we need to do to push this? Somehow this is stuck for way too long...

Hi, @klogg , @oleksiimoisieiev . Conflicts need to be solved first. It would be great if @oleksiimoisieiev were to do that and retest if everything works.

The plan would be to have this integrated in the next release, scheduled for December 18, 2023.

I will also arrange for a reboot of our Unikraft-EPAM calls to push on this.

razvand:
@oleksiimoisieiev , could you please rebase this PR and see our discussion on Discord: https://discord.com/channels/762976922531528725/1186531310479626250

NathanRignall:
Hello, tested this PR using [app-helloword](https://github.com/unikraft/app-helloworld) and kraftkit (0.7.1) with gcc-aarch64-linux-gnu (11.4.0)

The build fails with the following errors...
`make[2]: *** No rule to make target '/home/nathan/Documents/unikraft/plat/drivers/gic/gic-common.c', needed by '/home/nathan/Documents/app-helloworld/.unikraft/build/libxengic/gic-common.o'. Stop.`
`make[1]: *** [Makefile:1135: sub-make] Error 2`
`make: *** [Makefile:41: _all] Error 2`

xen/x86_64, qemu/x86_64, qemu/arm64 build successfully in this configuration

plat/xen/Config.uk
```diff
@@ -31,27 +34,27 @@ menu "Console Options"
 config XEN_KERNEL_HV_CONSOLE
 	bool "Hypervisor console for kernel prints"
 	default y
-	depends on ARCH_X86_64
+	depends on (ARCH_x86_64 || ARCH_ARM_64)
```
andreistan26:
```suggestion
	depends on (ARCH_X86_64 || ARCH_ARM_64)
```

plat/xen/arm/link64.lds.S
```diff
@@ -0,0 +1,118 @@
...
+	*(.exitcall.exit)
+  }
+
+  /* Stabs debugging sections.  */
+  .stab 0 : { *(.stab) }
+  .stabstr 0 : { *(.stabstr) }
+  .stab.excl 0 : { *(.stab.excl) }
+  .stab.exclstr 0 : { *(.stab.exclstr) }
+  .stab.index 0 : { *(.stab.index) }
+  .stab.indexstr 0 : { *(.stab.indexstr) }
```
andreistan26:
```suggestion
  DEBUG_SYMBOLS
```

andreistan26: Any reason why here we have `STABS` and not `DWARF`?

plat/xen/Config.uk
andreistan26: Typo `x` -> `X`.

lib/posix-socket/socket.c
```diff
@@ -341,9 +341,14 @@ int uk_sys_accept(const struct uk_file *sock, int blocking,
 	_socket_init(al, n->driver, new_data);
 
 	al_listener = __containerof(sock, struct socket_alloc, f);
+#if CONFIG_LIBPOSIX_SOCKET_EVENTS
```
michpappas: Perhaps @andreittr could have a look at this change.

plat/Config.uk
```diff
@@ -119,3 +119,25 @@ config FPSIMD
 config UKPLAT_AUXSP_PAGE_ORDER
 	int "Maximum size allocated for the auxiliary stack"
 	default 4
+
```
michpappas: Since we introduced `libukintctlr` this should not be needed. The proper way is to add `HAVE_INTCTLR` in `plat/xen/Config.uk`, and then `imply LIBUKINTCTLR_GICV[2,3]` for the GIC version you would like to have as the default, which you have done already 👍🏼 So this can go away, along with `UKPLAT_ALLOW_GIC` in `plat/kvm/Config.uk`.

plat/kvm/Config.uk
```diff
@@ -170,4 +170,8 @@ config RTC_PL031
        default y if ARCH_ARM_64
        depends on ARCH_ARM_64
 
+config UKPLAT_ALLOW_GIC
```
michpappas: This can go away, see the relevant comment on `plat/Config.uk`

plat/xen/Makefile.uk
```diff
@@ -89,14 +90,19 @@ LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_32) += $(LIBXENPLAT_BASE)/arm/arch_time.c
...
-LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(LIBXENPLAT_BASE)/arm/arch_time.c
+LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(LIBXENPLAT_BASE)/arm/mm.c
 LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(LIBXENPLAT_BASE)/arm/hypercalls64.S
+LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/lcpu.c|arm64_common
+LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/traps_arm64.c
+
 ifeq ($(CONFIG_HAVE_SMP),y)
 LIBXENPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_COMMON_BASE)/arm/lcpu_start.S
 endif
+ifeq ($(CONFIG_UKPLAT_ALLOW_GIC),y)
```
michpappas: This can go away, see the relevant comment in `plat/Config.uk`. And since `plat/xen/arm/irq.c` is not compiled anyway (`ALLOW_GIC` was no longer set in `plat/xen/Config.uk`) maybe it can go away too.

plat/xen/arm/link64.lds.S
```diff
@@ -0,0 +1,130 @@
...
+
+  EXCEPTION_SECTIONS
+
+  CTORTAB_SECTION
+
+  INITTAB_SECTION
+
+  EVENTTAB_SECTION
+
+  _ctors = .;
```
andreistan26, line 74:
```suggestion
  . = ALIGN(0x8)
  _ctors = .;
```

andreistan26: This should be aligned to 8 bytes.

include/uk/essentials.h
```diff
@@ -45,6 +45,10 @@ extern "C" {
 #define __align256b    __align32
 #define __align512b    __align64
 
+#ifndef __aligned
+#define __aligned(x) __attribute__((__aligned__(x)))
```
michpappas: We recently moved the compiler definitions from `essentials.h` to `compiler.h`. Maybe you can use `_align(x)` from `include/uk/compiler.h` instead?

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -185,6 +185,13 @@ int uk_intctlr_irq_fdt_xlat(const void *fdt, int nodeoffset, __u32 index,
 
 #endif /* __ASSEMBLY__ */
 
+#if defined(__aarch64__) && defined(CONFIG_PLAT_XEN)
```
michpappas: It would be great if we could avoid platform-specific changes in generic code, and fulfill any platform-specific requirements by generic additions to the `ukintctlr` API.

I post here the consensus from our discussion on discord:
* Declare a function in `lib/ukintctlr/include/uk/intctlr.h` for platform-specific initialization, eg `uk_intctlr_platform_init(void *)` or similar, along with documentation on who calls it, with what parameters, and when (details below).
* That function will be invoked by `uk_intctlr_probe()`, which is implemented by the driver, once the device is discovered by the firmware and before its memory regions are mapped (I am using a more generic description that fdt and GIC as that function can be invoked by other drivers). The driver will pass driver-specific data, that are declared in the driver's public API (eg `drivers/ukintctlr/gic/include/uk/`).
* The default implementation can be defined in `libukintctlr` as a `__weak` function that returns `0`. Platforms that require this functionality can override the weak definition with their own.
* In the case where other platforms require hooks at other stages of probing, the scheme can be extended with additional functions (eg `uk_intctlr_platform_init0()`, `uk_intctlr_platform_init1()` and so on).

michpappas (changes requested):
Hi @oleksiimoisieiev, thank you very much for updating this (once again), and apologies for taking so long to review. Besides the inline comments, please move the Xen drivers to the newly introduced `drivers/` subsystem at the top level directory (i.e. `unikraft/drivers/xen/`). You can have a look at the the top-level `Makefile.uk` and the rest of the drivers for the general layout and directives you need to define.

Also, the commit titled "DNM fixes" contains mixed changed, but perhaps you are planning to address that in the next round.

andreistan26 (changes requested):
Hi @oleksiimoisieiev, thanks for the update. I have only a few minor things to get re checked. I've also tested `x86-64` and it works as before, so no problems there.

oleksiimoisieiev:
> Hi @oleksiimoisieiev, thank you very much for updating this (once again), and apologies for taking so long to review. Besides the inline comments, please move the Xen drivers to the newly introduced `drivers/` subsystem at the top level directory (i.e. `unikraft/drivers/xen/`). You can have a look at the the top-level `Makefile.uk` and the rest of the drivers for the general layout and directives you need to define.
> 
> Also, the commit titled "DNM fixes" contains mixed changed, but perhaps you are planning to address that in the next round.

Hi @michpappas. As I can see staging branch already has xen related drivers in /plat/xen/drivers directory. So I will move them in the separate commit. DNM commit will be removed. I've just added it to simplify testing.

plat/kvm/Config.uk
oleksiimoisieiev: fixed

plat/Config.uk
oleksiimoisieiev: fixed

plat/xen/Makefile.uk
oleksiimoisieiev: fixed

include/uk/essentials.h
oleksiimoisieiev: fixed

lib/ukintctlr/include/uk/intctlr.h
oleksiimoisieiev: fixed

plat/xen/Config.uk
oleksiimoisieiev: thanks. Fixed

plat/xen/arm/link64.lds.S
oleksiimoisieiev: This came from Mini-OS. Wasn't looking deep into it. Fixed

plat/xen/arm/link64.lds.S
oleksiimoisieiev: thanks. fixed

plat/kvm/Config.uk
```diff
@@ -170,8 +170,4 @@ config RTC_PL031
        default y if ARCH_ARM_64
        depends on ARCH_ARM_64
 
-config UKPLAT_ALLOW_GIC
```
michpappas: This commit can be merged to the one that introduces the change, i.e. bd8af6dd7. No need to introduce and later revert a change in a given series.

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -45,6 +45,17 @@ struct uk_intctlr_irq {
 	unsigned int trigger;
 };
 
+/** Platform specific initialization data */
```
michpappas: This is GIC-specific, so it should be provided by the GIC driver. A more appropriate place would be `drivers/ukintctlr/gic/include/uk/gic.h`. The GIC driver exposes its headers to the rest of Unikraft, so you should be able to include `uk/intctlr/gic.h` from `plat/xen/`.

drivers/ukintctlr/gic/gic-v3.c
```diff
@@ -682,8 +683,15 @@ static int gicv3_do_probe(void)
 		return r;
 	}
 
-	ukplat_irq_setup(gicv3_drv.dist_mem_addr, gicv3_drv.rdist_mem_addr,
-		&gicv3_drv.dist_mem_addr, &gicv3_drv.rdist_mem_addr);
+	in.dist_addr = gicv3_drv.dist_mem_addr;
+	in.rdist_addr = gicv3_drv.rdist_mem_addr;
+	r = uk_intctlr_plat_probe(&in, &out);
+	if (r) {
```
michpappas: Please use `if unlikely(r)` here.

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -597,8 +598,15 @@ static int gicv2_do_probe(void)
 		return r;
 	}
 
-	ukplat_irq_setup(gicv2_drv.dist_mem_addr, gicv2_drv.cpuif_mem_addr,
-			 &gicv2_drv.dist_mem_addr, &gicv2_drv.cpuif_mem_addr);
+	in.dist_addr = gicv2_drv.dist_mem_addr;
+	in.rdist_addr = gicv2_drv.cpuif_mem_addr;
+	r = uk_intctlr_plat_probe(&in, &out);
+	if (r) {
```
michpappas: Please use `if unlikely(r)` here.

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -183,14 +194,21 @@ int uk_intctlr_irq_free(unsigned int *irqs, __sz count);
...
-#endif /* __ASSEMBLY__ */
+/**
+ * Platform-specific initialization function for interrupts
+ *
+ * If defined by the platform code then additional platform-specific
+ * initialization steps should be performed.
+ * For example some architectures with mmu support requires address
+ * mapping to access registers.
+ *
+ * @param in pointer to the uk_intctlr_plat_data structure
```
michpappas: Since we are moving the definition of `struct uk_intctlr_plat_data` to the driver, please also update to:

"driver-provided data to be interpreted by the platform. This structure is defined by the driver."

Also below:

"platform data that should be interpreted by the driver."

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -183,14 +194,21 @@ int uk_intctlr_irq_free(unsigned int *irqs, __sz count);
...
+ * @param out result structure that should be interpreted by interrupt
+ * 	controller
+ */
+int uk_intctlr_plat_probe(struct uk_intctlr_plat_data *in, void *out);
```
michpappas: Would it be more flexible to have both parameters defined as `void *` and let the driver choose what to define them to? An alternative way would be to have `struct uk_intctlr_driver_data` and `struct uk_intctlr_plat_data` for input and output, respectively.

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -183,14 +194,21 @@ int uk_intctlr_irq_free(unsigned int *irqs, __sz count);
...
+ * Platform-specific initialization function for interrupts
+ *
+ * If defined by the platform code then additional platform-specific
+ * initialization steps should be performed.
+ * For example some architectures with mmu support requires address
+ * mapping to access registers.
+ *
+ * @param in pointer to the uk_intctlr_plat_data structure
+ * @param out result structure that should be interpreted by interrupt
+ * 	controller
```
michpappas: checkpatch complains about spaces before tabs in this line, please use only spaces to align doxygen comments.

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -597,6 +597,9 @@ static int gicv2_do_probe(void)
 		return r;
 	}
 
+	ukplat_irq_setup(gicv2_drv.dist_mem_addr, gicv2_drv.cpuif_mem_addr,
```
michpappas: This change is obsoleted by `uk_intctlr_plat_probe()`, so it's not needed to be in the git history.

drivers/ukintctlr/gic/gic-v3.c
```diff
@@ -682,6 +682,9 @@ static int gicv3_do_probe(void)
 		return r;
 	}
 
+	ukplat_irq_setup(gicv3_drv.dist_mem_addr, gicv3_drv.rdist_mem_addr,
```
michpappas: This change is obsoleted by `uk_intctlr_plat_probe()`, so it's not needed to be in the git history.

lib/ukintctlr/include/uk/intctlr.h
michpappas: Thanks a lot. In that case this change can be removed from this commit too.

plat/kvm/Config.uk
michpappas: Thanks, please remove from here no need to introduce a new commit in the series.

plat/Config.uk
```diff
@@ -119,3 +119,9 @@ config FPSIMD
...
+config LIBOFW
```
michpappas: Unless I'm missing something, this can go away too.

drivers/xen/net/Makefile.uk
```diff
@@ -0,0 +1,9 @@
+$(eval $(call addlib_s,libxennetfront,$(CONFIG_XEN_NETFRONT)))
+
+LIBXENNETFRONT_EXPORTS          = $(LIBXENNETFRONT_BASE)/exportsyms.uk
```
michpappas: Is there a reason you export `exportsyms.uk`?

drivers/xen/net/Makefile.uk
```diff
@@ -0,0 +1,9 @@
...
+LIBXENNETFRONT_ASFLAGS-y       += $(LIBXENPLAT_ASFLAGS-y)
+LIBXENNETFRONT_ASINCLUDES-y    += $(LIBXENPLAT_ASINCLUDES-y)
```
michpappas: If you don't have any headers included by assembly files this directive is not necessary.

drivers/xen/net/Makefile.uk
```diff
@@ -0,0 +1,9 @@
+$(eval $(call addlib_s,libxennetfront,$(CONFIG_XEN_NETFRONT)))
+
+LIBXENNETFRONT_EXPORTS          = $(LIBXENNETFRONT_BASE)/exportsyms.uk
+LIBXENNETFRONT_ASFLAGS-y       += $(LIBXENPLAT_ASFLAGS-y)
```
michpappas: Is `ASFLAGS` required by this library?

drivers/xen/blk/exportsyms.uk
michpappas, line 1: IIRC if a library does not export any symbols you can leave `exportsyms.uk` empty.

drivers/xen/net/Config.uk
```diff
@@ -0,0 +1,7 @@
+config XEN_NETFRONT
+	bool "Xenbus Netfront Driver"
+	default y if LIBUKNETDEV
+	default n
```
michpappas: `default n` is redundant, as in Kconfig `bool` symbols default to `n`.

drivers/xen/net/Config.uk
```diff
@@ -0,0 +1,7 @@
...
+	depends on LIBUKNETDEV && XEN_XENBUS
```
michpappas: The dependency model of libraries is to `select` their dependencies, unless that causes a circular depenency, in which case we use `depends on`, and enable from the platform's `Config.uk`.

**Example:**
 The `virito_net` driver does `select LIBVIRTIO_BUS`, `select LIBUKSGLIST`, and `depends on LIBUKNETDEV`. And in `plat/kvm/Config.uk` we have `imply LIBVIRTIO_NET if LIBUKNETDEV`. That way if the user selects `KVM` and `LIBUKNETDEV` (or a library that depends on netdev like lwip) from menuconfig, all dependencies are resolved.

You should do the same for all Xen drivers. In the case of netfront that would be:
```
config LIBXEN_NETFRONT
	bool "Xenbus Netfront Driver"
	select LIBXEN_XENBUS
	help
		Driver for netfront devices
```
and in `plat/xen/Config.uk`:
```
imply LIBXEN_NETFRONT if LIBUKNETDEV
```

drivers/xen/net/Config.uk
```diff
@@ -0,0 +1,7 @@
+config XEN_NETFRONT
```
michpappas: The convention is to prefix with `LIB`, so that would be `LIBXEN_NETFRONT`. Similarly for the other drivers.

drivers/xen/net/Makefile.uk
```diff
@@ -0,0 +1,9 @@
+$(eval $(call addlib_s,libxennetfront,$(CONFIG_XEN_NETFRONT)))
```
michpappas: Also here `CONFIG_LIBXEN_NETFRONT`.

drivers/xen/xenbus/exportsyms.uk
michpappas, line 1: This should only contain symbols made available from this library to other libraries. If there are such symbols (for example the `xs_watch` family), these must be in a header under `include/<driver>/`. For example, assuming that the `xs_watch.h` *only* contains public definitions (if not you should split them into one internal and one external header) then you should have:
```
<unikraft_root>/drivers/xen/xenbus/include/uk/xenbus/xs_watch.h
```
and in the `Makefile.uk` of xenbus, you should made that header public as:
```
CINCLUDES-$(CONFIG_LIBXEN_XENBUS) += $(LIBXEN_XENBUS)/include
```
Drivers that depend on this library should then do `#include <uk/xenbux/xs_watch.h>`.

Finally, if you *really* want to go by the book, the public interfaces should be named as `uk_<lib>_<name>`, eg `uk_xenbux_xs_watch_path()`, but if you think this is excessive we can leave this for a separate PR later (we haven't fixed these yet in other drivers either).

michpappas (changes requested):
@oleksiimoisieiev thanks for the updates. Besides the inline comments, I also spotted a couple of issues in the namespaces of the commit message titles:

* `arm64: Fix MAIR configuration` -> `plat/xen/arm64: ...`
* `xen/pvcalls` -> `plat/xen/pvcalls`
* `/plat/xen/drivers` -> `plat/xen/drivers`
* `xenbus` -> `plat/xen/xenbus`

Also I notice some use a prefix of `plat/xen/arm` while others `plat/xen/arm64`. If there are no changes for 32-bit arm, then `plat/xen/arm64` will do.

Another thing that would be nice to run `support/scripts/checkpatch.uk --strict -g <git_sha>` to all patches.

plat/kvm/Config.uk
oleksiimoisieiev: fixed

arch/arm/arm64/include/uk/asm/lcpu.h
oleksiimoisieiev: fixed

drivers/ukintctlr/gic/gic-v2.c
oleksiimoisieiev: fixed

drivers/ukintctlr/gic/gic-v3.c
oleksiimoisieiev: fixed

drivers/xen/net/Config.uk
oleksiimoisieiev: fixed

drivers/xen/net/Config.uk
oleksiimoisieiev: fixed

drivers/xen/net/Config.uk
oleksiimoisieiev: fixed

drivers/xen/net/Makefile.uk
oleksiimoisieiev: fixed

drivers/xen/net/Makefile.uk
oleksiimoisieiev: removed

drivers/xen/net/Makefile.uk
oleksiimoisieiev: removed

drivers/xen/net/Makefile.uk
oleksiimoisieiev: removed

lib/ukintctlr/include/uk/intctlr.h
oleksiimoisieiev: fixed

lib/ukintctlr/include/uk/intctlr.h
oleksiimoisieiev: added

lib/ukintctlr/include/uk/intctlr.h
oleksiimoisieiev: fixed

lib/ukintctlr/include/uk/intctlr.h
oleksiimoisieiev: switched to void *

plat/Config.uk
oleksiimoisieiev: removed

drivers/xen/xenbus/exportsyms.uk
oleksiimoisieiev: both headers are used by xs.c and they are local as far as I can see.

oleksiimoisieiev: moved exported headers from plat/xen/include

drivers/ukintctlr/gic/gic-v2.c
michpappas: Looks like you missed removing this in the last update :)

drivers/ukintctlr/gic/gic-v3.c
michpappas: Looks like you missed removing this in the last update :)

lib/ukintctlr/include/uk/intctlr.h
michpappas: Looks like you missed removing this in the last update :)

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
```diff
@@ -134,6 +134,17 @@ struct _gic_dev {
 	struct _gic_operations ops;
 };
 
+/** Platform specific initialization data */
+struct uk_intctlr_plat_data {
+#if defined(__aarch64__)
```
michpappas: Somehow I missed this earlier, these should be `CONFIG_ARCH_ARM_64` and `CONFIG_ARCH_ARM`.

drivers/xen/9p/Config.uk
```diff
@@ -0,0 +1,22 @@
+menuconfig LIBXEN_9PFRONT
+	bool "Xenbus 9pfront Driver"
+	select LIBXEN_XENBUS
```
michpappas, line 3: Also add `depends on PLAT_XEN` here to avoid showing the driver in menuconfig if Xen is not enabled.

drivers/xen/blk/Config.uk
```diff
@@ -0,0 +1,23 @@
+menuconfig LIBXEN_BLKFRONT
+	bool "Xenbus Blkfront Driver"
+	select LIBXEN_XENBUS
```
michpappas, line 3: Also add `depends on PLAT_XEN` here to avoid showing the driver in menuconfig if Xen is not enabled.

drivers/xen/net/Config.uk
```diff
@@ -0,0 +1,5 @@
+config LIBXEN_NETFRONT
+	bool "Xenbus Netfront Driver"
+	select LIBXEN_XENBUS
```
michpappas, line 3: Also add `depends on PLAT_XEN` here to avoid showing the driver in menuconfig if Xen is not enabled.

drivers/xen/xenbus/Config.uk
```diff
@@ -0,0 +1,10 @@
+config LIBXEN_XENBUS
+	bool "Xenbus/XenStore driver"
+	default y if XEN_PV
+	depends on XEN_GNTTAB && PLAT_XEN
```
michpappas: Here you only need to keep `depends on PLAT_XEN`. And in `plat/xen/Config.uk` do:
```diff
diff --git a/plat/xen/Config.uk b/plat/xen/Config.uk
index ce869545..1f849edd 100644
--- a/plat/xen/Config.uk
+++ b/plat/xen/Config.uk
@@ -74,7 +74,7 @@ config XEN_PV_BUILD_P2M
 config XEN_GNTTAB
        bool "Grant table support"
        default y if XEN_PV
-       default n
+       select LIBXEN_XENBUS
        depends on (ARCH_X86_64 || ARCH_ARM_64)
        select LIBUKALLOC
        select LIBUKLOCK
```

michpappas (changes requested):
@oleksiimoisieiev last one! Just a few minor updates and it's good from my side. Besides the inline comments, also please break down the commit "lib/ukintctlr: Introducing platform-specific init func to ukintctlr" into one commit per library:
* lib/ukintctlr: Introducing platform-specific init func to ukintctlr
* drivers/ukintctlr/gic: Add platform init support
* plat/xen: <title of your choice>

That is because we normally require separate commits per library, as it makes it easier to track each library's history and facilitates review as often libraries have different maintainers.

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
oleksiimoisieiev: fixed

drivers/xen/9p/Config.uk
oleksiimoisieiev: fixed

drivers/xen/blk/Config.uk
oleksiimoisieiev: fixed

drivers/xen/net/Config.uk
oleksiimoisieiev: fixed

drivers/xen/xenbus/Config.uk
oleksiimoisieiev: fixed

michpappas (approved):
@oleksiimoisieiev many thanks for going through all the rework 🚀 

andreistan26 (approved):
@oleksiimoisieiev, thanks for the PR, works great.

Reviewed-by: Andrei Stan <andreistan2003@gmail.com>

michpappas:
oleksiimoisieiev many thanks.

craciunoiuc (approved):
Don't mind me, just leaving the tag again.

PR #734, opened 2023-01-30
`lib/ukstreambuf`: (C-string) buffer helper

skuenzer:
### Base target

 - Architecture(s): all
 - Platform(s): all

### Description of changes

This PR introduces `lib/ukstreambuf` a library that makes it easier to enqueue data to a buffer. It automatically keeps track of the current seek position and left space on a buffer. Especially for C-string buffers there is also special handling available for taking care of correct `'\0'`-termination so that C-strings can be easily appended.


marcrittinghaus:
Hey @skuenzer 😃

just saw this PR popping up and noticed some minor things. I saw that you used a different style for the function documentation than we currently have in the [coding conventions](https://docs.google.com/document/d/1_iRkdBjQtnqOj798ZNr23OQTWQcXPdpDsnfBikm6Pcg/edit?disco=AAAAmQ30v1Y). Also for the header, we said we use the [shorter](https://docs.google.com/document/d/1_iRkdBjQtnqOj798ZNr23OQTWQcXPdpDsnfBikm6Pcg/edit?disco=AAAAg_W_G0U) one.

Since it took me a moment to understand the purpose of the lib and since we are currently trying to do this for all libs, would it be possible to add a `README.md` with a short description and usage sample?

I know that you are under high pressure to get things ready, so never mind if this is not possible to change in time.

skuenzer:
> Hey @skuenzer 😃
> 
> just saw this PR popping up and noticed some minor things. I saw that you used a different style for the function documentation than we currently have in the [coding conventions](https://docs.google.com/document/d/1_iRkdBjQtnqOj798ZNr23OQTWQcXPdpDsnfBikm6Pcg/edit?disco=AAAAmQ30v1Y). Also for the header, we said we use the [shorter](https://docs.google.com/document/d/1_iRkdBjQtnqOj798ZNr23OQTWQcXPdpDsnfBikm6Pcg/edit?disco=AAAAg_W_G0U) one.
> 
> Since it took me a moment to understand the purpose of the lib and since we are currently trying to do this for all libs, would it be possible to add a `README.md` with a short description and usage sample?
> 
> I know that you are under high pressure to get things ready, so never mind if this is not possible to change in time.

Hey @marcrittinghaus,

sorry, you are right, a left open TODO, also `exportsyms.uk` was missing. I updated the headers. The `README.md` I would delay this time so that our binary compat release is not getting in danger. I'll provide it as part of the docs project. The library will be used by the strace and netlink implementations.

Simon

lib/ukstreambuf/include/uk/streambuf.h
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * Current length of data on the buffer of a streambuf
+ * HINT: If `UK_STREAMBUF_C_TERMSHIFT` is _not_ configured, the length is equal
+ *       to the seek offset.
+ *
+ * @param sb Streambuf object
+ * @return Current length on the corresponding buffer
+ */
+#define uk_streambuf_len(sb)						\
+	((sb)->seek)
```
eduardvintila, line 153: Why do we consider the termination byte to be part of the length of the data when `UK_STREAMBUF_C_TERMSHIFT` is configured?  
The `uk_streambuf_left` macro doesn't do this, for example. It reports the terminating byte as a free byte which can be overwritten. 

My intuition was that it should be similar to `strlen` when that option is selected and that `uk_streambuf_len(sb)` should just be equal to `uk_streambuf_seek(sb)`. Of course, tests would need to be updated accordingly. It's possible I'm missing something out and this isn't really an issue, though.

lib/ukstreambuf/include/uk/streambuf.h
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * Left bytes on the buffer of a streambuf object
+ * NOTE: If `UK_STREAMBUF_C_TERMSHIFT` is configured, the reported number of
+ *       bytes includes the terminating byte of a previous append operation
+ *       (e.g., `uk_streambuf_printf()`). As a result, a buffer configured
+ *       with `UK_STREAMBUF_C_TERMSHIFT` will still report 1 byte left for
+ *       overwriting the termination byte.
+ *
+ * @param sb Streambuf object
+ * @return Current length on the corresponding buffer
```
eduardvintila:
```suggestion
 * @return Current number of bytes left on the corresponding buffer
```

lib/ukstreambuf/include/uk/streambuf.h
```diff
@@ -0,0 +1,277 @@
...
+	((sb)->buflen - (__sz) uk_streambuf_seek((sb)))
+
+/**
+ * Report if content on buffer got truncated
+ * This happens when an appending operation (e.g., `uk_streambuf_printf()`)
+ * cannot write all bytes to the corresponding buffer because of limited buffer
+ * space.
+ *
+ * @param sb Streambuf object
+ * @return Current length on the corresponding buffer
```
eduardvintila:
```suggestion
 * @return Non-zero if the content got truncated, zero otherwise.
```

eduardvintila (commented):
Hey, @skuenzer. This library looks really good and concise! I have left a tiny review, please take a look when you can. Thanks.

lib/ukstreambuf/include/uk/streambuf.h
skuenzer: Hi @eduardvintila, this is a good question. My idea behind was that a stream buffer is also working on binary data without any notion of termination symbols. I did consider `uk_streambuf_len(sb)` as the size of actual used data on the buffer - irrespective if there are termination symbols. If `UK_STREAMBUF_C_TERMSHIFT` is set, `uk_streambuf_seek(sb)` returns the number that you are asking for. In principle I consider that `uk_streambuf_len(sb)` has a relation to `uk_streambuf_buflen(sb)` while `uk_streambuf_seek(sb)` has a relation to `uk_streambuf_left(sb)`. I think, to avoid confusion, I would add additional notes to the function descriptions so that this is more clear. What do you think?

lib/ukstreambuf/include/uk/streambuf.h
skuenzer: Right, thanks!

lib/ukstreambuf/include/uk/streambuf.h
skuenzer: Also true, sorry, a copy&paste mistake.

lib/ukstreambuf/include/uk/streambuf.h
eduardvintila: @skuenzer Yep, sure, makes sense to me and a few notes on this would be really good to have  👍🏻

skuenzer:
@eduardvintila I pushed an update

eduardvintila (approved):
This looks great, thanks @skuenzer!

skuenzer:
@eduardvintila I just pushed another update: I rebased the PR to latest staging and moved the include of `<uk/essentials.h>` to `<uk/streambuf.h>`.

razvand (approved):

nderjung:
This was implicitly merged in #737.

PR #744, opened 2023-02-03
lib/vfscore: Cast initrd vbase to (void *)

skuenzer:
PR #722 introduced an update to `struct ukplat_memregion_desc` where the virtual base of a memory region is now specified as a __vaddr_t integer value. Since it is a virtual address we can cast it to (void *) to avoid build warnings in `lib/vfscore/rootfs.c`.


marcrittinghaus (approved):
Hey @skuenzer,

thanks very much! Missed the build warnings within the other ones. I tested it and it works fine. Warnings are gone 😃 

PR #747, opened 2023-02-04
lib/uksched: Align allocated stacks to arch requirement

skuenzer:
This PR introduces aligns stack allocation that are done by `lib/uksched` to the requirement of the target architecture. Without aligned stacks, random crashed may occur because the compiler is assuming such an alignment. An incorrect stack alignment can cause crashes because the compiler may placed instructions in the code that will fail when accessing values on the stack if not properly aligned.

### DIFF
--- a/lib/uksched/thread.c
+++ b/lib/uksched/thread.c
@@ -368,7 +368,7 @@ static int _uk_thread_struct_init_alloc(struct uk_thread *t,
 	int rc;
 
 	if (a_stack && stack_len) {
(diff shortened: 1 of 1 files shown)


razvand:
Hi, @skuenzer, @mschlumpp, this is OK. I tested it, all is good.

@mschlumpp, please approve it and add the reviewed tag, and then I'll do the final approval.

jschlumpp (approved):
One stack alignment bug less ✨ 

razvand (approved):

PR #758, opened 2023-02-09
lib/posix-futex: Implement subset of FUTEX_WAIT_BITSET

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBPOSIX_FUTEX=y`

### Description of changes

Many modern applications use `FUTEX_WAIT_BITSET` to be able to set an absolute timeout. To accomplish that they set all bits in the `val3`/mask parameter, which makes the operation equivalent to a normal `FUTEX_WAIT` (except the absolute timeout value).

### DIFF
--- a/lib/posix-futex/futex.c
+++ b/lib/posix-futex/futex.c
@@ -59,6 +59,7 @@
 #include <uk/print.h>
 #include <uk/spinlock.h>
 #include <uk/plat/lcpu.h>
+#include <uk/plat/time.h>
 
 /** @struct uk_futex
  *  @brief Futex structure.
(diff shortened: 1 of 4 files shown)


lib/posix-futex/futex.c
```diff
@@ -276,10 +276,43 @@ UK_LLSYSCALL_R_DEFINE(int, futex, uint32_t *, uaddr, int, futex_op,
...
 		      uint32_t *, uaddr2, uint32_t, val3)
 {
+	__nsec timeout_ns;
+
+	/*
+	 * N.B. CLOCK_(MONOTONIC|REALTIME|...) are at the moment all the same in
+	 * Unikraft. Therefore, we can just use CLOCK_MONOTONIC for timeouts in
+	 * the following code.
+	 */
 	switch (futex_op) {
```
csvancea: The command stored inside `futex_op` can be OR'ed with `FUTEX_CLOCK_REALTIME`. In this case, the execution goes to the default case and the syscall returns `-ENOSYS`.

Considering that both monotonic and realtime clocks are the same in Unikraft, changing `switch (futex_op)` to `switch (futex_op & FUTEX_CMD_MASK)` fixes the issue for me (and the `*_PRIVATE` cases can be dropped too!)

csvancea (commented):
Hey @mschlumpp,

Thanks for looking into this bug!

Certain pthreads functions (pthread_join) may generate futex syscalls with `futex_op=FUTEX_WAIT_BITSET_PRIVATE | FUTEX_CLOCK_REALTIME` and `timeout=NULL`, which I suppose it means that the thread should wait indefinitely until it is waken up by a `FUTEX_WAKE` command.

However, Unikraft (+app_elfloader) appears to crash right after the `FUTEX_WAIT_BITSET` command, probably when doing the context switch. I am not sure whether this bug is related to this PR or not, so I will open a different issue and provide some repro steps.

LE: Issue #796

jschlumpp:
@csvancea That error is likely related to the currently broken `clone3` system call. Our current implementation seems to clear all registers but glibc passes the function to be executed in the [`rdx` register](https://sourcegraph.com/github.com/bminor/glibc@e2756903329365134089d23548e9083d23bc3dd9/-/blob/sysdeps/unix/sysv/linux/x86_64/clone3.S?L33). But after the clearing it will just jump to `0x0`.

csvancea (approved):
Retested the changes with app-elfloader + pthread test app and confirmed that the futex calls made by the Linux ELF succeed.

Thank you, @mschlumpp 

Reviewed-by: Cosmin Vancea csvancea@gmail.com

adinasm (approved):
Tested it too with app-elfloader (and changed clone3) and didn't notice any problems with the futex calls.

Reviewed-by: Adina Smeu <adina.smeu@gmail.com>

razvand (approved):

PR #752, opened 2023-02-09
plat/x86: Enable NX bit when initializing paging

jschlumpp:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_PAGING=y`

### Description of changes

There are a few reasons to enable the NX bit in the paging init routine:
* Setting the bit before checking for actual support is risky. Currently, the check happens in the paging initialization routine.
* The bit is not used, when the Unikraft paging support is not enabled.
* Every boot entry point (32-bit/64-bit) would have a duplicated enable.

PR #751, opened 2023-02-09
lib/ukvmem: Compile all source files for ISR context

jschlumpp:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBUKVMEM=y`

### Description of changes

The main and individual pagefault handlers are executed in an ISR context. Not marking the files as `isr`, will cause the usual trouble (clobbered vector registers).

PR #760, opened 2023-02-13
plat/kvm: Firecracker boot protocol support

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This introduces support booting Unikraft on the firecracker VMM. 

This depends on the following PRs:
 * #752 
 * #751 

The commits of these PRs are included and will be removed once these PRs are merged. Update: done :heavy_check_mark:


plat/kvm/x86/setup.c
```diff
@@ -273,15 +273,23 @@ static __sz cmdline_len;
...
-	char *cmdl = (bi->cmdline) ? (char *)bi->cmdline : CONFIG_UK_NAME;
+	char *cmdl;
 
-	cmdline_len = strlen(cmdl) + 1;
+	if (bi->cmdline) {
+		cmdl = (char *)bi->cmdline;
+		cmdline_len = bi->cmdline_len;
+	} else {
+		cmdl = CONFIG_UK_NAME;
+		cmdline_len = strlen(cmdl);
```
mogasergiu: How about `sizeof(CONFIG_UK_NAME)` instead of `strlen`.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,75 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+			/* Some part of the inserted region is before the
+			 * overlapping region. Try to insert that part if it's
+			 * large enough.
+			 */
+			mrdc.pbase = pstart;
+			mrdc.vbase = pstart + voffset;
+			mrdc.len   = mrdp->pbase - pstart;
+
+			if (min_size && mrdc.len >= min_size)
+				ukplat_memregion_list_insert(list, &mrdc);
```
mogasergiu: Shouldn't here be a check for the return value as well?

plat/kvm/x86/multiboot.c
```diff
@@ -133,33 +124,21 @@ void multiboot_entry(struct lcpu *lcpu, struct multiboot_info *mi)
...
-
-					start = mrdp->pbase + mrdp->len;
-				}
-
-				if (end - start < PAGE_SIZE)
-					continue;
+				rc = ukplat_memregion_list_insert_split_phys(
+				    &bi->mrds, &mrd, __PAGE_SIZE);
+				if (rc < 0)
+					multiboot_crash("unable to add region",
```
mogasergiu: Just a nitpick, but perhaps it would look nicer if all crash messages would begin with a capital letter, instead of having messages that begin with capital letters as well as messages that begin with lowercase letters. I personally find it inconsistent.

plat/kvm/x86/lxboot.S
```diff
@@ -0,0 +1,50 @@
...
+ * enabled with ident mapping for loaded kernel, zero page (i.e., struct
+ * lxboot_params) and command line buffer. Interrupts disabled.
+ */
+.code64
+.section .text.64.boot
+ENTRY(_lxboot_entry)
+	cmpl	$LXBOOT_HDR_HEADER_MAGIC, LXBOOT_HDR_HEADER_OFFSET(%rsi)
+	jne	no_lxboot
+
+	movq	$lcpu_boot_startup_args, %rdi	/* startup args for boot CPU */
```
mogasergiu: I believe you could re-write this to be position independent, like `leaq lcpu_boot_startup_args(%rip), %rdi`. Same for the instruction below that moves `x86_bpt_pml4`.

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,160 @@
...
+	mrd.vbase = initrd_addr;
+	mrd.pbase = initrd_addr;
+	mrd.len = initrd_size;
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "initrd", sizeof("initrd"));
+#endif
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+	if (rc < 0)
+		lxboot_crash(rc, "unable to add ram mapping");
```
mogasergiu: Same comment regarding uppercase/lowercase letters at the beginning of messages. Even though, from what I can tell by the definition of the macro, the message will never get printed.

plat/kvm/Config.uk
```diff
@@ -12,6 +12,29 @@ menuconfig PLAT_KVM
...
+	prompt "Hypervisor"
+	default KVM_HYPERVISOR_QEMU
+
+config KVM_HYPERVISOR_QEMU
+	bool "QEMU"
+	select KVM_BOOT_PROTO_MULTIBOOT
+
+config KVM_HYPERVISOR_FIRECRACKER
+	bool "Firecracker"
+	select KVM_BOOT_PROTO_LXBOOT
```
nderjung: As part of this addition, would differentiating the output artifact name now be appropriate? E.g.:
- `helloworld_qemu-x86_64`
- `helloworld_fc-x86_64`

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,75 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+	pend = mrd->pbase + mrd->len;
+
+	memcpy(&mrdc, mrd, sizeof(mrdc));
+
+	for (i = 0; i < (int)list->count; i++) {
+		ukplat_memregion_get(i, &mrdp);
+		if (!ukplat_memregion_desc_overlap(mrdp, pstart, pend))
+			continue;
+
+		if (!mrdp->type)
```
mogasergiu, line 163: By the way you seem to be checking here for a non-zero mrd type. But what if the region you are about to split is of a type that would require a higher priority than the one you are comparing it to? For example, what if you are adding a mrd of type RESERVED whose pstart < mrdp->pbase and whose pend > mrdp->pbase + mrdp->len. Then you would have to remove mrdp entirely from the list, since the reserved mrd you are about to add takes priority.

jschlumpp: A priority system was not really intended for this function and I currently don't see a need to implement one for the `RESERVED` case (the other types don't really have a clear priority ordering anyway).

plat/kvm/Config.uk
jschlumpp: This PR places the firecracker support in a strange state where it's not really a full different platform like linuxu or Xen, which can be compiled with the same `.config` file, so I'm not sure whether doing the same differentiation as those cases is a good approach. While the user could quickly distinguish between a image for QEMU and FC, they would also risk having stale images around after switching the configuration option. By keeping the image name the incorrect image would just not boot.
In the long-term the plan is moving to a proper linux boot protocol implementation anyway and that can be loaded by QEMU, too. (Most KVM hypervisors do support that protocol for obvious reasons).

So I would prefer leave the current file name as it is.

nderjung: I see the conundrum, thanks. Whilst the long-term would be to support a more "universal" boot protocol, couldn't the short-term solution to the image being stale on a config switch just be to create a relevant Make-target for the FC/MB config options (one that could enable GC of the image via `make clean`)?

The reason for the suggestion is the attempt to prevent the user from ever being in the position of booting the incorrect image because "fc" and "qemu" are now in the name.  (That said, it would probably be ok if some error occurs during boot that provides enough information for the user to realise the mistake).  My fear is that with the boot protocol it might be too early to output an error (though correct me if I am wrong) and a (new) user is stuck with an image they don't know what to do with.

include/uk/plat/memory.h
```diff
@@ -84,6 +84,27 @@ struct ukplat_memregion_desc {
...
+ *   the end of the physical memory region
+ * @return
+ *   zero if the two specified regions have no overlap, a non-zero value
+ *   otherwise
+ */
+static inline int
+ukplat_memregion_desc_overlap(const struct ukplat_memregion_desc *mrd,
+			      __paddr_t pstart, __paddr_t pend)
+{
+	return ((pend > mrd->pbase) && (pstart < mrd->pbase + mrd->len));
```
mogasergiu: Perhaps worth mentioning, but I have just noticed there already exists a macro for this:

https://github.com/unikraft/unikraft/blob/staging/include/uk/essentials.h#L264

Would it be worth considering using the macro instead? Or perhaps there are too many additional checks...

plat/kvm/Config.uk
```diff
@@ -12,6 +12,29 @@ menuconfig PLAT_KVM
...
+if (ARCH_X86_64)
+
+config KVM_BOOT_PROTO_MULTIBOOT
+	bool
+
+config KVM_BOOT_PROTO_LXBOOT
+	bool
+
+choice
+	prompt "Hypervisor"
```
Krechals: nit
```suggestion
	prompt "Virtual Machine Monitor"
```

plat/kvm/x86/setup.c
```diff
@@ -273,15 +276,23 @@ static __sz cmdline_len;
...
+		cmdl = CONFIG_UK_NAME;
+		cmdline_len = sizeof(CONFIG_UK_NAME) - 1;
+	}
 
-	cmdline = bootmemory_palloc(cmdline_len, UKPLAT_MEMRT_CMDLINE);
+	cmdline = bootmemory_palloc(cmdline_len + 1, UKPLAT_MEMRT_CMDLINE);
 	if (unlikely(!cmdline))
 		return -ENOMEM;
 
 	strncpy(cmdline, cmdl, cmdline_len);
```
Krechals: Can we use `memcpy` instead? Since we are already using `cmdline_len`, I don't think stopping at a potential \0 is even desired.

plat/common/include/uk/plat/common/bootinfo.h
```diff
@@ -28,17 +28,20 @@ struct ukplat_bootinfo {
 	/** Null-terminated boot protocol identifier */
 	char bootprotocol[16];
 
-	/** Address of null-terminated kernel command line */
+	/** Address of the kernel command line */
 	__u64 cmdline;
```
Krechals: Let's specify this is not null-terminated.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -89,6 +90,10 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
 	struct ukplat_memregion_desc *p;
 	__u32 i;
 
+	/* Ignore zero-length regions to simplify platform-specific code */
+	if (unlikely(mrd->len == 0))
+		return 0;
```
Krechals: Do we treat this case as an error?
Returning 0 should mean that the element was placed at the beginning of the list.
I would say we should return a status code instead.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,80 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+ *   The minimum size a inserted region has to have. Regions smaller than this
+ *   size won't be added. Settings this parameter zero will disable this
+ *   behavior.
+ * @returns
+ *   zero if the operation was successful, an negative errno otherwise
+ */
+static inline int
+ukplat_memregion_list_insert_split_phys(struct ukplat_memregion_list *list,
+					const struct ukplat_memregion_desc *mrd,
+					__sz min_size)
```
Krechals: nit
```suggestion
static inline int
ukplat_memregion_list_insert_split_phys(struct ukplat_memregion_list *list,
					const struct ukplat_memregion_desc *mrd,
					const __sz min_size)
```

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,80 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+	int rc;
+
+	voffset = mrd->vbase - mrd->pbase;
+	pstart = mrd->pbase;
+	pend = mrd->pbase + mrd->len;
+
+	memcpy(&mrdc, mrd, sizeof(mrdc));
+
+	for (i = 0; i < (int)list->count; i++) {
+		ukplat_memregion_get(i, &mrdp);
```
Krechals: Should we check the return value?

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,80 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+		if (!ukplat_memregion_desc_overlap(mrdp, pstart, pend))
+			continue;
```
Krechals, line 159: Since the list is sorted, I believe we can break the iteration just when we find the first element that has: 
`mrdp->pbase > pend`

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +117,80 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+			 * large enough.
+			 */
+			mrdc.pbase = pstart;
+			mrdc.vbase = pstart + voffset;
+			mrdc.len   = mrdp->pbase - pstart;
+
+			if (min_size && mrdc.len >= min_size) {
+				rc = ukplat_memregion_list_insert(list, &mrdc);
+				if (rc < 0)
+					return -ENOMEM;
```
Krechals: Currently `ukplat_memregion_list_insert` traverses the list (which is actually an array) from the tail to the head to find the right index at which the insertion should happen.

The thing is that we already know the position at which we want to insert the element: `i`. 
Not sure if we really want to do it in this PR, as it's already getting large. Still, adding a TODO as a comment would be nice.

Krechals: On a different note, currently we are traversing 2 times the same element whenever an insertion happens:

Let's say i = 0 and we want to insert a new element. The element will be inserted as the first element, and the current `mrdp` will become the second in the list. In the next iteration, we will traverse the same `mrdp`.

I think here we should increment i by 1.

Krechals (commented):
Thanks for doing this, Marco! I think it generally looks good and I don't see anything major that needs to be done. There are only a few places I didn't check so far (makefile, linker), but I'll take a look tomorrow.

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,160 @@
...
+	mrd.pbase = cmdline_addr;
+	mrd.vbase = cmdline_addr;
+	mrd.len = cmdline_size;
+	mrd.type = UKPLAT_MEMRT_CMDLINE;
+	mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_MAP;
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "cmdline", sizeof("cmdline"));
+#endif
+
+	rc = ukplat_memregion_list_insert_nonempty(&bi->mrds, &mrd);
```
mogasergiu: Why use the `nonempty` variation here? Are you not already checking for the size above? Same for initrd.

plat/kvm/x86/setup.c
```diff
@@ -194,7 +194,10 @@ static int paging_init(void)
 
 		rc = ukplat_page_map(&kernel_pt, vaddr, paddr,
 				     len >> PAGE_SHIFT, prot, 0);
-		if (unlikely(rc))
+		/* Unmappings are currently not performed on the low-mem.
```
mogasergiu, line 197: IIRC this is why the multiboot.c code drops memory regions below 1MiB. You can do so as well.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,113 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+/**
+ * Inserts a region into the memory region list. See
+ * ukplat_memregion_list_insert() for a detailed description. Compared to the
+ * that function, this one ignores empty regions.
+ * @param list The memory region list to insert the range into
+ * @param mrd The memory range to insert.
+ * @return Zero if the region was added successfully, a negative errno otherwise
+ */
+static inline int
+ukplat_memregion_list_insert_nonempty(struct ukplat_memregion_list *list,
```
mogasergiu: I don't see the need for an additional function just for this. You are unlikely to use this outside of the boot code, so why not just do an `if` statement there outside the insertion function? Maybe I am missing other use cases?

plat/common/x86/cpu_native.c
```diff
@@ -48,13 +48,19 @@ unsigned long read_cr2(void)
...
+#endif
+
 	/*
 	 * Perform an ACPI shutdown by writing (SLP_TYPa | SLP_EN) to PM1a_CNT.
 	 * Generally speaking, we'd have to jump through a lot of hoops to
 	 * collect those values, however, for QEMU, those are static. Should be
 	 * harmless if we're not running on QEMU, especially considering we're
 	 * already shutting down, so who cares if we crash.
 	 */
+
```
mogasergiu: Is this newline left here by mistake?

mogasergiu (changes requested):
Thanks for applying the changes! See my other comments

plat/kvm/x86/setup.c
jschlumpp: Firecracker places the command line below 1MiB.

plat/common/include/uk/plat/common/memory.h
jschlumpp: All the functions in the header are only used in the boot code ^^
I assume you meant that the behavior doesn't make any difference in the current state of the PR. Indeed, with the last update to the PR the primary use-case was refactored away (I would say it's safe to assume that the firmware doesn't give us zero-length regions).

mogasergiu: Hmm, I had in mind this [commit](https://github.com/unikraft/unikraft/pull/848/commits/9258f201b65d074be2667bca263080d68397c5d9), with this you could dynamically allocate from free memory regions regardless of where you are in the code. But yes, I agree, no point in over-engineering this stuff over assumptions of a faulty previous boot phase :). If anything, there should be only one call, of a function called at the end of the boot phase that does all this checking and removes empty regions or coalesces the regions. I thought of something like [this](https://github.com/unikraft/unikraft/pull/848/commits/f53ba355a9f2dc2f36ffd8119b82a67fa4394591)? But,  for now, it's outside of the scope of this PR :).

plat/kvm/x86/setup.c
mogasergiu: That's unfortunate...

plat/common/x86/cpu_native.c
```diff
@@ -48,6 +48,11 @@ unsigned long read_cr2(void)
 
 void system_off(void)
 {
+#if CONFIG_KVM_VMM_FIRECRACKER
+	/* Trigger a reset via the PS/2 controller */
+	outb(0x64, 0xFE);
```
mogasergiu, line 55: I don't think this is Firecracker specific. I guess this is just one of the legacy ways of pulsing the CPU's reset pin thruogh the PS/2 Controller? Pretty sure this should work on QEMU as well. If anything,  I would personally just place it at the end of the function without any ifdef's, as a last resort, following the `isa-debug-exit`.

jschlumpp: The problem is that it will cause a reset instead of a shutdown on non-firecracker systems. While QEMU has a command-line configuration for that (`-no-reboot`), it's not the default. 
And poking random I/O registers seems like a bad idea. So maybe it makes more sense to put the QEMU ACPI/ISA stuff in an `#ifdef`, too.

mogasergiu: > The problem is that it will cause a reset instead of a shutdown on non-firecracker systems. While QEMU has a command-line configuration for that (`-no-reboot`), it's not the default. And poking random I/O registers seems like a bad idea. So maybe it makes more sense to put the QEMU ACPI/ISA stuff in an `#ifdef`, too.

Or, perhaps, it should have been placed  in a `reset()` function like [ARM](https://github.com/unikraft/unikraft/blob/staging/plat/common/arm/cpu_native.c#L51) has?

jschlumpp: On firecracker it is a power off and not a reset trigger. So putting it in a reset function would be wrong.

mogasergiu: This inconsistency is unfortunate :/. Yeah I guess an ifdef should do.

plat/kvm/x86/setup.c
```diff
@@ -273,15 +273,23 @@ static __sz cmdline_len;
 
 static inline int cmdline_init(struct ukplat_bootinfo *bi)
 {
-	char *cmdl = (bi->cmdline) ? (char *)bi->cmdline : CONFIG_UK_NAME;
+	char *cmdl;
 
-	cmdline_len = strlen(cmdl) + 1;
+	if (bi->cmdline) {
```
michpappas: Perhaps it's better to use `bi->cmdline_len` here, as there is a chance that in the case of no args, `bi->cmdline` will point to a `NUL` byte.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,96 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+	__paddr_t pstart, pend;
+	__vaddr_t voffset;
+	int i;
+	int rc;
+
+	voffset = mrd->vbase - mrd->pbase;
+	pstart = mrd->pbase;
+	pend = mrd->pbase + mrd->len;
+
+	memcpy(&mrdc, mrd, sizeof(mrdc));
```
Krechals: Can we use assignment instead of memcpy? - Easier to read (IMO) and it might let the compiler do some  optimisations 
```suggestion
	mrdc = *mrd;
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,160 @@
...
+		lxboot_crash(rc, "Unable to add ram mapping");
+}
+
+static void
+lxboot_init_mem(struct ukplat_bootinfo *bi, struct lxboot_params *bp)
+{
+	struct ukplat_memregion_desc mrd = {0};
+	struct lxboot_e820_entry *entry;
+	__u64 start, end;
+	int i, rc;
```
Krechals: nit; Though I expected we had the flags enabled for spotting such things. 
```suggestion
	int rc;
	__u8 i;
```

Krechals (approved):
Awesome work! LGTM. Only left a few comments. Thank you!

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,96 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+ *   The memory region list to insert the range into
+ * @param mrd
+ *   The memory range to insert
+ * @param min_size
+ *   The minimum size a inserted region has to have. Regions smaller than this
+ *   size won't be added. Settings this parameter zero will disable this
+ *   behavior.
+ * @returns
+ *   zero if the operation was successful, an negative errno otherwise
+ */
```
mogasergiu: Just some typo's I believe.

```suggestion
 * Insert a new region into the memory region list. This extends
 * ukplat_memregion_list_insert to carve out the area of any pre-existing
 * overlapping regions. The virtual addresses will leave out any carved out
 * region.
 * If there is not enough space for all the resulting regions, the function will
 * insert as many as possible and then return an error.
 * @param list
 *   The memory region list to insert the range into
 * @param mrd
 *   The memory range to insert
 * @param min_size
 *   The minimum size a inserted region has to have. Regions smaller than this
 *   size won't be added. Settings this parameter zero will disable this
 *   behavior.
 * @returns
 *   Zero if the operation was successful, a negative errno otherwise
 */
```

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,96 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+			mrdc.pbase = pstart;
+			mrdc.vbase = pstart + voffset;
+			mrdc.len   = mrdp->pbase - pstart;
+
+			if (mrdc.len > min_size) {
+				rc = ukplat_memregion_list_insert(list, &mrdc);
+				if (rc < 0)
+					return rc;
+
+				UK_ASSERT(i == rc);
```
mogasergiu: I think this could prove to be a bad idea to use `UK_ASSERT` in such functions, as they are most likely used before console initialization, right?

mogasergiu (approved):
Other than these 2 comments and the capital letter at the beginning of that one `multiboot_crash`, everything looks good to me :). Very clean PR!

plat/kvm/x86/lxboot.c
jschlumpp: I though that the compiler was smart enough to realize that an `int` will work fine. But for testing I used `__s8` and the compiler also didn't complain... (However, `clang-tidy` does recognize that the upper-bound would be problematic).

nderjung:
❗ This PR contains a major breaking change regarding output binary name. As part of this addition, differentiating the output artifact now results in, for example:
- `helloworld_qemu-x86_64`
- `helloworld_fc-x86_64`

Making this comment visible for the upcoming release note.

jschlumpp:
Small fix: I [changed the split function](https://github.com/unikraft/unikraft/compare/fb38f7f54dcdd161c47e3e574f2a88fe530e6c68..ed8774d69dcd9b334625ec6267e06a2be459e97a) to not use the global bootinfo table.

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+static void
+lxboot_init_mem(struct ukplat_bootinfo *bi, struct lxboot_params *bp)
+{
+	struct ukplat_memregion_desc mrd = {0};
+	struct lxboot_e820_entry *entry;
+	__u64 start, end;
+	int rc;
+	__s8 i;
+
+	for (i = 0; i < bp->e820_entries; i++) {
```
mogasergiu, line 101: I have just stumbled upon this while working on something else but, IIRC, witth paging disabled, the allocator expects that the free memory regions that it owns are all below 4GiB since that is what the static boot page tables map. Just making sure, but are all of these entries below 4GiB or should this loop here stop when it reaches the first memory region above 4GiB if paging is disabled (assuming they are sorted)?

I think that the multiboot code does not have to worry about that. Although the address field of the entry is 64 bits, the `mem_upper` field is always 32 bits so I guess that allows it to safely assue that no memory region is above 4GiB.

jschlumpp: That's a good point, but I don't think the boot entry code is the proper place to fix this problem. 
Also, if the point about multiboot is true, wouldn't that also mean that it currently doesn't support more than 4GiB of memory (even with paging)?

mogasergiu: Multiboot1 does not but the e820 entry does. That's why I said its address field is 64-bit.

mogasergiu: Oh and I agree the boot entry point is not the place to fix this, you are right. The boot entry point should do everything it can to fill the `bootinfo` struct the best it can. Therefore, perhaps it should be the job of what comes after it to accomodate what it is given for what it can offer. That being said, should this check be done in `heap_init` in `libukboot` instead?

jschlumpp: I just talked to Marc and he pointed me to [this](https://github.com/unikraft/unikraft/blob/195cc32f7e5e65d8929ae65b8fa9203ddf246ba6/plat/kvm/x86/setup.c#L233). So the boot code already prunes the high memory areas.

mogasergiu: Ah I see. I must have missed it.
 I was just making sure. Thanks for checking :).

plat/kvm/Linker.uk
```diff
@@ -11,7 +15,13 @@ KVM_BINFO_FLAGS := -a $(CONFIG_UK_ARCH)
 ##
 ## Link image
 ##
-KVM_IMAGE := $(BUILD_DIR)/$(CONFIG_UK_NAME)_kvm-$(CONFIG_UK_ARCH)
+ifeq ($(CONFIG_KVM_VMM_QEMU),y)
+KVM_VMM := qemu
+else ifeq ($(CONFIG_KVM_VMM_FIRECRACKER),y)
+KVM_VMM := fc
+endif
```
marcrittinghaus: If `CONFIG_UK_ARCH` is `arm64` the option is `Config.uk` is not defined. I assume the `arm64` image will thus be `unikraft_-arm64`.

plat/common/x86/cpu_native.c
```diff
@@ -48,6 +48,11 @@ unsigned long read_cr2(void)
...
+#endif
```
marcrittinghaus:
```suggestion
#endif /* CONFIG_KVM_VMM_FIRECRACKER */
```

plat/common/x86/cpu_native.c
```diff
@@ -48,6 +48,11 @@ unsigned long read_cr2(void)
 
 void system_off(void)
 {
+#if CONFIG_KVM_VMM_FIRECRACKER
+	/* Trigger a reset via the PS/2 controller */
```
marcrittinghaus: If this is a power off for FC then the comment should reflect this and not say `reset` 😃

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
 	return (int)i;
 }
 
+/**
+ * Insert a new region into the memory region list. This extends
+ * ukplat_memregion_list_insert to carve out the area of any pre-existing
+ * overlapping regions. The virtual addresses will leave out any carved out
+ * region.
+ * If there is not enough space for all resulting regions, the function will
+ * insert as many as possible and then signal return an error.
```
marcrittinghaus:
```suggestion
 * insert as many as possible and then return an error.
```

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+ * @param list
+ *   The memory region list to insert the range into
+ * @param mrd
+ *   The memory range to insert
+ * @param min_size
+ *   The minimum size a inserted region has to have. Regions smaller than this
+ *   size won't be added. Settings this parameter zero will disable this
+ *   behavior.
```
marcrittinghaus:
```suggestion
 *   The minimum size an inserted region has to have. Regions smaller than this
 *   size won't be added. Setting this parameter to zero will disable this
 *   behavior.
```

include/uk/plat/memory.h
```diff
@@ -84,6 +84,27 @@ struct ukplat_memregion_desc {
...
+ * the physical address space.
+ *
+ * @param mrd
+ *   Pointer to the memory region descriptor to check against
+ * @param pstart
+ *   the start of the physical memory region
+ * @param pend
+ *   the end of the physical memory region
+ * @return
+ *   zero if the two specified regions have no overlap, a non-zero value
```
marcrittinghaus:
```suggestion
 *   Start of the physical memory region
 * @param pend
 *   End of the physical memory region
 * @return
 *   Zero if the two specified regions have no overlap, a non-zero value
```

include/uk/essentials.h
```diff
@@ -253,20 +253,12 @@ extern "C" {
...
  *        the first value that is outside of a range.
  */
 #define RANGE_OVERLAP(base0, len0, base1, len1)				\
-	(IN_RANGE((base1), (base0), (len0))				\
-	 || ((((base1) + (len1)) > (base0))				\
-	     && (((base1) + (len1)) <= (base0) + (len0)))		\
-	 || (((base1) <= (base0))					\
-	     && (((base1) + (len1)) >= ((base0) + (len0)))))
+	((((base1) + (len1)) > (base0))) &&				\
+	((base1) < ((base0) + (len0)))
```
marcrittinghaus: I would add another pair of parenthesis around the entire expression.

plat/kvm/x86/multiboot.c
```diff
@@ -22,20 +22,11 @@
...
-}
-
 static inline void mrd_insert(struct ukplat_bootinfo *bi,
 			      const struct ukplat_memregion_desc *mrd)
 {
 	int rc;
 
-	if (unlikely(mrd->len == 0))
-		return;
-
```
marcrittinghaus: Why is this removed?

plat/kvm/x86/multiboot.c
```diff
@@ -131,33 +124,21 @@ void multiboot_entry(struct lcpu *lcpu, struct multiboot_info *mi)
...
-					start = mrdp->pbase + mrdp->len;
-				}
-
-				if (end - start < PAGE_SIZE)
-					continue;
+				rc = ukplat_memregion_list_insert_split_phys(
+				    &bi->mrds, &mrd, __PAGE_SIZE);
+				if (rc < 0)
+					multiboot_crash("Unable to add region",
+							rc);
```
marcrittinghaus:
```suggestion
				rc = ukplat_memregion_list_insert_split_phys(
					&bi->mrds, &mrd, __PAGE_SIZE);
				if (unlikely(rc < 0))
					multiboot_crash("Unable to add region",
							rc);
```

plat/kvm/x86/multiboot.c
```diff
@@ -131,33 +124,21 @@ void multiboot_entry(struct lcpu *lcpu, struct multiboot_info *mi)
...
+				continue;
```
marcrittinghaus:
```suggestion
```

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+		if (pstart < mrdp->pbase) {
+			/* Some part of the inserted region is before the
+			 * overlapping region. Try to insert that part if it's
+			 * large enough.
+			 */
+			mrdc.pbase = pstart;
+			mrdc.vbase = pstart + voffset;
+			mrdc.len   = mrdp->pbase - pstart;
+
+			if (mrdc.len > min_size) {
```
marcrittinghaus: According to the function description it should be:
```suggestion
			if (mrdc.len >= min_size) {
```

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+				rc = ukplat_memregion_list_insert(list, &mrdc);
+				if (rc < 0)
```
marcrittinghaus: It is `unlikely` that one day I don't have to add a comment because of this? 😄

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+					return rc;
+			}
+		}
+
+		pstart = mrdp->pbase + mrdp->len;
+	}
+
+	if (pend - pstart <= min_size)
```
marcrittinghaus: Should be `<` according to the function description.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,87 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
...
+		return 0;
+
+	mrdc.pbase = pstart;
+	mrdc.vbase = pstart + voffset;
+	mrdc.len = pend - pstart;
+	/* Add the remaining region */
```
marcrittinghaus, line 192: For increased symmetry to the other case, I would add a newline.

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+		return;
+
+	mrd.pbase = cmdline_addr;
+	mrd.vbase = cmdline_addr;
+	mrd.len = cmdline_size;
+	mrd.type = UKPLAT_MEMRT_CMDLINE;
+	mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_MAP;
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "cmdline", sizeof("cmdline"));
+#endif
```
marcrittinghaus:
```suggestion
#endif /* CONFIG_UKPLAT_MEMRNAME */
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+		return;
+
+	mrd.type = UKPLAT_MEMRT_INITRD;
+	mrd.flags = UKPLAT_MEMRF_MAP | UKPLAT_MEMRF_READ;
+	mrd.vbase = initrd_addr;
+	mrd.pbase = initrd_addr;
+	mrd.len = initrd_size;
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "initrd", sizeof("initrd"));
+#endif
```
marcrittinghaus:
```suggestion
#endif /* CONFIG_UKPLAT_MEMRNAME */
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+	__sz cmdline_size;
+	int rc;
+
+	cmdline_addr = bp->ext_cmd_line_ptr;
+	cmdline_addr <<= 32;
+	cmdline_addr |= bp->hdr.cmd_line_ptr;
+
+	cmdline_size = bp->hdr.cmd_line_size;
+
+	if (cmdline_addr == 0)
```
marcrittinghaus:
```suggestion
	if (unlikely(cmdline_addr == 0))
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+		lxboot_crash(-EINVAL, "Command line address is zero");
+
+	if (cmdline_size == 0)
+		return;
+
+	mrd.pbase = cmdline_addr;
+	mrd.vbase = cmdline_addr;
+	mrd.len = cmdline_size;
+	mrd.type = UKPLAT_MEMRT_CMDLINE;
```
marcrittinghaus:
```suggestion
	mrd.len   = cmdline_size;
	mrd.type  = UKPLAT_MEMRT_CMDLINE;
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+	mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_MAP;
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "cmdline", sizeof("cmdline"));
+#endif
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+	if (rc < 0)
```
marcrittinghaus: 👀

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+	initrd_size |= bp->hdr.ramdisk_size;
+
+	if (initrd_addr == 0 || initrd_size == 0)
+		return;
+
+	mrd.type = UKPLAT_MEMRT_INITRD;
+	mrd.flags = UKPLAT_MEMRF_MAP | UKPLAT_MEMRF_READ;
+	mrd.vbase = initrd_addr;
+	mrd.pbase = initrd_addr;
+	mrd.len = initrd_size;
```
marcrittinghaus:
```suggestion
	mrd.type  = UKPLAT_MEMRT_INITRD;
	mrd.flags = UKPLAT_MEMRF_MAP | UKPLAT_MEMRF_READ;
	mrd.vbase = initrd_addr;
	mrd.pbase = initrd_addr;
	mrd.len   = initrd_size;
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+#ifdef CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "initrd", sizeof("initrd"));
+#endif
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+	if (rc < 0)
```
marcrittinghaus: 🦜

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+		mrd.len = end - start;
+
+		if (entry->type == LXBOOT_E820_TYPE_RAM) {
+			mrd.type = UKPLAT_MEMRT_FREE;
+			mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE;
+
+			rc = ukplat_memregion_list_insert_split_phys(
+			    &bi->mrds, &mrd, __PAGE_SIZE);
+			if (rc < 0)
+				lxboot_crash(rc, "Unable to add ram mapping");
```
marcrittinghaus:
```suggestion
			rc = ukplat_memregion_list_insert_split_phys(
			    	&bi->mrds, &mrd, __PAGE_SIZE);
			if (unlikely(rc < 0))
				lxboot_crash(rc, "Unable to add ram mapping");
```

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		mrd.type = UKPLAT_MEMRT_RESERVED;
+		mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_MAP;
+
+		/* We assume that reserved regions cannot
+		 * overlap with loaded modules.
+		 */
+		rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+		if (rc < 0)
```
marcrittinghaus: You know the drill 🙊

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+	bi = ukplat_bootinfo_get();
+	if (unlikely(!bi))
+		lxboot_crash(-EINVAL, "Incompatible or corrupted bootinfo");
+
+	if (unlikely(bp->hdr.boot_flag != 0xAA55))
+		lxboot_crash(-EINVAL, "Invalid magic number");
+
+	lxboot_init_cmdline(bi, bp);
+	lxboot_init_initrd(bi, bp);
+	lxboot_init_mem(bi, bp);
```
marcrittinghaus, line 155: Actually, I like that you split the function in different subroutines. We could do the same in multiboot.

plat/kvm/x86/lxboot.c
```diff
@@ -0,0 +1,161 @@
...
+
+			rc = ukplat_memregion_list_insert_split_phys(
+			    &bi->mrds, &mrd, __PAGE_SIZE);
+			if (rc < 0)
+				lxboot_crash(rc, "Unable to add ram mapping");
+			continue;
+		}
+
+		mrd.type = UKPLAT_MEMRT_RESERVED;
+		mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_MAP;
```
marcrittinghaus: On multiboot you put this into an `else` scope. I don't have a preference. Just make it equal so it is easy for the eye to spot the pattern.

marcrittinghaus (commented):
Hey @mschlumpp,

thanks for the PR! Looks really good. Just small nitpicks 😛

marcrittinghaus:
> ❗ This PR contains a major breaking change regarding output binary name. As part of this addition, differentiating the output artifact now results in, for example:
> 
> * `helloworld_qemu-x86_64`
> * `helloworld_fc-x86_64`
> 
> Making this comment visible for the upcoming release note.

@nderjung Do you think this is a problem besides having to add a note / update documentation? I don't know how kraftkit will react to this.

include/uk/essentials.h
jschlumpp: Rewrite-it-in-lisp :smile:

plat/kvm/x86/multiboot.c
jschlumpp: There was an iteration where the `list_insert` function took care of that. But now it's necessary again.

marcrittinghaus (approved):
Thank you very much @mschlumpp! Also thanks to @Krechals for providing the initial PoC that made this work possible 🚀 And of course @mogasergiu for the feedback!

PR #772, opened 2023-03-01
Add PIE support for Xen and KVM

mogasergiu:
### Prerequisite checklist

### Base target

 - Platform(s): `kvm`, `xen`
 - Application(s):

### Additional configuration

`CONFIG_OPTIMIZE_PIE`
`CONFIG_UKPLAT_MEMREGION_MAX_COUNT`

### Description of changes

Drop the `elf32-i386` linker flag through a python script that converts an ELF64's Header, Section Headers and Program Headers to their 32-bit equivalents.

Make the Unikernel position independent with the help of  `struct uk_reloc`, a new structure accompanied by helper symbol-generating macros and a script that analyzes the final Unikernel binary such symbols as well as the `.rela.dyn` section's entries to build a binary blob inside a new custom ELF section `.uk_reloc` made up of `struct uk_reloc` entries that an early self relocator can make use of.

### NOTE


mogasergiu:
v2:
- increase commit granularity
- Place the *dyn* related sections into SECT_STRIP_FLAGS to automatically get
stripped
- Place every *dyn* and .uk_reloc sections in a separate ld script that is
only considered if CONFIG_OPTIMIZE_PIE is enabled
- fix checkpatch warnings/errors
- add a new UK_RELOC_INJECT_COUNT to guarantee that uk_reloc does not spill
into other sections + mkukreloc.py checks the spill is not happening
- check against relocations generated by sections beyond .comment section
- wrap uk_reloc and elf64_to_32 scripts into build functions: build_elf64_to_32
and build_ukreloc
- remove unneeded arch specific `st_curr_baddr` in favor of `__BASE_ADDR` macro
- removed no longer needed `volatile` keyword leftovers from debugging
- tested on different UEFI machines (loaded at random addresses) with UBSAN
and ~28k relocations

mogasergiu:
v3:
- remove unnecessary final patching in `mkukreloc.py` as it doubles the script's execution time unnecessarily and makes the wrong assumption that offset in file == offset in memory, thus ending up causing wrong relocations in some unikernel images that do not happen to have the same layout by coincidence.
- rebase

dragosp27 (approved):
Great job!

Reviewed-by: Dragos Petre <dragos.petre27@gmail.com>

SerbanSo:
Looking good

Reviewed-by: Serban Sorohan serban.sorohan@gmail.com

SerbanSo (approved):

support/scripts/elf64_to_32.py
```diff
@@ -0,0 +1,207 @@
...
+    if opt.output is None:
+        raise Exception("Please provide the path to the output ELF32 binary")
+
+    elf64 = Elf64_to_32(opt.elf64_file)
+
+    elf32_ehdr = elf64.get_elf32_ehdr()
+    elf32_phdrs = elf64.get_elf32_phdrs()
+    elf32_shdrs = elf64.get_elf32_shdrs()
+
+    with open(opt.elf64_file, 'r+b') as elf64_file:
```
jschlumpp: Is there a reason to open it also as writeable?

support/scripts/elf64_to_32.py
```diff
@@ -0,0 +1,207 @@
...
+        elf64_buf = elf64_file.read()
+
+    # Create the empty file
+    with open(opt.output, 'w') as elf32_file:
```
jschlumpp: Is there are reason to create the file first before writing?

support/scripts/elf64_to_32.py
```diff
@@ -0,0 +1,207 @@
...
+        pass
+
+    with open(opt.output, 'r+b') as elf32_file:
```
jschlumpp: The file doesn't seem to be read so opening it for writing should be enough.

lib/ukreloc/include/uk/reloc.h
```diff
@@ -0,0 +1,122 @@
...
+ * and is seen as a relocation by the linker.
+ *
+ * @param sym
+ *   The symbol to use for the relocation.
+ */
+#if defined(__X86_64__)
+#define RELA_DYN_ENTRY_TYPE R_X86_64_64
+#elif defined(__ARM_64__)
+#define RELA_DYN_ENTRY_TYPE R_AARCH64_ABS64
+#endif
```
jschlumpp, line 35: I would also add a `#else` branch that uses an `#error` pragma to fail the compilation with a message that indicates that a new case has to be added.

support/scripts/mkukreloc.py
jschlumpp, line 1: I think the general direction was to avoid introducing further python scripts into the build process and instead implement these tools as small C programs using elfutils (libelf).

lib/ukreloc/reloc.lds.S
```diff
@@ -0,0 +1,62 @@
...
+
+SECTIONS
+{
+	. = ALIGN(__PAGE_SIZE);
+	.ukreloc :
+	{
+		_ukreloc_start = .;
+		. += 4;
+		KEEP(*(.ukreloc))
+	} :data
```
jschlumpp: You may want to use the [defines](https://github.com/unikraft/unikraft/blob/staging/plat/common/include/uk/plat/common/common.lds.h#L37-L45) for the segment annotations.

plat/kvm/x86/lcpu_start.S
jschlumpp, line 1: It might be a good idea to add a note if relocations in this file change, that the corresponding `IMPORT_START16_UKRELOC_SYM` statements also have to be updated.

jschlumpp (commented):
Added a few notes

support/scripts/elf64_to_32.py
mogasergiu: Hi Marco! Thank you a lot for the input :smiley:! Regarding your `elf64_to_32.py` comments, 4 months ago when I was writing the first versions of this script I was having two separate binaries: first I would just read the initial binary and then create the ELF32 equivalent. Unlike the current version that just prepends the ELF32 Headers. Indeed, the permissions wouldn't have made sense anyway even then, I think I was battling some permission issues at the time although I do not remember. I agree with you, I think I should have just done what I already do for [mkefi.py](https://github.com/unikraft/unikraft/pull/909/commits/e4a83a02b25cfac86414e1604710086dbd0f5e05#diff-7e2ddda8a9cccf09ea61481a4c03e5ffcff4bb1c0e671984681041a9d0e4acecR195) instead. So I am going to change these python lines accordingly :+1: .

lib/ukreloc/arch/x86_64/include/arch/reloc.h
```diff
@@ -54,6 +54,159 @@
...
+/* Expects the lower 32 bits of the base virtual address in %edi
+ * and the higher 32 bt in %esi (hopefully a KASLR randomized value).
+ * Also place the base load address in %edx. The macro can either create its
+ * own scratch stack to save register state or use the already existing one,
+ * if there.
+ *
+ * @param have_stack Boolean value to tell whether the caller already has a
+ *                   stack available or not
+ */
+.macro do_ukreloc32	have_stack:req
```
mogasergiu: I just noticed that this is going to break if, in the future, we make use of the MEMRT_KERNEL memory regions related to the ELF Segments as, unlike the C written relocator, it does not relocate the ELF Segment memory regions...

Unsure whether it is worth doing this iteration now or later when the time we will actually make use of the ELF Segment memory regions comes.

mogasergiu:
Note for me. Rename `ukreloc` and related definitions to use `uk_reloc`

plat/kvm/x86/multiboot.S
```diff
@@ -77,10 +78,18 @@ ENTRY(_multiboot_entry)
 	cmpl	$MULTIBOOT_BOOTLOADER_MAGIC, %eax
 	jne	no_multiboot
 
-	movl	$lcpu_boot_startup_args, %edi	/* startup args for boot CPU */
+	/* Hardcoding for now I guess... */
+	movl    $0x00100000, %edi
```
mogasergiu, line 82: These 3 lines I do not like, maybe some way to add an additional configuration option for the load address in the linker script, but while allowing three different versions (e.g. if a user wants to build KVM, Xen and linuxu images with the same configurations at the same time). So RFC.

lib/ukreloc/include/uk/reloc.h
```diff
@@ -0,0 +1,122 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UKRELOC_H__
+#define __UKRELOC_H__
+
+#define UKRELOC_FLAGS_PHYS_REL		(1 << 0)
```
razvanvirtan, line 10: Is there one extra-tab here? :thinking:

lib/ukreloc/arch/x86_64/include/arch/reloc.h
mogasergiu: Introduced [`do_ukreloc_kmrds`](https://github.com/unikraft/unikraft/pull/772/commits/16ac1dbffc94f7570aab07c4f858893e6bf4cc54) for this.

plat/linuxu/arm64/entry64.S
```diff
@@ -35,6 +35,13 @@
 
 .section .text
 
+/* For now, although inaccurate, set the
```
michpappas, line 38: Although I understand what this comment refers to, it's only because I'm looking at it through the context of the PIE PR. Yet chances are that it's not clear to someone looking at this file out of scope. Perhaps you can provide some reasoning on both why this is added here, and why it doesn't matter in this platform. Same to the other variants where this comment applies.

support/scripts/elf64_to_32.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # Check EI_DATA for endianness
+        if self.elf64_ehdr[5] == 1:
+            self.endian = 'little'
+        else:
+            self.endian = 'big'
+        _endian = '>' if self.endian == 'big' else '<'
+
+        # Check e_machine
+        if sum(self.elf64_ehdr[18:20]) == ELF_MACHINE['arm64']:
```
michpappas: From my understanding this script is not used in `arm`/`arm64`, so perhaps this and the ELF magic for arm are not required?

plat/kvm/Linker.uk
```diff
@@ -67,6 +67,9 @@ $(KVM_IMAGE): $(KVM_IMAGE).dbg
 			$< -o $@ 2>&1 | \
 			{ grep -v "Empty loadable segment detected" || true; })
 	$(call build_bootinfo,$@)
+ifeq ($(ELF64_TO_32),y)
+	$(call build_elf64_to_32,$@)
```
michpappas, line 76: Leaving this here as GitHub does not allow commenting on the commit messages. Since we now do support lxboot (to some extent), the commit message can be updated to use a more accurate description.

lib/ukreloc/Config.uk
```diff
@@ -0,0 +1,3 @@
+config LIBUKRELOC
+	bool "ukreloc: Self relocation and symbol based relocations"
+	default y
```
michpappas: Why does `libukreloc` need to be enabled by default?

lib/ukreloc/reloc.lds.S
```diff
@@ -0,0 +1,63 @@
...
+	dynamic PT_DYNAMIC;
+}
+
+SECTIONS
+{
+	. = ALIGN(__PAGE_SIZE);
+	.ukreloc :
+	{
+		_ukreloc_start = .;
+		. += 4;
```
michpappas: Ah, the good old magic 4... 🔮  Maybe use an ALIGN here?

plat/common/arm/cpu_native.c
```diff
@@ -56,10 +56,8 @@ void reset(void)
 	 * NO PSCI or invalid PSCI method, we can't do reset, just
 	 * halt the CPU.
 	 */
-	if (!smccc_psci_call) {
-		uk_pr_crit("Couldn't reset system, HALT!\n");
```
michpappas: So these messages are generally useful, and I know for fact that the pl011 and ns16550 drivers have no issue if the UART is not initialized, as they just return, so if the problem is caused by the x86 drivers a fix can be applied there instead?

If you do apply this change, also remember to update the commit message of the previous commit:
```
    `apply_ukreloc` may cause `UBSAN` to issue false positives, in
    the case of the `x86` architecture which leads to the Unikernel crashing
    since the serial console is not initialized yet, as it is mostly intended
    to be called within the early self relocator. So make sure that
    `UBSAN` does not touch `apply_ukreloc` because it makes sense, for
    `x86`, to make unaligned accesses (especially in Kernel Space) in order
    to resolve non-compliant relocations.
```
With the opportunity, I find the argumentation regarding UART a bit weak here. If unaligned accesses are normal here (which as discussed they are on x86 early boot code), then that should be enough reasoning regardless of the UART.

lib/ukreloc/arch/arm64/include/arch/reloc.h
michpappas, line 1: The plat-rearch way is to create a hierarchy as `arch/<arch>/include/uk/asm/` and then in your `include/uk/reloc.h` you do
```
#include <asm/reloc.h>
```
To get this to work you need to update your `Makefile.uk` to also add the arch-specific path to the `CINCLUDES`.

And finally update any code to indluce `uk/reloc.h` instead of `arch/reloc.h`.

arch/arm/arm64/ctx.S
```diff
@@ -132,7 +132,8 @@ ENTRY(ukarch_ctx_switch)
 	 * Record the restore point for switch out thread to restore
 	 * its callee-saved registers in next switch to time.
 	 */
-	ldr x30, =restore_point
+	adrp x30, restore_point
```
michpappas, line 135: Any reason `ur_ldr` is not used here?

lib/ukreloc/reloc.c
```diff
@@ -79,18 +78,8 @@ void __used do_ukreloc(__paddr_t r_paddr, __vaddr_t r_vaddr)
...
-	 * no been resolved yet and contains the link time address.
-	 */
-	ukplat_memregion_foreach(&mrdp, UKPLAT_MEMRT_KERNEL, 0, 0) {
-		mrdp->pbase -= (__paddr_t)lt_baddr;
-		mrdp->pbase += r_paddr;
-		mrdp->vbase -= (__vaddr_t)lt_baddr;
-		mrdp->vbase += r_vaddr;
-	}
+	/* Relocate UKPLAT_MEMRT_KERNEL memory region descriptors */
+	do_ukreloc_kmrds(r_paddr, r_vaddr);
```
michpappas: This breaks builds that don't enable `OPTIMIZE_PIE`, maybe make it conditional.

support/scripts/mkukreloc.py
```diff
@@ -0,0 +1,325 @@
...
+# Offset | Info | Type | Sym. Value | Sym. Name + Addend
+# However, since this is linked with -static-pie, all that we need can be
+# extracted from the .rela.dyn section(hopefully no .rel section since this
+# script does not account for those) as indicated by .dynamic. In our case
+# this section is expected to be filled with R_X86_64_RELATIVE or
+# R_AARCH64_RELATIVE (in the case of the latter, the E may not fit in the
+# command's output...).
+# Luckily we do not need to deal with .symtab since `readelf -r` nicely shows
+# the corresponding st_value.
+R_RELATIVE_TYPE_EXP = r'R_(?:X86_64|AARCH64)_RELATIV[E]{0,1}'
```
michpappas, line 44: Perhaps it would be useful to additionally check for unexpected relocation types.

Config.uk
```diff
@@ -95,6 +95,13 @@ config OPTIMIZE_LTO
 		will increase overall building time but creates more efficient
 		Unikraft binaries.
 
+config OPTIMIZE_PIE
```
razvand, line 98: Why is this `OPTIMIZE_PIE` and not just `PIE`? It's not an optimization, is it?

Makefile.uk
```diff
@@ -88,15 +88,32 @@ M4FLAGS      += -DUK_VERSION=$(UK_VERSION).$(UK_SUBVERSION)
 # If GCC supports "-no-pie" flag, we will add this flag to link flags to
 # override "pie" option, because some distributions will set
 # "--enable-default-pie" by default.
+ifeq ($(CONFIG_OPTIMIZE_PIE),y)
+COMPFLAGS-$(call gcc_version_ge,6,1)	+= -fPIE
+LDFLAGS-$(call gcc_version_ge,6,1)	+= -static-pie
```
razvand: From my understanding, the `static-pie` flag is used beginning with GCC 8: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81498

Makefile.uk
```diff
@@ -88,15 +88,32 @@ M4FLAGS      += -DUK_VERSION=$(UK_VERSION).$(UK_SUBVERSION)
...
+LDFLAGS-$(call have_clang)	+= -z notext -z norelro
+SECT_STRIP_FLAGS-$(CONFIG_OPTIMIZE_PIE) += -R .dynamic
+SECT_STRIP_FLAGS-$(CONFIG_OPTIMIZE_PIE) += -R .dynsym
+SECT_STRIP_FLAGS-$(CONFIG_OPTIMIZE_PIE) += -R .dynstr
+SECT_STRIP_FLAGS-$(CONFIG_OPTIMIZE_PIE) += -R .rela.dyn
+else
 COMPFLAGS-$(call gcc_version_ge,6,1)	+= -fno-PIC
 LDFLAGS-$(call gcc_version_ge,6,1)	+= -no-pie
+COMPFLAGS-$(call have_clang)	+= -fno-PIC
+LDFLAGS-$(call have_clang)	+= -no-pie
```
razvand, line 111: Why only for clang? These should apply to GCC as well, right? In case the `OPTIMIZE_PIE` option is not selected.

lib/ukreloc/Config.uk
```diff
@@ -0,0 +1,3 @@
+config LIBUKRELOC
+	bool "ukreloc: Self relocation and symbol based relocations"
```
razvand:
```suggestion
	bool "ukreloc: Self relocation and symbol-based relocations"
```

lib/ukreloc/arch/arm64/include/arch/reloc.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef __ARM_64_UK_RELOC_H__
+#define __ARM_64_UK_RELOC_H__
+
+#include "../../../include/uk/reloc.h"
+
+#ifdef __ASSEMBLY__
+
+/*
+ * Macro to replaced the non-relocation fiendly `ldr` instruction. The usage
```
razvand:
```suggestion
 * Macro to replace the non-relocation-friendly `ldr` instruction. The usage
```

lib/ukreloc/arch/x86_64/include/arch/reloc.h
```diff
@@ -0,0 +1,220 @@
...
+ */
+
+#ifndef __X86_64_UK_RELOC_H__
+#define __X86_64_UK_RELOC_H__
+
+#include "../../../include/uk/reloc.h"
+
+#ifdef __ASSEMBLY__
+
+/* Relocation friendly ur_mov to replace mov instructions incompatible with
```
razvand:
```suggestion
/* Relocation-friendly ur_mov to replace mov instructions incompatible with
```

lib/ukreloc/arch/x86_64/include/arch/reloc.h
```diff
@@ -0,0 +1,220 @@
...
+ * a PIE binary. Usage example:
+ * ```
+ * ur_mov	gdt64_ptr, %eax, 4
+ * ```
+ * The above will make mkukreloc.py process the symbol gdt64_ptr_ukeloc_imm4
```
razvand:
```suggestion
 * The above will get mkukreloc.py to process the symbol gdt64_ptr_ukeloc_imm4
```

lib/ukreloc/arch/x86_64/include/arch/reloc.h
```diff
@@ -0,0 +1,220 @@
...
+#define SAVE_REG32_COUNT				6
+
+/* Expects the lower 32 bits of the base virtual address in %edi
+ * and the higher 32 bt in %esi (hopefully a KASLR randomized value).
+ * Also place the base load address in %edx. The macro can either create its
+ * own scratch stack to save register state or use the already existing one,
+ * if there.
+ *
+ * NOTE: Unlike its C-variant, do_ukreloc(), this does not relocate
+ * the Kernel's static, build-time memory region descriptors!
```
razvand:
```suggestion
 * the kernel's static, build-time memory region descriptors!
```

plat/kvm/x86/multiboot.S
razvand: I would leave them as such for the time being. Let's have something working first and wait for the time when a configurable setup is required.

razvand (changes requested):
@mogasergiu, did a first quick pass. Will also test extensively and come back with another deeper pass.

lib/ukreloc/include/uk/reloc.h
```diff
@@ -0,0 +1,217 @@
...
+#define UKRELOC_PLACEHOLDER		0xB00B
```
razvand:
```suggestion
#define UKRELOC_PLACEHOLDER		0xB0B0
```

lib/ukreloc/include/uk/reloc.h
```diff
@@ -0,0 +1,217 @@
...
+#define UKRELOC_SIGNATURE		0xBADB0111
```
razvand:
```suggestion
#define UKRELOC_SIGNATURE		0x0BADB0B0
```

lib/ukreloc/include/uk/reloc.h
```diff
@@ -0,0 +1,217 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UKRELOC_H__
+#define __UKRELOC_H__
```
razvand:
```suggestion
#ifndef __UK_RELOC_H__
#define __UK_RELOC_H__
```

michpappas (changes requested):
@mogasergiu great work 🚀 Some minor review comments below.

Config.uk
mogasergiu: I took @skuenzer's [suggestion](https://github.com/unikraft/unikraft/pull/239#discussion_r668293448) into consideration.

Makefile.uk
mogasergiu: I need this to be `-static-pie` and not just `-pie`, because otherwise you are risking having GOT-related relocations generated which I do not and will not handle just to support an older version of the compiler because it does not make any sense to have this kind of relocations present in a kernel.

EDIT: I am going to have a try with `-static -pie` though in the hopes that it will work the same :crossed_fingers:

Makefile.uk
mogasergiu: But they are, right above these lines, aren't they? :confused:

arch/arm/arm64/ctx.S
mogasergiu: IIRC the inclusion path for `arch/` assembly files do not take libraries into account. So I left it like this in the hopes that plat re-arch will fix this.

lib/ukreloc/Config.uk
mogasergiu: `libukreloc` needs to be enabled by default to avoid surrounding calls to `ur_*` macros with `#ifdef's`. The `ur_*` macro's, when `CONFIG_OPTIMIZE_PIE` is disables, revert back to using absolute addressing.

lib/ukreloc/reloc.c
mogasergiu: I will create a stub for it just like for `do_ukreloc`

lib/ukreloc/reloc.lds.S
mogasergiu: That's supposed to be there to make room for the 4byte signature, but you're right I should have a locally defined macro for it. `ALIGN`? What for?

support/scripts/mkukreloc.py
mogasergiu: That is done by this [check](https://github.com/unikraft/unikraft/pull/772/files#diff-1b6f1a9686178d25fb2b6e0f9ce8a803518660fed6a76c425c53a78df8e44acfR307). If the relocation is unexpected it will not be detected through the regex, leading to us finding less relocations than `.dynamic` reports.

Makefile.uk
mogasergiu: :disappointed: `readelf: Error: no .dynamic section in the dynamic segment`. I would leave it with `-static-pie` for now... If the time comes that this would cause issues... I will open up an issue and look into it.

razvand (changes requested):
For commit https://github.com/unikraft/unikraft/pull/772/commits/7d3053e9df912d9df0955f2da06143f4cf5490fc, use `program headers, the last ...` (aka lower case `p` and `h` and `t`).

plat/common/include/uk/plat/common/sections.h
```diff
@@ -43,6 +43,9 @@
 /* _dtb: start of device tree */
 extern char _dtb[];
 
+/* _base_addr: base reference address as of link time */
```
razvand: I think this comment could be improved. If my understanding is correct, I would reword this to "This is the base address of the unikernel image. It is set at link time. It is used to compute other addresses, as reference to this.".

Makefile.uk
razvand: I'm stupid. You're right.

Makefile.uk
razvand: My point was that you have to check for GCC >= 8 for this to work. If GCC < 8, you have to issue an error when `CONFIG_OPTIMIZE_PIE` is enabled.

support/scripts/elf64_to_32.py
```diff
@@ -0,0 +1,201 @@
...
+        # Save previous ELF64
+        elf64_buf = elf32.read()
+
+        # Go back to the beginning
+        elf32.seek(0)
+
+        # Write equivalent ELF32 header
+        elf32.write(elf32_ehdr)
+
+        # Write the ELF32 equivalent Program Headers at the end of the binary
```
mogasergiu: s/binary/ELF32 header

razvand (changes requested):

razvand (changes requested):
Be wary of the reference to `ukplat_memregion_list_coalesce()` which is only later defined in #848. Other than that, this is a great PR: commits, comments, everything looks great.

I'll do extensive testing after you integrate all changes.

lib/ukreloc/reloc.lds.S
michpappas: Okay, in that case a `#define` would do

support/scripts/mkukreloc.py
michpappas: Yes I did not expect that this is done by the regex of course. Didn't notice the [check](https://github.com/unikraft/unikraft/pull/772/files#diff-1b6f1a9686178d25fb2b6e0f9ce8a803518660fed6a76c425c53a78df8e44acfR307), thanks 👍🏼

lib/ukreloc/reloc.c
michpappas: I think the right way is to make it conditional, similarly to calls from other libraries. As discussed, in the future the `ur_` macros will become unconditionally available and the library won't be enabled by default, but that doesn't mean that all other definitions from this library will become unconditionally available too.

mogasergiu: I do not like having `#ifdef`'s that could have been easily hidden away from the client code. I like having something similar to what linux does: if the CONFIG for this function is not enabled, then use a stub instead. If you place `#ifdef`'s throughout client code it's going to end up being an unreadable, unmaintainable mess. My opinion is that this kind of situations should be transparently handled by the server of the API's.

arch/arm/arm64/ctx.S
michpappas: Yes that's true. Let's do that then 👍🏼

lib/ukreloc/reloc.c
michpappas: Linux is not compiled into individual libraries. In any case, using `#ifdef` is the way currently done, so any proposal to change this scheme shoud be discussed in a broader scope rather being applied ad-hoc.

mogasergiu: > Linux is not compiled into individual libraries.

Fair, but the build process is largely the same. The burden of `#ifdef`-ing should not fall on the client code. IMO it would be better if it would be handled by the build system.

> In any case, using #ifdef is the way currently done,

And do you not agree with me that it looks like a mess? It's enough to look at `ukboot/boot.c`.

> any proposal to change this scheme shoud be discussed in a broader scope rather being applied ad-hoc.

This library is going to stay implicitly enabled for a long time. When that does not happen I think we should put the burden of `#ifdef`-ing on the build system, e.g. if the library is not enabled, include a dummy header with stubs from `include/uk`.

> And do you not agree with me that it looks like a mess

If you agree with me in this regard, why not begin with this library.

michpappas: > This library is going to stay implicitly enabled for a long time

Actually not, the plan is to address this in the next release.

> If you agree with me in this regard, why not begin with this library.

Because it is not just my call, and it needs to be agreed by the rest of the maintainers.

michpappas: > I will create a stub for it just like for `do_ukreloc`

Actually I was referring to calls to `do_ukreloc` and `do_ukreloc_mrds`, but accidentally placed my comment in the definition instead.

The same comment applies to `do_ukreloc`.

mogasergiu: I know, I still do not agree.

michpappas (approved):
After offline discussion with @mogasergiu, @mschlumpp, @razvand there are still open questions, yet due to: 1. the extensive length of this PR, 2. the amount of testing required with every update, and 3. the time pressure to the upcoming release the consensus was to merge as is and provide a follow-up update after the maintainers team concludes on the remaining topics.

CC: @skuenzer 

razvand (approved):

PR #774, opened 2023-03-02
build: Implement a more robust compiler detection

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Instead of checking the compiler executable name, inspect the compiler's version output. This also works when there are symlinks involved or the executable name has additional text such as version numbers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -629,6 +629,8 @@ CFLAGS		+= -DCC_VERSION=$(CC_VERSION)
 CXXFLAGS	+= -DCC_VERSION=$(CC_VERSION)
 GOCFLAGS	+= -DCC_VERSION=$(CC_VERSION)
 
+CC_NAME		:= $(shell $(CC) -v 2>&1 | grep -E "\\w+ version" | sed 's/\s.*$$//')
+
 # ensure $(BUILD_DIR)/kconfig, $(BUILD_DIR)/include and $(BUILD_DIR)/include/uk exists
 $(call mk_sub_build_dir,kconfig)
 $(call mk_sub_build_dir,include)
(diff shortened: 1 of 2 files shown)


vladandrew:
Hi Marco,

Just tested this and it works great. I was wondering, in what case would that sed be useful?

jschlumpp:
Hey,
`sed` is used to extract the first word, which is the compiler name. So in the case of `gcc --version`:
```
gcc (GCC) 12.2.1 20230201
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
It would retrieve `gcc` and store it in the variable.

vladandrew:
Hmm, on my side it doesn't seem to match correctly:

```bash
❯ gcc --version | head -n1 | sed 's/\s.*$$//'
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
```

I think there is an extra `$` at the end:

```bash
❯ gcc --version | head -n1 | sed 's/\s.*$//'
gcc
```

jschlumpp:
The double dollar sign is necessary because you need to escape dollar signs in the Makefile. So in the actually executed `sed` command there is only one dollar sign left.

mariasfiraiala (commented):
Thanks @mschlumpp! I've tested this on `x86` and it works fine with `gcc` and `clang`, however, on `AArch64` with both `aarch64-linux-gnu-gcc` and `gcc-arm-11.2-2022.02-x86_64-aarch64-none-linux-gnu` the build fails with

```console
/home/maria/freestanding/unikraft/arch/arm/arm64/Makefile.uk:93: *** Require GCC version >= 4.9 found 11.2.  Stop.
```

jschlumpp:
@mariasfiraiala I pushed an update, can you check whether the error still occurs?

mariasfiraiala (approved):
It's working fine now, thanks @mschlumpp.

vladandrew:
All good!

vladandrew (approved):
All good!

skuenzer (approved):

Issue #766, opened 2023-02-17
lib/ukboot: Critical error when running `app-helloworld` on linuxu


RaduNichita:
### Describe the bug

When running the linuxu application for helloworld, it crashes due to a problem with memory allocator from `boot.c`

### Steps to reproduce

Run the [app-helloworld](https://github.com/unikraft/app-helloworld) application with Unikraft v0.12.

### Expected behavior

The program to dsplay the message "Hello world".

### Which architectures were you using or does this bug affect?

`linuxu`

### Which platforms were you using or does this bug affect?

`x86_64`

### Relevant log output

```shell
[ 2692.514525] Info: [libukboot] <boot.c @  245> Unikraft constructor table at 0x40f000 - 0x40f008
[ 2692.514558] Info: [libukboot] <boot.c @  264> Initialize memory allocator...
[ 2692.514571] CRIT: [libukboot] <boot.c @  268> Failed to initialize memory allocator
```

Daparrag:
hi @RaduNichita 

I add some changes that will fix this problem.

razvand:
**Hackathon points**: 40

PR #782, opened 2023-03-07
support/scripts:Fix Python ascii codec can't decode byte

bokket:
### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): ubuntu20.04
 - Application(s): [e.g. `app-helloworld` or N/A]

### Additional configuration
none

### Description of changes
Ref to https://stackoverflow.com/questions/21129020/how-to-fix-unicodedecodeerror-ascii-codec-cant-decode-byte
I think Python has exceeded the length of unicode as an intermediate encoding when doing the encoding conversion.
image

I changed the decoding to 'utf-8'. It makes it in the `KVM image`.But failed in the `linuxu-x86_64`.This problem links #766 
image
image
image

### DIFF
--- a/support/scripts/mkbootinfo.py
+++ b/support/scripts/mkbootinfo.py
@@ -64,7 +64,7 @@ def main():
     # to create a binary blob that has exactly this size so we can replace it
     # in the binary.
     out = subprocess.check_output(["objdump", "-h", opt.kernel])
-    match = re.findall(SECINFO_EXP, out.decode('ASCII'), re.MULTILINE)
+    match = re.findall(SECINFO_EXP, out.decode('utf-8'), re.MULTILINE)
 
     if len(match) != 1:
         raise Exception(".uk_bootinfo section not found")
@@ -73,7 +73,7 @@ def main():
 
     # Retrieve info about ELF segments in unikernel
     out = subprocess.check_output(["objdump", "-p", opt.kernel])
-    phdrs = re.findall(PHDRS_EXP, out.decode('ASCII'), re.MULTILINE)
+    phdrs = re.findall(PHDRS_EXP, out.decode('utf-8'), re.MULTILINE)
 
     # The boot info is a struct ukplat_bootinfo
     # (see plat/common/include/uk/plat/common/bootinfo.h) followed by a list of


razvand:
Hi, @bokket , could you please elaborate how can we make the issue appear? And also add that to the commit description message. And add a complete name + email signature as the commit author and as the signed off author.

flpostolache:
Hello, @bokket! I tried replicating the problem but I couldn't reproduce the crash. I noticed from your screenshots that you are using a different system and keyboard language. Can you please specify both so I can try to reproduce the problem?

bokket:
> Hello, @bokket! I tried replicating the problem but I couldn't reproduce the crash. I noticed from your screenshots that you are using a different system and keyboard language. Can you please specify both so I can try to reproduce the problem?

Sorry for my later reply.I do use the Chinese character set,Python3 uses utf-8 by default.The function parameter may be str to convert from unicode to encoded text / bytes.
“The issue is that when you call str(), python uses the default character encoding to try and encode the bytes you gave it, which in your case are sometimes representations of unicode characters. To fix the problem, you have to tell python how to deal with the string you give it by using .encode('whatever_unicode'). Most of the time, you should be fine using utf-8.”
It‘s iusse ref:
https://stackoverflow.com/questions/9942594/unicodeencodeerror-ascii-codec-cant-encode-character-u-xa0-in-position-20
https://docs.python.org/2.7/howto/unicode.html

bokket:
> 

Sorry.I will sign my personal information and  squash all commits and force push(if the pr make happy or be approved:)

bokket:
> Hi, @bokket , could you please elaborate how can we make the issue appear? And also add that to the commit description message. And add a complete name + email signature as the commit author and as the signed off author.

Sorry.I will sign my personal information and squash all commits and force push in next days(if the pr make happy or be approved:).I have other things to do

flpostolache:
I managed to recreate the issue, @razvand. You have to make sure to have your system language in chinese . ( I used traditional chinese ). Then, if you try to build a `hello world` for kvm for example, the mkbootinfo script will fail. @bokket, I think the change is ok.

marcrittinghaus:
Thanks @bokket for the PR!

Could you please add a short description about the issue and the fix in the commit message. Please see other commits for the format (line length, new lines etc.). Please also add a space in the commit title after `support/scripts:`.

@maniatro111: I assume you verified that the patch fixes the problem, right?

flpostolache:
@marcrittinghaus, the patch solves the issue. I've verified it again now. One more question that is kinda in topic. In the `mkbootinfo.py` script there is also the `-n` unused option which from my understanding also includes the app's name in the image. Should those bytes also be converted with utf-8 format or is it ok to leave it as ASCII?

razvand:
@bokket , could you take care of the updates requested? Otherwise, would it be OK if we updated your PR to include it in the next release?

nderjung:
Closing as https://github.com/unikraft/unikraft/pull/873 is merged.

PR #784, opened 2023-03-07
Add per-lcpu variable definition

andraprs:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64` , `arm64`
 - Platform(s): `kvm`
 - Application(s): `app-helloworld`

### Additional configuration

Set CONFIG_UKPLAT_LCPU_MAXCOUNT to a value higher than 1, the default one e.g. CONFIG_UKPLAT_LCPU_MAXCOUNT=3.

### DIFF
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -229,6 +229,22 @@ int ukplat_lcpu_wakeup(const __lcpuidx lcpuidx[], unsigned int *num);
 #define ukplat_lcpu_count()	(1)
 #endif /* CONFIG_HAVE_SMP */
 
+/* Per-LCPU variable definition */
+
+#define UKPLAT_PER_LCPU_DEFINE(var_type, var_name) \
+	var_type var_name[CONFIG_UKPLAT_LCPU_MAXCOUNT]
+#define ukplat_per_lcpu(var_name, lcpu_idx) \
+	var_name[lcpu_idx]
+#define ukplat_per_lcpu_current(var_name) \
+	ukplat_per_lcpu(var_name, ukplat_lcpu_idx())
+
+#define UKPLAT_PER_LCPU_ARRAY_DEFINE(var_type, var_name, size) \
+	var_type var_name[CONFIG_UKPLAT_LCPU_MAXCOUNT][size]
+#define ukplat_per_lcpu_array(var_name, lcpu_idx, idx) \
+	var_name[lcpu_idx][idx]
+#define ukplat_per_lcpu_array_current(var_name, idx) \
+	ukplat_per_lcpu_array(var_name, ukplat_lcpu_idx(), idx)
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/uksched/include/uk/sched_impl.h
+++ b/lib/uksched/include/uk/sched_impl.h
@@ -84,11 +84,14 @@ unsigned int uk_sched_thread_gc(struct uk_sched *sched);
 static inline
 void uk_sched_thread_switch(struct uk_thread *next)
 {
-	struct uk_thread *prev = __uk_sched_thread_current;
+	struct uk_thread *prev;
+
+	prev = ukplat_per_lcpu_current(__uk_sched_thread_current);
 
 	UK_ASSERT(prev);
 
-	__uk_sched_thread_current = next;
+	ukplat_per_lcpu_current(__uk_sched_thread_current) = next;
+
 	prev->tlsp = ukplat_tlsp_get();
 	if (prev->ectx)
 		ukarch_ectx_store(prev->ectx);
--- a/lib/uksched/include/uk/thread.h
+++ b/lib/uksched/include/uk/thread.h
@@ -34,6 +34,7 @@
 #include <uk/arch/lcpu.h>
 #include <uk/arch/time.h>
 #include <uk/arch/ctx.h>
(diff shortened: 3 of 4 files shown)


andraprs:
I dropped the 2 commits related to the paging and ukvmem logic, so that the changes for the virtual memory are handled separately and looked into more thoroughly.

Later edit:

I included the patches corresponding to these 2 commits in this GitHub branch: https://github.com/andraprs/unikraft/tree/andraprs/per-lcpu-var-paging-ukvmem.

razvanvirtan (approved):
Hi, @andraprs!
I checked and tested this, everything looks good and I think it will be really useful in our further SMP efforts, thanks!
Also, as a small note, I tested the per-lcpu variables support on ARM64 and it works, so I think the scope of this PR can be safely extended.

andraprs:
> Hi, @andraprs! I checked and tested this, everything looks good and I think it will be really useful in our further SMP efforts, thanks! Also, as a small note, I tested the per-lcpu variables support on ARM64 and it works, so I think the scope of this PR can be safely extended.
> 
> Reviewed-by: Razvan Virtan [virtanrazvan@gmail.com](mailto:virtanrazvan@gmail.com)

Thank you for the shared info, Razvan, and for trying out an Arm setup as well. :)

include/uk/plat/lcpu.h
```diff
@@ -41,6 +41,27 @@
...
+	ukplat_per_lcpu_var(var_name, ukplat_lcpu_idx())
+
+#define UKPLAT_PER_LCPU_FIRST_MULTI_DIM_VAR_DEFINE(var_type, var_name, size) \
+	var_type var_name[CONFIG_UKPLAT_LCPU_MAXCOUNT][size]
+#define ukplat_per_lcpu_first_multi_dim_var(var_name, lcpu_idx, idx) \
+	var_name[lcpu_idx][idx]
+#define ukplat_per_lcpu_first_multi_dim_var_current(var_name, idx) \
+	ukplat_per_lcpu_first_multi_dim_var(var_name, ukplat_lcpu_idx(), idx)
+
+#define UKPLAT_PER_LCPU_SECOND_MULTI_DIM_VAR_DEFINE(var_type, var_name, size) \
```
skuenzer: What is the idea behind the two definitions (first and second)? I understand that you I swap the index and end up with different localities...

include/uk/plat/lcpu.h
```diff
@@ -41,6 +41,27 @@
 extern "C" {
 #endif
 
+#define UKPLAT_PER_LCPU_VAR_DEFINE(var_type, var_name) \
```
skuenzer: I am wondering if we can shorten the macro names a bit, so that we do not reach the 80 char limit too quickly:
For example:
`UKPLAT_PER_LCPU_VAR_DEFINE()` -> `UKPLAT_PER_LCPU_DEFINE()`
`UKPLAT_PER_LCPU_FIRST_MULTI_DIM_VAR_DEFINE` -> `UKPLAT_PER_LCPU_ARRAY_DEFINE`

michpappas: +1 from me, I had the same thought on that.

include/uk/plat/lcpu.h
michpappas: The array support was a feature request from me, as with `UKPLAT_PER_LCPU_VAR_DEFINE` you could do `UKPLAT_PER_LCPU_VAR_DEFINE(<type>, arr[count])` to define arrays in the form of `arr[count][lcpu]` but not `arr[lcpu][count]`. I cannot think of a use-case for `arr[count][lcpu]`.

andraprs: Thank you, both, for the feedback. I can keep only one definition for the array.

include/uk/plat/lcpu.h
andraprs: Ack, I'll update to the proposed shorter naming. Thank you.

include/uk/plat/lcpu.h
```diff
@@ -41,6 +41,20 @@
...
+#define ukplat_per_lcpu_current(var_name) \
+	ukplat_per_lcpu(var_name, ukplat_lcpu_idx())
+
+#define UKPLAT_PER_LCPU_ARRAY_DEFINE(var_type, var_name, size) \
+	var_type var_name[CONFIG_UKPLAT_LCPU_MAXCOUNT][size]
+#define ukplat_per_lcpu_array(var_name, lcpu_idx, idx) \
+	var_name[lcpu_idx][idx]
+#define ukplat_per_lcpu_array_current(var_name, idx) \
+	ukplat_per_lcpu_array(var_name, ukplat_lcpu_idx(), idx)
+
```
marcrittinghaus: Hey @andraprs,
I would move this down here, because it practically depends on `ukplat_lcpu_idx()` and also feels a bit out of context right at the top:
```C
#else /* CONFIG_HAVE_SMP */
#define ukplat_lcpu_idx()	(0)
#define ukplat_lcpu_count()	(1)
#endif /* CONFIG_HAVE_SMP */

/* Per-LCPU variable definition */
<----------

#ifdef __cplusplus
}
#endif
```

Other than that I think this is fine 😃

andraprs: Done. Thank you for review.

marcrittinghaus (approved):
Thanks @andraprs 😃

PR #795, opened 2023-03-11
lib/vfscore: Add test suite for dup() sys call

toast3rx:
Add check if dup works as intended for valid/invalid file descriptor

### Prerequisite checklist

### Base target

 - Architecture(s):x86_64, arm

### Description of changes

Create two tests that checks if dup() system call creates a new file descriptor when used on a valid fd and returns an error when it's called with -1 as argument.


razvand:
Hi, @toast3rx , thanks for the contribution.

`vfscore` is a huge beast and part of it (the top-layer) should actually be part of a standard C library such as [`nolibc`](https://github.com/unikraft/unikraft/tree/staging/lib/nolibc) or [`lib-musl`](https://github.com/unikraft/lib-musl). The `dup()` system call tests fall in that category. They should be tested using [`lib-libc-test`](https://github.com/unikraft/lib-libc-test). See if `lib-libc-test` has `dup()` tests and, if not, add it to them.

That being said, the testing infrastructure for `vfscore` is good. But aim to use it for internal `vfscore` API testing. Anything that's not part of a libc/POSIX-like interface is a good candidate.

toast3rx:
@razvand  I've amended the PR and now it contains some tests for `unlink()`.  I plan on adding some more tests for `vfscore` in the future. Do i need to open PRs for each function or do I need to keep this PR as a draft?

razvand:
> @razvand I've amended the PR and now it contains some tests for `unlink()`. I plan on adding some more tests for `vfscore` in the future. Do i need to open PRs for each function or do I need to keep this PR as a draft?

I think `unlink` is also part the libc testing.

Use **this** PR (only one) and add commits for each (internal) separate function you will test.

mariapana:
Hi! Taking a quick look at the code, things look good. The only thing that comes to mind at first glance is actually related to the PR itself. It mentions `dup()` testing, but the contents of your contribution are actually about `unlink()`. Are the PR title and description outdated or is there something else I might be missing? Thanks!

PR #798, opened 2023-03-11
lib/posix-event: Update `uio_offset` and `uio_resid` on successful reads

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [all]

### Additional configuration

 - `CONFIG_LIBPOSIX_EVENT=y`

### Description of changes

This PR updates (on a successful read) the `uio_offset` and `uio_resid` members of the `uio` structure passed to the `eventfd_vfscore_read` function. Without updating those members, the `read` function from the vfscore layer will always report that no bytes have been read at all.

The `eventfd_vfscore_write` function doesn't have the same bug, as it already updates its `uio` members after successful writes.

### DIFF
--- a/lib/posix-event/eventfd.c
+++ b/lib/posix-event/eventfd.c
@@ -171,6 +171,9 @@ static int eventfd_vfscore_read(struct vnode *vnode,
 
 	uk_mutex_unlock(&efd->lock);
 
+	buf->uio_resid = 0;
+	buf->uio_offset = sizeof(uint64_t);
+
 	return 0;
 }
 


craciunoiuc (approved):
Tested it and it works, thanks!

DeliaPavel (approved):

razvand:
Hi, @eduardvintila , could you point out a use case / simple example of this that causes an issue without your fix?

Also, it would be great if you could translate that fix into [a unit test](https://unikraft.org/docs/develop/writing-tests/) for `posix-event`.

eduardvintila:
> Hi, @eduardvintila , could you point out a use case / simple example of this that causes an issue without your fix?

Hey, @razvand. Here is a simple example which showcases the problem (it's just a simple app which only needs `posix-event`):

```C
#include <stdio.h>
#include <unistd.h>
#include <sys/eventfd.h>

int main(void) {
    int ret, initVal = 0;
    uint64_t w, r;
    int fd = eventfd(initVal, 0);

    if (fd < 0) {
      printf("Open eventfd failed\n");
      return -1;
    }

    w = 0xDEADBEEF;
    ret = write(fd, &w, sizeof(w));
    if (ret != sizeof(w)) {
        printf("unexpected write\n");
        close(fd);
        return -1;
    }

    ret = read(fd, &r, sizeof(r));
    if (ret != sizeof(r) || r != w) { /* ret should be 8, r should be equal to w */
        printf("unexpected read. ret = %d, r = %lx\n", ret, r);
        close(fd);
        return -1;
    }


    printf("All good!\n");
    close(fd);
    return 0;
}
```

Without the changes from the PR, this is the current output:
```console
[    0.137209] Info: [libukboot] <boot.c @  369> Calling main(2, ['build/app-helloworld-eventfd_kvm-x86_64', 'console=ttyS0'])
unexpected read. ret = 0, r = deadbeef
[    0.142455] Info: [libukboot] <boot.c @  378> main returned -1, halting system
[    0.145215] Info: [libkvmplat] <shutdown.c @   35> Unikraft halted
```

As you can see, one would expect the `read()` function to actually return the number of bytes that were read. But it just returns 0.

With the fixes from the PR, we get this output:

```console
[    0.137487] Info: [libukboot] <boot.c @  369> Calling main(2, ['build/app-helloworld-eventfd_kvm-x86_64', 'console=ttyS0'])
All good!
[    0.142540] Info: [libukboot] <boot.c @  378> main returned 0, halting system
[    0.145570] Info: [libkvmplat] <shutdown.c @   35> Unikraft halted
```

Which is as expected.

> Also, it would be great if you could translate that fix into [a unit test](https://unikraft.org/docs/develop/writing-tests/) for `posix-event`.

Sure, I can try to write some tests. But I'm not sure if I should make a new PR for this or just add a new commit to this one

razvand:
> Sure, I can try to write some tests. But I'm not sure if I should make a new PR for this or just add a new commit to this one.

Create a new PR that adds a [unit test](https://unikraft.org/docs/develop/writing-tests/) to `posix-event`.

razvand (approved):
Thanks, @eduardvintila. This works.

eduardvintila:
Just rebased on top of staging.

PR #801, opened 2023-03-13
lib/ukalloc: Synchronize the bbuddy allocator

razvanvirtan:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`, `arm64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Additional configuration

 - `LIBUKLOCK=y`

### Description of changes

The bbuddy allocator provided by Unikraft is not SMP safe. Synchronize the allocator with a lock-based approach, with a minimum impact on the allocator performance.

We associate a spinlock with each level of the allocator, where level k is the list of blocks of size 2^k pages. Therefore, two different cores will block only when trying to allocate / free / check blocks of the same size.

The only global_lock protects the list of memory areas, in case of concurrent calls to `bbuddy_addmem()`.

### Testing

For quick testing, you can use [these test files](https://github.com/unikraft/unikraft/files/14028770/bbuddy_smp_tests.zip)

Running example: `qemu-system-x86_64 --cpu host --enable-kvm -smp 9 --nographic --kernel build/app-helloworld_qemu-x86_64 -m 8G`

**Expected behavior:** without this PR, the application may crash on multi-core systems because of overlapping `malloc()` and `free()` calls. If not crashing, `malloc()` may result in overlapping allocated regions.

### DIFF
--- a/lib/ukallocbbuddy/bbuddy.c
+++ b/lib/ukallocbbuddy/bbuddy.c
@@ -51,6 +51,11 @@
 #include <uk/assert.h>
 #include <uk/page.h>
 
+#ifdef CONFIG_HAVE_SMP
+#include <uk/spinlock.h>
+#endif
+
+
 typedef struct chunk_head_st chunk_head_t;
 typedef struct chunk_tail_st chunk_tail_t;
 
@@ -79,11 +84,23 @@ struct uk_bbpalloc_memr {
 	unsigned long *mm_alloc_bitmap;
 };
 
+/*
+ * We associate a spinlock with each level of the allocator, where level k is
+ * the list of blocks of size 2^k pages. Therefore, two different cores will
+ * block only when trying to allocate / free / check blocks of the same size.
+
+ * The only global_lock protects the list of memory areas, in case of concurrent
+ * calls to `bbuddy_addmem()`.
+ */
 struct uk_bbpalloc {
 	unsigned long nr_free_pages;
 	chunk_head_t *free_head[FREELIST_SIZE];
 	chunk_head_t free_tail[FREELIST_SIZE];
 	struct uk_bbpalloc_memr *memr_head;
+#ifdef CONFIG_HAVE_SMP
+	uk_spinlock level_locks[FREELIST_SIZE];
+	uk_spinlock global_lock;
+#endif
 };
 
 #if CONFIG_LIBUKALLOCBBUDDY_FREELIST_SANITY
@@ -224,16 +241,20 @@ static void map_alloc(struct uk_bbpalloc *b, uintptr_t first_page,
 	end_off = end_page_idx & (PAGES_PER_MAPWORD - 1);
 
 	if (curr_idx == end_idx) {
-		memr->mm_alloc_bitmap[curr_idx] |=
-		    ((1UL << end_off) - 1) & -(1UL << start_off);
+		uk_or(&memr->mm_alloc_bitmap[curr_idx],
+			((1UL << end_off) - 1) & -(1UL << start_off));
 	} else {
-		memr->mm_alloc_bitmap[curr_idx] |= -(1UL << start_off);
+		uk_or(&memr->mm_alloc_bitmap[curr_idx], -(1UL << start_off));
 		while (++curr_idx < end_idx)
 			memr->mm_alloc_bitmap[curr_idx] = ~0UL;
-		memr->mm_alloc_bitmap[curr_idx] |= (1UL << end_off) - 1;
+		uk_or(&memr->mm_alloc_bitmap[curr_idx], (1UL << end_off) - 1);
 	}
 
+#ifdef CONFIG_HAVE_SMP
+	uk_fetch_sub(&b->nr_free_pages, nr_pages);
+#else
 	b->nr_free_pages -= nr_pages;
+#endif
 }
 
 static void map_free(struct uk_bbpalloc *b, uintptr_t first_page,
@@ -262,16 +283,20 @@ static void map_free(struct uk_bbpalloc *b, uintptr_t first_page,
 	end_off = end_page_idx & (PAGES_PER_MAPWORD - 1);
 
 	if (curr_idx == end_idx) {
-		memr->mm_alloc_bitmap[curr_idx] &=
-		    -(1UL << end_off) | ((1UL << start_off) - 1);
+		uk_and(&memr->mm_alloc_bitmap[curr_idx],
+			-(1UL << end_off) | ((1UL << start_off) - 1));
 	} else {
-		memr->mm_alloc_bitmap[curr_idx] &= (1UL << start_off) - 1;
+		uk_and(&memr->mm_alloc_bitmap[curr_idx], (1UL << start_off) - 1);
 		while (++curr_idx != end_idx)
 			memr->mm_alloc_bitmap[curr_idx] = 0;
-		memr->mm_alloc_bitmap[curr_idx] &= -(1UL << end_off);
+		uk_and(&memr->mm_alloc_bitmap[curr_idx], -(1UL << end_off));
 	}
 
+#ifdef CONFIG_HAVE_SMP
+	uk_fetch_add(&b->nr_free_pages, nr_pages);
+#else
 	b->nr_free_pages += nr_pages;
+#endif
 }
 
 /* return log of the next power of two of passed number */
@@ -310,8 +335,14 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
 
 	/* Find smallest order which can satisfy the request. */
 	for (i = order; i < FREELIST_SIZE; i++) {
+#ifdef CONFIG_HAVE_SMP
+		uk_spin_lock(&b->level_locks[i]);
+#endif
 		if (!FREELIST_EMPTY(b->free_head[i]))
 			break;
+#ifdef CONFIG_HAVE_SMP
+		uk_spin_unlock(&b->level_locks[i]);
+#endif
 	}
 	if (i >= FREELIST_SIZE)
 		goto no_memory;
@@ -320,6 +351,15 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
 	alloc_ch = b->free_head[i];
 	b->free_head[i] = alloc_ch->next;
 	alloc_ch->next->pprev = alloc_ch->pprev;
+
+	map_alloc(b, (uintptr_t)alloc_ch, 1UL << order);
+
+	/* A free block was found on level i in the for above, so this level
+	 * was not unlocked until the list has been modified.
+	 */
+#ifdef CONFIG_HAVE_SMP
+	uk_spin_unlock(&b->level_locks[i]);
+#endif
 
 	/* We may have to break the chunk a number of times. */
 	while (i != order) {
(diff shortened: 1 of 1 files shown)


lib/ukallocbbuddy/Config.uk
```diff
@@ -1,6 +1,16 @@
...
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
 	select LIBUKDEBUG
 	select LIBUKALLOC
+
+if LIBUKALLOCBBUDDY
+	config LIBUKALLOCBBUDDY_SMP_SAFE
+	bool "SMP safe allocator"
+	default n
+	depends on ARCH_ARM_64
```
razvanvirtan, line 12: Should remove / change when we will have proper spinlock support on more architectures.

lib/ukallocbbuddy/Config.uk
```diff
@@ -1,6 +1,15 @@
...
+menuconfig LIBUKALLOCBBUDDY
 	bool "ukallocbbuddy: Binary buddy page allocator"
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
 	select LIBUKDEBUG
 	select LIBUKALLOC
+
+if LIBUKALLOCBBUDDY
+	config LIBUKALLOCBBUDDY_SMP_SAFE
+	bool "SMP safe allocator"
```
Sairajkodilkar: Hi @razvanvirtan, You do not need to define a separate flag to enable SMP safe bbuddy allocator. This is because there exists a global flag `CONFIG_HAVE_SMP` .
The spinlock library internally uses this flag. If the CONFIG_HAVE_SMP is not set then it will map the functions to the dummy spinlock functions. Hence you do not need to handle SMP case

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -81,6 +86,10 @@ struct uk_bbpalloc {
 	chunk_head_t *free_head[FREELIST_SIZE];
 	chunk_head_t free_tail[FREELIST_SIZE];
 	struct uk_bbpalloc_memr *memr_head;
+#ifdef CONFIG_LIBUKALLOCBBUDDY_SMP_SAFE
```
Sairajkodilkar: Remove the config flag check; this is because on the uniprocessor system, the uk_spinlock is mapped to an empty structure hence required memory is 0 bytes.
```suggestion
```

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -81,6 +86,10 @@ struct uk_bbpalloc {
...
+	struct uk_spinlock level_locks[FREELIST_SIZE];
+	struct uk_spinlock global_lock;
+#endif
```
Sairajkodilkar, line 103:
```suggestion
```

Sairajkodilkar (changes requested):

Sairajkodilkar:
Hi @razvanvirtan, I went through your implementation and noticed that you are not handling the race condition in function `bbuddy_addmem()`.  The user can use this function to add the memory regions after initialization. So, you need to protect lines 506, 507, and 539, which involve the modification of struct uk_alloc.
Along with this, you need to protect line 426 in function `bbuddy_pmaxalloc()` with a spinlock

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -173,8 +182,14 @@ static void map_alloc(struct uk_bbpalloc *b, uintptr_t first_page,
 			memr->mm_alloc_bitmap[curr_idx] = ~0UL;
 		memr->mm_alloc_bitmap[curr_idx] |= (1UL << end_off) - 1;
 	}
-
+	
+#ifdef CONFIG_LIBUKALLOCBBUDDY_SMP_SAFE
+	uk_spin_lock(&b->global_lock);
 	b->nr_free_pages -= nr_pages;
+	uk_spin_unlock(&b->global_lock);
```
Sairajkodilkar: Instead of the spinlock, you can use the gnu builtins `__atomic_sub_fetch()` and `__atomic_load_n()` to write and read `b->nr_free_pages` atomically. The `uk/arch/atomic.h` provides a wrapper on top of these atomic builtins. currently, unikraft does not have a wrapper for `__atomic_sub_fetch()` and `__atomic_add_fetch()`, Hence you need to add them.

see [gnu atomics](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) and [atomic.h](https://github.com/unikraft/unikraft/blob/staging/include/uk/arch/atomic.h) for more details

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -212,7 +227,13 @@ static void map_free(struct uk_bbpalloc *b, uintptr_t first_page,
 		memr->mm_alloc_bitmap[curr_idx] &= -(1UL << end_off);
 	}
 
+#ifdef CONFIG_LIBUKALLOCBBUDDY_SMP_SAFE
+	uk_spin_lock(&b->global_lock);
+	b->nr_free_pages += nr_pages;
```
Sairajkodilkar: use `__atomic_add_fetch()`

Sairajkodilkar (changes requested):

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -173,8 +182,14 @@ static void map_alloc(struct uk_bbpalloc *b, uintptr_t first_page,
 			memr->mm_alloc_bitmap[curr_idx] = ~0UL;
 		memr->mm_alloc_bitmap[curr_idx] |= (1UL << end_off) - 1;
 	}
-
+	
+#ifdef CONFIG_LIBUKALLOCBBUDDY_SMP_SAFE
+	uk_spin_lock(&b->global_lock);
```
Sairajkodilkar: you should disable the interrupts before acquiring the lock and reenable them after releasing the lock.
```
irqf = ukplat_lcpu_save_irqf();
uk_spin_lock();
ukplat_lcpu_restore_irqf(irqf);
```
`

Sairajkodilkar (changes requested):

lib/ukallocbbuddy/bbuddy.c
razvanvirtan: Do we expect to have buddy allocator calls from interrupt handlers?

razvanvirtan:
Hi @Sairajkodilkar, thanks for the thorough review! Here are a few points regarding this round of updates:

- I created [an additional PR ](https://github.com/unikraft/unikraft/pull/864 )for introducing a wrapper on the `__atomic_fetch_sub` builtin. I will drop the associated commit, once the PR is merged.

- I removed the custom config option, but still kept some guards in the places where we are using atomic builtins, in order to avoid extra assembly instructions on single-core systems. (For instance, on arm64 one `LDAXR` and one `STLXR` guard the simpe `sub` / `add`.)

- Regarding the `bbuddy_pmaxalloc()`, I am not sure we can guarantee SMP safe results. Even if we use a `spinlock` to protect the entire function until return, we still need to release the lock before returning and I think we will still have a small possibility of race conditions. (Also, we can't make sure that the allocator situation won't change between the `bbuddy_pmaxalloc()` call and the moment when the user will actually use its results, but we may consider it is the caller's duty to take this into account). What do you think, maybe we should document this function as not being safe in multi-core environments? (I don't have a strong opinion on this)

lib/ukallocbbuddy/Config.uk
```diff
@@ -1,6 +1,7 @@
-config LIBUKALLOCBBUDDY
+menuconfig LIBUKALLOCBBUDDY
```
andraprs: For the commit message title could use the "lib/ukallocbbuddy" prefix instead of "lib/ukalloc", as there are no changes e.g. in files from https://github.com/unikraft/unikraft/tree/staging/lib/ukalloc and just to be a bit more specific.

Thank you for the synchronization updates from this PR. :)

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -81,6 +84,8 @@ struct uk_bbpalloc {
 	chunk_head_t *free_head[FREELIST_SIZE];
 	chunk_head_t free_tail[FREELIST_SIZE];
 	struct uk_bbpalloc_memr *memr_head;
+	uk_spinlock level_locks[FREELIST_SIZE];
```
andraprs, line 101: Could add a comment for the "uk_bbpalloc" struct and can include a brief description for the level / global lock(s) usage (for example why they are needed) e.g. can be parts from the PR "Description of changes" section.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -249,6 +262,11 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
 
 	/* Find smallest order which can satisfy the request. */
 	for (i = order; i < FREELIST_SIZE; i++) {
+		uk_spin_lock(&b->level_locks[i]);
+		if (!FREELIST_EMPTY(b->free_head[i]))
+			break;
+		uk_spin_unlock(&b->level_locks[i]);
+
```
andraprs: There is a duplicate codebase after the uk_spin_unlock() call.

```
		if (!FREELIST_EMPTY(b->free_head[i]))
			break;
```

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -260,6 +278,10 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
 	b->free_head[i] = alloc_ch->next;
 	alloc_ch->next->pprev = alloc_ch->pprev;
 
+	map_alloc(b, (uintptr_t)alloc_ch, 1UL << order);
+
+	uk_spin_unlock(&b->level_locks[i]);
```
andraprs, line 361: Could mention in a comment where the corresponding "uk_spin_lock()" for this "uk_spin_unlock()" comes from e.g. the for loop above. Or can unlock earlier (e.g. before "break") and add another set of lock() / unlock() for this codebase block, to keep lock / unlock pairs.

The "map_alloc()" call was moved earlier in the codebase flow to be part of this synchronized code block? Does this reorder have any effects given the follow-up while loop?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -355,6 +386,8 @@ static void bbuddy_pfree(struct uk_alloc *a, void *obj, unsigned long num_pages)
 
 	freed_ch->next->pprev = &freed_ch->next;
 	b->free_head[order] = freed_ch;
+
+	uk_spin_unlock(&b->level_locks[order]);
```
andraprs: The comment above (the corresponding "uk_spin_lock()" for this "uk_spin_unlock()" ) applies for this codebase logic as well.

andraprs (commented):
Added a couple of comments for the bbuddy allocator updates.

Thank you.

lib/ukallocbbuddy/bbuddy.c
razvanvirtan: I don't think unlocking before the break and then locking again would be safe. In this time window, another thread could enter the `if (!FREELIST_EMPTY(b->free_head[i]))` statement and find the same free block, resulting in a race condition.

Yes, the `map_alloc()` call was moved so that it is part of the synchronized  code block. Otherwise, if the free block that has been selected for this allocation resulted from a previous split, there will be a chance that it will be merged by a `free()` running on a parallel thread, even if we are in the process of allocating it (see [here](https://github.com/unikraft/unikraft/pull/801/files#diff-7a854c8df048714d0efcab86bedab15bf77f36afe24270e946fd1bdb578c14afR355)).

The reorder shouldn't have any effects. The loop only works on lower orders of the allocator, while the `map_alloc` call makes sure that this block won't be merged with its buddy and added to a higher order list.

razvanvirtan:
@andraprs Updated according to your comments, thank you!

lib/ukallocbbuddy/bbuddy.c
andraprs: Thank you, Razvan, for the explanation and the updates.

andraprs (dismissed):
Thank you for the updates.

Should rebase to the latest staging branch, there are a couple of conflicts for the "README.md" and "version.mk" files mentioned in the PR and an additional commit that shouldn't be part of this PR.

andreistan26 (approved):
All good, works well. Thanks!

PR #804, opened 2023-03-16
plat/pci_ecam: Fix FDT interrupt node lacking address-cells

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [ N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The `gen_pci_irq_parse` function fails when an interrupt parent node is missing the `#address-cells` property. Even though the Devicetree spec states that this property shall be explicitly defined, some DTBs (such as the RISC-V DTB generated by QEMU <= 7.1) do not include this property for the interrupt controller. However, when the parent unit address is missing from the DTB, we can assume that the parent node's `#address-size` property is equal to 0 to maintain compatibility with such DTBs. The specification even says that "If a unit address component is not required, `#address-cells` shall be explicitly defined to be zero.".

We can thus introduce a new CONFIG option, `DTB_INTERRUPT_ADDRESS_CELLS_ERRATA`, which should provide an errata for uncomplying device trees when parsing the interrupt nodes.

Since this particular issue doesn't manifest on the ARM `virt` machine from QEMU, one can possibly test this with other platforms, or with the RISC-V PR: https://github.com/unikraft/unikraft/pull/461.

### DIFF
--- a/drivers/ukbus/pci/pci_ecam.c
+++ b/drivers/ukbus/pci/pci_ecam.c
@@ -350,10 +350,16 @@ int gen_pci_irq_parse(const fdt32_t *addr, struct fdt_phandle_args *out_irq)
 			prop = fdt_getprop(dtb, newpar, "#address-cells",
 					   &plen);
 			if (prop == NULL) {
-				uk_pr_info(" -> parent lacks #address-cells!\n");
+#ifdef CONFIG_DTB_INTERRUPT_ADDRESS_CELLS_ERRATA
+				uk_pr_info(" -> parent lacks #address-cells, assume 0 cells\n");
+				newaddrsize = 0;
+#else
+				uk_pr_info(" -> parent lacks #address-cells! abort parsing...\n");
 				goto fail;
+#endif
+			} else {
+				newaddrsize = fdt32_to_cpu(prop[0]);
 			}
-			newaddrsize = fdt32_to_cpu(prop[0]);
 
 			uk_pr_debug(" -> newintsize=%d, newaddrsize=%d\n",
 			    newintsize, newaddrsize);
--- a/plat/Config.uk
+++ b/plat/Config.uk
@@ -115,3 +115,12 @@ config FPSIMD
 	depends on ARCH_ARM_64
 	help
 		Enable support FPU usage in application
+
+config DTB_INTERRUPT_ADDRESS_CELLS_ERRATA
+	bool "Errata for uncomplying device trees"
+	default n
+	depends on ARCH_ARM_64
+	help
+		Errata for parsing certain uncomplying device trees
+		which lack the #address-cells property for interrupt
+		controllers.


kubanrob:
> However, when the parent unit address is missing from the DTB, we can assume that the parent node's #address-size property is equal to 0, as per Devicetree specification v0.3

From my reading, this is exactly the opposite of what the specification states.

> If a unit address component is not required, #address-cells shall be explicitly defined to be zero.

Explicitly means it needs to be in the device tree, where as this patch implicitly assumes it is zero.

From the same specification:
> If missing, a client program should assume a default value of 2 for #address-cells, and a value of 1 for
#size-cells.

**So I think assuming zero if its missing goes against the specification.**

Anyway, it seem to be a good solution for this particular problem and is not in generic DT parsing code, maybe this could be enabled by a config option to relax the strictness of these checks?

eduardvintila:
Hey, @kubanrob! Indeed, this patch would not comply per-se with the specification, and I may have worded the commit description a bit weird, sorry for that.

> If missing, a client program should assume a default value of 2 for #address-cells, and a value of 1 for
#size-cells.


Yes, I was aware of this. Unfortunately, if one would assume that, the compatibility with the aforementioned RISC-V DTB would break and the parsing of the interrupt controller's properties will fail. Also, the following is stated in the specification, which in my opinion makes it clear that this old RISC-V DTB violates the specification:

> Both the child node and the interrupt parent node are required to have #address-cells and #interrupt-
cells properties defined.

Fortunately, starting with QEMU 7.2, the RISC-V DTB has been fixed and now has the properties explicitly defined, just like the specification states it:

```
		plic@c000000 {
			phandle = <0x03>;
			riscv,ndev = <0x60>;
			reg = <0x00 0xc000000 0x00 0x600000>;
			interrupts-extended = <0x02 0x0b 0x02 0x09>;
			interrupt-controller;
			compatible = "sifive,plic-1.0.0\0riscv,plic0";
			**#address-cells = <0x00>;** 
			**#interrupt-cells = <0x01>;** 
		};
```

I fully agree with introducing a CONFIG option for this, since it's essentially an ERRATA for the uncomplying RISC-V DTB found in older versions of QEMU. I will also update the description to clearly illustrate this.

Thanks a lot for your input! : )

razvanvirtan:
Hi, @eduardvintila!
Thank you for this! Can you please add the config option that was discussed above, so that we will merge this for 0.13?

eduardvintila:
Hey, @razvanvirtan @kubanrob.  I've introduced a new CONFIG option in `Platform Configuration` which can trigger this errata explicitly only when selected. Please let me know if the changes are okay, when you have the time. Many thanks!

razvanvirtan:
All good on my side, thank you @eduardvintila!

razvanvirtan (dismissed):
All good on my side, thank you @eduardvintila!

michpappas:
Hi @eduardvintila. I like this change, especially when provided as an erratum for non-complying implementations. I am only concerned, given that this practice was so widespread in the past, whether the fix should be specific to the interrupt controller or be part of generic parsing helpers such as `fdt_get_address()`, `fdt_interrupt_cells()` and so on.

eduardvintila:
Hey, @michpappas. I think this issue is too particular to be embedded into the generic FDT parsers; I'm not sure how we can do this, since only in this instance (with regards to interrupt controllers), we assume that `#address-cells` is equal to 0. Otherwise, for generic nodes, `#address-cells` is already [assumed to be 2](https://github.com/unikraft/unikraft/blob/staging/lib/fdt/fdt_addresses.c#L42) when the property is missing, as per the Devicetree spec:

>"If missing, a client program should assume a default value of 2 for #address-cells, and a value of 1 for
#size-cells."

eduardvintila:
@michpappas  Or... we could do something like this: modify `pci_ecam.c` code to no longer use the `fdt_getprop` function to retrieve the `#address-cells` property when parsing the interrupt tree, but rather `fdt_get_address()`. We then add a special check inside this function for nodes marked as `interrupt-controller` to return 0 as the number of `address-cells` when the property is missing, **if** the `CONFIG_ERRATA` option is selected. If this option is not selected, we return a negative value, marking that the devicetree is uncomplying.

michpappas:
@eduardvintila yes you're right sorry 👍🏼 I'm just a bit concerned that this is only inside the PCI-ECAM driver, and later the same issue will appear in other drivers. What are your thoughts towards moving that logic into a new helper in `plat/drivers/ofw` and applying that change there?

eduardvintila:
> I'm just a bit concerned that this is only inside the PCI-ECAM driver, and later the same issue will appear in other drivers. 


@michpappas, oh, you're right! Yes, implementing this kind of interrupt parsing from the FDT as part of the `ofw` library makes total sense to me. I can work on migrating the existing logic to `ofw` as part of a helper function 👍🏻

kubanrob (commented):
> I can work on migrating the existing logic to ofw as part of a helper function 👍🏻

That seems sensible. I am holding off with approving until this is either done or it's decided that will be another PR.

Otherwise looks good to me.

michpappas:
Hey @eduardvintila 👋🏼 Is there any chance we can revive this for 0.17?

eduardvintila:
> Hey @eduardvintila 👋🏼 Is there any chance we can revive this for 0.17?

Hey, @michpappas! Sorry, was very busy IRL, too late to get it for 0.17 now. Will leave for 0.18 🙈

PR #805, opened 2023-03-16
plat/drivers/ofw: Fix issues in FDT functions

eduardvintila:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR fixes issues found in three functions: `fdt_translate_one` and
`fdt_node_offset_idx_by_compatible_list` / `fdt_node_offset_by_compatible_list` :

`fdt_translate_one`:
When a node is missing the "ranges" property, offset should be set to 0. The current implementation sets offset to the address to be translated, which ignores the fact that the `fdt_default_translate function` is later called. The latter function adds the offset to the same address, resulting in doubling the effective address, when in fact a 1-to-1 mapping is necessary.

`fdt_node_offset_idx_by_compatible_list` / `fdt_node_offset_by_compatible_list`:
The current implementation returns as soon as a node that matches a compatible string is found. However, we should iterate through all of the compatible list to make sure that we return the node with the least offset value such that future searches will not skip any matching nodes.

As with https://github.com/unikraft/unikraft/pull/804, the issues happen to not arise on QEMU's ARM `virt` machine. x86 targets do not use this driver at all, so no luck there either. As such, one can maybe test this PR on other ARM platforms, or with the RISC-V PR (https://github.com/unikraft/unikraft/pull/461), which is directly affected by this problem.

### DIFF
--- a/plat/drivers/ofw/fdt.c
+++ b/plat/drivers/ofw/fdt.c
@@ -300,14 +300,17 @@ int fdt_get_address(const void *fdt, int nodeoffset, uint32_t index,
 int fdt_node_offset_by_compatible_list(const void *fdt, int startoffset,
 				  const char * const compatibles[])
 {
-	int idx, offset;
+	int idx, min_offset = INT_MAX, offset;
 
 	for (idx = 0; compatibles[idx] != NULL; idx++) {
 		offset = fdt_node_offset_by_compatible(fdt, startoffset,
 				  compatibles[idx]);
-		if (offset >= 0)
-			return offset;
+		if (offset >= 0 && offset < min_offset)
+			min_offset = offset;
 	}
+
+	if (min_offset != INT_MAX)
+		return min_offset;
 
 	return -FDT_ERR_NOTFOUND;
 }
@@ -315,18 +318,22 @@ int fdt_node_offset_by_compatible_list(const void *fdt, int startoffset,
 int fdt_node_offset_idx_by_compatible_list(const void *fdt, int startoffset,
 			  const char * const compatibles[], int *index)
 {
-	int idx, offset;
+	int idx, min_offset = INT_MAX, offset, compatible_idx;
 
 	for (idx = 0; compatibles[idx] != NULL; idx++) {
 		offset = fdt_node_offset_by_compatible(fdt, startoffset,
 				  compatibles[idx]);
-		if (offset >= 0) {
-			*index = idx;
-			return offset;
+		if (offset >= 0 && offset < min_offset) {
+			min_offset = offset;
+			compatible_idx = idx;
 		}
 	}
 
-	*index = idx;
+	if (min_offset != INT_MAX) {
+		*index = compatible_idx;
+		return min_offset;
+	}
+
 	return -FDT_ERR_NOTFOUND;
 }
 
--- a/plat/drivers/ofw/fdt.c
+++ b/plat/drivers/ofw/fdt.c
@@ -158,7 +158,7 @@ static int fdt_translate_one(const void *fdt, int parent, fdt32_t *addr,
 	if (!ranges)
 		return 1;
 	if (rlen == 0) {
-		offset = fdt_reg_read_number(addr, na);
+		offset = 0;
 		uk_pr_debug("empty ranges, 1:1 translation\n");
 		goto finish;
 	}
@@ -179,7 +179,7 @@ static int fdt_translate_one(const void *fdt, int parent, fdt32_t *addr,
 	memcpy(addr, ranges + na, 4 * pna);
 
 finish:
-	uk_pr_info("parent translation for:%p %x", addr, pna);
+	uk_pr_info("parent translation for:%p %x\n", addr, pna);
 
 	/* Translate it into parent bus space */
 	return fdt_default_translate(addr, offset, pna);


plat/drivers/ofw/fdt.c
```diff
@@ -315,18 +315,22 @@ int fdt_node_offset_by_compatible_list(const void *fdt, int startoffset,
 int fdt_node_offset_idx_by_compatible_list(const void *fdt, int startoffset,
 			  const char * const compatibles[], int *index)
 {
-	int idx, offset;
+	int idx, min_offset = INT_MAX, compatible_idx;
 
 	for (idx = 0; compatibles[idx] != NULL; idx++) {
-		offset = fdt_node_offset_by_compatible(fdt, startoffset,
+		int offset = fdt_node_offset_by_compatible(fdt, startoffset,
```
razvanvirtan: Maybe we could declare this at the beginning of the function, like the other local variables?

razvanvirtan (commented):
Hi @eduardvintila!
This looks good, thanks for the nice catches!
I have only left an inline comment.

plat/drivers/ofw/fdt.c
eduardvintila: Good point, thanks!

razvanvirtan (approved):
Thank you, @eduardvintila!

michpappas:
Hi again @eduardvintila 

> fdt_translate_one:
> When a node is missing the "ranges" property, offset should be set to 0. The current implementation sets offset to the address to be translated, which ignores the fact that the fdt_default_translate function is later called. The latter function adds the offset to the same address, resulting in doubling the effective address, when in fact a 1-to-1 mapping is necessary.

Looks good 👍🏼

> fdt_node_offset_idx_by_compatible_list:
> The current implementation returns as soon as a node that matches a compatible string is found. However, we should iterate through all of the compatible list to make sure that we return the node with the least offset value such that future searches will not skip any matching nodes.

Could you please elaborate a bit on the requirement to return the nearest node? My expectation is that a driver would eventually care for all compatible nodes.

eduardvintila:
Hey, @michpappas!


> > fdt_node_offset_idx_by_compatible_list:
> > The current implementation returns as soon as a node that matches a compatible string is found. However, we should iterate through all of the compatible list to make sure that we return the node with the least offset value such that future searches will not skip any matching nodes.
> 
> Could you please elaborate a bit on the requirement to return the nearest node? My expectation is that a driver would eventually care for all compatible nodes.

Sure! This requirement is even specified in the documentation of the function, in [fdt.h](https://github.com/unikraft/unikraft/blob/staging/plat/drivers/include/ofw/fdt.h#L171):

```
fdt_node_offset_idx_by_compatible_list() returns the offset of the
 **first matched** node after startoffset, which has a 'compatible'
 property which lists the given compatible string; or if
 startoffset is -1, the very first such node in the tree.
```

In the current implementation, the offset returned by the function in the FDT  is actually influenced by the order of the elements in the `compatibles` list. For example, I stumbled upon this problem when working on the RISC-V port. The RISC-V FDT for the `virt` QEMU platform is laid out differently than the ARM one, so when searching nodes that match this compatible list from [platform_bus.c](https://github.com/unikraft/unikraft/blob/staging/plat/common/platform_bus.c#L145):

```
static const char *pf_device_compatible_list[] = {
	"virtio,mmio",
	"pci-host-ecam-generic",
	NULL
};
```

...  the node with the `virtio,mmio` property is found and returned, but it just happens that its offset in the tree is bigger than the `pci-host-ecam-generic` node. This means that a further call to the function will miss this node, since the search is starting from the offset of the last node that was found, which essentially results in no longer initializing the PCI host controller in our case.

Also... I have just noticed that I've missed the `fdt_node_offset_by_compatible_list()` function in my PR. It seems to do the same job, it's only missing the `idx` parameter in which it places the index of the compatible string that was found in the node. If the changes are okay, I'll also update that function, as well.

Thanks!

michpappas:
> This means that a further call to the function will miss this node, since the search is starting from the offset of the last node that was found, which essentially results in no longer initializing the PCI host controller in our case.

@eduardvintila aha, that was the part I was missing. Thanks for the explanation. 

Btw, I find the mere existence this helper function a bit questionable, as the driver could do a two-level iteration over all occurrences of each compatible node, and that would had avoided the extra logic. But since we have it, it definitely needs to behave correctly.

> Also... I have just noticed that I've missed the fdt_node_offset_by_compatible_list() function in my PR. It seems to do the same job, it's only missing the idx parameter in which it places the index of the compatible string that was found in the node. If the changes are okay, I'll also update that function, as well.

Okay 👍🏼

eduardvintila:
@michpappas, I've updated the PR to also include the changes in the related function, `fdt_node_offset_by_compatible_list` 👍🏻

michpappas:
@eduardvintila thanks a lot. Unless @kubanrob has any comments, I'm good to merge.

kubanrob:
Looks good to merge to me too. Did not test, but the explanations make sense and to not see any problems from my side.

kubanrob (approved):

michpappas:

Thanks all

michpappas (approved):

Issue #447, opened 2022-04-13
On pointer comparison


RekhaSigma:
### Describe the bug

Referring to the following pointer comparison in L113 in memmove_isr():

https://github.com/unikraft/unikraft/blob/6a1e028f8c498ae7d83cac7a12844a916a3ca48b/lib/isrlib/string.c#L108-L116

the C Standard says the behaviour of pointer comparison is undefined when the objects pointed to are different objects.
In that case, wouldn't it be better to fix the semantics by type casting, like
```
if ((intptr_t) src > (intptr_t) dst)
...
```

vladandrew:
Hi,

This is correct, although I think this not a problem under GCC or clang since it does the (intptr_t) casting behind the hoods.

RekhaSigma:
Thanks for the info, Vlad.

marcrittinghaus:
Right, but does not hurt to fix. Thanks @RekhaSigma !

@vladandrew
If you create a PR for this, if you like, you may also take a look at the other candidate functions in `nolibc` and the `isr` lib.

vladandrew:
Will do!

srujanm111:
Hello, I am representing a group of undergraduate UT Austin students aiming to contribute to open-source virtualization projects on behalf of a virtualization course. Could we have this issue assigned to us?

PR #825, opened 2023-04-02
lib/nolibc, lib/isrlib: Pointer comparison

ranfun:
### Prerequisite checklist

### Description of changes

The memmove_isr function is used to copy a block of memory from a source address to a destination address. The function is used in interrupt service routines (ISRs) where it is important to copy the memory as quickly as possible.

In the previous implementation of memmove_isr, the function assumed that the source address is either greater than the destination address or less than it. If both the conditions were false, the function would not work as expected and may cause undefined behavior. This assumption could lead to issues where the memory blocks overlap.

The modified implementation of memmove_isr takes into account both overlapping and non-overlapping cases. The function typecasts the src and dst pointers to intptr_t to ensure that the values being compared are treated as pointers rather than integers. This can help prevent errors that might occur if the pointer values are interpreted as signed integers.  ( Issue [#447](https://github.com/unikraft/unikraft/issues/447) ).

The changes were implemented in both nolibc and isrlib libraries to ensure that the code is consistent across all libraries and to prevent any issues when switching between them.

Overall, the changes ensure that memmove_isr works correctly in all cases, including when the memory blocks overlap, and that the code is robust and less prone to errors.

### DIFF
--- a/lib/isrlib/string.c
+++ b/lib/isrlib/string.c
@@ -107,21 +107,27 @@ void *memrchr_isr(const void *m, int c, size_t n)
 
 void *memmove_isr(void *dst, const void *src, size_t len)
 {
-	uint8_t *d = dst;
-	const uint8_t *s = src;
+	uint8_t *d = dst; 
+    const uint8_t *s = src; 
 
-	if (src > dst) {
-		for (; len > 0; --len)
-			*(d++) = *(s++);
-	} else {
-		s += len;
-		d += len;
+    if ((intptr_t)src == (intptr_t)dst) {
+        return dst;
+    }
 
-		for (; len > 0; --len)
-			*(d--) = *(s--);
-	}
+    if ((intptr_t)src > (intptr_t)dst) {
+        for (; len > 0; --len) {
+            *(d++) = *(s++);
+        }
+    } else {
+        s += len;
+        d += len;
 
-	return dst;
+        for (; len > 0; --len) {
+            *(d--) = *(s--);
+        }
+    }
+
+    return dst;
 }
 
 int memcmp_isr(const void *ptr1, const void *ptr2, size_t len)
--- a/lib/isrlib/string.c
+++ b/lib/isrlib/string.c
@@ -109,11 +109,10 @@ void *memmove_isr(void *dst, const void *src, size_t len)
 {
 	uint8_t *d = dst;
     const uint8_t *s = src;
+    const uint8_t *s = src;
 
-    if ((intptr_t)src == (intptr_t)dst) {
-        return dst;
-    }
-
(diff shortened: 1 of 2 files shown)


lib/isrlib/string.c
```diff
@@ -108,18 +108,22 @@ void *memrchr_isr(const void *m, int c, size_t n)
...
-		d += len;
-
-		for (; len > 0; --len)
-			*(d--) = *(s--);
-	}
+    const uint8_t *s = src;
+
+    if ((intptr_t)src == (intptr_t)dst) {
+        return dst;
+    }
```
mariasfiraiala:
```suggestion
    if ((intptr_t)src == (intptr_t)dst)
        return dst;
```

unnecessary braces

lib/nolibc/string.c
```diff
@@ -109,7 +109,11 @@ void *memmove(void *dst, const void *src, size_t len)
 	uint8_t *d = dst;
 	const uint8_t *s = src;
 
-	if (src > dst) {
+	if ((intptr_t)src == (intptr_t)dst) {
+        return dst;
+    }
```
mariasfiraiala: Suspect code indent and unnecessary braces

mariasfiraiala (changes requested):
Thanks @ranfun, the changes look ok. However, a few cosmetic aspects should be taken into consideration before merging this:

* Run `checkpatch.pl` on your commits using the instructions from [here](https://unikraft.org/docs/contributing/testing/#using-checkpatchpl) and fix your coding style errors. You'll find a couple of examples in my comments below.

* Squash your commits. The 4 of them can be combined into a single one, which suffices.

* Modify your commit message such as the verbs are all at Present Tense. Use the prefix `lib/{nolibc,isrlib}` before the description of the changes in your commit message. A good practice is to also have less than 50 characters for the commit message, so try to accomplish that.

eduardvintila (commented):
Hey, @ranfun. Thanks for the changes! Please follow @mariasfiraiala's suggestions so we can merge this PR : )

ranfun:
yes will do!

razvand:
@ranfun, can you take care of this, please?

razvand:
@ranfun, please see the checkpatch warnings

ranfun:
The latest patch passed the checkpatch warnings

mariasfiraiala:
> The latest patch passed the checkpatch warnings

This is exactly why you should drop the other 5 commits.  We shouldn't have upstream unnecessary patches, such as the first 5 commits that **do not** pass the coding style checker.

So, @ranfun, drop them.

mariasfiraiala:
Closed via #880.

PR #826, opened 2023-04-04
Dynamically allocate `struct mount` paths

dinhngtu:
### Prerequisite checklist

### Description of changes

Currently `m_path` and `m_special` are statically-allocated arrays, meaning their storage spaces are wasted in `static struct mount` variables.

Dynamically allocate these variables instead to save on binary size.

This PR saves ~8K of uncompressed binary size per `static struct mount` object.

### DIFF
--- a/lib/posix-event/epoll.c
+++ b/lib/posix-event/epoll.c
@@ -41,10 +41,13 @@
 #include <uk/alloc.h>
 #include <uk/syscall.h>
 #include <uk/config.h>
+#include <uk/init.h>
 
 #include <inttypes.h>
 #include <stdlib.h>
 #include <errno.h>
+#include <string.h>
+#include <stdlib.h>
 
 static uint64_t e_inode;
 
@@ -371,3 +374,28 @@ int epoll_pwait(int epfd, struct epoll_event *events, int maxevents,
 	return ret;
 }
 #endif /* UK_LIBC_SYSCALLS */
+
+static int epoll_mount_init(void)
+{
+	int ret;
+
+	epoll_mount.m_path = strdup("");
+	if (!epoll_mount.m_path) {
+		ret = -ENOMEM;
+		goto err_out;
+	}
+
+	epoll_mount.m_special = strdup("");
+	if (!epoll_mount.m_special) {
+		ret = -ENOMEM;
+		goto err_free_m_path;
+	}
+
+	return 0;
+
+err_free_m_path:
+	free(epoll_mount.m_path);
+err_out:
+	return ret;
+}
+uk_lib_initcall(epoll_mount_init);
--- a/lib/posix-event/eventfd.c
+++ b/lib/posix-event/eventfd.c
@@ -45,10 +45,13 @@
 #include <uk/mutex.h>
 #include <uk/list.h>
 #include <uk/config.h>
+#include <uk/init.h>
 
 #include <sys/eventfd.h>
 #include <inttypes.h>
 #include <errno.h>
+#include <string.h>
+#include <stdlib.h>
 
 struct eventfd {
 	/** Current value of this eventfd */
@@ -438,3 +441,28 @@ int eventfd(unsigned int initval, int flags)
 	return ret;
 }
 #endif /* UK_LIBC_SYSCALLS */
+
+static int eventfd_mount_init(void)
+{
+	int ret;
+
+	eventfd_mount.m_path = strdup("");
+	if (!eventfd_mount.m_path) {
+		ret = -ENOMEM;
+		goto err_out;
+	}
+
+	eventfd_mount.m_special = strdup("");
+	if (!eventfd_mount.m_special) {
+		ret = -ENOMEM;
+		goto err_free_m_path;
+	}
+
+	return 0;
+
+err_free_m_path:
+	free(eventfd_mount.m_path);
+err_out:
+	return ret;
+}
+uk_lib_initcall(eventfd_mount_init);
--- a/lib/posix-socket/socket_vnops.c
+++ b/lib/posix-socket/socket_vnops.c
@@ -43,8 +43,11 @@
 #include <vfscore/mount.h>
 #include <vfscore/fs.h>
 #include <uk/errptr.h>
+#include <uk/init.h>
 #include <inttypes.h>
 #include <errno.h>
+#include <string.h>
+#include <stdlib.h>
 
 static struct mount posix_socket_mount;
 
@@ -342,3 +345,28 @@ static struct vfsops posix_socket_vfsops = {
 static struct mount posix_socket_mount = {
 	.m_op = &posix_socket_vfsops
 };
+
+static int posix_socket_mount_init(void)
+{
+	int ret;
+
+	posix_socket_mount.m_path = strdup("");
+	if (!posix_socket_mount.m_path) {
+		ret = -ENOMEM;
+		goto err_out;
+	}
+
+	posix_socket_mount.m_special = strdup("");
+	if (!posix_socket_mount.m_special) {
+		ret = -ENOMEM;
+		goto err_free_m_path;
+	}
+
+	return 0;
+
+err_free_m_path:
+	free(posix_socket_mount.m_path);
+err_out:
+	return ret;
+}
+uk_lib_initcall(posix_socket_mount_init);
--- a/lib/vfscore/mount.c
+++ b/lib/vfscore/mount.c
@@ -161,7 +161,7 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	/*
 	 * Create VFS mount entry.
 	 */
-	mp = malloc(sizeof(struct mount));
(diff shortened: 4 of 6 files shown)


RaduNichita (approved):
I tested with `app-nginx`, `app-redis` and `app-sqlite` on x86 and everything was OK. 

Thanks @dinhngtu  for providing this PR.

Reviewed-by:  Radu Nichita <radunichita99@gmail.com>

lib/vfscore/mount.c
```diff
@@ -171,8 +171,8 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	mp->m_flags = flags;
 	mp->m_dev = device;
 	mp->m_data = NULL;
-	strlcpy(mp->m_path, dir, sizeof(mp->m_path));
-	strlcpy(mp->m_special, dev, sizeof(mp->m_special));
+	mp->m_path = strndup(dir, PATH_MAX - 1);
```
skuenzer, line 174: Can you also update the code to abort operation if the allocation failed (`ENOMEM`)?

skuenzer (changes requested):

lib/vfscore/mount.c
```diff
@@ -236,6 +244,8 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	if (dp_covered)
 		drele(dp_covered);
  err2:
+	free(mp->m_special);
+	free(mp->m_path);
 	free(mp);
  err1:
```
skuenzer, line 252: I'd actually rename this label to `err_close_device:`

lib/vfscore/mount.c
```diff
@@ -236,6 +244,8 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	if (dp_covered)
 		drele(dp_covered);
  err2:
+	free(mp->m_special);
+	free(mp->m_path);
 	free(mp);
```
skuenzer, line 251: `err_free_mp:`

lib/vfscore/mount.c
```diff
@@ -236,6 +244,8 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	if (dp_covered)
 		drele(dp_covered);
  err2:
+	free(mp->m_special);
+	free(mp->m_path);
```
skuenzer, line 249: `err_free_mp_m_path:`

lib/vfscore/mount.c
```diff
@@ -236,6 +244,8 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	if (dp_covered)
 		drele(dp_covered);
  err2:
+	free(mp->m_special);
```
skuenzer, line 247: If a jump needs to go here: `err_free_mp_m_special:`

lib/vfscore/mount.c
```diff
@@ -171,8 +171,16 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	mp->m_flags = flags;
 	mp->m_dev = device;
 	mp->m_data = NULL;
-	strlcpy(mp->m_path, dir, sizeof(mp->m_path));
-	strlcpy(mp->m_special, dev, sizeof(mp->m_special));
+	mp->m_path = strndup(dir, PATH_MAX - 1);
+	if (!mp->m_path) {
+		error = ENOMEM;
+		goto err2;
```
skuenzer: On errors `mp->m_path` and `mp->m_special` are empty, so you should go to free'ing `mp`, instead.
```c
goto err_free_mp;
```

lib/vfscore/mount.c
```diff
@@ -171,8 +171,16 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
...
+	}
+	mp->m_special = strndup(dev, PATH_MAX - 1);
+	if (!mp->m_special) {
+		error = ENOMEM;
+		goto err2;
```
skuenzer:
```c
goto err_free_m_path;
```

skuenzer (changes requested):
Hi @dinhngtu,

thanks a lot for the revision. Can you update the jump labels? Normally we avoid doing `free(NULL)`, so you would need to split your `err2:` block.

Thanks,

Simon

dinhngtu:
I have updated the jump labels as requested.

nderjung (changes requested):
Thanks for this. Just a note on the Git title formats. The two commit titles should be:
```
lib: Add initcalls to initialize `struct mounts`
lib/vfscore: Dynamically allocate struct mount paths
```
Thanks!

dinhngtu:
I've renamed them. The first commit message was shortened slightly to fit into the length limit.

lib/posix-event/epoll.c
```diff
@@ -371,3 +373,17 @@ int epoll_pwait(int epfd, struct epoll_event *events, int maxevents,
...
 #endif /* UK_LIBC_SYSCALLS */
+
+static int epoll_mount_init(void)
+{
+	epoll_mount.m_path = strdup("");
+	if (!epoll_mount.m_path)
+		return -ENOMEM;
+	epoll_mount.m_special = strdup("");
+	if (!epoll_mount.m_special) {
+		free(epoll_mount.m_path);
```
skuenzer: I'd actually prefer a goto logic as we do in general for error exits. The reason is that this can be extended more easily later:

```c
static int epoll_mount_init(void)
{
	epoll_mount.m_path = strdup("");
	if (!epoll_mount.m_path)
		goto err_out;
	epoll_mount.m_special = strdup("");
	if (!epoll_mount.m_special) {
		git err_free_m_path;
	return 0;

err_free_m_path:
	free(epoll_mount.m_path);
err_out:
	return -ENOMEM;
}
```

lib/posix-event/eventfd.c
```diff
@@ -438,3 +440,17 @@ int eventfd(unsigned int initval, int flags)
...
 #endif /* UK_LIBC_SYSCALLS */
+
+static int eventfd_mount_init(void)
+{
+	eventfd_mount.m_path = strdup("");
+	if (!eventfd_mount.m_path)
+		return -ENOMEM;
+	eventfd_mount.m_special = strdup("");
+	if (!eventfd_mount.m_special) {
+		free(eventfd_mount.m_path);
```
skuenzer: same here

lib/posix-socket/socket_vnops.c
```diff
@@ -342,3 +344,17 @@ static struct vfsops posix_socket_vfsops = {
...
 };
+
+static int posix_socket_mount_init(void)
+{
+	posix_socket_mount.m_path = strdup("");
+	if (!posix_socket_mount.m_path)
+		return -ENOMEM;
+	posix_socket_mount.m_special = strdup("");
+	if (!posix_socket_mount.m_special) {
+		free(posix_socket_mount.m_path);
```
skuenzer: same here

lib/vfscore/pipe.c
```diff
@@ -718,3 +719,17 @@ int mkfifo(const char *path __unused, mode_t mode __unused)
...
 #endif /* UK_LIBC_SYSCALLS */
+
+static int pipe_mount_init(void)
+{
+	p_mount.m_path = strdup("");
+	if (!p_mount.m_path)
+		return -ENOMEM;
+	p_mount.m_special = strdup("");
+	if (!p_mount.m_special) {
+		free(p_mount.m_path);
```
skuenzer: same here

skuenzer (commented):
Hi @dinhngtu,

thanks for the update. I'd still prefer a different prefix for the second commit, something like `lib/posix-event, lib/posix-socket, lib/vfscore:`. Unfortunately, this is pretty long.
However, I noticed that this change is related to the first commit. Probably you should just merge them together into a single commit and describe in the commit subject that you allocate empty strings for mount structs for lib/posix-event, lib/posix-socket, lib/vfscore, too. Btw, `NULL` or just `""` instead of `strdup("")` would not work? Are these mount structs ever free'd?

Thanks,

Simon

dinhngtu:
I've changed the init logic accordingly and squashed the commits.

I didn't want to use `NULL` or `""` for the mount paths in these structs since I wanted to kept these strings being dynamically-allocated strings like other mount structs. I think it'd be confusing to have some mount path strings being dynamically allocated and others not.

skuenzer:
> I've changed the init logic accordingly and squashed the commits.
> 
> I didn't want to use `NULL` or `""` for the mount paths in these structs since I wanted to kept these strings being dynamically-allocated strings like other mount structs. I think it'd be confusing to have some mount path strings being dynamically allocated and others not.

I agree with `""` but `NULL` would be still okay, however, I am not sure if a lot of code needs to be patched for this case.
Other from this the changes are fine to me.

skuenzer (approved):

PR #832, opened 2023-04-11
lib/ukvmem: Allow non-writable shared file mappings

csvancea:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Non-writable shared file mappings are now treated as anonymous read-only file mappings. Note that any writes made to the underlying file while the mapping is established will not be reflected in memory.

### DIFF
--- a/lib/ukvmem/include/uk/vmem.h
+++ b/lib/ukvmem/include/uk/vmem.h
@@ -225,6 +225,8 @@ struct uk_vma_ops {
 	 *   The length of the virtual memory area in bytes
 	 * @param data
 	 *   Pointer to VMA-type specific data supplied to the map function
+	 * @param attr
+	 *   Page attributes for the pages in the memory area
 	 * @param[in,out] flags
 	 *   The flags supplied to the map function. The handler may
 	 *   set/unset the following flags (besides VMA-type specific ones):
@@ -237,7 +239,8 @@ struct uk_vma_ops {
 	 *   0 on success, a negative errno error otherwise
 	 */
 	int (*new)(struct uk_vas *vas, __vaddr_t vaddr, __sz len, void *data,
-		   unsigned long *flags, struct uk_vma **vma);
+		   unsigned long attr, unsigned long *flags,
+		   struct uk_vma **vma);
 
 	/**
 	 * Frees any references that this VMA might hold. The memory for the
--- a/lib/ukvmem/vma_dma.c
+++ b/lib/ukvmem/vma_dma.c
@@ -29,8 +29,8 @@ static __vaddr_t vma_op_dma_get_base(struct uk_vas *vas __unused,
 #endif /* CONFIG_LIBUKVMEM_DMA_BASE */
 
 int vma_op_dma_new(struct uk_vas *vas, __vaddr_t vaddr __unused,
-		   __sz len __unused, void *data, unsigned long *flags __unused,
-		   struct uk_vma **vma)
+		   __sz len __unused, void *data, unsigned long attr __unused,
+		   unsigned long *flags __unused, struct uk_vma **vma)
 {
 	struct uk_vma_dma_args *args = (struct uk_vma_dma_args *)data;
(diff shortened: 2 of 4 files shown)


marcrittinghaus:
Hey @csvancea,

looks very good! Thanks! 

One thing I am still thinking about is, if we should do the fix in the file implementation in `ukvmem` instead. Because there is also the "Shared is not supported"-check, which we could extend. Then `posix-mmap` behaves according to specification and just `ukvmem` includes the workaround. Which is fair because it also has the limitation. In the future, when we extend `ukvmem`, we automatically would remove the workaround with the "Shared is not supported"-check, whereas the modification in `posix-mmap` you need to explicitly remember and revert.

What do you think?

csvancea:
Hello @marcrittinghaus,

The initial plan was to apply the workaround to `ukvmem` here: 
https://github.com/unikraft/unikraft/blob/fd188c0ebda3a685e48fa3fc224757d573abe3bd/lib/ukvmem/vma_file.c#L46-L48

but that requires us to change the signature of the `new` VMA operation and add the page protection as an extra argument. If that's okay, I will make the changes and move the protection check to `vma_op_file_new`.

marcrittinghaus:
> but that requires us to change the signature of the `new` VMA operation and add the page protection as an extra argument. If that's okay, I will make the changes and move the protection check to `vma_op_file_new`.

Yes, sounds good. I think the positive aspects outweigh the fact that we have to change the signature.

andraprs (approved):
Reviewed-by: Andra Paraschiv <andra@unikraft.io>

Thank you.

marcrittinghaus:
Hey @csvancea,

this looks good. One last thing that worries me is that you can still set write permissions on the memory later and this can lead to subtle bugs as changes would not be flushed to disk or shared with additional mappings of the same file.

I suggest we also prevent the program from changing the protections to writeable later. In case an application attempts to do, we quickly see this with the failed `mprotect`/ukvmem syscall. And maybe the app stops early instead of running in some hard to debug silent inconsistency.

We can easily do this by introducing `vma_op_file_set_attr()`. You can use the same function body as `vma_op_set_attr()` except that you check for `vma->flags & UK_VMA_FILE_SHARED` and `WRITEABLE` and return `-EPERM` in this case.

It is not perfect since the inconsistency with `write()` still exists, but it would at least remove one source of problems.

What do you think?

csvancea:
Hey @marcrittinghaus,

Great suggestions! I have amended the PR and ran some tests to make sure that everything works as intended.

However, testing revealed that any failed `set_attr` call triggers a 'kernel' panic [1], which probably is not the best way to handle `mprotect(..., PROT_WRITE)`. Shall we leave it as it is (since it is likely that programs won't check the result of `mprotect` and they would crash anyway at a later point when they try to write to that memory -- crashing right away at least makes the error obvious), or propagate the error back to the caller and hope that the user handles it accordingly?

[1] https://github.com/unikraft/unikraft/blob/eb9f3a92df2019cc962d96255c20df6cf04f5d37/lib/ukvmem/vmem.c#L774-L783

marcrittinghaus (approved):
Thanks! All fine 😃

marcrittinghaus:
> However, testing revealed that any failed `set_attr` call triggers a 'kernel' panic [1], which probably is not the best way to handle `mprotect(..., PROT_WRITE)`. Shall we leave it as it is (since it is likely that programs won't check the result of `mprotect` and they would crash anyway at a later point when they try to write to that memory -- crashing right away at least makes the error obvious), or propagate the error back to the caller and hope that the user handles it accordingly?

Lets keep it like this for now, so we have the "feature" in. I have to revisit if propagating the error is easily possible. This is because we are potentially splitting areas etc. when changing the protections. And as you said, this will fail probably anyways later in the application and it is more direct now.

John-Ted (approved):

PR #839, opened 2023-04-16
xen/x86/mm.c: fix unchecked uk_palloc value

hlef:
### Prerequisite checklist

### Base target

 - Architecture(s): x86
 - Platform(s): xen
 - Application(s): N/A

### Description of changes

The return value of uk_palloc is not checked in _arch_init_p2m.

Allocation failure at that point should never happen, but it's not a reason not to check.

At this stage, in the even of a failure, we should probably abort the boot altogether.

This bug was detected using the following Coccinelle spatch:

    @call@
    expression ptr;
    position p;
    @@

    ptr@p = uk_palloc(...);

    @ok@
    expression ptr;
    position call.p;
    @@

    ptr@p = uk_palloc(...);
    ... when != ptr
    (
     (ptr == NULL || ...)
    |
     (ptr != NULL || ...)
    )

    @depends on !ok@
    expression ptr;
    position call.p;
    @@

    ptr@p = uk_palloc(...);
    + if (ptr == NULL) return;

Note: I did not test these changes because I do not have a Xen setup.

### DIFF
--- a/plat/xen/x86/mm.c
+++ b/plat/xen/x86/mm.c
@@ -651,9 +651,13 @@ void _arch_init_p2m(struct uk_alloc *a)
 		UK_CRASH("Error: Too many pfns.\n");
 
 	l3_list = uk_palloc(a, 1);
+	if (l3_list == NULL)
+		UK_CRASH("Error: Cannot allocate l3_list.\n");
 	for (pfn = 0; pfn < max_pfn; pfn += P2M_ENTRIES) {
 		if (!(pfn % (P2M_ENTRIES * P2M_ENTRIES))) {
 			l2_list = uk_palloc(a, 1);
+			if (l2_list == NULL)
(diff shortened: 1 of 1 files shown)


skuenzer (changes requested):
Thanks for this update. Can you also fix the prefix of the commit title here?
For example: `plat/xen: Fix unchecked uk_palloc() value in x86/mm.c`

StefanJum:
@hlef Everything works fine, one small typo in the commit message:
`At this stage, in the even of a failure`: `even -> event`.

hlef:
@skuenzer @StefanJum thanks, I have updated the commit message.

StefanJum (approved):
All good now @hlef, thanks. I'll leave the tag after @skuenzer goes through it.

skuenzer (approved):
@hlef: Thanks for the update!

Approved-by: Simon Kuenzer  <simon@unikraft.io>

PR #843, opened 2023-04-16
lib/vfscore: fix potential free(NULL)

hlef:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N?A
 - Application(s): N/A

### Description of changes

Very clearly this code is not correct. If `names` is always non-NULL then the `if` is useless. Otherwise `names` should be freed in the `if`. Since I'm not sure I am just playing it safe and freeing `names` in the `if`.

This bug was found by Coccinelle with the following spatch:

    @@
    expression E;
    @@

    if(E) {...}
    -free(E);

NOTE: I only minimally tested this. Please carefully double check it.

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -1054,10 +1054,11 @@ int scandir(const char *path, struct dirent ***res,
 	closedir(d);
 
 	if (errno) {
-		if (names)
+		if (names) {
 			while (cnt-->0)
 				free(names[cnt]);
-		free(names);
+			free(names);
+		}
 		return -1;
 	}
 	errno = old_errno;


hlef:
Actually I just realized that `free(NULL)` is legal. We might want to add a comment here instead, since it will probably be confusing for most people.

dinhngtu (approved):
Like you said, `free(NULL)` is safe but this change should eliminate some confusion.

marcrittinghaus (approved):
I agree. `free(NULL)` is valid, but I also prefer to include a check. This explicitly documents that the pointer might be `NULL` and you were thinking about this when writing the code. It is also a reminder for people extending the code later.

Just removed the approved-by tag. Please rename the commit to `lib/vfscore: Fix potential free(NULL)`.

nderjung (changes requested):
Hi, though it wasn't caught in checkpatch, the title of the commit should be:

```
lib/vfscore: fix potential free(NULL)
```

marcrittinghaus:
Ha, noticed it right before your message 😄

Starnox (approved):
I also agree. This straightforward way of handling this case is more explicit in intention.

Reviewed-by: Eduard-Florin Mihailescu <mihailescu.eduard@gmail.com>

marcrittinghaus (approved):

PR #848, opened 2023-04-19
Integrate `ukplat_memregion` into `Xen` `x86`, `AArch64` `KVM` and `linuxu`

mogasergiu:
### Prerequisite checklist

### Base target

 - Application(s):

### Description of changes

Integrate `ukplat_memregion_*` into `KVM` `ARM64` and `Xen` `x86` and `linuxu` to ease memory region management. Furthermore, add an `ARM64` specific intermediary page-fault handler to `ukvmem` to enable `ARM64` to have access to this library and better virtual memory management.

Note: This PR depends on #772 and since it has not been merged it, it contains its commits.

### NOTE


support/scripts/elf64_to_32.py
jschlumpp, line 1: Out of curiosity: Is there a reason to use this over `objcopy --output-target elf32-i386`? 
But I have maybe an even better idea: maybe the script could *prepend* the 32-bit ELF header + multiboot header to the image. Then we would have the proper 64-bit ELF header in the address space and we could also load the symbol table in the future for symbolization of crash back traces.

mogasergiu: This is related to the PIE PR linked in the description. I thought that a custom script offered more control, as objcopy does some things differently, for example it re-adjusts the offset to the Program Headers, while the script leaves everything as it is. Besides, unrelated to ELF32, but I had some issues whenever I would use it with `--target efi-app-x86_64 --subsystem=10` on the 64-bit PIE and I thought that having something I control would be better in general, as I found `objcopy` to sometimes prove unreliable when converting formats.

I agree, that is a great idea, I will change this so it prepends it instead :) good point.

mogasergiu:
I will bring some more changes to this PR in the following 2 days (it's still a draft). Afterwards, hopefully, I will mark it as ready for review, together with the linked PIE PR.

support/scripts/elf64_to_32.py
mogasergiu: One problem that I have just thought of though: this implies building the code as if it  were aware of the elf64 header. So I guess this means that we should include it into a PT_LOAD program header?

mogasergiu: Or better yet, perhaps make the script change the in-file offset/size of the first PT-LOAD to also include the elf64 header.

jschlumpp: Yes, basically the kernel would end up in configuration as if the loader would have loaded the original ELF64.
Actually using ELF64 for a x86_64 binary would be the more sensible behavior (Currently it doesn't matter because there is no code in staging that reads the ELF header). Using ELF32 was more of a hack when I removed the multiboot `AOUT_KLUDGE` and `--omagic` flag.
It's not necessary to use the script to insert a new segment for the ELF header (including program headers). Adjusting the linker script to include them in the first PT_LOAD segment is enough.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,39 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
 	return (int)i;
 }
 
+/**
+ * Inserts the region into the memory region list by index. The list is kept in
```
michpappas, line 173: >  The list is kept in ascending order of physical base addresses. Order is only preserved among
> ranges that do not overlap. 

Inserting to index does not maintain the ordering by paddr, unless the caller knows what they're doing, so the description should be fixed. I see this is used for a very specific purpose in " plat/common: Track index in ukplat_memregion_list_insert_split_phys " where the caller knows what they're doing indeed.

Given that the requirement is that memregs must be in order and not overlapping, it would be best if it was made clear that this function should not be used unless strictly needed. For that it would be best if:
1. Update the description
2. Remove doxygen documentation

plat/kvm/x86/setup.c
```diff
@@ -220,7 +220,11 @@ static inline int cmdline_init(struct ukplat_bootinfo *bi)
 		cmdline_len = sizeof(CONFIG_UK_NAME) - 1;
 	}
 
-	cmdline = ukplat_memregion_alloc(cmdline_len + 1, UKPLAT_MEMRT_CMDLINE);
+	/* This is not the original command-line, but one that will be thrashed
```
michpappas, line 42: What problem does this change fix? Is that only a change at the semantics level?

plat/common/memory.c
```diff
@@ -137,3 +139,217 @@ void *ukplat_memregion_alloc(__sz size, int type)
...
+			m[i] = m[i + 1];
+			m[i + 1] = tmp;
+		}
+		ml = &m[i];
+		mr = &m[i + 1];
+		ml_prio = get_mrd_prio(ml);
+		mr_prio = get_mrd_prio(mr);
+
+		/* If they overlap */
+		if (RANGE_OVERLAP(ml->pbase,  ml->len, mr->pbase, mr->len)) {
```
michpappas, line 261: This checks whether the regions overlap, but not whether they are part of the same mapping, which causes such regions not to be coalesced, which eventually manifests into a `-EXIST` error when the regions are mapped.

Fixing this here can actually be a bit tricky as you will need to know the granularity these regions are mapped with. If you decide to address this by allocating at least page-sized regions, you effectively change the requirements of the memregion API, as you will need to enforce that when allocating / adding new regions. CC: @marcrittinghaus.

Finally, notice that theoretically speaking this is still not fail-safe, as you have no guarantees that objects placed in memory by the bootloader (initrd, dtb, cmdline) are placed with granularity > `PAGE_SIZE`. I think in practice this is a non-issue, as boot protocols come with requirements for these objects (e.g. https://www.kernel.org/doc/Documentation/arm64/booting.txt), yet it's something worth to keep in mind.

plat/kvm/arm/entry64.S
```diff
@@ -36,6 +36,13 @@
 #include <uk/config.h>
 #include <arch/reloc.h>
 
+/* Prefer using in-image stack, to avoid conflicts when unmapping is done by
```
michpappas, line 39: How come we didn't come across this problem before? Not necessarily in this PR but it would be nice if we didn't waste a page for the bootstack.

plat/common/bootinfo.c
```diff
@@ -6,6 +6,12 @@
 
 #include <uk/essentials.h>
 #include <uk/plat/common/bootinfo.h>
+#include <uk/plat/common/sections.h>
+#ifdef CONFIG_LIBFDT
+#include <libfdt.h>
+#endif /* CONFIG_LIBFDT */
+
+#define ukplat_bootinfo_crash(s)		ukplat_crash()
```
michpappas: What is the purpose of this?

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+{
+	struct ukplat_bootinfo *bi;
+
+	bi = ukplat_bootinfo_get();
+	if (unlikely(!bi))
+		ukplat_bootinfo_crash("Invalid bootinfo");
+
+	fdt_bootinfo_fdt_mrd(bi, fdtp);
+	fdt_bootinfo_mem_mrd(bi, fdtp);
+	ukplat_memregion_list_coalesce(&bi->mrds);
```
michpappas: As discussed: although not strictly required because `fdt_bootinfo_cmdl_mdr()` calls `ukplat_memregion_alloc()`, I think it's better form if you placed `ukplat_memregion_list_coalesce()` in the end, as it shows that you don't make any assumptions about the implementation of the various `fdt_bootinfo_*_mrd()` functions.

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+	naddr = fdt_address_cells(fdtp, nmem);
+	if (unlikely(naddr < 0 || naddr >= FDT_MAX_NCELLS))
+		ukplat_bootinfo_crash("Could not find proper address cells!");
+
+	nsz = fdt_size_cells(fdtp, nmem);
+	if (unlikely(nsz < 0 || nsz >= FDT_MAX_NCELLS))
+		ukplat_bootinfo_crash("Could not find proper size cells!");
+
+	/*
+	 * QEMU will always provide us at least one bank of memory.
```
michpappas: Is there a reason not to support >1 memory banks? IIRC the comment about QEMU supporting one bank originates from `plat/kvm/arm/setup.c`, yet it's time to get rid of the QEMU-specific mentality as Unikraft should be capable of running on any platform that provides a dtb on `arm64` (and this code is not part of `plat/qemu`). So if there is a blocker to support multiple banks, I would prefer if the comment was rephrased with the actual reasoning.

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+static void fdt_bootinfo_mem_mrd(struct ukplat_bootinfo *bi, void *fdtp)
+{
+	struct ukplat_memregion_desc mrd = {0};
+	int prop_len, prop_min_len;
+	__u64 mem_base, mem_sz;
+	int naddr, nsz, nmem;
+	const __u64 *regs;
+	int rc;
+
+	/* search for assigned VM memory in DTB */
```
michpappas: This checks the validity of the header, so the comment and the error message are misleading (besides the fact that assigned and reserved memory are opposite things). Perhaps by this point checking against the header is redundant? See also the relevant comment in `fdt_bootinfo_fdt_mrd()`.

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+
+	mem_sz = fdt64_to_cpu(regs[1]);
+	mem_base = fdt64_to_cpu(regs[0]);
+	if (unlikely(!RANGE_CONTAIN(mem_base, mem_sz,
+				    __BASE_ADDR, __END - __BASE_ADDR)))
+		ukplat_bootinfo_crash("Image outside of RAM");
+
+	mrd.vbase = (__vaddr_t)mem_base;
+	mrd.pbase = (__paddr_t)mem_base;
+	mrd.len   = __BASE_ADDR - mem_base;
```
michpappas: If the image is placed at the top of the region this will insert a region with zero length.

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+	mrd.type  = UKPLAT_MEMRT_FREE;
+	mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE;
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+	if (unlikely(rc < 0))
+		ukplat_bootinfo_crash("Could not add free memory descriptor");
+
+	mrd.vbase = (__vaddr_t)__END;
+	mrd.pbase = (__paddr_t)__END;
+	mrd.len   = mem_base + mem_sz - __END;
```
michpappas: If the image is placed at the end of the region this will insert a region of zero length.

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+	 * and size, each of which is 8-bytes.
+	 * TODO: Support more than one memory@ node/regs/ranges properties.
+	 */
+	prop_len = 0;
+	prop_min_len = (int)sizeof(fdt32_t) * (naddr + nsz);
+	regs = fdt_getprop(fdtp, nmem, "reg", &prop_len);
+	if (unlikely(!regs || prop_len < prop_min_len))
+		ukplat_bootinfo_crash("Bad 'reg' property or more than one memory bank.");
+
+	mem_sz = fdt64_to_cpu(regs[1]);
```
michpappas: Doesn't `fdt64_to_cpu()` imply 2 cells?

plat/common/bootinfo.c
```diff
@@ -113,3 +119,142 @@ void ukplat_bootinfo_print(void)
...
+	bi->cmdline = (__u64)cmdl;
+	bi->cmdline_len = (__u64)cmdl_len;
+}
+
+static void fdt_bootinfo_fdt_mrd(struct ukplat_bootinfo *bi, void *fdtp)
+{
+	struct ukplat_memregion_desc mrd = {0};
+	int rc;
+
+	rc = fdt_check_header(fdtp);
```
michpappas: Related to the comment in `fdt_bootinfo_mem_mrd()`: Perhaps you could move this check in `ukplat_bootinfo_fdt_setup()` after `ukplat_bootinfo_get()`.

plat/common/paging.c
```diff
@@ -741,7 +741,7 @@ static int pg_page_mapx(struct uk_pagetable *pt, __vaddr_t pt_vaddr,
 			pt->nr_lx_pages[lvl]++;
 #endif /* CONFIG_PAGING_STATS */
 
-		if (PT_Lx_PTE_PRESENT(orig_pte, lvl))
+		if (PT_Lx_PTE_PRESENT(orig_pte, lvl) && pt == pg_active_pt)
```
michpappas, line 744: Okay, yet it seems that this change is not relevant to this series which contains enough changes to review already. If that's the case, please move into a separate PR.

plat/common/paging.c
```diff
@@ -1418,3 +1421,101 @@ void ukplat_page_kunmap(struct uk_pagetable *pt, __vaddr_t vaddr,
...
+		prot |= PAGE_ATTR_PROT_READ;
+	if (flags & UKPLAT_MEMRF_WRITE)
+		prot |= PAGE_ATTR_PROT_WRITE;
+	if (flags & UKPLAT_MEMRF_EXECUTE)
+		prot |= PAGE_ATTR_PROT_EXEC;
+
+	return prot;
+}
+
+int ukplat_paging_init(void)
```
michpappas, line 1441: I think the right place to move this is `ukboot`,as the paging subsystem is (and should be) oblivious of `ukplat_memregion`. CC: @marcrittinghaus.

plat/common/memory.c
```diff
@@ -380,3 +381,116 @@ int ukplat_memregion_get(int i, struct ukplat_memregion_desc **mrd)
...
+	 */
+	unmap_start = ALIGN_DOWN(bpt_unmap_mrd.vbase, __PAGE_SIZE);
+	unmap_end = unmap_start + ALIGN_DOWN(bpt_unmap_mrd.len, __PAGE_SIZE);
+
+	/* After Kernel image */
+	rc = ukplat_memregion_list_insert(&bi->mrds,
+			&(struct ukplat_memregion_desc){
+				.vbase = ALIGN_UP(__END, __PAGE_SIZE),
+				.pbase = 0,
+				.len   = unmap_end -
```
michpappas, line 380: Are there any conditions where any of the unmap regions has zero length?

plat/common/paging.c
```diff
@@ -1508,7 +1508,7 @@ int ukplat_paging_init(void)
 
 		rc = ukplat_page_map(&kernel_pt, vaddr, paddr,
 				     len >> PAGE_SHIFT, prot, 0);
-		if (unlikely(rc && rc != -EEXIST))
+		if (unlikely(rc))
```
michpappas, line 1520: Leaving this here as github does not allow commenting on the commit message:
```
before commit <generalize mem init commit sha here>, Unikraft was
```
You can't add a commit hash here even when the PR is finalized, because the hashes will changed by tags added by GitHub upon push.

plat/common/include/uk/plat/common/memory.h
```diff
@@ -112,6 +112,32 @@ ukplat_memregion_list_insert(struct ukplat_memregion_list *list,
 	return (int)i;
 }
 
+#if defined(__X86_64__)
+#define	X86_HI_MEM_START		0xA0000
+#define X86_HI_MEM_LEN			0x50000
+
+static inline int
+ukplat_memregion_list_insert_legacy_hi_mem(struct ukplat_memregion_list *list)
```
michpappas, line 121: Similarly for the following `SIPI` patch, these are too `x86_64` specific to be here. I understand that in the current state of the sources there may be not clear where these should be. According to the plat re-arch conventions that should probably be under `lib/ukboot/arch/x86_64/`.

plat/common/Makefile.rules
```diff
@@ -1,3 +1,15 @@
 define build_elf64_to_32 =
 	@$(SCRIPTS_DIR)/elf64_to_32.py $(1) -o $(1)
 endef
+
+BINFO_FLAGS := -a $(CONFIG_UK_ARCH)
```
michpappas, line 5: According to plat re-arch these should go into the `Makefile.rules` of `lib/ukboot` so that they can be re-used by both internal and external platforms.

plat/common/bootinfo.c
```diff
@@ -10,8 +10,8 @@
 /**
  * Space reservation for boot information. Will be populated by build script.
  */
-__section(".uk_bootinfo") __align(8) __used
-static char bi_bootinfo_sec[UKPLAT_BOOTINFO_SIZE(32)];
+__section(".uk_bootinfo") __align(8) __used static char
```
michpappas, line 13: I'm not very keen on the `fdt` code being in this file that contains generic functionality. By plat re-arch `bootinfo` will live under `lib/ukboot`, and different implementations that populate bootinfo like `multiboot`, `mbr`, `uefi` etc will be under `drivers/ukboot/xxx/`. Perhaps we should start moving things now instead of creating more work for re-arch?

plat/common/memory.c
```diff
@@ -91,6 +92,13 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 		ostart = mrd->pbase;
 		olen   = mrd->len;
 
+		/* Check whether we are allocating from an in-image memory hole
```
michpappas, line 96: Can you please provide some more context for this change? What is an example of such an allocation? What I read here is that if the allocation is outside the image, then we map. This makes the "hole" term confusing to me. Moreover, on the comment and in the description in the commit message you refer to "in-image holes" while the title is about "memory holes" in general.

plat/kvm/arm/bpt64.S
```diff
@@ -203,6 +203,6 @@ arm64_bpt_l1_pt1:
  * 0x0000000040000000 - 0x00000000401fffff	Kernel	@ 1GiB
  */
 .align 12
+.globl arm64_bpt_l0_pt0
 arm64_bpt_l0_pt0:
-	pte_fill	0x0000000040000000, 256, 0, PTE_PAGE_NORMAL_RO
-	pte_fill	0x0000000040100000, 256, 0, PTE_PAGE_NORMAL_RWX
+	pte_fill	0x0000000040000000, 512, 0, PTE_PAGE_NORMAL_RWX
```
michpappas, line 208: Is this change in the scope of this commit? If so it should be described in the commit message.
Since we're there, also the commit message contains a bit too many "Furthermores", you may want to rephrase  :)

plat/common/memory.c
```diff
@@ -137,3 +139,217 @@ void *ukplat_memregion_alloc(__sz size, int type)
...
+
+		/* If they overlap */
+		if (RANGE_OVERLAP(ml->pbase,  ml->len, mr->pbase, mr->len)) {
+			/* If they are not of the same priority */
+			if (ml_prio != mr_prio) {
+				/* If the overlapping memory region of type
+				 * UKPLAT_MEMRT_KERNEL is of lower priority,
+				 * then something is definetely wrong => abort!
+				 */
+				if ((ml_prio < mr_prio &&
```
michpappas: I don't quite follow this, even with the comment, can you please provide an example?

plat/common/memory.c
```diff
@@ -137,3 +139,217 @@ void *ukplat_memregion_alloc(__sz size, int type)
...
+				/* Remove dropped regions */
+				if (ml->len == 0)
+					ukplat_memregion_list_delete(list, i);
+				else if (mr->len == 0)
+					ukplat_memregion_list_delete(list,
+								     i + 1);
+				else
+					i++;
+
+			/* If they are of the same priority and different flags,
```
michpappas: I'm a bit concerned about the silent keeping one region and deleting another. Would it be better to return an error unconditionally instead?

michpappas (changes requested):
@mogasergiu these changes are a big improvement! 🚀 Due to the sheer amount of changes I limited my review to common code and arm, ie I skipped linuxu, xen, and x86_64. Perhaps another reviewer can help with these.

You will notice that in several places I mention how the code should be laid out accoriding to the plat re-arch guidelines. Not sure if this should be done in the scope of this PR, perhaps we can discuss this offline.

plat/kvm/x86/setup.c
mogasergiu: > What problem does this change fix? Is that only a change at the semantics level?

It bothered me how the type of this memory region was inconsistent with reality. This was not the command-line per se, but the command-line whose arguments will be separated by `\0`'s. This is inconsistent with the memory region that actually contains the command-line as a whole string. So I would like to keep this change. :)

plat/common/memory.c
mogasergiu: I would solve this by allocating at least page-sized regions, as you have said. But, as you have also mentioned, you end up having to enforce this in the insertion API's as well, meaning that the previous boot phase that passes memory map related stuff will have to ensure that the raw received memory regions do not exist in the same mapping.

This is the solution that I would come up with now: a separate function, called something among the lines of `ukplat_memregion_list_page_overlap_adjust`, that should be called after `ukplat_memregion_list_coalesce`, that simply iterates through memory regions and removes `UKPLAT_MEMRF_MAP` if an in-page overlap is found. I would definetely not do this in the coalescing function as it would get out of its signature's scope.

So that is one suggestion. However, I still do not like it, as I believe that `ukplat_memregion_*` API's should have nothing to do with virtual memory or care about it as that kind of beats the purpose of having a paging API and a memregion API. I believe that `ukplat_memregion_*` API should only do what their names imply and take care of maintaining a proper raw memory map representation. Therefore, I think this case should be taken care of by `ukplat_paging_init`'s loop that maps memory regions with `UKPLAT_MEMRF_MAP`. Especially since it also aligns them to page-sized regions as well, making it already very easy to keep track of which page was mapped previously as they are already sorted ascendingly.

Let me know what you think of these alternate solutions. I would go with the latter.

plat/kvm/arm/entry64.S
mogasergiu: I do not think a page is that much in our usual scenarios. The problem I encountered related to this line was 4 months ago, but from what I vaguely remember, it caused problems when setting up the memory regions, because I would have had to manually add a hardcoded separate function call just to insert the related memory region. Furthermore, at the time of doing this change I also did not have the `umap_bpt_mrd` idea, so a lot has changed since then. I will check again what happens if I do not do this, as I do not remember very well my intentions here.

plat/common/bootinfo.c
mogasergiu: Just a wrapper like you would see in the initial Firecracker/Multiboot/UEFI code. Since console is not initialized at that point, I thought I would align this code with what has already been written for what would be executed before the console is initialized.

plat/common/bootinfo.c
mogasergiu: This code was just taken from the place you have already pointed to, I did not bother with improving it at that time. If I were to make this suggested change, it would have been in another commit. If you really want me to, I can update it to look for other memory banks as well, it should not be difficult and I would not mind it at all :).

plat/common/bootinfo.c
mogasergiu: Very good point! I thought it would not matter since `ukplat_paging_init` ignores such regions and nothing would break. But yeah, the memory map ends up having an unnecessary memory region descriptor. Will add a check for it!

plat/common/paging.c
mogasergiu: Yep, that would more sense for me as well, you are right :D.

EDIT: Nevermind, I think `ukplat_paging_init` should stay here until `setup.c`'s are merged. This is needed to make it easier to get rid of the BPT's while also being able to have early, pre-ukboot, access to important memory regions that need to be accessed early, like ACPI stuff. I believe this function should be part of the final, merged `setup.c`, when it will happen.

plat/common/memory.c
mogasergiu: `ukplat_page_unmap` ignores them and they are removed after `ukplat_paging_init` is done. So I think it is fine. I would keep it that way so I do not have to add a check when removing the two unmap regions.

plat/common/paging.c
mogasergiu: I believe I wanted to have this commit in a separate PR, so that when it would have gotten merged I could have just amended the message. Oh well, guess I will just keep the commit message headline.

plat/common/bootinfo.c
mogasergiu: Ok, I will have these placed in a `bootinfo_fdt.c` then.

plat/common/memory.c
mogasergiu: I am actually unsure whether this is needed anymore. I wrote this when writing UEFI for arm, at the beginning, before I was even met with the memregion missing for ARM... *sigh* :rofl:. At that time, ARM would set up its memory relative to ld symbols.

plat/common/memory.c
mogasergiu: Some ELF segments are separated by holes less than page-size, due to alignments and my coalesce function would mark those holes as free memory so that nothing goes to waste. If you allocate from there say a stack or cmdl through memregion_alloc, it would have MEMRF_MAP. And you get an EEXIST, because the ELF segments are mapped there already.

plat/kvm/x86/setup.c
michpappas: Yes you're right. In fact I was also confused in the past with the two `cmdline` entries, so maybe this is a good change 👍🏼

plat/common/memory.c
michpappas: > I would solve this by allocating at least page-sized regions

Just to be clear: multiple of page size , not at least page-sized

> Let me know what you think of these alternate solutions. I would go with the latter.

I understand the argumentation towards separation, but in practice I don't think you can ignore PT requirements until you reach the point you map, otherwise you could potentially end up with regions with conflicting attributes.

Consider for example the bug that triggered this discussion in the first place, where the `cmdl` and `dtb` regions were in the same page. The paging API could merge these into a single mapping if they had consistent protections and other PT attributes. But in any other case, you would end up with `-EEXIST`, which is what we are trying to avoid here. Because of that I see no choice other than enforcing regions to be a multiple of page size when they are created, and this is also the reason boot specs define alignment requirements on images. Similarly in the UEFI spec (e.g. Sect. 2.3.6 AArch64 Platforms - but similarly for other arch too):
> Any UEFI memory descriptor that requests a virtual mapping via the EFI_MEMORY_DESCRIPTOR having
the EFI_MEMORY_RUNTIME bit set must be aligned on a 4 KiB boundary and must be a multiple of 4 KiB
in size.

Given the above, I see no other way to address this problem. Unless I missed something in your proposal, that is :)

plat/common/bootinfo.c
michpappas: Okay, thanks for clarifying

plat/common/bootinfo.c
michpappas: Okay. Adding support for multiple banks would go together with excluding reserved memory. Perhaps to avoid delay merging this PR further we can add it later. Maybe just reduce the comment to something like "for now we only support one memory bank"?

plat/common/paging.c
michpappas: Yes exactly, this will be in the merging of `setup.c` logic into `ukboot`. On whether to keep it into `ukplat_paging_init()` until then, perhaps @marcrittinghaus can provide his feedback, being the author of the paging API.

plat/common/memory.c
mogasergiu: > otherwise you could potentially end up with regions with conflicting attributes.

Ah, very good point, yes! Right, I am only going to make sure this does not happen with `ukplat_memregion_alloc` (allocate page-aligned in multiple of page-sized) only though. I will leave the rest the same and only let this function keep this condition in mind (should also add a comment for this). This means I will also remove the code snippet that takes care of "in-image memory holes".

plat/kvm/arm/entry64.S
mogasergiu: @michpappas I just checked, the reason why I did this is because the paging initialization code, that also unmaps everything outside the kernel, uses the earliest boot stack and therefore ends up unmaps our stack and we end up using an unmapped stack while the paging initialization code is running. I will update the comment to be more verbose.

plat/common/bootinfo.c
mogasergiu: I just remembered why I did this. `fdt_bootinfo_cmdl_mrd` makes use of `ukplat_memregion_alloc` and, therefore, if we were to call `ukplat_memregion_list_coalesce` after it, it would imply that the list has already been coalesced/sorted before doing that allocation. So I will keep it like this and add a clarifying comment.

michpappas (approved):

razvand (approved):

PR #683, opened 2022-12-05
`lib/posix-fdtab`: Introduce file descriptor library

jschlumpp:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBPOSIX_FDTAB=y`

### Description of changes

This separates the file descriptor handling from vfscore, so that libraries such as posix-socket work without pulling in a filesystem dependency.

Additionally, this also changes the error code semantic to use negative errors. This makes it possible to use positive return values in the success-case. In this PR this possibility is enabled for ioctl. 

This hopefully also includes all fixes done to vfscore after I started working on posix-fdtab.

PR #855, opened 2023-04-20
/lib/devfs, /lib/ukswran: Fix ioctl() on device files (+ cleanup) 

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR fixes the issue, where an `ioctl()` on one of the `/dev/` files causes a crash and an assert error in debug builds. The problem is that `devfs` expects all device operations to be properly set to a handler function. However, this is not the case for all device files in Unikraft. The PR fixes this by completing all device operation definitions and adding a `UK_ASSERT` for all operations to the registration. This way, missing handlers are detected earlier instead of requiring an explicit call.


John-Ted:
Is the checkpatch warning something that needs to be fixed? The PR looks good otherwise.

lib/devfs/device.c
```diff
@@ -77,95 +78,75 @@ device_lookup(const char *name)
...
-	uk_mutex_unlock(&devfs_lock);
-}
-
-
 /*
  * device_create - create new device object.
  *
  * A device object is created by the device driver to provide
  * I/O services to applications.  Returns device ID on
  * success, or 0 on failure.
```
razvand: This is wrong. It doesn't return `0` on error. It returns a positive (?) error code.

Also, given the `checkpatch` complaint: why is the return value positive (i.e. `ENOMEM` instead of `-ENOMEM`)?

razvand (commented):
Thanks, @marcrittinghaus. See my comments.

@John-Ted, @Alex-deVis, @rares-miculescu, please test this yourself.

lib/devfs/device.c
marcrittinghaus: Hey @razvand,

good point. I fixed the stale comment. It was already broken before. Unfortunately, all functions in devfs happen to return positive errno values (🤮) because they are used in `devfs_vnops.c` to handle vnode operations. At this point, vfscore still expects positive error codes so I stayed with the positive codes here, too.

The fdtab #683 PR changes this so that vfscore finally expects negative errno values to denote errors. The PR changes this also for `device_create` and the other functions in `devfs`. So the fdtab PR just needs to be rebased.

John-Ted (approved):

razvand (approved):

PR #862, opened 2023-04-22
plat/common/x86: Use FSGSBASE for TLS if available

andreittr:
FSGSBASE is an ISA extension that provides a way to access the FS and GS registers directly, without a need for an MSR read/write. This change adds support for using the `rdfsbase` and `wrfsbase` instructions to manipulate the TLS pointer, increasing performance on context switch.

### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

 - `CONFIG_MARCH_X86_64_COREI7AVXI=y` or any other x86 arch that has FSGSBASE

### DIFF
--- a/plat/common/include/x86/cpu.h
+++ b/plat/common/include/x86/cpu.h
@@ -97,6 +97,21 @@ static inline __u64 rdtsc(void)
 	return (h << 32) | l;
 }
 
+#ifdef __FSGSBASE__
+static inline __u64 rdfsbase(void)
+{
+	__u64 ret;
+
+	__asm__ __volatile__ ("rdfsbase %%rax" : "=a" (ret));
+	return ret;
+}
+
+static inline void wrfsbase(__u64 fsbase)
+{
+	__asm__ __volatile__ ("wrfsbase %0" : : "r" (fsbase));
+}
+#endif /* __FSGSBASE__ */
+
 
 /* accessing devices via port space */
 static inline __u8 inb(__u16 port)
--- a/plat/common/include/x86/tls.h
+++ b/plat/common/include/x86/tls.h
@@ -30,16 +30,22 @@
 #ifndef __PLAT_CMN_X86_TLS_H__
 #define __PLAT_CMN_X86_TLS_H__
 
-#include <x86/cpu.h> /* rdmsrl, wrmsrl */
+#include <x86/cpu.h> /* rdmsrl, wrmsrl, rdfsbase, wrfsbase */
 #include <x86/cpu_defs.h>
 
-/* TODO: On latest CPUs a faster instruction can be used:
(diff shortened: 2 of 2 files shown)


plat/common/include/x86/cpu.h
```diff
@@ -97,6 +97,21 @@ static inline __u64 rdtsc(void)
 	return (h << 32) | l;
 }
 
+#ifdef __FSGSBASE__
+static inline __u64 rdfsbase(void)
+{
+	__u64 ret;
+
+	__asm__ __volatile__ ("rdfsbase %%rax" : "=a" (ret));
```
jschlumpp, line 105: This could also use the compiler builtins `__builtin_ia32_rdfsbase64` and `__builtin_ia32_wrfsbase64`.

jschlumpp (commented):
I saw the message on discord and the performance improvement is pretty impressive. :+1: 

I also checked whether we could use `-march=x86-64-v3` to enable this by default, but I think the compiler support for these levels are still a bit too fresh (GCC 11/LLVM 12).

plat/common/include/x86/cpu.h
andreittr: Yeah, good point; or maybe even intrinsics, that way it's more compiler-agnostic; patch incoming.

edit: nvm, intrinsics are a pain to use without pulling in all of the std headers :/ I'll have a look at what's the most portable way.

i-Pear: It seems that simply adding "-mfsgsbase" in the compilation parameters is sufficient, without the need to include the header file?
Also, "-mfsgsbase" is provided since gcc6, which is not as fresh as "-march=x86-64-v3".

jschlumpp: Yes, for the builtins (for example `__builtin_ia32_rdfsbase64` the parameter is enough). There are also intrinsics in `immintrin.h` (for example `_readfsbase_u64`). But the builtins are documented anyway and shouldn't change.

The feature levels are recent, yes. They are only a shortcut to enable instruction set extensions for a "generation" of processors. Of course, enabling these individually is an option, but the resulting code will still only work on Ivy Bridge generation-ish processors and using the feature level would also enable all the other nice to have instructions (AVX, movbe, etc.).

skuenzer: Do these `__builtin_ia32_rdfsbase64` depend on to compile and optimize for a certain CPU? IT should be better independent so that we can still run on a wider range of x86 CPU generations. The right way is probably cpuid and skipping cpuid if we configured Unikraft to compile for a recent CPU generation (Architecture options), right?

mogasergiu:
Closing since it is revived by #1316 @mschlumpp @skuenzer I assigned you to it

PR #863, opened 2023-04-22
arch/x86: Add sanitization function for ECTX memory & use it in the syscall shim handler

andreittr:
### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): N/A
 - Application(s): app-elfloader

### Additional configuration

 - `CONFIG_MARCH_X86_64_COREI7AVX=y` or any x86 arch that supports AVX

### Description of changes

Most context saving instructions/routines overwrite the target area with a valid representation that can be loaded back. x86 XSAVE/XSAVEOPT are an exception, as garbage data can remain in the context area causing the following XRSTOR to fault.
Thus any potentially dirty memory passed to XSAVE* must be (minimally) sanitized before its first use.
This change introduces `ukarch_ectx_sanitize` as a function to perform this sanitization if needed.
It differs from `ukarch_ectx_init` in that it does only the minimum work necessary to make memory ready for `ukarch_ectx_store`, instead of preparing a valid argument for `ukarch_ectx_load`.

`ukplat_syscall_handler` is changed to add this sanitization call before storing ECTX, as the memory area is allocated uninitialized on the stack. Another option would be to zero out the entire area, but that would add unacceptable overhead to the syscall hot path, especially for ectx methods which do not require it (x86 FSAVE & FXSAVE, arm & arm64).

### DIFF
--- a/arch/arm/ectx.c
+++ b/arch/arm/ectx.c
@@ -60,6 +60,12 @@ __sz ukarch_ectx_align(void)
 	return ECTX_ALIGN;
 }
 
+void ukarch_ectx_sanitize(struct ukarch_ectx *state __maybe_unused)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, ECTX_ALIGN));
+}
+
 void ukarch_ectx_init(struct ukarch_ectx *state)
 {
 	UK_ASSERT(state);
--- a/arch/x86/ectx.c
+++ b/arch/x86/ectx.c
@@ -110,6 +110,30 @@ __sz ukarch_ectx_align(void)
 	return ectx_align;
 }
 
+void ukarch_ectx_sanitize(struct ukarch_ectx *state)
+{
+	UK_ASSERT(ectx_align); /* Do not call when not yet initialized */
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, ectx_align));
+
+	switch (ectx_method) {
+	case X86_SAVE_XSAVE:
+	case X86_SAVE_XSAVEOPT:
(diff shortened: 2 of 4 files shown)


arch/x86/ectx.c
```diff
@@ -110,6 +110,30 @@ __sz ukarch_ectx_align(void)
 	return ectx_align;
 }
 
+void ukarch_ectx_sanitize(struct ukarch_ectx *state)
+{
+	UK_ASSERT(ectx_align); /* Do not call when not yet initialized */
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, ectx_align));
```
mogasergiu, line 117: Hmm, strange, I would have expected `checkpatch.pl` to issue a warning regarding whitespace between the cast and the variable itself, but the entire file seems to have it at every cast inside `IS_ALIGNED`.

arch/arm/ectx.c
```diff
@@ -60,6 +60,12 @@ __sz ukarch_ectx_align(void)
 	return ECTX_ALIGN;
 }
 
+void ukarch_ectx_sanitize(struct ukarch_ectx *state __maybe_unused)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, ECTX_ALIGN));
```
mogasergiu, line 66: Ditto.

mogasergiu (approved):
Looks good otherwise.

arch/x86/ectx.c
skuenzer: This whitespace in front of the casted variable is actually correct.

skuenzer (approved):
From my side, this PR looks good.

PR #866, opened 2023-04-25
lib/vfscore: Change vnops to take constant strings

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The current vnode operations define strings as mutable, which makes it more difficult to do optimizations. Since no buffer size is supplied the strings can only be modified on a char-for-char basis anyways, which also limits the use-cases of their mutability. In addition, no implementation of the vnode operations makes use of the it.

This PR thus changes all vnode operations to take constant strings. The PR further cleans up `sys_symlink` and removes an unnecessary path copy.

### DIFF
--- a/lib/9pfs/9pfs_vnops.c
+++ b/lib/9pfs/9pfs_vnops.c
@@ -246,7 +246,8 @@ static int uk_9pfs_close(struct vnode *vn __unused, struct vfscore_file *file)
 	return 0;
 }
 
-static int uk_9pfs_lookup(struct vnode *dvp, char *name, struct vnode **vpp)
+static int uk_9pfs_lookup(struct vnode *dvp, const char *name,
+			  struct vnode **vpp)
 {
 	struct uk_9pfs_mount_data *md = UK_9PFS_MD(dvp->v_mount);
 	struct uk_9pdev *dev = md->dev;
@@ -363,7 +364,8 @@ static int uk_9pfs_inactive(struct vnode *vp)
 	return 0;
 }
 
-static int uk_9pfs_create_generic(struct vnode *dvp, char *name, mode_t mode)
+static int uk_9pfs_create_generic(struct vnode *dvp, const char *name,
+				  mode_t mode)
 {
 	struct uk_9pdev *dev = UK_9PFS_MD(dvp->v_mount)->dev;
 	struct uk_9pfid *fid;
@@ -382,7 +384,7 @@ static int uk_9pfs_create_generic(struct vnode *dvp, char *name, mode_t mode)
 	return -rc;
 }
 
-static int uk_9pfs_create(struct vnode *dvp, char *name, mode_t mode)
+static int uk_9pfs_create(struct vnode *dvp, const char *name, mode_t mode)
 {
 	struct uk_9pfs_mount_data *md = UK_9PFS_MD(dvp->v_mount);
 
@@ -412,7 +414,7 @@ static int uk_9pfs_remove_generic(struct vnode *dvp, struct vnode *vp)
 }
 
 static int uk_9pfs_remove(struct vnode *dvp, struct vnode *vp,
-		char *name __unused)
+			  const char *name __unused)
 {
 	struct uk_9pfs_node_data *nd = UK_9PFS_ND(vp);
 	int rc = 0;
@@ -425,7 +427,7 @@ static int uk_9pfs_remove(struct vnode *dvp, struct vnode *vp,
 	return rc;
 }
 
-static int uk_9pfs_mkdir(struct vnode *dvp, char *name, mode_t mode)
+static int uk_9pfs_mkdir(struct vnode *dvp, const char *name, mode_t mode)
 {
 	if (!S_ISDIR(mode))
 		return EINVAL;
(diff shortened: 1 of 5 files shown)


lib/vfscore/syscalls.c
```diff
@@ -859,63 +859,48 @@ sys_rename(char *src, char *dest)
...
 	/* newpath should not already exist */
-	if (namei_last_nofollow(np, newdirdp, &newdp) == 0) {
+	if (unlikely(namei_last_nofollow(np, newdirdp, &newdp) == 0)) {
 		drele(newdp);
 		error = EEXIST;
-		goto out;
+		goto out_unlock;
 	}
 
+	UK_ASSERT(newdp == NULL);
```
mogasergiu: Nitpick, `checkpatch.pl` reports a `CHECK: Comparison to NULL could be written "!newdp"` on my end. Same thing, but perhaps worth considering. Does the CI/CD not run `checkpatch.pl` with the `--strict` flag?

lib/9pfs/9pfs_vnops.c
```diff
@@ -434,7 +436,7 @@ static int uk_9pfs_mkdir(struct vnode *dvp, char *name, mode_t mode)
 }
 
 static int uk_9pfs_rmdir(struct vnode *dvp, struct vnode *vp,
-		char *name __unused)
+			 const char *name __unused)
```
mogasergiu, line 439: Perhaps not related to this PR, but a lot of the `vfscore` code seems to be suffering from this argument misaligned placement in the function signature...

lib/vfscore/syscalls.c
```diff
@@ -859,63 +859,48 @@ sys_rename(char *src, char *dest)
...
-	char		*name;
-
-	if (oldpath == NULL || newpath == NULL) {
-		return (EFAULT);
-	}
+	int error;
+	char np[PATH_MAX];
+	struct dentry *newdp = NULL;
+	struct dentry *newdirdp = NULL;
+	char *name;
```
mogasergiu: Again, maybe just my preference, but now that there seems to be some sort of refactoring process on-going I guess, I think everything would look more beautiful if local variable declarations would be done in a reverse xmas tree order w.r.t. declaration line length. But just my preference. Another very helpful thing, I believe, would be adding the include files in alphabetical order, as that seems to be the consensus on some other large OSS projects, though this is not necessarily related to this PR :).

lib/9pfs/9pfs_vnops.c
```diff
@@ -246,7 +246,8 @@ static int uk_9pfs_close(struct vnode *vn __unused, struct vfscore_file *file)
 	return 0;
 }
 
-static int uk_9pfs_lookup(struct vnode *dvp, char *name, struct vnode **vpp)
+static int uk_9pfs_lookup(struct vnode *dvp, const char *name,
```
mogasergiu, line 249: I may be wrong, but would it be worth to consider writing this as a `const char *const name`? Many of the function signatures changed in this PR seem to be able to benefit from that additional `const`.

mogasergiu (commented):
Looks good to me. See my other comments.

lib/9pfs/9pfs_vnops.c
marcrittinghaus: The vfscore code does suffer from a lot of issues, alignment being more on the harmless side of things 😄 For all bug fixes I do I also try to cleanup some stuff. And with more PRs coming in also from others, I hope that at some point the majority of the codebase has experienced some update/cleanup 🙏🏻

lib/vfscore/syscalls.c
marcrittinghaus: Yes, I like reverse xmas also better. Changed it.

lib/vfscore/syscalls.c
marcrittinghaus: Thanks, changed it! 

No it is not using `--strict`.

lib/9pfs/9pfs_vnops.c
marcrittinghaus: What optimizations do you have in mind? Can you elaborate? 

I assume that the compiler is intelligent enough to see that a variable is not changed within a certain scope. I am asking because `const char * const` is not used frequently. From the coding practice, I agree that explicitly removing mutability where it's not needed is a good idea. However, we are using this nowhere else in the codebase. It is also a bit bulky.

mogasergiu: > What optimizations do you have in mind? Can you elaborate?
> 
> I assume that the compiler is intelligent enough to see that a variable is not changed within a certain scope. I am asking because `const char * const` is not used frequently. From the coding practice, I agree that explicitly removing mutability where it's not needed is a good idea. However, we are using this nowhere else in the codebase. It is also a bit bulky.

Nope, no optimizations in mind. Was just wondering if there could be a benefit to having more verbosity in such cases. I can agree that generally this is useless and only slightly may help readablity. But, you're right, this is not done anywhere else and needlessly increases the type length.

mogasergiu (approved):

razvand (approved):

PR #869, opened 2023-04-28
plat/common/x86: Copy AP bootcode when num cores is exceeded

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `common`
 - Application(s): N/A

### Description of changes

Copy the AP bootcode when a higher number of cores than the one specified in CONFIG_UKPLAT_LCPU_MAXCOUNT is detected, to allow existing cores to boot up.

### DIFF
--- a/plat/common/x86/lcpu.c
+++ b/plat/common/x86/lcpu.c
@@ -172,7 +172,7 @@ int lcpu_arch_mp_init(void *arg __unused)
 			 * just stop here.
 			 */
 			uk_pr_warn("Maximum number of cores exceeded.\n");
-			return 0;
+			break;
 		}
 	}
 	UK_ASSERT(bsp_found);


marcrittinghaus (changes requested):
Hey @michpappas,

I was just checking the patch. Looks simple enough 😄

However, I actually think there is a problem. When we leave the loop early, we might not have found the BSP. With your patch, we will hit the assert and crash. So either we remove the `bsp_found` variable altogether or we need to make the assert dependent also on if we reached the max number of CPUs.

In the latter case, it could look like this:
```C
	struct lcpu *lcpu = lcpu_get(0);
	__lcpuid bsp_cpu_id = lcpu->id;
	__lcpuid cpu_id;

	...
	for (...) {
		...
	}
	UK_ASSERT(bsp_found || !lcpu);
```

michpappas:
> However, I actually think there is a problem. When we leave the loop early, we might not have found the BSP. 

You're right. In fact, we may also miss other MADT entries for APICs and the like. I think the right way to fix that is to replace `break` with `continue` to let it iterate through all entries, what do you think?

> So either we remove the bsp_found variable altogether

I doubt that assert it will ever be hit (famous last words), but it's good practice to keep it there.

> or we need to make the assert dependent also on if we reached the max number of CPUs.

I think it's okay to handle this gracefully as we do now, as it gives some flexibility on deploying the same image to different hosts.

marcrittinghaus:
> You're right. In fact, we may also miss other MADT entries for APICs and the like. I think the right way to fix that is to replace `break` with `continue` to let it iterate through all entries, what do you think?

Since we are just iterating to enumerate CPUs, I think it is fine to `break`, isn't it?

> I doubt that assert it will ever be hit (famous last words), but it's good practice to keep it there.
> I think it's okay to handle this gracefully as we do now, as it gives some flexibility on deploying the same image to different hosts.

Agree 😄

razvanvirtan (changes requested):
Thanks,@michpappas!
I agree with Marc that skipping some MADT entries in this case won't have any functional effect, so I think updating the assert condition [here](https://github.com/unikraft/unikraft/pull/869/files#diff-efa03d78316bf5ab020327794d6f2133ba19162958c3ab72b8d1aea1af117575L178) should be enough. Once this is done I can approve 👍

mogasergiu:
The assertion should not be removed IMO, it is specified by the ACPI standard that the BSP must always come first. If we are dealing with uncompliant firmware, which is not unlikely in a real world baremetal scenario (although it is a bit far-fetched, kind of a hard to make mistake), I would rather we knew about it from the very beginning. I like the second part though `UK_ASSERT(bsp_found || !lcpu);`, provided `lcpu` is initialized to NULL at its declaration and, given what has been previously mentioned, it is a `&&` instead and `lcpu` is tested for `NULL` (MADT should never be empty). I would put it in a second, separate, assertion though, as the crash message would show both `bsp_found && lcpu` not letting you know which of the checks failed.

PR #870, opened 2023-04-28
plat/common: Introduce ukplat_pt_init_clone

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

The existing ukplat_pt_clone function allows cloning an existing pagetable, with both pagetables using the same frame allocator.
    
Introduce ukplat_pt_init_clone to allow initializing a new pagetable with a dedicated allocator, as a clone of an existing one.

### DIFF
--- a/include/uk/plat/paging.h
+++ b/include/uk/plat/paging.h
@@ -85,7 +85,8 @@ struct uk_pagetable {
 	(((flag) >> PAGE_FLAG_SIZE_SHIFT) & PAGE_FLAG_SIZE_MASK)
 
 /* Page table clone flags */
-#define PAGE_FLAG_CLONE_NEW	0x01 /* Create an empty page table */
+#define PAGE_FLAG_CLONE_NEW		0x01 /* Create an empty page table */
+#define PAGE_FLAG_CLONE_INITIALIZED	0x02 /* PT is already initialized */
 
 /**
  * Returns the active page table (the one that defines the virtual address
@@ -166,6 +167,30 @@ int ukplat_pt_add_mem(struct uk_pagetable *pt, __paddr_t start, __sz len);
  */
 int ukplat_pt_clone(struct uk_pagetable *pt, struct uk_pagetable *pt_src,
 		    unsigned long flags);
+
+/**
+ * Initializes a new page table as a clone of an existing pagetable. The
+ * new pagetable is assigned physical address range for allocations and
+ * mappings.
+ *
+ * @param pt_dst
+ *   An uninitialized page table that will receive a clone of the source page
+ *   table
+ * @param pt_src
+ *   The source page table that will be cloned
+ * @param start
+ *   Start of the physical address range that will be available
+ *   for allocations of physical memory (e.g., mapping with __PADDR_ANY).
+ *   The function may reserve some memory in this area for own purposes. The
+ *   range must not be assigned to other page tables.
+ * @param len
+ *   The length (in bytes) of the physical address range
+ *
+ * @return
+ *   0 on success, a non-zero value otherwise
+ */
+int ukplat_pt_init_clone(struct uk_pagetable *pt_dst, struct uk_pagetable *pt_src,
+			 __paddr_t start, __sz len);
 
 /**
  * Frees the given page table by recursively releasing the page table hierarchy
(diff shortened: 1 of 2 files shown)


michpappas:
@marcrittinghaus I am not entirely happy with the current approach. An alternative approach would be to change the default behavior of `ukplat_pt_clone()` to *not* use the same frame allocator by default, and make that behavior conditional to a flag. Moreover add the `start` and `len`. The prototype then would become:

```C
/**
 * Initializes a new page table as clone of another page table.
 *
 * @param pt_dst
 *   An uninitialized page table that will receive a clone of the source page
 *   table
 * @param pt_src
 *   The source page table that will be cloned
 * @param flags
 *   Clone flags (PAGE_FLAG_CLONE_*)
 *
 *   If PAGE_FLAG_CLONE_NEW is specified, a new (empty) top-level page table is
 *   created. Note that this page table will be completely empty and thus do
 *   not map any code or data segments of the kernel.
 *  If PAGE_FLAG_CLONE_SHARE_ALLOC is specified, the new page
 *  table will be sharing the physical address range available for new allocations and
 *  mappings with the source page table.
 * @return
 *   0 on success, a non-zero value otherwise
 */
int ukplat_pt_clone(struct uk_pagetable *pt_dst, struct uk_pagetable *pt_src,
		      __paddr_t start, __sz len, unsigned long flags);
```

marcrittinghaus:
> ```
> int ukplat_pt_clone(struct uk_pagetable *pt_dst, struct uk_pagetable *pt_src,
> 		      __paddr_t start, __sz len, unsigned long flags);
> ```

Would you then specify `0, 0` in case you set `PAGE_FLAG_CLONE_SHARE_FALLOC`?

To add to this: First step would be to create a new frame allocator for the provided range and then use this in the internal `pt_clone` function, right?

michpappas:
> Would you then specify 0, 0 in case you set PAGE_FLAG_CLONE_SHARE_FALLOC?

Correct, although I didn't document that. If we go that way I will though.

> To add to this: First step would be to create a new frame allocator for the provided range and then use this in the internal pt_clone function, right?

Yes, just how it's done in the current state of the patch:

```
+       memset(pt_dst, 0, sizeof(struct uk_pagetable));
+
+       rc = pgarch_pt_init(pt_dst, start, len);
+       if (unlikely(rc))
+               return rc;
```

Another option would had been to initialize the pagetable (and its frame allocator) outside `ukplat_pt_clone()`, and add a flag to `ukplat_pt_clone()` that indicates that the passed pagetable is already initialized. But that would require to expose `pgarch_pt_init()` to the public API, which I wanted to avoid.

michpappas:
> Would you then specify 0, 0 in case you set PAGE_FLAG_CLONE_SHARE_FALLOC?

Correct, although I didn't document that. If we go that way I will though.

> To add to this: First step would be to create a new frame allocator for the provided range and then use this in the internal pt_clone function, right?

Yes, just how it's done in the current state of the patch:

```C
+       memset(pt_dst, 0, sizeof(struct uk_pagetable));
+
+       rc = pgarch_pt_init(pt_dst, start, len);
+       if (unlikely(rc))
+               return rc;
```

Another option would had been to initialize the pagetable (and its frame allocator) outside `ukplat_pt_clone()`, and add a flag to `ukplat_pt_clone()` that indicates that the passed pagetable is already initialized. But that would require to expose `pgarch_pt_init()` to the public API, which I wanted to avoid.

craciunoiuc (approved):
Your favourite PR-Pusher says: All good here. Thanks!

Edit: I'm out of context with the discussion, but if the bot wants me to review, I review.

marcrittinghaus:
> Another option would had been to initialize the pagetable (and its frame allocator) outside `ukplat_pt_clone()`, and add a flag to `ukplat_pt_clone()` that indicates that the passed pagetable is already initialized. But that would require to expose `pgarch_pt_init()` to the public API, which I wanted to avoid.

Yes, I agree. That I would have not liked either. I think your alternative approach of changing `ukplat_pt_clone()` to include a memory range feels more natural. 

Actually, I wonder if we should merge `ukplat_pt_init()` and `ukplat_pt_clone()` into one function then. I would really like that, because it would be more elegant to call one existing function with a certain combination of parameters than to have one special function that you should only call once during initialization.

So we could add another flag that instructions to hijack the currently configured HW PT. This would of course implicitly mean you acknowledge that the page table frames must not/cannot be released. Just like it is with the static PT when you call `ukplat_pt_init()`.

@michpappas What do you think?

rares-miculescu:
Looks good, but shouldn't it be rebased to the latest changes?

PR #874, opened 2023-05-01
lib/vfscore: Handle `FIONBIO` command in `pipe_ioctl()`

eduardvintila:
### Prerequisite checklist

### Base target

### Description of changes

The current `ioctl` implementation for pipes does not handle the `FIONBIO` command, which leads to a negative error value being returned. This PR introduces a simple case which sets the FNONBLOCK flag when such a command is passed.

Without this PR, `app-redis` won't respond to any PING requests, for example, since Redis relies on nonblocking pipes to properly work.

### DIFF
--- a/lib/vfscore/pipe.c
+++ b/lib/vfscore/pipe.c
@@ -462,6 +462,9 @@ static int pipe_ioctl(struct vnode *vnode,
 		*((int *) data) = pipe_buf_get_available(pipe_buf);
 		uk_mutex_unlock(&pipe_buf->rdlock);
 		return 0;
+	case FIONBIO:
+		/* sys_ioctl() already sets f_flags, no need to do anything */
+		return 0;
 	default:
 		return -EINVAL;
 	}


dinhngtu (commented):
Why was `pipe_file->flags` set instead of `vfscore_file->f_flags` as checked by `pipe_read` (see below)? I'm not quite clear on the purpose of `pipe_file->flags` either, it seems to be unused until this PR.

https://github.com/unikraft/unikraft/blob/5cfaf423c04d0f00b16d5726849bca5fd8187606/lib/vfscore/pipe.c#L364

Starnox:
> Why was `pipe_file->flags` set instead of `vfscore_file->f_flags` as checked by `pipe_read` (see below)? I'm not quite clear on the purpose of `pipe_file->flags` either, it seems to be unused until this PR.
> 
> https://github.com/unikraft/unikraft/blob/5cfaf423c04d0f00b16d5726849bca5fd8187606/lib/vfscore/pipe.c#L364

I have the same question, as I just started looking at the library. If you could provide a little bit more details, that would be great, thanks!

eduardvintila:
> I'm not quite clear on the purpose of `pipe_file->flags` either, it seems to be unused until this PR.

Hey, @dinhngtu @Starnox! Indeed, `pipe_file->flags` isn't used anywhere, just yet. I'm not sure of its purpose, either.  I considered setting it because maybe in the future it will be used when extending/updating/refactoring our current internal `pipes` implementation.

> Why was `pipe_file->flags` set instead of `vfscore_file->f_flags` as checked by `pipe_read` (see below)? 

All `ioctl` calls are first routed through the generic [`sys_ioctl()`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/syscalls.c#L382) function, which already sets the `vfscore_file`'s `f_flags` member with the [desired flags](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/syscalls.c#L405), before calling the actual implementation for the given file (which, for pipes, is `pipe_ioctl()`). 

Therefore, in our case, we don't need to explicitly set `vfscore_file->f_flags`, since the aforementioned function already does that before `pipe_ioctl()` is called.

dinhngtu (changes requested):
Per [this comment](https://github.com/unikraft/unikraft/blob/e1604e97a54c1195368463c0c1f2ccb726dac6d1/lib/vfscore/main.c#L2057-L2065), the plan going forward is to consume `f_flags` directly rather than syncing it with our own flags whenever possible. We already do this in our pipes implementation so IMO we should leave `pipe_file->flags` alone and rewrite this as:

```c
case FIONBIO:
    /* sys_ioctl() already sets f_flags, no need to do anything */
    return 0;
```

eduardvintila:
Good point, @dinhngtu! I've updated the PR to reflect this 👍🏻

dinhngtu (approved):
That's all from me, thanks.

Starnox (approved):
Looks good to me! Thanks!

Reviewed-by: Eduard-Florin Mihailescu <mihailescu.eduard@gmail.com>

EDIT by @marcrittinghaus: Fixed email format in tag.

razvand (approved):

PR #875, opened 2023-05-05
lib/vfscore: `fs0` for automounting 9pfs

skuenzer:
### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [all]

### Additional configuration

 - `LIBVFSCORE_AUTOMOUNT_ROOTFS=y`
 - `LIBVFSCORE_ROOTFS_9PFS=y`

### Description of changes

This PR changes the default device/tag name for an automatically mounted 9pfs root volume from `rootfs` to `fs0`. This is done for convenience because `fs0` is the tag name that the `qemu-guest` support script uses to name the first 9pfs share.

### DIFF
--- a/lib/vfscore/Config.uk
+++ b/lib/vfscore/Config.uk
@@ -74,7 +74,7 @@ if LIBVFSCORE_AUTOMOUNT_ROOTFS
 	config LIBVFSCORE_ROOTDEV
 	string "Default root device"
 	depends on !LIBVFSCORE_ROOTFS_RAMFS && !LIBVFSCORE_ROOTFS_INITRD
-	default "rootfs" if LIBVFSCORE_ROOTFS_9PFS
+	default "fs0" if LIBVFSCORE_ROOTFS_9PFS
 	default ""
 	help
 		Device to mount the filesystem from (e.g., on 9PFS this


razvand:
This is OK with me. It's about time we used consistent naming for filesystem mounts. `fs0` is to be used throughout.

@mariasfiraiala, @StefanJum, see if it's OK, review it and approve it.

StefanJum (commented):
The perfect pr :partying_face: 
I would change the commit message a little to be more clear
`lib/vfscore: Use fs0 for 9pfs default device name` or something like that.

marcrittinghaus:
Coincidentally, there has been a discussion with @antoineco and @nderjung a few days ago about this and they were asking for `rootfs` because they find it more intuitive than `fs0`. 

I do not have a strong preference as I never found `fs0` to be misleading and I actually use `fs1` and higher for additional volumes. So it makes it easy to organize, from my perspective.

antoineco:
For the case where we take a container filesystem and mount it directly (in binary compatibility mode for instance) I think `rootfs` feels more familiar. But if there can be multiple ones `fs[0-9]` sounds like a sane default. We can always name it differently (even abstract it away) in higher level tools like KraftKit, which is the level we were discussing with Alex.

razvand:
@marcrittinghaus , @antoineco , @nderjung , @skuenzer , we will go for `fs0`.

@mariasfiraiala , @StefanJum , you can mark this as approved, so we can merge it.

mariasfiraiala (approved):
Except for @StefanJum's comment, all good on my side.

skuenzer:
> 

I applied this suggestion, thanks a lot.

StefanJum (approved):
All good, thanks.

skuenzer:
> Coincidentally, there has been a discussion with @antoineco and @nderjung a few days ago about this and they were asking for `rootfs` because they find it more intuitive than `fs0`.
> 
> I do not have a strong preference as I never found `fs0` to be misleading and I actually use `fs1` and higher for additional volumes. So it makes it easy to organize, from my perspective.

For completeness, I'd like to refer to the discussion we had on [Discord](https://discord.com/channels/762976922531528725/859811422683004985/1104147714540851330).
The main arguments for the change are basically:
- Change for easier onboarding of new people using `qemu-guest` and simpler HowTo's. Changing the default tag name can be an obstacle to get started with Unikraft development.
- `qemu-guest` does not allow to specify the tag name and such an option would make the tool more complicated to use. Tag names are auto-generated: `fs0`, `fs1`, and so on. For `qemu-guest`, the passed paths are of no meaning and the tool should not make assumptions on those. If `qemu-guest` would handover the first share as `rootfs`, might be unexpected for `qemu-guest` users.
- The change only affects the default automount configuration value. It can be still changed with `menuconfig` or even at runtime via a kernel parameter (`uklibparam`).

Overall this PR is all about improving user experience.

razvand (approved):

PR #881, opened 2023-05-08
build: Disable compiler control flow protection

andreittr:
### Description of changes
Compiler flag -fcf-protection sets control-flow integrity measures that improve security, and sometimes defaults to enabled on distro compilers. Some aspects of this feature are implemented in the compiler header <cet.h> for e.g., stack unwinders. If this header is missing some programs fail to compile.
Since we don't (yet) support pulling compiler headers into unikraft build, this change disables cf-protection globally.

This change fixes missing <cet.h> errors for system compilers that default cf-protection to on.

This is part of a patch set consisting of:
 - [libcxx](https://github.com/unikraft/lib-libcxx/pull/28)
 - [libcxxabi](https://github.com/unikraft/lib-libcxxabi/pull/4)
 - [lib-compiler-rt](https://github.com/unikraft/lib-compiler-rt/pull/12)
 - [libunwind](https://github.com/unikraft/lib-libunwind/pull/7)
 - [musl](https://github.com/unikraft/lib-musl/pull/45)
 - [unikraft/build](https://github.com/unikraft/unikraft/pull/881)

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

System compiler that defaults cf-protection to on (check with `cc -E -dM - < /dev/null | grep CET`).

### DIFF
--- a/Makefile.uk
+++ b/Makefile.uk
@@ -11,6 +11,8 @@ COMPFLAGS-$(call have_gcc)	+= -fno-tree-sra -fno-split-stack -nostdinc
 ifneq ($(HAVE_STACKPROTECTOR),y)
 COMPFLAGS    += -fno-stack-protector
 endif
+COMPFLAGS    += -fcf-protection=none
+
 COMPFLAGS    += -Wall -Wextra
 COMPFLAGS-$(call have_clang)	+= -Wdocumentation -Wdocumentation-pedantic
 


Makefile.uk
```diff
@@ -11,6 +11,8 @@ COMPFLAGS-$(call have_gcc)	+= -fno-tree-sra -fno-split-stack -nostdinc
 ifneq ($(HAVE_STACKPROTECTOR),y)
 COMPFLAGS    += -fno-stack-protector
 endif
+COMPFLAGS    += -fcf-protection=none
```
mariasfiraiala, line 14:
```suggestion
ifneq ($(CONFIG_OPTIMIZE_CFC),y)
COMPFLAGS    += -fcf-protection=none
endif
```

Wouldn't it be better if we added a config option for this? And have the C++ libraries enable that flag automatically? 

Basically, we would have the `-fcf-protection=none` missing for all apps/libs, besides cpp ones, which will disable the optimization.

mariasfiraiala (changes requested):
Thanks @andreittr! This works, however I have a few kind of pedantic comments before merging this. You'll find the suggestion below.

Makefile.uk
andreittr: I agree that cf-protection should eventually become a Kconfig build option and have smth like this in the Makefile
```
ifeq ($(CONFIG_OPTIMIZE_CFC),y)
COMPFLAGS += -fcf-protection=full
else
COMPFLAGS += -fcf-protection=none
endif
```
(although I'd rather not use "OPTIMIZE" for build flags that do security hardening => "CONFIG_HARDEN_CFC" or similar)

However:
- `-fcf-protection` sets the `__CET__` macro which any code can check for. This is not limited to C++ unwinders, but arbitrary code that is aware of control flow protections (e.g. off the top of my head: JIT compilers that need to know to emit `endbr64` at the beginning of generated functions).
- if `__CET__` is defined, the header `<cet.h>` is expected to be provided. Unfortunately at this time unikraft cannot provide compiler headers in its include path (and providing them without also pulling in system includes is not exactly trivial) so any CET-aware code will fail to compile.
- even if we somehow provide `<cet.h>`, since CET is a hardware feature (well, two features: SHSTK and IBT), proper use of it requires explicit support code in the "kernel" (i.e., unikraft core) in addition to compiler flags. We do not yet have this code so any incidental use of cf-protection does not actually add to security (and might even break things if e.g., CET-aware code tries to access a non-existent shadow stack).

Due to these issues I consider that force-disabling cf-protection is the right call at this time, with proper implementation in the form of (1) a user-selectable Kconfig option, (2) support code in unikraft core, and (3) cet.h to come in the future.

mariasfiraiala (approved):
Alright, then, thanks for the clarification @andreittr! All good on my side now.

Makefile.uk
razvand: @mariasfiraiala, please add an issue that points out the issue you raised and @andreittr's input for a future solution with the 3 steps above.

razvand:
@andreittr, @mariasfiraiala, does this PR break anything? I.e. should we include it part of release 0.13, or should we leave it for the "after-the-release inclusion" together with the other PRs:

 - [libcxx](https://github.com/unikraft/lib-libcxx/pull/28)
 - [libcxxabi](https://github.com/unikraft/lib-libcxxabi/pull/4)
 - [lib-compiler-rt](https://github.com/unikraft/lib-compiler-rt/pull/12)
 - [libunwind](https://github.com/unikraft/lib-libunwind/pull/7)
 - [musl](https://github.com/unikraft/lib-musl/pull/45)

mariasfiraiala:
> @andreittr, @mariasfiraiala, does this PR break anything?

Not on my end. And it makes sense to have this merged with the 0.13 release, as it does make sense globally, without relating it to the C++17 series of patches.

RaduNichita (approved):
All good also from my side.

razvand (approved):

PR #885, opened 2023-05-11
lib/posix-process: Start tid numbering from one

flpostolache:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): `N/A`

### Description of changes

These commits make so that PID/TID 0 cannot be assigned to any thread. Some library calls can fail when using binary compatibility mode because they assume that this value is reserved by the kernel.

### DIFF
--- a/lib/posix-process/Config.uk
+++ b/lib/posix-process/Config.uk
@@ -12,8 +12,8 @@ if LIBPOSIX_PROCESS
 if LIBPOSIX_PROCESS_PIDS
 		config LIBPOSIX_PROCESS_MAX_PID
 		int "Largest PID"
-		range 1 32768
-		default 32
+		range 1 32767
+		default 31
 
 		config LIBPOSIX_PROCESS_INIT_PIDS
 		bool "Assign PID during boot"
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -83,9 +83,11 @@ struct posix_thread {
 
 /**
  * System global lists
+ * NOTE: We pre-allocate PID/TID 0 which is reserved by the kernel.
+ *       An application should never get PID/TID 0 assigned.
  */
 static struct posix_thread *tid_thread[TIDMAP_SIZE];
-static unsigned long tid_map[UK_BITS_TO_LONGS(TIDMAP_SIZE)];
+static unsigned long tid_map[UK_BITS_TO_LONGS(TIDMAP_SIZE)] = { [0] = 0x01UL };
 
 /**
  * Thread-local posix_thread reference
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -328,8 +328,7 @@ static void pprocess_release(struct posix_process *pprocess)
 
 static void pprocess_kill(struct posix_process *pprocess)
 {
-	struct posix_thread *pthread, *pthreadn;
-	bool self_destruct = false;
+	struct posix_thread *pthread, *pthreadn, *pthread_self = NULL;
 
 	/* Kill all remaining threads of the process */
 	uk_list_for_each_entry_safe(pthread, pthreadn,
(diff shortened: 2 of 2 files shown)


lib/posix-process/process.c
```diff
@@ -40,6 +40,8 @@
 #include <uk/config.h>
 #include <uk/syscall.h>
 
+#define PIDMAP_SIZE (CONFIG_LIBPOSIX_PROCESS_MAX_PID + 1)
```
skuenzer: Maybe calling it `TIDMAP_SIZE` makes more sense in this source file. I am sorry, we spoke offline about it and the suggestion came from me.

lib/posix-process/Config.uk
```diff
@@ -14,6 +14,8 @@ if LIBPOSIX_PROCESS_PIDS
 		int "Largest PID"
 		range 1 32768
 		default 32
+		help 
+				This value represents the maximum number of pids that can be created
```
skuenzer: With the changes you did, it is now really the largest PID/TID number that can be created and not the number of PIDs.

lib/posix-process/Config.uk
```diff
@@ -14,6 +14,8 @@ if LIBPOSIX_PROCESS_PIDS
 		int "Largest PID"
 		range 1 32768
 		default 32
```
skuenzer: Maybe set the default to 31 so that the bitmap smoothly fits into 4 `long`s.

lib/posix-process/Config.uk
```diff
@@ -14,6 +14,8 @@ if LIBPOSIX_PROCESS_PIDS
 		int "Largest PID"
 		range 1 32768
```
skuenzer: It may make sense to set the maximum to 32767 in order to fully use the available bitmap in the max case.

lib/posix-process/process.c
```diff
@@ -83,9 +83,10 @@ struct posix_thread {
 
 /**
  * System global lists
+ * NOTE: PID/TID 0 are reserved by the kernel and should never be available
```
skuenzer:
```diff
- * NOTE: PID/TID 0 are reserved by the kernel and should never be available
+ * NOTE: We pre-allocate PID/TID 0 which is reserved by the kernel.
+ *       An application should never get PID/TID 0 assigned.
```

skuenzer (changes requested):
Thanks a lot for this changes. Great!

Can you adopt the commit titles to the following?
`lib/posix-process: Correct MAX_PID config option`
`lib/posix-process: Reserve PID/TID 0 for kernel`

Additionally, a debug message confused me but in the end I think this is correct. If there is no parent, PID `0` is printed at line 283:
```c
uk_pr_debug("Process PID %d created (parent PID: %d)\n",
    (int) pprocess->pid,
    (int) ((pprocess->parent) ? pprocess->parent->pid : 0));
```
However, this is not reflected in the getppid system call (line 532). We should probably also return PID 0 here in the case that there is no parent instead of the own PID. At the moment it makes more sense to me to say this process was directly created by the kernel. Because we reserve PID 0 for the kernel, you can do this change in the same commit as well.

lib/posix-process/Config.uk
flpostolache: By that help string, I meant that, including the start thread, you can create that many threads. With the new changes it is indeed the maximum number that can be created. Because it is easier to understand, I am inclined to completely erase the help message or change it to what you commented. What is your opinion?

skuenzer: Good question, I think because the option says `Largest PID`, we can live without the help text.

skuenzer (approved):
Thanks a lot for your work and fixing this!

andraprs (approved):

Thank you for the updates.

PR #886, opened 2023-05-12
lib/uk9p, lib/9pfs: Fix I/O error on fsync with 9P2000.u

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [QEMU]
 - Application(s): [N/A]

### Description of changes

With commit 6d07017e38fdbdec6037ba6f58daa6670ca7ab74, I introduced support for `fsync` with 9P2000.u. However, a bug in the implementation causes applications to receive an I/O error, whereas the stubbed version would let them continue to run. This issue has been discovered by @StefanJum.

This PR fixes the issue by
1) using the correct `fid` in the 9P request, which is the actual bug
2) introduces a workaround for a bug in QEMU, where QEMU returns a zero-length response for a successful `fsync`.

The bug fix can be tested by running `app-sqlite`. Trying to create a new table with `CREATE TABLE tab (d1 int);` will fail with an I/O error. The operation should complete with the PR applied. This can also by tested by simply calling `fsync` on a file in a simple `main`.

### DIFF
--- a/lib/9pfs/9pfs_vnops.c
+++ b/lib/9pfs/9pfs_vnops.c
@@ -767,11 +767,11 @@ out:
 	return -rc;
 }
 
-static int uk_9pfs_setattr(struct vnode *vp, struct vattr *attr)
+static int uk_9pfs_do_setattr(struct vnode *vp, struct uk_9pfid *fid,
+			      struct vattr *attr)
 {
 	struct uk_9pfs_mount_data *md = UK_9PFS_MD(vp->v_mount);
 	struct uk_9pdev *dev = md->dev;
-	struct uk_9pfid *fid = UK_9PFS_VFID(vp);
 
 	if (md->proto == UK_9P_PROTO_2000L) {
 		uint32_t valid = 0;
@@ -882,6 +882,11 @@ static int uk_9pfs_setattr(struct vnode *vp, struct vattr *attr)
 	}
 }
 
+static int uk_9pfs_setattr(struct vnode *vp, struct vattr *attr)
+{
+	return uk_9pfs_do_setattr(vp, UK_9PFS_VFID(vp), attr);
+}
+
 static int uk_9pfs_fsync(struct vnode *vp, struct vfscore_file *fp)
 {
 	struct uk_9pfs_mount_data *md = UK_9PFS_MD(vp->v_mount);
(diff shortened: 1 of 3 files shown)


lib/9pfs/9pfs_vnops.c
```diff
@@ -900,7 +905,7 @@ static int uk_9pfs_fsync(struct vnode *vp, struct vfscore_file *fp)
 
 static int uk_9pfs_truncate(struct vnode *vp, off_t off)
 {
-	return uk_9pfs_setattr(vp, &(struct vattr){
+	return uk_9pfs_do_setattr(vp, UK_9PFS_VFID(vp), &(struct vattr){
```
StefanJum, line 908: Isn't this just a call to the `uk_9pfs_setattr()`, since the `fid` is `UK_9PFS_VFID(vp)`?

StefanJum (commented):
Works fine, thanks @marcrittinghaus. See the one comment about the wrapper call.

lib/9pfs/9pfs_vnops.c
marcrittinghaus: Yes, you could leave it like that. Question is, if it really makes sense. Since this is a trivial direct call and `uk_9pfs_setattr()` does not do any extra work, you just increase the number of hops (=lower performance). Furthermore, this would create a dependency between `truncate` and `setattr` whereas the actual dependency is between `truncate` and `do_setattr`.

Does that make sense?

StefanJum: Yes, that makes sense.
I don't have a strong opinion on this, I generally just hate having to use wrappers, but when it's needed it's needed.
All looks good then, thank you :rocket:

StefanJum (approved):

lib/9pfs/9pfs_vnops.c
marcrittinghaus: Yes, it would be nice if we could avoid the `_do_` variant altogether. But since `uk_9pfs_setattr` is a vnop that does not work.

eduardvintila (approved):
Thanks for the fix, @marcrittinghaus. Looks very good.

razvand (approved):

PR #476, opened 2022-06-14
Synchronization

Sairajkodilkar:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): x86_64, arm64
 - Platform(s): N/A
 - Application(s): N/A

### Description

The pull request introduces irq save API for the spinlock as well as SMP safe mutex implementation.
Previously for UP arch, the Mutex disabled the interrupt, this implementation also synchronized the wait queue. 
Along with this the previous implementation also used the two conditions to indicate the lock that lock is free, which was
`owner == current || lock_count == 0
`
but if we observe closely we can separate these conditions since if the owner is the current thread then we can increment the lock count, otherwise, we can wait for the current to be null and can acquire the lock. This implementation does not disable interrupts but we can definitely introduce an API that does. 
One requirement of this commit is that we need to separately implement the SMP safe wait queue. currently, these wait queue is only used in uklock lib but SMP safe wait queues can be used anywhere in the kernel.

PR #888, opened 2023-05-12
lib/vfscore: Avoid recursive locking in sys_rename

marcrittinghaus:
If the source and destination directories are the same in sys_rename, the vnode is locked two times. This is problematic with non-recursive mutexes. Also the double checks for access are not needed. This commit adds a check to make sure in this case, only a single lock is done and redundant checks are avoided.

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

If the source and destination directories are the same in sys_rename,
the vnode is locked two times. This is problematic with non-recursive
mutexes. Also the double checks for access are not needed. This commit
adds a check to make sure in this case, only a single lock is done and
redundant checks are avoided.

The issue can be encountered in `app-python3` when applying #476

### DIFF
--- a/lib/vfscore/syscalls.c
+++ b/lib/vfscore/syscalls.c
@@ -839,19 +839,23 @@ sys_rename(char *src, char *dest)
 	vn_lock(dvp1);
 
 	dvp2 = ddp2->d_vnode;
-	vn_lock(dvp2);
+	if (dvp1 != dvp2) {
+		vn_lock(dvp2);
 
-	/* Source and destination directions should be writable) */
-	if ((error = vn_access(dvp1, VWRITE)) != 0)
-	    goto err3;
-	if ((error = vn_access(dvp2, VWRITE)) != 0)
-	    goto err3;
+		/* Source and destination must be on the same file system */
+		if (dvp1->v_mount != dvp2->v_mount) {
+			error = EXDEV;
+			goto err3;
+		}
 
-	/* The source and dest must be same file system */
-	if (dvp1->v_mount != dvp2->v_mount) {
-		error = EXDEV;
-		goto err3;
+		/* Destination directory must be writable */
+		if ((error = vn_access(dvp2, VWRITE)) != 0)
+			goto err3;
 	}
+
+	/* Source directory must be writable */
+	if ((error = vn_access(dvp1, VWRITE)) != 0)
+		goto err3;
 
 	error = VOP_RENAME(dvp1, vp1, sname, dvp2, vp2, dname);
 	if (error)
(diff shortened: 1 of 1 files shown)


lib/vfscore/syscalls.c
```diff
@@ -839,20 +839,23 @@ sys_rename(char *src, char *dest)
...
-	vn_lock(dvp2);
+	if (dvp1 != dvp2) {
+		vn_lock(dvp2);
 
-	/* Source and destination directions should be writable) */
-	if ((error = vn_access(dvp1, VWRITE)) != 0)
-	    goto err3;
-	if ((error = vn_access(dvp2, VWRITE)) != 0)
-	    goto err3;
+		/* The source and dest must be same file system */
```
mariasfiraiala:
```suggestion
		/* The source and destination must be in the same file system */
```

Not sure about the meaning of the previous comment, so I amended it a bit.

lib/vfscore/syscalls.c
```diff
@@ -839,20 +839,23 @@ sys_rename(char *src, char *dest)
...
+		if (dvp1->v_mount != dvp2->v_mount) {
+			error = EXDEV;
+			goto err3;
+		}
 
-	/* The source and dest must be same file system */
-	if (dvp1->v_mount != dvp2->v_mount) {
-		error = EXDEV;
-		goto err3;
+		/* Source and destination directions should be writable) */
```
mariasfiraiala:
```suggestion
		/* Source and destination directions should be writable */
```

Not sure about "directions" here. Is it ok?

mariasfiraiala (changes requested):
Thank you @marcrittinghaus. I've tested this with `app-python3` and #476 and it worked alright. I left some (nit-picky) suggestions regarding the old comments. I'll add my tag, anyway, since, from my pov, the functionality is fine.

lib/vfscore/syscalls.c
StefanJum: I think it's `directories`.

StefanJum (approved):
Besides @mariasfiraiala 's comments everything looks fine to me, thanks.
I'll add the tag after @razvand takes a look.

marcrittinghaus:
> Thank you @marcrittinghaus. I've tested this with `app-python3` and #476 and it worked alright. I left some (nit-picky) suggestions regarding the old comments. I'll add my tag, anyway, since, from my pov, the functionality is fine.
> 
> Reviewed-by: Maria Sfiraiala [maria.sfiraiala@gmail.com](mailto:maria.sfiraiala@gmail.com)

Thanks @mariasfiraiala for pointing this out. That was a good opportunity to fix this. I did not see this 🐮💩. It is fixed 👍🏻 

Thanks @StefanJum for testing 🙏🏻

razvand (commented):
It's OK.

@StefanJum, please add your tag and then I'll do the final approval step.

StefanJum (approved):
All good, go @razvand.

razvand (approved):

PR #893, opened 2023-05-15
lib/ukargparse: Support escaping of quotes

skuenzer:
### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_LIBUKARGPARSE=y`

### Description of changes

This PR introduces the support of escaping single (`'`) and double quotes (`"`) with the escape character `\`. The behavior is similar to what one expects from Unix shells, like `sh` or `bash`.

Single-quoted characters and character sequences preserve each character. Double-quoted characters and character sequences preserve each character except the backslash `\`. The backslash can be used to switch of the special meaning of a character, like the double quote (`\"`) or a backslash (`\\`).

This enables handing over of strings with nested quotes, like:
`"\"my string\""` --`uk_argparse()`--> `"my string"`
This can be especially useful for setting c-string values through `lib/uklibparam`.

### DIFF
--- a/lib/ukargparse/argparse.c
+++ b/lib/ukargparse/argparse.c
@@ -1,9 +1,10 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Authors: Simon Kuenzer <simon.kuenzer@neclab.eu>
+ * Authors: Simon Kuenzer <simon@unikraft.io>
  *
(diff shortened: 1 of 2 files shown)


razvand (approved):

PR #903, opened 2023-05-23
fix/eventfd: O_NONBLOCK

skuenzer:
### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_LIBPOSIX_EVENT=y`

### Description of changes

Commit fc96850 introduced an complete implementation of handling `FNONBLOCK` and `FASYNC` with `fcntl`, including forwarding of error codes to the caller. Because a standard `EINVAL` handler for `ioctl` was used for `eventfd` file descriptors, setting `O_NONBLOCK` caused a failing assertion.


John-Ted (approved):
Both `eventfd` with `EFD_NONBLOCK` and `fcntl` with `O_NONBLOCK` now work, without any assertion fails.

lib/posix-event/eventfd.c
```diff
@@ -292,9 +293,25 @@ static int eventfd_vfscore_poll(struct vnode *vnode, unsigned int *revents,
...
 }
 
+static int eventfd_vfscore_ioctl(struct vnode *vnode __unused,
+				 struct vfscore_file *fp __unused,
+				 unsigned long request, void *buf __unused)
+{
+
+	switch (request) {
+	case FIONBIO:
+		/* eventfd operations look for O_NONBLOCK flag. We just
```
razvand: I would expand this to say that `FIONBIO` is the `ioctl` flag corresponding to the `O_NONBLOCK` flag. Otherwise, it seems weird to mention `O_NONBLOCK` with no reference to `FIONBIO`.

razvand (changes requested):

flpostolache:
@skuenzer Thank you for this PR. There is a typo in the first commit message `an complete implementation`. Also, would it be better if we change the assertion in `eventfd.c` from `UK_ASSERT(ret >= 0)` to `UK_ASSERT(ret == 0);` to be more restrictive? 0 is the only value that can be returned in case of success when fcntl is used with `F_SETFL`. Otherwise, the PR looks good and works on my side.

lib/posix-event/eventfd.c
skuenzer: 👍🏻

skuenzer:
> @skuenzer Thank you for this PR. There is a typo in the first commit message `an complete implementation`. Also, would it be better if we change the assertion in `eventfd.c` from `UK_ASSERT(ret >= 0)` to `UK_ASSERT(ret == 0);` to be more restrictive? 0 is the only value that can be returned in case of success when fcntl is used with `F_SETFL`. Otherwise, the PR looks good and works on my side.

Sure, I updated it. I was seeing it more relaxed that `fcntl` should just not fail. However, you are right, specification requirs it to return `0` in this case. :+1:
I also fixed the typo that you found. Thanks for finding it.

flpostolache (approved):
Looks good! @razvand please have another look when you can.

razvand (approved):

PR #906, opened 2023-05-23
lib/ukgcov: Introduce library for code coverage

Starnox:
Initial setup with config options for console, binary file and memory output.

### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`, `linuxu`]
 - Application(s): [N/A]

### Additional configuration

- Select gcov from `Library Configuration` KConfig menu.
- Select the way you want to extract the `gcov` information from the kernel
- Add [`musl`](https://github.com/unikraft/lib-musl)

### Description of changes

Introducing a new internal library taking advantage of gcc-13 newest `Profiling and Test Coverage in Freestanding Environments`.

Depends on: #1042 #1043 #1044


lib/gcov/Config.uk
```diff
@@ -0,0 +1,25 @@
+menuconfig LIBGCOV
+	bool "gcov: Code coverage support"
+	default n
+	help
+		This library implements the gcov support for code coverage
```
michpappas: "implements freestanding gcov support"

lib/gcov/Config.uk
```diff
@@ -0,0 +1,25 @@
...
+		analysis.
+
+if LIBGCOV
+
+choice LIBGCOV_OUTPUT
+	prompt "Coverage data output"
+	default LIBGCOV_OUTPUT_SERIAL_HEXDUMP
+config LIBGCOV_OUTPUT_SERIAL_HEXDUMP
```
michpappas: Right now the console option outputs binary data, which cannot be captured. These need to be converted to printable characters, similarly to `embedded-gcov`. Since it uses a permissive license, you can even port the relevant code from there, if needed.

lib/gcov/include/ukgcov.h
```diff
@@ -0,0 +1,59 @@
...
+
+/* Set all counters to zero.  */
+
+extern void __gcov_reset (void);
+
+/* Write profile information to a file.  */
+
+extern void __gcov_dump (void);
+
+/* Convert the gcov information referenced by INFO to a gcda data stream.
```
michpappas: ⚠️ **Careful here!** This is from https://github.com/gcc-mirror/gcc/blob/master/libgcc/gcov.h, which **is licensed under the GPL** 💀

Moreover, only `dump_gcov_info()` needs to be public. The rest of the definitions can be moved to `ukgcov.c`. The documentation is not necessary, you can add a comment with a link to https://github.com/gcc-mirror/gcc/blob/master/libgcc/gcov.h, if you think it's necessary.

According to Unikraft's coding convention's, the public headers need to be name-spaced, so please create a `uk` subdirectory under `include/`, ie `include/uk/`. And since this header is now name-spaced, you can now call the header `gcov.h` (without the `uk` prefix). Your users will then do:
```
#include <uk/gcov.h>
```

Finally, I would also prefix the exported symbol with `gcov`, ie rename `dump_gcov_info()` to `gcov_dump_info()`.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
+/* SPDX-License-Identifier: MIT */
+/* Copyright © 2005-2020 Rich Felker, et al.
```
michpappas: Where is this license and copyright derived from? Bellow you (correctly) reference https://gcc.gnu.org/onlinedocs/gcc/Freestanding-Environments.html, but I don't see anything relevant there.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+static __attribute__ ((used)) size_t gcov_output_buffer_pos;
+static size_t gcov_output_buffer_size = INITIAL_BUFFER_SIZE;
+#endif
+
+/* This function shall produce a reliable in order byte stream to transfer the
+   gcov information from the target to the host system.  */
+static void dump(const void *d, unsigned n, void *arg)
+{
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+    (void)arg;
```
michpappas: Use `__unused` in the parameter declaration instead, ie
```
static void dump(const void *d, unsigned n, void *arg __unused)
```

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+static unsigned char __attribute__ ((used)) *gcov_output_buffer;
+static __attribute__ ((used)) size_t gcov_output_buffer_pos;
+static size_t gcov_output_buffer_size = INITIAL_BUFFER_SIZE;
+#endif
+
+/* This function shall produce a reliable in order byte stream to transfer the
+   gcov information from the target to the host system.  */
+static void dump(const void *d, unsigned n, void *arg)
+{
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
```
michpappas: It would be cleaner to move the implementation of each output type to its own static function, and call it from here. Similarly for the dump function.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+    (void)arg;
+    const unsigned char *c = d;
+    if (gcov_output_buffer_pos + n >= gcov_output_buffer_size) {
+        unsigned char *temp_buffer = (unsigned char *) realloc(gcov_output_buffer, gcov_output_buffer_size <<= 1);
```
michpappas: No space after casting, here and everywhere else.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+   certain conditions.  */
+static void *allocate(unsigned length, void *arg)
+{
+	(void)arg;
+	return malloc(length);
+}
+
+int dump_gcov_info (void)
+{
+    const struct gcov_info *const *info = __gcov_info_start;
```
michpappas: Inconsistent indentation. Please run `./support/scripts/checkpatch.uk -g HEAD` and fix the findings.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+	const struct gcov_info *const *end = __gcov_info_end;
+
+	/* Obfuscate variable to prevent compiler optimizations.  */
```
michpappas: Does GCC optimize this out? If so, a more elegant way is to add `__used` in the declarations above. See `include/uk/essentials.h`

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+	__asm__("" : "+r"(info));
+
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+    gcov_output_buffer = (unsigned char *) malloc(gcov_output_buffer_size);
+    if (gcov_output_buffer == NULL) {
+        return -1;
```
michpappas: It's better to return `errno` types (in this case `-ENOMEM`) to provide the caller with some insight.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+    }
+#endif
+
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_FILE
+    FILE *file = fopen(GCOV_OUTPUT_BINARY_FILENAME, "w");
+    if (file == NULL) {
+        return -1;
```
michpappas: `return -ENOENT;`

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,105 @@
...
+	while (info != end) {
+#ifdef GCOV_OPT_OUTPUT_SERIAL_HEXDUMP
+		void *arg = (void *) stdout;
+#elif defined(GCOV_OPT_OUTPUT_BINARY_FILE)
+        void *arg = (void *) file;
+#else
+        void *arg = NULL;
+#endif
+		__gcov_info_to_gcda(*info, filename, dump, allocate, arg);
+		fputc('\n', stderr);
```
michpappas: This should be conditional to console dump.

michpappas (changes requested):
Hi @Starnox, great work 👍🏼 Besides the inline comments,

* Please consider adding a `README.md`. Eventually library READMEs will be automatically merged to form Unikraft's documentation.
* Add a check in `Makefile.uk` for the minimum GCC version. You can do this with `$(call error_if_gcc_version_t,13,0)` (along with an explanatory comment).
* Optionally add a section in your `README.md` on how to compile with GCC-13 in environments that don't support this version yet. A link to https://hub.docker.com/_/gcc/ will suffice.

From the TODOs in your cover letter, before merging we definitely need the scripts to generate the lcov report, and the remaining work on the console. If the latter becomes a blocker, we can merge first and add console support as a followup item.

razvand:
@Starnox , @michpappas , isn't this superseded by embedded-gcov?

michpappas:
> @Starnox , @michpappas , isn't this superseded by embedded-gcov?

@razvand this intends to supersede embedded-gcov

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,190 @@
...
+	return malloc(length);
+}
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+static int gcov_dump_info_memory(void)
+{
+	const struct gcov_info *const *info = __gcov_info_start;
+	const struct gcov_info *const *end = __gcov_info_end;
+
+	gcov_output_buffer = (unsigned char *)malloc(gcov_output_buffer_size);
```
razvand:
```suggestion
	gcov_output_buffer = (unsigned char *) malloc(gcov_output_buffer_size);
```

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,190 @@
...
+{
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+	return gcov_dump_info_memory();
+#elif defined(GCOV_OPT_OUTPUT_BINARY_FILE)
+	return gcov_dump_info_file();
+#elif defined(GCOV_OPT_OUTPUT_SERIAL_HEXDUMP)
+	return gcov_dump_info_serial();
+#endif
+
+	// Should never reach this point
```
razvand:
```suggestion
	/* Should never reach this point */
```

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,67 @@
+#  SPDX-License-Identifier: BSD-3-Clause
```
razvand: Shouldn't we use the same license header, with `Unikraft authors ...` and simplified content? @michpappas

lib/gcov/scripts/gcov_serial_parse.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python3
```
razvand: Add license header.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,67 @@
...
+#  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+################################################################################
+# GCC version check
+################################################################################
+$(call error_if_gcc_version_lt,13,0)
+
+################################################################################
+# App registration
```
razvand: Why is this called "App registration"? Shouldn't it be library registration?

lib/gcov/scripts/gcov_process.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+set -e
+
+if [[ $# -ne 3 ]]; then
+	echo "Usage: $0 [-b <binary_dump> | -c <console_text>] <build_dir>"
+	exit
```
razvand:
```suggestion
	exit 1
```

lib/gcov/scripts/gcov_process.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if [ -n "${CROSS_COMPILE}" ]; then
+	GCOV=${CROSS_COMPILE}/gcov
+else
+	GCOV=$(which gcov)
+fi
+
+if [ -z "$GCOV" ]; then
+	echo "$0: Could not find gcov"
+	exit
```
razvand:
```suggestion
	exit 1
```

lib/gcov/scripts/gcov_process.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+if [ -n "${CROSS_COMPILE}" ]; then
+	GCOV=${CROSS_COMPILE}/gcov
+else
+	GCOV=$(which gcov)
+fi
+
+if [ -z "$GCOV" ]; then
+	echo "$0: Could not find gcov"
```
razvand:
```suggestion
	echo "$0: Could not find gcov" 1>&2
```

lib/gcov/scripts/gcov_process.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+set -e
+
+if [[ $# -ne 3 ]]; then
+	echo "Usage: $0 [-b <binary_dump> | -c <console_text>] <build_dir>"
```
razvand:
```suggestion
	echo "Usage: $0 [-b <binary_dump> | -c <console_text>] <build_dir>" 1>&2
```

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,190 @@
...
+extern const struct gcov_info *const __gcov_info_start[];
+extern const struct gcov_info *const __gcov_info_end[];
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+static void dump_memory(const void *d, unsigned n, void *arg __unused)
+{
+	const unsigned char *c = d;
+
+	if (gcov_output_buffer_pos + n >= gcov_output_buffer_size) {
+		unsigned char *temp_buffer = (unsigned char *)realloc(gcov_output_buffer, gcov_output_buffer_size <<= 1);
```
razvand:
```suggestion
		unsigned char *temp_buffer = (unsigned char *) realloc(gcov_output_buffer, gcov_output_buffer_size <<= 1);
```

razvand (changes requested):
Hi, @Starnox. This looks good. I need to test it.

Please see some minor comments.

Apart from those:

- Add a `README.md` file detailing how to use the library and the role of different scripts.
- If you haven't already, pass shell scripts through [`shellcheck`](https://www.shellcheck.net/) and Python scripts through [`pylint`](https://pypi.org/project/pylint/).

lib/gcov/ukgcov.c
Starnox: I saw similar styling like the one I used in `vfscore`. Can you please confirm which is the right convention here?

lib/gcov/Config.uk
```diff
@@ -0,0 +1,26 @@
+menuconfig LIBGCOV
+	bool "gcov: Code coverage support"
```
michpappas: We need a decision on whether this library should be called `gcov` or `ukgcov`. @razvand, @skuenzer, did we discuss about this before?

lib/gcov/Config.uk
```diff
@@ -0,0 +1,26 @@
...
+	default n
+	select VFSCORE
```
michpappas: If this is only used on binary file output (please confirm my understanding) this `select` can be moved to `LIBGCOV_OUTPUT_BINARY_FILE`.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,14 @@
...
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#ifndef __UKGCOV_H__
+#define __UKGCOV_H__
+
+static const size_t INITIAL_BUFFER_SIZE = 100 * (1 << 10); // 100 KB
+
+/* Dump the gcov information to whatever output is configured in KConfig */
```
michpappas: Since this is a public function this should be in doxygen format and document the return value.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,193 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the MIT License (the "License", see COPYING.md).
+ * You may not use this file except in compliance with the License.
+ */
+/*
+ * The code in this file was derived from gcc-13-onlinedocs:
```
michpappas: 💀 **Careful here** 💀 What is the license for code derived from the GCC docs? Above you are licensing this code as `BSD-3. Clause`. Is there any information available on that? Also, is that part of the code or all code in this file?

The easiest way to resolve this is to replace derived code with original work, ie written from scratch without referring to the non-compatible licensed work. If not possible, then the licensing needs to be carefully reviewed by someone before we merge this PR (and I'm not a lawyer) ⚠️ 

**CC:** @razvand 

PS: I see also a few coding style issues, but I'll wait until the  licensing topic is resolved before fully reviewing this file.

support/scripts/gcov/genhtml_report.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
michpappas: For any scripts that are 3rd party work imported, please add information that allows complete traceability in your commit message, i.e. filename, repository, branch, commit.

support/scripts/gcov/gcov_process.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
michpappas: What's the license of this file?

michpappas (changes requested):
Hi @Starnox, please have a look at the licensing question, unfortunately I didn't see it earlier.

lib/gcov/README.md
```diff
@@ -0,0 +1,77 @@
+# Unikraft Library for Profiling and Test Coverage
+
+This library introduces initial setup with config options for console, binary file, and memory output. It takes advantage of gcc-13's newest feature of Profiling and Test Coverage in Freestanding Environments.
```
michpappas: Here and elsewhere: `s/gcc-13/GCC-13/g`. Also although it's newest now, in the future it will look stale. Prefer something like `GCC's Profiling and Test Coverage...`. This takes care of the next comment about minimum version.

lib/gcov/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Prerequisites
+
+One way to build with `gcc-13` is to use a `Docker` container - [DockerImage](https://hub.docker.com/_/gcc/)
```
michpappas: Although you mentioned it already on the previous section looking at this paragraph is a bit sudden and without context to someone who looks directly at this section. Please add a sentence before that this requires a minimum version of GCC-13.

lib/gcov/README.md
```diff
@@ -0,0 +1,77 @@
...
+./support/scripts/gcov/gcov_process.sh -c <console_output> <build_directory>
+```
+
+**For Binary File:**
+
+```bash
+./support/scripts/gcov/gcov_process.sh -b <binary_output> <build_directory>
+```
+
+## Conclusion
```
michpappas: A conclusion section does not provide much value into this type of document. Maybe you can instead add a `Known Issues` section with the problem with duplicate filenames (but also keep in mind that this needs to be updated once the issue is resolved).

michpappas (changes requested):
@Starnox thanks for that, please see the comments below.

lib/gcov/Config.uk
Starnox: Hi! This is not only used for binary file output. Is is required in order for the gcov objects to link in all three cases.

michpappas: So if `LIBGCOV_OUTPUT_BINARY_FILE` requires `VFSCORE` and the code that implements binary file output is wrapped around `#if CONFIG_LIBGCOV_OUTPUT_BINARY_FILE`, why is vfscore a linkage requirement for other output types when binary output is disabled?

lib/gcov/ukgcov.c
michpappas: The convention is no space.

lib/gcov/Config.uk
```diff
@@ -0,0 +1,26 @@
...
+	default LIBGCOV_OUTPUT_SERIAL_HEXDUMP
+config LIBGCOV_OUTPUT_SERIAL_HEXDUMP
+	bool "Console"
+config LIBGCOV_OUTPUT_BINARY_FILE
+	bool "File"
+config LIBGCOV_OUTPUT_BINARY_MEMORY
+	bool "Memory"
+endchoice
+
+config LIBGCOV_OUTPUT_BINARY_FILENAME
```
michpappas: From going through the code it looks like that selecting binary output and not specifying a filename causes a build error. To handle this case, we have two options:

1. Set a default
2. Issue an error from the Makfile if a filename is not defined

Since we can't make a safe assumption about the configuration of the system Unikarft will be running on (permissions etc), it may be hard to make a reasonable assumption about what would be a good location to place the file by default, so issuing an error sounds like a more failsafe approach IMO.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,17 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#ifndef __UKGCOV_H__
```
michpappas: `__UK_GCOV_H__`

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+static size_t __used gcov_output_buffer_pos;
+static size_t gcov_output_buffer_size = INITIAL_BUFFER_SIZE;
+#endif
+
+/* Gcov symbols: (Source) https://github.com/gcc-mirror/gcc/blob/master/libgcc/gcov.h */
+struct gcov_info;
+extern void __gcov_reset(void);
+extern void __gcov_dump(void);
+extern void __gcov_info_to_gcda(const struct gcov_info *__info,
+			void (*__filename_fn)(const char *, void *),
```
michpappas: Indentation here is incorrect, these need to be vertically aligned to `(const struct gcov_invo *__info,` ie you need one more tab. On the next definition too. If the problem is that this results into long lines, you can try something like:

```C
typedef void (*uk_gcov_fname_fn)(const char *fname, void *arg);

extern void __gcov_info_to_gcda(const struct gcov_info *__info, 
                                uk_gcov_fname_fn fname_fn,

```

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <string.h>
+#include <uk/essentials.h>
+#include <uk/gcov.h>
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
```
michpappas: Except from include guards, preprocessor conditionals should be in the form of:
```
#if CONDITION
...
#endif /* CONDITION */
```
and
```
#if CONDITION
...
#else /* !CONDITION */ --- negation here because it is like "#elif !CONDITION"
...
#endif /* !CONDITION */ <--- repeats the last condition we had
```
The above are reversed for negative conditions, ie:
```
#if !CONDITION
...
#endif /* !CONDITION */
```
and
```
#if !CONDITION
...
#else /* CONDITION */ --- negation here because it is like "#elif CONDITION"
...
#endif /* CONDITION */ <--- repeats the last condition we had
```

This has been established very recently and it has not been documented yet, we need to open a PR in the coding conventions.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+			void *__arg);
+
+/* The start and end symbols are provided by the linker script.  We use the
+ * array notation to avoid issues with a potential small-data area.
+ */
+extern const struct gcov_info *const __gcov_info_start[];
+extern const struct gcov_info *const __gcov_info_end[];
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+static void dump_memory(const void *d, unsigned n, void *arg __unused)
```
michpappas: You can use more descriptive names for the arguments in these functions.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,17 @@
...
+#define __UKGCOV_H__
+
+static const size_t INITIAL_BUFFER_SIZE = 100 * (1 << 10); // 100 KB
```
michpappas: I take it that this is placed in this header and not in `gcov.c` so that the user can easily adjust it to a different value? If that's the case a comment would be useful to describe what this buffer is use for and instruct the user to adjust it as / if needed.

lib/gcov/Config.uk
Starnox: Sorry for the delay, I was on holiday. Because it activates support for the file operations API that comes with PR #1043 and makes available symbols required by gcov in order to build.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+# Licensed under the MIT License (the "License", see COPYING.md).
+# You may not use this file except in compliance with the License.
+
+################################################################################
+# GCC version check
+################################################################################
+$(call error_if_gcc_version_lt,13,0)
```
michpappas: You need make the GCC version check conditional to `CONFIG_LIBGCOV` as the Makefile is evaluated regardless of whether or not the library is enabled.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
michpappas: Normally we don't specify a license header on `Makefiles.uk`. See other libraries for reference.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
...
+# Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+# Licensed under the MIT License (the "License", see COPYING.md).
+# You may not use this file except in compliance with the License.
+
+################################################################################
+# GCC version check
+################################################################################
+$(call error_if_gcc_version_lt,13,0)
+
+################################################################################
```
michpappas: Normally we don't use decorative headers like these on `Makefiles.uk`. See other libraries for reference.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
...
+# Library includes
+################################################################################
+CINCLUDES-y += -I$(LIBGCOV_BASE)/include
+LIBGCOV_SRCS-y += $(LIBGCOV_BASE)/ukgcov.c
+
+################################################################################
+# Library flags
+################################################################################
+CFLAGS-$(CONFIG_LIBGCOV) += --coverage -fprofile-info-section
+LDFLAGS-$(CONFIG_LIBGCOV) += --coverage
```
michpappas: You can align these two. Perhaps also move them at the top together with `CINCLUDES`, and move the `SRCS` towards the bottom before the linker script.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
...
+
+LIBGCOV_CFLAGS-$(CONFIG_LIBGCOV_OUTPUT_SERIAL_HEXDUMP)     += -DGCOV_OPT_OUTPUT_SERIAL_HEXDUMP
```
michpappas: You don't need to pass these via `-D` parameters, the build system does this already for you 😉  You can include `uk/config.h` in `ukgcov.c` and check directly against `CONFIG_MY_OPTION`.

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,39 @@
...
+
+################################################################################
+# Library registration
+################################################################################
+$(eval $(call addlib_s,libgcov,$(CONFIG_LIBGCOV)))
+
+################################################################################
+# Library includes
+################################################################################
+CINCLUDES-y += -I$(LIBGCOV_BASE)/include
```
michpappas: Global includes must be conditional to the library, otherwise they will be appended to each library's gcc options even if this library is not enabled.

lib/gcov/Config.uk
```diff
@@ -0,0 +1,30 @@
+menuconfig LIBGCOV
+	bool "gcov: Code coverage support"
+	default n
+	select LIBVFSCORE
+	select LIBSYSCALL_SHIM
+	select LIBPOSIX_ENVIRON
+	select LIBUKMMAP
+
```
michpappas: No need for a blank line here

lib/gcov/Config.uk
```diff
@@ -0,0 +1,30 @@
...
+	help
+		This library implements freestanding gcov support
```
michpappas: I think "gcov support analysis" doesn't make sense. You can say something like "provides support for coverage analysis with gcov", or "implements freestanding gcov analysis", or something like that.

lib/gcov/Config.uk
```diff
@@ -0,0 +1,30 @@
+menuconfig LIBGCOV
+	bool "gcov: Code coverage support"
+	default n
```
michpappas: `default n` is redundant, as it is the default behavior anyway. You can remove it.

lib/gcov/libgcov.lds.S
```diff
@@ -0,0 +1,9 @@
+SECTIONS {
+	. = ALIGN(8);
+	.gcov_info : {
+        PROVIDE (__gcov_info_start = .);
```
michpappas: Mixed indentation here, use tabs everywhere.

lib/gcov/ukgcov.c
michpappas, line 1: I suggest you run checkpatch here, as there are many styling issues to address:
 `support/scripts/checkpatch.uk --strict -f ukgcov.c`.

lib/gcov/Config.uk
michpappas: So I did a quick experiment and from what I see GCC is looking for these as soon as you pass `--coverage`, I guess that's what you meant above. In that case maybe it would be worth to add a comment above these dependencies in  `Config.uk`.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+#include <uk/essentials.h>
+#include <uk/gcov.h>
+
+#ifdef GCOV_OPT_OUTPUT_BINARY_MEMORY
+static unsigned char __used *gcov_output_buffer;
+static size_t __used gcov_output_buffer_pos;
+static size_t gcov_output_buffer_size = INITIAL_BUFFER_SIZE;
+#endif
+
+/* Gcov symbols: (Source) https://github.com/gcc-mirror/gcc/blob/master/libgcc/gcov.h */
```
michpappas: A better phrasing would be `gcov symbols defined in http://...`

michpappas (changes requested):
Hi @Starnox, sorry this took a while. I did a first pass to address the high level issues like conventions and style. Once updated I'll do another round focusing on the actual implementation.

lib/gcov/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Example:
+
+```bash
+dump memory memdump.bin gcov_output_buffer gcov_output_buffer+gcov_output_buffer_pos
+```
+
+## Creating Coverage Report
+
+Process the output using `ukgcov`'s `gcov_process.sh` script, a wrapper around tools provided by [`embedded-gcov`](https://github.com/nasa-jpl/embedded-gcov). Install dependencies `dox2unix` and `lcov`, then invoke `gcov_process.sh` with parameters depending on the output method selected.
```
michpappas: `dos2unix`

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,19 @@
+$(eval $(call addlib_s,libgcov,$(CONFIG_LIBGCOV)))
+
+ifeq ($(CONFIG_LIBGCOV),y)
+$(call error_if_gcc_version_lt,13,0)
+endif
+
+CINCLUDES-$(CONFIG_LIBGCOV) 	+= -I$(LIBGCOV_BASE)/include
```
michpappas: Similarly to the linux kernel, tab-stops should be set to 8 spaces. Please update the alignment based on that, both here and in `ukgcov.c` / `gcov.h`.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef __UK_GCOV_H__
+#define __UK_GCOV_H__
+
+/*
+ * Buffer size used to store the coverage information in memory
+ * that doubles every time the buffer is full. Adjust based on
+ * memory constraints or if you know the resulted file size
+ * in order to avoid unnecessary reallocations.
+ */
+static const size_t INITIAL_BUFFER_SIZE = 100 * (1 << 10); // 100 KB
```
michpappas: According to the coding style, always use block comments, ie `/* 100 KiB */`. Also `KiB` is better, as `KB` can be ambiguous.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+#include <uk/gcov.h>
+#include <uk/config.h>
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
+static unsigned char __used *gcov_output_buffer;
+static size_t __used gcov_output_buffer_pos;
+static size_t gcov_output_buffer_size = INITIAL_BUFFER_SIZE;
+#endif /* CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY */
+
+/* gcov symbols defined in https://github.com/gcc-mirror/gcc/blob/master/libgcc/gcov.h */
```
michpappas: These are declarations defined by libgcc, not symbols. Moreover, being declarations these don't need to be `extern`.  Also:
* `__gcov_dump` is not used, so you can remove it.
* `__gcov_reset` is not used, but it's potentially useful to obtain coverage info per test, so perhaps you can add a wrapper to the API exposed by this library.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+								void (*__filename_fn)(const char *, void *),
+								void (*__dump_fn)(const void *, unsigned int, void *),
+								void *(*__allocate_fn)(unsigned int, void *),
+								void *__arg);
+extern void __gcov_filename_to_gcfn(const char *__filename,
+								void (*__dump_fn)(const void *, unsigned int, void *),
+								void *__arg);
+
+/* The start and end symbols are provided by the linker script.  We use the
+ * array notation to avoid issues with a potential small-data area.
```
michpappas: > We use the array notation to avoid issues with a potential small-data area

What does that comment mean? Also I think I've seen it somewhere before 🧐

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+ */
+extern const struct gcov_info *const __gcov_info_start[];
+extern const struct gcov_info *const __gcov_info_end[];
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
```
michpappas: As requested in a previously,  please fix these to:
```
#if CONFIG_FEATURE
...
#endif /* CONFIG_FEATURE */
```
Similarly for all other conditionals to Kconfig.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+static void dump_memory(const void *source, unsigned int n, void *arg __unused)
+{
+	const unsigned char *c = source;
```
michpappas: With the correct tab-stop value (8 spaces) there are still `LONG_LINE` checkpatch errors that can be easily fixed. Eg here what you can do is:
```diff
+       unsigned char *buf;
 
        if (gcov_output_buffer_pos + n >= gcov_output_buffer_size) {
-               unsigned char *temp_buffer = (unsigned char *)realloc(gcov_output_buffer, gcov_output_buffer_size <<= 1);
+               buf = (unsigned char *)realloc(gcov_output_buffer,
+                                              gcov_output_buffer_size <<= 1);
 
-               if (!temp_buffer) {
-                       free(gcov_output_buffer);
+               if (unlikely(!buf)) {
+                       free(buf);
                        return;
                }
-               gcov_output_buffer = temp_buffer;
+               gcov_output_buffer = buf;
```

Also notice that it's a good practice to optimize checks for error conditions with `unlikely`.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+
+	if (gcov_output_buffer_pos + n >= gcov_output_buffer_size) {
+		unsigned char *temp_buffer = (unsigned char *)realloc(gcov_output_buffer, gcov_output_buffer_size <<= 1);
+
+		if (!temp_buffer) {
```
michpappas: Use `unlikely` here.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
+static int gcov_dump_info_memory(void)
+{
+	const struct gcov_info *const *info = __gcov_info_start;
+	const struct gcov_info *const *end = __gcov_info_end;
+
+	gcov_output_buffer = (unsigned char *)malloc(gcov_output_buffer_size);
+	if (!gcov_output_buffer)
```
michpappas: Use `unlikely` here.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+struct gcov_info;
+extern void __gcov_reset(void);
+extern void __gcov_dump(void);
+extern void __gcov_info_to_gcda(const struct gcov_info *__info,
+								void (*__filename_fn)(const char *, void *),
+								void (*__dump_fn)(const void *, unsigned int, void *),
+								void *(*__allocate_fn)(unsigned int, void *),
+								void *__arg);
+extern void __gcov_filename_to_gcfn(const char *__filename,
+								void (*__dump_fn)(const void *, unsigned int, void *),
```
michpappas: Even if you remove `extern` this will exceed the 80 chars limit, if you use 8 spaces per tab. You can avoid that with:
```diff
-extern void __gcov_info_to_gcda(const struct gcov_info *__info,
-                                                               void (*__filename_fn)(const char *, void *),
-                                                               void (*__dump_fn)(const void *, unsigned int, void *),
-                                                               void *(*__allocate_fn)(unsigned int, void *),
-                                                               void *__arg);
-extern void __gcov_filename_to_gcfn(const char *__filename,
-                                                               void (*__dump_fn)(const void *, unsigned int, void *),
-                                                               void *__arg);
-
+typedef void (*__filename_fn)(const char *, void *);
+typedef void (*__dump_fn)(const void *, unsigned int, void *);
+typedef void *(*__allocate_fn)(unsigned int, void *);
+
+void __gcov_info_to_gcda(const struct gcov_info *__info,
+                        __filename_fn filename_fn, __dump_fn dump_fn,
+                        __allocate_fn allocate_fn, void *__arg);
+void __gcov_filename_to_gcfn(const char *__filename, __dump_fn dump_fn,
+                            void *__arg);

```

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
+	dump_memory(d, n, arg);
+#elif defined(CONFIG_LIBGCOV_OUTPUT_BINARY_FILE)
+	dump_file(d, n, arg);
+#elif defined(CONFIG_LIBGCOV_OUTPUT_SERIAL_HEXDUMP)
+	dump_serial(d, n, arg);
+#endif /* CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY */
+}
+
+/* The filename is serialized to a gcfn data stream by the
```
michpappas: This comment also looks familiar 🧐

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+ * __gcov_filename_to_gcfn() function.  The gcfn data is used by the
+ * "merge-stream" subcommand of the "gcov-tool" to figure out the filename
+ * associated with the gcov information.
+ */
+static void filename(const char *f, void *arg)
+{
+	__gcov_filename_to_gcfn(f, dump, arg);
+}
+
+/* The __gcov_info_to_gcda() function may have to allocate memory under
```
michpappas: Hmmm 🧐 🔎

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,191 @@
...
+{
+	FILE *file = arg;
+	const unsigned char *c = source;
+
+	fwrite(c, n, 1, file);
+}
+#endif
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_SERIAL_HEXDUMP
+static inline unsigned char *encode(unsigned char c, unsigned char buf[2])
```
michpappas: You can make this a bit more elegant with something like:
```diff
+#if CONFIG_LIBGCOV_OUTPUT_SERIAL_HEXDUMP
+
+#define e(c)   (((c) % 16) < 10 ? ((c) % 16) + '0' : ((c) % 16) - 10 + 'a')
 
-#ifdef CONFIG_LIBGCOV_OUTPUT_SERIAL_HEXDUMP
 static inline unsigned char *encode(unsigned char c, unsigned char buf[2])
 {
-       buf[0] = (c % 16) < 10 ? (c % 16) + '0' : (c % 16) - 10 + 'a';
-       c >>= 4;
-       buf[1] = (c % 16) < 10 ? (c % 16) + '0' : (c % 16) - 10 + 'a';
+       buf[0] = e(c);
+       buf[1] = e(c >> 4);
+
        return buf;
 }
```

michpappas (changes requested):
Hi @Starnox here's another round of comments. I think we're getting close 🚀

lib/gcov/ukgcov.c
Starnox: It's a thing I read about called SDA (small-data area) and some toolchains might treat simple pointers (non-array) differently when it assumes they might point to the small-data area. By declaring the symbols as arrays, we are ensuring that these pointers will not be mistaken for small-data pointers, which actually breaks the program if it does (I tried without).

lib/gcov/Makefile.uk
```diff
@@ -0,0 +1,19 @@
+$(eval $(call addlib_s,libgcov,$(CONFIG_LIBGCOV)))
+
+ifeq ($(CONFIG_LIBGCOV),y)
+$(call error_if_gcc_version_lt,13,0)
+endif # CONFIG_LIBGCOV
+
+CINCLUDES-$(CONFIG_LIBGCOV)	+= -I$(LIBGCOV_BASE)/include
```
michpappas: Sorry I missed that earlier, but you also need to add here
```diff
+CXXINCLUDES-$(CONFIG_LIBGCOV)  += -I$(LIBGCOV_BASE)/include
```
along with guards for C++ linkage in `gcov.h`:
```diff
+#ifdef __cplusplus
+extern "C" {
+#endif
... 
+#ifdef __cplusplus
+}
+#endif
```
You can have a look at how we do it in other libraries too.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,30 @@
...
+ * You may not use this file except in compliance with the License.
+ */
+#ifndef __UK_GCOV_H__
+#define __UK_GCOV_H__
+
+/* The linker script supplies the start and end symbols.
+ * We employ array syntax to circumvent potential
+ * small-data area problems
+ */
+extern const struct gcov_info *const __gcov_info_start[];
```
michpappas: Actually if these are not used by the caller, they don't need to be part of the public API. If that's the case, you can move them back to `ukgcov.c`. IMO this comment is not necessary, as we declare these using the array notation in several other libraries.

lib/gcov/include/uk/gcov.h
```diff
@@ -0,0 +1,30 @@
...
+extern const struct gcov_info *const __gcov_info_end[];
+
+/*
+ * Buffer size used to store the coverage information in memory
+ * that doubles every time the buffer is full. Adjust based on
+ * memory constraints or if you know the resulted file size
+ * in order to avoid unnecessary reallocations.
+ */
+static const size_t INITIAL_BUFFER_SIZE = 100 * (1 << 10); /* 100 KiB */
```
michpappas: By convention we use `#define` for these. Saying that, this should probably be provided as a Kconfig option instead 🤔

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,203 @@
...
+typedef void (*__dump_fn)(const void *, unsigned int, void *);
+typedef void *(*__allocate_fn)(unsigned int, void *);
+
+void __gcov_info_to_gcda(const struct gcov_info *__info,
+			 __filename_fn filename_fn, __dump_fn dump_fn,
+			 __allocate_fn allocate_fn, void *__arg);
+void __gcov_filename_to_gcfn(const char *__filename, __dump_fn dump_fn,
+			     void *__arg);
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
```
michpappas: Here and everywhere else: For `CONFIG` options we use `#if`, not `#ifdef`.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,203 @@
...
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_FILE
+static int gcov_dump_info_file(void)
+{
+	const struct gcov_info *const *info = __gcov_info_start;
+	const struct gcov_info *const *end = __gcov_info_end;
+
+	FILE *file = fopen(GCOV_OUTPUT_BINARY_FILENAME, "w");
+
+	if (!file)
```
michpappas: `unlikely`

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,203 @@
...
+	return buf;
+}
+
+static void dump_serial(const void *source, unsigned int n, void *arg __unused)
+{
+	const unsigned char *c = source;
+	unsigned char buf[2];
+
+	for (unsigned int i = 0; i < n; ++i) {
+		char *encoded = encode(c[i], buf);
```
michpappas: According to the coding conventions we generally avoid declaring block scope vars, so it's perhaps better to move the declaration of `char *encoded` to the top.

lib/gcov/ukgcov.c
```diff
@@ -0,0 +1,203 @@
...
+			 __filename_fn filename_fn, __dump_fn dump_fn,
+			 __allocate_fn allocate_fn, void *__arg);
+void __gcov_filename_to_gcfn(const char *__filename, __dump_fn dump_fn,
+			     void *__arg);
+
+#ifdef CONFIG_LIBGCOV_OUTPUT_BINARY_MEMORY
+static void dump_memory(const void *source, unsigned int n, void *arg __unused)
+{
+	const unsigned char *c = source;
+	unsigned char *buff;
```
michpappas: Cool kids use `buf` 😛 It's also consistent with the other functions in this file that use `buf` too.

support/scripts/gcov/gcov_process.sh
```diff
@@ -0,0 +1,60 @@
...
+	gcov-tool merge-stream < "$CONVERTED_FILE"
+else
+	gcov-tool merge-stream < "$OUTPUT_FILE"
+fi
+
+find "$BUILD_DIR" -type d -name objs -prune -o -name \*.gcda -exec cp {} objs/ \;
+
+echo
+
+./lcov_newcoverage.sh
```
michpappas: Looking at the scripts it seems that we have a bit too many. The ones we import from `embedded-gcov` especially do some parameter validation to handle cases of per-test runs and invoke a single command. Perhaps we can simplify this to the following and remove the scripts altogether.
```
-./lcov_newcoverage.sh
-./lcov_combine_new_base.sh
-./genhtml_report.sh
+# Capture converage info from this run
+lcov --gcov-tool gcov --capture --directory objs/ \
+       --ignore-errors gcov,source -o results/run.info
+
+# Combine baseline and coverage info from this run
+lcov -a results/base.info -a results/uk.info -o results/run.info
+
+# Generate HTML coverage report
+genhtml -o results/html results/total.info --show-details --legend
```
That would leave us with two scripts: `gcov_process.sh`, and `gcov_serial_parse.py`. So maybe it would be even better if we provided a single python script of our own, for all processing.

PS: This would also require updating `README.md`.

support/scripts/gcov/gcov_process.sh
michpappas, line 1: This will create `results/` and `objs/` under `support/scripts/gcov`. This is definitely not good, we should not be tainting the source tree. So any output should be either at a temporary directory, the local directory, or a user supplied directory. Perhaps the latter is the best, as it covers the other two ✌🏼

lib/gcov/README.md
```diff
@@ -0,0 +1,71 @@
+# Unikraft Library for Profiling and Test Coverage
+
+This library introduces initial setup with config options for console, binary file, and memory output. It takes advantage of `GCC's`` Profiling and Test Coverage in Freestanding Environments.
+
+## Prerequisites
+
+This feature was introduced in `GCC-13`, as such, this library won't work with any version lower than that.
+One easy way to build with a specific `GCC` version is to use a `Docker` container - [DockerImage](https://hub.docker.com/_/gcc/)
```
michpappas: It would be worth to add a **Notice:** here that processing should also use with the version of `lcov` that corresponds to the compiler version that supports coverage in freestanding evnironments.

michpappas (changes requested):
@Starnox last round of comments on the main logic. After that I think we only have the scripts left 🚀

lib/gcov/Config.uk
```diff
@@ -0,0 +1,39 @@
+menuconfig LIBGCOV
+	bool "gcov: Code coverage support"
+	# These dependencies are required when using the `--coverage` flag
```
michpappas:
```
      # The dependencies below are required by GCC's `--coverage`
      # parameter unconditionally to the output method we use.
```

michpappas (changes requested):
@Starnox thanks for the updates. Last round before merge! 🚀

Besides a small comment in `Config.uk` below, the outcome for the library name is that it should be `ukgcov`, because we define our own API rather than using an existing one provided by the library (eg `__gcov_info_dump`). This requires to:
* Rename the container directory `ukgcov`
* Update the library definition to `libukgcov`, and similarly for to derived options in `Makefile.uk`
* Update the options in `Config.uk`
* Update the public API and `exportsyms.uk`

Hope I didn't miss anything :)

lib/ukgcov/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Configuration
+
+To utilize this library, you will need to select and configure some options:
+
+- Select `ukgcov` from `Library Configuration` KConfig menu.
+- Choose the way you want to extract the `gcov` information from the kernel.
+
+### File System Configuration for extracting via File Output
+
+- In `plat` -> `kvm` -> `virtio`, select:
```
michpappas: This has now been moved  to drivers, and we have also automatic dependency resolution of libraries, so the driver is automatically selected by `libvfscore`, so this can be removed.

lib/ukgcov/README.md
```diff
@@ -0,0 +1,75 @@
...
+  - `virtio PCI device support`
+  - `virtio 9P device`
+- In `lib` -> `vfscore`, select:
+  - `Default root filesystem (9PFS)`
+
+- Create a directory `fs0` where the resulting filename with the name provided in `KConfig` will reside.
+
+## Usage
+
+To extract the coverage information, you can call the `gcov_dump_info()` function at the end of the program. Various output methods are provided:
```
michpappas: This has been renamed to `ukgcov_dump_info()`, however, the convention for the pubilc APIs is `uk_`, so that should be renamed to `uk_gcov_dump_info()` both here and in the implementation / header / exportsyms.

lib/ukgcov/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```bash
+./support/scripts/ukgcov/gcov_process.sh -o <result_dir>  -c <console_output> <build_directory>
+```
+
+**For Binary File:**
+
+```bash
+./support/scripts/ukgcov/gcov_process.sh -o <result_dir> -b <binary_output> <build_directory>
+```
```
michpappas, line 73: So I got bitten by this _again_ 🤦🏼 Although it's temporary as GCC 13 will eventually make it into distros, perhaps it would be a good idea to move the notice in line 10 here and rephrase to:

**Notice:** In order for the processing script to work, `lcov` must come from the toolchain used for compiling. If you are using a docker container to build with GCC-13, make sure to run `gov_process.sh` from within the container.

lib/ukgcov/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Select `ukgcov` from `Library Configuration` KConfig menu.
+- Choose the way you want to extract the `gcov` information from the kernel.
+
+### File System Configuration for extracting via File Output
+
+- In `plat` -> `kvm` -> `virtio`, select:
+  - `virtio PCI device support`
+  - `virtio 9P device`
+- In `lib` -> `vfscore`, select:
```
michpappas, line 19: One step is missing here:
```
- Automatically mount a root filesystem (/)
```

michpappas (changes requested):
Hi @Starnox I did another round of tests and left a few minor comments in `README.md`. Besides those, I didn't manage to get memory output to work due to the following errors during processing:
```
Scanning objs for .gcda files ...
Found 123 data files in objs
Processing print.gcda
/usr/src/uk_gcov/app-helloworld/gcov_out/objs/print.gcda:stamp mismatch with notes file
geninfo: WARNING: GCOV failed for /usr/src/uk_gcov/app-helloworld/gcov_out/objs/print.gcda!

... repeats for every gcda ...

geninfo: WARNING: GCOV failed for /usr/src/uk_gcov/app-helloworld/gcov_out/objs/errno.gcda!
Processing mmap.gcda
/usr/src/uk_gcov/app-helloworld/gcov_out/objs/mmap.gcda:stamp mismatch with notes file
geninfo: WARNING: GCOV failed for /usr/src/uk_gcov/app-helloworld/gcov_out/objs/mmap.gcda!
Finished .info-file creation
lcov: ERROR: no valid records found in tracefile results/uk.info
Combining tracefiles.
Reading tracefile results/baseline.info
Reading tracefile results/uk.info
```
Can you reproduce at your end?

michpappas (approved):
Regarding the issue with the memory output: It is no longer reproducible with a clean build. I think we can merge, and if it turns up again we can have another look.

@Starnox thank you very much for going through this lengthy process to add coverage support in Unikraft (without giving up)!  🥂 

razvand (approved):

PR #907, opened 2023-05-25
arch/arm/arm64: Fix `stxr`'s used in `spinlock`s register size

mogasergiu:
Our `spinlock`s have a size of 4-bytes and an alignment of 8, which, in our `semaphore` implementation, causes a padding of 4 bytes. This is fine, but it is an inconsistency.

With `QEMU` direct kernel boot, all of our free memory is zeroed out. However, when using a previous boot phase such as `UEFI`, this inconsistency shows itself through an always locked `spinlock`. `UEFI` firmware "poisons" all of its free memory with `0xaf` bytes which leads to our `stxr`, whose second register argument is 64-bit instead of 32-bit, also atomically storing the `poisoned` padding.

Thus, make sure that our `stxr` uses 32-bit register for the to be transferred register.

GitHub-Fixes: #289

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

PR #908, opened 2023-05-25
Add `ukvmem` `Aarch64` page fault handler

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Add the required functionalities to implement a basic `ukvmem` page fault handler for `Aarch64`. This PR depends on #907 #772 #848 , therefore it includes them.

### NOTE

### DIFF
--- a/arch/arm/arm64/include/uk/asm/paging.h
+++ b/arch/arm/arm64/include/uk/asm/paging.h
@@ -173,6 +173,31 @@ static inline __paddr_t PT_Lx_PTE_PADDR(__pte_t pte, unsigned int lvl)
 	return paddr;
 }
 
+static inline __paddr_t PT_Lx_PTE_SET_PADDR(__pte_t pte, unsigned int lvl,
+					     __paddr_t paddr)
+{
+	static __u64 pte_lx_map_paddr_mask[] = {
+		PTE_L0_PAGE_PADDR_MASK,
+		PTE_L1_BLOCK_PADDR_MASK,
+		PTE_L2_BLOCK_PADDR_MASK
+	};
+
+	if (PAGE_Lx_IS(pte, lvl)) {
+#ifdef CONFIG_LIBUKDEBUG
+		UK_ASSERT(lvl <= PT_MAP_LEVEL_MAX);
+#endif /* CONFIG_LIBUKDEBUG */
+		paddr &= pte_lx_map_paddr_mask[lvl];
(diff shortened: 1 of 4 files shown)


arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -104,6 +104,10 @@ struct ukarch_pagetable {
 #define PAGE_ATTR_SHAREABLE_IS		(1 << PAGE_ATTR_SHAREABLE_SHIFT)
 #define PAGE_ATTR_SHAREABLE_OS		(2 << PAGE_ATTR_SHAREABLE_SHIFT)
 
+/* Page fault error code bits */
```
michpappas, line 107: We define the rest of ESR bits in `arch.h` maybe you could move these for consistency there too

michpappas (changes requested):
@mogasergiu only a few minor comments from me 👍🏼 

In the titles of the commit messages we don't put the whole path, but rather a namespace identifier (when in doubt check the git history). In this case you can replace `include/arch/arm/arm64` with `arch/arm64`.

Also, if I'm not mistaken, you can remove the dependencies and rebase so that this can be merged.

mogasergiu:
> Also, if I'm not mistaken, you can remove the dependencies and rebase so that this can be merged.

Hi, thanks a lot for taking a look :D. Unfortunately, this PR depends on #848, so that would probably need to be merged first :/.

arch/arm/arm64/include/uk/asm/paging.h
mogasergiu: Actually, I am not so sure, I now remember that the reason why I did this is because x86 also does this in `arch/x86/x86_64/include/uk/asm/paging.h`.

michpappas: Yes, but in the scope of plat re-arch these will be moved into `arch/` too.

mogasergiu: I understand. IMO this should still stay this way to be consistent with x86. When plat re-arch comes, I believe this migration should be all part of the same commit. Therefore, I would like to keep it like this for now, because there is going to be some time until plat re-arch comes around.

michpappas (approved):

razvand (approved):

PR #909, opened 2023-05-25
Add `EFI` stub support

mogasergiu:
### Prerequisite checklist

### Base target

 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Add support for `EFI` stub, so that we can boot from an `EFI` environment without a bootloader as well as have access to `EFI` structures. Furthermore, add corresponding configurations, as well as a `QEMU Virt` specific configuration to distinguish, in the case of `Aarch64` between the two booting methods.

This PR depends on #907 #772 #848 and #908, therefore it includes them.

### NOTE


mogasergiu:
Removed `pe_header.S`. Now, the entire creation of the equivalent PE/COFF headers is done through `mkefi.py`.

plat/kvm/Makefile.uk
```diff
@@ -134,7 +134,6 @@ LIBKVMPLAT_SRCS-y                          += $(UK_PLAT_COMMON_BASE)/memory.c|co
 LIBKVMPLAT_SRCS-y                          += $(UK_PLAT_KVM_DEF_LDS)
 ifeq ($(CONFIG_KVM_BOOT_EFI_STUB),y)
 LIBKVMPLAT_SRCS-y                          += $(LIBKVMPLAT_BASE)/efi.c|common
-LIBKVMPLAT_SRCS-y                          += $(LIBKVMPLAT_BASE)/pe_header.S|common
```
mogasergiu: Uuh... oops looks like I forgot to squash this commit while rebasing sorry about that.

plat/kvm/efi.c
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/* Runtime Services memory regions in the Memory Attribute Table have a higher
+ * granularity regarding sizes and permissions: the ones resulted from
+ * GetMemoryMap only differentiate between Runtime Services Data/Code, while
+ * the MAT also differentiates between permissions of the Runtime Services'
+ * PE sections (Runtime Services can basically be thought of as loaded Portable
+ * Executable format drivers).
+ */
+static void uk_efi_rt_md_to_bi_mrds(struct ukplat_memregion_desc **rt_mrds,
```
mogasergiu, line 229: Apparently this table is optional. Just booted on a laptop that is missing the MAT. Will rework this so that if MAT is not found `uk_efi_md_to_bi_mrd` will add the memory region instead of returning `-EEXIST`

plat/kvm/efi.c
```diff
@@ -0,0 +1,342 @@
...
+	/* Fetch the Runtime Services memory regions from the MAT */
+	uk_efi_rt_md_to_bi_mrds(&rt_mrds, &rt_mrds_count);
+	for (i = 0; i < rt_mrds_count; i++) {
+		rc = ukplat_memregion_list_insert(&bi->mrds, &rt_mrds[i]);
+		if (unlikely(rc < 0))
+			uk_efi_crash("Failed to insert rt_mrd\n");
+	}
+
+	/* We no longer need the list of Runtime Services memory regions */
+	status = uk_efi_bs->free_pool(rt_mrds);
```
michpappas, line 316: I'm clearly missing something here. Inserting regions does not memcpy. How does coalesce work once we unmap this memory?

plat/kvm/efi.c
```diff
@@ -0,0 +1,342 @@
...
+		return -EEXIST;
+	case UK_EFI_BOOT_SERVICES_CODE:
+	case UK_EFI_BOOT_SERVICES_DATA:
+	case UK_EFI_CONVENTIONAL_MEMORY:
+		/* These are freed after ExitBootServices is called */
+		mrd->type = UKPLAT_MEMRT_FREE;
+
+		mrd->flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE;
+
+		break;
```
michpappas, line 164: No `default` case need here?

plat/kvm/efi.c
```diff
@@ -0,0 +1,342 @@
...
+	case UK_EFI_MEMORY_MAPPED_IO:
+	case UK_EFI_MEMORY_MAPPED_IO_PORT_SPACE:
+		mrd->type = UKPLAT_MEMRT_RESERVED;
+		mrd->flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE |
+			     UKPLAT_MEMRF_MAP;
+
+		break;
+	case UK_EFI_RUNTIME_SERVICES_CODE:
+	case UK_EFI_RUNTIME_SERVICES_DATA:
+		/* Already added through uk_efi_rt_md_to_bi_mrds() */
```
michpappas: Out of curiosity, why then not fetch these regions from here in the first place? does the MAT contain other regions besides rt?

plat/kvm/efi.c
```diff
@@ -302,6 +304,141 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
...
+	if (uk_img_hndl->load_options && uk_img_hndl->load_options_size) {
+		len = (uk_img_hndl->load_options_size >> 1) + 1;
+
+		status = uk_efi_bs->allocate_pool(UK_EFI_LOADER_DATA, len,
+						  (void **)&cmdl);
+		if (unlikely(status != UK_EFI_SUCCESS))
+			uk_efi_crash("Failed to allocate memory for cmdl\n");
+
+		/* Update  actual size */
+		len = utf16_to_ascii(uk_img_hndl->load_options, cmdl);
```
michpappas: So here we use `load_options_size` to allocate memory for `cmdl`, but then we don't use that length in `utf16_to_ascii()`. Feels a bit uncomfortable. Would it make sense to restrict the length based by adding a parameter in `utf16_to_ascii()` and in the cases where that is unknown pass some sort of max size, eg `UK_EFI_MAXPATHLEN`?

**Edit:** For a use-case on why we need this, see the comment on line 353.

plat/kvm/efi.c
```diff
@@ -302,6 +304,141 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
...
+	 * such partition
+	 */
+	status = sfs_proto->open_volume(sfs_proto, &volume);
+	if (unlikely(status != UK_EFI_SUCCESS))
+		uk_efi_crash("Failed to open Volume\n");
+
+	/* UEFI only knows UTF-16 */
+	len16 = ascii_to_utf16(file_name, (char *)file_name16);
+	if (unlikely(len16 > UK_EFI_MAXPATHLEN))
+		uk_efi_crash("File path too long\n");
```
michpappas, line 389: Too late for tears, by that point the buffer has been overflown already 😬

plat/kvm/efi.c
```diff
@@ -302,6 +304,141 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
...
+					    (void **)&uk_img_hndl);
+	if (unlikely(status != UK_EFI_SUCCESS))
+		uk_efi_crash("Failed to handle loaded image protocol\n");
+
+	return uk_img_hndl;
+}
+
+/* Read a file from a device, given a file name */
+static void uk_efi_read_file(uk_efi_hndl_t dev_h, const char *file_name,
+			     char *const *buf, __sz *len)
```
michpappas: `const char **buf`

support/scripts/mkefi.py
```diff
@@ -0,0 +1,267 @@
...
+    # Fetch base address, end address and bss start address from ld script
+    base_addr = get_sym_val(elf_dbg, r'_base_addr')
+    bss_addr = get_sym_val(elf_dbg, r'__bss_start')
+
+    # Consider the first PT_LOAD Segment as the first in the file.
+    ld_phdrs = get_loadable_phdrs(opt.elf)
+
+    # Make sure they are sorted by their addresses
+    ld_phdrs = sorted(ld_phdrs, key=lambda x: x['VirtAddr'])
+    end_addr = ld_phdrs[-1]["VirtAddr"] + ld_phdrs[-1]["MemSiz"]
```
michpappas, line 165: If there is a section of zero `MemSiz`, this results into a broken PE header.

Consider replacing with:
```C
ld_phdrs = sorted(ld_phdrs, key=lambda x: x['VirtAddr'] + x['MemSiz'])
```

support/scripts/mkefi.py
```diff
@@ -0,0 +1,267 @@
...
+                "PointerToLinenumbers": (0x0, 0x4),
+                "NumberOfRelocations" : (0x0, 0x2),
+                "NumberOfLinenumbers" : (0x0, 0x2),
+                "Characteristics"     : (PE_SEC_CHRS, 0x4),
+           }
+
+# Get the absolute value of symbol, as seen through `nm`
+def get_sym_val(elf, sym):
+    exp = r'^\s*' + r'([a-f0-9]+)' + r'\s+[A-Za-z]\s+' + sym + r'$'
+    out = subprocess.check_output(['nm', elf])
```
michpappas, line 113: I find shell + regex a bit flaky in general, because you rely on the output of a tool you can't control. Are you not using a python elf parsing library here to minimize dependencies on non-standard python libs?

In any case, if you choose to stick to `readelf`, `nm` and the like, perhaps adding a check that these are installed would be useful.

support/scripts/mkefi.py
```diff
@@ -0,0 +1,267 @@
...
+    # - all the other ELF Program Headers as PE sections with all permissions
+    # enabled (RWX)
+    with open(opt.elf, 'r+b') as f:
+        elf_file = f.read()
+
+        # Get Machine
+        if sum(elf_file[18:20]) == ELF_MACHINE["arm64"]:
+            MZ_PE_HDR["Machine"][0] = PE_MACHINE["arm64"]
+        else:
+            MZ_PE_HDR['Machine'][0] = PE_MACHINE["x86_64"]
```
michpappas, line 181: Maybe here explicitly check for all supported architectures and throw an error on unknown arch. That will let people porting to a new architecture (eg RISC-V) know that they have work to do. In they current state they will have to figure it out by debugging why the generated image doesn't work on their architecture.

plat/kvm/arm/efi_arm.c
```diff
@@ -0,0 +1,42 @@
+#include <kvm/efi.h>
```
michpappas: SPDX and legal blurb missing here.

And since github does not allow to comment on the commit message:

>    This is supposed to finish the architecture specific setupin order

"setup in order"

>    to be able to have a Unikraft valid initial environment state before
>    `_libkvmplat_start` is execited.

"is executed "

plat/kvm/arm/efi_arm.c
michpappas, line 1: Since this is already under `kvm/arm/` the `_arm` suffix in the filename is redundant.  Perhaps something like `efi_post.c` or similar would be more appropriate? Doesn't matter too much though, feel free to resolve if you think it's unnecessary.

plat/kvm/arm/efi_arm.c
```diff
@@ -0,0 +1,42 @@
...
+extern __paddr_t vector_table;
+extern void start_mmu(void);
+
+static __u8 __align(16) uk_efi_bootstack[__PAGE_SIZE];
+
+/* We cannot set the system register directly */
+#define uk_efi_set_sp_el1(sp)						\
+	__asm__ __volatile__("mov sp, %0\n" ::"r" (sp))
+
+void __noreturn uk_efi_jmp_to_kern(void)
```
michpappas: How is that different than the existing `entry64.S`? Should we be using unified entry code?

plat/kvm/Config.uk
```diff
@@ -29,6 +29,12 @@ config KVM_BOOT_PROTO_LXBOOT
 	help
 		Linux 64-bit Boot Protocol
 
+config KVM_BOOT_QEMU_VIRT
+	bool "QEMU Virt"
```
michpappas: Good change 👍🏼  Saying that, the machine name is `virt`, not `Virt`: https://qemu.readthedocs.io/en/latest/system/arm/virt.html. Also everywhere else, incl. the commit message.

plat/kvm/Config.uk
```diff
@@ -29,6 +29,12 @@ config KVM_BOOT_PROTO_LXBOOT
...
+	depends on KVM_VMM_QEMU && !KVM_VMM_FIRECRACKER && ARCH_ARM_64
```
michpappas, line 34: If QEMU and FC are mutually exclusive in Kconfig already, is it enough to state here `depends on ARCH_ARM_64 && KVM_VMM_QEMU`?

plat/kvm/Config.uk
```diff
@@ -63,6 +63,19 @@ config KVM_BOOT_EFI_STUB_DTB_FNAME
 	string "Name of the Devicetree Blob file"
 	default "$(UK_NAME).dtb"
 
+config KVM_BOOT_EFI_STUB_RST_ATK_MITIGATION
+	bool "TCG Reset Attack Mitigation"
+	default n
```
michpappas, line 68: > Since this is not really implemented in `OVMF`'s `NVRAM` variables we disable this by default.

I don't follow, can you please elaborate on that?

plat/kvm/efi.c
```diff
@@ -518,12 +518,45 @@ static inline void uk_efi_exit_bs(void)
...
+	uk_efi_status_t status;
+	__u8 enable = 1;
+
+	status = uk_efi_rs->get_variable((__s16 *)var_name,
+					 MEMORY_ONLY_RESET_CONTROL_GUID,
+					 NULL, &data_sz, NULL);
+	/* There is either no such variable in the firmware database, or no
+	 * variable storage is supported
+	 */
+	if (status == UK_EFI_UNSUPPORTED || status == UK_EFI_NOT_FOUND)
```
michpappas, line 589: I take it that these are these not `unlikely` because what is described in the comment above is a common case?

plat/kvm/efi.c
```diff
@@ -518,12 +518,45 @@ static inline void uk_efi_exit_bs(void)
 		uk_efi_crash("Failed to to exit Boot Services\n");
 }
 
+static inline void uk_efi_reset_attack_mitigation_enable(void)
```
michpappas: It would be useful to add a comment here with a reference to:
```
Sect 4. of TCG Platform Reset Attack Mitigation Specification Version 1.10 Rev. 17
```

plat/kvm/shutdown.c
```diff
@@ -26,12 +26,45 @@
...
 #include <uk/plat/bootstrap.h>
+#include <kvm/efi.h>
+#include <uk/plat/common/bootinfo.h>
 
 static void cpu_halt(void) __noreturn;
 
-/* TODO: implement CPU reset */
-void ukplat_terminate(enum ukplat_gstate request __unused)
+#ifdef CONFIG_KVM_BOOT_EFI_STUB
+static void uk_efi_rs_reset_system(uk_efi_reset_type_t reset_type)
```
michpappas: What if UEFI does not provide Runtime Services?

plat/kvm/arm/link64.lds.S
```diff
@@ -73,6 +73,7 @@ SECTIONS {
 	_text = .;
 	.text :
 	{
+		KEEP (*(.data.boot))
```
michpappas: Sorry, maybe I'm missing something but where do you use this in `arm64`?

support/scripts/mkefi.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Also add the fake .reloc section
+        MZ_PE_HDR["NumberOfSections"][0] = len(ld_phdrs) + 1
+        MZ_PE_HDR["ImageBase"][0] = base_addr
+        MZ_PE_HDR["SizeOfHeaders"][0] = PE_SEC_SZ * (len(ld_phdrs) + 1)
+        MZ_PE_HDR["SizeOfHeaders"][0] += 334  # See MZ_PE_HDR Offset comments
+        MZ_PE_HDR["SizeOfHeaders"][0] = MZ_PE_HDR["SizeOfHeaders"][0] & ~0xfff
```
michpappas, line 197: From around this point on this gets a bit hard to review. I'm sure it's possible with a bit of effort, but let's say it's not something that you go through casually. A bit more commenting would definitely help.

michpappas (changes requested):
@mogasergiu great work as always 🚀 Review comments follow.

plat/kvm/efi.c
mogasergiu: It does `memmove` though :confused:. Those are just memory regions built specifically from  the MAT and since they get `memmove`d to a different address we do not need them anymore, so I free that memory. and that memory does end up being unmapped.

plat/kvm/efi.c
mogasergiu: Good point, should have a `default` returning an error.

plat/kvm/efi.c
```diff
@@ -0,0 +1,342 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#include <kvm/efi.h>
+#include <uk/plat/common/bootinfo.h>
+
+uk_efi_runtime_services_t *uk_efi_rs;
+uk_efi_boot_services_t *uk_efi_bs;
```
mogasergiu: Note for me, these should have been `static`. Just so I do not forget.

support/scripts/mkefi.py
mogasergiu: Pretty sure I only have standard python libs here. If `readelf` and `nm` are not installed, the user would not be able to build the kernel in the first place so they would not get to the point where the script would run anyway. `readelf` and `nm` are part of the toolchain.

plat/kvm/arm/efi_arm.c
mogasergiu: When I initially wrote these, they were all under `plat/common`. After plat re-arch I intend to move these to a separate `lib/` (along with multiboot/lxboot etx.). Yes, it makes more sense to rename these now, good point :).

plat/kvm/arm/efi_arm.c
mogasergiu: Did you mean `lcpu_start.S`? `entry64.S` does not really make sense to me as it does some `virt` specific stuff. If you did mean `lcpu_start`, then yes it should work and I took into consideration. Basically that function does not touch some registers I touch here, so it would work if I touch those registers here and then jump to `lcpu_start` to let it touch the others. The reason why I decided to keep everything here is because I would only save 3 lines of code if I called `lcpu_start`. Furthermore, if I have everything here instead, it makes it much more clear what it takes to bridge the gap between UEFI and Unikraft, IMHO. In the future, as things get merged between architectures and code looks more uniform, I might even merge the two architectures' UEFI pre-kernels. For now, I would like to keep it this way. The way it is now I do not like it at all either... the main culprit for this though, I believe, is the lack of a unified, properly written IRQ subsystem.

plat/kvm/arm/link64.lds.S
mogasergiu: Leftover from when I had `pe_header.S` and that was an assembly written source for this specific section, so that it transparently works for both arch's. It's fine to remove now. Nice catch!

plat/kvm/shutdown.c
mogasergiu: Is that even possible? It would be as if you would not be able to do system resets, bios updates or manipulate firmware database variables. Nevertheless, I should definetely check for a `NULL` pointer as good practice anyway indeed.

plat/kvm/efi.c
mogasergiu: Yes, it is quite a common case. By default, default distribution installed `OVMF` does not have such variables enabled, but you can configure it to be if you manually build `OVMF`, but I doubt that is the case for your usual user. Many baremetal machines do have it though.

plat/kvm/Config.uk
mogasergiu: Thanks, but it's weird actually, this commit should have been higher up in the commit history, no idea what happened and how it ended up last.

plat/kvm/Config.uk
mogasergiu: I thought so too, but it would often be the case, whenever I would switch between the two, that this config would stay enabled, even though it does not make any sense. I will try to delete it again and see if it still happens.

plat/kvm/efi.c
michpappas: Bruh... 🤦🏼 I never asked

plat/kvm/efi.c
mogasergiu: Adding the clarification here as well for completion. The memory descriptors from the MAT are sub-memory descriptors of the runtime services' memory descriptors that are returned by `GetMemoryMap()`. The difference is that the former has higher page granularity and permissions.

plat/kvm/Config.uk
mogasergiu: I missed this comment, but it is the same answer

> Yes, it is quite a common case. By default, default distribution installed OVMF does not have such variables enabled, but you can configure it to be if you manually build OVMF, but I doubt that is the case for your usual user. Many baremetal machines do have it though.

plat/kvm/arm/entry64.S
```diff
@@ -49,7 +49,7 @@ lcpu_bootstack:
  */
 
 .text
-ENTRY(_libkvmplat_entry)
+ENTRY(qemu_virt_entry)
```
michpappas: In anticipation of Firecracker support on arm64, please refrain from renaming this to QEMU virt

support/scripts/mkefi.py
```diff
@@ -0,0 +1,267 @@
...
+PE_SEC_CHRS = (
+                0x00000040 |  # IMAGE_SCN_CNT_INITIALIZED_DATA
+                0x00d00000 |  # IMAGE_SCN_ALIGN_4096BYTES
+                0x20000000 |  # IMAGE_SCN_MEM_EXECUTE
+                0x40000000 |  # IMAGE_SCN_MEM_READ
+                0x80000000    # IMAGE_SCN_MEM_WRITE
+)
+
+MZ_PE_HDR = {
+    # MS-DOS Stub (Image Only)
```
michpappas, line 40: ~~It would be better if the MZ header is defined in a separate dictionary, as that will make it easier to add support for the linux variant on arm64.~~

**Edit:** I think I should do that in a separate script instead. Background: For FC and QEMU virt with initrd we need an [arm64 linux boot image](https://www.kernel.org/doc/Documentation/arm64/booting.txt). That requires only the MZ header, ie without PE/COFF, being prepended to a raw binary. My original intention was to add that functionality into `mkefi.py`, but:

1. The header also applies to non-EFI images
2. I don't have a use-case for EFI images that use the linux arm64 boot header

Therefore, what I will do is create a separate script, and to cover any future requirement for UEFI images with a linux arm64 boot header, I will add an option to overwrite the MZ header of an existing UEFI image.

plat/kvm/Config.uk
```diff
@@ -26,13 +29,28 @@ config KVM_BOOT_PROTO_LXBOOT
...
 		Linux 64-bit Boot Protocol
 
+config KVM_BOOT_EFI_STUB
+	bool "EFI stub"
+	depends on KVM_VMM_QEMU && !KVM_VMM_FIRECRACKER
+	select OPTIMIZE_PIE
+	select UKPLAT_ACPI
+	depends on !PLAT_LINUXU
+	help
+		Prepend a fake, architecture specific, PE header to the
```
michpappas: Too many implementation details here. This is unnecessary because:
* If the implementation changes this will rot
* It  is not the information the user is looking for at this point

A generic description of the feature would suffice.

plat/kvm/arm/entry64.S
mogasergiu: Could you give me a suggestion? Maybe just `virt`? :thinking:

michpappas: Just revert. Because now that we will be supporting both QEMU virt and Firecracker, KVM is technically correct. On re-arch we will break this into individual platforms.

StefanJum:
I tested this (as part of #912) and it works :tada:, just few things that I bumped into that might be worth mentioning:

* The `KVM_BOOT_EFI_STUB` config options automatically sets some file names (cmdl, initrd, dtb), and the booting will fail if the files are not present.
  Maybe warn instead of failing, or leave the `*_FNAME` config options empty by default, and have the user set them if they are needed.
* The number of memory regions need to be increased from the default `128`, again a warning could help

mogasergiu:
> I tested this (as part of #912) and it works

Hi @StefanJum, thanks a ton for testing this! Feedback is always very much welcome :heart:.
 
> * The `KVM_BOOT_EFI_STUB` config options automatically sets some file names (cmdl, initrd, dtb), and the booting will fail if the files are not present.
>   Maybe warn instead of failing, or leave the `*_FNAME` config options empty by default, and have the user set them if they are needed.

Yes, indeed, it was something that I thought of as well at some point. The way it is now, in order for booting not to fail, the user must place an empty string in the name of the resource, which does look kind of inflexible indeed. A warning instead, as you've said, might prove to be better :D.

> * The number of memory regions need to be increased from the default `128`, again a warning could help

I chose `128` because that is also what Linux chooses by default and it is highly dependent on the platform, rather than whether it is `UEFI` or not. I would keep this as `128`, however, probably just failing if there is not enough available space for the memory regions might be too strict and we should just continue as soon as that happens and warn the user instead :thinking:. @michpappas What is your opinion on this?

michpappas:
> I chose 128 because that is also what Linux chooses by default and it is highly dependent on the platform, rather than whether it is UEFI or not. I would keep this as 128, however, probably just failing if there is not enough available space for the memory regions might be too strict and we should just continue as soon as that happens and warn the user instead thinking. @michpappas What is your opinion on this?

I think issuing a warning is not an option, as we have no guarantees on the system's correct operation at that point. So perhaps providing a clear error message would be more appropriate.

plat/kvm/Config.uk
```diff
@@ -7,24 +7,89 @@ menuconfig PLAT_KVM
...
+		Multiboot Boot Protocol Version 1
 
 config KVM_BOOT_PROTO_LXBOOT
-	bool
+	bool "Lxboot"
+	depends on !KVM_VMM_QEMU && KVM_VMM_FIRECRACKER && ARCH_X86_64
+	help
+		Linux 64-bit Boot Protocol
+
+config KVM_BOOT_QEMU_VIRT
```
michpappas: We can define the QEMU boot protocol as what QEMU uses when you don't pass a kernel image, ie what we have now. That is:
* ELF image at `0x40100000`
* dtb at `0x40000000`

For consistency with the rest of the boot protocols this can then be renamed to `KVM_BOOT_PROTOCOL_QEMU_VIRT`. I'll need that distinction on an upcoming PR that enables  `KVM_BOOT_PROTOCOL_LXBOOT` on `arm64`.

razvand (changes requested):
If you check out at commit `56f5b4e4548ad907c57b77640e53d0c1c4c1d1d7`, it fails at build time:

```
/home/unikraft/maintainer-tools/workdir/unikraft/plat/kvm/Linker.uk:20: *** extraneous 'endif'.  Stop.
```

Due to commit `5b72d4e1e64291acee7641c04516f7c19d3806b8`, the Unikernel won't properly shutdown. It will hanve. You can build and run a simple helloworld to test that.

michpappas (approved):

razvand (approved):

blenessy:
Hi!

I'm trying to test this out with the [helloworld library in Unikraft catalog](https://github.com/unikraft/catalog/tree/main/library/helloworld):

Configured with `kraft menu --plat qemu --arch x86_64`:
1. Selected `EFI Stub` as boot protocol.
2. Set `CONFIG_KVM_BOOT_PROTO_EFI_STUB_CMDLINE_FNAME=""`
3. Set `CONFIG_KVM_BOOT_PROTO_EFI_STUB_INITRD_FNAME=""`
4. Set `CONFIG_KVM_BOOT_PROTO_EFI_STUB_DTB_FNAME=""`
5. Disabled the  VGA console driver

Build with `kraft menu --plat qemu --arch x86_64`: 

Tested with:

```shell
qemu-system-x86_64 -bios /usr/share/qemu/OVMF.fd -kernel .unikraft/build/helloworld_qemu-x86_64 -nographic
[    0.000000] CRIT: [libkvmplat] <memory.c @   81> Assertion failure: (mrd)->pg_count * 0x1000UL == (mrd)->len
```

What am I doing wrong, please help ?

PR #911, opened 2023-05-25
Refactor the `ACPI` code and change its location.

mogasergiu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Re-write the `ACPI` structures definitions and, since `ACPI` is platform and architecture agnostic, move it to a location where it reflects that.

This PR depends on #772 #848 #908 #909 and #910, therefore it includes them.

### NOTE
This is part of a larger set of Pull Requests. To make things easier and to ensure that things build as expected, it is recommended that testing is done based on #912, which includes everything. The splitting has been done to ease the review process.

PR #910, opened 2023-05-25
Add the ability to build `EFI` `ISO` and `DISK` images.

mogasergiu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): `x86_64`, `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Add the ability to build `EFI` `ISO` and `DISK` images. Usage example:
- for `DISK`
```
support/scripts/mkukimg -k ${UK_IMG} -f disk -b ukefi -a ${ARCH} -c ${UK_CMDL} -i ${UK_INITRD} -d ${UK_DTB} -o ${OUTPUT_IMG}
```
```
sudo ${QEMU} \
			-fsdev local,id=myid,path=$(pwd)/fs0,security_model=none \
                        -device virtio-9p-pci,fsdev=myid,mount_tag=fs0,disable-modern=on,disable-legacy=off \
                        -netdev bridge,id=en0,br=virbr0 \
                        -device virtio-net-pci,netdev=en0 \
			-drive file=${OUTPUT_IMG},if=none,id=drive0,format=raw \
			-device virtio-blk,drive=drive0 \
			-drive if=pflash,format=raw,readonly=on,file=${OVMF_CODE} \
			-drive if=pflash,format=raw,file=${OVMF_VARS}
			-nographic \
			-m 8G \
			-enable-kvm \
			-cpu host -smp 4
```

- for `ISO`
```
support/scripts/mkukimg -k ${UK_IMG} -f iso -b ukefi -a ${ARCH} -c ${UK_CMDL} -i ${UK_INITRD} -d ${UK_DTB} -o ${OUTPUT_IMG}
```
```
sudo ${QEMU} \
			-fsdev local,id=myid,path=$(pwd)/fs0,security_model=none \
                        -device virtio-9p-pci,fsdev=myid,mount_tag=fs0,disable-modern=on,disable-legacy=off \
                        -netdev bridge,id=en0,br=virbr0 \
                        -device virtio-net-pci,netdev=en0 \
			-cdrom ${OUTPUT_IMG} \
[...]

PR #912, opened 2023-05-25
Add support for setting up `SMP` on `ARM64` through `ACPI`

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Add support for setting up `SMP` on `ARM64` through `ACPI`. 

This PR depends on #772 #848 #908 #909 #910 and #911, therefore it includes.
### NOTE

To make testing easier, use the script from #910 to build the Unikraft images. You will need an `OVMF BIOS` image first.

### DIFF
--- a/plat/common/arm/lcpu.c
+++ b/plat/common/arm/lcpu.c
@@ -30,6 +30,8 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+#include <uk/plat/common/bootinfo.h>
+#include <uk/plat/common/acpi.h>
 #include <uk/plat/lcpu.h>
 #include <uk/plat/io.h>
 #include <arm/smccc.h>
@@ -73,7 +75,6 @@ void __noreturn lcpu_arch_jump_to(void *sp, ukplat_lcpu_entry_t entry)
 #define FDT_SIZE_CELLS_DEFAULT 0
 #define FDT_ADDR_CELLS_DEFAULT 2
 
-static void *dtb;
 void lcpu_start(struct lcpu *cpu);
 static __paddr_t lcpu_start_paddr;
 extern struct _gic_dev *gic;
@@ -92,7 +93,59 @@ int lcpu_arch_init(struct lcpu *this_lcpu)
 	return ret;
 }
 
-int lcpu_arch_mp_init(void *arg)
+#ifdef CONFIG_UKPLAT_ACPI
+static int do_arch_mp_init(void *arg __unused)
+{
+	__lcpuid bsp_cpu_id = lcpu_get(0)->id;
+	int bsp_found __maybe_unused = 0;
+	union {
+		struct acpi_madt_gicc *gicc;
+		struct acpi_subsdt_hdr *h;
+	} m;
+	struct lcpu *lcpu;
+	struct acpi_madt *madt;
+	__lcpuid cpu_id;
+	__sz off, len;
+
+	uk_pr_info("Bootstrapping processor has the ID %ld\n", bsp_cpu_id);
+
+	/* Enumerate all other CPUs */
+	madt = acpi_get_madt();
+	UK_ASSERT(madt);
+
+	len = madt->hdr.tab_len - sizeof(*madt);
+	for (off = 0; off < len; off += m.h->len) {
+		m.h = (struct acpi_subsdt_hdr *)(madt->entries + off);
+
+		if (m.h->type != ACPI_MADT_GICC ||
+		    !(m.gicc->flags & ACPI_MADT_GICC_FLAGS_EN))
+			continue;
+
+		cpu_id = m.gicc->mpidr & CPU_ID_MASK;
+
+		if (bsp_cpu_id == cpu_id) {
+			UK_ASSERT(!bsp_found);
+
+			bsp_found = 1;
+			continue;
+		}
+
+		lcpu = lcpu_alloc(cpu_id);
+		if (unlikely(!lcpu)) {
+			/* If we cannot allocate another LCPU, we probably have
+			 * reached the maximum number of supported CPUs. So
+			 * just stop here.
+			 */
+			uk_pr_warn("Maximum number of cores exceeded.\n");
+			return 0;
+		}
+	}
+	UK_ASSERT(bsp_found);
+
+	return 0;
+}
+#else
+static int do_arch_mp_init(void *arg)
 {
 	int fdt_cpu;
 	const fdt32_t *naddr_prop, *nsize_prop, *id_reg;
@@ -103,17 +156,12 @@ int lcpu_arch_mp_init(void *arg)
 	__lcpuid bsp_cpu_id;
 	char bsp_found __maybe_unused = 0;
 	struct lcpu *lcpu;
+	void *dtb;
 
 	/* MP support is dependent on an initialized GIC */
 	UK_ASSERT(gic);
 
 	dtb = arg;
-	/**
-	 * We have to provide the physical address of the start routine when
-	 * starting secondary CPUs. We thus do the translation here once and
-	 * cache the result.
-	 */
-	lcpu_start_paddr = ukplat_virt_to_phys(lcpu_start);
 
 	bsp_cpu_id = lcpu_arch_id();
 	uk_pr_info("Bootstrapping processor has the ID %ld\n",
@@ -215,6 +263,19 @@ int lcpu_arch_mp_init(void *arg)
 	UK_ASSERT(bsp_found);
 
 	return 0;
+}
+#endif
+
+int lcpu_arch_mp_init(void *arg)
+{
+	/**
+	 * We have to provide the physical address of the start routine when
+	 * starting secondary CPUs. We thus do the translation here once and
+	 * cache the result.
+	 */
+	lcpu_start_paddr = ukplat_virt_to_phys(lcpu_start);
+
+	return do_arch_mp_init(arg);
 }
 
 int lcpu_arch_start(struct lcpu *lcpu, unsigned long flags __unused)
--- a/plat/drivers/gic/gic-common.c
+++ b/plat/drivers/gic/gic-common.c
@@ -41,7 +41,7 @@
 #error At least one GIC driver should be selected!
 #endif
 
-int _dtb_init_gic(const void *fdt, struct _gic_dev **dev)
+int init_gic(struct _gic_dev **dev)
 {
 	int rc;
 
--- a/plat/drivers/gic/gic-common.c
+++ b/plat/drivers/gic/gic-common.c
@@ -54,7 +54,7 @@ int _dtb_init_gic(const void *fdt, struct _gic_dev **dev)
 
 #ifdef CONFIG_LIBGICV3
 	/* GICv2 is not present, try GICv3 */
-	rc = gicv3_probe(fdt, dev);
+	rc = gicv3_probe(dev);
 	if (rc == 0)
 		return 0;
 #endif /* CONFIG_LIBGICV3 */
--- a/plat/drivers/gic/gic-common.c
+++ b/plat/drivers/gic/gic-common.c
@@ -47,7 +47,7 @@ int _dtb_init_gic(const void *fdt, struct _gic_dev **dev)
 
 #ifdef CONFIG_LIBGICV2
 	/* First, try GICv2 */
-	rc = gicv2_probe(fdt, dev);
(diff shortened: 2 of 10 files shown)


plat/kvm/arm/setup.c
```diff
@@ -189,7 +196,7 @@ void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
 	if (unlikely(rc))
 		UK_CRASH("Failed to initialize bootstrapping CPU: %d\n", rc);
 
-#ifdef CONFIG_HAVE_SMP
+#if defined(CONFIG_HAVE_SMP)
```
michpappas: Is there a reason for this change?

plat/kvm/arm/setup.c
```diff
@@ -48,28 +48,51 @@
 
 smccc_conduit_fn_t smccc_psci_call;
 
-static void _dtb_get_psci_method(void *fdt)
+/* TODO: Write all PSCI stuff as a proper driver... */
```
michpappas: If I may, I find this `TODO` unnecessary given that we are all aware of this, and that `psci` will be moved into `drivers` shortly (re-arch).

plat/drivers/include/gic/gic-v2.h
```diff
@@ -35,6 +35,9 @@
 
 #include <gic/gic.h>
 
+/* GICv2 GICC register map size page aligned up according to the spec. */
```
michpappas: According to what spec?

plat/drivers/include/gic/gic-v3.h
```diff
@@ -67,6 +67,9 @@
  */
 #define GICC_CTLR_EL1_EOImode_drop	(1U << 1)
 
+/* Default size from the specification */
```
michpappas: Which spec? Also clarify in the commit message please.

plat/drivers/gic/gic-common.c
```diff
@@ -34,6 +34,7 @@
 #include <gic/gic.h>
 #include <gic/gic-v2.h>
 #include <gic/gic-v3.h>
+#include <uk/plat/common/acpi.h>
```
michpappas, line 37: Needs to be conditionally included or else build breaks when ACPI is not enabled.

plat/drivers/gic/gic-v2.c
```diff
@@ -39,6 +39,8 @@
 #include <uk/asm.h>
 #include <uk/arch/limits.h>
 #include <uk/plat/lcpu.h>
+#include <uk/plat/common/acpi.h>
```
michpappas, line 43: Include conditionally?

plat/drivers/gic/gic-v2.c
```diff
@@ -490,9 +492,8 @@ static int gicv2_initialize(void)
 	return 0;
 }
 
-static int gicv2_do_probe(const void *fdt)
+static inline void gicv2_set_ops(void)
```
michpappas: Personally I find `inline` when used with `static` to be meaningful only in headers, plus compilers is smart enough nowadays to inline code when needed, but I don't have a strong opinion.

plat/drivers/gic/gic-v2.c
```diff
@@ -509,6 +510,71 @@ static int gicv2_do_probe(const void *fdt)
 
 	/* Set driver functions */
 	gicv2_drv.ops = drv_ops;
+}
+
+#if defined(CONFIG_UKPLAT_ACPI)
+/* Default page-aligned up size for GICv2 CPU interface according to spec */
```
michpappas: Spec which

plat/drivers/gic/gic-v2.c
```diff
@@ -509,6 +510,71 @@ static int gicv2_do_probe(const void *fdt)
...
+#define GICV2_GICC_MEM_SZ	0x2000
```
michpappas: Perhaps that should be in `gic-v2.h` together with `GICD_V2_MEM_SZ`?

plat/drivers/gic/gic-v2.c
```diff
@@ -509,6 +510,71 @@ static int gicv2_do_probe(const void *fdt)
...
+
+	rc = acpi_get_gicc(&gicv2_drv);
+	if (unlikely(rc < 0))
+		return rc;
+
+	rc = acpi_get_gicd(&gicv2_drv);
+	if (unlikely(rc < 0))
+		return rc;
+
+	if (unlikely(gicv2_drv.dist_mem_size != GICD_V2_MEM_SZ))
```
michpappas, line 563: Sorry maybe I missed something, but under what circumstances can this condition happen?

plat/drivers/gic/gic-v3.c
```diff
@@ -44,6 +44,8 @@
 #include <uk/bitops.h>
 #include <uk/asm.h>
 #include <uk/plat/lcpu.h>
+#include <uk/plat/common/acpi.h>
```
michpappas, line 48: Include conditionally

plat/drivers/gic/gic-v3.c
```diff
@@ -580,6 +581,71 @@ static int gicv3_do_probe(const void *fdt)
...
+		acpi_madt_gicr_t *gicr;
+		acpi_subsdt_hdr_t *h;
+	} m;
+	acpi_madt_t *madt;
+	__sz off, len;
+
+	madt = acpi_get_madt();
+	UK_ASSERT(madt);
+
+	/* We do not count the Redistributor regions, instead we rely on QEMU
```
michpappas: No assumptions that we run on QEMU please. I understand your point in the comment, but I would rephrase to something more generic.

plat/drivers/gic/gic-v3.c
```diff
@@ -589,42 +655,42 @@ static int gicv3_do_probe(const void *fdt)
 	/* Get address and size of the GIC's register regions */
 	r = fdt_get_address(fdt, fdt_gic, 0, &gicv3_drv.dist_mem_addr,
 				&gicv3_drv.dist_mem_size);
-	if (r < 0) {
+	if (unlikely(r < 0)) {
 		uk_pr_err("Could not find GICv3 distributor region!\n");
 		return r;
 	}
 
+	/* TODO: Redistributors may be spread out in different regions.
```
michpappas, line 665: Okay to add this here (and good point 👍🏼) but IMO unnecessary to mention this addition in the commit message.

plat/common/arm/lcpu.c
```diff
@@ -92,7 +93,59 @@ int lcpu_arch_init(struct lcpu *this_lcpu)
 	return ret;
 }
 
-int lcpu_arch_mp_init(void *arg)
+#ifdef CONFIG_UKPLAT_ACPI
+static int _lcpu_arch_mp_init(void *arg __unused)
```
michpappas: As per our coding conventions, no leading underscores for `static` functions please. Prefer something else like `do_arch_mp_init()` or so.

michpappas (changes requested):
@mogasergiu nice thanks, only minor comments on this one 🚀

plat/kvm/arm/setup.c
mogasergiu: An orphan change, I will delete it. If you were to look through the changelog, you would see that before there was a `#if defined(CONFIG_HAVE_SMP) && defined(CONFIG_UKPLAT_ACPI)`. But I removed it at some point and this change remained as a leftover from rebasing. Nice catch!

plat/drivers/gic/gic-common.c
mogasergiu: This happened with some x86 specific code as well, but I caught it in time. Ashamed that I did not catch it for the superior arch as well :stuck_out_tongue:.

plat/drivers/gic/gic-v2.c
mogasergiu: I don't have one either, maybe I will revert it, maybe I won't :man_dancing:, depends on the mood. On a serious note though, I agree, it's in the same file and it is one less keyword.

plat/drivers/gic/gic-v2.c
mogasergiu: This is because `GICC` entries exist for `GICv3` as well, so `GICv2`'s `acpi_get_gicc` may find `GICC`'s but they are not for the `GICv2`. Now that I take a look at it, since the `GICD` is common among both version, it should be checked for in `plat/drivers/gic/gic-common.c` and then decide upon which `GIC` version to actually probe. However, this cannot be done because `Devicetree` probing must also be accommodated for.... unless we assume `GICD` register map sizes between versions, which I would rather not do. The whole initial `_dtb_init_gic` looks dumb to me.

So, since the `GICD` is common for both, I would make the `gicd` pointer in `acpi_get_gicd` a `static` pointer so that we do not have to search for it a second time if both interrupt controllers end up being enumerated. Furthermore, I would make sure that `acpi_get_gicd` is called first in the `ACPI` probing. Let me know your thoughts.

mogasergiu: Actually, why can both versions exist in the first place? This is contrary to `ACPI`'s `GICD` structure:
> One, and only one, GIC distributor structure must be present in the MADT for an ARM based system

Why aren't `LIBGICV2` and `LIBGICV3` mutually exclusive?

plat/drivers/gic/gic-v3.c
mogasergiu: But I do like being mean in commit messages :wine_glass::moyai:. On a serious note though, the reason why I thought it should be mentioned in the commit message is because otherwise it looks like an unrelated, orphan change. So I still think it is somewhat necessary, an additional commit message line does not hurt anybody IMHO. So, if you don't mind, I would like to keep it there :).

michpappas: Okay 👍🏼

plat/drivers/gic/gic-v2.c
michpappas: > This is because GICC entries exist for GICv3 as well, so GICv2's acpi_get_gicc may find GICC's but they are not for the GICv2

Can you have a system with both a GICv2 and a GICv3? I haven't see one so far.

> Why aren't LIBGICV2 and LIBGICV3 mutually exclusive?

It's not because you can have a system with > 1 distributors. It's there because it allows a single build to be tested, released, and deployed for different machine types, that use GICv2 and GICv3.

mogasergiu: > Can you have a system with both a GICv2 and a GICv3? I haven't see one so far.

Not to my knowledge, that is why I asked why aren't the two configs mutually exclusive. What I meant is that if both are enabled, the GICv2 is probed first and it will look for GICC entries, which do not give you the GIC version, so the ACPI probe of GICv2 cannot know the version until the GICD is probed. That is why I said that I should have probably checked for the GICD first and also cache its result in a function locally declared `static` variable, so that when GICv3 is probed, it can fetch the result from when GICv2 was probed.

> It's not because you can have a system with > 1 distributors. It's there because it allows a single build to be tested, released, and deployed for different machine types, that use GICv2 and GICv3.

Ah, yes, it makes sense then. However, what drivers are (de)selected should not be reflected in the core codebase, this should have been transparently probed for in a generic, properly written, platform bus driver. But it's a discussion for later.

> So, since the GICD is common for both, I would make the gicd pointer in acpi_get_gicd a static pointer so that we do not have to search for it a second time if both interrupt controllers end up being enumerated. Furthermore, I would make sure that acpi_get_gicd is called first in the ACPI probing.

Unless you have any objections, I would proceed as I previously suggested.

plat/kvm/arm/setup.c
```diff
@@ -48,28 +48,51 @@
...
+}
+#else
+static int get_psci_method(struct ukplat_bootinfo *bi)
 {
-	int fdtpsci, len;
 	const char *fdtmethod;
+	int fdtpsci, len;
+	void *fdt;
+
+	fdt = bi->dtb;
```
razvand: Should be `fdt = (void *) bi->dtb`. Clang complains otherwise.

razvand (changes requested):

razvanvirtan:
@mogasergiu I looked on the code and tested with `nginx`, things look good to me :+1: 

One minor comment: in the https://github.com/unikraft/unikraft/pull/912/commits/980154623623641eecbba106d0791172ad8f8d04 commit message, you have one repetition: "fetch the `GIC Distributor`'s `GIC Distributor`'s physical base"

razvanvirtan (approved):
This looks good to me, thanks for the great work @mogasergiu!

michpappas (approved):

razvand (approved):

PR #923, opened 2023-05-31
Fix schedcoop idling and x86 timer interrupts

FedeParola:
### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): kvm
 - Application(s): N/A

### Description of changes

880f27c020ecfbd70fae8c1bd7efecb3cce9f170 fixes a problem in ukschedcoop in which the idle thread keeps busy looping instead of halting the cpu when no wakeup time is set (i.e., wake_up_time == 0). To test the problem simply apply [app-helloworld-diff.txt](https://github.com/unikraft/unikraft/files/11619568/app-helloworld-diff.txt) to the helloworld application (the diff causes the main thread to block) and run it. You will see the CPU usage at 100%.

655ea17d3c46904578dac921348a267234502855 fixes a problem in which the i8254 timer doesn't switch completely to software-triggered strobe in the init function.  To switch the i8254 to sw-triggered strobe mode a value must be written into the counter. Previously the timer remained in rate generator mode and kept generating interrupts until the counter was written (e.g., for a sleep()). To test add some debug message in the timer interrupt ISR and run the previous modified helloworld, you will see a lot of interrupts are triggered. E.g., [time-diff.txt](https://github.com/unikraft/unikraft/files/11619569/time-diff.txt).

### DIFF
--- a/lib/ukschedcoop/schedcoop.c
+++ b/lib/ukschedcoop/schedcoop.c
@@ -213,11 +213,13 @@ static __noreturn void idle_thread_fn(void *argp)
 		wake_up_time = (volatile __nsec) c->idle_return_time;
 		now = ukplat_monotonic_clock();
 
-		if (wake_up_time > now) {
-			if (wake_up_time)
+		if (!wake_up_time || wake_up_time > now) {
+			if (wake_up_time) {
 				ukplat_lcpu_halt_to(wake_up_time);
-			else
+			} else {
 				ukplat_lcpu_halt_irq();
+				ukplat_lcpu_enable_irq();
+			}
 
 			/* handle pending events if any */
 			ukplat_lcpu_irqs_handle_pending();
--- a/plat/kvm/x86/tscclock.c
+++ b/plat/kvm/x86/tscclock.c
@@ -292,6 +292,8 @@ int tscclock_init(void)
 	 * Initialise i8254 timer channel 0 to mode 4 (one shot).
 	 */
 	outb(TIMER_MODE, TIMER_SEL0 | TIMER_ONESHOT | TIMER_16BIT);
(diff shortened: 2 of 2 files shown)


razvand:
Nice fixes, @FedeParola. I will use your test programs to validate the solution, but, just by browsing the code, it makes sense.

I would ask you to start the commit messages with capital letter in the two commits of this PR, i.e. `Fix idling without wakeup time` (capital `F`) and `Switch i8254 to sw-triggered strobe on init` (capital `S`).

@mschlumpp, @mogasergiu, @skuenzer, @marcrittinghaus, please also take a look.

razvand (approved):
Hi, @FedeParola. I tested and, indeed, the problems are as you mentioned and the solutions as you provided. I'm approving this PR.

We'll wait for an approval step from @skuenzer. At that point, I will also add the reviewer tag, conforming to [Unikraft's review process](https://unikraft.org/docs/contributing/review-process/).

I've one question related to the i8254 sw-triggered configuration. Why are two writes (`outb`) required to `TIMER_CNTR`? Indeed, I tried only using on write (`outb`) but it did not work.

FedeParola:
Hi @razvand and thanks for the review. The i8254 has a 16 bit counter, but the register to program it is only 8 bit, so two `outb` are needed to perform a complete write. It is also possible to configure the timer to only input the high or low 8 bits of the counter (remaining bits are automatically set to 0), but keeping the two steps configuration allows later configurations to retain full resolution.

skuenzer (approved):
Dear @FedeParola,

a well explained PR with good code examples. Thanks for these fixes!

razvand (approved):

PR #924, opened 2023-06-02
lib/uksignal: Use correct kernel struct sigaction

andreittr:
### Description of changes

The structs passed to the `sigaction` library call and the `rt_sigaction` Linux syscall have different sizes and layouts; uksignal previously used the incorrect library struct in the syscall, leading to possible memory corruptions when called from a real libc (musl, newlib). This change corrects this behavior by using the kernel struct.

### Prerequisite checklist

### Base target

 - Architecture(s): X86_64, aarch64
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Application which uses sigaction through a real libc (musl, newlib).

### DIFF
--- a/lib/uksignal/ksigaction.h
+++ b/lib/uksignal/ksigaction.h
@@ -1,0 +1,28 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_KSIGACTION_H__
+#define __UK_KSIGACTION_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Imported from musl 1.2.3 */
+/* In line with the Linux kernel struct for x86_64 and aarch64. */
+
+struct k_sigaction {
+        void (*handler)(int);
+        unsigned long flags;
+        void (*restorer)(void);
+        unsigned mask[2];
+};
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __UK_KSIGACTION_H__ */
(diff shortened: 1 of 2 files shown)


lib/uksignal/include/ksigaction.h
```diff
@@ -0,0 +1,9 @@
+/* Imported from musl 1.2.3 */
```
StefanJum: Should we add the licence header here?

StefanJum (changes requested):
One comment @andreittr, everything works fine.

lib/uksignal/include/ksigaction.h
razvand: @andreittr, indeed its contents should be something like

```
 */

/* Imported from musl 1.2.3 */
/* In line with the Linux kernel struct for x86_64 and aarch64. */
#ifndef __KSIGACTION_H__
#define __KSIGACTION_H__

struct k_sigaction {
        void (*handler)(int);
        unsigned long flags;
        void (*restorer)(void);
        unsigned mask[2];
};

#endif /* __KSIGACTION_H__ */
```

@skuenzer, would this be OK with you?

Moreover, should this be **outside** the `include/` directory, since it's internal to `uksignal`, not exported? If it is OK to be in `include/`, then it should be in `include/uk/`, right?

Would it make sense to have the C++ guards?

```
#ifdef __cplusplus
extern "C" {
#endif

...

#ifdef __cplusplus
}
#endif 
``

razvand (changes requested):
Thanks, @andreittr, this works. See our comments.

andreittr:
Update: addressed comments: include license information, header guards, C++ guards.

StefanJum (approved):
All good on my side. I'll wait for the others to review and then add the tag.

razvand (commented):
@StefanJum , this looks OK with me. Please mark it approved and I'll do the final approval step.

StefanJum (approved):
All good, go @razvand 

razvand (approved):

PR #937, opened 2023-06-07
plat: Move register definitions into arch

rares-miculescu:
### Description of changes

All architectural definitions from `plat/` are moved to `arch/` to prevent duplication and maintenance problems.

### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): N/A
 - Application(s): N/A


michpappas:
@rares-miculescu I don't see any register definitions moved in the current state of this PR.

arch/x86/x86_64/include/uk/asm/arch.h
```diff
@@ -113,3 +121,80 @@
 #define X86_EFER_TCE		(1 << 15)
 
 #endif /* __PLAT_CMN_X86_CPU_DEFS_H__ */
+
+/* registers from apic_defs.h */
+#ifndef __PLAT_CMN_X86_APIC_DEFS_H__
+#define __PLAT_CMN_X86_APIC_DEFS_H__
```
mogasergiu: ~I would move the APIC definitions to `plat/common/include/x86/apic.h` instead. Also, take note of the guards
`__PLAT_CMN_X86_APIC_DEFS_H__`, they should be aligned to where the headers actually are and now you basically have both `__PLAT_CMN_X86_APIC_DEFS_H__` and `__PLAT_CMN_X86_CPU_DEFS_H__`, so it cannot be right.~

EDIT: Disregard this. It was decided to go forward like this actually.

mogasergiu (changes requested):
Hi, Rares! Thanks for the PR! Left a minor comment. By the way, it is very good practice to try and expand the commit description so that whoever reads this stuff in the future can gain some context about what you tried to do. Make it so that even future you, years later when you don't remember, would be able to understand what your intention/motivation was :laughing:.

arch/x86/x86_64/include/uk/asm/arch.h
```diff
@@ -9,4 +11,89 @@
 #include <uk/asm/apic.h>
 #include <uk/asm/pic.h>
 
+#ifndef __PLAT_CMN_X86_CPU_DEFS_H__
```
michpappas: These include guards are not necessary given that you have `__UK_ARCH_X86_64_H__` above. Saying that, `arch.h` in `arm64` is not guarded, but instead it is implicitly included via `lcpu.h`:
```
#ifndef __UKARCH_LCPU_H__
#error Do not include this header directly
#endif
```
So for the sake of consistency you can probably do the same for this file, as any common code that includes `lcpu.h` will also include `arch.h` in both architectures. The alternatlive route, and probably what will happen in the long term, is to decouple `arch.h` in `arm64` from `lcpu.h` and see what breaks, but I would refrain from doing that at this stage and leave it for later.

michpappas (changes requested):
Hi @rares-miculescu. Although there are more `x86_64` register definitions besides the ones in `lcpu.h`, I think this is a good start for establishing `arch.h` in `x86_64`, so I'm happy with this change. Later we can move more register definitions there, along with any new ones defined in the future. I have only one request for a change, please see the comment below.

plat/kvm/x86/lcpu_helpers.S
```diff
@@ -37,7 +37,7 @@
 #ifndef __PLAT_KVM_X86_LCPU_HELPERS_S__
 #define __PLAT_KVM_X86_LCPU_HELPERS_S__
 
-#include <x86/cpu_defs.h>
+#include <uk/asm/arch.h>
```
michpappas: With the latest change to forbid `arch.h` to be included directly, this would normally break. The only reason it doesn't is because `lcpu_helpers.S` is included by `lcpu_start.S` which *happens* to first include `lcpu.h`, and thus define `__UKARCH_LCPU_H__` and pass the check in `arch.h`. Conclusion: This change must be reverted, please include `cpu_defs.h` as before.

plat/xen/x86/entry64.S
```diff
@@ -25,7 +25,7 @@
 
 #include <uk/arch/types.h>
 #include <uk/arch/limits.h>
-#include <x86/cpu_defs.h>
+#include <uk/asm/arch.h>
```
michpappas: Same here.

arch/x86/x86_64/include/uk/asm/arch.h
```diff
@@ -1,12 +1,94 @@
...
+#define X86_EFLAGS_OF           (1 << 11)   /* overflow flag             */
+#define X86_EFLAGS_NT           (1 << 14)   /* nested task flag          */
+#define X86_EFLAGS_RF           (1 << 16)   /* resume flag               */
+#define X86_EFLAGS_VM           (1 << 17)   /* virtual 8086 mode flag    */
+#define X86_EFLAGS_AC           (1 << 18)   /* alignment check flag      */
+#define X86_EFLAGS_VIF          (1 << 19)   /* virtual interrupt flag    */
+#define X86_EFLAGS_VIP          (1 << 20)   /* virtual interrupt pending */
+#define X86_EFLAGS_ID           (1 << 21)   /* ID flag                   */
+
+
```
michpappas: This line is redundant, you can get rid of it.

michpappas (changes requested):
@rares-miculescu thanks. Almost there now. Besides the comments below please also rebase so that this can get merged.

plat/common/include/x86/cpu.h
```diff
@@ -31,7 +31,6 @@
 #define __PLAT_COMMON_X86_CPU_H__
 
 #include <uk/arch/types.h>
-#include <x86/cpu_defs.h>
```
michpappas, line 34: As a general rule, all files must include headers they depend on. For example, if a file uses a register definition, it must *explicitly* include the header that contains that definition, or *implicitly* include it, by including a header that includes that header.

In this case, this file uses register definitions like `CPUID3` that you moved to `arch`. You should therefore include the header that defines these, i.e. `#include <uk/arch/lcpu.h>`.

plat/common/include/x86/paging.h
```diff
@@ -44,7 +44,6 @@
 
 #include <errno.h>
 
-#include "cpu_defs.h"
```
michpappas, line 47: Same here.

plat/common/include/x86/tls.h
```diff
@@ -31,7 +31,6 @@
 #define __PLAT_CMN_X86_TLS_H__
 
 #include <x86/cpu.h> /* rdmsrl, wrmsrl */
-#include <x86/cpu_defs.h>
```
michpappas, line 34: Same here.

drivers/ukintctlr/xpic/pic.c
```diff
@@ -22,7 +22,6 @@
  */
 /* Taken from solo5 platform_intr.c */
 
-#include <uk/asm/arch.h>
```
michpappas, line 25: Same here.

plat/kvm/x86/lcpu_start.S
```diff
@@ -38,7 +38,6 @@
 #include <uk/asm.h>
 #include <uk/plat/common/lcpu.h>
 #include <uk/reloc.h>
-#include <x86/cpu_defs.h>
```
michpappas, line 41: This one is okay as above it already includes `#include <uk/plat/common/lcpu.h>`, which implicitly includes `uk/arch/lcpu.h`.

michpappas (changes requested):
@rares-miculescu I saw one more issue, one more iteration, sorry.

michpappas (approved):
Thanks @rares-miculescu

razvand (approved):

PR #938, opened 2023-06-08
`lib/uklibid`: Library identifiers and `SUBBUILD` build system feature

skuenzer:
### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [all]

### Additional configuration

 - `CONFIG_LIBUKLIBID=y`

### Description of changes

This PR introduces `lib/uklibid` which is a library that generates at compile-time unique IDs for each library. These IDs are starting from `0` and linearly increase so that they can be used for table lookups with O(1) complexity. `lib/ukdebug` and `lib/ukstore` are first users of this library and adopted accordingly.

As required ground work, this PR also introduces a `SUBBUILD` feature to the build system. This feature adds the ability to place generated files (e.g., pre-processed, compiled) in a subdirectory of the corresponding library build directory instead of just placing them in the top-level directory. For this purpose, the build system looks for a file-scoped variable with the suffix `_SUBBUILD`. Such a variable can contain a path that is created below the library build directory.

Example (`Makefile.uk`):
```Makefile
# Create `myheader.h` under `build/libmylib/include/uk/`
LIBMYLIB_SRCS-y += $(LIBMYLIB_BASE)/myheader.m4>.h
LIBMYLIB_MYHEADER_SUBBUILD += include/uk
```


support/build/Makefile.rules
```diff
@@ -59,8 +59,8 @@ endef
 ################################################################################
 
 define mk_sub_build_dir =
-$(if $(shell mkdir -p "$(BUILD_DIR)/$(1)" && cd "$(BUILD_DIR)/$(1)" >/dev/null && pwd),,\
-     $(error could not create directory "$(BUILD_DIR)/$(1)"))
+$(if $(shell mkdir -p "$(BUILD_DIR)/$(notdir $(1))" && cd "$(BUILD_DIR)/$(notdir $(1))" >/dev/null && pwd),,\
```
michpappas, line 88: If `addlib` and friends expect a name, would make more sense to return an error here if someone passes a path instead of silently handling with `notdir`?

support/build/Makefile.rules
```diff
@@ -69,13 +69,13 @@ sub_build_dir = $(addprefix $(BUILD_DIR)/,$(notdir $(1)))
 fileext = $(subst .,,$(suffix $(1)))
 
 # converts a list of library names to paths pointing to their corresponding object library file
-# libname2olib $libname
+# libname2olib $1:libname
```
michpappas, line 101: `mk_sub_build_dir` above could be documented too

support/build/Makefile.rules
```diff
@@ -830,28 +850,26 @@ endef
...
-  $($(call vprefix_src,$(1),$(2),PREPINCLUDES-y))\
-  $($(call vprefix_src,$(1),$(2),PREPFLAGS))\
-  $($(call vprefix_src,$(1),$(2),PREPFLAGS-y))\
-  $(if $(4),\
-   $($(call vprefix_srcv,$(1),$(2),$(4),PREPINCLUDES))\
-   $($(call vprefix_srcv,$(1),$(2),$(4),PREPINCLUDES-y))\
-   $($(call vprefix_srcv,$(1),$(2),$(4),PREPFLAGS))\
-   $($(call vprefix_srcv,$(1),$(2),$(4),PREPFLAGS-y))\
-  ,)\
+$(eval subbuild=$($(call vprefix_src,$(1),$(2),$(4),SUBBUILD)))
```
michpappas, line 883: At first I found the `SUBBUILD` name a bit confusing as it implies a _sub-build_ (ie a build within a build) rather than _a sub-dir under `build/`_, so something along the lines of `BUILD_SUBDIR` would had been be more clear. On the other hand I see that `mk_sub_build_dir` exists for some time now, and to be consistent you would have to also update the rest of `Makefile.rules`...

lib/uklibid/include/uk/libid.h
```diff
@@ -0,0 +1,9 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#ifndef LIBUKLIBID_H_
```
michpappas: I suspect the motivation for not using the usual format `__LIBNAME_H__` is coding conventions? If so, using leading underscores in include guards is actually permitted.

lib/uklibid/include/uk/libid.h
```diff
@@ -6,4 +6,53 @@
...
+#define _uk_libid_static(l) \
+	__UKLIBID_ ## l
+
+/**
+ * Compile-time mapping of a library name to its unique library identifier
+ * WARNING: Any unknown library name (e.g., library not part of build) will
+ *          make this pre-processor macro fail.
+ *
+ * @param libname Non-quoted name of the library (no C-string!)
+ * @return Library identifier (unsigned int)
```
michpappas: Intuitively speaking, would a fixed type like `__u64` make more sense for identifiers?

lib/uklibid/include/uk/libid.h
```diff
@@ -6,4 +6,53 @@
...
+ */
+#define uk_libid_static(libname) \
+	_uk_libid_static(libname)
+
+#ifndef __LIBNAME__
```
michpappas, line 33: What is the use-case of this?

lib/uklibid/include/uk/libid.h
```diff
@@ -6,4 +6,53 @@
...
+#endif
+
+#define UKLIBID_NONE __UI_MAX
+
+#define _uk_libid_static(l) \
+	__UKLIBID_ ## l
+
+/**
+ * Compile-time mapping of a library name to its unique library identifier
+ * WARNING: Any unknown library name (e.g., library not part of build) will
```
michpappas, line 24: Is it possible to have a user of this header that is not part of the build?

lib/uklibid/libname.c
```diff
@@ -16,3 +19,21 @@ const char *uk_libname(unsigned int libid)
 
 	return namemap[libid];
 }
+
+#if CONFIG_HAVE_LIBC || CONFIG_LIBNOLIBC
+unsigned int uk_libid(const char *libname)
+{
+	unsigned int libid;
+
+	if (unlikely(!libname))
```
michpappas: An assertion here would avoid an unnecessary conditional, even if it's an optimized one

lib/uklibid/include/uk/libid.h
```diff
@@ -51,6 +51,25 @@ extern "C" {
...
+ */
+const char *uk_libname(unsigned int libid);
+
+/**
+ * Return the library name of the current library compilation
+ *
+ * @return Reference to a '\0'-terminated C-string that contains the library
+ *         name
+ */
+#define uk_libname_self(libid) \
```
michpappas: Looks like the `libid` parameter creeped in here.

lib/ukstore/Makefile.uk
```diff
@@ -1,53 +1,19 @@
...
-
-$(LIBUKSTORE_GEN_INCLUDES_PATH)/uk/bits/store_array.h: $(LIBUKSTORE_BASE)/gen_array.h.awk $(LIBUKSTORE_BUILD)/uk_libs.in
-	$(call build_cmd,AWK,libukstore,$(notdir $@), \
-		$(AWK) -F '-' -f $(LIBUKSTORE_BASE)/gen_array.h.awk \
-		$(LIBUKSTORE_BUILD)/uk_libs.in > $@)
-
+CXXINCLUDES-$(CONFIG_LIBUKSTORE) += -I$(LIBUKSTORE_BUILD)/include
+
+# Cross-reference libraries catalog generated by lib/uklibid
+LIBUKSTORE_LIBRARIES_IN = $(call sub_build_dir,libuklibid)/libraries.in
```
michpappas, line 13: Overall this change is good and cleans up a lot of clutter, one thing that worries me slightly is whether `libraries.in` is a part of `uklibid`'s inner workings, ie considered private. On the other hand I can't think of a better alternative to solve this problem right now.

michpappas (changes requested):
Hi @skuenzer nice PR! 🚀 Just a couple of minor issues and some clarifications from my side.

support/build/Makefile.rules
skuenzer: True, it seems I missed that. I'll add it.

lib/uklibid/include/uk/libid.h
skuenzer: Time for me to study the conventions 😄. I can go back to the (previous) scheme we used with the other libs: `__LIBUKLIBID_H__`.

lib/uklibid/include/uk/libid.h
skuenzer: I was thinking also about this, but rather `__u32` or even `__u16`. Then I thought, we likely have less than `__U16_MAX` libraries in a build (16K libraries...). Then I thought, let the compiler decide an optimal default for the architecture because the values will be at most in the hundreds. A theoretical 8bit target architecture for Unikraft could still support 255 libraries. On the other hand, I agree with you too, using a fixed size integers better communicate size assumptions to calling code. The history tells us that space rockets can explode because calling code may misuse functions simply because not being aware of their original integer size assumptions. A fixed-size integer is probably also better for the case when we need to reserve space for writing libid's out somewhere... in short: you changed my mind about `unsigned int` 😄; I'll change it to `__u16`.

lib/uklibid/include/uk/libid.h
skuenzer: Pre-processor: This allows `libname` to be resolved when an alias (via macro or macro function) is handed over as argument

lib/uklibid/libname.c
skuenzer: actually, right :+1:. There is no good reason to call: `uk_libid(NULL)`

lib/uklibid/include/uk/libid.h
skuenzer: Unikraft's build system should be restrictive enough that this should not be possible. This was a try to document assumptions that were taken during development just for the case we suddenly export headers from Unikraft to externally handled builds.

lib/uklibid/include/uk/libid.h
skuenzer: It did... nasty little parameter... 😄

lib/ukstore/Makefile.uk
skuenzer: I agree with you one hundred percent. It is probably a good idea to just document this as bad practice but still best way for this case. I will add such a comment.

support/build/Makefile.rules
skuenzer: I went through the same thoughts and chose this name because it was there already (also my mistake) and I was able to keep the variable suffix short. I wanted to avoid anything that has an underscore because resolving uses a lot of underscores already: 
`<LIBNAME>_<SOURCENAME>_<VARIANTNAME>_SUBBUILD`
The thing is that the variable suffix (`_SUBBUILD`) will be used in `Makefile.uk`'s in the future while it is unlikely that somebody uses `mk_sub_build_dir` directly (although there might be exceptions). We can still choose a better name. I also had (something with) `TARGET` in mind... :thinking:

support/build/Makefile.rules
skuenzer: I'd prefer avoiding this because for achieving this we would need to do more costly string parsings with Make.

michpappas (approved):
Looks good, thanks! @razvand it's on you now.

lib/uklibid/Makefile.uk
```diff
@@ -3,3 +3,23 @@ $(eval $(call addlib_s,libuklibid,$(CONFIG_LIBUKLIBID)))
...
 CXXINCLUDES-$(CONFIG_LIBUKLIBID) += -I$(LIBUKLIBID_BASE)/include -I$(LIBUKLIBID_BUILD)/include
+
+$(LIBUKLIBID_BUILD)/libraries.in.new:
+	$(call build_cmd,GEN,libuklibid,$(notdir $@), \
+		echo $(foreach P,$(UK_PLATS) $(UK_PLATS-y),\
+		     $(foreach L,$($(call uc,$(P))_LIBS) $($(call uc,$(P))_LIBS-y), \
+		     $(L) \
+		     )) \
+		     $(UK_LIBS) $(UK_LIBS-y) \
+		| tr ' ' '\n' > $@)
```
kubanrob: It would be good if this list would be the stable for the same library names and the exactly same config for the same application.

This can be useful for different reasons:
- if the order of the list changes, compiled olibs need to be recompiled
- following from that: It is impossible to distribute precompiled olibs if they use ukdebug internally
- the ID could be used for logging or tracing
- in general, the build becomes more reproducible

For that reasons, it would be good if there would be either:
- a fixed order of this list (eg. alphabetically)
- known dependencies that need to be met to make the order deterministic, for example: _Make sure that application Makefile includes the libraries in the same order (`L` variable)._

skuenzer: This makes sense, I added a sorting step to ensure determinism. :+1:

kubanrob:
I see the same error as `integration / libc-test` on arm64: `E  tar: /github/workspace/_helloworld/.unikraft/build/libmusl/origin: Cannot open: No such file or directory`.

lib/ukdebug/outf.h
```diff
@@ -56,7 +56,7 @@ struct out_dev {
 		/* OUTDEV_KERN, OUTDEV_DEBUG */
 		struct {
 			int lvl; /* OUTDEV_KERN only */
-			const char *libname;
+			__u16 libid;
```
kubanrob, line 60: When building along with lib-musl on arm64: `error: unknown type name ‘__u16’`

skuenzer: Good point, will fix. An include is then only missing.

skuenzer:
> I see the same error as `integration / libc-test` on arm64: `E tar: /github/workspace/_helloworld/.unikraft/build/libmusl/origin: Cannot open: No such file or directory`.

Right, sorry, I applied the fix in the latest version that I just pushed.

nderjung:
Hi, please see the builds, they have all failed on:

```
No rule to make target '/github/workspace/lib/uklibid/selfid.c', needed by '/github/workspace/_helloworld/.unikraft/build/libkvmplat/selfid.uklibid.o'
```

skuenzer:
> Hi, please see the builds, they have all failed on:
> 
> ```
> No rule to make target '/github/workspace/lib/uklibid/selfid.c', needed by '/github/workspace/_helloworld/.unikraft/build/libkvmplat/selfid.uklibid.o'
> ```

It's now there 😇

skuenzer:
As we discussed offline, this release now includes support for shipping precompiled libraries, although `uklibid` identifiers are used. For a more detailed explanation, see commit 948604f52b2c993d359b59bd9933cda9821ef325.

kubanrob (approved):
Looks good to me, thank you @skuenzer!

razvand (approved):

PR #939, opened 2023-06-08
lib/ukstore: Introduce dynamic store API

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): [e.g. `x86_64` or N/A]
 - Platform(s): [e.g. `kvm`, `xen` or N/A]
 - Application(s): [e.g. `app-python3` or N/A]

### Description of changes

This series adds dynamic entry support to uk_store. Libraries can dynamically register uk_store objects at runtime, each one of which is associated with one or more entries. Similarly to static entries, each dynamic entry is associated with a getter and / or a setter.

Consumers are notified for the lifetime of objects via events.

GitHub-Supersedes: #460
GitHub-Depends-On: #938

PR #943, opened 2023-06-13
lib/uknetdev: Export stats via uk_store

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

Introduce `CONFIG_LIBUKNETDEV_STATS` to enable exporting stats via `ukstore`.

### Description of changes

This series adds `uknetdev` the capability to export aggregate and per-interface stats via `ukstore`.

GitHub-Depends-On: #939
GitHub-Depends-On: #938
GitHub-Supersedes: #226


craciunoiuc:
Looks as good as it can get 😉 

I'll let the others take a peek too

michpappas:
Fixed a checkpatch warning and rebased to the latest state of #939

lib/ukstore/include/uk/store.h
```diff
@@ -74,207 +74,272 @@ enum uk_store_entry_type {
...
+/**
+ * Looks up an entry in an object of a library.
+ *
+ * @param libid    owner library (ignored for dynamic entries)
+ * @param object   target object (NULL for static entries)
+ * @param entry_id the id of the entry to look up
+ * @return the matched entry or NULL
+ */
+const struct uk_store_entry *
+uk_store_get_entry(__u16 libid, struct uk_store_object *object_id, __u64 entry_id);
```
skuenzer: I remember we spoke about this function and I am wonder now if we should split this into two calls, like:
```c
const struct uk_store_entry *
uk_store_get_entry(__u16 libid, __u64 entry_id);
```
```c
const struct uk_store_object_entry *
uk_store_get_entry(struct uk_store_object *object_id, __u64 entry_id);
```
What do you think?

michpappas: Saw this comment just now, it belongs to #939 (this one is about netdev but I included the commits of #939 as a dependency). Will look at it tomorrow.

michpappas:
Rebased to latest state of in #939 and squashed the dependencies and marked more clearly in the commit title to avoid further confusion.

lib/ukstore/include/uk/store.h
michpappas: Addressed in #939

craciunoiuc (approved):
All good here. Thanks!

lib/uknetdev/include/uk/netdev_core.h
```diff
@@ -457,6 +461,46 @@ struct uk_netdev_einfo {
...
+	size_t frame;
+};
+
+struct uk_netdev_stats {
+	struct uk_netdev_tx_stats tx_m;
+	struct uk_netdev_rx_stats rx_m;
+};
+
+#ifdef CONFIG_LIBUKNETDEV_STATS
+int uk_netdev_stats_init(struct uk_netdev *dev);
```
skuenzer: We probably should not export this API call outside of the library. Instead, we should create a library-local header in the library base folder for cross-referencing between library source files. I even did not see `uk_netdev_stats_init()` in `exportsyms.uk` which confirms that the function is intended to be library-internal.

skuenzer (changes requested):

skuenzer:
Hum, I think we should remove the global/aggregated stats from the PR. I am not sure how useful they are and we should not implement them in this way. First of all we would need another lock with this implementation to protect the global stats (imagine multiple CPUs and multiple network devices) and secondly, this way would cause contention on the global stats (multi-core & multi-netif). I think that the aggregated stats should be computed within each of the global stats getter by iterating over all device stats and summing the corresponding values up. Assumption is that a getter is less often called versus the tx or rx function.
Ultimately and because of the same reasons, we should even do this on a device level, meaning collecting independent stats per device-queue. the ukstore per-device getters should then sum up the corresponding value from each queue stat. But let's keep it in this way as first version and improve the implementation afterwards.

skuenzer (approved):
All good! Thanks a lot for your work!

PR #944, opened 2023-06-13
build: Use -fcf-protection for GCC >= 8 or Clang >= 7

Krechals:
### Prerequisite checklist

### Base target

### Description of changes

The build fails on every app when gcc-8.0+ isn't used.

-fcf-protection was introduced in [gcc-8.0](https://gcc.gnu.org/gcc-8/changes.html) and in [clang-7.0](https://releases.llvm.org/7.0.0/tools/clang/docs/UsersManual.html).

The build still fails for clang-7.0, since `__asm__ goto` from `plat/kvm/x86/traps.c` was introduced in clang-9.0.

GitHub-Depends-On: https://github.com/unikraft/unikraft/pull/983

### DIFF
--- a/Makefile.uk
+++ b/Makefile.uk
@@ -11,7 +11,8 @@ COMPFLAGS-$(call have_gcc)	+= -fno-tree-sra -fno-split-stack -nostdinc
 ifneq ($(HAVE_STACKPROTECTOR),y)
 COMPFLAGS    += -fno-stack-protector
 endif
-COMPFLAGS    += -fcf-protection=none
+COMPFLAGS-$(call gcc_version_ge,8,0)	+= -fcf-protection=none
+COMPFLAGS-$(call have_clang)	+= -fcf-protection=none
 
 COMPFLAGS    += -Wall -Wextra
 COMPFLAGS-$(call have_clang)	+= -Wdocumentation -Wdocumentation-pedantic


support/build/Makefile.rules
```diff
@@ -39,17 +39,36 @@ define have_clang =
 $(if $(filter Clang,$(CC_NAME)),y,n)
 endef
 
+# test whether a compiler version is greater than or equal to the minimum requirement
+# compiler_version_ge $compiler_major,$compiler_minor
+define compiler_version_ge =
+$(shell if [ $(CC_VER_MAJOR) -gt $(1) ] || ([ $(CC_VER_MAJOR) -eq $(1) ] && [ $(CC_VER_MINOR) -ge $(2) ]) ; then echo -n y ; fi)
```
StefanJum: Isn't `[ $(CC_VER_MAJOR) -gt $(1) ] || ([ $(CC_VER_MAJOR) -eq $(1) ]` just an `[ $(CC_VER_MAJOR) -ge $(1) ]`?

StefanJum (changes requested):
This looks fine @Krechals, thank you, see the one comment.

About the build still failing with `clang 7`, should we add a global check for the `clang` version and fail if it's <= 9 (it could be only for `x86_64` or whatever setup fails to build)?
Something like [this](https://github.com/unikraft/unikraft/blob/staging/Makefile#L689-L691).

--- LATER EDIT: This will be done by #949 

I would also split the PR into 2 commits, one that adds the `-fcf-protection` flag and one that adds the `error_if_clang_version_lt` rule to the build system.

razvand:
@Krechals, could you please take a look in this PR to have it ready for upstream? See @StefanJum's comments. @mariasfiraiala , please also take a look.

mariasfiraiala (changes requested):
This works fine, thanks! @Krechals can you please address @StefanJum's comments in order to have this merged? I think they make sense.

razvand:
@Krechals , could you please rebase this PR on top of #983 ? Simply add the `-fcf-protection` specific checks commit and use the Clang-specific Makefile updates from #983 .

support/build/Makefile.rules
Krechals: Solved in https://github.com/unikraft/unikraft/pull/988

razvand:
@Krechals , could you rebase this, now that PR #983 is in?

mariasfiraiala (approved):
All good, thanks!

StefanJum (changes requested):
[This commit](https://github.com/unikraft/unikraft/commit/77221a5be2a715b10034d7ec0e729f9bef3ce3af) adds a minimum compiler version check, which is `9.0` for `clang`. So we only need the check for `gcc`.

Something like
```make
ifeq ($(call have_gcc),y)
COMPFLAGS-$(call gcc_version_ge,8,0)	+= -fcf-protection=none
else
COMPFLAGS-y += -fcf-protection=none
endif
```

or

```make
COMPFLAGS-$(call gcc_version_ge,8,0)	+= -fcf-protection=none
COMPFLAGS-$(call have_clang) += -fcf-protection=none
```

could be better?

StefanJum:
All good now @Krechals, just please remove the id `(101668)` from the commit authored-by and sign-off tags, we can merge after that.

StefanJum (approved):
All good, thanks.

razvand (approved):

PR #946, opened 2023-06-17
lib/ukallocbbuddy: Allocate physical memory in the buddy allocator 

razvanvirtan:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]

### Additional configuration

Tweak this threshold in the buddy allocator configuration menu:
 - `CONFIG_UKLIBALLOCBUDDY_THRESHOLD`

### Description of changes
Currently, the buddy allocator only handles virtual memory reservation in a huge memory area that is mapped at boot time, whereas physical memory is handled with demand paging.
    
This commit introduces physical memory allocation for the blocks that are provided by the buddy allocator, minimizing the number of page faults generated by subsequent accesses.

This is created on the top of the buddy allocator synchronization PR (that is why it is still in draft). For testing, please use [this branch](https://github.com/razvanvirtan/unikraft/tree/buddy_ukvmem) that removes some problematic commits (see the conversations from [here](https://discord.com/channels/762976922531528725/883603425076584448))

### DIFF
--- a/lib/ukallocbbuddy/Config.uk
+++ b/lib/ukallocbbuddy/Config.uk
@@ -5,3 +5,15 @@ menuconfig LIBUKALLOCBBUDDY
 	select LIBUKDEBUG
 	select LIBUKALLOC
 
+if LIBUKALLOCBBUDDY
+
+config UKLIBALLOCBUDDY_THRESHOLD
+	int "Maximum level for chunk physical allocation"
+	range 1 52
+	default 26
+	help
+		Maximum level of a chunk for which we allocate physical memory even
+		if the chunk will be split.
+
+endif
+
--- a/lib/ukallocbbuddy/Config.uk
+++ b/lib/ukallocbbuddy/Config.uk
@@ -1,6 +1,7 @@
-config LIBUKALLOCBBUDDY
+menuconfig LIBUKALLOCBBUDDY
 	bool "ukallocbbuddy: Binary buddy page allocator"
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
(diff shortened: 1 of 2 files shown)


lib/ukallocbbuddy/Config.uk
```diff
@@ -5,3 +5,15 @@ menuconfig LIBUKALLOCBBUDDY
 	select LIBUKDEBUG
 	select LIBUKALLOC
 
+if LIBUKALLOCBBUDDY
+
+config UKLIBALLOCBUDDY_THRESHOLD
```
skuenzer, line 10: Please always use `LIBUKALLOCBBUDDY_` prefix for all options.

skuenzer: I'd also prefer if this option is optional, so that `ukalloc_bbuddy` works still without physical allocations. Can you add an option to enable this behavior (it would be okay to have it default on if phys memory mgmt is available)? ... or maybe better: Could you provide another initializer function for `bbuddy` that enables phys. allocation (and a given initial virtual memory area) and allows handing over a physical context? `lib/ukboot` could then initialize `bbuddy` differently depending on what is available (only virtual memory, like linuxu (potentially solo5) or phys. memory mgmt is there). For both options, they should only be available if `ukvmem` or `ukfalloc` is available (depending which interface you use).

skuenzer (commented):
Hi @razvanvirtan, I am just adding some few comments to this PR. I wasn't doing a review. I'd also like to hear @marcrittinghaus's opinion.

PR #947, opened 2023-06-18
arch/x86: Fix interrupts on qemu-microvm/qboot

### LINKED ISSUE #516: No interrupts when running on QEMU's microvm machine type
### Describe the bug

A VM running on x86_64 QEMU/KVM does not receive any interrupts when using [microvm machine type](https://qemu.readthedocs.io/en/latest/system/i386/microvm.html) without acpi.

### Steps to reproduce

- A program
```c
#include <stdio.h>
#include <unistd.h>

int main(){
        sleep(1);
        printf("ok\n");
}
```

- Run with microvm machine time
```
% sudo qemu-system-x86_64 \
    -cpu host -enable-kvm \
    -M microvm,pit=on,pic=on,isa-serial=on,rtc=on,acpi=off \
    -kernel build/hello_kvm-x86_64 -nographic
```

- Output
```
Powered by
                 Hyperion 0.9.0~a858888
```
The program stops because sleep() never returns due to no timer interrupts.

### Expected behavior

The VM works correctly if we use q35 machine type (default).

```sh
sudo qemu-system-x86_64 \
    -M q35 -cpu host -enable-kvm \
    -kernel build/hello_kvm-x86_64 -nographic
```

- Output
```
SeaBIOS (version rel-1.16.0-0-gd239552ce722-prebuilt.qemu.org)


iPXE (http://ipxe.org) 00:03.0 CA00 PCI2.10 PnP PMM+07F910D0+07EF10D0 CA00



Booting from ROM..Powered by
                 Hyperion 0.9.0~a858888
ok
```

After some investigation, I found that microvm receives interrupts if we disable APIC explicitly (clear 11th bit of APIC base address).

- Patch
```diff
diff --git a/plat/kvm/x86/intctrl.c b/plat/kvm/x86/intctrl.c
index dc40555..9097c66 100644
--- a/plat/kvm/x86/intctrl.c
+++ b/plat/kvm/x86/intctrl.c
@@ -83,6 +83,17 @@ static void PIC_remap(int offset1, int offset2)

 void intctrl_init(void)
 {
+
+       __u32 eax, ebx, ecx, edx;
+       cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+#define CPUID_FEAT_EDX_APIC  (1 << 9)
+#define IA32_APIC_BASE_MSR 0x1B
+#define IA32_APIC_BASE_MSR_ENABLE 0x800
+       if (edx & CPUID_FEAT_EDX_APIC) {
+               uint64_t base = rdmsrl(IA32_APIC_BASE_MSR);
+               wrmsrl(IA32_APIC_BASE_MSR, (base & 0xfffff0000));
+       }
+
        PIC_remap(32, 40);
 }
```

With this patch, the VM also works fine with microvm machine type. However, I'm not really sure if this is the correct approach. The microvm machine type without acpi uses qboot instead of seabios and I guess this relates to the problem.
 

x86_64

kvm

### Relevant log output

_No response_

felixmoebius:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

Most firmware will put the APIC in "virtual wire mode" before starting the OS, which will raise interrupts received from the legacy PIC on LINT0 to the BSP. This is well explained [here (section 3.6.2.2)](https://web.archive.org/web/20121002210153/http://download.intel.com/design/archives/processors/pro/docs/24201606.pdf).
The relevant part where this is set up in SeaBIOS can be found [here](https://github.com/coreboot/seabios/blob/cd933454b5e3e1f86379a44b5ae1852c2a01a485/src/fw/smp.c#L128).

[qboot](https://github.com/bonzini/qboot) which is the default bios for QEMU's microvm machine type does not do this, so if we want to boot with qboot we need to do this setup ourselves.

I believe it is save to assume that we will boot in xAPIC mode (?), in which case we can do the necessary configuration while still in that mode, before switching to x2APIC mode for smp boot.

2e370e526ee9458fb77e512bd8628dd66dd54ec2 adds some MMIO read/write functions similar to the ones that already exist for arm64 to simplify interacting with the xAPIC MMIO interface.

fixes #516 

Booting on qemu-microvm works with the following command

```
qemu-system-x86_64 \
  -M microvm,rtc=on,pit=on,pic=on \
  -kernel "build/app-helloworld_qemu-x86_64" \
  -nographic
```

Booting on normal QEMU with qboot

```
qemu-system-x86_64 \
  -kernel "build/app-helloworld_qemu-x86_64" \
  -bios "/usr/share/qemu/qboot.rom" \
  -cpu host \
  -enable-kvm \
  -nographic
```

### DIFF
--- a/arch/x86/x86_64/include/uk/asm/lcpu.h
+++ b/arch/x86/x86_64/include/uk/asm/lcpu.h
@@ -114,9 +114,75 @@ struct __regs {
 #ifndef nop
 #define nop()   __asm__ __volatile__ ("nop" : : : "memory")
 #endif
-#endif /* !__ASSEMBLY__ */
 
-#ifndef __ASSEMBLY__
+static inline __u8 ioreg_read8(const volatile __u8 *address)
+{
+	__u8 value;
+
+	__asm__ __volatile__("movb %1, %0"
+		: "=q" (value) : "m" (*address) : "memory");
+
+	return value;
+}
+
+static inline __u16 ioreg_read16(const volatile __u16 *address)
+{
+	__u16 value;
+
+	__asm__ __volatile__("movw %1, %0"
+		: "=r" (value) : "m" (*address) : "memory");
+
+	return value;
+}
+
+static inline __u32 ioreg_read32(const volatile __u32 *address)
+{
+	__u32 value;
+
+	__asm__ __volatile__("movl %1, %0"
+		: "=r" (value) : "m" (*address) : "memory");
+
+	return value;
+}
+
+static inline __u64 ioreg_read64(const volatile __u64 *address)
+{
+	__u64 value;
+
+	__asm__ __volatile__("movq %1, %0"
+		: "=r" (value) : "m" (*address) : "memory");
+
+	return value;
+}
+
+static inline void ioreg_write8(const volatile __u8 *address,
+				__u8 value)
+{
+	__asm__ __volatile__("movb %0, %1"
+		: : "q" (value), "m" (*address) : "memory");
+}
+
+static inline void ioreg_write16(const volatile __u16 *address,
+				 __u16 value)
+{
+	__asm__ __volatile__("movw %0, %1"
+		: : "r" (value), "m" (*address) : "memory");
+}
+
+static inline void ioreg_write32(const volatile __u32 *address,
+				 __u32 value)
+{
+	__asm__ __volatile__("movl %0, %1"
+		: : "r" (value), "m" (*address) : "memory");
+}
+
+static inline void ioreg_write64(const volatile __u64 *address,
+				 __u64 value)
+{
+	__asm__ __volatile__("movq %0, %1"
+		: : "r" (value), "m" (*address) : "memory");
+}
+
 static inline unsigned long ukarch_read_sp(void)
 {
 	unsigned long sp;
--- a/plat/common/include/x86/apic.h
+++ b/plat/common/include/x86/apic.h
@@ -39,35 +39,58 @@
 
 #include <uk/assert.h>
 #include <uk/arch/limits.h>
+#include <uk/plat/common/lcpu.h>
 #include <x86/cpu.h>
 #include <x86/apic_defs.h>
 
 #include <errno.h>
 
-static inline int x2apic_enable(void)
+static inline int apic_enable(struct lcpu *this_lcpu)
 {
 	__u32 eax, ebx, ecx, edx;
+	__u32 x2apic_support;
 
 	/* Check for x2APIC support */
 	cpuid(1, 0, &eax, &ebx, &ecx, &edx);
-	if (!(ecx & X86_CPUID1_ECX_x2APIC))
-		return -ENOTSUP;
+	x2apic_support = ecx & X86_CPUID1_ECX_x2APIC;
 
-	/* Check if APIC is active */
+	/*
+	 * We expect to be started in xAPIC mode. This means firmware has not
+	 * irreversibly disabled the APIC and not yet enabled x2APIC mode.
+	 */
 	rdmsr(APIC_MSR_BASE, &eax, &edx);
-	if (!(eax & APIC_BASE_EN))
+	UK_BUGON(!(eax & APIC_BASE_EN));
+	if (x2apic_support)
+		UK_BUGON(eax & APIC_BASE_EXTD);
+
+	/* Set APIC software enable flag */
+	eax = ioreg_read32(APIC_MMIO_SVR);
+	eax |= APIC_SVR_EN;
+	ioreg_write32(APIC_MMIO_SVR, eax);
+
+	/*
+	 * Enable "virtual wire mode" to receive PIC interrupts on the BSP.
+	 * Local interrupts are masked after reset and will remain masked on
+	 * AP cores.
+	 */
+	if (lcpu_is_bsp(this_lcpu)) {
+		eax =  APIC_LVT_DELIVERY_MODE_EXTINT;
+		ioreg_write32(APIC_MMIO_LINT0, eax);
+
+		eax = APIC_LVT_DELIVERY_MODE_NMI;
+		ioreg_write32(APIC_MMIO_LINT1, eax);
+	}
+
+#ifdef CONFIG_HAVE_SMP
+	/* We currently need x2APIC mode for SMP boot */
+	if (!x2apic_support)
 		return -ENOTSUP;
 
 	/* Switch to x2APIC mode */
+	rdmsr(APIC_MSR_BASE, &eax, &edx);
 	eax |= APIC_BASE_EXTD;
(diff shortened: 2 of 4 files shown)


plat/common/include/x86/apic.h
```diff
@@ -44,30 +44,43 @@
...
-	cpuid(1, 0, &eax, &ebx, &ecx, &edx);
-	if (!(ecx & X86_CPUID1_ECX_x2APIC))
-		return -ENOTSUP;
-
-	/* Check if APIC is active */
+	/*
+	 * We expect to be started in xAPIC mode. This means firmware has not
+	 * irreversibly disabled the APIC and not yet enabled x2APIC mode.
+	 */
 	rdmsr(APIC_MSR_BASE, &eax, &edx);
```
dinhngtu, line 55: shouldn't this be an IO read for xAPIC?

plat/common/include/x86/apic.h
```diff
@@ -149,7 +162,6 @@ static inline void x2apic_ack_interrupt(void)
 }
 
 /* We only support x2APIC at the moment */
```
dinhngtu: this comment should perhaps be updated or removed

arch/x86/x86_64/include/uk/asm/lcpu.h
```diff
@@ -36,6 +36,7 @@
 
 #ifndef __ASSEMBLY__
 #include <uk/arch/types.h>
+#include <stdint.h>
```
dinhngtu: can Unikraft types be used instead of stdint.h types?

dinhngtu (commented):
Hi, thanks for the PR.

The patch works with a single vCPU on QEMU, QEMU-microvm and Firecracker. I haven't tested SMP.

I've left a few comments.

plat/common/include/x86/apic.h
```diff
@@ -44,30 +44,43 @@
...
-	wrmsr(APIC_MSR_BASE, eax, edx);
+	eax = APIC_LVT_TRIGGER_MODE_LEVEL | APIC_LVT_DELIVERY_MODE_NMI;
+	ioreg_write32(APIC_MMIO_LINT1, eax);
 
-	/* Set APIC software enable flag if necessary */
-	rdmsr(APIC_MSR_SVR, &eax, &edx);
-	if ((eax & APIC_SVR_EN) == 0) {
-		eax |= APIC_SVR_EN;
-		wrmsr(APIC_MSR_SVR, eax, edx);
+#ifdef CONFIG_HAVE_SMP
```
dinhngtu, line 84: this means a `!CONFIG_HAVE_SMP` build will succeed in `apic_enable` instead of failing with `-ENOTSUP` like before

plat/common/include/x86/apic.h
```diff
@@ -44,30 +44,43 @@
...
-
-	/* Check if APIC is active */
+	/*
+	 * We expect to be started in xAPIC mode. This means firmware has not
+	 * irreversibly disabled the APIC and not yet enabled x2APIC mode.
+	 */
 	rdmsr(APIC_MSR_BASE, &eax, &edx);
-	if (!(eax & APIC_BASE_EN))
-		return -ENOTSUP;
+	UK_BUGON(!(eax & APIC_BASE_EN) || (eax & APIC_BASE_EXTD));
```
dinhngtu: this check should only be done if the CPU has x2apic support

plat/common/x86/lcpu.c
```diff
@@ -60,13 +60,11 @@ __lcpuid lcpu_arch_id(void)
 
 int lcpu_arch_init(struct lcpu *this_lcpu)
 {
-#ifdef CONFIG_HAVE_SMP
 	int rc;
 
 	rc = apic_enable();
```
dinhngtu: we should have a comment here explaining why `apic_enable` must be used even if `CONFIG_HAVE_SMP` is not enabled

dinhngtu:
Just another note: the LINT# lines should be masked on all but 1 CPU, and that they should be configured as edge-triggered (going from the example in the MPS).

plat/common/include/x86/apic.h
felixmoebius: `!CONFIG_HAVE_SMP` did not call into `apic_enable` previously, or am I missing something here?

plat/common/include/x86/apic.h
felixmoebius: I would assume `APIC_BASE_EXTD` to be set to zero in case x2apic mode is not supported (at least that is what qemu seems to do), but it is probably better to not rely on it.

felixmoebius:
Sorry for taking so long to respond. I have pushed a v2 for review.

felixmoebius:
> Just another note: the LINT# lines should be masked on all but 1 CPU, and that they should be configured as edge-triggered (going from the example in the MPS).

Regarding edge vs level triggered mode for LINT0/LINT1: I am actually not sure which one is right. I have tested both and I don't see any difference. The current setup matches the one in SeaBIOS.

Regarding masking them on other CPUs: This was not the case previously, so I didn't make any changes in that regard. I'm not familiar enough with the SMP support in Unikraft to really test if that breaks anything.

plat/common/include/x86/apic.h
dinhngtu: you're right, I forgot about that

dinhngtu:
> Regarding edge vs level triggered mode for LINT0/LINT1: I am actually not sure which one is right. I have tested both and I don't see any difference. The current setup matches the one in SeaBIOS.

I read the SDM again and it turns out the trigger mode bit doesn't matter for NMI/ExtInt as both are always edge/level respectively.

> Regarding masking them on other CPUs: This was not the case previously, so I didn't make any changes in that regard. I'm not familiar enough with the SMP support in Unikraft to really test if that breaks anything.

For the LINT# lines, a single interrupt will end up signaling all CPUs. For NMIs this is common but I don't think it's the case for ExtInt.

The rest all look fine to me.

razvand:
@felixmoebius , could you please update the PR according to feedback?

felixmoebius:
> For the LINT# lines, a single interrupt will end up signaling all CPUs. For NMIs this is common but I don't think it's the case for ExtInt.

According to the intel manuals the mask bit in the LVT is set after reset. I think it should be enough to leave the LVT entries on AP cores as is without masking them explicitly.
I have made adjustments to only set up ExtINT and NMI interrupts on the BSP.

dinhngtu (approved):
That's everything from me, thanks.

razvand:
@felixmoebius , could you please rebase this PR?

razvand:
@felixmoebius , reminder here, could you please rebase the PR?

nderjung:
Bump on rebasing this please @felixmoebius  :)

razvand:
I've moved this to release 0.18. @mogasergiu will work on an updated version of it.

PR #950, opened 2023-06-20
lib: Move `atomic.h` arhitecture specific code to new `ukatomic`…

Starnox:
As part of the effort towards a platform re-arch, created a new design for an internal `uk` library that contains all the `arch` specific logic in different subdirectories and includes the right one using the build system.
Removed the `arch` prefix in all relevant locations.

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

Select `ukatomic` library in `KConfig` menu.

### Description of changes

As part of the effort towards a platform re-arch, created a new design
for an internal `unikraft` library that contains all the `arch` specific
logic in different subdirectories and includes the right one using
the build system.


michpappas (changes requested):
Hi @Starnox, please break down this PR into separate commits:
* One commit with that migration of atomic into own library.
* Separate commits for each updated calling library.

In general it is a good practice to update libraries in separate commits, as different libraries may have different maintainers. It also simplifies the review process in general.

Similarly for `plat`. Given that right now it's not so structured, you can make logic separations like `plat/common/lcpu`, `drivers/virtio`, and `plat/xen`.

Ditto for headers in `include/uk/`.

Starnox:
> Hi @Starnox, please break down this PR into separate commits:
> 
>     * One commit with that migration of atomic into own library.
> 
>     * Separate commits for each updated calling library.
> 
> 
> In general it is a good practice to update libraries in separate commits, as different libraries may have different maintainers. It also simplifies the review process in general.
> 
> Similarly for `plat`. Given that right now it's not so structured, you can make logic separations like `plat/common/lcpu`, `drivers/virtio`, and `plat/xen`.
> 
> Ditto for headers in `include/uk/`.

Done! Let me know if the separation is satisfactory.

lib/ukatomic/Config.uk
```diff
@@ -0,0 +1,3 @@
+menuconfig LIBUKATOMIC
+	bool "ukatomic: Atomic operations"
+	default y
```
michpappas, line 3: A `help` section with some additional info on the contents of this library would be useful.

michpappas (changes requested):
@Starnox thanks for the update, it looks good now 👍🏼 Some last nitpicks:

```
arch/lib: Move atomic operations logic from `arch` to new `ukatomic` library
```

Normally this denotes a hierarchy but there's no `lib` under `arch`. I think it's okay to just use `lib` here.

```    
    As part of the plat re-arch efforts, atomic operations will be implemented
    in an internal `uk` library.
```

I wouldn't mention the plat re-arch effort here, given that we there will be many more commits to come and certainly it would be too much to provide that context in all of them. A simple description of the changes (ie move headers into own library) would suffice.

In the rest of the commits:
```
    - Modify include path of `atomic.h`
    - Removed `arch prefix`
```
Since you use assertive voice in the commit title and the first bullet, you might as well use `Remove` instead of `Removed`.

Other than the above it looks great. Thanks again!

lib/ukatomic/Config.uk
```diff
@@ -0,0 +1,8 @@
...
+	help
+	  This option enables the atomic operations library, implemented
```
michpappas: This message can be improved a bit, as currently it provides either redundant information, or information not useful to the user. I explain:

> implemented for all supported architectures

Architecture support should be addressed by Kconfig dependencies, ie libraries not supported by a given architecture normally don't appear to the menu.

> This library is required by other libraries and applications, and should be enabled by default.

This is redundant as you already enable this by setting `default y`.

It's better to provide something as simple as `Atomic operations` or a more elaborate description like `Architecture-specific implementation of atomic operations. The set of operations is the same as the one provided by GCC's memory model aware atomic built-ins.`

michpappas (changes requested):
Hi @Starnox sorry that I'll be ultra-pedantic here but I see that this PR is used as base for others like #959, so we better get this right.

In the commit message:
```
 lib: Move atomic operations logic from arch to new ukatomic library

- Move headers from `arch` directory to `ukatomic`
- Move header from `include` directory to `ukatomic`
- Provide config option which is by default selected
``` 

These bullets essentially provide the same information as `git diff`. A more concentrated description is normally preferred:
```
lib: Move atomic operations from arch to new ukatomic library

Introduce `lib/ukatomic` and migrate `ukatomic` implementation from `arch`.
This library is enabled by default.
```

or...

```
lib: Move atomic operations from arch to new ukatomic library

Migrate `ukatomic` implementation from `arch` into newly introduced `lib/ukatomic`.
This library is enabled by default.
```

...or some similar wording of your choice :)

razvand:
@Starnox , could you please rebase this?

lib/ukatomic/Makefile.uk
```diff
@@ -0,0 +1,59 @@
+#  SPDX-License-Identifier: BSD-3-Clause
```
michpappas: Normally we don't put copyright in makefiles (at least in recently introduced libraries). You can have a look at recently introduced libraries like `uktimeconv`, `uknofault` or `ukvmem` for reference.

lib/ukatomic/arch/arm/include/uk/asm/atomic.h
```diff
@@ -44,7 +44,7 @@
  * @return The index of the least significant 1-bit of x, or if x is zero,
  *   the result is undefined. ffs(1)=0, ffs(3)=0, ffs(0x80000000)=31
  */
-static inline unsigned int ukarch_ffs(unsigned int x)
+static inline unsigned int uk_ffs(unsigned int x)
```
michpappas: Actually these are bit ops, not atomic, so they should not had been part of `atomic.h` in the first place. Perhaps you should move these into their own `ukbitops` library and leave `ukatomic` with what is in `include/uk/atomic.h`.

michpappas (changes requested):
Hi @Starnox, thanks for updating this. Besides the comments below, please squash these commits together. Also apologies for requesting that because I know we previously asked to have these in separate commits, but there have been discussions during re-arch among maintainers where people expressed the opinion that for re-arch it makes more sense to have these changes atomically (for changes not related to re-arch the rule of one commit per library remains). Thanks a lot! 🥂

lib/ukatomic/arch/arm/include/uk/asm/atomic.h
Starnox: I think this should have it's own separate PR after this is merged

Starnox:
> Hi @Starnox, thanks for updating this. Besides the comments below, please squash these commits together. Also apologies for requesting that because I know we previously asked to have these in separate commits, but there have been discussions during re-arch among maintainers where people expressed the opinion that for re-arch it makes more sense to have these changes atomically (for changes not related to re-arch the rule of one commit per library remains). Thanks a lot! 🥂

Hi! I rebased, updated and tested this PR with the latest changes from staging

lib/ukatomic/exportsyms.uk
```diff
@@ -0,0 +1,4 @@
+uk_ffs
```
michpappas: These are now part of `ukbitops`, so they can be removed from `ukatomic`.

lib/ukatomic/Makefile.uk
```diff
@@ -0,0 +1,5 @@
+$(eval $(call addlib_s,libukatomic,$(CONFIG_LIBUKATOMIC)))
+
+CINCLUDES-$(CONFIG_LIBUKATOMIC)	 		+= -I$(LIBUKATOMIC_BASE)/include
+ASINCLUDES-$(CONFIG_LIBUKATOMIC)	  	+= -I$(LIBUKATOMIC_BASE)/include
```
michpappas: Since `atomic.h` does not contain any directives that can be interpreted by the assembler, `ASINCLUDES` is strictly speaking not necessary.

lib/ukatomic/Config.uk
```diff
@@ -0,0 +1,5 @@
+menuconfig LIBUKATOMIC
+	bool "ukatomic: Atomic operations"
+	default y
+	help
+	  Support for atomic operations.
```
michpappas: The linux kernel uses two spaces to indent `help`, and personally I prefer that too, but the Unikraft convention is to use a tab here too, so let's stick to that 🙂

lib/ukatomic/include/uk/atomic.h
```diff
@@ -27,83 +27,81 @@
...
 
-#ifndef __UKARCH_ATOMIC_H__
-#define __UKARCH_ATOMIC_H__
+#ifndef __UK_ATOMIC_H__
+#define __UK_ATOMIC_H__
 
 #include <uk/arch/lcpu.h>
 
 #ifdef __cplusplus
 extern "C" {
```
michpappas, line 36: I'm leaving this here as github doesn't allow to comment on the closing brace: I think there is no reason to have `ACCESS_ONCE` and `UK_READ_ONCE` / `UK_WRITE_ONCE` outside the C++ guards. I checked on the commit message that introduced these changes and didn't see any specific reason mentioned, so I speculate it was an omission.

lib/ukbitops/Makefile.uk
```diff
@@ -0,0 +1,20 @@
+$(eval $(call addlib_s,libukbitops,$(CONFIG_LIBUKBITOPS)))
+
+CINCLUDES-$(CONFIG_LIBUKBITOPS)	 		+= -I$(LIBUKBITOPS_BASE)/include
+ASINCLUDES-$(CONFIG_LIBUKBITOPS)	  	+= -I$(LIBUKBITOPS_BASE)/include
```
michpappas: Similarly to `atomic.h`, I doubt anything defined in `bitops.h` can be used by asm code, so all the `ASINCLUDES` in this file can go away.

lib/ukbitops/Makefile.uk
```diff
@@ -0,0 +1,20 @@
...
+CXXINCLUDES-$(CONFIG_LIBUKBITOPS)		+= -I$(LIBUKBITOPS_BASE)/include
```
michpappas, line 4: Since you are defining `CXXINCLUDES` here, you should be adding C++ guards in the headers.

lib/ukbitops/Config.uk
```diff
@@ -0,0 +1,5 @@
+menuconfig LIBUKBITOPS
+	bool "ukbitops: Bit operations"
+	default y
+	help
+	  Support for common bit operations.
```
michpappas: Same as `libukatomic`, please indent with a tab here.

lib/ukbitops/Makefile.uk
```diff
@@ -0,0 +1,20 @@
...
+
+ifeq ($(CONFIG_ARCH_X86_64),y)
```
michpappas: You can reduce these to a couple of lines by declaring:
```
CINCLUDES-$(CONFIG_LIBUKBITOPS)	 		+= -I$(LIBUKBITOPS_BASE)/arch/$(CONFIG_UK_ARCH)/include
CXXINCLUDES-$(CONFIG_LIBUKBITOPS)		+= -I$(LIBUKBITOPS_BASE)/arch/$(CONFIG_UK_ARCH)/include
```

lib/ukatomic/include/uk/atomic.h
```diff
@@ -27,83 +27,81 @@
...
 
 /**
  * Perform an atomic increment/decrement operation and return the
  * previous value.
  */
-#define ukarch_inc(src) \
-	ukarch_fetch_add(src, 1)
-#define ukarch_dec(src) \
-	ukarch_fetch_sub(src, 1)
+#define uk_inc(src) \
```
michpappas, line 71: There is one instance of `ukarch_inc` remaining in `unixsock.c`:
```
lib/posix-unixsocket/unixsock.c:		if (!ukarch_inc(&data->listen.count))
```
Perhaps it was added after you did this work.

michpappas (changes requested):
Hi @Starnox thank you very much for making these changes. A few comments below, I think once these are resolved we can merge.

michpappas (approved):
@Starnox thanks a lot for this work.

razvand (approved):

PR #59, opened 2020-09-10
New boot logo and coloured output

skuenzer:
This series enables ANSI color support for lib/ukdebug.

PR #961, opened 2023-06-26
lib/ukboot: Zero out main thread TLS

alexhoppus:
The main thread TLS is allocated but not cleaned, what leads to undefined behavior.

### Prerequisite checklist

### Base target

 - Architecture(s): any
 - Platform(s): any
 - Application(s): any with musl-libc

### Additional configuration

 - `CONFIG_LIBMUSL=y`

### Description of changes

The allocation sequence for tls memory is:

uk_memalign => uk_do_memalign => a->memalign => uk_memalign_compat => uk_posix_memalign_ifpages => uk_palloc => bbuddy_palloc

it seems there is no memory cleaning in the call chain (the posix memalign also doesn't give such guarantees). Also memory is not cleaned manually after tls have been allocated. If memset memory with some pattern, the pthread_t tls related fields end up with this pattern value, what means no further initialization of pthread_t structure is performed.

**Actual issue with pthread_t fields corruption we encountered**

Arch: arm64 (but could be reproduced on x86 also)
How bug looks like from the perspective of observer: sometimes randomly main thread called pthread_exit() & was terminated.

To understand the actual issue we need to know, that during posix calls canceled pthread could terminate. For example, here is sem_wait call, which we invoked

```
9 int sem_timedwait(sem_t *restrict sem, const struct timespec *restrict at)      
 10 {                                                                               
 11 >---pthread_testcancel();                                                       
 12                                                                                 
 13 >---if (!sem_trywait(sem)) return 0;                                            
 14                                                                                 
 15 >---int spins = 100;                                                            
 16 >---while (spins-- && sem->__val[0] <= 0 && 
...
```
at the beginning it is checked whether we have pending cancellation request.

the cancellation flag is located together with canceldisable in pthread_t structure which is located in TLS. Both flags are checked before the thread is canceled.

```
 76 void __testcancel()                                                             
 77 {                                                                               
 78                                                                                 
 79 >---pthread_t self = __pthread_self();                                          
 82 >---if (self->cancel && !self->canceldisable)                                   
 83 >--->---__cancel();                                                             
 84 }
```

The main thread TLS is allocated using unikraft memalign call
```

275 >---/* Allocate a TLS for this execution context */                             
276 >---tls = uk_memalign(a,                                                        
277 >--->--->---  ukarch_tls_area_align(),                                          
278 >--->--->---  ukarch_tls_area_size());                                          
279 >---if (!tls)                                                                   
280 >--->---UK_CRASH("Failed to allocate and initialize TLS\n"); 
```

just after the memalign I have printed the buffer, the printing shows, that memory returned is not zeroed.
```
[    0.000000] ERR:  [libukboot] <boot.c @  287> 910328810d40a680 449088a8c6b900b1 30bdbb08b8020d9b 2209880dcc010c80   <---- here we have value which ends up in pthread->cancel flags
```

there is a chance, that the memory is initialized somewhere further, however when I put print in __testcancel it could be clearly seen that some of this uninitialized data (showed above) end up in pthread_t fields (cancel, canceldisable and others of course). Also if I memset tls memory with some pattern, I can clearly see this pattern in self->cancel and self->canceldisable

```
__pthread_testcancel:11
__testcancel:80 tid = 0 cancel 30bdbb08 canceldisable 1 self 0x409de028
```

You can use this example to reproduce/verify on your side.
To reproduce it you need:
1) musl
2) app-helloworld which calls sem_wait in main function
3) mem buf returned by uk_memalign for tls could be zeroed, but there is no guarantee, to emulate garbage fill it via pattern.
4) print cancel flag from __testcancel, you will see pattern there

### DIFF
--- a/lib/ukboot/boot.c
+++ b/lib/ukboot/boot.c
@@ -303,6 +303,7 @@ void ukplat_entry(int argc, char *argv[])
 	if (!tls)
 		UK_CRASH("Failed to allocate and initialize TLS\n");
 
+	memset(tls, 0, ukarch_tls_area_size());
 	/* Copy from TLS master template */
 	ukarch_tls_area_init(tls);
 	/* Activate TLS */


lib/ukboot/boot.c
```diff
@@ -303,6 +303,7 @@ void ukplat_entry(int argc, char *argv[])
 	if (!tls)
 		UK_CRASH("Failed to allocate and initialize TLS\n");
 
+	memset(tls, 0, ukarch_tls_area_size());
```
skuenzer, line 306: :thinking: This should be done by `ukarch_tls_area_init()` to avoid wiping areas which will be again initialized with values directly afterwards because they are copied from the TLS template. By just reading the code, I can see that the snippets are there, see: [arch/x86/x86_64/tls.c:190](https://github.com/unikraft/unikraft/blob/staging/arch/x86/x86_64/tls.c#L190) and [arch/x86/x86_64/tls.c:38](https://github.com/unikraft/unikraft/blob/staging/arch/x86/x86_64/tls.c#L38). Do you use the Arm version?

alexhoppus: @skuenzer  Hi. Thank you for response. The original issue appeared on arm64, but I don't think it is arch specific. Per my understanding, the code snippets you are referring to, they clear tbss, what is per thread bss section (uninitialized global data with __thread specifier goes there). However, I'm talking about pthread_t structure which resides in tls and used by pthreads. So these are different part of tls memory. Maybe I need to add a little bit more context how is that reproduced with pthreads.

skuenzer: Okay :thinking:, this should be done by musl that should provide `ukarch_tls_tcb_init()` (expected by [arch/x86/x86_64/tls.c:201](https://github.com/unikraft/unikraft/blob/staging/arch/x86/x86_64/tls.c#L201)): https://github.com/unikraft/lib-musl/blob/staging/__uk_init_tls.c#L177 . Can you check this function?

alexhoppus: @skuenzer maybe you mean `ukarch_tls_tcb_init()` -> `__uk_init_tls() `->` __uk_init_tp()` ? I don't see how `uk_thread_uktcb_init()` (https://github.com/unikraft/lib-musl/blob/staging/__uk_init_tls.c#L177) (looks like it is not called for init thread, but I'm not sure) appears in this chain. In this 3 functions I have mentioned above I don't see any tls cleaning.

jschlumpp: I think the problem is that the pthread structure is not initialized in the main thread. So zeroing the structure can fix some of the problems but in the end there are also list structures, which must not be initialized to zero. I think we could call `uk_thread_uktcb_init` for the main thread.

alexhoppus: @mschlumpp yes the problem is in non initialized `pthread_t` for main thread. Okay, then let me look how exactly `uk_thread_uktcb_init` can be used for init thread here.

skuenzer: @alexhoppus Sorry, my mistake. You are right, it is [lib-musl:__uk_init_tls.c:L219](https://github.com/unikraft/lib-musl/blob/staging/__uk_init_tls.c#L219) that then calls `__uk_init_tls()` and `__uk_init_tp()`. In fact, some clearing is missing here, so you should patch somewhere here.
Normally, `ukarch_tls_tcb_init()` is called for every thread before `uk_thread_uktcb_init()` is called for thread creations using `lib/uksched`-API calls. pthreads initialize the TCB manually behind musl's pthread APIs (however `ukarch_tls_tcb_init()` should still be visited) but the main thread only goes through `ukarch_tls_tcb_init()`. So, it makes sense to initialize a zeroed (or somewhat meaningful) pthread structure for fixing the problem.
Ideally, such a fix may also be an enabler for builds with musl but without threading (`lib/uksched`). As correct behavior, the pthread API could then not be used, the calls should fail because thread creation is not supported in such a case. As background: We ignored musl without `lib/uksched` for the first version but we intended to enable it later again.

alexhoppus: @skuenzer I think we should initialize pthread in `__uk_init_tp()` in the same way as pthread_create do. What do you think?

alexhoppus: I have opened new PR https://github.com/unikraft/lib-musl/pull/59 in musl

skuenzer: Yes, I agree, great. This means, we can close this PR and continue review with [PR #59](https://github.com/unikraft/lib-musl/pull/59) with lib-musl.

skuenzer:
Closing this PR because it is superseded by [PR #59](https://github.com/unikraft/lib-musl/pull/59) with lib-musl.

PR #964, opened 2023-06-30
drivers/arm-cca: Add libukrsi to support Arm CCA

zhxj9823:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): `app-helloworld`

### Additional configuration

This PR adds a new config `CONFIG_ARM64_FEAT_RME`. Enabling this config is required to run an application in the CCA platform.

### Description of changes

The changes include:

- The page table in `unikraft/plat/kvm/arm/bpt64.S` maps memory regions according to the memory layout of kvmtool.
  The device region is marked as unprotected.
- Kvmtool supports the ns16550 serial console, so it replaces the default pl011 in Unikraft.
  Besides, the ns16550 serial console in kvmtool uses byte-width registers.
- Kvmtool puts the device tree at the end of the memory region and passes its address in the `x0` register.
  So `_init_dtb_mem` excludes the dtb regions from the stack and heap area.
- `ukplat_rsi_init` function is called to initialize relevant features of the CCA platform.
  A `ukplat_rsi_setup_memory` function is added to the end of `_init_dtb_mem`, which marks the memory area as `RSI_RAM`.
- RSI commands required during the boot process are added.


lib/ukrsi/Config.uk
```diff
@@ -0,0 +1,13 @@
+menuconfig LIBUKRSI
+	bool "ukrsi: Realm Services Interface library"
+	default n
```
michpappas: `default n` is redundant, as `no` is the default option in Kconfig, so this can be removed.

lib/ukrsi/Config.uk
```diff
@@ -0,0 +1,13 @@
...
+	help
+		Provides functions for RSI.
```
michpappas: Please provide a more elaborate description here, eg:
```
Realm Services Interface (RSI) support. Enable if you are planning to run Unikraft as a CCA Realm.
```

lib/ukrsi/Config.uk
```diff
@@ -0,0 +1,13 @@
+menuconfig LIBUKRSI
+	bool "ukrsi: Realm Services Interface library"
```
michpappas: Please add `depends on ARM64_FEAT_RME` here. Although the entire menu is wrapped around the same condition (see below), it's still a good practice to have this dependency on the library configuration.

arch/arm/arm64/Config.uk
```diff
@@ -211,6 +211,17 @@ config ARM64_FEAT_RNG
 	  from a TRNG.
 endmenu
 
+menu "Armv9-A Extensions"
+	depends on MCPU_ARM64_NONE
+
+config ARM64_FEAT_RME
+	bool "Armv9 Realm Management Extension"
+	select LIBUKRSI
```
michpappas: `ARM64_FEAT_RME` essentially means "the platform I am running on is Armv9 that supports RME", and as such it may enable other functionality in the future, such as `RMI`. Because of that `select RSI` needs to be removed from here. Instead we need it to "unlock" the CCA features. See the next comment on how to do that.

Also `select PAGING` must also be moved to `ukrsi/Config.uk`, as it's a library dependency not a feature (ie `RME`) dependency. However, as part of the functionality provided by `librsi` also works witout paging, you can omit it altogether, and wrap around the calling code that depends on paging with `#if CONFIG_PAGING`.

lib/Makefile.uk
```diff
@@ -59,3 +59,4 @@ $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/vfscore))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukrust))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukreloc))
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukofw))
+$(eval $(call _import_lib,$(CONFIG_UK_BASE)/lib/ukrsi))
```
michpappas: This library needs to be moved into `drivers/arm-cca/`.

At the top of `drivers/Config.uk`:
```
+if ARM64_FEAT_RME
+menu "Arm CCA"
+source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-arm-cca.uk' -r '$(KCONFIG_DRIV_BASE)/arm-cca' -l '$(KCONFIG_DRIV_BASE)/arm-cca')"
+endmenu
+endif
+
```
also while you're there please remove the spaces between the `menu`/`endmenu` directives and the `source` in other subsystems to make them more compact.

In `drivers/Makefile.uk`:
```
+$(eval $(call _import_lib,$(UK_DRIV_BASE)/arm-cca))
```

lib/ukrsi/Makefile.uk
```diff
@@ -0,0 +1,13 @@
+$(eval $(call addlib_s,libukrsi,$(CONFIG_LIBUKRSI)))
+
+CINCLUDES-$(CONFIG_LIBUKRSI)   += -I$(LIBUKRSI_BASE)/arch/$(CONFIG_UK_ARCH)/include
```
michpappas: Please align as:
```
CINCLUDES-$(CONFIG_LIBUKRSI)   += -I$(LIBUKRSI_BASE)/arch/$(CONFIG_UK_ARCH)/include
CXXINCLUDES-$(CONFIG_LIBUKRSI) += -I$(LIBUKRSI_BASE)/arch/$(CONFIG_UK_ARCH)/include
```

lib/ukrsi/arch/arm64/include/arch/rsi.h
```diff
@@ -0,0 +1,215 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
```
michpappas: All new files must use the following copyright:
```
 */
```

lib/ukrsi/rsi.c
```diff
@@ -0,0 +1,311 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2023, Xingjian Zhang <zhxj9823@qq.com>.
```
michpappas: Please update the copyright here too.

lib/ukrsi/tests/test_rsi.c
```diff
@@ -0,0 +1,297 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
```
michpappas: Update copyright.

lib/ukrsi/arch/arm64/include/arch/rsi.h
michpappas, line 1: Since `ukrsi` is by definition exclusive to `arm64`, the header needs to be moved to `ukrsi/include/uk/rsi.h`

lib/ukrsi/arch/arm64/include/arch/rsi.h
```diff
@@ -0,0 +1,215 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __PLAT_COMMON_ARM64_RSI_H__
```
michpappas: Update these to `__UK_RSI_H__`

lib/ukrsi/arch/arm64/include/arch/rsi.h
```diff
@@ -0,0 +1,215 @@
...
+#define __PLAT_COMMON_ARM64_RSI_H__
+
+#define RSI_ATTESTATION_TOKEN_CONTINUE 0xC4000195
```
michpappas: All values provided by the spec need to live under `arch/`. Please move all `#defines` to `arch/arm64/include/uk/asm/rsi.h`.

lib/ukrsi/arch/arm64/include/arch/rsi.h
```diff
@@ -0,0 +1,215 @@
...
+#define RSI_SUCCESS 0x0
+#define RSI_ERROR_INPUT 0x1
+#define RSI_ERROR_STATE 0x2
+#define RSI_INCOMPLETE 0x3
+
+/* RIPAS values */
+#define RSI_EMPTY 0x0
+#define RSI_RAM 0x1
+#define RSI_DESTROYED 0x2
+
```
michpappas: Since you (correctly) export this header via `CXXINCLUDES` you need to add guards for C++ linkage:
```
#ifdef __cplusplus__
export "C"
{
#endif
...
#ifdef __cplusplus__
}
#endif
```

lib/ukrsi/rsi.c
```diff
@@ -0,0 +1,311 @@
...
+
+#include <arm/smccc.h>
+#include <arch/rsi.h>
+#include <errno.h>
+#include <uk/essentials.h>
+#include <uk/arch/paging.h>
+#include <uk/plat/common/sections.h>
+extern __u64 mem_base, mem_sz;
+
+#ifdef CONFIG_PAGING
```
michpappas: The current guidelines for guarding libraries are to use `#if` instead of `#ifdef` and comment as:
```
#if CONFIG_LIBXYZ

#endif /* CONFIG_XYZ */
```
and
```
#if CONFIG_LIBXYZ

#else /* !CONFIG_XYZ */ <--- same as "elif !CONFIG_XYZ"

#endif /* !CONFIG_XYZ */
```
The reverse applies for negation, ie:
```
#if !CONFIG_LIBXYZ

#endif /* !CONFIG_XYZ */
```
and
```
#if !CONFIG_LIBXYZ

#else /* CONFIG_XYZ */ <--- same as "elif CONFIG_LIBXYZ"

#endif /* CONFIG_XYZ */
```

Please update these everywhere in this series.

plat/kvm/arm/entry64.S
```diff
@@ -36,7 +36,11 @@
 #include <uk/config.h>
 #include <uk/reloc.h>
 
+#ifdef CONFIG_ARM64_FEAT_RME
```
michpappas: Here and everwhere else, make these conditional to `CONFIG_LIBUKRSI` instead.

plat/kvm/arm/setup.c
```diff
@@ -163,7 +168,13 @@ void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
 
 	fdt = (void *)bi->dtb;
 
+#ifdef CONFIG_ARM64_FEAT_RME
+	rc = uk_rsi_init();
```
michpappas: Please add a comment to explain why we defer console initialization when we are running as a realm.

lib/ukrsi/rsi.c
```diff
@@ -221,22 +222,24 @@ int uk_rsi_init(void)
...
 
-void uk_rsi_setup_memory(unsigned long base, unsigned long end)
+unsigned long
+uk_rsi_setup_memory(unsigned long base, unsigned long end, unsigned char ripas)
 {
 	unsigned long ret;
 	unsigned long new_base;
 
-	UK_ASSERT(base <= end);
+	// UK_ASSERT(base <= end);
```
michpappas: Why is this commented out?

lib/ukrsi/rsi.c
```diff
@@ -0,0 +1,311 @@
...
+#include <errno.h>
+#include <uk/essentials.h>
+#include <uk/arch/paging.h>
+#include <uk/plat/common/sections.h>
+extern __u64 mem_base, mem_sz;
+
+#ifdef CONFIG_PAGING
+#include <uk/arch/lcpu.h>
+#include <uk/plat/paging.h>
+#define RAM_BASE_ADDR 0x80000000
```
michpappas: This should not be hardcoded here. `plat/kvm/include/kvm-arm64/kvm-image.h` provides this value, you should take it from there. Also because this is conditional to `CONFIG_PAGING` the build fails if you disable paging.

Until `plat` is cleaned up you can do it as follows.

In `ukrsi/Makefile.uk` add:
```
LIBUKRSI_CINCLUDES-y += -I$(LIBKVMPLAT_BASE)/include
```

and in this file:
```
#include <kvm-arm64/image.h>
```

plat/common/bootinfo_fdt.c
```diff
@@ -10,12 +10,12 @@
 #include <libfdt.h>
 
 #define ukplat_bootinfo_crash(s)		ukplat_crash()
+__u64 mem_base, mem_sz;
```
michpappas: This is a hack. You need to get this information from bootinfo in `libukrsi`. You can update its prototype to:
```
void uk_rsi_init(struct uk_bootinfo *bi)
```
We follow this pattern in other drivers too.

michpappas (changes requested):
Hi @zhxj9823 overall good work 👍🏼  I did a first pass and added a some comments regarding the structure based on our latest guidelines (some of them are not documented yet, sorry).

To avoid having too many review comments I'll stop here and do another round after you address the review comments and you have pushed an update.

Besides the comments below, please squash 4b66d916 to afeff8d5. The improvements done during development process should be not be visible in the final commits.

Finally in the commit message:
```
    This commit changes the size of the boot stack to 16KB to accommodate
    larger objects used in the boot flow.
    
    It also adds uk_rsi_init function to initialize the RSI commands and
    move kvm_console_init after the initialization of paging.
```
As the main change is the initialization of `librsi` please reverse these two statements, ie:
```    
    This commit adds uk_rsi_init function to initialize the RSI commands and
    move kvm_console_init after the initialization of paging.

    It also changes the size of the boot stack to 16KB to accommodate
    larger objects used in the boot flow.
```
Also since you mention that you defer console initialization in the commit message, please add a brief explanation on the reason behind this change and mention that this is only when `libukrsi` is enabled.

drivers/Config.uk
```diff
@@ -1,11 +1,13 @@
...
 source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-bus.uk' -r '$(KCONFIG_DRIV_BASE)/ukbus' -l '$(KCONFIG_DRIV_BASE)/ukbus')"
-
 endmenu
 
 menu "Interrupt controller"
-
 source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-intctlr.uk' -r '$(KCONFIG_DRIV_BASE)/ukintctlr' -l '$(KCONFIG_DRIV_BASE)/ukintctlr')"
+endmenu
 
+if ARM64_FEAT_RME
```
michpappas: Please move this to the top of this file so that driver subsystems are alphabetically shown in menuconfig.

drivers/Makefile.uk
```diff
@@ -8,3 +8,4 @@ UK_DRIV_BASE := $(CONFIG_UK_BASE)/drivers
 
 $(eval $(call _import_lib,$(UK_DRIV_BASE)/ukbus))
 $(eval $(call _import_lib,$(UK_DRIV_BASE)/ukintctlr))
+$(eval $(call _import_lib,$(UK_DRIV_BASE)/arm-cca))
```
michpappas: Here too, before ukbus.

drivers/arm-cca/ukrsi/Config.uk
```diff
@@ -0,0 +1,13 @@
+menuconfig LIBUKRSI
+	bool "ukrsi: Realm Services Interface library"
+	depends on ARM64_FEAT_RME
+	select PAGING
```
michpappas: Regarding the dependency to `PAGING`: I was envisioning that we could also provide CCA capabilities even to platforms that for one reason or another choose to build Unikraft without paging. Such platforms would need to set up all protections at their boot page-tables (incl. devices). In that case, `select PAGING` can go away from here, and functions that use paging - along with callers of these functions - can be made conditional to `CONFIG_PAGING`.

drivers/arm-cca/ukrsi/Makefile.uk
```diff
@@ -0,0 +1,14 @@
+$(eval $(call addlib_s,libukrsi,$(CONFIG_LIBUKRSI)))
+
+CINCLUDES-$(CONFIG_LIBUKRSI)   += -I$(LIBUKRSI_BASE)/arch/$(CONFIG_UK_ARCH)/include
```
michpappas: Sorry perhaps I didn't phrase it clearly here. What I meant was to move this to the top-level arch, ie `unikraft/arch/arm/arm64/include/uk/asm/rsi.h`. The plan of re-arch is to keep anything architecture-defined (in the loose sense) there, and anything implementation defined in libraries. Also once that header is moved, this directive is no longer needed, as all headers under `arch/` are available to all libraries.

drivers/arm-cca/ukrsi/arch/arm64/include/uk/asm/rsi.h
```diff
@@ -0,0 +1,53 @@
...
+#define RSI_ERROR_INPUT 0x1
+#define RSI_ERROR_STATE 0x2
+#define RSI_INCOMPLETE 0x3
+
+/* RIPAS values */
+#define RSI_EMPTY 0x0
+#define RSI_RAM 0x1
+#define RSI_DESTROYED 0x2
+
+struct rsi_realm_config {
```
michpappas: Once you move this header to `arch/` it should only contain things defined by the spec. This means that `struct rsi_realm_config` needs to move to `drivers/arm-cca/ukrsi/uk/include/rsi.h` and `struct rsi_host_call_data` can be moved to `drivers/arm-cca/ukrsi/tests/test_rsi.c` as it's only used there.

Also the C++ linkage guards, should be in `drivers/arm-cca/ukrsi/uk/include/rsi.h` instead, as that's the header that is exported via `CXXINCLUDES`.

drivers/arm-cca/ukrsi/arch/arm64/include/uk/asm/rsi.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef __UKARCH_RSI_H__
+#define __UKARCH_RSI_H__
+
+#ifdef __cplusplus__
+extern "C"
+{
+#endif
+
+#define RSI_ATTESTATION_TOKEN_CONTINUE 0xC4000195
```
michpappas: The naming can be improved a bit if these are namespaced like `RSI_CMD_...`.

Also please add a couple of tabs to align the values of `#define`s, similarly for the others in this file.

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,167 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_RSI_H__
+#define __UK_RSI_H__
+
+#include <uk/asm/rsi.h>
+#include <uk/plat/common/bootinfo.h>
+
```
michpappas, line 11: As stated in the other comment, please add the C++ linkage guards here.

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,167 @@
...
+extern unsigned long unprotected_mask;
```
michpappas: Since this is a global it should be prefixed like the rest of public symbols, ie `uk_rsi_unprotected_mask`.

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Generate an attestation token
+ *
+ * @param addr IPA of the Granule to which the token will be written
+ * @param challenge Challenge value
+ * @param[out] size Attestation token size in bytes
+ *
+ * @return Command return status
+ */
+unsigned long uk_generate_attestation_token(unsigned long addr,
```
michpappas: `uk_rsi_generate_attestation_token`

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,167 @@
...
+					    unsigned long *size);
+
+/**
+ * Set the memory region as protected
+ *
+ * @param addr
+ * @param numpages
+ * @return 0 for success, other for failure
+ */
+int set_memory_protected(unsigned long addr, int numpages);
```
michpappas: `uk_rsi_set_memory_protected`

drivers/arm-cca/ukrsi/include/uk/rsi.h
michpappas, line 1: Please add an `exportsyms.uk` file in `ukrsi/` with all the public symbols (incl. globals).

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Set the memory region as unprotected
+ *
+ * @param addr
+ * @param numpages
+ * @return 0 for success, other for failure
+ */
+int set_memory_shared(unsigned long addr, int numpages);
```
michpappas: `uk_rsi_set_memory_shared`

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+#include <uk/arch/paging.h>
+#include <uk/plat/paging.h>
+#include <uk/plat/io.h>
+#include <uk/plat/common/sections.h>
+#include <uk/rsi.h>
+#include <kvm-arm64/image.h>
+
+unsigned long unprotected_mask;
+
+unsigned long uk_rsi_attestation_token_continue(unsigned long addr,
```
michpappas: Since the parameters and return value are not meant to be portable, and their size is defined by the spec, you can use fixed type variables like:
```
__u64 uk_rsi_attestation_token_continue(__u64 addr, __sz size)
```

Similarly for the other functions, where needed, both for parameters and local variables. Please notice the "where needed" part, there are cases where this is not required, eg when a function returns an errno value of course we use `int`. A good rule of thumb is whether a fixed size is defined by the spec.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+	struct smccc_args args = {0};
+
+	args.a0 = RSI_VERSION;
+
+	smccc_invoke(&args);
+
+	return args.a0;
+}
+
+void uk_rsi_init(struct ukplat_bootinfo *bi)
```
michpappas: Due to the nature of CCA it's important that `uk_rsi` functions return an error value, so that the handler can take a decision whether to proceed or not. In fact, I would even use `__check_result` here to force the caller to take action if `uk_rsi_init` fails. So this can be changed to:
```
int __check_result uk_rsi_init(struct ukplat_bootinfo *bi)
```
For the return values you can use somethning from `errno.h`, eg `-ENOTSUP`.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+{
+	struct rsi_realm_config __attribute__((aligned(PAGE_SIZE))) config;
```
michpappas: Here and everywhere else, you can instead use `__aligned(PAGE_SIZE)` from `essentials.h`.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+	const struct ukplat_memregion_desc *mrd;
+	__paddr_t pstart, pend, len;
+	unsigned long version;
+	unsigned long ret;
+	__u32 i;
+
+	version = uk_rsi_version();
+	uk_pr_info("RSI version %lu.%lu\n", (version >> 16),
```
michpappas: Maybe instead of magic numbers you can define in `arch/arm/arm64/include/uk/asm/rsi.h`:
```
+#define RSI_VERSION_MAJOR_SHIFT        16
+#define RSI_VERSION_MAJOR_MASK         0x7fff
+
+#define RSI_VERSION_MINOR_SHIFT        0
+#define RSI_VERSION_MINOR_MASK         0xffff

```

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+		   (version & 0xFFFF));
+
+	for (i = 0; i < bi->mrds.count; i++) {
+		mrd = &bi->mrds.mrds[i];
+
+		pstart = ALIGN_UP(mrd->pbase, __PAGE_SIZE);
```
michpappas: These relate a bit to #1060. We can discuss about it tomorrow.

drivers/arm-cca/ukrsi/arch/arm64/include/uk/asm/rsi.h
```diff
@@ -0,0 +1,53 @@
...
+ * RsiCommandReturnCode type
+ * Represents a return code from an RSI command.
+ */
+#define RSI_SUCCESS 0x0
+#define RSI_ERROR_INPUT 0x1
+#define RSI_ERROR_STATE 0x2
+#define RSI_INCOMPLETE 0x3
+
+/* RIPAS values */
+#define RSI_EMPTY 0x0
```
michpappas: Here also you can improve the naming of the RIPAS values a bit, ie `RSI_RIPAS_EMPTY` etc.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+	while (base != end) {
+		ret = uk_rsi_ipa_state_set(base, end, ripas, 0, &new_base);
+		base = new_base;
+		if (ret != RSI_SUCCESS)
+			break;
+	}
+	return ret;
+}
+
+void uk_rsi_setup_device(void)
```
michpappas: This function makes a very strong assumption about the platform, i.e. that the device memory is 1x page at the beginning of the RAM. To remove this dependency you can update the prototype to:
```
int uk_rsi_setup_device(__u64 base, __sz size)
```
and also add assertions that the base and size are page-aligned.

Then you can define the device regions in `plat/kvm/include/kvm-arm64/image.h`:
```
diff --git a/plat/kvm/include/kvm-arm64/image.h b/plat/kvm/include/kvm-arm64/image.h
index 8464601b..99f7b17a 100644
--- a/plat/kvm/include/kvm-arm64/image.h
+++ b/plat/kvm/include/kvm-arm64/image.h
@@ -7,11 +7,15 @@
 #define __KVM_ARM64_IMAGE_H__
 
 #if CONFIG_KVM_VMM_QEMU
-#define RAM_BASE_ADDR 0x40000000
+#define RAM_BASE_ADDR          0x40000000
+#define DEVICE_BASE_ADDR       0x08000000
+#define DEVICE_LENGTH          0x38000000
 #elif CONFIG_KVM_VMM_FIRECRACKER
-#define RAM_BASE_ADDR 0x80000000
+#define RAM_BASE_ADDR          0x80000000
+#define DEVICE_BASE_ADDR       0x00000000
+#define DEVICE_LENGTH          0x80000000
 #endif /* CONFIG_KVM_VMM_FIRECRACKER  */
 
-#define DTB_RESERVED_SIZE 0x100000
+#define DTB_RESERVED_SIZE      0x100000
 
 #endif /*  __KVM_ARM64_IMAGE_H__ */
```
And finally update `plat/common/memory.h` as:
```
diff --git a/plat/common/memory.c b/plat/common/memory.c
index 0960cdc7..384bb31f 100644
--- a/plat/common/memory.c
+++ b/plat/common/memory.c
@@ -33,6 +33,7 @@
  */
 #include <stdbool.h>
 #include <stddef.h>
+#include <kvm-arm64/image.h>
 #include <uk/plat/common/sections.h>
 #include <uk/asm/limits.h>
@@ -425,7 +426,7 @@ int ukplat_mem_init(void)
        ukplat_memregion_list_delete(&bi->mrds, 0);
 
 #if CONFIG_LIBUKRSI
-       uk_rsi_setup_device();
+       uk_rsi_setup_device(DEVICE_BASE_ADDR, DEVICE_LENGTH);
 #endif /* CONFIG_LIBUKRSI */
 
        return 0;
```

**Important:** Is marking all device memory as shared a hard requirement of CCA (ie so that the device can access that memory), or is it only required for devices emulated by the host? How do devices fit into the security model of the CCA? Is it guaranteed that device memory is always visible by the host?

Either way, I expect that this function will be deprecated later when we introduce `ioremap`, as we will be able to set all device attributes (inc. `PAGE_NS_SHARED`) from there.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+{
+	unsigned long ret;
+	unsigned long prot;
+	unsigned long phys_addr;
+
+	prot = PAGE_ATTR_PROT_RW;
+	ret = ukplat_page_set_attr(ukplat_pt_get_active(), addr, numpages, prot,
+				   0);
+
+	phys_addr = ukplat_virt_to_phys(addr);
```
michpappas: You need a cast here otherwise you'll get a GCC warning:
```
-       phys_addr = ukplat_virt_to_phys(addr);
+       phys_addr = ukplat_virt_to_phys((void *)addr);
```

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+	return ret;
+}
+
+int set_memory_shared(unsigned long addr, int numpages)
+{
+	unsigned long ret;
+	unsigned long prot;
+	unsigned long phys_addr;
+
+	phys_addr = ukplat_virt_to_phys(addr);
```
michpappas: cast here too

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,288 @@
...
+	ret = uk_rsi_setup_memory(phys_addr, phys_addr + (numpages * PAGE_SIZE),
+				  RSI_EMPTY);
+
+	prot = PAGE_ATTR_PROT_RW | PAGE_NS_SHARED;
```
michpappas: I have reservations on setting `PAGE_ATTR_PROT_RW` on `uk_rsi_set_memory_protected()` and `uk_rsi_set_memory_shared()` because we don't know the intentions of the user. Mapping shared memory `RO` is a very legitimate use-case. Moreover, before setting memory to protected / shared the caller will most likely update the protections anyway or perform some other action, like map the region even.

Given the above, I propose _one_ of the following:
* Remove the call to `ukplat_page_set_attr()` altogether, and expect (_and document_) that the user calls sets `PAGE_NS_SHARED` manually either via `ukplat_page_set_attr()` or via `ukplat_page_map()`. That reduces `uk_rsi_set_memory_protected()` and `uk_rsi_set_memory_shared()` to wrappers to `uk_rsi_setup_memory()`. The advantage is that it saves a call to `ukplat_page_set_attr()`.
*  We add an optional parameter `unsigned long attr` that allows the user to pass additional attributes. The downside is that we potentially make an additional call to `ukplat_page_set_attr()`, (which with the current direct-mapped based implementation is not too expensive anyway) but the interface remains simple.

drivers/arm-cca/ukrsi/tests/test_rsi.c
```diff
@@ -0,0 +1,363 @@
...
+ */
+
+#include <uk/assert.h>
+#include <uk/essentials.h>
+#include <uk/arch/paging.h>
+#include <uk/plat/io.h>
+#include <uk/rsi.h>
+#include <uk/test.h>
+
+#define FID_SMCCC_VERSION	0x80000000
```
michpappas: Except from `FID_INVALID`, all other `#define`s are already provided by `smccc.h`.

plat/kvm/arm/setup.c
```diff
@@ -163,7 +167,11 @@ void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
 
 	fdt = (void *)bi->dtb;
 
+#if CONFIG_LIBUKRSI
```
michpappas, line 173: Please remember to add the comment here, as stated in the previous feedback round.

plat/kvm/arm/setup.c
```diff
@@ -46,6 +46,10 @@
 #include <uk/arch/memtag.h>
 #endif /* CONFIG_HAVE_MEMTAG */
 
+#ifdef CONFIG_LIBUKRSI
```
michpappas: s/`#ifdef`/`#if` /

michpappas (changes requested):
Hi @zhxj9823, thanks for the updates. Here's another round of comments, I think we're getting closer 🚀

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,284 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <arm/smccc.h>
+#include <errno.h>
+#include <uk/essentials.h>
+#include <uk/arch/lcpu.h>
+#include <uk/arch/paging.h>
+#include <uk/plat/paging.h>
```
michpappas: Unlike `uk/arch/paging.h`, `uk/plat/paging.h` needs to be included conditionally to `CONFIG_PAGING`.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,284 @@
...
+		ret = uk_rsi_setup_memory(pstart, pend, RSI_RIPAS_RAM);
+		if (ret != RSI_SUCCESS)
+			return -ENOTSUP;
+	}
+
+	ret = uk_rsi_realm_config((__u64)&config);
+	if (ret != RSI_SUCCESS)
+		return -ENOTSUP;
+
+	uk_pr_info("RSI realm config ipa_width %lu, hash_algo %lu:\n",
```
michpappas: This causes a warning, `hash_algo` has to be `%u`

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,284 @@
...
+
+	/* set up device memory region */
+	pages = DIV_ROUND_UP(size, PAGE_SIZE);
+	prot = PAGE_ATTR_PROT_RW | PAGE_NS_SHARED;
+
+	return ukplat_page_set_attr(ukplat_pt_get_active(), base, pages, prot,
+				    0);
+}
+
+int uk_rsi_set_memory_protected(__u64 addr, int numpages)
```
michpappas: `numpages` cannot be negative so this should not be `int`, but `unsigned long` as defined in the prototype of `ukplat_page_set_attr()`, unless you prefer to use `__sz`, depending on how you like your interface to look like.

drivers/arm-cca/ukrsi/rsi.c
```diff
@@ -0,0 +1,284 @@
...
+	phys_addr = ukplat_virt_to_phys((void *)addr);
+	ret = uk_rsi_setup_memory(phys_addr, phys_addr + (numpages * PAGE_SIZE),
+				  RSI_RIPAS_RAM);
+	if (ret != RSI_SUCCESS)
+		return -ENOTSUP;
+
+	return 0;
+}
+
+int uk_rsi_set_memory_shared(__u64 addr, int numpages)
```
michpappas: `unsigned long numpages` or `__sz numpages`.

arch/arm/arm64/include/uk/asm/rsi.h
```diff
@@ -0,0 +1,41 @@
...
+#define RSI_RIPAS_RAM		0x1
+#define RSI_RIPAS_DESTROYED	0x2
+
+/* version shift */
+#define RSI_VERSION_MAJOR_SHIFT	16
+#define RSI_VERSION_MAJOR_MASK	0x7fff
+#define RSI_VERSION_MINOR_SHIFT	0
+#define RSI_VERSION_MINOR_MASK	0xffff
+
+
```
michpappas, line 39: There's an extra line here

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,188 @@
+ /* SPDX-License-Identifier: BSD-3-Clause */
```
michpappas: Please remove this leading whitespace

drivers/arm-cca/ukrsi/include/uk/rsi.h
```diff
@@ -0,0 +1,188 @@
...
+#include <uk/asm/rsi.h>
+#include <uk/plat/common/bootinfo.h>
+
+#ifdef __cplusplus__
+extern "C"
+{
+#endif
+
+extern __u64 uk_rsi_unprotected_mask;
+#define PAGE_NS_SHARED ((uk_rsi_unprotected_mask))
```
michpappas: Perhaps we should be using our own name for this. Since the purpose of this `#define` is to be used as a PTE bit, I propose to rename it to `PTE_RME_UNPROTECTED_BIT` to be consistent with the rest of the PTE definitions in `arch/arm/arm64/include/uk/asm/arch.h`. Notice that if `PAGE_NS_SHARED` had a fixed value it would be required to be in a header under `arch/arm/arm64/include/uk/asm/`, but headers under `arch/` must not depend on libraries, so we can't move it there.

Moreover, the right way to define an additional attribute in the paging API is to update `arch/arm/arm64/include/uk/asm/paging.h` as shown below (please double check whether that value is not used by another attribute, in case I missed something):
```diff
diff --git a/arch/arm/arm64/include/uk/asm/paging.h b/arch/arm/arm64/include/uk/asm/paging.h
index c13e8e9c..73108627 100644
--- a/arch/arm/arm64/include/uk/asm/paging.h
+++ b/arch/arm/arm64/include/uk/asm/paging.h
@@ -86,6 +86,8 @@ struct ukarch_pagetable {
 #define PAGE_ATTR_PROT_WRITE           0x02
 #define PAGE_ATTR_PROT_EXEC            0x04
 
+#define PAGE_ATTR_RME_UNPROTECTED      0x05
+
 #define PAGE_ATTR_TYPE_MASK            0x07
 #define PAGE_ATTR_TYPE_SHIFT           5
 ```

and then in `plat/common/include/arm/arm64/paging.h`:
```diff
diff --git a/plat/common/include/arm/arm64/paging.h b/plat/common/include/arm/arm64/paging.h
index fa904641..5f22a720 100644
--- a/plat/common/include/arm/arm64/paging.h
+++ b/plat/common/include/arm/arm64/paging.h
@@ -123,11 +123,13 @@ pgarch_pte_create(__paddr_t paddr, unsigned long attr, unsigned int level,
                pte |= PTE_ATTR_XN;
 
 #if CONFIG_LIBUKRSI
-       if (attr & PAGE_NS_SHARED)
-               pte |= PAGE_NS_SHARED;
+       if (attr & PAGE_ATTR_RME_UNPROTECTED)
+               pte |= PTE_RME_UNPROTECTED_BIT;
        else
-               pte &= (~PAGE_NS_SHARED);
-#endif /* CONFIG_LIBUKRSI */
+               pte &= (~PTE_RME_UNPROTECTED_BIT);
+#else /* !CONFIG_LIBUKRSI */
+       UK_ASSERT(!(attr & PAGE_ATTR_RME_UNPROTECTED));
+#endif /* !CONFIG_LIBUKRSI */
 
        switch (attr & (PAGE_ATTR_SHAREABLE_MASK <<
                PAGE_ATTR_SHAREABLE_SHIFT)) {
```

plat/kvm/arm/setup.c
```diff
@@ -163,7 +167,14 @@ void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
 
 	fdt = (void *)bi->dtb;
 
+	/* With ukrsi, we defer console init after paging initialization */
```
michpappas: This comment as is does not add any value as it describes what is exactly below, similarly to:
```c
/* Print hello */ <---- not useful
printf("Hello\n");
```
Useful comments explain why we do a behavior that is not immediately obvious to the reader of that code. Please update.

arch/arm/arm64/include/uk/asm/rsi.h
```diff
@@ -0,0 +1,41 @@
...
+#define RSI_CMD_MEASUREMENT_EXTEND		0xC4000193
+#define RSI_CMD_MEASUREMENT_READ		0xC4000192
+#define RSI_CMD_REALM_CONFIG			0xC4000196
+#define RSI_CMD_VERSION				0xC4000190
+
+/**
+ * RsiCommandReturnCode type
+ * Represents a return code from an RSI command.
+ */
+#define RSI_SUCCESS	0x0
```
michpappas: Please fix the indentation of the values. Use tabs and align all values to those of `RSI_CMD_*`

drivers/arm-cca/ukrsi/tests/test_rsi.c
```diff
@@ -0,0 +1,360 @@
...
+ */
+UK_TESTCASE(ukrsi, uktest_test_rsi_realm_config)
+{
+	struct rsi_realm_config config __align(PAGE_SIZE);
+	unsigned long ret;
+
+	ret = uk_rsi_realm_config(ukplat_virt_to_phys(&config));
+	UK_TEST_EXPECT_SNUM_EQ(ret, RSI_SUCCESS);
+
+	printf("Realm config: ipa_width %lu, hash_algo %lu\n", config.ipa_width,
```
michpappas: This causes a warning, `hash_algo` has to be `%u`

michpappas (changes requested):
Hi @zhxj9823 we're almost there! 🚀 I also ran the tests, it all looks good 👍🏼 

Besides the comments below, please consider the following:
* Run `support/scripts/checkpatch.uk --strict -g <commit-hash>` on every commit and address any issues.
* Add a brief description of the changes in the commit message of `4c19415a` and `7da74606`.
* Make sure the code compiles with and without `CONFIG_PAGING` and that in either case the PR does not introduce any warnings.

plat/kvm/arm/setup.c
```diff
@@ -159,6 +159,36 @@ static void __noreturn _ukplat_entry2(void)
 	ukplat_lcpu_halt();
 }
 
+#if CONFIG_LIBUKRSI && CONFIG_EARLY_UART_NS16550
+static int uk_rsi_setup_early_uart(struct ukplat_bootinfo *bi)
```
michpappas: This function is generally reusable for setting up early devices so it must be part of lib-ukrsi. Please update the parameters to the following and move it there.
```
__u64 uk_rsi_set_early_unprotected(__u64 base, __sz len)
```

The caller must provide their regular address (eg for early UART in kvmtool that would be `0x01000000`), and the function will return the unprotected IPA alias.

Regarding bootinfo, you can get a pointer within the function by calling `ukplat_bootinfo_get()`.

Once you have done that you need to update **both** the `ns16550` and `pl011` drivers to call that function to retrieve the unprotected IPA alias if `CONFIG_EARLY_UART && CONFIG_LIBUKRSI` are enabled. You can do that as follows:
1. Define two states for the `initialzied` global. `INITIALIZED` and `MAP_UNPROTECTED`.
2. Set the default to `MAP_UNPROTECTED` if `CONFIG_EARLY_UART && CONFIG_LIBUKRSI` 
3. Change all checks against `initialized` in the driver to calls to a new function `initialized()` that will do:
```c
static uint8_t initialized(void)
{
#if CONFIG_LIBUKRSI
    if (unlikely(initialized == MAP_UNPROTECTED)) {
        base = uk_rsi_map_early_unprotected();
        initialized = INITIALIZED;
    }
#endif
    return initialized;
}
```

This will allow you to derive the unprotected alias on the first call to UART without changing any of the existing semantics.

plat/kvm/arm/setup.c
```diff
@@ -159,6 +159,36 @@ static void __noreturn _ukplat_entry2(void)
...
+{
+	int rc;
+	struct ukplat_memregion_desc mrd = {0};
+
+	__u64 aliased_IPA = CONFIG_EARLY_UART_NS16550_BASE;
```
michpappas: As we discussed (many times) before, the user cannot provide the aliased IPA because that depends on the maximum address bits of the hardware, and therefore we can only derive the aliased IPA at runtime. Because of that, here you need to use `uk_rsi_unprotected_mask` and apply it to the user provided address.

Moreover, because you can't guarantee that no call to this function happens happen before `uk_rsi_init()`, you need to check if `uk_rsi_unprotected_mask` is has been initialized here, and if it hasn't to derived it at this moment. But also you use the early UART in `uk_rsi_init()` so it's best if you move the call to the top of that function:
``` diff
diff --git a/drivers/arm-cca/ukrsi/rsi.c b/drivers/arm-cca/ukrsi/rsi.c
index 02b7e48f..ff4869d3 100644
--- a/drivers/arm-cca/ukrsi/rsi.c
+++ b/drivers/arm-cca/ukrsi/rsi.c
@@ -199,12 +199,19 @@ int __check_result uk_rsi_init(struct ukplat_bootinfo *bi)
        __u64 version, ret;
        __u32 i;
 
+       /* call before early UART access */
+       ret = uk_rsi_realm_config((__u64)&config);
+       if (ret != RSI_SUCCESS)
+               return -ENOTSUP;
+
        version = uk_rsi_version();
        uk_pr_info(
            "RSI version %lu.%lu\n",
            (version >> RSI_VERSION_MAJOR_SHIFT) & RSI_VERSION_MAJOR_MASK,
            (version >> RSI_VERSION_MINOR_SHIFT) & RSI_VERSION_MINOR_MASK);
 
+       uk_pr_info("RSI realm config ipa_width %lu\n", config.ipa_width);
+
        for (i = 0; i < bi->mrds.count; i++) {
                mrd = &bi->mrds.mrds[i];
 
@@ -217,12 +224,6 @@ int __check_result uk_rsi_init(struct ukplat_bootinfo *bi)
                        return -ENOTSUP;
        }
 
-       ret = uk_rsi_realm_config((__u64)&config);
-       if (ret != RSI_SUCCESS)
-               return -ENOTSUP;
-
-       uk_pr_info("RSI realm config ipa_width %lu\n", config.ipa_width);
-
        /* set the mask of the unprotected bit */
        uk_rsi_unprotected_mask = (1UL) << (config.ipa_width - 1);
 
```

plat/kvm/arm/setup.c
```diff
@@ -159,6 +159,36 @@ static void __noreturn _ukplat_entry2(void)
...
+	__sz uart_mmio_len = __PAGE_SIZE; /* page-aligned */
```
michpappas: Here you should pass the user provided length. Also it's a good idea if you put some assertions at the top of the file to check whether the user-supplied base and len are page-aligned.

plat/kvm/arm/setup.c
```diff
@@ -159,6 +159,36 @@ static void __noreturn _ukplat_entry2(void)
...
+
+	/* Add a memreg for the Unprotected IPA alias of the early
+	 * UART to prevent from unmapping at paging_init().
```
michpappas: Please remove mentions to early uart from this comment, as this function will be generic.

plat/kvm/arm/setup.c
```diff
@@ -159,6 +159,36 @@ static void __noreturn _ukplat_entry2(void)
...
+	mrd.type  = UKPLAT_MEMRT_REALM;
+	mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE | UKPLAT_MEMRF_MAP;
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
+	if (unlikely(rc < 0)) {
+		uk_pr_err("Could not add free memory descriptor");
+		return rc;
+	}
+
+	ukplat_memregion_list_coalesce(&bi->mrds);
```
michpappas: It's not a good idea to ignore error code of `ukplat_memregion_list_coalesce()`. You can have this function return that.

michpappas (changes requested):
Hi @zhxj9823 please see the comments on early UART (and early device access in general). I will have a look at your changes on low level mapping later today.

PR #967, opened 2023-07-01
drivers: Move virtio from `plat/drivers/` to `drivers/` 

rares-miculescu:
### Prerequisite checklist

### Description of changes

Move virtio drivers into newly introduce drivers subsystem.
Virtio drivers should have been scattered around the drivers/ filesystem,
but an exception was made, to keep them altogether.


drivers/virtio/9p/Makefile.uk
```diff
@@ -0,0 +1,6 @@
+$(eval $(call adddriver_s,virtio9p,$(CONFIG_VIRTIO9P)))
+
+CINCLUDES-$(CONFIG_VIRTIO9P) += -I. -I$(UK_DRIVERS_BASE)/include/virtio
```
mogasergiu: Do not use `-I.`, notice that when you moved the headers you forgot to create the `include` directory. They were not part of the same directory as the driver itself. I would create an `include/` subdirectory in each driver's directory. Pretty sure headers are kept in the same directory as the driver when your intention is to make those header contents unexportable, private to the driver only. Thus to keep things consistent:

```
virtio
    ├── 9p
    │   ├── Config.uk
    │   ├── Makefile.uk
    │   ├── virtio_9p.c
    │   └── include
    │          └── virtio
    │                  ├── virtio_9p.h
```

Furthermore, try to have one include per line:
```
CINCLUDES-$(CONFIG_SOME_CONFIG) += -I$(SOME_INCLUDE1)
CINCLUDES-$(CONFIG_SOME_CONFIG) += -I$(SOME_INCLUDE2)
```
Same goes for the other headers.

drivers/virtio/Makefile.uk
```diff
@@ -0,0 +1,12 @@
...
+
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/9p))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/blk))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/bus))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/mmio))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/net))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/pci))
+$(eval $(call _import_driver,$(CONFIG_VIRTIO)/ring))
+
+
```
mogasergiu: No unnecessary newlines.

drivers/virtio/Config.uk
```diff
@@ -0,0 +1,2 @@
+config UKDRIVER_VIRTIO
+    bool
```
mogasergiu: Add a short description, `bool "Description"`.

plat/kvm/Makefile.uk
```diff
@@ -156,57 +156,61 @@ LIBKVMPF_CINCLUDES-$(CONFIG_ARCH_ARM_64)   += -I$(LIBKVMPLAT_BASE)/include
...
-LIBKVMVIRTIONET_ASINCLUDES-y   += -I$(UK_PLAT_COMMON_BASE)/include
-LIBKVMVIRTIONET_CINCLUDES-y    += -I$(UK_PLAT_COMMON_BASE)/include
-LIBKVMVIRTIONET_ASINCLUDES-y   += -I$(UK_PLAT_DRIVERS_BASE)/include
-LIBKVMVIRTIONET_CINCLUDES-y    += -I$(UK_PLAT_DRIVERS_BASE)/include
-LIBKVMVIRTIONET_SRCS-y +=\
-			$(UK_PLAT_DRIVERS_BASE)/virtio/virtio_net.c
+##LIBKVMVIRTIONET_ASINCLUDES-y   += -I$(LIBKVMPLAT_BASE)/include
+##LIBKVMVIRTIONET_CINCLUDES-y    += -I$(LIBKVMPLAT_BASE)/include
+##LIBKVMVIRTIONET_ASINCLUDES-y   += -I$(UK_PLAT_COMMON_BASE)/include
+##LIBKVMVIRTIONET_CINCLUDES-y    += -I$(UK_PLAT_COMMON_BASE)/include
```
mogasergiu: No need to comment these, just delete them.

mogasergiu (changes requested):
Hi, Rares! Thanks for the contribution, making great progress :smile:. Left some comments.

drivers/virtio/9p/Makefile.uk
razvand: @rares-miculescu , once you solve a comment, please the click the `Resolve conversation` button. Then re-request the review by clicking the double arrow button next to the reviewer name in the top right corner.

StefanJum (changes requested):
Thank you @rares-miculescu, few comments:

* Please add newlines at the end of the files that show up with the red sign
* Use the [Unikraft commit guidlines](https://unikraft.org/docs/contributing/submitting-changes/#commits) (add a component/selector, use backticks, add `/` at the end of directory names).
* As @mogasergiu said, do not use `-I.`. You can use `-I$(LIB...._BASE)/....`
* If the headers will be internal (not used by any other internal / external library), do not add them in the `CINCLUDES`, use them by including `"headername"` instead of `<headername>`. If they are not internal, and they will be used by other libraries, place them inside an `include/uk/`, `include/uk/component/` or some similar directory, and add them to the `CINCLUDES` with `-I$(LIB..._BASE)/include`. Final point is that `-I.` should not be needed.

drivers/ukvirtio/Config.uk
michpappas, line 1: This is now obsolete. Instead you need to add an entry under `drivers/Config.uk`:
```
+menu "Virtio"
+
+source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-virtio.uk' -r '$(KCONFIG_DRIV_BASE)/virtio' -l '$(KCONFIG_DRIV_BASE)/virtio')"
+
+endmenu
```

drivers/Makefile.uk
```diff
@@ -8,3 +8,4 @@ UK_DRIVERS_BASE := $(CONFIG_UK_BASE)/drivers
 
 $(eval $(call _import_lib,$(UK_DRIVERS_BASE)/ukbus))
 $(eval $(call _import_lib,$(UK_DRIVERS_BASE)/uktty))
+$(eval $(call _import_lib,$(UK_DRIVERS_BASE)/ukvirtio))
```
michpappas: Virtio is a bit of a special case. Normally the individual virtio drivers would be spread across different drivers subsystems, ie `virtio_net` under `drivers/uknetdev/`, `virtio_block` under `drivers/ukblkdev`, `virtio_mmio` and `virtio_pci` under `drivers/ukbus/` and so on. Consensus was to make an exception here and keep Virtio code together under `drivers/virtio/`. Notice that the name also indicates the deviation from the convention (`virtio` instead of `ukvirtio`).

drivers/ukvirtio/9p/Config.uk
```diff
@@ -0,0 +1,10 @@
+config VIRTIO_9P
+       bool "Virtio 9P device"
+       default y if LIBUK9P
+       default n
+       depends on LIBUK9P
+       imply VIRTIO_PCI if ARCH_X86_64
```
michpappas: AFAIK there is no hard requirement that there are any transport restrictions on Virtio devices on any architecture. If in doubt, you can always check with the Virtio spec. A `git blame` shows that this change was made by:
```
commit 1804ab7d2c3845afe22afa7ca8cd54c92b75fb20
Author: Stefan Teodorescu <stefanl.teodorescu@gmail.com>
Date:   Thu Oct 17 18:51:15 2019 +0300

    plat/kvm: Automatically select Virtio PCI device for Virtio devices.
    
    When selecting Virtio Net device or Virtio 9P, the PCI device wouldn't
    be automatically selected, even if it is a strong dependency.
    
    Signed-off-by: Stefan Teodorescu <stefanl.teodorescu@gmail.com>
    Reviewed-by: Sharan Santhanam <sharan.santhanam@neclab.eu>
```
I think what happened is that this was set at the wrong place - judging from the date - during Unikraft's prehistoric era. The right place to enable that is at the platform, ie platforms must `imply` the drivers they would like to enable by default, so this line can be dropped from here.

drivers/ukvirtio/9p/Config.uk
```diff
@@ -0,0 +1,10 @@
...
+       select VIRTIO_BUS
```
michpappas: Here we also need to select the parent library, ie `LIBUK9P`.
Also add `select VIRTIO`, I will provide more insight on that in a follow up comment on `virtio_bus`.

drivers/ukvirtio/9p/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_9p,$(CONFIG_LIBUKVIRTIO_9P)))
```
michpappas: `addplatlib_s` should be replaced with `addlib_s`. Also there are no dependencies from the platform anymore, so you can get rid of any `INCLUDES` directives. Finally, similarly to the rest of the drivers, any includes must be made conditional. With these in mind, you end up with:
```
+$(eval $(call addlib_s,libvirtio_9p,$(CONFIG_VIRTIO_9P)))
+
+LIBVIRTIO_9P_ASINCLUDES-$(CONFIG_VIRTIO_9P) += -I$(LIBVIRTIO_9P_BASE)/include
+LIBVIRTIO_9P_CINCLUDES-$(CONFIG_VIRTIO_9P)  += -I$(LIBVIRTIO_9P_BASE)/include
+
+# common virtio headers
+LIBVIRTIO_9P_ASINCLUDES-$(CONFIG_VIRTIO_9P) += -I$(UK_DRIV_VIRTIO_BASE)/include
+LIBVIRTIO_9P_CINCLUDES-$(CONFIG_VIRTIO_9P)  += -I$(UK_DRIV_VIRTIO_BASE)/include
+
+LIBVIRTIO_9P_SRCS-y += $(LIBVIRTIO_9P_BASE)/virtio_9p.c
```

drivers/ukvirtio/Makefile.uk
```diff
@@ -0,0 +1,15 @@
+################################################################################
+#
+# Driver registrations
+#
+################################################################################
+
+UK_DRIVERS_VIRTIO_BASE := $(UK_DRIVERS_BASE)/ukvirtio
```
michpappas: `DRIVERS` has now been replaced with `DRIV_` everywhere in this file, and as stated in a previous comment `ukvirtio` should be `virtio`.

drivers/ukvirtio/blk/Config.uk
```diff
@@ -0,0 +1,9 @@
+config VIRTIO_BLK
+	bool "Virtio Block Device"
+	default y if LIBUKBLKDEV
```
michpappas: This can go away, as we can't tell with certainty with drivers we should select if `blkdev` is enabled. This logic belongs to the platform configuration.

drivers/ukvirtio/blk/Config.uk
```diff
@@ -0,0 +1,9 @@
...
+	default n
+	depends on LIBUKBLKDEV
```
michpappas: This should instead be `select LIBUKBLKDEV`
Also add `select VIRTIO`, I will provide more insight on that in a follow up comment on `virtio_bus`.

drivers/ukvirtio/blk/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_blk,$(CONFIG_LIBUKVIRTIO_BLK)))
```
michpappas: Same comments as in `virtio/9p/Makefile.uk` apply here too.

drivers/ukvirtio/bus/Config.uk
```diff
@@ -0,0 +1,10 @@
+config VIRTIO_BUS
```
michpappas: So `virtio_bus` and `virtio_ring` are required for all Virtio devices, and it makes no sense to select them without selecting a device. Therefore we can hide them from `menuconfig` altogether by making them invisible options:
```
config VIRTIO_BUS
      bool
      default n
      select LIBUKBUS
      select LIBUKALLOC
      select LIBUKSGLIST
      select VIRTIO
```

Moreover, since we need both `virtio_bus` and `virtio_ring`, it makes sense to have config options of any Virtio device to enable these together.  We can do that by introducing yet another invisible variable, `CONFIG_VIRTIO` that is selected by all Virtio devices (see the earlier comments on `virtio_9p` and `virtio_blk`).

To do that, you should create `drivers/virtio/Config.uk` with:
```
+config VIRTIO
+       bool
```

drivers/ukvirtio/bus/Config.uk
```diff
@@ -0,0 +1,10 @@
...
+      bool  "Virtio bus driver"
+      default y
+      depends on (ARCH_X86_64 || ARCH_ARM_64)
+      depends on LIBUKBUS
+      select LIBUKALLOC
+      select LIBUKSGLIST
+      help
```
michpappas: Since this is now invisible this can go away.

drivers/ukvirtio/bus/Makefile.uk
```diff
@@ -0,0 +1,10 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_bus,$(CONFIG_LIBUKVIRTIO_BUS)))
```
michpappas: This can be simplified a lot. The comments on other virtio devices apply here too, but this library must be conditional to `CONFIG_VIRTIO` instead, for reasons discussed on a previous comment:
```
+$(eval $(call addlib_s,libvirtio_bus,$(CONFIG_VIRTIO)))
+
+# common virtio headers
+LIBVIRTIO_BUS_ASINCLUDES-$(CONFIG_VIRTIO) += -I$(UK_DRIV_VIRTIO_BASE)/include
+LIBVIRTIO_BUS_CINCLUDES-$(CONFIG_VIRTIO)  += -I$(UK_DRIV_VIRTIO_BASE)/include
+
+LIBVIRTIO_BUS_SRCS-y += $(LIBVIRTIO_BUS_BASE)/virtio_bus.c
```

drivers/ukvirtio/mmio/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_mmio,$(CONFIG_LIBUKVIRTIO_MMIO)))
```
michpappas: Same comment apply here as in `virtio_9p`, `virtio_net`, etc.

You also need to create a `Config.uk` for the MMIO transport:
```
+config VIRTIO_MMIO
+       bool "Virtio MMIO transport"
+       default n
+       select VIRTIO_BUS
+       select LIBUKOFW
+       select LIBUKINTCTLR_GIC
+       depends on ARCH_ARM_64
+       help
+              Virtio MMIO transport.
```

drivers/ukvirtio/net/Config.uk
```diff
@@ -0,0 +1,10 @@
+config VIRTIO_NET
+       bool "Virtio Net device"
+       default y if LIBUKNETDEV
```
michpappas: This should go

drivers/ukvirtio/net/Config.uk
```diff
@@ -0,0 +1,10 @@
...
+       default n
+       depends on LIBUKNETDEV
```
michpappas: This should become `select`

drivers/ukvirtio/net/Config.uk
```diff
@@ -0,0 +1,10 @@
...
+       imply VIRTIO_PCI if ARCH_X86_64
```
michpappas: This can go too.

drivers/ukvirtio/net/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_net,$(CONFIG_LIBUKVIRTIO_PCI)))
```
michpappas: Same as `virtio_9p` etc.

drivers/ukvirtio/pci/Config.uk
```diff
@@ -0,0 +1,8 @@
+config VIRTIO_PCI
+       bool "Virtio PCI device support"
+       default y if (VIRTIO_NET || VIRTIO_9P || VIRTIO_BLK)
```
michpappas: This should go, these devices can also use the MMIO transport.

drivers/ukvirtio/pci/Config.uk
```diff
@@ -0,0 +1,8 @@
+config VIRTIO_PCI
+       bool "Virtio PCI device support"
```
michpappas: A better description would be `Virtio PCI transport`

drivers/ukvirtio/pci/Config.uk
```diff
@@ -0,0 +1,8 @@
...
+       default y if (VIRTIO_NET || VIRTIO_9P || VIRTIO_BLK)
+       default n
+       depends on LIBUKBUS_PCI
```
michpappas: This should be `select`

drivers/ukvirtio/pci/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_pci,$(CONFIG_LIBUKVIRTIO_PCI)))
```
michpappas: Same changes as in `virtio_9p` etc.

drivers/ukvirtio/ring/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addplatlib_s,kvm,libukvirtio_ring,$(CONFIG_LIBUKVIRTIO_RING)))
```
michpappas: Same changes as in `virtio_bus`. Also conditional to `CONFIG_VIRTIO`:
```
+$(eval $(call addlib_s,libvirtio_ring,$(CONFIG_VIRTIO)))
+
+LIBVIRTIO_RING_ASINCLUDES-$(CONFIG_VIRTIO) += -I$(LIBVIRTIO_RING_BASE)/include
+LIBVIRTIO_RING_CINCLUDES-$(CONFIG_VIRTIO)  += -I$(LIBVIRTIO_RING_BASE)/include
+
+# common virtio headers
+LIBVIRTIO_RING_ASINCLUDES-$(CONFIG_VIRTIO) += -I$(UK_DRIV_VIRTIO_BASE)/include
+LIBVIRTIO_RING_CINCLUDES-$(CONFIG_VIRTIO)  += -I$(UK_DRIV_VIRTIO_BASE)/include
+
+LIBVIRTIO_RING_SRCS-y += $(LIBVIRTIO_RING_BASE)/virtio_ring.c
```

michpappas (changes requested):
@rares-miculescu please see the review comments below. Also since `virtio_pci` depends on `ukbus/pci`, please rebase this to #1024

drivers/virtio/mmio/Config.uk
```diff
@@ -0,0 +1,9 @@
+config VIRTIO_MMIO
+    bool "Virtio MMIO transport"
+    default n
+    select VIRTIO_BUS
```
michpappas: You also need to add `depends on LIBUKBUS` here, because:
1. This driver actually registers to `ukbus`
2. In menuconfig it should not be visible (and selectable) if `ukbus` is not enabled

drivers/virtio/9p/Config.uk
```diff
@@ -0,0 +1,9 @@
+config VIRTIO_9P
+       bool "Virtio 9P device"
```
michpappas: Please indent using tabs in all `Config.uk`

drivers/virtio/9p/Config.uk
```diff
@@ -0,0 +1,9 @@
...
+       default n
```
michpappas: In all `Config.uk`: `default n` is redundant, as `n`  it's the default behavior of Kconfig,  so it can be removed.

drivers/virtio/blk/Config.uk
```diff
@@ -0,0 +1,9 @@
+config VIRTIO_BLK
+	bool "Virtio Block Device"
```
michpappas: For consistency with the other options let's change this to "Virtio Block device"

drivers/virtio/blk/Config.uk
```diff
@@ -0,0 +1,9 @@
...
+	default n
+	select VIRTIO
+	depends on LIBUKBLKDEV
```
michpappas, line 3: There's no such rule, but since you have put the dependency first in the other configs, maybe you can be consistent and also place it also here before the `select`s.

drivers/virtio/9p/Makefile.uk
```diff
@@ -0,0 +1,10 @@
+$(eval $(call addlib_s,libvirtio_9p,$(CONFIG_VIRTIO_9P)))
+
+LIBVIRTIO_9P_ASINCLUDES-$(CONFIG_VIRTIO_9P) += -I$(LIBVIRTIO_9P_BASE)/include
```
michpappas: I had a discussion with @skuenzer and the convention is to add `ASINCLUDES` only when they are actually needed (ie when a header is included by asm code). So these can be removed everywhere in virtio Makefiles.

drivers/virtio/net/Config.uk
```diff
@@ -0,0 +1,8 @@
+config VIRTIO_NET
+       bool "Virtio Net device"
+       default n
+       depends on LIBUKNETDEV
+       select VIRTIO
```
michpappas: Should also add `select VIRTIO_BUS` here, as this - like all virtio drivers - registers into it.

drivers/virtio/bus/Config.uk
```diff
@@ -0,0 +1,7 @@
+config VIRTIO_BUS
+      bool "Virtio Bus Device"
```
michpappas: As virtio bus is Unikraft specific, perhaps a better description would be "Virtio ukbus driver".

drivers/virtio/Config.uk
```diff
@@ -0,0 +1,2 @@
+config VIRTIO
```
michpappas: Now that `VIRTIO_BUS` is visible, this global option no longer makes sense as it is only useful for `VIRTIO_RING`. So this can be simplified to:

1. Replace this with `ring/Config.uk` that contains a single invisible item named `VIRTIO_RING`
2. Update `ring/Makefile.uk` accordingly
3.  Remove `select VIRTIO` from driver and have them all `select VIRTIO_BUS`.
4. Have `VIRTIO_BUS` to `select VIRTIO_RING`

drivers/virtio/9p/Config.uk
```diff
@@ -0,0 +1,9 @@
+config VIRTIO_9P
```
michpappas: After a brief discussion with @skuenzer: Please update all `Config.uk` options in virtio to from `VIRTIO_` to `LIBVIRTIO_`. Remember to update your Makefiles and anywhere else where these maybe be used.

michpappas (changes requested):
Hi @rares-miculescu overall good job 😎 Besides the inline comments, please also see the items below:

Update `plat/kvm/Config.uk`. At the top level:
```
-imply LIBUKBUS_PCI if LIBUKBUS
+imply LIBVIRTIO_PCI if LIBUKBUS
+imply LIBVIRTIO_MMIO if (LIBUKBUS && ARCH_ARM_64)
```
Both QEMU and Firecracker support virtio block and virtio net, so at the top level also add:
```
+imply LIBVIRTIO_BLK if LIBUKBLKDEV
+imply LIBVIRTIO_NET if LIBUKNETDEV
```
QEMU also supports 9pfs, so in `KVM_VMM_QEMU `:
```
+imply LIBVIRTIO_9P if LIBUK9PFS
```
If all the Kconfig dependencies are set correctly, you should only need to select the top level libraries and the rest should be automatically enabled in menuconfig. 💪🏼 Please test 🤓 

In your commit messages:
```
    Notice that the migration of this driver is performed in two steps:
    
    1. Migrate the sources into driver verbatim (does not compile)
    2. Update the sources for this migration (next commit)
```
This is actually not true, as the top-most commit only updates include guards and no paths. The individual commits should be compiling successfully.

In the topmost commit:
```
    drivers/ukbus/platform: Update include guards
```
That would be `drivers/virtio` 😛

drivers/virtio/9p/Makefile.uk
```diff
@@ -0,0 +1,8 @@
+$(eval $(call addlib_s,libvirtio_9p,$(CONFIG_LIBVIRTIO_9P)))
+
+LIBVIRTIO_9P_CINCLUDES-$(CONFIG_LIBVIRTIO_9P)  += -I$(LIBVIRTIO_9P_BASE)/include
+
+# common virtio headers
+LIBVIRTIO_9P_CINCLUDES-$(CONFIG_LIBVIRTIO_9P)  += -I$(UK_DRIV_LIBVIRTIO_BASE)/include
```
michpappas: I unkerstand the motivation, but still I don't like very much that `virtio_bus.h` is under `virtio/include/` and not under  `virtio/bus/`. Similarly for `virtio_ring.h`. What you can do is:
1. Move `virtio_bus.h` to `bus/include/virtio/` and `virtio_ring.h` to `ring/incluce/virtio/`. IIRC virtio-ring does not depend on ukbus, so in that Makefile you will only need to include its own headers path.
2. Include these directories in every `Makefile.uk`

```diff
diff --git a/drivers/virtio/bus/Makefile.uk b/drivers/virtio/bus/Makefile.uk
index 28c6713b..eddc3505 100644
--- a/drivers/virtio/bus/Makefile.uk
+++ b/drivers/virtio/bus/Makefile.uk
@@ -2,5 +2,7 @@ $(eval $(call addlib_s,libvirtio_bus,$(CONFIG_LIBVIRTIO_BUS)))
 
 # common virtio headers
 LIBVIRTIO_BUS_CINCLUDES-y += -I$(UK_DRIV_LIBVIRTIO_BASE)/include
+LIBVIRTIO_BUS_CINCLUDES-y += -I$(UK_DRIV_LIBVIRTIO_BASE)/bus/include
+LIBVIRTIO_BUS_CINCLUDES-y += -I$(UK_DRIV_LIBVIRTIO_BASE)/ring/include
 
 LIBVIRTIO_BUS_SRCS-y += $(LIBVIRTIO_BUS_BASE)/virtio_bus.c
```

drivers/virtio/9p/Makefile.uk
```diff
@@ -0,0 +1,8 @@
+$(eval $(call addlib_s,libvirtio_9p,$(CONFIG_LIBVIRTIO_9P)))
+
+LIBVIRTIO_9P_CINCLUDES-$(CONFIG_LIBVIRTIO_9P)  += -I$(LIBVIRTIO_9P_BASE)/include
```
michpappas: Looking makefiles under `lib/` it seems that only public headers (ie `CINCLUDES`) need to be conditionally enabled. Private ones (`LIBXYZ-INCLUDES`) are private by default. So in all makefiles:
```diff
- LIBVIRTIO_9P_CINCLUDES-$(CONFIG_LIBVIRTIO_9P)  += -I$(LIBVIRTIO_9P_BASE)/include
+ LIBVIRTIO_9P_CINCLUDES-y  += -I$(LIBVIRTIO_9P_BASE)/include
```
PS: You can also do a simple experiment by looking at the generated `.cmd` files to confirm that behavior.
PS2: I know I initially asked you to change it, sorry about that, it only became apparent to me now too

drivers/virtio/9p/include/virtio/virtio_9p.h
```diff
@@ -30,8 +30,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef __PLAT_DRV_VIRTIO_9P_H
-#define __PLAT_DRV_VIRTIO_9P_H
+#ifndef __LIBVIRTIO_9P_H__
```
michpappas: All include guards should be `__VIRTIO_XX_H__` instead of `__LIBVIRTIO_XX_H__`. Consider that in the normal case we have a libraries named `libukxx` and the include guards are in the form of `__UK_XX_H__`. Consequently, for a library without the `uk`, like `libvirtioxx`, the include guards would be `__XX_H__` (for simplicity I omit additional parts of the include guard like `__VIRTIO_XX_FEAT_H__`).

drivers/virtio/mmio/Config.uk
```diff
@@ -0,0 +1,9 @@
+config LIBVIRTIO_MMIO
```
michpappas, line 4: Do I see spaces here? 🧐

drivers/virtio/ring/Config.uk
```diff
@@ -0,0 +1,2 @@
+config LIBVIRTIO_RING
+    bool
```
michpappas: Are these spaces? 🧐

plat/kvm/Config.uk
```diff
@@ -1,20 +1,24 @@
...
+	depends on (ARCH_X86_64 || (ARCH_ARM_64 && !HAVE_SYSCALL))
+	select LIBUKDEBUG
+	select LIBUKALLOC
+	select LIBUKTIMECONV
+	select LIBNOLIBC if !HAVE_LIBC
+	select LIBFDT if ARCH_ARM_64
+	select ELF64_TO_32 if ARCH_X86_64
+	select LIBUKRELOC if OPTIMIZE_PIE
+	select LIBUKOFW if ARCH_ARM_64
+	imply  LIBUKBUS_PCI if LIBUKBUS
```
michpappas: I spent big part of yesterday discussing with @skuenzer about Kconfig. It all started by the fact that nothing selects `libukbus` right now. There are some additional requirements, eg not to show virtio-pci or virtio-9p to the menu if they are not supported by a given VMM (eg Firecracker), not to show the virtio-bus and virtio transports when no device is selected, show only the relevant transports that are supported by a given VMM and so on.

Based on these:
1. Please add a commit to this PR to add `HAVE_PCI`. We need this so that we can hide PCI by platforms that don't have this ability.
```
diff --git a/drivers/ukbus/pci/Config.uk b/drivers/ukbus/pci/Config.uk
index 57387898..77b5f767 100644
--- a/drivers/ukbus/pci/Config.uk
+++ b/drivers/ukbus/pci/Config.uk
@@ -1,7 +1,11 @@
+config HAVE_PCI
+       bool
+
 config LIBUKBUS_PCI
        bool "PCI bus"
+       depends on HAVE_PCI
-       depends on LIBUKBUS
+       select LIBUKBUS
        help
                PCI bus driver for probing and operating PCI devices
 ```
 You will see below how that is used.

2. Add a commit in this series to have libofw to `select` its dependencies. The idea is to allows chained dependency resolution and use `depends on` only when there is a dependency loop, as we discussed before.
```
diff --git a/lib/ukofw/Config.uk b/lib/ukofw/Config.uk
index 8e2d1816..0e05ca25 100644
--- a/lib/ukofw/Config.uk
+++ b/lib/ukofw/Config.uk
@@ -1,4 +1,4 @@
 config LIBUKOFW
        bool "ukofw: Device tree helper functions"
        default n
-       depends on LIBFDT
+       select LIBFDT
```

3. In `virtio`. Add a top-level `Config.uk` as:
```
config VIRTIO_DEVICE
	bool
	default y if (LIBVIRTIO_9P || LIBVIRTIO_BLK || LIBVIRTIO_NET)
```
This will be used as syntactic sugar below to avoid typing all devices one by one. Yes yes I have asked you a few times to change things around, I'm sorry, things are still evolving. This is the last time I promise.

4. In `virtio-pci`. Have it select it's dependenies (ie `LIBUKBUS_PCI`) and make it available only to devices that support PCI (`HAVE_PCI`). Finally we also have the transports (`virtio-pci`, `virtio-mmio`) to no longer select `virtio-bus`, but instead make them invisible until `virtio-bus` is selected. You will see below under which conditions that happens.
```
diff --git a/drivers/virtio/pci/Config.uk b/drivers/virtio/pci/Config.uk
index 7f63e1ad..4afcc642 100644
--- a/drivers/virtio/pci/Config.uk
+++ b/drivers/virtio/pci/Config.uk
@@ -1,6 +1,7 @@
 config LIBVIRTIO_PCI
        bool "Virtio PCI transport"
-       depends on LIBUKBUS_PCI
-       select LIBVIRTIO_BUS
+       depends on VIRTIO_DEVICE
+       depends on HAVE_PCI
+       select LIBUKBUS_PCI
        help
                Support virtio devices on PCI bus
```
5. In `virtio-mmio`, similar changes like `virito-pci`, but since it doesn't have a parent we add `HAVE_MMIO` here:
```
diff --git a/drivers/virtio/mmio/Config.uk b/drivers/virtio/mmio/Config.uk
index 4bfce30c..85e66c7f 100644
--- a/drivers/virtio/mmio/Config.uk
+++ b/drivers/virtio/mmio/Config.uk
@@ -1,9 +1,11 @@
+config HAVE_MMIO
+       bool
+
 config LIBVIRTIO_MMIO
-    bool "Virtio MMIO transport"
-    depends on LIBUKBUS
-    depends on ARCH_ARM_64
-    select LIBVIRTIO_BUS
-    select LIBUKOFW
-    select LIBUKINTCTLR_GIC
-    help
-        Virtio MMIO transport.
+       bool "Virtio MMIO transport"
+       depends on ARCH_ARM_64
+       depends on VIRTIO_DEVICE
+       select LIBUKOFW
+       select LIBUKINTCTLR_GIC
+       help
+               Virtio MMIO transport.
```
6. In `virtio-bus` make it visible if at least one virtio device is enabled (`HAVE_VIRTIO`), have it `select` its dependency of `LIBUKBUS`. Here we don't need to use `depends on` as there is does no circular dependency, and in fact solves the problem that nothing selected `ukbus` until now.  Also enable by defaul tthe transports (mmio / pci) for those platforms that can use them (`HAVE_[MMIO,PCI]`).
```
diff --git a/drivers/virtio/bus/Config.uk b/drivers/virtio/bus/Config.uk
index d2f594d2..dc107a26 100644
--- a/drivers/virtio/bus/Config.uk
+++ b/drivers/virtio/bus/Config.uk
@@ -1,6 +1,9 @@
 config LIBVIRTIO_BUS
        bool "Virtio ukbus driver"
-       depends on LIBUKBUS
+       depends on VIRTIO_DEVICE
        select LIBUKALLOC
+       select LIBUKBUS
        select LIBUKSGLIST
        select LIBVIRTIO_RING
+       imply LIBVIRTIO_PCI if HAVE_PCI
+       imply LIBVIRTIO_MMIO if HAVE_MMIO
```

7. Now on to the devices. These don't need any changes, as they `select` their direct dependencies, and otherwise mind their own business. No dependency loops here either as they already `depend on` their API library (`ukblkdev`, `uknetdev`, `libuk9p`).

8. Finally the on the platform declare which features are supported by each VMM, and have the drivers selected if their corresponding feature is requested by the application.
```
diff --git a/plat/kvm/Config.uk b/plat/kvm/Config.uk
index aca9bef7..a14d5332 100644
--- a/plat/kvm/Config.uk
+++ b/plat/kvm/Config.uk
@@ -10,13 +10,11 @@ menuconfig PLAT_KVM
        select LIBUKOFW if ARCH_ARM_64
-       imply  LIBUKBUS_PCI if LIBUKBUS
-       imply  LIBVIRTIO_PCI if LIBUKBUS
-       imply  LIBVIRTIO_MMIO if (LIBUKBUS && ARCH_ARM_64)
-       imply  LIBVIRTIO_BLK if LIBUKBLKDEV
-       imply  LIBVIRTIO_NET if LIBUKNETDEV
-       imply  LIBUKBUS_PLATFORM if (LIBUKBUS && ARCH_ARM_64)
        select LIBUKINTCTLR if ARCH_X86_64
+       imply LIBUKBUS_PLATFORM if ARCH_ARM_64
+       imply LIBVIRTIO_9P if LIBUK9P
+       imply LIBVIRTIO_NET if LIBUKNETDEV
+       imply LIBVIRTIO_BLK if LIBUKBLKDEV
        help
                Create a Unikraft image that runs as a KVM guest
 
@@ -96,11 +94,13 @@ choice
 config KVM_VMM_QEMU
        bool "QEMU"
        imply LIBUKINTCTLR_GICV2
-       imply LIBVIRTIO_9P if LIBUK9P
+       select HAVE_PCI
+       select HAVE_MMIO
 
 config KVM_VMM_FIRECRACKER
        bool "Firecracker"
        select KVM_BOOT_PROTO_LXBOOT
+       select HAVE_MMIO
        imply LIBUKINTCTLR_GICV3
        depends on ARCH_X86_64 || ARCH_ARM_64
        
```

**Review by michpappas — CHANGES_REQUESTED** [review]
Hi @rares-miculescu this should be the last round 🚀 Please see the comments below.

File: drivers/virtio/9p/Makefile.uk (hunk 36 above)
Comment (rares-miculescu) [reply drivers/virtio/9p/Makefile.uk]: In order to be consistent, souldn't be better to write the include paths like this?
```
+LIBVIRTIO_BUS_CINCLUDES-y += -I$(LIBVIRTIO_BUS_BASE)/include
+LIBVIRTIO_BUS_CINCLUDES-y += -I$(LIBVIRTIO_RING_BASE)/include
```

File: drivers/virtio/ring/include/virtio/virtio_ring.h [hunk 42]
```diff
@@ -0,0 +1,196 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
razvand: Whee is this file (`virtio_ring.h`) coming from. It seems like a new file. Shouldn't it appear as moved from some place to this place? As is the case with `virtio_ring.c`?

drivers/virtio/bus/include/virtio/virtio_bus.h
```diff
@@ -0,0 +1,376 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
razvand: How did the `virtio_bus.h` file come to be? It should have been moved, similar to `virtio_bus.c`.

razvand (changes requested):

drivers/virtio/ring/Makefile.uk
```diff
@@ -0,0 +1,7 @@
+$(eval $(call addlib_s,libvirtio_ring,$(CONFIG_LIBVIRTIO_RING)))
+
+# common virtio headers
+LIBVIRTIO_RING_CINCLUDES-y  += -I$(UK_DRIV_LIBVIRTIO_BASE)/include
```
michpappas, line 4: You can align `+=` in these two lines.

michpappas (changes requested):
Hi @rares-miculescu thanks for the updates! 🚀 Overall it's good. Some final polishing and then we can finally merge 🎉 

Please always add a newline after include guards:

```
bash$ pcregrep -r -M -H '#define.*\n#include' drivers/virtio/
drivers/virtio//9p/include/virtio/virtio_9p.h:#define __VIRTIO_9P_H__
#include <uk/config.h>
drivers/virtio//net/include/virtio/virtio_net.h:#define __VIRTIO_NET_H__
#include <uk/config.h>
drivers/virtio//blk/include/virtio/virtio_blk.h:#define __VIRTIO_BLK_H__
#include <virtio/virtio_ids.h>
```

Please always add a newline between the copyright and the include guards:
```
bash$ pcregrep -r -M -H '\*\/.*\n#ifndef' drivers/virtio/
drivers/virtio//include/virtio/virtio_ids.h: */
#ifndef __VIRTIO_IDS_H__
drivers/virtio//include/virtio/virtio_types.h: */
#ifndef __VIRTIO_TYPES_H__
drivers/virtio//net/include/virtio/virtio_net.h: */
#ifndef __VIRTIO_NET_H__
drivers/virtio//blk/include/virtio/virtio_blk.h: */
#ifndef __VIRTIO_BLK_H__
drivers/virtio//blk/include/virtio/virtio_blk.h:/* Legacy feature bits */
#ifndef VIRTIO_BLK_NO_LEGACY
drivers/virtio//mmio/include/virtio/virtio_mmio.h: */
#ifndef __VIRTIO_MMIO_H__
```

Some improvements in your commit messages:
```
    lib/ukofw: Fix configuration dependencies
    
    Replace `depends on` with `select` for `LIBFDT`.
    This way we allow chained dependency resolution
    and use `depends on` only when we have a
    dependency loop.
```
no need to state the general rule, it's enough to say:
```
    lib/ukofw: Fix configuration dependencies
    
    Replace `depends on` with `select` for `LIBFDT` to allow
    chained dependency resolution.
```
Also in the PCI commit: 
```
    drivers/ukbus: Add `HAVE_PCI`
    
    Add `HAVE_PCI`, so we can hide PCI from platforms
    that don't have this feature.
```
you can say:
```
    drivers/ukbus: Add `HAVE_PCI`
    
    Add `HAVE_PCI` to limit menuconfig PCI options to
    platforms that provide this feature.
```

michpappas (approved):
Thanks a lot @rares-miculescu, it now looks good. 🥳 Also thanks for the patience with the conflicting requests from time to time. It's was quite a ride, but I think overall this PR helped us to define many parts we didn't cover until now.

razvand (approved):

PR #973, opened 2023-07-05
lib/nolibc,lib/isrlib: Fix strncpy

haesbaert:
strncpy will produce different results for an empty string depending on the preceding value of dst:

char buf[4];

bzero(buf, sizeof(buf));
buf[0] = 0;
buf[1] = 9;
strncpy(&buf[1], "", 3);

"clen" will be 0 since strnlen("") is 0, which will then look into dst[-1] (buf[0]) to decide if it terminates dst or not.

Instead of working around, replace strncpy for a conformant version, taken from OpenBSD libkern, as a bonus we zero out the remaining buffer.

Easier to test by renaming strncpy to something else otherwise the compiler will inline it most of the small cases.

### Prerequisite checklist

### DIFF
--- a/lib/nolibc/string.c
+++ b/lib/nolibc/string.c
@@ -152,17 +152,21 @@ size_t strnlen(const char *str, size_t len)
 
 char *strncpy(char *dst, const char *src, size_t len)
 {
-	size_t clen;
+	if (len != 0) {
+		char *d = dst;
+		const char *s = src;
 
-	clen = strnlen(src, len);
-	memcpy(dst, src, clen);
+		do {
+			if ((*d++ = *s++) == 0) {
+				/* NUL pad the remaining n-1 bytes */
+				while (--len != 0)
+					*d++ = 0;
+				break;
+			}
+		} while (--len != 0);
+	}
 
-	/* instead of filling up the rest of left space with zeros,
-	 * append a termination character if we did not copy one
-	 */
-	if (clen < len && dst[clen - 1] != '\0')
-		dst[clen] = '\0';
-	return dst;
+	return (dst);
 }
 
 char *strcpy(char *dst, const char *src)
--- a/lib/isrlib/string.c
+++ b/lib/isrlib/string.c
@@ -152,17 +152,21 @@ size_t strnlen_isr(const char *str, size_t len)
 
 char *strncpy_isr(char *dst, const char *src, size_t len)
 {
-	size_t clen;
+	if (len != 0) {
+		char *d = dst;
+		const char *s = src;
 
-	clen = strnlen_isr(src, len);
-	memcpy_isr(dst, src, clen);
+		do {
+			if ((*d++ = *s++) == 0) {
+				/* NUL pad the remaining n-1 bytes */
+				while (--len != 0)
+					*d++ = 0;
+				break;
+			}
+		} while (--len != 0);
+	}
 
-	/* instead of filling up the rest of left space with zeros,
-	 * append a termination character if we did not copy one
-	 */
-	if (clen < len && dst[clen - 1] != '\0')
-		dst[clen] = '\0';
-	return dst;
+	return (dst);
 }
 
 char *strcpy_isr(char *dst, const char *src)


lib/nolibc/string.c
```diff
@@ -152,17 +152,21 @@ size_t strnlen(const char *str, size_t len)
...
 {
-	size_t clen;
-
-	clen = strnlen(src, len);
-	memcpy(dst, src, clen);
+	if (len != 0) {
+		char *d = dst;
+		const char *s = src;
+
+		do {
```
RaduNichita, line 159: I have a suggestion here: could we move the zeroing part of the destination after copying the effective data?

Something like:

```c
 while (--len != 0 && (*d++ = *s++) != 0));
 /* NUL pad the remaining n-1 bytes */
 while (--len != 0)
	*d++ = 0;
```

Would it affect the performance in some way?

RaduNichita (changes requested):
Nice work, @haesbaert! I have provided a suggestion regarding the code. 

Furthermore, do you know if the openBSD license is compatible with BSD-3-Clause?  Also, you should add a link to the openBSD implementation of the functions in your commit message.

lib/nolibc/string.c
haesbaert: I don't think it would make any difference, it's worth noting that this is a somewhat inneficcient strncpy, as it copies things in bytes instead of words, an efficient strncpy is a bit of a monstruosity (if you check the ones included in libc's, they're scary!) and don't think it belongs in nolibc

haesbaert:
> Nice work, @haesbaert! I have provided a suggestion regarding the code.
> 
> Furthermore, do you know if the openBSD license is compatible with BSD-3-Clause? Also, you should add a link to the openBSD implementation of the functions in your commit message.

It's the same: https://github.com/openbsd/src/blob/master/sys/lib/libkern/strncpy.c, I'll change the commit message to include it

lib/nolibc/string.c
RaduNichita: I would go for the shortest approach, since it's more easier to follow, in my opinion. @razvand  do you have any opinion regarding this?

razvand: @RaduNichita , @haesbaert , I don't have a strong opinion. @haesbaert , you should decide what you think is best and that would be it.

lib/nolibc/string.c
```diff
@@ -152,17 +152,21 @@ size_t strnlen(const char *str, size_t len)
...
+		} while (--len != 0);
+	}
 
-	/* instead of filling up the rest of left space with zeros,
-	 * append a termination character if we did not copy one
-	 */
-	if (clen < len && dst[clen - 1] != '\0')
-		dst[clen] = '\0';
-	return dst;
+	return (dst);
```
RaduNichita, line 169: Why do you need brackets here?

RaduNichita (commented):
Tested and it has the expected behavior.

I left one question more out of curiosity, I will approve after that

lib/nolibc/string.c
haesbaert: Heya, sorry for the delay I was on holidays.
I'd keep as it is, since it's tested and I don't see an advantage in changing it.

lib/nolibc/string.c
haesbaert: oh the parenthesis are not needed, it's just common in BSD KNF, feel free to remove if wanted.
It's somewhat common in the unikraft codebase though:
```
sam:unikraft: grep -r return\ \( . |wc -l
388
````

Starnox (approved):
Nice catch! Works for me properly!

Reviewed-by: Eduard-Florin Mihailescu [mihailescu.eduard@gmail.com](mailto:mihailescu.eduard@gmail.com)

lib/nolibc/string.c
skuenzer: Okay, we'll remove the brackets while upstreaming.

skuenzer (approved):

PR #979, opened 2023-07-13
lib/vfscore: Implement individual volume automounting

mogasergiu:
Provide a command-line argument `vfs.fstab` that is meant to contain a list of whitespace separated strings with the following format:
```
	vfs.fstab=[
		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
		...
	]
```

The core will parse the provided strings as volume information and attempt to mount them accordingly.

### Prerequisite checklist

### Base target


lib/vfscore/Makefile.uk
```diff
@@ -10,6 +10,7 @@ LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/file.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/stdio.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/mount.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/vnode.c
+LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/volumes.c
```
skuenzer: Can we name the file `automount.c`? I think as a little clean-up it makes sense to rename `rootfs.c` and then add your code on top. Then we have it within one file.

lib/vfscore/volumes.c
```diff
@@ -0,0 +1,59 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <string.h>
+#include <sys/mount.h>
+#include <uk/init.h>
+#include <uk/libparam.h>
+
+#define UK_MAX_VFSCORE_VOLUMES_COUNT			64
```
skuenzer: Can we make this available via menu config? Also make it configurable if automount is available?

lib/vfscore/volumes.c
```diff
@@ -0,0 +1,59 @@
...
+		rc = mount(sdev, path, drv, 0, opts);
+		if (unlikely(rc))
+			uk_pr_err("Failed to mount %s: error %d\n", sdev, rc);
+	}
+
+	return 0;
+}
+
+/* Should be preferrably called after vfscore's rootfs is mounted (prio 4) */
+uk_rootfs_initcall_prio(vfscore_volumes_mount, 5);
```
skuenzer: great! It is a bit clearer maybe if we have the registrations in the same file. We could also include your code to the rootfs mount code.

lib/vfscore/Makefile.uk
mogasergiu: Are you sure? Because that would make volumes depend on `CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS`. And there is a case where you may want to have atuomatic mount of volumes without having the `rootfs.c` functionality.

skuenzer: Both true but you could just guard the ROOTFS code in the file and keep the source file part of the build, independent of the switch, right? My point is that it is easier to lookup any automount related code and that it is close together from an organization perspective.

mogasergiu: So guard two thirds of the file? Does it not look like bad design since the separation is so big to not have them in separate files so that the separation is transparent from the source code point of view?

StefanJum (commented):
This seems to work fine @mogasergiu, I'll add the tag after you resolve Simon's comments.

lib/vfscore/volumes.c
```diff
@@ -0,0 +1,59 @@
...
+
+static char *vfscore_volumes[UK_MAX_VFSCORE_VOLUMES_COUNT];
+
+UK_LIBPARAM_PARAM_ARR_ALIAS(automount, &vfscore_volumes, charp,
+			    UK_MAX_VFSCORE_VOLUMES_COUNT, "VFSCore Volumes");
+
+/**
+ * Expected command-line argument format:
+ *	vfs.automount=[
+ *		"<source_device>:<mount_target>:<fs_driver>:<options>"
```
RaduNichita:
```suggestion
 *		"<source_device>:<mount_target>:<fs_driver>:<options>",
```
I think there should be a comma here.

lib/vfscore/volumes.c
```diff
@@ -0,0 +1,59 @@
...
+		/* For now, mount() flags are set to 0 */
+		rc = mount(sdev, path, drv, 0, opts);
+		if (unlikely(rc))
+			uk_pr_err("Failed to mount %s: error %d\n", sdev, rc);
+	}
+
+	return 0;
+}
+
+/* Should be preferrably called after vfscore's rootfs is mounted (prio 4) */
```
RaduNichita: Could you elaborate a little bit about priorities. Why it should be 4 here?

RaduNichita (changes requested):
Great work, @mogasergiu :muscle: . I left some minor nitpicks.
I will approve after that!

lib/vfscore/volumes.c
mogasergiu: @skuenzer AFAIK, the `uklibparam` design makes use of whitespaces in such cases :thinking:.

lib/vfscore/volumes.c
mogasergiu: What I meant was that `rootfs.c` has a `uk_rootfs_initcall_prio(vfscore_rootfs, 4);`. Therefore, in order to ensure this can only be called after an explicit rootfs, give this a priority of `4 + 1 = 5`. I was referring to the `rootfs`. I will make the comment more clear :100:.

lib/vfscore/volumes.c
skuenzer: Yes, no commas. This has actually a pragmatic reason because that missing comma saves us parsing code. We can just rely on `ukargparse` for doing the argument split and array parsing. `ukargparse` also handles quotes and escaped quotes correctly. `uklibparam` automatically benefits from this.

lib/vfscore/volumes.c
skuenzer: I think if the `uk_rootfs_initcall_prio()` calls are next to each other (I am speaking about merged `rootfs.c` and `volumes.c`), then this is also clear. Additionally, `<uk/prio.h>` provides some computation macros so that this could be by using those also be a bit more expressive by the code. And the last alternative I see is to use only a single automount function that handles first `rootfs` and then `volumes`...

StefanJum (commented):
I retested this and everything works fine, thanks!

lib/vfscore/automount.c
```diff
@@ -123,8 +155,33 @@ static int vfscore_rootfs(void)
 		uk_pr_crit("Failed to mount %s (%s) at /: %d\n", rootdev, rootfs, errno);
 		return -1;
 	}
+#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS */
+
+	return 0;
+}
+
+static int vfscore_volumes_automount(void)
```
skuenzer: This is more a minor detail, more aesthetically: I'd prefer two functions at the same level:
`vfscore_automount_rootfs`
`vfscore_automount_volumes`
and a third function that calls these two and is registered to the init table.
btw, it is a clever idea to create empty function bodies depending on the configuration.

RaduNichita (approved):
All good now. Thanks, @mogasergiu :1st_place_medal: 

lib/vfscore/Config.uk
```diff
@@ -100,5 +100,19 @@ if LIBVFSCORE_AUTOMOUNT_ROOTFS
 		filesystem.
 endif
 
+config LIBVFSCORE_MAX_VOLUMES_COUNT
```
skuenzer: Can we also introduce an option that switches this on or off? I am thinking of something along these lines:
```
config LIBVFSCORE_AUTOMOUNT_VOLUMES
bool "Automatically mount list of volumes from kernel arguments"
selects LIBUKLIBPARAM
default n

if LIBVFSCORE_AUTOMOUNT_VOLUMES
    config LIBVFSCORE_AUTOMOUNT_VOLUMES_COUNT
    # ...
endif
```

StefanJum (commented):
Sill no bugs found in the wild :sunglasses: 
The only thing I would do is move the `help` section from under the `LIBVFSCORE_MAX_VOLUMES_COUNT` to the `LIBVFSCORE_AUTOMOUNT_VOLUMES` option. It's not really related to the max number of volumes.

razvanvirtan (approved):
All good for me, thanks!

mogasergiu:
> Sill no bugs found in the wild sunglasses The only thing I would do is move the `help` section from under the `LIBVFSCORE_MAX_VOLUMES_COUNT` to the `LIBVFSCORE_AUTOMOUNT_VOLUMES` option. It's not really related to the max number of volumes.

Good point, I simply copy-pasted. :rofl:

StefanJum (approved):
All good now

lib/vfscore/automount.c
```diff
@@ -0,0 +1,309 @@
...
+		return -1;
+	}
+
+	return 0;
+}
+
+#else /* CONFIG_LIBUKCPIO */
+static int vfscore_mount_initrd_volume(struct vfscore_volume *vv __unused)
+{
+	uk_pr_warn("Cannot extract InitRD. Enable CONFIG_LIBRAMFS.\n");
```
mogasergiu: Oops should have been `CONFIG_LIBUKCPIO`

skuenzer: Probably we should just fall back to a default message, something like saying: `Unknown filesystem: initrd`. I guess we have something like this already built-in with `mount()`. So, I think a separate handler for "initrd-support-is-not-present" is not needed.

lib/vfscore/automount.c
```diff
@@ -0,0 +1,309 @@
...
+#ifdef CONFIG_LIBVFSCORE_FSTAB
+static int vfscore_automount_fstab_volumes(void)
+{
+	struct vfscore_volume vv;
+	int rc, i;
+
+	for (i = 0; i < CONFIG_LIBVFSCORE_FSTAB_SIZE && vfscore_fstab[i]; i++) {
+		vfscore_fstab_fetch_volume_args(vfscore_fstab[i], &vv);
+
+		uk_pr_info("Mount %s to %s...\n", vv.sdev, vv.path);
```
skuenzer: What is `vv.sdev` saying in the initrd case? Maybe you could place this message under the else case and print a message for initrd like: "Extracting initrd to %s...\n"

mogasergiu: `vv.sdev` is a mandatory argument as per `fstab` syntax:

```
vfs.fstab=[
		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
		...
	]
```
`vv.sdev` is therefore `<src_dev>`.

EDIT: The initrd specific printing is already done by `vfscore_mount_initrd_volume`

lib/vfscore/automount.c
mogasergiu: Right then, I will just let this be an empty stub.

lib/vfscore/automount.c
```diff
@@ -0,0 +1,312 @@
...
+
+	/* TODO: Support multiple Initial RAM Disks */
+	rc = ukplat_memregion_find_initrd0(&initrd);
+	if (unlikely(rc < 0)) {
+		uk_pr_crit("Could not find an initrd!\n");
+
+		return -1;
+	}
+
+	uk_pr_info("Mount ramfs to /...\n");
```
skuenzer: I think this message is not needed anymore and it is incorrect. initrd can go now to any submountpoint, right?

lib/vfscore/automount.c
```diff
@@ -0,0 +1,312 @@
...
+	struct ukplat_memregion_desc *initrd;
+	int rc;
+
+	UK_ASSERT(vv);
+
+	/* TODO: Support multiple Initial RAM Disks */
+	rc = ukplat_memregion_find_initrd0(&initrd);
+	if (unlikely(rc < 0)) {
+		uk_pr_crit("Could not find an initrd!\n");
+
```
skuenzer, line 166: I think these spaces between message and return code is not necessary for these small if blocks. I leave it up to you if you want to keep them or not.

lib/vfscore/automount.c
skuenzer: My question was that the initrd mounting code does not need a source device as parameter. According to the code we accept anything that is in there which is actually fine. :+1:

lib/vfscore/automount.c
mogasergiu: Now that you mention it, [mount](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/mount.c#L122) already has a similar message. Therefore the message should be dropped from `vfscore_automount_fstab_volumes` as well :+1:.

skuenzer:
I just noticed, that in your second commit that removes the `vfs.roofs` parameter, etc. you also need to update `Config.uk`:
```diff
 config LIBVFSCORE_AUTOMOUNT_ROOTFS
 bool "Automatically mount a root filesysytem (/)"
 default n
-imply LIBUKLIBPARAM
 help
-	Automatically mounts '/' during boot. If `libuklibparam` is
-	compiled in, the default root filesystem and mount options can
-	be changed with the following library parameters:
-	'vfs.rootfs', 'vfs.rootdev', 'vfs.rootflags', and 'vfs.rootopts'
+	Automatically mounts '/' during boot. The parameters are
+	compiled-in and cannot be changed at runtime.
```

This is a minor comment: Instead of keeping the parameters set from config options from line 76 to 98, you could directly fill the struct in line 226 with those.

I aslo jsut notice that you are using `#ifdef CONFIG_ABC` although you can use `#if CONFIG_ABC` which is slightly more powerful because it catches the case when you manually do `#define CONFIG_ABC 0`. If you use `#if` on an undefined pre-processor macro, it will return false.

lib/vfscore/automount.c
```diff
@@ -0,0 +1,307 @@
...
+	for (i = 0; i < CONFIG_LIBVFSCORE_FSTAB_SIZE && vfscore_fstab[i]; i++) {
+		vfscore_fstab_fetch_volume_args(vfscore_fstab[i], &vv);
+
+		if (!strncmp(vv.drv, "initrd", sizeof("initrd") - 1))
+			rc = vfscore_mount_initrd_volume(&vv);
+		else
+			rc = vfscore_mount_volume(&vv);
+		if (unlikely(rc))
+			uk_pr_err("Failed to mount %s: error %d\n", vv.sdev,
+				  rc);
```
skuenzer, line 267: We should propagate failure as soon as one mount operation failed

mogasergiu:
> I aslo jsut notice that you are using `#ifdef CONFIG_ABC` although you can use `#if CONFIG_ABC` which is slightly more powerful because it catches the case when you manually do `#define CONFIG_ABC 0`. If you use `#if` on an undefined pre-processor macro, it will return false.

Is this just general advice? The problem is if you do `#if` on a string you are going to be met with a preprocessor error. So I think I will stick to `#ifdef`'s in the `CONFIG_LIBVFSCORE_ROOT*` macro's :thinking:.

skuenzer (approved):
Thanks a lot for the work!

PR #980, opened 2023-07-13
plat/kvm/arm: Do not enable interrupts before `wfi`

kubanrob:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

None

### Description of changes

There is the opportunity to miss an IRQ on the path to `wfi` from the idle thread after deciding to sleep:
The interrupts are enable just before exuting `wfi`. If the interrupt already arrived before that, it will be executed before `wfi`, and `wfi` will go to sleep afterward, not being woken up by the interrupt.

This patch removes calls to enable/disable interrupts around `wfi`. `wfi` will be woken up by interrupts regardless of the state of the CPSR register.

### Feedback needed 

- It is probably a good idea to assert IRQ disabled in `ukplat_lcpu_halt_irq` similar to #941, however, it is not completely clear if this will work with the other platforms.

### DIFF
--- a/include/uk/plat/lcpu.h
+++ b/include/uk/plat/lcpu.h
@@ -96,6 +96,9 @@ void ukplat_lcpu_halt_irq_until(__nsec until);
 /**
  * Halts the current logical CPU. Execution is resumed when an interrupt/signal
  * arrives.
+ *
+ * NOTE: This must be called with IRQ's disabled. On return, IRQ's are not
+ *        re-enabled.
  */
 void ukplat_lcpu_halt_irq(void);
 
--- a/plat/kvm/arm/lcpu.c
+++ b/plat/kvm/arm/lcpu.c
@@ -46,6 +46,8 @@ void ukplat_lcpu_disable_irq(void)
 
 void ukplat_lcpu_halt_irq(void)
 {
+	UK_ASSERT(ukplat_lcpu_irqs_disabled());
+
 	ukplat_lcpu_enable_irq();
 	halt();
 	ukplat_lcpu_disable_irq();
--- a/plat/kvm/arm/lcpu.c
+++ b/plat/kvm/arm/lcpu.c
@@ -48,9 +48,12 @@ void ukplat_lcpu_halt_irq(void)
 {
 	UK_ASSERT(ukplat_lcpu_irqs_disabled());
 
-	ukplat_lcpu_enable_irq();
+	/* Note: If priority masking is enabled
+	 * interrupts need to be unmasked in the GIC.
+	 *
+	 * See Linux `cpu_do_idle(void)` implementation
+	 */
 	halt();
-	ukplat_lcpu_disable_irq();
 }
 
 unsigned long ukplat_lcpu_save_irqf(void)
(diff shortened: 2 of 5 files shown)


razvand (approved):
Thanks, @kubanrob, this works for me.

mogasergiu (changes requested):
This is going to look inconsistent between arm and x86, but I guess this turns out to be good since `wfi`, as you said, is entering a state where you can wake up regardless of IRQ's being masked or not... unlike x86's `hlt` which would expect to be preceeded by an instruction that implies interrupt shadowing to achieve the same behavior... so, I guess, if anything, it would reflect reality as far as the difference between the two goes.

Besides this turns out to be a good quick fix for an issue @razvand has just encountered, so this is a good addition.

One thing though, I think we could benefit from adding a `UK_ASSERT(ukplat_lcpu_irqs_disabled());`, as it becomes more explicit what the expected behavior of this function is  (halt and wait for the next IRQ), because although it ends up being a wrapper for `halt()`, it should be clear the usecase is not the same. Otherwise, it implies that the occurrence of an IRQ before `halt()` is fine.

kubanrob:
> One thing though, I think we could benefit from adding a `UK_ASSERT(ukplat_lcpu_irqs_disabled());`

I added a commit to add an assert similar to #941 in the implementations. While doing that, I noticed both linuxu and xen both enable the interrupts before the halt, which will most likely lead to the  same problem as on arm.

I added two additional commits changing these platform, but I can't invest to much time into testing/researching these.

Please let me know if I should:
1. just drop the xen/linuxu commits from this PR
1. open one or two additional PRs for these two changes
1. let somebody familiar with xen/linuxu review it here

mogasergiu:
@kubanrob The first commit is a good addition (btw `s/ukplat_lcpu_irq/ukplat_lcpu_halt_irq/` for the commit message :wink:).

I would say:
- keep the first two commits in this PR, we need it in anyway as it fixes an existing bug
- open a **DRAFT** PR with an `[RFC]` prefix (something among the lines of `[RFC] plat/{linuxu, xen}: Do not enable interrupts before halting`) that contains the other two commits related to Xen and linuxu

I am not sure the same thing applies to those platforms either, so let's see what others from the community can say about those changes. Thank you for the thoughtful additional changes :100:.

mogasergiu (approved):
Thank you for the quick responses :100:.

nderjung:
This PR broke the build on staging: https://github.com/unikraft/unikraft/actions/runs/5742321655

cc @kubanrob @razvand @mogasergiu 

```
 E  /usr/bin/ld: /github/workspace/_helloworld/.unikraft/build/libkvmplat.o: in function `ukplat_lcpu_halt_irq':
 E  /github/workspace/plat/kvm/x86/lcpu.c:50: undefined reference to `UK_ASSERT'
 E  collect2: error: ld returned 1 exit status
 ```

craciunoiuc:
The first victim of the upcoming CI/CD system 🥰

PR #985, opened 2023-07-23
plat/drivers: Add two configurations to ns16550

zhxj9823:
### Prerequisite checklist

### Base target

 - Architecture(s): `ARM64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

The register shift of ns16550 is hard-coded in the driver, but it does not match the hardware on some platforms. The device-tree specification has a property "reg-shift" to describe the register shift of a device.
To match the specification, the register shift of ns16550 is retrieved from the device tree instead. If no such property is found, the default value 0 is used.

The register width of ns16550 varies on different platforms. u-boot uses a reg-io-width property in the device tree to determine it. If no such property is found, the default value 1 is used. The read and write macros are replaced by functions with the same name.

These two configurations break the compatibility of the driver. To keep the compatibility, the device tree should contain these properties.

### DIFF
--- a/plat/drivers/uart/ns16550.c
+++ b/plat/drivers/uart/ns16550.c
@@ -43,8 +43,6 @@
 #define NS16550_LSR_OFFSET	0x05U
 #define NS16550_MSR_OFFSET	0x06U
 
-#define NS16550_REG_SHIFT	0x02U
-
 #define NS16550_LCR_DLAB	0x80U
 #define NS16550_IIR_NO_INT	0x01U
 #define NS16550_FCR_FIFO_EN	0x01U
@@ -59,7 +57,7 @@
  * use ns16550_uart_initialized as an extra variable to check
  * whether the UART has been initialized.
  */
-#if defined(CONFIG_EARLY_UART_NS16550) && defined(CONFIG_EARLY_UART_NS16550_BASE)
+#if defined(CONFIG_EARLY_UART_NS16550)
 static uint8_t ns16550_uart_initialized = 1;
 static uint64_t ns16550_uart_base = CONFIG_EARLY_UART_NS16550_BASE;
 #else
@@ -67,11 +65,53 @@ static uint8_t ns16550_uart_initialized;
 static uint64_t ns16550_uart_base;
 #endif
 
+/* The register shift. Default is 0 (device-tree spec v0.4 Sect. 4.2.2) */
+static uint32_t ns16550_reg_shift = CONFIG_UART_NS16550_REG_SHIFT;
+
+/* The register width. Default is 1 (8-bit register width) */
+static uint32_t ns16550_reg_width = CONFIG_UART_NS16550_REG_WIDTH;
+
 /* Macros to access ns16550 registers with base address and reg shift */
-#define NS16550_REG(r)			((uint16_t *)(ns16550_uart_base + \
-					(r << NS16550_REG_SHIFT)))
-#define NS16550_REG_READ(r)		ioreg_read16(NS16550_REG(r))
-#define NS16550_REG_WRITE(r, v)	ioreg_write16(NS16550_REG(r), v)
+#define NS16550_REG(r) (ns16550_uart_base + (r << ns16550_reg_shift))
+
+static uint32_t ns16550_reg_read(uint32_t reg)
+{
+	uint32_t ret;
+
+	switch (ns16550_reg_width) {
+	case 1:
+		ret = ioreg_read8((uint8_t *)NS16550_REG(reg)) & 0xff;
+		break;
+	case 2:
+		ret = ioreg_read16((uint16_t *)NS16550_REG(reg)) & 0xffff;
+		break;
+	case 4:
+		ret = ioreg_read32((uint32_t *)NS16550_REG(reg));
+		break;
+	default:
+		UK_CRASH("Invalid register width: %d\n", ns16550_reg_width);
+	}
+	return ret;
+}
+
+static void ns16550_reg_write(uint32_t reg, uint32_t value)
+{
+	switch (ns16550_reg_width) {
+	case 1:
+		ioreg_write8((uint8_t *)NS16550_REG(reg),
+			     (uint8_t)(value & 0xff));
+		break;
+	case 2:
+		ioreg_write16((uint16_t *)NS16550_REG(reg),
+			      (uint16_t)(value & 0xffff));
+		break;
+	case 4:
+		ioreg_write32((uint32_t *)NS16550_REG(reg), value);
+		break;
+	default:
+		UK_CRASH("Invalid register width: %d\n", ns16550_reg_width);
+	}
+}
 
 static void init_ns16550(uint64_t base)
 {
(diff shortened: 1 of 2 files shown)


plat/drivers/uart/ns16550.c
```diff
@@ -65,13 +63,14 @@ static uint64_t ns16550_uart_base = CONFIG_EARLY_UART_NS16550_BASE;
...
 static uint64_t ns16550_uart_base;
+static uint32_t ns16550_reg_shift;
 #endif
 
 /* Macros to access ns16550 registers with base address and reg shift */
-#define NS16550_REG(r)			((uint16_t *)(ns16550_uart_base + \
-					(r << NS16550_REG_SHIFT)))
-#define NS16550_REG_READ(r)		ioreg_read16(NS16550_REG(r))
-#define NS16550_REG_WRITE(r, v)	ioreg_write16(NS16550_REG(r), v)
+#define NS16550_REG(r)			((uint8_t *)(ns16550_uart_base + \
```
michpappas: Here, besides reg shift, you also change the reg width. How is that change justified? Is it backwards compatible? Should it also be configurable? In any case this change should also be described in the commit message.

plat/drivers/uart/ns16550.c
```diff
@@ -65,13 +63,14 @@ static uint64_t ns16550_uart_base = CONFIG_EARLY_UART_NS16550_BASE;
 #else
 static uint8_t ns16550_uart_initialized;
 static uint64_t ns16550_uart_base;
+static uint32_t ns16550_reg_shift;
```
michpappas: Please add a comment that the default value of zero is specified by the device-tree spec along with the relevant section, eg
```
static uint32_t ns16550_reg_shift; /* device-tree spec v0.4 Sect. 4.2.2 */
```

michpappas (changes requested):
Hi @zhxj9823, thanks for that! 🚀 Besides the comments below, since this change breaks backwards compatibility, it is important to mention this in the commit message. Moreover, please include in the commit message justification for that, ie a reference to the spec.

plat/drivers/uart/ns16550.c
zhxj9823: Regarding the register width of ns16550, I find that [this document](https://www.ele.uva.es/~jesus/UltimatePutchar.pdf) mentions that 

> The UART registers are all 8-bit wide and only the 8 least significant bits of variables are actually read or written.

But I have not found a specification explicitly saying so. 

On the other side, u-boot uses a [`reg-io-width` property](https://elixir.bootlin.com/u-boot/latest/source/drivers/serial/ns16550.c#L556) to determine the register width, but this does not appear in the [device tree specification](https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.4). If this property is used in the real device trees, we can use this property to determine the register width as well.

michpappas: Yes I think it's a good idea to support this nevertheless, just in case some device deviates. That way we minimize the risk of breaking existing implementations that use 16-bit accesses (CC: @kubanrob). Worst case everyone uses the default.

razvand:
Hi, @zhxj9823. This has been marked as a draft PR. What needs to happen to make this a full PR?

razvand (commented):
@zhxj9823 , could you also elaborate what is the register shift? As someone with little knowledge about UART / NS16550, I would benefit from this information as part of the commit message. Likely others as well.

michpappas:
> @zhxj9823 , could you also elaborate what is the register shift? As someone with little knowledge about UART / NS16550, I would benefit from this information as part of the commit message. Likely others as well.

Personally I am against trivialization (in general), it is expected that the audience has some familiarity with the topic. Look at the commit history of other OSes for example. Besides, where would you draw the line. And this is a simple concept, more complex topics would require way more elaborate explanations. 

Saying that, if you really think it's necessary, one sentence is okay 😛

plat/drivers/uart/ns16550.c
zhxj9823: Please check the changes I made later. One thing to be noted is that [reg-io-width property](https://elixir.bootlin.com/u-boot/latest/source/drivers/serial/ns16550.c#L556) is only used in a case. So I am not sure if this change suits the use cases in the platforms Unikraft supprts.

michpappas:
@zhxj9823 thanks for the updates. As requested by @razvand please remove the `Draft` status from your PR. 

One small thing that could be improved could be explicitly setting the default reg_shift:
```
+static uint32_t ns16550_reg_shift = 0; /* device-tree spec v0.4 Sect. 4.2.2 */
```
This is of course not required for functionality, but makes it more clear when reading the code that the default value is derived by the spec.

Other than that, from my side it looks good 👍🏼

@kubanrob I am adding as a reviewer since that change affects you.

plat/drivers/uart/ns16550.c
```diff
@@ -65,26 +63,64 @@ static uint64_t ns16550_uart_base = CONFIG_EARLY_UART_NS16550_BASE;
 #else
 static uint8_t ns16550_uart_initialized;
 static uint64_t ns16550_uart_base;
+static uint32_t ns16550_reg_shift = 0; /* device-tree spec v0.4 Sect. 4.2.2 */
+static uint32_t ns16550_reg_width = 1; /* 8-bit register width by default */
```
kubanrob: These variables are missing for the early uart (see the lines above). It would make sense to introduce the corresponding config options (default values matching the device tree ones).

kubanrob (commented):
@michpappas I investigated for our plattform and do not see problems with switching to 8bit register-width in this driver. On the other hand, having the register-width configurable is will also not hurt anybody, and the reduces the risk for anybody currently using this driver.

Anyways any new variable should also be properly initialized if early boot UART is activated.

plat/drivers/uart/ns16550.c
michpappas: Good point, thanks! 👍🏼 @zhxj9823 will you please update to add these?

zhxj9823: I have added the config options. Please check if these options work fine.

It also raises the question of how to switch between `pl011` and `ns16550` as the early debug console.

michpappas: > I have added the config options. Please check if these options work fine.

I think these two options should not be limited to `EARLY_UART`, but should be overriding the defaults when set, unconditionally. That would allow the driver to work correctly in any future platforms that do not use dts.

The priority should therefore be:
1. dts (overrides everything)
2. Kconfig (if set override defaults)
3. Defaults (according to the spec)

> It also raises the question of how to switch between pl011 and ns16550 as the early debug console.

For that we should merge the options of the different drivers into a single one, and most likely also consider an option in the command line, but this is outside the scope of this PR.

zhxj9823: > I think these two options should not be limited to `EARLY_UART`, but should be overriding the defaults when set, unconditionally. That would allow the driver to work correctly in any future platforms that do not use dts.
> 
> The priority should therefore be:
> 
> 1. dts (overrides everything)
> 2. Kconfig (if set override defaults)
> 3. Defaults (according to the spec)

I update the patch. But I think it would be better if the two configs depend on something like `uart_ns16550`.

michpappas: I had a brief look on how the reg-shift and width for the `earlycon` are specified on linux. There these are not compiled in, but rather derived from the command-line options:
* https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html
* https://elixir.bootlin.com/linux/latest/source/drivers/tty/serial/earlycon.c#L93

The latter derives the regshift from the mmio width, ie 0 by default, 1 for MMIO16, and 2 for MMIO32 access. Right now we use 16-bit access and a regshift of 2 which doesn't make sense. @kubanrob should that be 32-bit access instead? Also would the command line be a valid option for earlycon in your case, assuming we implemented that? I imagine that since it's used on linux, it should be no problem for Unikraft.

**Edit:** Actually nevermind, as right now `libukparam` only supports paramters in the the form of `<libname>.<paramname>=`, so parameters without a scope like `earlycon=` or `console=` won't work.

michpappas:
@zhxj9823 please move these changes to the new location of `ns16550` under `drivers` (rebase to #1026).

michpappas:
> @zhxj9823 please move these changes to the new location of `ns16550` under `drivers` (rebase to #1026).

@zhxj9823 will you kindly rebase this on top of #1048 instead? It only needs a small change in `Config.uk` other than that it applies cleanly. Thanks a lot! 🚀

razvand (approved):
@zhxj9823, @michpappas, this is OK from my side. I've tested it together with #1048 and it works OK.

kubanrob (approved):
Looks good to me, thanks you @zhxj9823!

michpappas:
@zhxj9823 now that the parent PR has been merged, please rebase to `staging` to remove the dependency commit from this PR, so that we can merge it too. Thanks 🚀

michpappas (approved):
Thank you @zhxj9823 🚀 

PR #986, opened 2023-07-23
plat/drivers: Provide a single API for UART drivers

zhxj9823:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64` and `ARM64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This PR migrates the console API into a new `libukconsole`. Each console provides an ops structure that implements the console API. The APIs include `coutk`, `coutd`, `cink`, and `init`.

A unified `uk_console_init` function takes `bootinfo` as a parameter to initialize the available console in both x86 and ARM platforms. The implementation of `uk_console_init` is platform-specific.

The header for ANSI characters is moved into a separate one in `include/uk`.

GitHub-Depends-On: #989


razvand:
Hi, @zhxj9823, this is marked as draft work. What needs to happen to make this a full PR?

lib/ukconsole/console.c
```diff
@@ -27,12 +28,67 @@
...
+{
+	int offset;
+
+	/* find the device by name */
+	if (fdt_node_offset_by_compatible(dtb, -1, "arm,pl011") >= 0)
+		uk_console_ops = &uk_console_pl011_ops;
+
+	if ((offset = fdt_node_offset_by_compatible(dtb, -1, "ns16550")) >= 0
+	    && (offset = fdt_node_offset_by_compatible(dtb, -1, "ns16550a"))
+		   >= 0) {
```
razvand: See comment from Codacy. Where is offset used?

razvand (changes requested):
For the last two commits of this PR, please add a motivation for their existence. Such as why move the x86 console to a COM driver under `plat/drivers/uart/`.

Also, see some minor typos: `migrate` -> `migrates`, `an addition` -> `an additional`.

Also it would be ideal to add tests as part of the new `ukconsole` library, before upstreaming it, via the [`uktest` framework](https://unikraft.org/docs/develop/writing-tests/). @michpappas , what do you think? I'm OK with upstreaming it as part of the 0.14 release even if the tests aren't there.

zhxj9823:
>> Also it would be ideal to add tests as part of the new `ukconsole` library, before upstreaming it, via the [`uktest` framework](https://unikraft.org/docs/develop/writing-tests/). @michpappas , what do you think? I'm OK with upstreaming it as part of the 0.14 release even if the tests aren't there.

> The input and output of this library are from and to the serial console. Can this be fitted into the `uktest` framework?

Does not fit. In general devices are a bit more complex to write tests for, we would need to introduce some abstractions so it would take time. I propose that we skip it for now. @razvand.

lib/ukconsole/Config.uk
```diff
@@ -0,0 +1,68 @@
+menuconfig LIBUKCONSOLE
+	bool "ukconsole: Abstraction for serial console"
+	default y
+
+if LIBUKCONSOLE
+
+config KVM_KERNEL_SERIAL_CONSOLE
```
michpappas: KVM configuration should be in `plat/kvm/Config.uk`. The location of driver-specific configuraiton is a bit tricky right now because we don't have not moved drivers out of `plat` yet, so there is a gap on where these should go. Maybe you can move it back to `plat/kvm` for now, and I'll try to come back with an update.

lib/ukconsole/console.c
```diff
@@ -27,12 +28,67 @@
...
+{
+	return uk_console_ops->coutd(buf, len);
+}
+
+int ukplat_cink(char *buf, unsigned int maxlen)
+{
+	return uk_console_ops->cink(buf, maxlen);
+}
+
+#if CONFIG_ARCH_X86_64
```
michpappas: If you need per-architecture variants you need to place architecture-specific variants under `lib/ukconsole/arch/<arch>/<name>.c` and update your `Makefile.uk` accordingly.

lib/ukconsole/console.c
```diff
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 /*
- * Copyright (c) 2021 OpenSynergy GmbH
```
michpappas: Here it looks like derived work from an existing file, which is not the case. Please introduce a new file. The copyright for new files should be:
```
 */
```

lib/ukconsole/console.c
```diff
@@ -27,12 +28,67 @@
...
+	/* Initialize the corresponding console */
+	uk_console_ops->init();
+
+	uk_pr_info("Console init finished\n");
+}
+#else  /* !CONFIG_ARCH_X86_64 */
+extern const struct uk_console_ops uk_console_pl011_ops;
+extern const struct uk_console_ops uk_console_ns16550_ops;
+
+void uk_console_init(const void *dtb)
```
michpappas: To avoid having different function prototypes per architecture, and to support in the future other discovery methods like ACPI, you can instead pass all instances a parameter to the bootinfo. In the case of the dtb you should get the address from there.

As discussed, that depends on #912, so please rebase and mark the dependency in the cover letter of your PR using `GitHub-Depends-On`. For reference you can see https://github.com/unikraft/unikraft/pull/988 and https://github.com/unikraft/unikraft/pull/989.

lib/ukconsole/include/uk/console.h
```diff
@@ -100,6 +100,17 @@ extern "C" {
 #define UK_ANSI_COLOR_CYAN		6
 #define UK_ANSI_COLOR_WHITE		7
 
+/* console operations */
+struct uk_console_ops {
+	int (*coutk)(const char *buf, unsigned int len);
+	int (*coutd)(const char *buf, unsigned int len);
+	int (*cink)(char *buf, unsigned int maxlen);
+#if CONFIG_ARCH_X86_64
```
michpappas: To avoid having different function prototypes per architecture, and to support in the future other discovery methods like ACPI, you can instead pass all instances a parameter to the bootinfo. In the case of the dtb you should get the address from there.

As discussed, that depends on #912, so please rebase and mark the dependency in the cover letter of your PR using `GitHub-Depends-On`. For reference you can see https://github.com/unikraft/unikraft/pull/988 and https://github.com/unikraft/unikraft/pull/989.

lib/ukconsole/console.c
```diff
@@ -27,12 +28,67 @@
...
+#if CONFIG_ARCH_ARM64
+#include <libfdt.h>
 #endif
+#include <uk/console.h>
+#include <uk/print.h>
+#include <uk/assert.h>
+
+static struct uk_console_ops *uk_console_ops;
+
+int ukplat_coutk(const char *buf, unsigned int len)
```
michpappas: These should no longer be `ukplat_`  but `uk_console`

plat/kvm/Makefile.uk
```diff
@@ -82,14 +82,15 @@ ifeq ($(findstring y,$(CONFIG_KVM_KERNEL_VGA_CONSOLE) $(CONFIG_KVM_DEBUG_VGA_CON
...
-LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBKVMPLAT_BASE)/x86/serial_console.c
+LIBKVMPLAT_SRCS-$(CONFIG_ARCH_X86_64) += $(UK_PLAT_DRIVERS_BASE)/uart/com.c
 endif
 
 ##
 ## Architecture library definitions for arm64
 ##
 ifeq ($(findstring y,$(CONFIG_KVM_KERNEL_SERIAL_CONSOLE) $(CONFIG_KVM_DEBUG_SERIAL_CONSOLE)),y)
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_DRIVERS_BASE)/uart/pl011.c
+LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_DRIVERS_BASE)/uart/ns16550.c
```
michpappas, line 94: Looks like this change does not belong to this commit.

plat/drivers/include/uart/com.h
```diff
@@ -19,11 +19,11 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#ifndef __KVM_SERIAL_CONSOLE__
-#define __KVM_SERIAL_CONSOLE__
+#ifndef __COM_H__
+#define __COM_H__
 
 void _libkvmplat_init_serial_console(void);
```
michpappas, line 25: COMx devices are not KVM specific: https://en.wikipedia.org/wiki/COM_(hardware_interface), so the interface provided by this driver should come with appropriate names. Also shouldn't this driver also provide a set of ops like the rest?

plat/drivers/uart/ns16550.c
```diff
@@ -173,14 +168,14 @@ static int ns16550_getc(void)
 	return (int)(NS16550_REG_READ(NS16550_RBR_OFFSET) & 0xff);
 }
 
-int ukplat_coutk(const char *buf, unsigned int len)
+int ns16550_coutk(const char *buf, unsigned int len)
```
michpappas: These should be `static` now. In the rest of the drivers too.

michpappas (commented):
Hi @zhxj9823 thanks for this work 🚀 I did a first pass, it's likely that more iterations will follow. 

Besides the inline comments, according to Unikraft's guidelines changes to each library must come on a separate commit. This applies to any updates to `nolibc`, `ukdebug`, and `ukboot`. Moreover, please introduce the `ukconsole` library on a separate commit too. Similarly for changes on each driver.

plat/drivers/include/uart/com.h
zhxj9823: This relates to the issue that `vga` console and `com` console are used concurrently in `unikraft/plat/kvm/x86/console.c`. In the current scheme, there can be an `ops` structure for this file. However, having a separate `ops` structure for `com` driver breaks the backward compatibility. Another choice is to enable multiple consoles instead of only using one.

lib/ukconsole/console.c
zhxj9823: In this case, would it make sense to divide the `.h` and `.c` files? For the `console.h` file, the ANSI characters and console APIs are used in different cases. For the `console.c` file, it can be divided into the arch-specific `console_init` function and the other three arch-independent functions.

michpappas:
I had a brief look on how the reg-shift and width for the `earlycon` are specified on linux. There these are not compiled in, but rather derived from the command-line options:
* https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html
* https://elixir.bootlin.com/linux/latest/source/drivers/tty/serial/earlycon.c#L93

The latter derives the regshift from the mmio width, ie 0 by default, 1 for MMIO16, and 2 for MMIO32 access. Right now we use 16-bit access and a regshift of 2 which doesn't make sense. @kubanrob should that be 32-bit access instead? Also would the command line be a valid option for earlycon in your case, assuming we implemented that? I imagine that since it's used on linux, it should be no problem for Unikraft.

michpappas:
Hi @zhxj9823. I opened a PR with some changes to introduce a drivers subsystem #1023, and migrated the `pl011` and `ns16550` drivers there (#1025 and #1026). You can update this PR to move the COM and VGA console drivers to `drivers/uktty` using `pl011` as a template. Moveover, please rename this library from `ukconsole` to `uktty`.

You can also consider refactoring the current PR to register the `pl011` and `ns16550` drivers as platform drivers using the `PF_REGISTER_DRIVER` macro of the the platform bus (see `plat/common/include/platform_bus.h`). That will allow you to remove the dts probing from this library and replace it with a `uktty_init()` that the drivers will call on `add_device` to pass their ops to the `uktty` library. Notice that this deprecates the architecture-specific parts from this library.

michpappas:
Obsoleted by #1464

PR #988, opened 2023-07-24
plat/kvm/arm64: Implement lxboot on arm64

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Additional configuration

Enables `CONFIG_BOOT_PROTO_LXBOOT` for `arm64`.

### Description of changes

This series implements the arm64 linux boot protocol as described in [1].  The protocol requires that image is prepended with the image with a modified MZ header (shown below) and passes the base address of the fdt to x0.

```
      u32 code0;                    /* Executable code */
      u32 code1;                    /* Executable code */
      u64 text_offset;              /* Image load offset, little endian */
      u64 image_size;               /* Effective Image size, little endian */
      u64 flags;                    /* kernel flags, little endian */
      u64 res2      = 0;            /* reserved */
      u64 res3      = 0;            /* reserved */
      u64 res4      = 0;            /* reserved */
      u32 magic     = 0x644d5241;   /* Magic number, little endian,"ARM\x64" */
      u32 res5;                     /* reserved (used for PE COFF offset) */
```

Moreover this PR adds initrd support on arm64 via dts. Although not part of the linux boot protocol, arm64 linux systems provide this region via the "linux,initrd-start" and "linux-initrd,end" properties in the `chosen` node, as:
```    
      chosen {
            linux,initrd-start = <0x0 0x80000000>;
            linux,initrd-end = <0x0 0x80002000>;
      }
```

[1] https://www.kernel.org/doc/Documentation/arm64/booting.txt

GitHub-Depends-On: #912


michpappas:
Marking as RFC to discuss the generation of the header, as I think this should be discussed in conjunction with `mkefi.py` in #909.

Right now this PR provides two alternatives:

1. asm / linker (enabled by default)
2. Python script (`mklinux.py`)

Some topics to consider:

The script requires the following information:

* A raw binary image to prepend the header with.
* Three addresses to calculate the various header fields from (`ram_base`, `image_base`, `entry_point`).

The image binary needs to be generated by `Linker.uk`. The addresses can be obtained in two ways:
 
1. Pass the ELF along with raw binary as a parameter to `mklinux.py` (ugly IMO), and get the addresses from the ELF image from functions defined in script, similarly to `mkefi.py` in #909. To avoid code duplication this requires moving that code out of `mkefi.py` into a common library. I don't like that either, as this starts to build up.
2. Use `plat/common/Makefile.rules` to get the addresses of the required symbols, and pass them as parameters to the script. This is how the script is implemented atm - the calling part is missing.

The asm/linker approach seems more complex and error-prone as it involves things like section flags and objcopy that sometimes behave in an unexpected way (although once it works it works). On the other hand it seems like the right way to do the job; defining the encoding of the branch instruction in `mklinux.py` or calculating offsets manually are IMO an indication of doing something wrong. The asm/linker approach can also potentially support relocation.

Another aspect to consider with the script is a previous comment by @mschlumpp that these should be migrating to C code (can't find a reference rn sorry). Perhaps that could also be a silver line between the two?

Last but not least, please keep in mind that this header may in the future be required to replace the default MZ header used by EFI on `arm64`, either conditionally or unconditionally. With asm/linker it should be a matter of a `#ifdef` (assuming a consistent way to generate headers with EFI). With the script we could: 

1. Merge `mkefi.py` with `mklinux.py`. This has the drawback of increased complexity of `mkefi.sh` as it would also need ot handle prepending the header into non-EFI images on `arm64`.
2. Add a parameter to `mklinux.py` to replace the header on an existing image and handle the rest in `Linker.uk`. 

@marcrittinghaus  @mogasergiu @mschlumpp

zhxj9823:
> Right now this PR provides two alternatives:
> 
> 1. asm / linker (enabled by default)
> 2. Python script (`mklinux.py`)

@michpappas I'd like to know the workflow for these two alternatives. Is `mklinux.py` called immediately after the `make` process? In this case, would it be better to integrate this into the `make` flow with proper parameters? Also, what part is missing for [linuxhdr.S](https://github.com/unikraft/unikraft/pull/988/commits/1c5c99b38a2356f017d8a30b543ba79d416a2f07#diff-a86ee19d40019caf5ce3d9cf1ec31da3badf78d88ff247a6f9e428284fcba421)?

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
...
+    #
+    # We arbitrarily set this to header size + image size aligned up to 2MiB
+    total_size = os.path.getsize(opt.bin) + LINUX_ARM64_HDR_SIZE
+    LINUX_ARM64_HDR["IMAGE_SIZE"][0] = align_up(total_size, 2 * 1024 * 1024)
+
+    # Create img
+    with open(opt.bin, 'r+b') as f:
+        img = f.read()
+        f.seek(0)
+        for field in [k for k in LINUX_ARM64_HDR.keys() if k != "SectionHeaders"]:
```
zhxj9823: What is `"SectionHeaders"` in this context?

plat/kvm/include/kvm-arm64/image.h
```diff
@@ -0,0 +1,36 @@
...
+#define LINUX_ARM64_HDR_FLAGS_4K	(1 << 1)
+#define LINUX_ARM64_HDR_FLAGS_16K	(2 << 1)
+#define LINUX_ARM64_HDR_FLAGS_64K	(3 << 1)
+#define LINUX_ARM64_HDR_FLAGS					\
+	(LINUX_ARM64_HDR_FLAGS_LE | LINUX_ARM64_HDR_FLAGS_4K)
+
+#define LINUX_ARM64_HDR_SYMBOLS					\
+	_uk_image_size = _end - _base_addr;			\
+	_uk_load_offs  = _base_addr - _start_ram_addr - 64;	\
+	_linux_arm64_hdr_base = _start_ram_addr + DTB_RESERVED_SIZE - 64;
```
zhxj9823: Is this `64` corresponds to the `LINUX_ARM64_HDR_SIZE` macro?

plat/kvm/include/kvm-arm64/image.h
```diff
@@ -0,0 +1,36 @@
...
+#define LINUX_ARM64_HDR_MAGIC		"ARM\x64"
+#define LINUX_ARM64_HDR_IMAGE_SIZE	_uk_image_size
+#define LINUX_ARM64_HDR_LOAD_OFFS	_uk_load_offs
+#define LINUX_ARM64_HDR_FLAGS_BE	0
+#define LINUX_ARM64_HDR_FLAGS_LE	1
+#define LINUX_ARM64_HDR_FLAGS_4K	(1 << 1)
+#define LINUX_ARM64_HDR_FLAGS_16K	(2 << 1)
+#define LINUX_ARM64_HDR_FLAGS_64K	(3 << 1)
+#define LINUX_ARM64_HDR_FLAGS					\
+	(LINUX_ARM64_HDR_FLAGS_LE | LINUX_ARM64_HDR_FLAGS_4K)
```
zhxj9823: Should there be options to use `LINUX_ARM64_HDR_FLAGS_16K` and `LINUX_ARM64_HDR_FLAGS_64K`?

plat/kvm/arm/linuxhdr.S
```diff
@@ -0,0 +1,18 @@
...
+	add	x13, x18, #0x16			/* code0 */
+	b	qemu_virt_entry			/* code1 */
+	.quad	LINUX_ARM64_HDR_LOAD_OFFS	/* text_offset */
+	.quad	LINUX_ARM64_HDR_IMAGE_SIZE	/* image_size */
+	.quad	LINUX_ARM64_HDR_FLAGS		/* kernel flags */
+	.quad	0				/* res2 */
+	.quad	0				/* res3 */
+	.quad	0				/* res4 */
+	.ascii	LINUX_ARM64_HDR_MAGIC		/* magic */
+	.long	0x40				/* offset to PE header */
```
zhxj9823: How can this offset be guaranteed if such a header is not available yet?

jschlumpp:
About migrating away from Python build scripts: It primarily came from a conversation with Marc but the wish to remove the python dependency also surfaced from time to time elsewhere. Python is a quite large dependency and on some platforms/distros it's an explicit additional large package you have to install on top of the usual toolchain.

Apart from the dependency issue, using C would also have the benefit of being able to use a proper ELF parser like the one in elfutils. So far the python scripts rely on parsing the output of `objdump` or `readelf`, which sometimes resulted in reliability issues. Using a proper python ELF library is also not really an option because then you would need to somehow install that dependency which is even more messy in the python world.

Regarding the post-process script vs linker shenanigans: I think a well written post-process script is easier to maintain and more robust than linker scripts. Linker scripts are very under-documented and sometimes fail in unexpected ways from my experience.

mogasergiu:
> Regarding the post-process script vs linker shenanigans: I think a well written post-process script is easier to maintain and more robust than linker scripts. Linker scripts are very under-documented and sometimes fail in unexpected ways from my experience.

This :point_up:. I also personally find postprocessing scripts significantly easier to debug when something image format related goes wrong and it feels like you have more control over them.

 > 2. Add a parameter to mklinux.py to replace the header on an existing image and handle the rest in Linker.uk.

I would go with this and have a new track open for converting postprocessing scripts to C. Since these python scripts are not here to stay long term, perhaps there is not much point in discussing their design and should just go with what's simple and straightforward, deferring this as a different task that would run in parallel so as this does not become an unnecessary blocker for tasks that currently depend on this feature to proceed forward.

plat/common/bootinfo.c
```diff
@@ -252,6 +296,8 @@ void ukplat_bootinfo_fdt_setup(void *fdtp)
 
 	fdt_bootinfo_fdt_mrd(bi, fdtp);
 	fdt_bootinfo_mem_mrd(bi, fdtp);
+
```
mogasergiu: Why the new line?

plat/common/Makefile.rules
```diff
@@ -10,10 +10,14 @@ endif
 
 define build_bootinfo =
 	$(call build_cmd,UKBI,,$(1).bootinfo,\
-		$(SCRIPTS_DIR)/mkbootinfo.py $(1) $(1).bootinfo $(BINFO_FLAGS) && \
+		$(SCRIPTS_DIR)/mkbootinfo.py --min-address $(2) $(1) $(1).bootinfo $(BINFO_FLAGS) && \
```
mogasergiu: You repeated yourself in the commit message of `plat/common: Exclude segments below _base_addr from bootinfo`: `exclude segments exclude segments`

plat/common/bootinfo.c
```diff
@@ -222,6 +222,50 @@ static void fdt_bootinfo_cmdl_mrd(struct ukplat_bootinfo *bi, void *fdtp)
...
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return;
+
+	mrd.vbase = initrd_addr(fdt_initrd_start[0], start_len);
+	mrd.pbase = initrd_addr(fdt_initrd_start[0], start_len);
+	mrd.len   = initrd_addr(fdt_initrd_end[0], end_len) -
+	initrd_addr(fdt_initrd_start[0], start_len);
```
mogasergiu: This does not look aligned to the `=` operator.

mogasergiu (changes requested):
Probably lame mentioning these tiny nitpicks on a DRAFT PR, but maybe worth throwing them out there just for quick notekeeping. :)

support/scripts/mklinux.py
michpappas: Was not supposed to be there, now fixed thanks 👍🏼

plat/kvm/include/kvm-arm64/image.h
michpappas: This change is now removed in favor of `mklinux.py`.

plat/kvm/include/kvm-arm64/image.h
michpappas: This change is now removed in favor of `mklinux.py`.

plat/kvm/arm/linuxhdr.S
michpappas: This change is now removed in favor of `mklinux.py`. Saying that, I use the same value there. The rationale is that there is no right value there when there is no PE header (i.e. a zero offset also points to a non-existent header), so by convention the expectation is that if this header is used on an EFI image (not supported right now), the PE header will immediately follow the MZ header.

plat/common/bootinfo.c
michpappas: Removed

plat/common/Makefile.rules
michpappas: This commit is now removed in favor of `mklinux.py`.

plat/common/bootinfo.c
michpappas: Thanks, fixed 👍🏼

michpappas:
Updated to address review comments and to use the python script to generate the arm64 linux header. Removing RFC status, this is now ready review. Will rebase as soon as dependencies are merged.

plat/common/bootinfo.c
```diff
@@ -222,6 +222,50 @@ static void fdt_bootinfo_cmdl_mrd(struct ukplat_bootinfo *bi, void *fdtp)
...
+				       &start_len);
+	if (unlikely(!fdt_initrd_start || start_len <= 0))
+		return;
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return;
+
+	mrd.vbase = initrd_addr(fdt_initrd_start[0], start_len);
```
mogasergiu: Nitpick. I know it looks good to have the `=` aligned in a column, but it is inconsistent with other parts of the codebase. Unless there is a mention of this in the coding convention, I think it would be better to just stick to the `whitespace before/after operator/operand`.

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
...
+LINUX_ARM64_HDR = {
+    "CODE0"                   : [0x0000000091005a4d, 0x04],  # Offset   0
+    "CODE1"                   : [0x0000000000000000, 0x04],  # Offset   4
+    "LOAD_OFFS"               : [0x0000000000000000, 0x08],  # Offset   8
+    "IMAGE_SIZE"              : [0x0000000000000000, 0x08],  # Offset  16
+    "KERNEL_FLAGS"            : [0x0000000000000000, 0x08],  # Offset  24
+    "RES2"                    : [0x0000000000000000, 0x08],  # Offset  32
+    "RES3"                    : [0x0000000000000000, 0x08],  # Offset  40
+    "RES4"                    : [0x0000000000000000, 0x08],  # Offset  48
+    "MAGIC"                   : [0x00000000644d5241, 0x04],  # Offset  56
```
mogasergiu: Unless you are going to change this in the future at runtime, which I doubt, perhaps it would be better to surround it with `()` isntead.

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import sys
+
+# Linux arm64 boot header
+# https://www.kernel.org/doc/Documentation/arm64/booting.txt
+LINUX_ARM64_HDR = {
+    "CODE0"                   : [0x0000000091005a4d, 0x04],  # Offset   0
```
mogasergiu: Ditto

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
...
+    "CODE1"                   : [0x0000000000000000, 0x04],  # Offset   4
+    "LOAD_OFFS"               : [0x0000000000000000, 0x08],  # Offset   8
+    "IMAGE_SIZE"              : [0x0000000000000000, 0x08],  # Offset  16
+    "KERNEL_FLAGS"            : [0x0000000000000000, 0x08],  # Offset  24
+    "RES2"                    : [0x0000000000000000, 0x08],  # Offset  32
+    "RES3"                    : [0x0000000000000000, 0x08],  # Offset  40
+    "RES4"                    : [0x0000000000000000, 0x08],  # Offset  48
+    "MAGIC"                   : [0x00000000644d5241, 0x04],  # Offset  56
+    "RES5"                    : [0x0000000000000040, 0x04],  # Offset  60
```
mogasergiu: Ditto for all `RES*` fields

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
...
+    # kernel_flags
+    #
+    # For now assume little-endian, 4KiB pages, image anywhere in memory
+    LINUX_ARM64_HDR["KERNEL_FLAGS"][0] = 0xa
+
+    # image_size
+    #
+    # We arbitrarily set this to header size + image size aligned up to 2MiB
+    total_size = os.path.getsize(opt.bin) + LINUX_ARM64_HDR_SIZE
+    LINUX_ARM64_HDR["IMAGE_SIZE"][0] = align_up(total_size, 2 * 1024 * 1024)
```
mogasergiu: This does not look python-esque :monocle_face:. `s/2 * 1024 * 1024/2 ** 21/` :triumph:.

support/scripts/mklinux.py
```diff
@@ -0,0 +1,89 @@
...
+
+def autoint(x):
+    return int(x, 0)
+
+def main():
+    parser = argparse.ArgumentParser()
+    description='Prepends image with linux arm64 boot header'
+    parser.add_argument("ram_base", type=autoint, help='ram base')
+    parser.add_argument("img_base", type=autoint, help='image base')
+    parser.add_argument("entry", type=autoint, help='entry point')
```
mogasergiu: Perhaps just personal taste. Is there a reason why you chose to instead offload the fetching of the symbol to a Makefile rule instead of doing so in the script with less arguments. As you've seen this is also how I proceeded in my scripts, I invoke `nm` through python so that the script is self contained and requires as little arguments as possible :thinking:.

mogasergiu (commented):
Some very minor comments, mostly script related. Awesome job bringing this to the superior arch :muscle: :dollar:.

plat/common/bootinfo.c
michpappas: Ok

support/scripts/mklinux.py
michpappas: Ok

support/scripts/mklinux.py
michpappas: Ok

support/scripts/mklinux.py
michpappas: Ok

support/scripts/mklinux.py
michpappas: When it comes to readability I find that `2 * 1024 * 1024` to be the most clear way, compared to `2 ** 21`, `0x400000`, or `4194304`. If you think it's important enough to justify a change, I can do that and append a comment.

support/scripts/mklinux.py
michpappas: I actually consider that but in the end favored against it to avoid code duplication between scripts. The alternative would had been to abstract common code into a library, which I didn't like very much either. On the other hand I don't particularly like the current way either, as much as I don't like invoking `nm` from python, but that's something we won't avoid until we move these to C. Either way I don't have a strong feeling towards any approach.

support/scripts/mklinux.py
mogasergiu: I am fine with this either way. I understand very well what it's like being reluctant to change large commits :laughing: :wink: :stuck_out_tongue_winking_eye:. It looks more hippie with `**` though :nerd_face: :point_up: .

support/scripts/mklinux.py
mogasergiu: > Either way I don't have a strong feeling towards any approach.

Me neither, we want to get rid of these scripts anyway :1234:. Either way is fine for me as well :smile:. Go with how you feel :raised_hands:.

support/scripts/mklinux.py
michpappas: Updated to `2 ** 21`. There is already a comment above that states this is 2MiB, so that should be clear enough.

support/scripts/mklinux.py
michpappas: Copied the `get_sym_val` function from `mkefi.py` and reduced the arguments to raw binary and elf file. Btw, that was another reason I didn't like this approach, as I found passing addresses more elegant than passing raw binary + ELF (and I didnt' want to call objcopy from the script).

michpappas:
Pushed update addressing review comments.

support/scripts/mklinux.py
mogasergiu: I am fine with either, especially since we intend to delete them anyway in favor of C-written ones. :angel:

plat/kvm/Config.uk
```diff
@@ -29,13 +29,13 @@ config KVM_BOOT_PROTO_LXBOOT
 	help
 		Linux 64-bit Boot Protocol
 
-config KVM_BOOT_PROTOCOL_QEMU_VIRT
```
mogasergiu, line 32:
```
Rename the Kconfig options for QEMU virt and EFI stub to make them
with the rest of boot protocol options.
```
to
```
Rename the Kconfig options for QEMU virt and EFI stub to align them
with the rest of boot protocol options.
```
for commit `plat/kvm: Make Kconfig boot protocol option naming consistent`

plat/common/Makefile.rules
```diff
@@ -17,3 +17,12 @@ endef
 define build_efi =
 	@$(SCRIPTS_DIR)/mkefi.py $(1)
 endef
+
+define get_symbol_addr =
```
mogasergiu: Is this used anywhere else now that you do the fetching in the script?

plat/kvm/Config.uk
```diff
@@ -45,25 +47,25 @@ config KVM_BOOT_EFI_STUB
...
-if KVM_BOOT_EFI_STUB
+if KVM_BOOT_PROTO_EFI_STUB
 
-config KVM_BOOT_EFI_STUB_DEBUG
+config KVM_BOOT_PROTO_EFI_STUB_DEBUG
 	bool "Enable EFI stub crash messages"
 	default n
 
-config KVM_BOOT_EFI_STUB_CMDLINE_FNAME
+config KVM_BOOT_PROTO_EFI_STUB_CMDLINE_FNAME
```
mogasergiu, line 58: Thank you so much for these changes! :pray: My apologies for having missed them.

mogasergiu (changes requested):
Aweseom PR as always! Some minor comments.

plat/common/Makefile.rules
michpappas: No, I forgot to remove it, thanks 👍🏼

plat/kvm/Config.uk
michpappas: Good catch, thanks 👍🏼

michpappas:
Rebased from `staging` and addressed reviewer comments.

mogasergiu (approved):
Beautiful!

razvand (approved):

PR #991, opened 2023-07-24
lib/vfscore: Accept negative timestamps

andreittr:
### Description of changes

Previously vfs code would reject timestamps that are more than 1 second before the epoch as invalid. This is wrong; unix time is defined as a signed offset and as such can and must support dates before the epoch. This change relaxes the overly-strict validity check, allowing the full range of valid file timestamps to be used.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Test with:
```
struct utimbuf t = {-1, -1};
int r = utime("file", &t);
if (r) perror("Bug in utime");
```

### DIFF
--- a/lib/vfscore/syscalls.c
+++ b/lib/vfscore/syscalls.c
@@ -1296,8 +1296,7 @@ sys_readlink(char *path, char *buf, size_t bufsize, ssize_t *size)
  */
 static int is_timeval_valid(const struct timeval *time)
 {
-	return (time->tv_sec >= 0) &&
-		   (time->tv_usec >= 0 && time->tv_usec < 1000000);
+	return (time->tv_usec >= 0 && time->tv_usec < 1000000);
 }
 
 /*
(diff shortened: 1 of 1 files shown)


StefanJum (changes requested):
@andreittr the changes make sense, but shouldn't they be made in [`is_timeval_valid`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/syscalls.c#L1297) too (since that is the function that is eventually called by the `utime` syscall)?

andreittr:
> @andreittr the changes make sense, but shouldn't they be made in [`is_timeval_valid`](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/syscalls.c#L1297) too (since that is the function that is eventually called by the `utime` syscall)?

Good catch, thanks! Updated with fix.

StefanJum (approved):
All good, thank you.

razvand (approved):

PR #952, opened 2023-06-21
lib/vfscore: Fix double-lock bug in symlink

andreittr:
### Description of changes

This change fixes an issue where the lock of the destination parent directory was being taken more than once in the  symlink syscall by using a locked variant of namei_last_nofollow, which this PR also adds. In implementing the locked version of the latter function there's also some factoring out of common code.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

`checkpatch.uk` complains, but I've purposely kept the style consistent within the file.

### Base target

 - Architecture(s): all
 - Platform(s): all
 - Application(s): all

### Additional configuration

Test by issuing a symlink syscall; staging fails with a lock assert, fixed code behaves as expected.

PR #633, opened 2022-11-04
lib/vfscore: Change `pipe()` call to `uk_syscall_r_pipe()`

eduardvintila:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [`aarch64`]
 - Platform(s): [`kvm`]
 - Application(s): [`app-redis`]

### Description of changes

The `pipe2` implementation in vfscore relies on `pipe`. This commit ensures that we call the `pipe` system call from unikraft core and not a libc wrapper which in turn might call `pipe2` and thus end in an infinite loop. Musl ends up wrapping `pipe` exactly this way on `aarch64`, for example.

Signed-off-by: Eduard Vintilă <eduard.vintila47@gmail.com>

PR #992, opened 2023-07-24
lib/ukcpio: Fix symlinks not being extracted

andreittr:
### Description of changes

Previously the CPIO extraction code would only handle sections whose header was of type directory or file, skipping symlinks. This change adds code to handle extracting symlinks from CPIO archives, as well as a warning message for unknown CPIO sections.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): [N/A

### Additional configuration
Test with an initrd root that contains symlinks.


Starnox (approved):
Not something that I used before but for sure can see the benefits. I tried an initRD config and it behaved as expected. As for the code, I don't have any issues with it, so from me, you have the green-light.

Reviewed-by: Eduard-Florin Mihailescu [mihailescu.eduard@gmail.com](mailto:mihailescu.eduard@gmail.com)

lib/ukcpio/cpio.c
```diff
@@ -292,6 +294,36 @@ read_section(struct cpio_header **header_ptr,
...
+			error = -UKCPIO_MALFORMED_INPUT;
+			goto out;
+		}
+
+		char target[header_filesize + 1];
+		memcpy(target, data_location, header_filesize);
+		target[header_filesize] = 0;
+
+		uk_pr_info("%s: Target is %s\n", path_from_root, target);
+		if (symlink(target, path_from_root)) {
```
StefanJum:
```suggestion
		if (uk_syscall_r_symlink(target, path_from_root)) {
```

lib/ukcpio/cpio.c
```diff
@@ -292,6 +294,36 @@ read_section(struct cpio_header **header_ptr,
...
 			goto out;
 		}
+	} else if (IS_SYMLINK(header_mode)) {
+		uk_pr_info("Creating symlink %s\n", path_from_root);
+
+		data_location = (char *)ALIGN_4(
+				(char *)(header) + sizeof(struct cpio_header)
+				+ header_namesize);
+
+		if ((uintptr_t)data_location + header_filesize > last) {
```
StefanJum, line 304:
```suggestion
		if (unlikely((uintptr_t)data_location + header_filesize > last)) {
```

lib/ukcpio/cpio.c
StefanJum:
```suggestion
		if (unlikely(symlink(target, path_from_root))) {
```

StefanJum (changes requested):
This looks good, few minor comments.
Besides those, when I try to run an application with a symlink in the initrd fs, I get the following error:
```
[    0.171998] Info: [libukcpio] <cpio.c @  298> Creating symlink /./file2
[    0.172907] Info: [libukcpio] <cpio.c @  316> /./file2: Target is file1
[    0.173839] dbg:  [libvfscore] <main.c @ 1532> (int) uk_syscall_r_symlink((const char*) 0x19beb0, (const char*) 0x1ac020)
[    0.175638] CRIT: [libvfscore] <vnode.c @  124> Assertion failure: m->owner != cur
```

Also please rebase this on top of staging so the ci/cd tests can run.

andreittr:
> This looks good, few minor comments. Besides those, when I try to run an application with a symlink in the initrd fs, I get the following error:

@StefanJum You might need to apply this PR to fix symlink bugs in general: https://github.com/unikraft/unikraft/pull/952.

lib/ukcpio/cpio.c
andreittr: The rest of the file uses stdlib-like file access functions like [open](https://github.com/unikraft/unikraft/pull/992/files#diff-16c7dcd3c3387d5b8d30d83b32014a1bd9f4354b64f2deb66bfde8e9eec83cccR237), [write](https://github.com/unikraft/unikraft/pull/992/files#diff-16c7dcd3c3387d5b8d30d83b32014a1bd9f4354b64f2deb66bfde8e9eec83cccR263), [chmod](https://github.com/unikraft/unikraft/pull/992/files#diff-16c7dcd3c3387d5b8d30d83b32014a1bd9f4354b64f2deb66bfde8e9eec83cccR276), etc., and I assume that's a design decision to not depend on syscall-shim. I kept the same convention for the symlink code.

lib/ukcpio/cpio.c
andreittr: I understand the point, however this being here only for the symlink case and not the others would suggest that symlinks are less likely to fail these checks compared to regular files or dirs (which isn't AFAICT accurate).
A cleaner way to integrate this change then would be to do it file- or lib-wide in its own changeset; thoughts?

StefanJum: Ok, makes sense, we can do this lib-wide.

lib/ukcpio/cpio.c
StefanJum: The problem with libc-style syscalls is that you will exit the "kernel space" code and land into an external libc, that you can not control. There were some issues when this would lead to an infinite loop and other problems (see [this pr](https://github.com/unikraft/unikraft/pull/633)), so we decided to use raw syscalls from whithin the Unikraft core, so the execution will not leave the core space.

The `uk_syscall_*` symbols are defined even if the `syscall-shim` is not enabled, so most likely the rest of the file was written before we ran into issues with the libc-style syscalls.

StefanJum:
> @StefanJum You might need to apply this PR to fix symlink bugs in general: #952.

Yes, that was it, it works now.

lib/ukcpio/cpio.c
andreittr: Right, I'm aware of the dangers of calling into libc from critical code but hadn't known about the raw syscall policy, nor that this file doesn't necessarily use libcalls on purpose; thanks for clarifying!

I'll use syscalls for symlink in this PR, and add fixing up the rest of the file (or module (or unikraft)) to my backlog to be handled in a future PR.

andreittr:
Update: rebased on staging; changed symlink to syscall
Deferring adding likely/unlikely to branches & changing over all other libcalls into syscalls to separate changesets.

StefanJum (approved):
All good now, thanks @andreittr :rocket: 

razvand (approved):

PR #996, opened 2023-07-24
lib/ramfs: Ensure stability of inode numbers

andreittr:
### Description of changes

Previously ramfs would assign a new unique inode number to a file on every successful filesystem lookup. This can lead to the same file being allocated multiple different inodes, breaking code that (correctly) uses inodes as reliable indicators of file identity.
This change makes inode numbers persistent, allocated along with the filesystem node, and stable across multiple lookups of the same file.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Test with:
```
int f = open("file", O_RDONLY);
fstat(f, &st1);
stat("file", &st2);
printf("Should be same: %ld %ld\n", st1.st_ino, st2.st_ino);
```

### DIFF
--- a/lib/ramfs/ramfs.h
+++ b/lib/ramfs/ramfs.h
@@ -47,6 +47,8 @@ struct ramfs_node {
 	 * else NULL
 	 */
 	struct ramfs_node *rn_child;
+	/* Inode number of this node, should be unique and stable */
+	uint64_t rn_ino;
 	/*
 	 * Entry type: regular file - VREG, symbolic link - VLNK,
 	 * or directory - VDIR
--- a/lib/ramfs/ramfs_vnops.c
+++ b/lib/ramfs/ramfs_vnops.c
@@ -92,6 +92,7 @@ ramfs_allocate_node(const char *name, int type)
 	}
 	strlcpy(np->rn_name, name, np->rn_namelen + 1);
 	np->rn_type = type;
+	np->rn_ino = inode_count++;
 
 	if (type == VDIR)
 		np->rn_mode = S_IFDIR|0777;
@@ -230,7 +231,7 @@ ramfs_lookup(struct vnode *dvp, const char *name, struct vnode **vpp)
 		uk_mutex_unlock(&ramfs_lock);
 		return ENOENT;
 	}
-	if (vfscore_vget(dvp->v_mount, inode_count++, &vp)) {
+	if (vfscore_vget(dvp->v_mount, np->rn_ino, &vp)) {
 		/* found in cache */
 		*vpp = vp;
 		uk_mutex_unlock(&ramfs_lock);


lib/ramfs/ramfs.h
```diff
@@ -47,6 +47,7 @@ struct ramfs_node {
 	 * else NULL
 	 */
 	struct ramfs_node *rn_child;
+	uint64_t rn_ino;
```
RaduNichita, line 51: Please add a comment about the need for this member in the structure.

lib/ramfs/ramfs_vnops.c
```diff
@@ -93,6 +93,8 @@ ramfs_allocate_node(const char *name, int type)
 	strlcpy(np->rn_name, name, np->rn_namelen + 1);
 	np->rn_type = type;
 
+	np->rn_ino = inode_count++;
```
RaduNichita, line 95: Why have an empty line before?

RaduNichita (changes requested):
This seems OK, thanks @andreittr! I left some minor nitpicks

RaduNichita:
@andreittr, is it intended to open the file without the `O_CREAT` flag?

I am asking this since for Linux, it creates a file, but the inode number differs at the second attempt or running (0 vs a large number).

andreittr:
> @andreittr, is it intended to open the file without the `O_CREAT` flag?
> 
> I am asking this since for Linux, it creates a file, but the inode number differs at the second attempt or running (0 vs a large number).

Yes, the test snippet assumes "file" exists and is a regular file created ahead of time (e.g. already included in the ramfs/initrd)

mariasfiraiala:
@andreittr The changes look fine, however I wasn't able to reproduce the issue. Could you please share a bit more regarding your setup?

This is what I've tested:

```C
int main(void)
{
	struct stat st1 = {0};
	struct stat st2 = {0};

	int f = open("file", O_CREAT);
	fstat(f, &st1);
	stat("file", &st2);
	printf("Should be same: %ld %ld\n", st1.st_ino, st2.st_ino);
}
```

And this is the output:

ttr-inode

Notice how the [break point at which the static variable should be incremented](https://github.com/unikraft/unikraft/blob/staging/lib/ramfs/ramfs_vnops.c#L233) is reached only once.

andreittr:
> @andreittr The changes look fine, however I wasn't able to reproduce the issue. Could you please share a bit more regarding your setup?
> 
> This is what I've tested:
> 
> ```c
> int main(void)
> {
> 	struct stat st1 = {0};
> 	struct stat st2 = {0};
> 
> 	int f = open("file", O_CREAT);
> 	fstat(f, &st1);
> 	stat("file", &st2);
> 	printf("Should be same: %ld %ld\n", st1.st_ino, st2.st_ino);
> }
> ```
> 
> And this is the output:
> 
> ttr-inode
> 
> Notice how the [break point at which the static variable should be incremented](https://github.com/unikraft/unikraft/blob/staging/lib/ramfs/ramfs_vnops.c#L233) is reached only once.

Hi! See the above comment as well as the test snippet in the description. `open()` is not supposed to be called with `O_CREAT`, rather just with `O_RDONLY` and "file" needs to already exist. This is necessary to cause multiple FS lookups of the file, resulting in different inode numbers.

andreittr:
Pushed changes addressing comments.

mariasfiraiala (approved):
I see, all good on my side, then. Let's merge this!

RaduNichita:
It works for me as well.

I used something more hackish:

```c
int main(void)
{
        struct stat st1, st2;
        int f = open("file", O_CREAT);
        fstat(f, &st1);
        close(f);

        f = open("file", O_RDONLY);
        stat("file", &st2);
        printf("Should be same: %ld %ld\n", st1.st_ino, st2.st_ino);
        close(f);
        return 0;
}
```

but it does the trick 

Thanks, @andreittr 

RaduNichita (approved):

razvand (approved):

PR #998, opened 2023-07-24
lib/vfscore: Fix create through broken symlink

andreittr:
### Description of changes

Previously a vfs syscall on a broken symlink would fail with ENOENT instead of attempting the syscall on the symlink target. This would make open(O_CREAT) to fail instead of creating the missing target file. This change fixes this behavior, bringing it in line with expectations.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Test with:
```
symlink("nonexistent", "broken-link");
int f = open("broken-link", O_WRONLY|O_CREAT, 0644);
if (f < 0) perror("bug in open()");
```

### DIFF
--- a/lib/vfscore/lookup.c
+++ b/lib/vfscore/lookup.c
@@ -96,13 +96,14 @@ namei_follow_link(struct dentry *dp, char *node, char *name, char *fp, size_t mo
 	return (0);
 }
 /*
- * Convert a pathname into a pointer to a dentry
+ * Resolve a pathname into a pointer to a dentry and a realpath.
  *
  * @path: full path name.
  * @dpp:  dentry to be returned.
+ * @realpath: if not NULL, return path after resolving all symlinks
  */
 int
-namei(const char *path, struct dentry **dpp)
+namei_resolve(const char *path, struct dentry **dpp, char *realpath)
 {
 	char *p;
 	char node[PATH_MAX];
@@ -122,6 +123,7 @@ namei(const char *path, struct dentry **dpp)
 	memset(fp, 0, PATH_MAX);
 	strlcpy(fp, path, PATH_MAX);
 
+	error = 0;
 	do {
 		need_continue = 0;
 		/*
@@ -129,7 +131,8 @@ namei(const char *path, struct dentry **dpp)
 		 * the local node in the file system.
 		 */
 		if (vfs_findroot(fp, &mp, &p)) {
-			return ENOTDIR;
+			error = ENOTDIR;
+			goto out;
 		}
 
 		size_t mountpoint_len = p - fp;
@@ -140,7 +143,7 @@ namei(const char *path, struct dentry **dpp)
 		if (dp) {
 			/* vnode is already active. */
 			*dpp = dp;
-			return 0;
+			goto out;
 		}
 		/*
 		 * Find target vnode, started from root directory.
@@ -189,7 +192,7 @@ namei(const char *path, struct dentry **dpp)
 				if (error) {
 					vn_unlock(dvp);
 					drele(ddp);
-					return error;
+					goto out;
 				}
 
 				dp = dentry_alloc(ddp, vp, node);
@@ -198,7 +201,8 @@ namei(const char *path, struct dentry **dpp)
 				if (!dp) {
 					vn_unlock(dvp);
 					drele(ddp);
-					return ENOMEM;
+					error = ENOMEM;
+					goto out;
 				}
 			}
 			vn_unlock(dvp);
@@ -209,7 +213,7 @@ namei(const char *path, struct dentry **dpp)
 				error = namei_follow_link(dp, node, name, fp, mountpoint_len);
 				if (error) {
 					drele(dp);
-					return (error);
+					goto out;
 				}
 
 				drele(dp);
@@ -223,7 +227,8 @@ namei(const char *path, struct dentry **dpp)
 				node[0] = 0;
 
 				if (++links_followed >= MAXSYMLINKS) {
-					return (ELOOP);
+					error = ELOOP;
+					goto out;
 				}
 				need_continue = 1;
 				break;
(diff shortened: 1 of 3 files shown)


lib/vfscore/lookup.c
```diff
@@ -223,21 +227,37 @@ namei(const char *path, struct dentry **dpp)
...
+	return error;
+}
+/*
+ * Convert a pathname into a pointer to a dentry
+ *
+ * @path: full path name.
+ * @dpp:  dentry to be returned.
+ */
+int
+namei(const char *path, struct dentry **dpp)
```
RaduNichita, line 258:
```suggestion
int namei(const char *path, struct dentry **dpp)
```

lib/vfscore/lookup.c
```diff
@@ -96,13 +96,14 @@ namei_follow_link(struct dentry *dp, char *node, char *name, char *fp, size_t mo
...
- * Convert a pathname into a pointer to a dentry
+ * Resolve a pathname into a pointer to a dentry and a realpath.
  *
  * @path: full path name.
  * @dpp:  dentry to be returned.
+ * @realpath: if not NULL, return path after resolving all symlinks
  */
 int
-namei(const char *path, struct dentry **dpp)
+namei_resolve(const char *path, struct dentry **dpp, char realpath[PATH_MAX])
```
mariasfiraiala:
```suggestion
namei_resolve(const char *path, struct dentry **dpp, char *realpath)
```

Let's have this without the specified size, just like it happens for the other argument.

lib/vfscore/lookup.c
RaduNichita: You can leave it on the same line, as mentioned in the coding - conventions PR: https://github.com/unikraft/docs/pull/248

> Prefer the Linux style with the return type on the same line as the function name.
In case of long functions + first argument, you can break after the return type.

lib/vfscore/vfs.h
```diff
@@ -684,6 +684,22 @@ int sec_vnode_permission(char *path);
...
+ *	The full path name
+ * @param[out] dpp
+ *	Dentry to be returned
+ * @param[out] realpath
+ *	if not NULL, return path after resolving all symlinks
+ * @return
+ *	- (0):  Completed successfully
+ *	- (<0): Negative value with error code
+ */
+int
```
RaduNichita: Same here

lib/vfscore/lookup.c
```diff
@@ -223,21 +227,37 @@ namei(const char *path, struct dentry **dpp)
...
+				goto out;
 			}
 		}
 	} while (need_continue);
 
 	*dpp = dp;
-	return 0;
+out:
+	if (realpath)
+		strcpy(realpath, fp);
```
mariasfiraiala:
```suggestion
		strlcpy(realpath, fp, PATH_MAX);
```

We can use `strlcpy()` in order to keep the consistency [within the file](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/lookup.c#L87) and for better security.

lib/vfscore/lookup.c
```diff
@@ -223,21 +227,37 @@ namei(const char *path, struct dentry **dpp)
...
+	return error;
+}
+/*
+ * Convert a pathname into a pointer to a dentry
```
RaduNichita, line 252: I think the commentary is not needed, since it's very similar in [vfs.h](https://github.com/unikraft/unikraft/blob/staging/lib/vfscore/vfs.h#L675)

RaduNichita (changes requested):

mariasfiraiala (changes requested):
Thanks @andreittr. See my few comments below. Other than that I wasn't able to reproduce the issue with the example you provided, though the changes make sense.

andreittr:
> Thanks @andreittr. See my few comments below. Other than that I wasn't able to reproduce the issue with the example you provided, though the changes make sense.

What behavior are you getting when trying to reproduce? If the symlink call fails you might need to apply this PR: https://github.com/unikraft/unikraft/pull/952.

lib/vfscore/lookup.c
andreittr: Indeed the comments are duplicates of the docstrings in vfs.h, however the file already has this style for exported functions and I don't want to break it without good reason.

lib/vfscore/lookup.c
andreittr: All of lookup.c uses this format, and I believe maintaining consistency is important.
If we however believe this style actively harms readability, we can address it for the entire file in a style change PR.

In vfs.h: sure, there it makes sense, and the rest of the file follows suit as well.

mariasfiraiala:
> If the symlink call fails you might need to apply this PR: https://github.com/unikraft/unikraft/pull/952. 

I had already applied that PR since I also reviewed it.

> What behavior are you getting when trying to reproduce?

The bug is not reproducible on `9pfs` (meaning that the error message doesn't get printed), but with `ramfs`, the behaviour is indeed the one you mentioned, so all good.

@andreittr I'll approve the PR after you apply the suggestions, thank you for this!

andreittr:
Pushed changes addressing comments.

> The bug is not reproducible on `9pfs` (meaning that the error message doesn't get printed), but with `ramfs`, the behaviour is indeed the one you mentioned, so all good.

That's really odd that 9pfs isn't affected, as the code paths are common between filesystems :shrug: .

mariasfiraiala (approved):
All good on my side now, thanks a lot for the fixes @andreittr!

RaduNichita:
All good also from my side. As @mariasfiraiala, I have only managed to reproduce the bug on ramfs, not 9pfs.

RaduNichita (approved):

razvand (approved):

PR #1001, opened 2023-07-24
build: Add tree builds

andreittr:
### Description of changes

This PR adds tree builds -- builds that plate the products in a directory tree mirroring the source files.
The root of this source tree relative to which source files are considered is taken from the `$(LIBNAME)_SRC` make variable, if present.
This feature is most useful in libraries with many files sharing the same basename (e.g. sources named after class names). 

It consists of 3 commits:
- Add a separate `build/` directory to place the build tree in (so source directories won't conflict with unikraft files); this feature is active if `$(LIBNAME)_BUILDTREE` is set to `y`.
- Adapt path processing functions to retain source hierarchy when `$(LIBNAME)_BUILDTREE` is set to `y`.
- Add make functions `addlib_tree` and `addlib_tree_s` that automate setting `$(LIBNAME)_BUILDTREE`.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Use addlib_tree[_s] to register a treebuild library.

### DIFF
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -82,17 +82,18 @@ endef
 #
 ################################################################################
 
+# Returns the fully-qualified path for a subfolder under the build output
+# directory
+sub_build_dir = $(addprefix $(BUILD_DIR)/,$(if $($(call uc,$(1))__BUILDTREE),\
+	$(notdir $(1))/build,$(notdir $(1))))
+
 # Creates a subfolder under the build output directory
 # mk_sub_build_dir $1:subfoldername
 define mk_sub_build_dir =
-$(if $(shell $(MKDIR) -p "$(BUILD_DIR)/$(notdir $(1))" && cd "$(BUILD_DIR)/$(notdir $(1))" >/dev/null && pwd),,\
-     $(error could not create directory "$(BUILD_DIR)/$(notdir $(1))"))
+$(eval _d := $(call sub_build_dir,$(1)))\
+$(if $(shell $(MKDIR) -p "$(_d)" && cd "$(_d)" >/dev/null && pwd),,\
+     $(error could not create directory "$(_d)"))
 endef
-
-# Returns the fully-qualified path for a subfolder under the build output
-# directory
-# sub_build_dir $1:subfoldername
-sub_build_dir = $(addprefix $(BUILD_DIR)/,$(notdir $(1)))
 
 # Returns the file extension
 fileext = $(subst .,,$(suffix $(1)))
@@ -269,14 +270,15 @@ define addplatlib_s =
 $(if $(filter y,$(3)),$(call addplatlib,$(1),$(2)),)
 endef
 
+sub_libbuild_dir = $(call sub_build_dir,$(1))/$(strip $(2))
+
 # creates a sub build directory for a library
 # mk_sub_libbuild_dir $1:libname,$2:subdir
 define mk_sub_libbuild_dir =
-$(if $(shell $(MKDIR) -p "$(BUILD_DIR)/$(notdir $(1))/$(strip $(2))" && cd "$(BUILD_DIR)/$(notdir $(1))/$(strip $(2))" >/dev/null && pwd),,\
-     $(error could not create directory "$(BUILD_DIR)/$(notdir $(1))/$(strip $(2))"))
+$(eval _d := $(call sub_libbuild_dir,$(1),$(2)))\
+$(if $(shell $(MKDIR) -p "$(_d)" && cd "$(_d)" >/dev/null && pwd),,\
+     $(error could not create directory "$(_d)"))
 endef
-
-sub_libbuild_dir = $(BUILD_DIR)/$(notdir $(1))/$(strip $(2))
 
 ################################################################################
 #
@@ -432,7 +434,7 @@ $(error $(2): missing extraction rule for archive type)\
 ))))
 UK_FETCH-y += $(BUILD_DIR)/$(1)/.origin \
 $(eval $(call vprefix_lib,$(1),ORIGIN) = $(BUILD_DIR)/$(1)/origin)
-$(call mk_sub_libbuild_dir,$(1),origin)
+$(shell $(MKDIR) -p $(BUILD_DIR)/$(1)/origin)
 
 .PRECIOUS: $(BUILD_DIR)/$(1)/origin
 
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -271,10 +271,23 @@ $(eval $(call uc,$(1))_EXPORTS += $(wildcard $(IMPORT_BASE)/exportsyms.uk))
 $(eval $(call uc,$(1))_LOCALS  += $(wildcard $(IMPORT_BASE)/localsyms.uk))
 endef
 
+# addlib_tree $1:libname
+define addlib_tree =
+$(eval $(call uc,$(1))__BUILDTREE := y)
+$(call addlib,$(1))
+endef
+
 # addlib_s $1:libname,$2:switch
 define addlib_s =
 ifeq ($(2),y)
 $(call addlib,$(1))
+endif
+endef
+
+# addlib_tree_s $1:libname,$2:switch
+define addlib_tree_s =
+ifeq ($(2),y)
+$(call addlib_tree,$(1))
 endif
 endef
 
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -104,17 +104,41 @@ libname2olib = $(addprefix $(BUILD_DIR)/,$(addsuffix .o,$(1)))
 
 libname2preolib = $(addprefix $(BUILD_DIR)/,$(addsuffix .ld.o,$(1)))
 
+# check if $1 is a prefix of $2; expands to $2 if true
+isprefix = $(findstring $(1)$(subst $(1),,$(2)),$(2))
+
+# convert a full path to a source file into a relative path
+# result is relative to $(libname)_SRC, if defined and a prefix of $source
+# otherwise it is the filename of $source
+# srcrelpath $libname,$source
+define srcrelpath =
+$(eval _libsrc := $(if $($(call uc,$(1))__BUILDTREE),$($(call uc,$(1))_SRC)))\
+$(if $(and $(_libsrc),$(call isprefix,$(_libsrc),$(2))),\
+	$(2:$(_libsrc)/%=%),$(notdir $(2)))
+endef
+
+# convert a full path to a build target into a path relative to
+# the build dir of $libname if it is prefix of $target
+# otherwise return the filename of $target
+# targrelpath $libname,$target
+define targrelpath =
+$(eval _libbuild := $(if $($(call uc,$(1))__BUILDTREE),\
+	$(call sub_build_dir,$(1))))\
+$(if $(and $(_libbuild),$(call isprefix,$(_libbuild),$(2))),\
+	$(2:$(_libbuild)/%=%),$(notdir $(2)))
+endef
+
 # converts a list of sources to paths pointing to their corresponding destination files
 # src2dst $1:libname,$2:source(s),$3:destsuffix,$4:variant(optional),$5:subbuild(optional)
 define src2dst =
 $(if $(5),\
 $(if $(4),\
-$(addprefix $(call sub_libbuild_dir,$(1),$(5))/,$(addsuffix .$(4)$(3),$(basename $(notdir $(2))))),\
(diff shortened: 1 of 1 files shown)


razvand:
@andreittr, this fails when I try to build [`app-helloworld`](https://github.com/unikraft/app-helloworld). Does it work on your side? I'm cloning [the `app-helloworld` repository](https://github.com/unikraft/app-helloworld), [the `unikraft` repository](https://github.com/unikraft/unikraft), and I'm applying this PR. And I get this error:

```console
$ make
make[1]: Entering directory '/media/razvan/c4f6765a-efa5-4ebd-9cf0-7da9908a0189/razvan/unikraft/scripts/workdir/unikraft'
mkdir: cannot create directory ‘’: No such file or directory
/media/razvan/c4f6765a-efa5-4ebd-9cf0-7da9908a0189/razvan/unikraft/scripts/workdir/unikraft/lib/ukstore/Makefile.uk:3: *** could not create directory "".  Stop.
Makefile:1033: recipe for target 'sub-make' failed
make[2]: *** [sub-make] Error 2
Makefile:32: recipe for target '_all' failed
make[1]: *** [_all] Error 2
make[1]: Leaving directory '/media/razvan/c4f6765a-efa5-4ebd-9cf0-7da9908a0189/razvan/unikraft/scripts/workdir/unikraft'
Makefile:6: recipe for target 'all' failed
make: *** [all] Error 2
```

StefanJum (commented):
Hi @andreittr. Can you give an idea on how the build directory should look like and on what application is this relevant to test?

StefanJum:
> @andreittr, this fails when I try to build [`app-helloworld`](https://github.com/unikraft/app-helloworld). Does it work on your side? I'm cloning [the `app-helloworld` repository](https://github.com/unikraft/app-helloworld), [the `unikraft` repository](https://github.com/unikraft/unikraft), and I'm applying this PR. And I get this error:

@razvand  are you adding anything else in the config menu or something? I tried it on my system and on the remote systems and it seems to work fine.

razvand:
> @razvand are you adding anything else in the config menu or something? I tried it on my system and on the remote systems and it seems to work fine.

I get this problem on a Ubuntu 18.04 setup. You could use the Docker-based configuration [here](https://github.com/unikraft-upb/maintainer-tools/tree/staging/docker/ubuntu-18.04).

StefanJum:
> I get this problem on a Ubuntu 18.04 setup. You could use the Docker-based configuration [here](https://github.com/unikraft-upb/maintainer-tools/tree/staging/docker/ubuntu-18.04).

Yes, I get the same error in the container. It fails [here](https://github.com/unikraft/unikraft/blob/staging/lib/ukstore/Makefile.uk#L3), tries to create the directory [here](https://github.com/unikraft/unikraft/pull/1001/files#diff-f535de85a92c17723a9b46772fd6d8d8252feb9802a5079056b253d1bdd01df3R67) and does it (`build/libukstore/` exists), but the `if` fails for some reason

support/build/Makefile.rules
```diff
@@ -203,13 +229,26 @@ $(eval $(call uc,$(1))_EXPORTS += $(wildcard $(_IMPORT_BASE)/exportsyms.uk))
 $(eval $(call uc,$(1))_LOCALS  += $(wildcard $(_IMPORT_BASE)/localsyms.uk))
 endef
 
+# addlib_tree $libname
```
jschlumpp: What is the benefit of having a function for defining `*_BUILDTREE`? Is the ordering important? (`_BUILDTREE` then `addlib`)

andreittr:
> Hi @andreittr. Can you give an idea on how the build directory should look like and on what application is this relevant to test?

Hi! Yes, this library PR right here: https://github.com/unikraft/lib-geos/pull/1 makes use of tree building. It uses `addlib_tree_s` & defines the LIBGEOS_SRC path.

mariasfiraiala (approved):
Thanks @andreittr, this works fine!

StefanJum (changes requested):
Thank you @andreittr. This seems to work fine, but we still have the issue mentioned above for ubuntu 18.04. 
Also please rebase the pr on top of the staging branch so the ci/cd tests can run

StefanJum:
This still fails on ubuntu 18.04:
```
mkdir: cannot create directory '': No such file or directory
/workdir/unikraft/lib/ukstore/Makefile.uk:3: *** could not create directory "".  Stop.
Makefile:1055: recipe for target 'sub-make' failed
```

It's all good besides that. @razvand do we merge it as it is or do we search for a fix?

razvand:
> It's all good besides that. @razvand do we merge it as it is or do we search for a fix?

No, we have to fix that. @andreittr, any idea what's causing this? Could you check on a container running Ubuntu 18.04?

andreittr:
> No, we have to fix that. @andreittr, any idea what's causing this? Could you check on a container running Ubuntu 18.04?

@razvand found the issue while rebasing against staging; it was a typo that was excercised by new code.

Pushed rebase w/ fix, should work ok now.

support/build/Makefile.rules
andreittr: The function is there as a more dev-friendly "API" of sorts. IMO it's cleaner than "remember to define this magic variable, probably before doing `addlib`, but it depends on deferred expansion" (although the system's use of `*_SRC` in `srcrelpath` and `targrelpath` could be better). It's also an abstraction layer in case we give it its own dedicated implementation instead of `*_BUILDTREE` var.

As for the ordering, as alluded to, it shouldn't matter for deferred-expanded stuff, but that isn't always the case, and sometimes you can get e.g., eagerly run `mkdir`s on all build directories, so I prefer to ensure `_BUILDTREE` is defined before lib registration.

razvand:
@StefanJum, @mschlumpp, please take another look.

support/build/Makefile.rules
```diff
@@ -271,13 +271,26 @@ $(eval $(call uc,$(1))_EXPORTS += $(wildcard $(IMPORT_BASE)/exportsyms.uk))
 $(eval $(call uc,$(1))_LOCALS  += $(wildcard $(IMPORT_BASE)/localsyms.uk))
 endef
 
+# addlib_tree $1:libname
+define addlib_tree =
```
skuenzer, line 275: Really good to have a different library registration call 👍🏻

support/build/Makefile.rules
```diff
@@ -271,13 +271,26 @@ $(eval $(call uc,$(1))_EXPORTS += $(wildcard $(IMPORT_BASE)/exportsyms.uk))
...
+$(eval $(call uc,$(1))_BUILDTREE := y)
```
skuenzer: Maybe we want to hide this variable with a `_` prefix. What do you think about naming it `_$(call uc,$(1))_TREEMODE`?

support/build/Makefile.rules
skuenzer: I agree and also prefer to have the mode selected together with the library registration because of exactly these "API friendlyness" reasons. Another variant could be to add another optional parameter to `addlib` and `addlib_s` for enabling tree mode. But the concept is the same.
```
addlib $1:libname,$2:tree_mode(optional, 'y' for on)
```
The optional parameter won't break existing libraries because Make will still call the same functions defintion with having the second parameter empty.

jschlumpp (changes requested):
I assume there is reason to introduce `_SRC` instead of relying on `_BASE`?
Edit: Thinking about it, I guess it is necessary for external sources from downloaded source trees.

Also `build: Add option to place build objects in a tree` seems to break the formatting of the build output by inserting a space in front of all filenames:
```
  LDS     libkvmplat:  bootinfo.lds
  LDS     libkvmplat:  link64.lds
  LD       libkvmplat.ld.o
  OBJCOPY  libkvmplat.o
  CC      libkvmpci:  pci_bus.common.o
```

StefanJum:
Looks good to me now @andreittr. Is it rebased on staging? It looks there are some weird ci/cd weird fails.

jschlumpp:
@StefanJum I think it was just GitHub actions being broken for some reason. I triggered a retry of the failed jobs and they are now successful.

support/build/Makefile.rules
andreittr: Yeah, good idea if the var is never meant to be exposed to the user. However, wouldn't it be cleaner to keep the libname prefix as-is and preceed the variable name itself with an underscore (i.e., `LIBNAME__TREEMODE` rather than `_LIBNAME_TREEMODE`) so that all of LIBNAME's variables are of the form `LIBNAME_*`?

andreittr:
> Also `build: Add option to place build objects in a tree` seems to break the formatting of the build output by inserting a space in front of all filenames:

Hmm, that seems to be due to use of the `$(eval ...)` line in `targrelpath`; fixed it by removing the preceding space when invoking `verbose_cmd`.

jschlumpp:
It's still a problem for commands without a library specifier:
```
  CC      libkvmplat: irq.o
  CC      libkvmplat: lcpu.common.o
  LD       libkvmplat.ld.o
  OBJCOPY  libkvmplat.o
  CC      libkvmpci: pci_bus.common.o
```

andreittr:
> It's still a problem for commands without a library specifier:

Should be fixed now by judicious application of `$(strip ...)`.

jschlumpp (approved):

StefanJum (approved):
All good, thanks.

support/build/Makefile.rules
skuenzer: Yeah, I think `LIBNAME__TREEMODE` is also a good idea :+1:, makes sense. So far we use `_` prefix to hide internal Make functions and until recently there was an internal variable: `_IMPORT_BASE` but none of them were local to a library. We can consider this as a new case.

support/build/Makefile.rules
```diff
@@ -82,18 +82,19 @@ endef
 #
 ################################################################################
 
+# Returns the fully-qualified path for a subfolder under the build output
+# directory
+sub_build_dir = $(addprefix $(BUILD_DIR)/,$(if $($(call uc,$(1))_BUILDTREE),\
+	$(1)/build,$(1)))
```
skuenzer: can we also keep the `notdir` operation for non `__BUILDTREE` files?

support/build/Makefile.rules
```diff
@@ -82,18 +82,19 @@ endef
...
+
 # Creates a subfolder under the build output directory
 # mk_sub_build_dir $1:subfoldername
 define mk_sub_build_dir =
-$(if $(shell mkdir -p "$(BUILD_DIR)/$(notdir $(1))" && cd "$(BUILD_DIR)/$(notdir $(1))" >/dev/null && pwd),,\
-     $(error could not create directory "$(BUILD_DIR)/$(notdir $(1))"))
+$(eval _d := $(call sub_build_dir,$(1)))\
+$(if $(shell mkdir -p "$(_d)" && cd "$(_d)" >/dev/null && pwd),,\
```
skuenzer: I tried to use $(MKDIR) (which already exists) for any invocation of `mkdir`. Maybe we should try to do the same, like here: [c00264def](https://github.com/unikraft/unikraft/pull/1034/commits/c00264def97a8c1d04a0211aa215949f0779a01d)

andreittr:
Rebased & addressed comments.

support/build/Makefile.rules
```diff
@@ -300,26 +324,28 @@ endif
 # Writes to a file during deferred expansion
 #
 # write_file_deferred $1:filepath,$2:content
-write_file_deferred = $(if $(UK_DEFERRED_EXPANSION),$(file > $(1),$(2)),$$(file > $(1),$(2)))
+write_file_deferred = $(if $(UK_DEFERRED_EXPANSION),\
+	$(shell mkdir -p $(dir $(1)))$(file > $(1),$(2)),\
+	$$(shell mkdir -p $(dir $(1)))$$(file > $(1),$(2)))
```
skuenzer: Can we use `$(MKDIR)` here, too?

support/build/Makefile.rules
```diff
@@ -269,15 +270,16 @@ define addplatlib_s =
 $(if $(filter y,$(3)),$(call addplatlib,$(1),$(2)),)
 endef
 
+sub_libbuild_dir = $(call sub_build_dir,$(1))/$(2)
```
skuenzer: For safety reasons, I'd prefer to keep the `strip` operation on the second parameter
```
sub_libbuild_dir = $(call sub_build_dir,$(1))/$(strip $(2))
```

andreittr: You mean `strip` or `notdir`? (latter is in line with `sub_build_dir`)

skuenzer (approved):
Thanks a lot for the work!

PR #1005, opened 2023-07-26
Makefile: Allow external Makefile.build and Provide GO 1.18 support

eduardvintila:
### Prerequisite checklist

### Base target

### Description of changes

This PR is part of a series of PRs that work in conjunction for the update of Go support to 1.18:

- https://github.com/unikraft/lib-libgo/pull/7
- https://github.com/unikraft/unikraft/pull/1005
- https://github.com/unikraft/lib-musl/pull/67
- https://github.com/unikraft/app-helloworld-go/pull/9
- https://github.com/unikraft/lib-compiler-rt/pull/16
- https://github.com/unikraft/lib-musl/pull/52

For testing, make sure to pass to QEMU the `-cpu host` argument (this is necessary because the paging API and ukvmem are required by `libgo` as part of Virtual Address space management, and somehow the default CPU just won't handle 1GB pages).

Only GCC12 and x86 is supported for the moment.

### DIFF
--- a/lib/posix-mmap/Config.uk
+++ b/lib/posix-mmap/Config.uk
@@ -2,7 +2,6 @@ menuconfig LIBPOSIX_MMAP
 	bool "posix-mmap: Memory mapping related functions"
 	default n
 	depends on LIBUKVMEM
-	depends on !LIBUKMMAP
 	help
 		This library implements memory mapping related functions such
 		as mmap() and munmap().
--- a/Makefile
+++ b/Makefile
@@ -582,7 +582,7 @@ LD		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CC		:= $(CONFIG_CROSS_COMPILE)$(CONFIG_COMPILER)
 CPP		:= $(CC)
 CXX		:= $(CPP)
-GOC		:= $(CONFIG_CROSS_COMPILE)gccgo-7
+GOC		:= $(CONFIG_CROSS_COMPILE)gccgo
 # We use rustc because the gcc frontend is experimental and missing features such
 # as borrowing checking
 ifneq ("$(origin LLVM_TARGET_ARCH)","undefined")
--- a/lib/ukmmap/Config.uk
+++ b/lib/ukmmap/Config.uk
@@ -3,6 +3,7 @@ menuconfig LIBUKMMAP
 	default n
 	select LIBNOLIBC if !HAVE_LIBC
 	select LIBUKALLOC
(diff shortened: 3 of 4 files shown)


lib/posix-mmap/Config.uk
```diff
@@ -2,7 +2,6 @@ menuconfig LIBPOSIX_MMAP
 	bool "posix-mmap: Memory mapping related functions"
 	default n
 	depends on LIBUKVMEM
-	depends on !LIBUKMMAP
```
StefanJum, line 5: Why change this to be the other way around? (have `ukmmap` depend on `!LIBPOSIX_MMAP` instead of `posix-mmap` depend on `!LIBUKMMAP`)?

StefanJum (changes requested):
This looks fine. Besides the one comment, please rebase this on staging so the ci/cd tests can run.

lib/posix-mmap/Config.uk
eduardvintila: @StefanJum, the reason is that most of the external libraries don't have mentions of `LIBPOSIX_MMAP`, but rather `LIBUKMMAP`. 

For example, without this commit, if you first select `compiler-rt` in the menuconfig, `ukmmap` gets selected because the library's `Config.uk` file includes the statement `imply UKMMAP`. If you afterwards select `libgo`, which has the `select LIBPOSIX_MMAP` requirement, it would indeed select `posix-mmap`, but for some reason it also keeps `ukmmap` selected as well, which leads to a double definition conflict.

The dependency switch that I've introduced prevents that from happening, but I do not have a perfect explanation for this behavior, unfortunately.

StefanJum (approved):
All good, thanks.

RaduNichita (approved):
All good from my side. Tested with gcc12 on x86_64 and it works.

lib/posix-mmap/Config.uk
skuenzer: Hum... one additional thing to do in order to avoid the problem is by patching `compiler_rt` with
`imply LIB_UKMMAP if !LIBPOSIX_MMAP`.
Additionally, we probably should add a warning or error with `support/scripts/configupdate` as soon as ukmmap and posix_mmap are both. I think this would be a reasonable workaround as long as `ukmmap` exists. Longer term the plan was to migrate the left functionality from `ukmmap` to `posix-mmap` anyways.

Overall, I will approve this PR and will open an issue for updating the `configupdate` script.

razvand (approved):

PR #1010, opened 2023-07-31
plat/{linuxu,xen}: Do not enable interrupts before halt

kubanrob:
### Prerequisite checklist

### Base target

 - Platform(s): `xen`, `linuxu`
 - Application(s): N/A

### Additional configuration

None.

### Description of changes

Based on #980. While working on #980, I noticed both `linuxu` and `xen` also enable the interrupts before the halt, which will most likely lead to the same problem as on arm64.

### DIFF
--- a/plat/linuxu/irq.c
+++ b/plat/linuxu/irq.c
@@ -93,9 +93,7 @@ void ukplat_lcpu_halt_irq(void)
 {
 	UK_ASSERT(ukplat_lcpu_irqs_disabled());
 
-	ukplat_lcpu_enable_irq();
 	halt();
-	ukplat_lcpu_disable_irq();
 }
 
 int ukplat_lcpu_irqs_disabled(void)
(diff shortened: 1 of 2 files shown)


jschlumpp:
I think these changes might still make sense. For linuxu I don't think it makes much of a difference and for Xen I can sadly not test it.

kubanrob:
@mschlumpp I am happy to rebase this commit / clean up for merging if needed.

However, I don't have a setup to test it for either Linuxu or Xen. So feedback from people familiar with these platforms would be appreciated.

jschlumpp:
I have the same problem with Xen. 
For linuxu they interrupt functionality is mapped to signals, but I can't see anything where they are used. So I don't know how one could test it there.

kubanrob:
Since apart from the testing / review there is no reason this should not be merged, I removed the draft status, as discussed with @razvand.

andreistan26 (commented):
I don't think this will work for `x86`, the `xen` platform shares the same lcpu code for both `arm` and `x86`. Making it architecture dependent and only modifying the `arm` variant could work.

Issue #665, opened 2022-11-28
`lib/ukprint`: Configurable log format and printing


nderjung:
### Feature request summary

This issue tracks the addition of a new library which aims to abstract the methods necessary for printing to the console (i.e. via `outk` and `outc`).  Currently, this mechanism is embedded inside of `ukdebug`. There are two reasons for wanting to abstract this functionality into its own library:

1. The methods exposed by `ukdebug` prefix messages with debug preamble.  Whilst useful, this is not necessary in all cases. These message prefix options should be customizable through a bitwise flag, such that printing these additional pieces of information are customizable, i.e. wanting to print only the timestamp prefix but no other.
2. By uitilising the print methods from `ukdebug`, this library now becomes a needless dependency only to facilitate outgoing information which is out of the scope of the library's purpose. Idealising the notion of "one library one purpose" sees this functionality already as already out-of-scope for `ukdebug` itself.
3. To enrich the `ukprint` library with alternative printing formats. Right now there is a simple plaintext (or ANSI color variant) of a "log-line" from Unikraft. To better facilitate alternative use cases, preferences or machine readability, `ukprint` should offer alternative formatting options, for example a JSON output.
4. Optionally, though additional discussion is necessary on this topic: it would be nice to be able to provide such format through a configuration option, e.g. `LIBUKDEBUG_LOG_FORMAT="{level} {time} {library}: {msg}"`.

### Additional context

This will resolve the inconsistent test output which was temporarily introduced in #662.

The library should expose a public `#include <uk/print.h>` header.

TomFinet:
I have time to work on this and have made a rough outline of what I understand needs doing:

1. move `snprintf`, `outf` and `print` files from ukdebug to a new ukprint lib.
2. update config variables (remove `CONFIG_LIBUKDEBUG_PRINT_TIME` etc, these are replaced by the mentioned flags).
3. add `int flags` parameter to `outf` and family of `_uk_print` functions and macros.
4. modify `_vprint` function to print the preamble according to the flags parameter.
5. update existing libs to use ukprint instead of ukdebug for generic printing.

Is this roughly what was held in mind? Wanted to check before I started.

skuenzer:
Hi @TomFinet, thanks a lot for your offer, this is great. Internally, we had discussion that let to a design draft already for this. I think best is probably to organize a kick-off call where I can guide you through. We can take it from there.

razvand:
[...]

PR #1012, opened 2023-08-01
.github/workflows: Introduce GitHub Actions Workflow integrations

nderjung:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

This commit introduces GitHub-centric CI/CD definition that is intended to be executed on incoming PRs such that the incoming PR can be determined as working.  This supplementary file is based on-top of the [KraftKit GitHub Action Workflow Step][0] which is a native-to-Unikraft integration.  This integration allows for seamless building, packaging, distributing and executing (via TCG-based emulation) of Unikraft unikernels in a clear and declarative manner.

In this first integration pipeline, incoming PRs will be built against the following variant of the helloworld application:

- `qemu/x86_64`
- `qemu/arm64`
- `xen/x86_64`
- `fc/x86_64`
- `linuxu/x86_64`
- `linuxu/arm64`

Additionally, a more comprensive test will be executed with [libc-test][1] via uktest in emulation mode for qemu on `x86_64` only in order to display maximum coverage.

[0]: https://github.com/unikraft/kraftkit/blob/staging/action.yml
[1]: https://github.com/unikraft/lib-libc-test

Note: GitHub will not execute this action until it is merged into the target branch. A demonstration of this pipeline can be viewed here: https://github.com/nderjung/unikraft/actions/runs/5728684586 Therefore, to reproduce this work, simply apply this PR onto your fork on branch `staging` to see the action perform.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -1,0 +1,186 @@
+name: integration
+
+on:
+  push:
+    branches: [staging, stable]
+
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches: [staging]
+
+jobs:
+  libc-test:
+    name: libc-test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - plat: qemu
+          arch: x86_64
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+
+    - name: Fetch helloworld
+      uses: actions/checkout@v3
+      with:
+        repository: unikraft/app-helloworld
+        fetch-depth: 1
+        path: _helloworld
+
+    - name: Build and run libc-test via uktest
+      uses: unikraft/kraftkit@staging
+      with:
+        loglevel: debug
+        workdir: _helloworld
+        plat: ${{ matrix.plat }}
+        arch: ${{ matrix.arch }}
+        execute: true
+        kraftfile: |
+          specification: '0.5'
+          name: helloworld
+          unikraft:
+            source: ../
+            kconfig:
+              CONFIG_LIBUKTEST: "y"
+              CONFIG_LIBUKTEST_ALL: "n"
+              CONFIG_LIBUKTEST_TEST_MYSELF: "n"
+              CONFIG_LIBUKDEBUG_ANSI_COLOR: "y"
+              CONFIG_LIBDEVFS: "y"
+              CONFIG_LIBDEVFS_DEV_STDOUT: "y"
+              CONFIG_LIBDEVFS_DEV_NULL_ZERO: "y"
+              CONFIG_LIBDEVFS_DEV_ZERO: "y"
+              CONFIG_STACK_SIZE_PAGE_ORDER: 4
+          libraries:
+            libc-test:
+              version: staging
+              kconfig:
+                CONFIG_LIBLIBCTEST: "y"
+                CONFIG_LIBLIBCTEST_STRING_TESTS: "y"
+                CONFIG_LIBLIBCTEST_PTHREAD_TESTS: "y"
+                CONFIG_LIBLIBCTEST_CONVERSION_TESTS: "y"
+                CONFIG_LIBLIBCTEST_REGEX_TESTS: "y"
+                CONFIG_LIBLIBCTEST_FILE_FOLDER_TESTS: "y"
+                CONFIG_LIBLIBCTEST_TIME_TESTS: "y"
+                CONFIG_LIBLIBCTEST_NETWORK_TESTS: "y"
+                CONFIG_LIBLIBCTEST_SORT_TESTS: "y"
+                CONFIG_LIBLIBCTEST_STRUCTURE_SEARCH_TESTS: "y"
+                CONFIG_LIBLIBCTEST_SEMAPHORE_TESTS: "y"
+                CONFIG_LIBLIBCTEST_RANDOM_TESTS: "y"
+                CONFIG_LIBLIBCTEST_CRYPT_TESTS: "y"
+                CONFIG_LIBLIBCTEST_ENV_TESTS: "y"
+                CONFIG_LIBLIBCTEST_MALLOC_TESTS: "y"
+                CONFIG_LIBLIBCTEST_TGMATH_TESTS: "y"
+                CONFIG_LIBLIBCTEST_ICONV_TESTS: "y"
+                CONFIG_LIBLIBCTEST_UDIV_TESTS: "y"
+                CONFIG_LIBLIBCTEST_MBFUNC_TESTS: "y"
+                CONFIG_LIBLIBCTEST_SETJMP_TESTS: "y"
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+    - name: Fetch helloworld
+      uses: actions/checkout@v3
+      with:
+        repository: unikraft/app-helloworld
+        fetch-depth: 1
+        path: _helloworld
+
+    - name: Build and run libc-test via uktest
+      uses: unikraft/kraftkit@staging
```
craciunoiuc, line 35:
```suggestion
      uses: unikraft/kraftkit@v0
```

Why not like this and we do a tag?

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+    - name: Fetch helloworld
+      uses: actions/checkout@v3
+      with:
+        repository: unikraft/app-helloworld
+        fetch-depth: 1
+        path: _helloworld
+
+    - name: Build helloworld
+      uses: unikraft/kraftkit@staging
```
craciunoiuc, line 172: Same as above

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,139 @@
...
+        submodules: recursive
+
+    - name: Fetch helloworld
+      uses: actions/checkout@v3
+      with:
+        repository: unikraft/app-helloworld
+        fetch-depth: 1
+        path: _helloworld
+
+    - name: Build helloworld
```
craciunoiuc, line 171: Why build both? Shouldn't this be run inside the app catalog by sending events somehow?

Does it mean that eventually we will also have all apps we want to check against, here?

craciunoiuc (commented):
Looks good, some questions from my side

.github/workflows/integration.yaml
nderjung: Because this represents a branch or tag, there is no branch `v0` or tag `v0` on the kraftkit repo.

.github/workflows/integration.yaml
nderjung: Of course the catalog repo will run it as well. But this will be run once it is in already in `staging`. Here, we are checking the incoming PR builds against helloworld, on-demand.

craciunoiuc: Ok, so a double check, fine

.github/workflows/integration.yaml
craciunoiuc: exactly, we could do it if you want ;)

A v0 branch wouldn't hurt, certainly easier to update than a tag

Should we do it?

nderjung: I fear a `v0` branch is extra maintenance on our side. Once we do a proper release in kraftkit, we can put `unikraft/kraftkit@stable`

i-Pear:
I see there's "Unhandled Trap 13 (general protection), error code=0x0" in the demo, when running `lib-libc-test`. I also encountered this while testing my PR.

To solve this, you can try setting `LIBUKTEST_ALL` to `n` and `CONFIG_LIBUKTEST_TEST_MYSELF` to `n`. It seems the self-tests have some side effects, which can let `lib-libc-test` to crash.

.github/workflows/integration.yaml
craciunoiuc: Okay 👍

i-Pear:
For greater flexibility, each job can also be individually split into separate repositories/branches, making the `unikraft` repo more concise. Initially, `unikraft/test-workflows` was intended for this purpose, and you can be free to write into it as needed.

i-Pear:
It may also be possible to save the test output into a text file and display the test results in a separate step. The extensive logs from `git clone` and `make` might make it difficult for users to locate the actual test results.

nderjung:
Hi @i-Pear

> I see there's "Unhandled Trap 13 (general protection), error code=0x0" in the demo, when running `lib-libc-test`. I also encountered this while testing my PR.
>
> To solve this, you can try setting `LIBUKTEST_ALL` to `n` and `CONFIG_LIBUKTEST_TEST_MYSELF` to `n`. It seems the self-tests have some side effects, which can let `lib-libc-test` to crash.

Thanks for the suggestion. For this, I have force-pushed an update that now separates libc-test and "self-test" into two separate jobs. This can be viewed here:

https://github.com/nderjung/unikraft/actions/runs/5739682373

> For greater flexibility, each job can also be individually split into separate repositories/branches, making the `unikraft` repo more concise. Initially, `unikraft/test-workflows` was intended for this purpose, and you can be free to write into it as needed.

Separating jobs into different branches is definitely more overhead in terms of maintenance in my opinion. Instead this PR has introduced the use of a generic GitHub action tool that is also adoptable outside of testing Unikraft's core internally by anyone who simply wishes to build unikernels programmatically in a CI/CD context via GitHub Actions. Since KraftKit is the SDK behind this action, as we maintain this project we will always ensure that the action works [since we test it continuously internally](https://github.com/unikraft/kraftkit/actions/workflows/github-action.yaml) (see [here another example](https://github.com/unikraft/kraftkit/actions/runs/5727503369/job/15520734987#step:10:1)).

The `Kraftfile` has also been included in-line and represents a working expression of building Unikraft against helloworld. By keeping this declaration in-tree, modifications to Unikraft's core which affect this expression of KConfig options will be caught before they are made to `staging`. Additionally, new platforms, architectures, etc. introduced into the core must modify this file allowing the PR which introduces new functionality at this level to be tested within the scope of the same PR.

> It may also be possible to save the test output into a text file and display the test results in a separate step. The extensive logs from `git clone` and `make` might make it difficult for users to locate the actual test results.

One step at a time. We can parse the output later on as a quality-of-life improvement. Right now it's important just to get a build in and it is still viewable through GitHub's web UI. We can add additional wrappers ([see composites](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)) to make markdown summaries readable in the PR checks tab. I should note this is definitely planned, but we are waiting for [`ukdebug` to be refactored into two libraries, introducing a configurable `ukprint` library capable of outputting JSON-formatted logs](https://github.com/unikraft/unikraft/issues/665) which will allow for easier parsing of the results from `uktest`.

craciunoiuc:
If needed, we could potentially limit the output to warnings, this should silence all the compilation messages

what do you think? @nderjung

nderjung:
I would not remove the compilation steps from the CI at this stage as it is important to view what files are compiled and when warnings are emitted since [silencing warnings is an open issue and is often addressed in PRs](https://github.com/unikraft/unikraft/pulls?q=is%3Apr+silence+warning) and warning themselves need to be addressed individually.

I think the parsing of the compilation logs and execution logs should come later when we do these quality-of-life improvements. This can be addressed in the github-action itself and as PR in KraftKit.

craciunoiuc (approved):
Ok, let's get this in and let's see if you're lucky @nderjung 

razvand (approved):
Thanks, @i-Pear, for the comments. Let's merge this as it is now, and then we can improve it later.

PR #1022, opened 2023-08-07
plat/common: Add support for pvpanic devices for x86

craciunoiuc:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Additional configuration

Use `-device pvpanic` when testing with `qemu-system-x86_64`.

### Description of changes

Unikraft already offers exit codes through the ISA device, but these
are only a convention and are not understood by QEMU itself.

When using `-device pvpanic`, a GUEST_PANICKED event[1] is sent to QEMU
when the unikernel crashes.
This event can be seen when using QMP[1] for example.
The bits corresponding to each even are set as per the documentation[2].

The main usage of this is in kraftkit where unikernels are detached
and exit codes are lost.
Using this can signal when the unikernel crashes.

[1] https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#qapidoc-152
[2] https://github.com/qemu/qemu/blob/master/docs/specs/pvpanic.txt

### DIFF
--- a/plat/common/x86/cpu_native.c
+++ b/plat/common/x86/cpu_native.c
@@ -55,6 +55,13 @@ unsigned long read_cr2(void)
 /* This corresponds to an 85 (42 << 1 | 1) return value from QEMU */
 #define QEMU_ISA_DEBUG_EXIT_CRASH	42
 
+/* The port used by QEMU by default for the pvpanic device */
+#define QEMU_PVPANIC_EXIT_PORT		0x505
+/* This corresponds to a GUEST_PANICKED event for QEMU */
+#define QEMU_PVPANIC_GUEST_PANICKED	(1 << 0)
+/* This corresponds to a GUEST_CRASHLOADED event for QEMU */
+#define QEMU_PVPANIC_GUEST_CRASHLOADED	(1 << 1)
+
 /**
  * Trigger an exit() in QEMU with the code `value << 1 | 1`.
  * @param value the value used in the calculation of the exit code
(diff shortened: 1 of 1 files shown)


plat/common/x86/cpu_native.c
```diff
@@ -55,13 +55,21 @@ unsigned long read_cr2(void)
...
+#define QEMU_PVPANIC_GUEST_CRASHLOADED	2
+
 /**
  * Trigger an exit() in QEMU with the code `value << 1 | 1`.
  * @param value the value used in the calculation of the exit code
  */
 static void qemu_debug_exit(int value)
 {
 	outw(QEMU_ISA_DEBUG_EXIT_PORT, value);
+	outw(QEMU_PVPANIC_EXIT_PORT, QEMU_PVPANIC_GUEST_PANICKED);
```
jschlumpp: That breaks the "normal" exit code (see line 109). And it does not fit into this function, because this function is for the isa-debug-exit device.

jschlumpp (changes requested):

plat/common/x86/cpu_native.c
craciunoiuc: Sure, I can just create another function

craciunoiuc: actually, it's a single place where that is used, I just moved it down

Also changed to `outb` cause it's a single byte

plat/common/x86/cpu_native.c
```diff
@@ -55,6 +55,13 @@ unsigned long read_cr2(void)
 /* This corresponds to an 85 (42 << 1 | 1) return value from QEMU */
 #define QEMU_ISA_DEBUG_EXIT_CRASH	42
 
+/* The port used by QEMU by default for the pvpanic device */
+#define QEMU_PVPANIC_EXIT_PORT		0x505
+/* This corresponds to a GUEST_PANICKED (1 << 0) event for QEMU */
+#define QEMU_PVPANIC_GUEST_PANICKED	1
+/* This corresponds to a GUEST_CRASHLOADED (1 << 1) event for QEMU */
```
jschlumpp: Is it defined as `1 << 1`? Then why not use that definition directly?

jschlumpp (changes requested):

plat/common/x86/cpu_native.c
jschlumpp: Also would be nice to have a reference to the QEMU documentation of the pvpanic device in the comments.

craciunoiuc: I mean, why not, it says 1st bit/2nd bit

jschlumpp (approved):

razvand (approved):

PR #1023, opened 2023-08-07
Introduce drivers subsystem

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This commit adds a drivers subsystem to Unikraft. Drivers are placed under the top-level `drivers/` directory.

In `menuconfig` drivers appear at a top-level section called "Device Drivers".

**Limitations:** No external drivers support. That is due to additional changes required to support the `<subsystem>/<driver>` hierarchy.


michpappas:
Changes in this update:
- Rebase to `RELEASE-0.14`
- Update build system to use `config-submenu.sh`
- Update all driver libraries to register using `addlib_s`
- Update the top-level `Config.uk` of every subsystem to provide a Kconfig `menu`  instead of a `menuconfig`
- Update the top-level `Config.uk` of every subsystem to source the configuration of each one if its drivers. This is required because the `-r` parameter of `config-submenu.sh` will recurse down to only one level.
- Add `exportsyms.uk` to all driver libraries

michpappas:
Changes in this update:
- Rebase to `RELEASE-0.14`
- Use `config-submenu.sh`

Makefile
```diff
@@ -474,6 +479,7 @@ include $(CONFIG_UK_BASE)/support/build/Makefile.rules
 $(foreach _M,$(wildcard $(addsuffix Makefile.rules,\
 	   $(CONFIG_UK_BASE)/arch/ $(CONFIG_UK_BASE)/arch/*/ \
 	   $(CONFIG_UK_BASE)/plat/*/ $(CONFIG_UK_BASE)/lib/*/ \
+	   $(CONFIG_UK_BASE)/drivers/*/ \
```
michpappas: TBD: Right now this allows `Makefile.rules` per sub-system (eg `drivers/ukbus/Makefile.rules`). The alternatives I see are:
* Change it to `drivers/*/*`. Allows drivers to define their own `Makefile.rules`, but makes the hierarchy of `drivers/<subsystem>/<driver>` a hard requirement.
* Implement more advanced logic that recursively search for `Makefile.rules` inside driver (not in the scope of this PR).
* Use both `drivers/*` and `drivers/*/*`. Less elegant, yet more flexible. Still somewhat limited to two-level hierarchies.

Config.uk
```diff
@@ -37,6 +37,10 @@ menu "Platform Configuration"
 	source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_PLAT_IN)' -r '$(KCONFIG_PLAT_BASE)' -l '$(KCONFIG_PLAT_BASE)' -c '$(KCONFIG_EPLAT_DIRS)')"
 endmenu
 
+menu "Device Drivers"
+	source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DRIV_IN)' -r '$(KCONFIG_DRIV_BASE)' -l '$(KCONFIG_DRIV_BASE)' -c '$(KCONFIG_EDRIV_DIRS)')"
```
michpappas: Notice that the `-r` option of `config-submenu.sh` does a single-level lookup instead of being fully recursive:
```bash
	r)
		if [ ! -z "${OPTARG}" ]; then
			for P in "${OPTARG}/"*; do
				[ -d "${P}" ] && ARG_PATHS+=("${P}")
			done
		fi
		;;
```
That requires that the driver subsystems (`drivers/ukbus/`) source the `Config.uk` of their individual drivers (`drivers/ukbus/pci/`). IMO this is okay as it allows wrapping everything together under a Kconfig `menu`:
```
menu "Bus drivers"

source "$(KCONFIG_DRIV_BASE)/ukbus/pci/Config.uk"
# more bus drivers...

endmenu
```
It is also aligned with the similar requirement in the subsystem's `Makefile` that also imports the individual drivers:
```
#
# Driver registrations
#

UK_DRIVERS_BUS_BASE := $(UK_DRIVERS_BASE)/ukbus

$(eval $(call _import_lib,$(UK_DRIVERS_BUS_BASE)/pci))
```

support/build/config-submenu.sh
```diff
@@ -79,25 +79,25 @@ if [ ! -z "$ARG_OUT" ]; then
 	fi
 	[ -f "$ARG_OUT" ] && rm -f "$ARG_OUT"
 
-	# open fd:7
-	exec 7<>"$ARG_OUT"
+	# create and open a new file descriptor
+	exec {fd}<>"$ARG_OUT"
```
skuenzer: Can we use capital letters for the `FD`, this would be inline with the rest of the style in this file. Otherwise, this is a great fix!

Makefile
```diff
@@ -323,6 +326,8 @@ UK_ALIBS:=
 UK_ALIBS-y:=
 UK_OLIBS:=
 UK_OLIBS-y:=
+UK_DRIV:=
+UK_DRIV-y:=
```
skuenzer: I don't think that we need a `UK_DRIV` list in the build system. Drivers would use the common library registration methods and drivers will appear under `UK_OLIBS`

Makefile
```diff
@@ -708,6 +714,8 @@ $(foreach E,$(ELIB_DIR), \
 $(eval $(call _import_lib,$(CONFIG_UK_BASE)/arch/$(UK_FAMILY)))
 # internal platform libraries
 $(eval $(call verbose_include,$(CONFIG_UK_BASE)/plat/Makefile.uk))
+# drivers
+$(eval $(call verbose_include,$(CONFIG_UK_BASE)/drivers/Makefile.uk))
```
skuenzer, line 710: Maybe put this before platforms because afterwards we are including external platforms and I think the platforms should stay together.

Config.uk
skuenzer: Hum, Iwas thinking that we could also use `config-submenu.sh` for these files, problem is that we need a unique a filename for each driver group, where the script can write the generated KConfig code to. 🤔 But maybe we can do this, it would be then inline with how platforms and libraries handles this automatically. An option could be to use `$(KCONFIG_DIR)` as basis to generate the filename in-place:
```
menu "Bus drivers"

source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-bus.uk' -r '$(KCONFIG_DRIV_BASE)/ukbus' -l '$(KCONFIG_DRIV_BASE)/ukbus')"

endmenu
```

skuenzer: Because I see it in your example: I think we should probably make the name of `$(UK_DRIVERS_BASE)` inline with the other variable names in the build system, meaning `$(UK_DRIV_BASE)`.

Makefile
skuenzer: Hum... maybe lets start with the third option and we could still improve with the 2nd option

Makefile
michpappas: Ooops, I guess it was misplaced there

michpappas:
@skuenzer I addressed all the review comments. Besides these, I also did:
* Simplify the fix for the stdout file descriptor redirection in `config-submenu.sh`, as discussed.
* Remove the `-c '${KCONFIG_EDRIV_DIRS}'` parameter from `config-submenu.sh` in `Config.uk`, as we don't support external drivers yet.
* Add `KCONFIG_DIR` to the environmental variables. This is required for things like `$(KCONFIG_DIR)/drivers-bus.uk` in `drivers/Config.uk`.

skuenzer (approved):

PR #229, opened 2021-06-14
ukalloc: Allocator Statistics

skuenzer:
This patch series introduces allocator statistics. With minimal instrumentation, allocators can report counters on allocation and freeing events that happened. For example, the following could be queried per allocator:

```c
 last_alloc_size:         230 B /* size of the last allocation */
 max_alloc_size:         4096 B /* biggest satisfied allocation size */
 min_alloc_size:           12 B /* smallest satisfied allocation size */
 tot_nb_allocs:            75   /* total number of satisfied allocations */
 tot_nb_frees:             22   /* total number of satisfied free operations */
 cur_nb_allocs:            53   /* current number of active allocations */
 max_nb_allocs:            53   /* maximum number of active allocations */
 max_mem_use:          218023 B /* maximum amount of memory that was in use */
 cur_mem_use:          217088 B /* current amount of memory that is in use */
 nb_enomem:                 0   /* number of times failing allocation requests */
```

Note: The meaning of `CONFIG_IFSTAT` changes with this patch series. The configuration option originally raised wrong expectations while it it just provided an API to query the amount of available memory from an allocator. `CONFIG_IFSTAT` is now enabling these detailed allocator statistics.

Note: Since this patch series changes some details on the internal ukalloc API (`<uk/alloc_impl.h>`), external memory allocators (like TLSF, mimalloc) need to be adopted as well. For retrieving statistics from them, they also need to be instrumented to count alloc and free events.

The first four commits are doing minor corrections of the API headers. Afterwards, the statistics API is introduced. Additionally, the last commits introduce a one consolidated global allocator statistics, as well as a per library allocator statistics. `lib/nolibc` had to instrumented in order to support per-library statistics properly. Similar to `nolibc`, `newlib` and `musl` need to be instrumented, too to properly support per-library statistics for libc allocation interfaces: `malloc()`, `calloc()`, `free()`, etc.

This series is the successor of the series [1574](https://patchwork.unikraft.org/project/unikraft/list/?series=1574&state=*) and [1607](https://patchwork.unikraft.org/project/unikraft/list/?series=1607&state=*) on patchwork.

PR #1042, opened 2023-08-12
lib/nolibc: Provide symbols for `malloc`, `free`, `calloc`, `realloc`...

Starnox:
To ensure that symbols such as malloc, calloc, and free are accessible during link time for internal gcc objects, such as those related to gcov, these functions must be designated as public.

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

- CONFIG_LIBUKALLOC=y

### Description of changes

Remove `static inline` modifiers from memory management functions in `nolibc`

### DIFF
--- a/lib/nolibc/include/stdlib.h
+++ b/lib/nolibc/include/stdlib.h
@@ -63,58 +63,15 @@ int atoi(const char *s);
 long atol(const char *s);
 
 #if CONFIG_LIBUKALLOC
-/* Allocate size bytes of memory. Returns pointer to start of allocated memory,
- * or NULL on failure.
- */
-static inline void *malloc(size_t size)
-{
-	return uk_malloc(uk_alloc_get_default(), size);
-}
 
-/* Release memory previously allocated by malloc(). ptr must be a pointer
- * previously returned by malloc(), otherwise undefined behavior will occur.
- */
-static inline void free(void *ptr)
-{
-	return uk_free(uk_alloc_get_default(), ptr);
-}
+void *malloc(size_t size);
+void free(void *ptr);
+void *calloc(size_t nmemb, size_t size);
+void *realloc(void *ptr, size_t size);
+int posix_memalign(void **memptr, size_t align, size_t size);
+void *memalign(size_t align, size_t size);
 
-/* Allocate memory for an array of nmemb elements of size bytes. Returns
- * pointer to start of allocated memory, or NULL on failure.
- */
-static inline void *calloc(size_t nmemb, size_t size)
-{
-	return uk_calloc(uk_alloc_get_default(), nmemb, size);
-}
-
-/* Change the size of the memory block pointed to by ptr to size bytes.
- * Returns a pointer to the resized memory area. If the area pointed to was
- * moved, a free(ptr) is done.
- */
-static inline void *realloc(void *ptr, size_t size)
-{
-	return uk_realloc(uk_alloc_get_default(), ptr, size);
-}
-
-/* Allocate size bytes of memory, aligned to align bytes, and return the
- * pointer to that area in *memptr. Returns 0 on success, and a non-zero error
- * value on failure.
- */
-static inline int posix_memalign(void **memptr, size_t align, size_t size)
-{
-	return uk_posix_memalign(uk_alloc_get_default(),
-				 memptr, align, size);
-}
-
-/* Allocate size bytes of memory, aligned to align bytes. Returns pointer to
- * start of allocated memory, or NULL on failure.
- */
-static inline void *memalign(size_t align, size_t size)
-{
-	return uk_memalign(uk_alloc_get_default(), align, size);
-}
 #endif /* CONFIG_LIBUKALLOC */
-
 void abort(void) __noreturn;
 
 void exit(int status) __noreturn;
--- a/lib/nolibc/stdlib.c
+++ b/lib/nolibc/stdlib.c
@@ -61,6 +61,60 @@
 
 #define __DECONST(type, var) ((type)(uintptr_t)(const void *)(var))
 
+
+#if CONFIG_LIBUKALLOC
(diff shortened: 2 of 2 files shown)


lib/nolibc/include/stdlib.h
```diff
@@ -61,60 +61,16 @@ unsigned long long strtoull(const char *nptr, char **endptr, int base);
  */
 int atoi(const char *s);
 long atol(const char *s);
-
```
skuenzer, line 64: You can keep this line.

lib/nolibc/include/stdlib.h
```diff
@@ -61,60 +61,16 @@ unsigned long long strtoull(const char *nptr, char **endptr, int base);
...
-}
-#endif /* CONFIG_LIBUKALLOC */
+void *malloc(size_t size);
+void free(void *ptr);
+void *calloc(size_t nmemb, size_t size);
+void *realloc(void *ptr, size_t size);
+int posix_memalign(void **memptr, size_t align, size_t size);
+void *memalign(size_t align, size_t size);
 
+#endif
```
skuenzer: Please use
```c
#endif /* CONFIG_LIBUKALLOC */
```

skuenzer (changes requested):
I think the title is mis-leading because no signature is changed but symbols are provided. How about:
`lib/nolibc: Provide symbols for 'malloc', 'free', etc.`

lib/nolibc/include/stdlib.h
```diff
@@ -61,60 +61,16 @@ unsigned long long strtoull(const char *nptr, char **endptr, int base);
...
-
-/* Allocate size bytes of memory, aligned to align bytes. Returns pointer to
- * start of allocated memory, or NULL on failure.
- */
-static inline void *memalign(size_t align, size_t size)
-{
-	return uk_memalign(uk_alloc_get_default(), align, size);
-}
-#endif /* CONFIG_LIBUKALLOC */
+void *malloc(size_t size);
```
RaduNichita, line 67: I suggest that we keep the comments regarding functions in the header file, not in the source file. 

@skuenzer @razvand any opinions here?

RaduNichita (changes requested):

lib/nolibc/include/stdlib.h
Starnox: I was uncertain about the coding standard in this case, as I believed the existing comments provided more detailed descriptions of the functions' operations. This seemed different from the Doxygen style comments, which, from what I understand, are typically placed in the header files. I'll wait for the other to tune in and give me the green-light regarding the decision.

razvand: I agree with @Starnox' s point. There must be a Doxygen-style comment in the header file, that documents the exposed API of the function, i.e. information that is related to the signature of the function (return value, parameters, parameter types).

However, internal details on the implementation of the function, i.e. its body, go to the source code. These either go inside the function body itself (inline comments), or on top of the function body + signature. These comments go in the source code because they would detail internal information pertaining to the implementation (internal variables, other functions being called, design / implementation decisions) that don't make sense to be in the header file, that doesn't provide any glimpse into the implementation.

skuenzer:
Hi @Starnox, I just noticed two things:
- The inclusion of `uk/alloc.h` is probably not needed anymore if you move the static functions ouf of the [stdlib.h](https://github.com/unikraft/unikraft/blob/14a8c1f110c981d7a37d8b40cd3e1cee6cc56221/lib/nolibc/include/stdlib.h#L32) header. Correct?
- Probably we should document (e.g., commit message) that the per-library stats from PR #229 will loose the fine-grained stats when using `nolibc`'s `malloc`/`free`/... calls. With your changes, each allocation operation via the libc interface will be accounted to `nolibc` instead to the individual libraries calling `malloc`, `free`, etc. - this is fine, but it should be documented.

Starnox:
> Hi @Starnox, I just noticed two things:
> 
>     * The inclusion of `uk/alloc.h` is probably not needed anymore if you move the static functions ouf of the [stdlib.h](https://github.com/unikraft/unikraft/blob/14a8c1f110c981d7a37d8b40cd3e1cee6cc56221/lib/nolibc/include/stdlib.h#L32) header. Correct?
> 
>     * Probably we should document (e.g., commit message) that the per-library stats from PR [ukalloc: Allocator Statistics #229](https://github.com/unikraft/unikraft/pull/229) will loose the fine-grained stats when using `nolibc`'s `malloc`/`free`/... calls. With your changes, each allocation operation via the libc interface will be accounted to `nolibc` instead to the individual libraries calling `malloc`, `free`, etc. - this is fine, but it should be documented.

1. Including that we also include some symbols that the function use like `uk_malloc` ... Or am I misunderstanding something?
2. I added new info in the commit message

lib/nolibc/stdlib.c
```diff
@@ -61,6 +61,60 @@
...
+
+/* Allocate memory for an array of nmemb elements of size bytes. Returns
+ * pointer to start of allocated memory, or NULL on failure.
+ */
+void *calloc(size_t nmemb, size_t size)
+{
+	return uk_calloc(uk_alloc_get_default(), nmemb, size);
+}
+
+/* Change the size of the memory block pointed to by ptr to size bytes.
```
RaduNichita, line 90: I think we may specify that the allocated memory remains unchanged if the call fails

RaduNichita (changes requested):
One small comment from my side.

andraprs (approved):
Thank you for the updates.

RaduNichita (approved):
All good from my side. Thanks, @Starnox !

skuenzer (approved):
Thanks!

razvand (approved):

PR #1043, opened 2023-08-12
lib/nolibc: Introduce 'FILE' object APIs

Starnox:
If `vfscore` is enabled, then add a small set of minimally implemented file handling functions like `fopen`, `fclose`, `fwrite` that are useful at link time for `gcc` objects such as the ones from `gcov`

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

- CONFIG_LIBVFSCORE=y

### Description of changes

Useful and simple file operations functions using `libvfscore`

### DIFF
--- a/lib/nolibc/include/nolibc-internal/shareddefs.h
+++ b/lib/nolibc/include/nolibc-internal/shareddefs.h
@@ -88,6 +88,11 @@ typedef struct {
 #define __DEFINED_max_align_t
 #endif
 
+#if (defined __NEED_FILE && !defined __DEFINED_FILE)
+typedef struct _nolibc_file FILE;
+#define __DEFINED_FILE
+#endif
+
 #if defined(__NEED_useconds_t) && !defined(__DEFINED_useconds_t)
 typedef unsigned useconds_t;
 #define __DEFINED_useconds_t
--- a/lib/nolibc/include/stdio.h
+++ b/lib/nolibc/include/stdio.h
@@ -36,6 +36,11 @@
 
 #include <uk/essentials.h>
 
+
+#if CONFIG_LIBVFSCORE
+#include <vfscore/file.h>
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -43,10 +48,8 @@ extern "C" {
 #define __NEED_NULL
 #define __NEED_size_t
 #define __NEED_ssize_t
+#define __NEED_FILE
 #include <nolibc-internal/shareddefs.h>
-
-struct _nolibc_fd;
-typedef struct _nolibc_fd FILE;
 
 extern FILE *stdin;
 extern FILE *stdout;
@@ -92,8 +95,17 @@ int fputs(const char *restrict s, FILE *restrict stream);
 int puts(const char *s);
 
 #if CONFIG_LIBVFSCORE
+void clearerr(FILE *stream);
 int rename(const char *oldpath, const char *newpath);
-#endif
+int feof(FILE *stream);
+int ferror(FILE *stream);
+int fseek(FILE *stream, long offset, int whence);
+int fclose(FILE *stream);
+FILE *fopen(const char *pathname, const char *mode);
+FILE *fdopen(int fd, const char *mode);
+size_t fread(void *ptr, size_t size, size_t nmemb, FILE *stream);
+size_t fwrite(const void *ptr, size_t size, size_t nmemb, FILE *stream);
+#endif /* CONFIG_LIBVFSCORE */
 
 #ifdef __STDIO_H_DEFINED_va_list
 #undef va_list
--- a/lib/nolibc/stdio.c
+++ b/lib/nolibc/stdio.c
@@ -59,8 +59,14 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdarg.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 
 #include <uk/essentials.h>
+#include <uk/arch/lcpu.h>
 #include <uk/plat/console.h>
 
 /* 64 bits + 0-Byte at end */
@@ -542,3 +548,230 @@ int puts(const char *s)
 {
 	return fputs_internal(s, stdout, 1);
 }
+
+#if CONFIG_LIBVFSCORE
+struct _nolibc_file {
+	int fd;
+	int errno;
+	bool eof;
+	off_t offset;
+};
+
+void clearerr(FILE *stream)
+{
+	stream->eof = 0;
+	stream->errno = 0;
+}
+
+/* The following code is derived from musl libc */
+static int __fmodeflags(const char *mode, int *flags)
+{
+	if (!strchr("rwa", *mode)) {
+		errno = EINVAL;
+		return -1;
+	}
+	if (strchr(mode, '+'))
+		*flags = O_RDWR;
+	else if (*mode == 'r')
+		*flags = O_RDONLY;
+	else
+		*flags = O_WRONLY;
+	if (strchr(mode, 'x'))
+		*flags |= O_EXCL;
+	if (strchr(mode, 'e'))
+		*flags |= O_CLOEXEC;
+	if (*mode != 'r')
+		*flags |= O_CREAT;
+	if (*mode == 'w')
+		*flags |= O_TRUNC;
+	if (*mode == 'a')
+		*flags |= O_APPEND;
+
+	return 0;
+}
+
+FILE *fopen(const char *pathname, const char *mode)
+{
+	int flags;
+	int fd;
+	int ret;
+
+	ret = __fmodeflags(mode, &flags);
+
+	if (ret < 0)
+		return NULL;
+
+	fd = open(pathname, flags, 0666);
+
+	if (fd < 0)
+		return NULL;
+
+	return fdopen(fd, mode);
+}
+
+int feof(FILE *stream)
+{
+	return stream->eof;
+}
+
+int ferror(FILE *stream)
+{
+	return stream->errno;
+}
+
+int fclose(FILE *stream)
+{
+	int ret = close(stream->fd);
+
+	free(stream);
+	return ret;
+}
+
+FILE *fdopen(int fd, const char *mode __unused)
+{
+	FILE *f = (FILE *)malloc(sizeof(FILE));
+
+	if (!f)
+		return NULL;
+	f->fd = fd;
+	f->errno = 0;
+	f->eof = 0;
+	f->offset = 0;
+	return f;
+}
+
+int fseek(FILE *stream, long offset, int whence)
+{
+	off_t new_offset;
+
+	switch (whence) {
+	case SEEK_SET:
+	{
+		// Absolute offset from the beginning
+		if (unlikely(offset < 0)) {
+			stream->errno = 1;
+			return -1;
+		}
+		new_offset = offset;
+		break;
+	}
+
+	case SEEK_CUR:
+	{
+		// Offset from the current position
+		new_offset = stream->offset + offset;
+		if (unlikely(new_offset < 0)) {
+			stream->errno = 1;
+			return -1;
+		}
+		break;
+	}
+
+	case SEEK_END:
+	{
+		struct stat st;
+
+		if (unlikely(fstat(stream->fd, &st) != 0)) {
+			stream->errno = 1;
+			return -1;
+		}
+
+		new_offset = st.st_size + offset;
+		if (unlikely(new_offset < 0)) {
+			stream->errno = 1;
+			return -1;
+		}
+	}
+	break;
+
+	default:
+	{
+		// Invalid whence
+		stream->errno = 1;
+		return -1;
+	}
+	}
+
+	// Update the stream's offset
+	stream->offset = new_offset;
+	return 0;
+}
+
+size_t fread(void *ptr, size_t size, size_t nmemb, FILE *stream)
+{
+	if (unlikely(!stream))
+		return 0;
+
+	if (unlikely(!ptr || !size || !nmemb)) {
+		stream->errno = EINVAL;
+		return 0;
+	}
+
+	if (unlikely(stream->fd < 0)) {
+		stream->errno = EBADF;
+		return 0;
+	}
+
+	if (unlikely(SIZE_MAX / size < nmemb)) {
+		stream->errno = EOVERFLOW;
+		return 0;
+	}
+
+	size_t total = 0;
+
+	while (total < size * nmemb) {
+		ssize_t ret = pread(stream->fd,
+				((char *) ptr) + total,
+				size * nmemb - total, stream->offset);
+		if (ret > 0)
+			stream->offset += ret;  // Update the offset
+
+		if (ret < 0)
+			return total / size;
+		if (ret == 0) {
+			stream->eof = 1;
+			return total / size;
+		}
+		total += ret;
+	}
+
+	return total / size;
+}
+
+size_t fwrite(const void *ptr, size_t size, size_t nmemb, FILE *stream)
+{
+	if (unlikely(!stream))
+		return 0;
+
+	if (unlikely(!ptr || !size || !nmemb)) {
+		stream->errno = EINVAL;
+		return 0;
+	}
+
+	if (unlikely(stream->fd < 0)) {
+		stream->errno = EBADF;
+		return 0;
+	}
+
+	if (unlikely(SIZE_MAX / size < nmemb)) {
+		stream->errno = EOVERFLOW;
+		return 0;
+	}
+
+	size_t total = 0;
+
+	while (total < size * nmemb) {
+		ssize_t ret = pwrite(stream->fd,
+				((char *) ptr) + total,
+				size * nmemb - total, stream->offset);
+		if (ret > 0)
+			stream->offset += ret;  // Update the offset
+
+		if (ret < 0)
+			return total / size;
+		total += ret;
+	}
+
+	return total / size;
+}
+#endif /* CONFIG_LIBVFSCORE */
--- a/lib/nolibc/exportsyms.uk
+++ b/lib/nolibc/exportsyms.uk
@@ -44,6 +44,17 @@ putchar
 fputs
 puts
 
+# file
+clearerr
+feof
+ferror
+fseek
+fopen
+fclose
+fdopen
+fread
+fwrite
+
 # stdlib
 abort
 exit


lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+	return read(stream->fd, ptr, size * nmemb);
+}
+
+size_t fwrite(const void *ptr, size_t size, size_t nmemb, struct vfscore_file *stream)
+{
+	if (!stream)
+		return 0;
+	return write(stream->fd, ptr, size * nmemb);
+}
+#endif
```
skuenzer: Please add a comment:
`#endif /* CONFIG_LIBVFSCORE */`

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
 {
 	return fputs_internal(s, stdout, 1);
 }
+
+#if CONFIG_LIBVFSCORE
+
+int __fmodeflags(const char *mode)
```
skuenzer: In case this function is copied from somewhere, could you add a comment saying from where it is from? Since you use `__fmodeflags` only within this file, you can declare the function as `static`.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+{
+	int flags;
+
+	if (strchr(mode, '+'))
+		flags = O_RDWR;
+	else if (*mode == 'r')
```
skuenzer, line 575: As style suggestion: I'd prefer `mode[0] ==` instead of `*mode` to make clear that we are checking the first letter of the string. However, I am also totally fine if you prefer your variant.

lib/nolibc/include/stdio.h
```diff
@@ -93,6 +98,13 @@ int puts(const char *s);
 
 #if CONFIG_LIBVFSCORE
 int rename(const char *oldpath, const char *newpath);
+int feof(struct vfscore_file *stream);
+int fseek(struct vfscore_file *stream, long offset, int whence);
+int fclose(struct vfscore_file *stream);
+struct vfscore_file *fopen(const char *pathname, const char *mode);
```
skuenzer: Instead of `struct vfscore_file *` the return type must be `FILE *`. Otherwise we aren't API compatible with other libcs.
You should declare a type under `include/nolibc-internal/shareddefs.h`.
```c
#if (defined __NEED_FILE && !defined __DEFINED_FILE)
struct _nolibc_file; /* <- leave out this line if the compiler is happy without it */
typedef struct _nolibc_file FILE;
#endif /* (defined __NEED_FILE && !defined __DEFINED_FILE) */
```
Note that we are not exporting the struct details to the API user, we will define internally in `stdio.c`. The incomplete declaration is only there to enable type checks by the compiler.

lib/nolibc/stdio.c
```diff
@@ -80,7 +82,7 @@ static char const hex2ascii_data[] = "0123456789abcdefghijklmnopqrstuvwxyz";
  * The buffer pointed to by `nbuf' must have length >= MAXNBUF.
  */
 static inline char *ksprintn(char *nbuf, uintmax_t num, int base, int *lenp,
-			     int upper)
+				 int upper)
```
skuenzer: Please do not change whitespaces if unnecessary.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+	return flags;
+}
+
+struct vfscore_file *fopen(const char *pathname, const char *mode)
+{
+	struct vfscore_file *f;
+	int flags;
+	int fd;
+
+	if (!strchr("rwa", *mode)) {
```
skuenzer: If you prefer, you could also use `mode[0]`.

lib/nolibc/include/stdio.h
```diff
@@ -36,6 +36,11 @@
 
 #include <uk/essentials.h>
 
+
+#if CONFIG_LIBVFSCORE
+#include <vfscore/file.h>
```
skuenzer, line 41: No, you should not use `struct vfscore_file`. Normally, `FILE` is a libc-internal implementation and we want to keep these `f*`-functions agnostic of an actual filesystem library (meaning independent of `vfscore`). As background, `FILE` objects typically have a buffer where you read from and write to so that the number of system calls for `SYS_read`, `SYS_write` can be reduced. It is fine to provide with nolibc pure compatibility wrappers that map for instance `fopen()` to `open()`, `fread()` to `read()`, and so on.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+		errno = EINVAL;
+		return 0;
+	}
+
+	flags = __fmodeflags(mode);
+	fd = open(pathname, flags, 0666);
+
+	return vfscore_get_file(fd);
```
skuenzer: Here I'd `malloc` memory for `struct _nolibc_file` and return the pointer of it. In order to reduce duplication, I think you should handover the `fd` from the open call to `fdopen` and do the allocation in that function.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
 {
 	return fputs_internal(s, stdout, 1);
 }
+
+#if CONFIG_LIBVFSCORE
+
```
skuenzer, line 559: Here you can define the `strcut _nolibc_file`. For now, it is probably just:
```c
struct _nolibc_file {
    int fd;
    bool eof; /* true if we reached end-of-file */
};
```
Typically, you would also add a buffer and additional fields that are needed to reduce syscall overheads but we can start minimal since we want a simple wrapper for `nolibc` only.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+	flags = __fmodeflags(mode);
+	fd = open(pathname, flags, 0666);
+
+	return vfscore_get_file(fd);
+}
+
+int feof(struct vfscore_file *stream)
+{
+	return 0;
```
skuenzer: I think this should be:
```c
int feof(FILE *stream)
{
    return stream->eof;
}
```

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+}
+
+int fclose(struct vfscore_file *stream)
+{
+	return close(stream->fd);
```
skuenzer: Additionally you would need to `free` the `FILE` object as well.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+}
+
+int fseek(struct vfscore_file *stream, long offset, int whence)
+{
+	return lseek(stream->fd, offset, whence);
+}
+
+size_t fread(void *ptr, size_t size, size_t nmemb, struct vfscore_file *stream)
+{
+	return read(stream->fd, ptr, size * nmemb);
```
skuenzer: Here you'd need to check a libc manual to understand the expected behavior. We probably need to call `read` multiple times until `size * nmemb` bytes are read to the buffer or if we reached the end-of-file or an error. In the EOF case, `stream->eof` should be set, too.
To my understanding, we effectively want to read `nmemb` number of objects, each of having the size of `size` bytes. Probably, the function is reading in increments of `size` bytes to the buffer and returns the number of successful read elements. But please double check a manual.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+}
+
+size_t fwrite(const void *ptr, size_t size, size_t nmemb, struct vfscore_file *stream)
+{
+	if (!stream)
+		return 0;
+	return write(stream->fd, ptr, size * nmemb);
```
skuenzer: Please check if a write or seek operation (see `fseek()`) resets the `eof` flag.

skuenzer (changes requested):

skuenzer:
I am thinking if the following title would be shorter for your commit:
`lib/nolibc: Introduce 'FILE' object APIs`
🤔

lib/nolibc/stdio.c
RaduNichita: I think you should also do the same thing as @skuenzer suggested for  `fread`

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+	return vfscore_get_file(fd);
+}
+
+int fseek(struct vfscore_file *stream, long offset, int whence)
+{
+	return lseek(stream->fd, offset, whence);
+}
+
+size_t fread(void *ptr, size_t size, size_t nmemb, struct vfscore_file *stream)
+{
```
RaduNichita, line 701: I think you should add the same check for `NULL` for `stream variable,  as you did for `fwrite()` to keep consistency.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,78 @@ int puts(const char *s)
...
+	int flags;
+	int fd;
+
+	if (!strchr("rwa", *mode)) {
+		errno = EINVAL;
+		return 0;
+	}
+
+	flags = __fmodeflags(mode);
+	fd = open(pathname, flags, 0666);
```
RaduNichita, line 604: Could you use something from mode instead of hardcoding `0666` here :thinking:

RaduNichita (changes requested):
Please see my comments, @Starnox

lib/nolibc/stdio.c
Starnox: I wrote the function myself, but yeah, I will make it static

lib/nolibc/stdio.c
Starnox: I really don't know how that got there, weird 🤔

lib/nolibc/include/stdio.h
Starnox: I am a bit confused, probably because of my lack of understanding. Won't this definition be in conflict with this one `typedef struct _nolibc_fd FILE` from line 54?

lib/nolibc/stdio.c
Starnox: `checkpatch.uk` recommended this way of doing it so I went along with it.

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,124 @@ int puts(const char *s)
 {
 	return fputs_internal(s, stdout, 1);
 }
+
+#if CONFIG_LIBVFSCORE
+static int __fmodeflags(const char *mode)
+{
+	int flags;
+
+	if (strchr(mode, '+'))
```
skuenzer, line 573: In general, are these parameters positional? ...like `"w+x"` and `"ar+"` are valid while `"wx+"` and `"+ar"` would be invalid? Imagine: `fopen(pathname, "+")`: Would be `"+"` a complete and valid operation mode and should this be resolved to `O_RDWR`?
If invalid, we should parse differently so that we can return `-EINVAL` for the wrong cases. Maybe a variable storing the current position in the string could be helpful to match at an expected position of the string `mode[p]`. Btw, does `musl` has such a function for this conversion (mode string => flags) that we could borrow?

I am also thinking that if we have this function doing the check and parsing the arguments at the same time, you need a way to return an error (like `-1` or `-EINVAL`) and the decoded flags at the same time: We could handover `flags` by reference. What do you think?

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,124 @@ int puts(const char *s)
...
+	return flags;
+}
+
+struct _nolibc_file *fopen(const char *pathname, const char *mode)
+{
+	struct _nolibc_file *f;
+	int flags;
+	int fd;
+
+	if (!strchr("rwa", mode[0])) {
```
skuenzer: I think the mode string validation should also go to `__fmodeflags` so that this is at one place.

lib/nolibc/stdio.c
Starnox: I updated with a slightly modified version from `musl` libc. Could you also look at the comment I left regarding the `FILE` discussion?

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,133 @@ int puts(const char *s)
...
+		return 0;
+	}
+
+	size_t total = 0;
+
+	while (total < size * nmemb) {
+		int ret = write(stream->fd,
+					   ((char *) ptr) + total,
+					   size * nmemb - total);
+
```
RaduNichita, line 769:
```suggestion
```

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,133 @@ int puts(const char *s)
...
+		return NULL;
+
+	f->fd = fd;
+	f->eof = 0;
+	return f;
+}
+
+int fseek(struct _nolibc_file *stream, long offset, int whence)
+{
+	/* TODO set eof */
```
RaduNichita: What is with this TODO, @Starnox ?

lib/nolibc/stdio.c
```diff
@@ -541,3 +543,133 @@ int puts(const char *s)
...
+		return NULL;
+
+	fd = open(pathname, flags, 0666);
+
+	if (fd < 0)
+		return NULL;
+
+	f = fdopen(fd, mode);
+
+	return f;
```
RaduNichita:
```suggestion
	return fdopen(fd, mode);
```

RaduNichita (changes requested):

lib/nolibc/stdio.c
Starnox: Forgot it there, my bad, thanks!

John-Ted (approved):
Looks good to me, thanks!
Reviewed-by: Ioan-Teodor Teugea <ioan_teodor.teugea@stud.acs.upb.ro>

nderjung:
Hi @Starnox, please check the integration tests as they are failing.

lib/nolibc/include/nolibc-internal/shareddefs.h
```diff
@@ -88,6 +89,15 @@ typedef struct {
 #define __DEFINED_max_align_t
 #endif
 
+#if (defined __NEED_FILE && !defined __DEFINED_FILE)
+struct _nolibc_file {
+	int fd;
```
skuenzer: Please move the struct definition to `stdio.c`, just keep the following here:
```c
#if (defined __NEED_FILE && !defined __DEFINED_FILE)
struct _nolibc_file;
typedef struct _nolibc_file FILE;
#endif
```

lib/nolibc/include/nolibc-internal/shareddefs.h
```diff
@@ -39,6 +39,7 @@
 
 #include <uk/config.h>
 #include <uk/arch/types.h>
+#include <stdbool.h>
```
skuenzer: As soon as you moved the struct defintion of `struct _nolibc_file` to stdio.c, `<stdbool.h>` is not needed here. Any std header should not be included in this file. It is rather that this file is a core component on which the std headers are based on.

lib/nolibc/stdio.c
```diff
@@ -80,7 +83,7 @@ static char const hex2ascii_data[] = "0123456789abcdefghijklmnopqrstuvwxyz";
  * The buffer pointed to by `nbuf' must have length >= MAXNBUF.
  */
 static inline char *ksprintn(char *nbuf, uintmax_t num, int base, int *lenp,
-			     int upper)
+							int upper)
```
skuenzer: Why this whitespace change?

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,147 @@ int puts(const char *s)
 {
 	return fputs_internal(s, stdout, 1);
 }
+
+#if CONFIG_LIBVFSCORE
+/* The following code is taken from musl libc */
```
skuenzer: I guess you should mention `derived` because you modified the copied code.

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,147 @@ int puts(const char *s)
...
+
+	return 0;
+}
+
+FILE *fopen(const char *pathname, const char *mode)
+{
+	int flags;
+	int fd;
+
+	int ret = __fmodeflags(mode, &flags);
```
skuenzer: Please split declaration and assignment:
```c
int ret;

ret = __fmodeflags(mode, &flags);
```

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,147 @@ int puts(const char *s)
...
+}
+
+int fseek(FILE *stream, long offset, int whence)
+{
+	return lseek(stream->fd, offset, whence);
+}
+
+size_t fread(void *ptr, size_t size, size_t nmemb, FILE *stream)
+{
+	if (!stream || !ptr || !size || !nmemb) {
```
skuenzer: Can you encapsulate the error paths with `unlikely`?

skuenzer (changes requested):

lib/nolibc/stdio.c
Starnox: It is an auto format from my editor, I checked with checkpatch.uk and this formatting is correct

RaduNichita: Yes, @Starnox is correct, we have that in the [coding-conventions](https://unikraft.org/docs/contributing/coding-conventions#breaking-long-lines)

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,153 @@ int puts(const char *s)
...
+		errno = EINVAL;
+		return 0;
+	}
+
+	if (stream->fd < 0) {
+		errno = EBADF; // Bad file descriptor
+		return 0;
+	}
+
+	if (SIZE_MAX / size < nmemb) {
```
RaduNichita: Maybe add an unlikely here, since it doesn't happen frequently?

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,153 @@ int puts(const char *s)
...
```
RaduNichita: same here

RaduNichita (changes requested):
Looks good @Starnox . I have left some very small nitpicks. Please all update your commit message explaining why is this change necessary.

RaduNichita (approved):
All good on my side.

lib/nolibc/stdio.c
skuenzer: The formatting was correct originally and should not be touched in such a case. In general, whitespace changes should only be done if code really did not match convention and then we should aim to focus it on functions that our changes are already anyways touching. Alternatively, we can also introduce a dedicated commit to correct formatting in a file, however those are more risky for acceptance. The reason is that pure whitespace changes risk to break other PR's on the same file that need to be rebased and adopted to such whitespace changes.

This particular whitespace change seems to me 1) to be too unrelated and 2) the original version was not incorrect.
Before we had arguments aligning according ["Align function arguments after break to open parenthesis" (coding-conventions)](https://unikraft.org/docs/contributing/coding-conventions#breaking-long-lines):
```c
static inline char *ksprintn(char *nbuf, uintmax_t num, int base, int *lenp,
			     int upper)
```
`checkpatch` is also not complaining with this original version.

Your suggested change is pushing `int upper` away from the opening parenthesis and align `int`s:
```c
static inline char *ksprintn(char *nbuf, uintmax_t num, int base, int *lenp,
							int upper)
```

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,153 @@ int puts(const char *s)
...
+}
+
+int fseek(FILE *stream, long offset, int whence)
+{
+	return lseek(stream->fd, offset, whence);
+}
+
+size_t fread(void *ptr, size_t size, size_t nmemb, FILE *stream)
+{
+	if (unlikely(!stream || !ptr || !size || !nmemb)) {
```
skuenzer: Can you also `#include <uk/arch/lcpu.h>` to make sure the `unlikely` macro is defined? At the moment it should be indirectly be available but it is always good to explicitly include things in case nothing breaks if indirect headers are changed.

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,153 @@ int puts(const char *s)
...
+		errno = EINVAL;
```
skuenzer: I just noticed that according to POSIX and C standards, `fread` should probably not return error on `errno`. The actual interfaces for determining an error (and reset its error flag) are `ferror()`, `clearerr()`: https://man7.org/linux/man-pages/man3/ferror.3.html . So, I think we need to add and an `errno` field to the `FILE` struct and make the behavior matching with whats documented.

lib/nolibc/stdio.c
```diff
@@ -541,3 +544,153 @@ int puts(const char *s)
...
+
+	if (unlikely(SIZE_MAX / size < nmemb)) {
+		errno = EOVERFLOW;
+		return 0;
+	}
+
+	size_t total = 0;
+
+	while (total < size * nmemb) {
+		ssize_t ret = read(stream->fd,
```
skuenzer: Hum, I am not sure if it is maybe safer to use `pread()` (when `fd` supports it) and store `seek`  by our own. A program may malfunction due to an incorrect `seek` i fa program `reads` by itself form the same `fd` while it also interacts with it via `fdopen()`. Do you know how `musl` solves this case?

Obviously, this comment applies also for `fwrite()`.

skuenzer (changes requested):

razvand (approved):

PR #1044, opened 2023-08-12
lib/nolibc: Add `__errno_location` implementation

Starnox:
At link time, objects such as the ones derived from `gcov` which comes with `gcc-13` require this symbol in order to build.
There was also the need to remove the internalisation of symbols starting with `_`. This is still a work in progress and requires feedback.

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### DIFF
--- a/lib/nolibc/Makefile.uk
+++ b/lib/nolibc/Makefile.uk
@@ -52,6 +52,3 @@ LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/random.c
 LIBNOLIBC_SRCS-$(CONFIG_LIBNOLIBC_SYSLOG) += $(LIBNOLIBC_BASE)/syslog.c
 
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/qsort.c
-
-# Localize internal symbols (starting with __*)
-LIBNOLIBC_OBJCFLAGS-y += -w -L __*
--- a/lib/nolibc/errno.c
+++ b/lib/nolibc/errno.c
@@ -32,9 +32,15 @@
  */
 
 #include <errno.h>
+#include <stddef.h>
 
 #if CONFIG_HAVE_SCHED
 __thread int errno;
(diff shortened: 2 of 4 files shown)


lib/nolibc/errno.c
```diff
@@ -32,9 +32,15 @@
...
 
 #if CONFIG_HAVE_SCHED
 __thread int errno;
 #else
 int errno;
 #endif /* !CONFIG_HAVE_SCHED */
+
+int *__errno_location(void)
+{
+	return NULL;
```
jschlumpp: Shouldn't this point to errno?
```suggestion
	return &errno;
```

lib/nolibc/Makefile.uk
```diff
@@ -52,6 +52,3 @@ LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/random.c
 LIBNOLIBC_SRCS-$(CONFIG_LIBNOLIBC_SYSLOG) += $(LIBNOLIBC_BASE)/syslog.c
 
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/qsort.c
-
-# Localize internal symbols (starting with __*)
-LIBNOLIBC_OBJCFLAGS-y += -w -L __*
```
jschlumpp, line 57: I do wonder why this was here in the first place. Having an `exportsyms.uk` would already localize all unlisted symbols. @skuenzer

skuenzer: True, this is likely a pre-exportsyms leftover.

lib/nolibc/errno.c
skuenzer: I think so, too. In such a case, we have an actual implementation and not a stub anymore 😄

jschlumpp (approved):

razvand (approved):

PR #1047, opened 2023-08-14
plat/kvm: Align the bootstack base address

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`,`x86_64`]
 - Platform(s): [`kvm`]

### Description of changes

The base address of the stack must be 16-byte aligned on both arm64 and x86_64. Fix the alignent where this is currently missing.

### DIFF
--- a/plat/kvm/arm/entry64.S
+++ b/plat/kvm/arm/entry64.S
@@ -79,8 +79,9 @@ ENTRY(_libkvmplat_entry)
 	isb
 
 	/* Set the boot stack */
-	ur_ldr x26, lcpu_bootstack
-	mov sp, x26
+	ur_ldr	x26, lcpu_bootstack
+	and	x26, x26, ~(__STACK_ALIGN_SIZE - 1)
+	mov	sp, x26
 
 	/* Set the context id */
 	msr contextidr_el1, xzr


plat/kvm/arm/entry64.S
```diff
@@ -40,6 +40,7 @@
  * paging_init (it invalidates our stack)
  */
 .section .bss
+.align 16
```
mogasergiu: I see that, unlike x86, we do not have an explicit aligning of the stack like the `andq    $~0xf, %rsp`. Would it be worth adding this to ARM as well, just to double make sure.

mogasergiu (changes requested):
Thank you for making these changes :pray:! One minor comment.

plat/kvm/arm/entry64.S
michpappas: Based on offline discussion, updated PR to:
* Remove changes in x86_64, the alignment is enforced there on runtime already
* Update arm64 to enforce alignment on runtime too

mogasergiu (approved):
I do not like how this is inconsistent with x86, as there is no `__STACK_ALIGN_SIZE`, but, alas, this is getting out of scope and will maybe be addressed by plat re-arch, hopefully.

michpappas:
> I do not like how this is inconsistent with x86, as there is no `__STACK_ALIGN_SIZE`, but, alas, this is getting out of scope and will maybe be addressed by plat re-arch, hopefully.
> 
> Reviewed-by: Sergiu Moga [sergiu@unikraft.io](mailto:sergiu@unikraft.io)

Agreed. In fact I didn't like the `__STACK_ALIGN_SIZE` naming either, but avoided to change it as it's part of a wider series of  `#defines`in that header and would need to change it in other places too. Thanks!

razvand (approved):

PR #1048, opened 2023-08-14
plat/kvm: enable 8250 / NS16550 for Firecracker on arm64

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [`kvm`]

### Description of changes

Update the KVM configuration to enable the 8250 / NS16550 UART driver. This driver is selected when KVM_VMM_FIRECRACKER is selected on arm64.

This is an interim replacement of #986  to enable UART on Firecracker for the `0.14` release.
CC: @zhxj9823


plat/kvm/Config.uk
```diff
@@ -126,35 +118,66 @@ config KVM_DEBUG_VGA_CONSOLE
         help
           Choose VGA console for the debug printing
 
+config KVM_KERNEL_SERIAL_CONSOLE
+        bool "Serial console for the kernel prints"
+        default y
+        depends on (ARCH_X86_64 || ARCH_ARM_64)
```
mogasergiu, line 124: I am not sure this dependency is fair. These are just UART compatible chips, they are not architecture dependent right? Especially the NS16650. Perhaps it would be better to just have them depend on `PLAT_KVM`?

plat/kvm/Makefile.uk
```diff
@@ -86,8 +86,12 @@ endif
 ## Architecture library definitions for arm64
 ##
 ifeq ($(findstring y,$(CONFIG_KVM_KERNEL_SERIAL_CONSOLE) $(CONFIG_KVM_DEBUG_SERIAL_CONSOLE)),y)
+ifeq ($(CONFIG_KVM_VMM_FIRECRACKER),y)
```
mogasergiu: Why does it need to be condittioned by the VMM?

plat/kvm/Makefile.uk
```diff
@@ -86,8 +86,12 @@ endif
...
+LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_DRIVERS_BASE)/uart/ns16550.c
```
mogasergiu: Perhaps just
```
LIBKVMPLAT_SRCS-$(CONFIG_UART_NS16650) += $(UK_PLAT_DRIVERS_BASE)/uart/ns16550.c
```

plat/kvm/Makefile.uk
```diff
@@ -86,8 +86,12 @@ endif
...
+else
 LIBKVMPLAT_SRCS-$(CONFIG_ARCH_ARM_64) += $(UK_PLAT_DRIVERS_BASE)/uart/pl011.c
```
mogasergiu: Ditto

plat/kvm/include/kvm-arm64/uart.h
```diff
@@ -0,0 +1,23 @@
...
+ */
+#if CONFIG_UART_NS16550
+#include <uart/ns16550.h>
+#endif /* CONFIG UART_NS16550 */
+
+#if CONFIG_UART_PL011
+#include <uart/pl011.h>
+#endif /* CONFIG_UART_PL011 */
+
+static inline void kvm_console_init(void *fdt)
```
mogasergiu, line 14: Why is this in ARM only? Will this be moved to arch agnostic place after plat re-arch?

mogasergiu (commented):
Hi, had a quick run. I am not sure of the choices for dependencies. Are these temporary just until plat re-arch is finished?

plat/kvm/Makefile.uk
michpappas: Because I forgot to update it, thanks

plat/kvm/include/kvm-arm64/uart.h
michpappas: Yes, this will be obsoleted by #986

plat/kvm/Config.uk
michpappas: Actually that was not introduced in this PR, so I kept it as is to avoid changing too many things since this is very temporary. All this is wrapped in a huge `if (PLAT_KVM)` so in general we could remove it. On the other hand such architectural definitions are also set in all other drivers in this file, so there may be some reasoning there I'm not aware of. Besides all this will be reworked in plat re-arch, so maybe we should leave it as is for now. What do you think?

mogasergiu: > there may be some reasoning there I'm not aware of

Good point, we don't have much context here yet.

> should leave it as is for now

Yes, definetely. Too many last minute changes in the last few days anyway... :smile:.

plat/kvm/Config.uk
```diff
@@ -126,35 +118,66 @@ config KVM_DEBUG_VGA_CONSOLE
...
+
+	config KVM_SERIAL_BAUD_19200
+		bool "19200"
+endchoice
+endif
+
+config EARLY_PRINT_PL011_UART_ADDR
+	hex "Early debug console pl011 serial address"
+	depends on UART_PL011
+	default 0x09000000
```
mogasergiu, line 166: Is there a reason why `PL011`'s early configs (`EARLY_PRINT_PL011_UART_ADDR`) is inconsistent with those of `NS16550` (`CONFIG_EARLY_UART_NS16550`*)? `CONFIG_EARLY_UART_NS16550` does not seem to be used anywhere, therefore I think we could only have `EARLY_PRINT_PL011_ADDR` and `CONFIG_EARLY_PRINT_NS16550`. I like not having `UART` in the name to accomodate for video devices that could register themselves as console drivers through their framebuffers (e.g. VGA/GOP)

michpappas: Yes, it's for backwards compatibility, these options are currently used by OpSy. So if we are going to change them it's best to do it once, in the scope of #986.

plat/kvm/Config.uk
```diff
@@ -126,35 +118,66 @@ config KVM_DEBUG_VGA_CONSOLE
...
+
+config EARLY_UART_NS16550
+	bool "Early console"
+	depends on UART_NS16550
+	default n
+
+config EARLY_UART_NS16550_BASE
+	hex "Early debug console ns16550 serial address"
+	depends on EARLY_UART_NS16550
+	help
```
zhxj9823, line 179: Should there be a default for this config as well?

plat/kvm/Makefile.uk
```diff
@@ -86,7 +86,8 @@ endif
 ## Architecture library definitions for arm64
 ##
 ifeq ($(findstring y,$(CONFIG_KVM_KERNEL_SERIAL_CONSOLE) $(CONFIG_KVM_DEBUG_SERIAL_CONSOLE)),y)
```
zhxj9823: It does not make sense to use if conditions here as `if (KVM_KVM_KERNEL_SERIAL_CONSOLE || KVM_DEBUG_SERIAL_CONSOLE)` is used in the config, so `CONFIG_UART_NS16550` and `CONFIG_UART_PL011` are already dependent on `KVM_KVM_KERNEL_SERIAL_CONSOLE` and `KVM_DEBUG_SERIAL_CONSOLE`.

plat/kvm/Config.uk
michpappas: For Firecracker it is not possible as it lays out the MMIO regions depending on the devices to use. Saying that, there should be no default value for any driver. That is because a given driver can be potentially used by different platforms. In other words, the address for early UART only makes sense as a command-line parameter.
The only reason I am leaving the default for `pl011` is backwards compatibility. In the scope of #986 we will be removing this too.

plat/kvm/Makefile.uk
michpappas: Good catch, thanks. Now fixed 👍🏼

mogasergiu (approved):

mariasfiraiala (approved):
All good on my part, thanks!

zhxj9823 (approved):

zhxj9823 (approved):
Reviewed-by: Xingjian Zhang [zhxj9823@qq.com](mailto:zhxj9823@qq.com)

michpappas:
@kubanrob, as stated in the cover letter: These are interim changes, before bigger rework on UART in 0.15. I tried to preserve backwards compatibility. If there is something that breaks for you please speak up, otherwise this has been tested and we're ready to merge.

razvand (approved):

PR #1049, opened 2023-08-14
plat/kvm/arm: Rework early cache clean & invalidate

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [`kvm`]

### Description of changes

During boot the following cache operations are required:
- Clean the cache to the PoC to avoid evicted dirty cache lines to overwrite subsequent writes.
- With the MMU / caches off, perform any accesses necessary
- Invalidate any memory accessed to avoid clean cache lines to shadow what we previously wrote into memory.

The arm64 linux boot protocol provides the requirements for the system's state before jumping into the kernel [1[. Among these it is required that upon entry:

* The MMU is off
* The D-cache for the region corresponging to the loaded image must be cleaned to the point of coherence.

These conditions provide the opportunity of optimization the existing boot code. Specifically we can skip the extremely expensive clean & invalidating of the entire image on LXBOOT and QEMU virt, and futhermore limit further cache invalidation to the regions accessed before enabling the MMU.

Additionally to the above, optimize the existing cache and invalidate functionality by using a single barrier at the end of the operation.

### DIFF
--- a/plat/kvm/arm/entry64.S
+++ b/plat/kvm/arm/entry64.S
@@ -36,11 +36,13 @@
 #include <uk/config.h>
 #include <uk/reloc.h>
 
+#define BOOTSTACK_SIZE	4096
+
 /* Prefer using in-image stack, to avoid conflicts when unmapping is done by
  * paging_init (it invalidates our stack)
  */
 .section .bss
-.space 4096
+.space BOOTSTACK_SIZE
 lcpu_bootstack:
 
 .text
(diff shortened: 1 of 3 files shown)


michpappas:
Marking as Draft to discuss PIE with @mogasergiu and get feedback from @kubanrob.

plat/kvm/arm/entry64.S
```diff
@@ -57,15 +59,22 @@ ENTRY(_libkvmplat_entry)
...
 
-	/*
-	 * We will disable MMU and cache before the pagetables are ready.
-	 * This means we will change memory with cache disabled, so we need to
-	 * invalidate the cache to ensure there is no stale data in it.
-	 * But we don't know the size of the RAM either. And it would be
-	 * expensive to invalidate the whole cache. In this case, just
-	 * just need to invalidate what we are going to use:
-	 * DTB, TEXT, DATA, BSS, and bootstack.
+#ifndef CONFIG_OPTIMIZE_PIE
```
kubanrob: Does `OPTIMIZE_PIE` imply that the Linux Boot Protocol is used? At a first glance this seems unrelated.

michpappas: It's unrelated indeed (also notice this is an `ifndef`). I do not make an explicit check for `CONFIG_KVM_BOOT_PROTO_LXBOOT` but rely on the MMU heuristic to also cover for QEMU virt, and potentially other platforms that don't use the linux boot protocol but exhibit the same behavior. Tbh I'm between two minds about this. On the one hand implicit detection looks like a bad idea, on the other hand it allows more flexibility. If we want to be on the safe side I can make this conditional to `BOOT_PROTO_LXBOOT` OR `BOOT_PROTO_QEMU_VIRT`.
What do you think?

michpappas:
After discussing with @mogasergiu I am updating this to also have PIE rely on the bootloader to have the cache clean, but still unconditionally invalidate the entire image region as `libukreloc` won't invalidate individual cache lines right now.

plat/kvm/arm/entry64.S
```diff
@@ -36,11 +36,13 @@
 #include <uk/config.h>
 #include <uk/reloc.h>
 
+#define BOOTSTACK_SIZE	4096
```
mogasergiu, line 39: In commit message of `plat/kvm: Do not unconditionally clean and invalidate the cache`:
s/[1[/[1]/
s/inalidate/invalidate

plat/kvm/arm/entry64.S
```diff
@@ -57,31 +59,58 @@ ENTRY(_libkvmplat_entry)
...
+	 * unpredicatable. In that case we clean-invalidate the cache for
+	 * the entire image and disable the MMU.
+	 *
+	 * Notice: A clean cache would suffice as long as we invalidate any
+	 *         memory we touch while the MMU is disabled. But since PIE
+	 *         can write anywhere we also invalidate the entire image.
+	 */
+	mrs	x19, sctlr_el1
+	and	x19, x19, #SCTLR_EL1_M_BIT
+#ifdef CONFIG_OPTIMIZE_PIE
```
mogasergiu: s/ifdef/if

plat/kvm/arm/entry64.S
```diff
@@ -57,31 +59,58 @@ ENTRY(_libkvmplat_entry)
...
-	msr sctlr_el1, x2
+	dsb	sy
+	mrs	x2, sctlr_el1
+	mov	x3, #SCTLR_EL1_M_BIT|SCTLR_EL1_C_BIT
+	bic	x2, x2, x3
+	msr	sctlr_el1, x2
 	isb
-
-	/* Set the boot stack */
+#ifdef CONFIG_OPTIMIZE_PIE
```
mogasergiu: Ditto

mogasergiu (changes requested):
Awesome optimizations! :100:
A few minor comments.

michpappas:
Updated commit message, fixed conditional includes according to coding style based on comments from @mogasergiu. Replaced `CONFIG_OPTIMIZE_PIE` with `CONFIG_LIBUKRELOC` based on internal discussion with @skuenzer.

michpappas:
This has been successfully tested on QEMU both natively and emulated, and on Firecracker. @kubanrob besides the above open question, please let me know if you need to do any testing on your side, otherwise we're close to merge.

mogasergiu:
> Updated commit message, fixed conditional includes according to coding style based on comments from @mogasergiu. Replaced `CONFIG_OPTIMIZE_PIE` with `CONFIG_LIBUKRELOC` based on internal discussion with @skuenzer.

Right, that too, forgot about it, thanks for keeping it in mind! :pray:

mogasergiu (approved):
It is very much appreciated that you took PIE into consideration. Thank you! :small_airplane: 

razvand (approved):

PR #1054, opened 2023-08-15
lib/posix-environ: Ensure that compiled-in strings are writable

mogasergiu:
Previously, the elements of `__init_env` would be compile-in strings
that would correspond to `CONFIG_LIBPOSIX_ENVIRON_ENVPx` configurable
strings. This lead to these strings being placed by in the `.rodata`
section and making them unmodifiable.

Unfortunately, Redis represents an application whose code may want to
modify such data (see
commit 6356cf6808be ("Set process name in ps output to make operations safer.").
Therefore, to appease such actions, make sure that this data is placed
in a modifiable section such as `.data` by using statically declared
global variables containing these very strings as an indirection.

To make it look somewhat nicer, do so with the help of two basic helper
macros: DECLARE_LIBPOSIX_ENVIRON_ENV_VAR to transparently declare such
variables and LIBPOSIX_ENVIRON_ENV_VAR to transparently access them.

Signed-off-by: Sergiu Moga <sergiu@unikraft.io>

### Prerequisite checklist

### Base target

 - Architecture(s): [All]
 - Platform(s): [All]
 - Application(s): [All]


michpappas (changes requested):
Looks good to me. One minor point is the leading underscore in the variable name. Unless there is a strong reason to be there it should be changed to something in compliance with our guidelines.

mogasergiu:
> Looks good to me. One minor point is the leading underscore in the variable name. Unless there is a strong reason to be there it should be changed to something in compliance with our guidelines.

I noticed that @skuenzer likes using those :stuck_out_tongue_closed_eyes:.

lib/posix-environ/environ.c
```diff
@@ -7,6 +7,28 @@
 #include <uk/essentials.h>
 #include "environ.h"
 
+#define DECLARE_LIBPOSIX_ENVIRON_ENV_VAR(x)			\
+	static char _e##x[] = CONFIG_LIBPOSIX_ENVIRON_ENVP##x
+
+#define LIBPOSIX_ENVIRON_ENV_VAR(x)		_e##x
```
skuenzer: If you swap `LIBPOSIX_ENVIRON_ENV_VAR(x)` with `DECLARE_LIBPOSIX_ENVIRON_ENV_VAR(x)` you can even use `LIBPOSIX_ENVIRON_ENV_VAR(x)` within `DECLARE_LIBPOSIX_ENVIRON_ENV_VAR(x)` :)

skuenzer:
> > Looks good to me. One minor point is the leading underscore in the variable name. Unless there is a strong reason to be there it should be changed to something in compliance with our guidelines.
> 
> I noticed that @skuenzer likes using those 😝.

I have to admit that this is true. I liked them for things to mark them as "do not touch", especially when they are part of a system declared in the file/library. But since we have our convention, we can change it. In this sense @michpappas is right that we should be a good example for other contributors.

michpappas (approved):

skuenzer (approved):
Thanks a lot for these changes!

PR #1060, opened 2023-08-21
plat: Let coalescing method ensure memregion alignment

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): [all]
 - Platform(s): [all]
 - Application(s): [all]

### Description of changes

In order to preserve alignment among memory region descriptors, we
would strive to align every manually inserted memory region. However,
this would also rely on the boot protocol itself to report initial aligned
memory regions, which seems not be the case for some boot protocols
(e.g. cmdline may not be aligned in `Multiboot1`, `initrd` start
address may not be aligned in Linux Boot Protocol).

Therefore, write a single, centralized solution, inside the memory
region coalescing function and make it handle misalignment cases
as well to cope with misbehaving boot protocols. To guarantee
alignment, take each memory region, align it to both ends and
then resolve overlappings. Just like before, free memory regions
are trimmed in favor of other types of memory regions. We know
free memory regions will always be aligned because resolving
conflicts between aligned memory region descriptors will always
yield aligned memory regions.

To preserve original sizes of user allocated memory region descriptors,
in case a deletion is not detected, restore the start address and
length of such region and make sure that the `vbase` field stays updated
with new values of `pbase`.

Finally, now that we offload alignments to the newly improved coalescing
method, remove manual, hardcoded alignments from early boot glue code.
If everything goes as planned and is implemented in a manner where
expectations are met, these alignments should not be needed as they
are transparently ensured by the coalescing method as well as the
paging initialization.

NOTE: Re-adjust the other parts of the code that are a client of the memregion
API's to throw an exception if a memregion is not already aligned. From now on,
client code of such API shall not attempt to align received memory regions,
but rather throw an exception if that has not happened already. This way we
will be able to catch more cases that the coalescing method does not care
of and update it accordingly, to maintain one single centralized solution.

### DIFF
--- a/lib/ukboot/boot.c
+++ b/lib/ukboot/boot.c
@@ -86,6 +86,7 @@
 #ifdef CONFIG_LIBUKSP
 #include <uk/sp.h>
 #endif
+#include <uk/arch/paging.h>
 #include <uk/arch/tls.h>
 #include <uk/plat/tls.h>
 #include "banner.h"
@@ -186,6 +187,11 @@ static struct uk_alloc *heap_init()
 	 * add every subsequent region to it.
 	 */
 	ukplat_memregion_foreach(&md, UKPLAT_MEMRT_FREE, 0, 0) {
+		UK_ASSERT(md->vbase == md->pbase);
+		UK_ASSERT(!(md->pbase & ~PAGE_MASK));
+		UK_ASSERT(md->len);
+		UK_ASSERT(!(md->len & ~PAGE_MASK));
+
 		uk_pr_debug("Trying %p-%p 0x%02x %s\n",
 			    (void *)md->vbase, (void *)(md->vbase + md->len),
 			    md->flags,
--- a/plat/common/memory.c
+++ b/plat/common/memory.c
@@ -65,16 +65,18 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 {
 	struct ukplat_memregion_desc *mrd, alloc_mrd = {0};
 	__vaddr_t unmap_start, unmap_end;
+	__sz unmap_len, desired_sz;
 	struct ukplat_bootinfo *bi;
 	__paddr_t pstart, pend;
 	__paddr_t ostart, olen;
-	__sz unmap_len;
 	int rc;
 
 	unmap_start = ALIGN_DOWN(bpt_unmap_mrd.vbase, __PAGE_SIZE);
 	unmap_end = unmap_start + ALIGN_DOWN(bpt_unmap_mrd.len, __PAGE_SIZE);
 	unmap_len = unmap_end - unmap_start;
 
+	/* Preserve desired size */
+	desired_sz = size;
 	size = ALIGN_UP(size, __PAGE_SIZE);
 	ukplat_memregion_foreach(&mrd, UKPLAT_MEMRT_FREE, 0, 0) {
 		UK_ASSERT(mrd->pbase <= __U64_MAX - size);
@@ -106,7 +108,7 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 		if (olen - (pstart - ostart) == size) {
 			mrd->pbase = pstart;
 			mrd->vbase = pstart;
-			mrd->len = pend - pstart;
+			mrd->len = desired_sz;
 			mrd->type = type;
 			mrd->flags = flags;
 
@@ -122,7 +124,7 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 		/* Insert allocated region */
 		alloc_mrd.vbase = pstart;
 		alloc_mrd.pbase = pstart;
-		alloc_mrd.len   = size;
+		alloc_mrd.len   = desired_sz;
 		alloc_mrd.type  = type;
 		alloc_mrd.flags = flags | UKPLAT_MEMRF_MAP;
 
--- a/plat/common/memory.c
+++ b/plat/common/memory.c
@@ -180,6 +180,8 @@ static inline int get_mrd_prio(struct ukplat_memregion_desc *const m)
  * contains the right memory region of higher priority, then split the left one
  * in two, by adjusting the current left one and inserting a new memory region
  * descriptor.
+ * The start of the left memory region in the list is always less
+ * or equal to the start of the right memory region in the list.
  */
 static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
 					 struct ukplat_memregion_desc *const ml,
@@ -217,22 +219,27 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
 		 * a new one if the left region is larger than the right region.
 		 */
 		} else {
+			__sz len = ml->pbase + ml->len - mr->pbase - mr->len;
+
 			if (RANGE_CONTAIN(ml->pbase, ml->len,
-					  mr->pbase, mr->len))
-				/* Ignore insertion failure as there is nothing
-				 * we can do about it and it is not worth caring
-				 * about.
+					  mr->pbase, mr->len && len))
+				/* len here is basically ml_end - mr_end. Thus,
+				 * len == 0 can happen only if mr is at the end
+				 * of the ml and we therefore ignore the rest.
+				 * If we ended up here then mr is actually
+				 * somewhere in the middle of ml and we are
+				 * inserting the mrd between mr_end and ml_end
 				 */
 				ukplat_memregion_list_insert_at_idx(list,
 					&(struct ukplat_memregion_desc){
 						.vbase = mr->pbase + mr->len,
 						.pbase = mr->pbase + mr->len,
-						.len   = ml->pbase + ml->len -
-							 mr->pbase - mr->len,
+						.len   = len,
 						.type  = ml->type,
 						.flags = ml->flags
 					}, ridx + 1);
 
+			/* Drop the fraction of ml that overlaps with mr */
 			ml->len = mr->pbase - ml->pbase;
 		}
 	}
@@ -279,6 +286,8 @@ static void ukplat_memregion_restore_mrd(struct ukplat_memregion_desc *mrd,
 /* Quick function to do potentially necessary swapping of two adjacent memory
  * region descriptors. Here just to modularize code because
  * ukplat_memregion_list_coalesce was quite large already.
+ * Guarantees the start of the left memory region in the list is always less
+ * or equal to the start of the right memory region in the list.
  */
 static void ukplat_memregion_swap_if_unordered(struct ukplat_memregion_list *l,
 					       __u32 l_idx, __u32 r_idx)
--- a/plat/common/memory.c
+++ b/plat/common/memory.c
@@ -154,19 +154,22 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
  * - the others are all allocated for Unikraft so they will have the same
  * priority
  */
+#define MRD_PRIO_FREE			0
+#define MRD_PRIO_KRNL_RSRC		1
+#define MRD_PRIO_RSVD			2
 static inline int get_mrd_prio(struct ukplat_memregion_desc *const m)
 {
 	switch (m->type) {
 	case UKPLAT_MEMRT_FREE:
-		return 0;
+		return MRD_PRIO_FREE;
 	case UKPLAT_MEMRT_INITRD:
 	case UKPLAT_MEMRT_CMDLINE:
 	case UKPLAT_MEMRT_STACK:
 	case UKPLAT_MEMRT_DEVICETREE:
 	case UKPLAT_MEMRT_KERNEL:
-		return 1;
+		return MRD_PRIO_KRNL_RSRC;
 	case UKPLAT_MEMRT_RESERVED:
-		return 2;
+		return MRD_PRIO_RSVD;
 	default:
 		return -1;
 	}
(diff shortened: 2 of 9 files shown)


mogasergiu:
Failing points met so far and why this should be done:
- QEMU's multiboot cmdline start address happened to be misaligned one time. This seems to be generally unreliable only in the case of the command-line. GRUB, the official reference implementation of multiboot, also does not explicitly align the command-line, it simply [allocates](https://github.com/coreos/grub/blob/2.02-coreos/grub-core/loader/i386/multiboot_mbi.c#L675) memory with an alignment of [0](https://github.com/coreos/grub/blob/2.02-coreos/grub-core/kern/mm.c#L380)
- ARM CCA GSOC student booting through the `Linux Boot Protocol` (our `lxboot`) was met with a misaligned initrd
start address on `kvmtool`. This actually seems to be [defined behavior](https://github.com/kvmtool/kvmtool/blob/e17d182ad3f797f01947fc234d95c96c050c534b/arm/kvm.c#L98)
- we need to have page-aligned non-overlapping memory regions that are flagged as
`UKPLAT_MEMRF_MAP` because we would have duplicate mappings, possibly with
conflicting flags (e.g. overlapping `UKPLAT_MEMRT_STACK` (`rw-` permissions) with `UKPLAT_MEMRT_DEVICETREE` (`r--` permissions), due to forced [alignments](https://github.com/unikraft/unikraft/blob/staging/plat/common/paging.c#L1506)),  and would end up making paging initialization issue an
`EEXIST`
- we need to have page-aligned free memory regions because otherwise they will
end up overlapping with `UKPLAT_MEMRF_MAP `memory regions and when the allocator
reaches that specific overlapping free memory region it will end up being
double mapped, resulting in an undefined behavior

plat/common/paging.c
```diff
@@ -1453,22 +1453,23 @@ int ukplat_paging_init(void)
...
-		len    = PAGE_ALIGN_DOWN(mrd->len - (paddr - mrd->pbase));
+		UK_ASSERT(!(mrd->pbase & ~PAGE_MASK));
+		UK_ASSERT(!(mrd->len & ~PAGE_MASK));
 
 		/* Not mapped */
 		mrd->vbase = __U64_MAX;
 		mrd->flags &= ~UKPLAT_MEMRF_PERMS;
 
-		if (unlikely(len == 0))
+		if (unlikely(mrd->len == 0))
```
michpappas: Is there a legitimate case of free regions with zero length? Should this be an assert instead?

plat/common/paging.c
```diff
@@ -1480,6 +1481,11 @@ int ukplat_paging_init(void)
 				  paddr, paddr + len, rc);
```
michpappas: These need to be updated to `mrd->paddr`, `mrd->len` too

lib/ukboot/boot.c
```diff
@@ -178,6 +178,7 @@ static struct uk_alloc *heap_init()
 	a = uk_alloc_init((void *)heap_base, alloc_pages << PAGE_SHIFT);
 #endif /* !CONFIG_LIBUKVMEM */
 #else /* CONFIG_LIBUKBOOT_HEAP_BASE */
+#define PAGE_OFF				(0xfff)
```
michpappas: You can include `uk/arch/paging.h` unconditionally to `CONFG_PAGING` and use `PAGE_MASK` from there.

plat/common/memory.c
```diff
@@ -106,7 +108,7 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 		if (olen - (pstart - ostart) == size) {
 			mrd->pbase = pstart;
 			mrd->vbase = pstart;
-			mrd->len = pend - pstart;
+			mrd->len = desired_sz;
```
michpappas, line 111: > A user may want however to fetch some memory regions and know the actual, intended size

From the above the *exact* motivation of this change is unclear to me, but I am a bit skeptical because this information may be later used by other code that performs operations at the page level. One example is `lib_ukrsi_init()` in (https://github.com/unikraft/unikraft/pull/964/) that sets the RIPAS for each region. Right now the implementation calls  `ALIGN_UP(mrd->len, __PAGE_SIZE)`, which makes me slightly uncomfortable as feels quite implicit.

One way to satisfy both requirements is to append an additional 'page_count' element into the mrd structure so that we preserve both values, similarly to what we do in `metadata_ifpages` of  `ukalloc` for MTE: https://github.com/unikraft/unikraft/blob/staging/lib/ukalloc/alloc.c#L78. That of course needs to be maintained for all mrds, including those that are not inserted via `ukplat_memregion_alloc`, and would require coalesce to also update the `page_count` as needed.

On the other hand one may argue that it's enough to rely to the guarantees provided by the mrds, and always assume that:
1. mrd->[p,v]addr is always page aligned, regardless of a region's origin origin (alloc or insert+coalesce). 
2. mrd->size may or may not be aligned (the latter isonly possible when the region was inserted via `ukplat_memregion_alloc`), but can always be safely aligned down to page size.

If we agree that these are indeed guarantees, then we can also keep it as is. What do you think?

plat/common/memory.c
```diff
@@ -235,45 +238,92 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
...
 		mr_prio = get_mrd_prio(mr);
+		if (unlikely(mr_prio < 0))
+			return -EINVAL;
 
-		/* If they overlap */
-		if (RANGE_OVERLAP(ml->pbase,  ml->len, mr->pbase, mr->len)) {
+		ukplat_memregion_align_mrd(ml, &ml_opbase, &ml_olen);
+		ukplat_memregion_align_mrd(mr, &mr_opbase, &mr_olen);
+
+		/* If they overlap. We assume MRD_PRIO_KRNL_RSRC will never
```
michpappas: "If they overlap." is not a sentence 😆 "If they overlap, we assume...", or "If they overlap: We assume..."

But more importantly, would an assertion be preferable over a "gentlemen's agreement"? 🎩

plat/common/memory.c
```diff
@@ -233,45 +238,92 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
...
 
+static void ukplat_memregion_align_mrd(struct ukplat_memregion_desc *mrd,
+				       __paddr_t *opbase, __sz *olen)
+{
+	*opbase = mrd->pbase;
+	*olen = mrd->len;
+
+	mrd->len = ALIGN_UP(mrd->pbase + mrd->len, __PAGE_SIZE);
+	mrd->pbase = ALIGN_DOWN(mrd->pbase, __PAGE_SIZE);
+	mrd->len -= mrd->pbase;
```
michpappas: Now that I understand your intention, everything is clear (and in fact a clever way to address the problem), but initially I found this sequence of statements initially confusing because `mrd->len` is being used as an intermediate variable for something different than its purpose. I don't have a strong opinion about it, as the confusion is only momentary, but just FYI in case you'd like to rephrase it using a temporary variable or leave a brief comment.

plat/common/memory.c
```diff
@@ -233,45 +238,92 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
...
+		ukplat_memregion_align_mrd(ml, &ml_opbase, &ml_olen);
+		ukplat_memregion_align_mrd(mr, &mr_opbase, &mr_olen);
+
+		/* If they overlap. We assume MRD_PRIO_KRNL_RSRC will never
+		 * overlap with MRD_PRIO_RSVD.
+		 */
+		if (RANGE_OVERLAP(ml->pbase, ml->len, mr->pbase, mr->len)) {
 			/* If they are not of the same priority */
 			if (ml_prio != mr_prio) {
 				overlapping_mrd_fixup(list, ml, mr, ml_prio,
```
michpappas, line 362: Unless I'm missing something. ;) Neither here or in `overlapping_mrd_fixup()` we take into account flags, and that's a bug waiting to happen, or worse a security issue. If that's correct, then this condition should check if the flags are the same before calling `overlappping_mrd_fixup()`, and bail with an error otherwise.

plat/common/memory.c
```diff
@@ -324,9 +379,32 @@ int ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
 			   ml_prio == mr_prio && ml->flags == mr->flags) {
 			ml->len += mr->len;
 			ukplat_memregion_list_delete(list, i + 1);
+			del = 1;
 		} else {
 			i++;
 		}
+
+restore_mrds:
+		/* We assume only MRD_PRIO_FREE can be dropped. We want
```
michpappas: Is there a reason this happens post-processing as opposed to deleting conditionally to these? Similarly for the updating of `mrd->vaddr`.

michpappas (changes requested):
Hi @mogasergiu this looks like a good approach to the problem. Please find some review comments below.

plat/common/memory.c
mogasergiu: > 1. mrd->[p,v]addr is always page aligned, regardless of a region's origin origin (alloc or insert+coalesce).

They may not be aligned, the aligned mappings are to be guaranteed by `ukplat_paging_init`'s last `for` loop that aligns all `UKPLAT_MEMRT_MAP` mrd's. It is the free memory regions whose [p,v]addr and len must always be aligned.

> 2.  mrd->len may or may not be aligned (the latter isonly possible when the region was inserted via > ukplat_memregion_alloc), but can always be safely aligned down to page size.

Yes, this we can assume.

In short, with this commit, the only mrd's that shall be aligned are the free ones, while the others keep their intended, original, values.

plat/common/memory.c
mogasergiu: I will add both a comment and a temporary variable :smile:.

plat/common/memory.c
mogasergiu: > "If they overlap." is not a sentence 😆 "If they overlap, we assume...", or "If they overlap: We assume..."

Oops, that is part of the old delete comment, forgot to remove that.

> But more importantly, would an assertion be preferable over a "gentlemen's agreement"? 🎩

Good point, I will add an `UK_ASSERT(mr_prio != MRD_PRIO_RSVD && ml_prio != MRD_PRIO_RSVD)`

plat/common/memory.c
mogasergiu: Hmm, actually I am assuming that only free mrd's may overlap with mrd's of other prio. So, instead, I would add another assertion: `UK_ASSERT(ml_prio == MRD_PRIO_FREE || mr_prio == MRD_PRIO_FREE);` to ensure that at least one of the overlapping regions is a free one that must e adjusted in favor of the other. Kernel rsrc type mrd's should never overlap with anything as those represent what we manually insert.

plat/common/memory.c
```diff
@@ -324,9 +379,32 @@ int ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
+		if (!del && mr_prio == MRD_PRIO_KRNL_RSRC)
+			ukplat_memregion_restore_mrd(mr, mr_opbase, mr_olen);
+
+		/* Update ml's vbase, since it might not be equal to pbase
+		 * anymore. Whether we deleted ml or mr it does not matter,
+		 * as ml is now equal to the remaining one, because
+		 * ukplat_memregion_list_delete() removes by `memmove()`ing.
+		 */
+		ml->vbase = ml->pbase;
 	}
```
mogasergiu, line 474: I should add a `m[i]->vbase = m[i]->pbase;` out of the loop here, as there may be a case where the last mrd is left unrestored.

plat/common/memory.c
mogasergiu: Thought I would only have one place where this is done :thinking:. I could have also wrapped this into a short function instead :thinking:.

plat/common/paging.c
mogasergiu: Good point, I did not think of much here I just updated it accordingly., this code was here before the memregion PR even I think.

lib/ukboot/boot.c
mogasergiu: I remember having tried this and was met with some errors. I always try to use already existing definitions. I might misremember, I will try again :+1: .

michpappas: It's `uk/plat/paging.h` that requires `CONFIG_PAGING`, but `uk/arch/paging.h` does not have such a requirement.

plat/common/memory.c
```diff
@@ -326,9 +379,32 @@ int ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
+		 * so that the kernel may use them (e.g. initrd start address).
+		 */
+		if (!del && ml_prio == MRD_PRIO_KRNL_RSRC)
+			ukplat_memregion_restore_mrd(ml, ml_opbase, ml_olen);
+
+		/* This here can only happen if two adjacent MRD_PRIO_KRNL_RSRC
+		 * regions are resolved without a deletion. Preserve mr's
+		 * original data as well.
+		 */
+		if (!del && mr_prio == MRD_PRIO_KRNL_RSRC)
```
michpappas: As discussed, it looks like we need a condition for ` != MRD_PRIO_FREE_RSRC`?

michpappas (approved):
Looks good thanks! 🚀 

mogasergiu:
@michpappas Added another commit to handle a case I have just encountered while working on something else. Let me know your thoughts. I am not sure if it would be good idea to use `unlikely` here though :thinking:, I probably should though.

razvand:
@michpappas , @zhxj9823 , could you take a final look, given @mogasergiu 's new commit?

michpappas:
> @michpappas , @zhxj9823 , could you take a final look, given @mogasergiu 's new commit?

@razvand the additional commit looks good, I think this one's good to merge now.

mogasergiu:
@michpappas Added two more commits :laughing: :ok_hand:

michpappas (changes requested):
@razvand please ignore my last comment and refrain from merging, as there are newly added commits.

plat/common/memory.c
```diff
@@ -217,22 +217,48 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
...
 		 */
 		} else {
+			/* len here is basically ml_end - mr_end. Thus,
+			 * len == 0 can happen only if:
+			 * - mr is at the end of the ml
+			 * - mr == ml, as mr can't be at the beginning
+			 * since ukplat_memregion_swap_if_unordered()
+			 * would have assigned mr as the actual ml.
+			 */
+			__sz len = ml->pbase + ml->len - mr->pbase - mr->len;
```
michpappas, line 222: I think you can simplify the logic a bit, as `RANGE_CONTAIN` already handles the case where the two regions are equal. And that's why it also works when `ml_prio > mr_prio`. Also in general the code would be easier to follow if the different cases are handled in an explicit and discrete way, ie:
```C
/* If the left memory region is contained in within the right region,
 * or the regions are equal, drop it entirely.
if (RANGE_CONTAIN()) {

/* If the left memory region is partially contained in the right region,
 * drop that part of the left region only.
 */
} else if () {

/* If the right region is contained within the left region,
 * keep the right region, and split the left region into two
 * surrounding parts.
 */
} else {

}
```
Together with documenting the guarantees you get from the caller regarding the relative position of the regions, then you have the code speak for itself and it doesn't require so much explanation in comments 😎

plat/common/memory.c
```diff
@@ -217,22 +217,48 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
 		 * a new one if the left region is larger than the right region.
 		 */
 		} else {
+			/* len here is basically ml_end - mr_end. Thus,
```
michpappas: This function heavily relies on assumptions about the regions (order, relative base / size, type). In this case it's a good practice to explicitly state these assumptions with comments accompanied with assertions, as it makes the function more robust, and also helps the reader understand it as a self-contained unit.

While some early-UART challenged architectures may have issues with such assertions, IMO opinion that should not prevent us from adding these assertions right now, because ultimately we cannot avoid dealing with fatal cases anyway and for that we use `UK_CRASH`, which also requires early UART. Moreover, we have plenty of calls to `UK_ASSERT` before UART is initialized in our code-base already, even in the memory subsystem. Finally, adding a fix to the x86 drivers to ignore early UART calls if early UART is not enabled - just how we do it on `arm64` - should be easy.

michpappas (changes requested):
Hi @mogasergiu, good stuff 💪🏼  Besides the comments below, the commit message initially describes the case where the regions are equal, but later also covers the case where the regions partially overlap. It then describes the pre-conditions at the very end. If I were you I would restructure that a bit to have the constraints in the beginning along with the description of the problem, followed by the solution, and also maybe leave the implementation details out.

michpappas (approved):
Looks good thanks!

razvand (approved):

PR #1061, opened 2023-08-21
plat/kvm: Use `AllocatePages()` to allocate memory regions in `UEFI`

mogasergiu:
According to the `UEFI` specifications, `AllocatePool()` is the safest allocation method, however there are two pitfalls:
- we don't know the BIOS's memory allocator or the system's memory map, so the resulted allocation start address may be out of the range that would be compatible with a Unikraft configuration that does not have `CONFIG_HAVE_PAGING`
- it does not guarantee page-aligned allocations and, even though we may be able to align the length or the end address, the allocation's start address may therefore not be page-aligned

Thus, replace such calls to `AllocatePool()` with calls to `AllocatePages()` to ensure page-aligned start addresses. Furthermore, preserve backwards compatibility with Unikraft configurations that do not have `CONFIG_HAVE_PAGING` by limiting the allocation's start address to be below the highest address of the static boot page tables through the help of `bpt_unmap_mrd`.

### Prerequisite checklist

### Base target

 - Architecture(s): [all]
 - Platform(s): [`kvm`]
 - Application(s): [all]

### DIFF
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -7,6 +7,16 @@
 #include <uk/arch/paging.h>
 #include <uk/plat/common/bootinfo.h>
 
+extern struct ukplat_memregion_desc bpt_unmap_mrd;
+static uk_efi_paddr_t uk_efi_alloc_max_paddr;
+
+/* We must ensure backwards compatibility with !CONFIG_HAVE_PAGING */
+#if CONFIG_HAVE_PAGING
+static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_ANY_PAGES;
+#else  /* !CONFIG_HAVE_PAGING */
+static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_MAX_ADDRESS;
+#endif /* !CONFIG_HAVE_PAGING */
+
 static struct uk_efi_runtime_services *uk_efi_rs;
 static struct uk_efi_boot_services *uk_efi_bs;
 static struct uk_efi_sys_tbl *uk_efi_st;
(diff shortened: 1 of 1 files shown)


plat/kvm/efi.c
```diff
@@ -208,8 +220,11 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 
 	/* Make sure the actual allocated buffer is bigger */
 	*map_sz += *desc_sz * UK_EFI_SURPLUS_MEM_DESC_COUNT;
-	status = uk_efi_bs->allocate_pool(UK_EFI_LOADER_DATA, *map_sz,
-					  (void **)map);
+	*map = (struct uk_efi_mem_desc *)uk_efi_alloc_max_paddr;
+	status = uk_efi_bs->allocate_pages(uk_efi_alloc_type,
+					   UK_EFI_LOADER_DATA,
+					   *map_sz / PAGE_SIZE + 1,
```
jschlumpp: Can't this use `ALIGN_UP`? (Same for the other occurrences)

mogasergiu: Yes, of course it can! Bruh, why did I go all big :brain: and decided to do division and  +1 :facepalm:

jschlumpp (approved):
~~I have no idea about UEFI, and wasn't able to get QEMU working with it. I assume that somebody else will test it.~~
Okay seems to work, but we should really increase the default memory region count. With the default memory region count the image does not boot on QEMU+OVMF.

eduardvintila (approved):
Looks very good to me. Thanks, @mogasergiu!

michpappas (approved):

PR #1075, opened 2023-08-29
lib/ramfs: Fix UAF bug when deleting open file

andreittr:
### Description of changes

Previously ramfs would eagerly free filesystem nodes when deleting a file regardless whether the file is open, leading to a possible use-after-free bug if the open file is later used. This change fixes this issue by marking nodes as deleted after they have been removed from the filesystem, then only releasing resources lazily once all open references have been garbage-collected.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Test with:
```c
#include <stdio.h>
#include <string.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdlib.h>

{
#define FN "tmpfile"

	int f;
	int r;
	void *buf;

	if ((f = open(FN, O_RDWR|O_CREAT, 0777)) < 0) {
		puts("Badf");
		return 1;
	}

	if ((r = unlink(FN))) {
		printf("Bad unlink %d\n", r);
		return 1;
	}

	buf = malloc(4000);
	printf("%p\n", buf);
	memset(buf, 0xff, 4000);

	if ((r = write(f, "asdf", 4)) < 0) {
		printf("Bad write %d\n", r);
		return 1;
	}

	if ((r = ftruncate(f, 0))) {
		printf("Bad truncate %d\n", r);
		return 1;
	}

	if ((r = close(f))) {
		printf("Bad close %d\n", r);
		return 1;
	}

	return 0;
}
```

### DIFF
--- a/lib/ramfs/ramfs_vnops.c
+++ b/lib/ramfs/ramfs_vnops.c
@@ -55,6 +55,12 @@
 #include <fcntl.h>
 #include <vfscore/fs.h>
 
+/* 16 bits are enough for file mode as defined by POSIX, the rest we can use */
+#define RAMFS_MODEMASK 0xffff
+#define RAMFS_DELMODE 0x10000
+#define RAMFS_IS_DELETED(np) ((np)->rn_mode & RAMFS_DELMODE)
+#define RAMFS_MARK_DELETED(np) (np)->rn_mode |= RAMFS_DELMODE
+
 static struct uk_mutex ramfs_lock = UK_MUTEX_INITIALIZER(ramfs_lock);
 static uint64_t inode_count = 1; /* inode 0 is reserved to root */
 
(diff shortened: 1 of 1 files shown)


lib/ramfs/ramfs_vnops.c
```diff
@@ -55,6 +55,12 @@
 #include <fcntl.h>
 #include <vfscore/fs.h>
 
+/* 16 bits are enough for file mode as defined by POSIX, rest we can use */
```
RaduNichita: It not very clear to me at what  "rest we can use". Could you be more explicit there, please?

RaduNichita (commented):
It works for me, I will approve after discussing my only comment.

lib/ramfs/ramfs_vnops.c
andreittr: I'll add a "the" for grammar pedantry.

DeliaPavel (approved):
Looks good to me!

RaduNichita (approved):
All good also from my side.

razvand (approved):

PR #1077, opened 2023-08-30
support/scripts: `checkpatch` `WARN` on commit summary > 70 chars

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Although `checkpatch.pl` checks for commit message lines to not be longer than `75` characters, in the case of `Github`'s UI, a commit summary longer than `70` character results in its line overflowing into the next.
E.g.
```
[Selector]/[Component]: Lorem ipsum dolor sit amet, consectetuer adipiscin
```
becomes
```
[Selector]/[Component]: Lorem ipsum dolor sit amet, consectetuer adip...
...iscin
```

Since this is not aesthetically pleasing, enforce a character limit for commit message summaries to not be longer than `70` characters.

### DIFF
--- a/support/scripts/checkpatch.pl
+++ b/support/scripts/checkpatch.pl
@@ -2548,6 +2548,25 @@ sub process {
 			     "Patch subject line does not follow Unikraft scheme: '[Selector]/[Component]: [Short message]'\n" . $herecurr);
 		}
 
+# Check that the first line in the commit message has less than
+# 70 characters to comply with Github UI. This helps
+# avoid commit message subjects placed at the beginning
+# to overflow into the second line when viewing through
+# the Github UI.
+		if ($in_header_lines &&
+		    $line =~ /^Subject:.*$/i) {
+			my $commit_subject_line0 = $1 if
+				$lines[$linenr - 1] =~ /^Subject: \[.*\] (.*)$/;
+			my $commit_subject_line1 = $lines[$linenr];
+			my $commit_subject = $commit_subject_line0 .
+					     $commit_subject_line1;
+			if (length($commit_subject) > 70) {
+				WARN("COMMIT_SUBJECT_LONG_LINE",
+				     "Possible unwrapped commit subject (prefer a maximum 70 chars)\n" . $commit_subject);
+				$commit_log_long_line = 1;
+			}
+		}
+
 # Check for old stable address
 		if ($line =~ /^\s*cc:\s*.*<?\bstable\@kernel\.org\b>?.*$/i) {
 			ERROR("STABLE_ADDRESS",


support/scripts/checkpatch.pl
```diff
@@ -2548,6 +2548,25 @@ sub process {
...
+# the Github UI.
+		if ($in_header_lines &&
+		    $line =~ /^Subject:.*$/i) {
+			my $commit_summary_line0 = $1 if
+				$lines[$linenr - 1] =~ /^Subject: \[.*\] (.*)$/;
+			my $commit_summary_line1 = $lines[$linenr];
+			my $commit_summary = $commit_summary_line0 .
+					     $commit_summary_line1;
+			if (length($commit_summary) > 70) {
+				WARN("COMMIT_LOG_LONG_LINE",
```
skuenzer: I think it would make sense to introduce a different switch for this. I'd call it `COMMIT_SUBJECT_LONG_LINE` since it is about the length of the subject line only.

support/scripts/checkpatch.pl
```diff
@@ -2548,6 +2548,25 @@ sub process {
...
+				     "Possible unwrapped commit summary (prefer a maximum 70 chars)\n" . $commit_summary);
```
skuenzer: What do you think of `summary` -> `subject`?

support/scripts/checkpatch.pl
StefanJum: Should we maybe have this as an error instead of warning? The GitHub interface messes up the commits pretty bad, so I would enforce this as much as possible.

StefanJum (commented):
Works fine, just one comment besides what @skuenzer already said.

support/scripts/checkpatch.pl
mogasergiu: I would keep `summary` so as not to create confusion between the actual `PATCH` subject and the commit subject. Or maybe you have other suggestions for a synonym? I do not have any preference here :smile:.

support/scripts/checkpatch.pl
mogasergiu: You have a good point :+1:. However, I would keep it aligned with that of the `75` characters, which is a `WARN` message :thinking:. But I really have no preference here. @skuenzer @razvand @michpappas I shall go with that the majority chooses here :raised_hands:.

support/scripts/checkpatch.pl
```diff
@@ -2548,6 +2548,25 @@ sub process {
 			     "Patch subject line does not follow Unikraft scheme: '[Selector]/[Component]: [Short message]'\n" . $herecurr);
 		}
 
+# Check that the first line in the commit message is below
```
RaduNichita:
```suggestion
# Check that the first line in the commit message has less than
```

RaduNichita (changes requested):
One small suggestion @mogasergiu

support/scripts/checkpatch.pl
michpappas: IMO as we should not strictly enforce it - ie block a PR - given that there could be legitimate reasons for the subject line to exceed that limit, it makes sense if it is a warning. On the other hand we don't really differentiate much between warnings and errors anyway (or do we)?

StefanJum (approved):
All good

support/scripts/checkpatch.pl
razvand: Agree with @michpappas . Keep this as a warning.

michpappas (approved):
This is a good change, as it aligns the output of `checkpatch` on this topic to our guidelines. Thanks @mogasergiu 🚀 

razvand (approved):

PR #1079, opened 2023-08-31
Make `elf64_to_32.py` also build the `multiboot` header and rename it to `multiboot.py`

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

Since `Multiboot` demands that its corresponding header be placed in the
first 8192 bytes and `GRUB` for whatever reason adds another restriction
requesting that the ELF Program Headers also have to be found in this
range (see GRUB commit `9a5c1ad`), offload the effort of building the
`Multiboot` header from the GAS directive defined header of `multiboot.S`
to `elf64_to_32.py`.

This will offer us more flexiblity and relieve us of the effort of
having anything ELF32 related inside the core codebase.

With this and since this script is only used for `multiboot` builds,
rename it accordingly, to reflect the fact the sole reason this script
exists is to cope with the demands of this ancient boot protocol.

The main reason this has to be done:
Apparently, the first ELF64 segment may also be placed beyond the first 8192,
even if there is no reason to do so. For example, we were met with a case where,
although the sum of the ELF header and ELF program headers was ~400 bytes,
the offset to the first segment would be padded to 8192, thus making the
`multiboot` signature end up being placed beyond the range in which
`multiboot` compliant bootloaders look for.


RaduNichita (approved):
Tested on helloworld and a few other apps and it works. Thanks, @mogasergiu 

support/scripts/multiboot.py
```diff
@@ -180,6 +207,11 @@ def main():
         # Write equivalent ELF32 header
         elf32.write(elf32_ehdr)
 
+        # Append the Multiboot header
+        # NOTE: Multiboot header must be 4-byte aligned - luckily that is the
```
jschlumpp, line 211: You could add an assert for this.

Also I think there should be an assertion that you don't accidentally overwrite data past the original ELF header.

mogasergiu: > Also I think there should be an assertion that you don't accidentally overwrite data past the original ELF header.

It supposedly does not exist until I write it, so there is nothing to overwrite :thinking:.

jschlumpp: But you are opening the original input file. In the beginning, there is an ELF64 header there. And then you write a ELF32 header there without potentially moving the "data" of the file backwards.

jschlumpp: Sorry, I misunderstood the script :+1:

jschlumpp (approved):

PR #1084, opened 2023-09-08
build: Always save architecture with savedefconfig

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

With `savedefconfig`, KConfig creates a file that contains only options that are different from the default selection. Because the build system automatically detects the host architecture and sets the default architecture to this detected architecture, the architecture is never part of the defconfig file if the architecture in the configuration matches the host architecture. This becomes problematic as soon as such a file is loaded on a build environment that has a different host architecture: The configuration can fail because of architectural configuration differences.

This PR ensures that the configured architecture is always stored in the defconfig file. The `savedefconfig` target appends the configured architecture to the file if the architectures between host and configuration match.

### DIFF
--- a/Config.uk
+++ b/Config.uk
@@ -3,18 +3,25 @@
 # see https://www.kernel.org/doc/Documentation/kbuild/Config.in-language.txt.
 #
 mainmenu "Unikraft/$(UK_FULLVERSION) Configuration"
+
 config UK_FULLVERSION
 	string
 	default "$(UK_FULLVERSION)"
+
 config UK_CODENAME
 	string
 	default "$(UK_CODENAME)"
+
 config UK_ARCH
 	string
 	default "x86_64" if ARCH_X86_64
 	default "arm64"  if ARCH_ARM_64
 	default "arm"	if ARCH_ARM_32
 	default "$(UK_ARCH)"
+
+config HOST_ARCH
+	string
+	default "$(HOST_ARCH)"
 
 config NO_APP
 	def_bool $(shell,test $(UK_BASE) = $(UK_APP) && echo y || echo n)
--- a/Makefile
+++ b/Makefile
@@ -872,13 +872,12 @@ DEFCONFIG = $(call qstrip,$(UK_DEFCONFIG))
 # We don't want to fully expand UK_DEFCONFIG here, so Kconfig will
 # recognize that if it's still at its default $(CONFIG_DIR)/defconfig
 COMMON_CONFIG_ENV = \
-	CC=$(CC)\
 	CONFIG_="CONFIG_" \
 	KCONFIG_CONFIG="$(UK_CONFIG)" \
 	KCONFIG_AUTOCONFIG="$(KCONFIG_AUTOCONFIG)" \
 	KCONFIG_AUTOHEADER="$(KCONFIG_AUTOHEADER)" \
 	KCONFIG_TRISTATE="$(KCONFIG_TRISTATE)" \
-	HOST_GCC_VERSION="$(HOSTCC_VERSION)" \
+	HOST_ARCH="$(HOSTARCH)" \
 	BUILD_DIR="$(BUILD_DIR)" \
 	UK_BASE="$(CONFIG_UK_BASE)" \
 	UK_APP="$(CONFIG_UK_APP)" \
--- a/Makefile
+++ b/Makefile
@@ -960,7 +960,6 @@ defconfig:
 
 savedefconfig:
 	@$(COMMON_CONFIG_ENV) $(KPYTHON) $(CONFIGLIB)/savedefconfig.py --kconfig $(CONFIG_CONFIG_IN) --out $(DEFCONFIG)
-	@$(COMMON_CONFIG_ENV) $(SCRIPTS_DIR)/configupdate $(UK_CONFIG) $(UK_CONFIG_OUT)
 
 oldconfig:
 	@$(COMMON_CONFIG_ENV) $(KPYTHON) $(CONFIGLIB)/oldconfig.py $(CONFIG_CONFIG_IN)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1035,8 +1034,6 @@ savedefconfig: $(KCONFIG_DIR)/conf
 	@$(COMMON_CONFIG_ENV) $< \
 		--savedefconfig=$(if $(DEFCONFIG),$(DEFCONFIG),$(CONFIG_DIR)/defconfig) \
 		$(CONFIG_CONFIG_IN)
-	@$(SED) '/UK_DEFCONFIG=/d' $(if $(DEFCONFIG),$(DEFCONFIG),$(CONFIG_DIR)/defconfig)
```
skuenzer, line 1038: This line is removed because it produces an unnecessary console output, it does not modify the resulting file. Additionally, `UK_DECONFIG=` was never stored to the `defconfig` file in my tests.

Config.uk
```diff
@@ -15,6 +15,9 @@ config UK_ARCH
 	default "arm64"  if ARCH_ARM_64
 	default "arm"	if ARCH_ARM_32
 	default "$(UK_ARCH)"
+config HOST_ARCH
```
michpappas, line 22: Nit: Missing newline before the new config directive.

Makefile
```diff
@@ -872,13 +872,12 @@ DEFCONFIG = $(call qstrip,$(UK_DEFCONFIG))
 # We don't want to fully expand UK_DEFCONFIG here, so Kconfig will
 # recognize that if it's still at its default $(CONFIG_DIR)/defconfig
 COMMON_CONFIG_ENV = \
-	CC=$(CC)\
```
michpappas, line 875: Why is this no longer needed?

arch/Makefile.rules
```diff
@@ -19,3 +19,10 @@ endef
 define aarch64_no_reserved_tcb_overlap =
 	$(eval LIBCONTEXT_CFLAGS-y+=-DCONFIG_AARCH64_NO_TCB_OVERLAP=1)
 endef
+
+# Converts an architecture string to the corresponding name of a configuration variable
+# Usage:
+#  $(call ukarch_str2cfg,"<architecture>")
+define ukarch_str2cfg =
+$(addprefix CONFIG_ARCH_,$(call uc,$(subst arm,ARM_32,$(subst arm64,ARM_64,$(1)))))
```
michpappas, line 36: This concerns me a bit wrt future architectures eg RISC, as it's essentially a bug lurking. I know it won't be as terse, but it would be more fail-safe if this used a controlled vocabulary for all known architectures and issue an error if an unknown arch passed. Perhaps it could be implemented as a separate macro that checks against known architectures.

michpappas (changes requested):
Hi @skuenzer, just a few comments below.

Makefile
skuenzer: It was already incorrect. The correct version would have to be `CC=$(HOSTCC)` because `HOSTCC` is the compiler intended for compiling host tools. I can check if `CC` is used by the sub build system for KConfig. If so, I'll replace this line with the corrected version.

arch/Makefile.rules
skuenzer: At the moment, external architectures are not supported, in this sense this won't break the status quo. Clearly it might need to get modified if we know how we support external architectures. However, the function is only handling the special cases of Arm where we have an extra `_` in the config option: `CONFIG_ARCH_ARM_32` although we call the architecture `arm32`. My intention was that this function is future-proof if we match the config name with the architecture name (like we do on x86): For example, if we name Risc-V `riscv`, then the matching architecture configuration option must be `CONFIG_ARCH_RISCV`.
What do you mean with separate macros?

Config.uk
skuenzer: Right, I'll add.

Makefile
skuenzer: Actually, the correct `CC` setting is already handed over when we invoke `Make` of KConfig, so this environment variable is overwritten and unnecessary to set with `COMMON_CONFIG_ENV`.

arch/Makefile.rules
michpappas: > At the moment, external architectures are not supported, in this sense this won't break the status quo. Clearly it might need to get modified if we know how we support external architectures

Yup, I was not referring to external architectures, as we already agreed that may not even be possible, but to new internal architectures.

> My intention was that this function is future-proof if we match the config name with the architecture name (like we do on x86

That expectation is exactly what I am concerned about. If the author of a new architecture is not be aware of this expectation (and how would they be?) then this feature will be broken on the new architecture. An explicit check, on the other hand, will reveal such an issue right away.

skuenzer: Hum... my concern is that a different expression which is extensible would be too complex for now and we also do not know yet how we will include external archs. Maybe there will be a more natural solution to the problem.
Example:
```Makefile
ukarch_str2cfg = $(addprefix CONFIG_ARCH_,$(CONFIG_ARCH_MAP_$(1)))
```
...which requires that each arch sets a special var:
```Makefile
CONFIG_ARCH_MAP_x86_64=X86_64
CONFIG_ARCH_MAP_arm32=ARM_32
CONFIG_ARCH_MAP_arm64=ARM_64
```
Additionally, code needs to be added to throw an error when the map variable is not defined.

However, I can comment what the function assumes so that this information helps as soon as we touch the code again.

michpappas: Works for me, thanks 👍🏼

michpappas (approved):
@skuenzer approving conditionally to adding that comment 😉 

Config.uk
skuenzer: Okay, I did not put it because it was part of a block without spaces. I am updating with this commit this particular block to add spaces in betweeen.

arch/Makefile.rules
skuenzer: 👍🏻

razvand:
@StefanJum , could you please review it and then I will approve it ASAP.

StefanJum (approved):
Works fine, thanks.

razvand (approved):

PR #1076, opened 2023-08-30
plat/virtio: Fix vring_avail_event macro

jschlumpp:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This fixes the `vring_avail_event` to not cause miscompilations by violating the strict aliasing rules.

PR #1089, opened 2023-09-12
drivers/virtio: Implement support for event index notification suppression

jschlumpp:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This mechanism is a more modern variant of the event disable bit and allows setting threshold until which interrupts are not necessary (Similar to the suppression using the bit it is only a hint that can be ignored). Firecracker only supports this index based mechanism and does not implement the bit-based variant.

Note: This introduces a dependency to virtio_bus in virtio_ring and I'm not sure whether we want that.

### DIFF
--- a/drivers/virtio/ring/include/virtio/virtio_ring.h
+++ b/drivers/virtio/ring/include/virtio/virtio_ring.h
@@ -169,7 +169,8 @@ static inline void vring_init(struct vring *vr, unsigned int num, uint8_t *p,
 	vr->avail = (struct vring_avail *) (p +
 			num * sizeof(struct vring_desc));
 	vr->used = (void *)
-	(((unsigned long) &vr->avail->ring[num] + align - 1) & ~(align - 1));
+		(((unsigned long) &vr->avail->ring[num] + sizeof(uint16_t) +
+			align - 1) & ~(align - 1));
 }
 
 static inline unsigned int vring_size(unsigned int num, unsigned long align)
@@ -188,7 +189,8 @@ static inline unsigned int vring_size(unsigned int num, unsigned long align)
 static inline int vring_need_event(__u16 event_idx, __u16 new_idx,
 				   __u16 old_idx)
 {
-	return (new_idx - event_idx - 1) < (new_idx - old_idx);
+	return (uint16_t)(new_idx - event_idx - 1) <
+		(uint16_t)(new_idx - old_idx);
 }
 
 #ifdef __cplusplus
--- a/drivers/virtio/ring/virtio_ring.c
+++ b/drivers/virtio/ring/virtio_ring.c
@@ -43,6 +43,7 @@
 #include <uk/plat/io.h>
 #include <virtio/virtio_ring.h>
 #include <virtio/virtqueue.h>
+#include <virtio/virtio_bus.h>
 #ifdef CONFIG_LIBUKVMEM
 #include <uk/arch/paging.h>
 #include <uk/plat/paging.h>
@@ -101,6 +102,12 @@ void virtqueue_intr_disable(struct virtqueue *vq)
 	UK_ASSERT(vq);
 
 	vrq = to_virtqueue_vring(vq);
+
+	if (vq->uses_event_idx) {
+		vring_used_event(&vrq->vring) =
+			vrq->last_used_desc_idx - vrq->vring.num - 1;
+		return;
+	}
 	vrq->vring.avail->flags |= (VRING_AVAIL_F_NO_INTERRUPT);
 }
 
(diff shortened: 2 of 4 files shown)


drivers/virtio/ring/virtio_ring.c
```diff
@@ -43,6 +43,7 @@
 #include <uk/plat/io.h>
 #include <virtio/virtio_ring.h>
 #include <virtio/virtqueue.h>
+#include <virtio/virtio_bus.h>
```
andreittr, line 46: Where do we actually use the declarations from `virtio_bus`?
I'm likely suffering a brainfart, but I can't find any in the PR diff.

michpappas: These are included from `Makefile.uk`:
```
# common virtio headers
LIBVIRTIO_RING_CINCLUDES-y  += -I$(UK_DRIV_LIBVIRTIO_BASE)/include
LIBVIRTIO_RING_CINCLUDES-y  += -I$(LIBVIRTIO_BUS_BASE)/include
LIBVIRTIO_RING_CINCLUDES-y  += -I$(LIBVIRTIO_RING_BASE)/include
```

jschlumpp: It's the feature defines (`VIRTIO_FEATURE_HAS`, etc.)
The header is possibly not the best place for them but they were moved into the bus library during the re-arch move.

andreittr: Re: bringing virtio_bus in as a dependency: it makes sense if it provides core virtio functions that ring relies on. However, if the reasons are only `VIRTIO_FEATURE_*`, which are macros for checking/setting/clearing bits, I think we should reconsider.

IMO, due to their very basic nature, `VIRTIO_FEATURE_*` macros would better live somewhere common, like `uk/virtio_types.h`, unless we can argue that the way they do bit ops is specific to virtio_bus and not applicable anywhere else.

edit: race condition w/ Marco's reply, corrected

drivers/virtio/ring/include/virtio/virtio_ring.h
```diff
@@ -157,7 +157,7 @@ struct vring {
  * versa. They are at the end for backwards compatibility.
  */
 #define vring_used_event(vr) ((vr)->avail->ring[(vr)->num])
-#define vring_avail_event(vr) (*(__virtio16 *)&(vr)->used->ring[(vr)->num])
+#define vring_avail_event(vr) (*(__virtio_le16 *)&(vr)->used->ring[(vr)->num])
```
michpappas: This is fixed in #1076

drivers/virtio/ring/virtio_ring.c
michpappas: Yes these headers were moved as part of re-arch verbatim, as in general re-arch avoids refactoring to limit the scope of work. And since the different virtio components were split into individual libraries `virtio_bus.h` naturally became part of `libvirtio_bus`.

From what I see the reason `VIRTIO_FEATURE_*` are defined in `virtio_bus.h` is likely because the same header defines `virtio_confg_ops` (which among others also include `features_get` / `features_set`) and that is effectively part of `struct virtio_dev` which is defined there too.

On the other hand there is also `virtio_config.h`  as part of the common headers, so that could possible be a better place?

PS: Sorry for the previous hasty and irrelevant answer 🙏🏼

drivers/virtio/ring/virtio_ring.c
```diff
@@ -186,9 +202,21 @@ static inline void virtqueue_detach_desc(struct virtqueue_vring *vrq,
...
 
 	UK_ASSERT(vq);
 	vrq = to_virtqueue_vring(vq);
+	if (vq->uses_event_idx) {
+		new = vrq->vring.avail->idx;
+		/* TODO: Use the actually submitted count instead of assuming
+		 *       that a single descriptor was submitted. This would be
+		 *       more efficient when we would have batching implemented.
+		 */
+		old = new - 1; /* FIXME: Use actual written count */
```
michpappas: @mschlumpp just to confirm, is this something that is still planned to be addressed later?

jschlumpp: I'll remove the `FIXME` comment and leave the `TODO` for now. In the future we can implement a batching mechanism, which we can use to implement this more efficiently.

drivers/virtio/ring/include/virtio/virtio_ring.h
jschlumpp: The rebase will remove this diff.

drivers/virtio/ring/virtio_ring.c
jschlumpp: Users of the feature test macros also have to access the `features` field of `virtio_device`. So they have to rely on the header.

drivers/virtio/ring/virtio_ring.c
michpappas: Okee 👍🏼

michpappas (approved):
LGTM, thanks

drivers/virtio/ring/Makefile.uk
```diff
@@ -4,6 +4,7 @@ $(eval $(call addlib_s,libvirtio_ring,$(CONFIG_LIBVIRTIO_RING)))
 LIBVIRTIO_RING_CINCLUDES-y  += -I$(UK_DRIV_LIBVIRTIO_BASE)/include
 LIBVIRTIO_RING_CINCLUDES-y  += -I$(LIBVIRTIO_BUS_BASE)/include
 LIBVIRTIO_RING_CINCLUDES-y  += -I$(LIBVIRTIO_RING_BASE)/include
+LIBVIRTIO_RING_CINCLUDES-y  += -I$(LIBVIRTIO_BUS_BASE)/include
```
andreittr: I believe this is a duplicate of line 5, might've slipped in through some rebase; we should remove it.

jschlumpp: I just checked: It got added in PR #1095. It's not very clear why that PR also added the include.

andreittr (approved):
Looks good, thanks!

razvand (approved):

PR #1093, opened 2023-09-14
lib/vfscore: Fix pipe error codes

andreittr:
### Description of changes

This change fixes incorrect return of error codes in the pipe module of vfscore. Specifically:
- Return positive error codes from internal functions; syscall handlers expect positive errors and will negate them as required
- Do not set errno, rather return the error code
- Correctly return EINVAL when attempting to open a pipe with O_DIRECT, as we do not support packet mode pipes, and calling code needs to be made aware of this (see `pipe2(2)`)

Checkpatch has no sodding clue about (the very messy) error conventions in our codebase; likely a false positive inherited from Linux, will probably fix at the source soonish.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A

### DIFF
--- a/lib/vfscore/pipe.c
+++ b/lib/vfscore/pipe.c
@@ -309,7 +309,7 @@ static int pipe_write(struct vnode *vnode,
 
 	if (!pipe_file->r_refcount) {
 		/* TODO before returning the error, send a SIGPIPE signal */
-		return -EPIPE;
+		return EPIPE;
 	}
 
 	uk_mutex_lock(&pipe_buf->wrlock);
@@ -447,8 +447,7 @@ static int pipe_seek(struct vnode *vnode __unused,
 			struct vfscore_file *vfscore_file __unused,
 			off_t off1 __unused, off_t off2 __unused)
 {
-	errno = ESPIPE;
-	return -1;
+	return ESPIPE;
 }
 
 static int pipe_ioctl(struct vnode *vnode,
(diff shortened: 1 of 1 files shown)


StefanJum (approved):
All good, thanks.

RaduNichita:
All good from my side.

lib/vfscore/pipe.c
```diff
@@ -699,6 +698,9 @@ UK_SYSCALL_R_DEFINE(int, pipe2, int*, pipefd, int, flags)
 {
 	int rc;
 
+	if (flags & O_DIRECT)
+		return -EINVAL;
```
RaduNichita, line 702: Here why are we returning the negative value?

RaduNichita (changes requested):

lib/vfscore/pipe.c
andreittr: If you mean why negative, it's because we're in the context of a raw syscall, which by API convention returns negative error values.

If you mean why error out on this flag, it's because an `O_DIRECT` pipe behaves differently to a regular pipe on Linux, operating in "packet mode" -- essentally like a datagram socket where writes match up with reads 1-to-1. Unikraft sockets don't implement such a mode, and for applications that rely on this behavior it's better to error out early than risk weird bugs.

RaduNichita: All good now, thanks.

RaduNichita (approved):
All good now, thanks.

razvand (approved):

PR #1095, opened 2023-09-15
drivers: Add PLAT_COMMON_BASE to the include path

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The `PCI` and `virtio` have dependencies on headers under `uk/plat/common`. Add `PLAT_COMMON_BASE` to the include path to prevent compile errors. Moreover, the `virtio-ring` driver includes `virtio/virtio-bus.h`. Add `LIBVIRTIO_BUS_BASE` to this driver library's include path.

PR #1103, opened 2023-09-18
lib/ukintctlr: Introduce uk_intctlr API - Re-arch IRQ handling

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR removes all interrupt-related code from `plat/`, and moves it into `lib/ukintctlr/` and `drivers/ukintctlr`, as needed. Both new and existing functionality is collected under `libukintctlr` into a single, newly introduced `uk_intctlr` API. Among others, that also includes unified interrupt handling. 

The changes are summarized as:
* All interrupt management, registration, and handling is moved to `lib/ukintctlr`.
* The GIC driver is updated to implement the new `uk_intctlr` API.
* The x86 IRQ controller drivers (PIC, x2APIC) are moved to the `drivers/ukintctlr` subsystem into a new driver `xPIC` and implement the `uk_intctlr` API.
* Explicit IRQ acknowledgement is removed from individual handlers and moved into the unified handler of `lib/ukintctlr`.
* Limits such as `MAX_IRQ` are defined by the individual drivers.
* New features like unregistering IRQ handlers, and IRQ allocation are added into the `uk_intctlr` API (@mschlumpp) 
* All platforms, libraries, and drivers are adapted to the changes introduced by the `uk_intctlr` API.

To minimize the scope of this work, changes to the implementation has been limited to the smallest possible.

Given the nature of this work, commits don't always follow the one commit per library convention. Instead I have tried to handle changes on a per-case basis, and pick what I found to be the most suitable pattern for each case.

GitHub-Depends-On: #1095


razvand:
@michpappas , this fails on 9pfs-based builds:

```
  LDS     libkvmplat: bootinfo.lds
  LDS     libkvmplat: libparam.lds
  CC      lib9pfs: 9pfs_vfsops.o
  CC      lib9pfs: 9pfs_vnops.o
  LDS     lib9pfs: libparam.lds
  CC      libdevfs: device.o
In file included from /home/unikraft/maintainer-tools/workdir/unikraft/lib/uk9p/include/uk/9p.h:40,
                 from /home/unikraft/maintainer-tools/workdir/unikraft/lib/9pfs/9pfs_vfsops.c:37:
/home/unikraft/maintainer-tools/workdir/unikraft/lib/uk9p/include/uk/9pdev.h:43:10: fatal error: uk/plat/irq.h: No such file or directory
   43 | #include <uk/plat/irq.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [/home/unikraft/maintainer-tools/workdir/unikraft/support/build/Makefile.build:27: /home/unikraft/maintainer-tools/workdir/apps/redis/build/lib9pfs/9pfs_vfsops.o] Error 1
make[3]: *** Waiting for unfinished jobs....
In file included from /home/unikraft/maintainer-tools/workdir/unikraft/lib/uk9p/include/uk/9p.h:40,
                 from /home/unikraft/maintainer-tools/workdir/unikraft/lib/9pfs/9pfs_vnops.c:40:
/home/unikraft/maintainer-tools/workdir/unikraft/lib/uk9p/include/uk/9pdev.h:43:10: fatal error: uk/plat/irq.h: No such file or directory
   43 | #include <uk/plat/irq.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [/home/unikraft/maintainer-tools/workdir/unikraft/support/build/Makefile.build:27: /home/unikraft/maintainer-tools/workdir/apps/redis/build/lib9pfs/9pfs_vnops.o] Error 1
```

razvand:
Thanks, @michpappas , this works now.

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -1,42 +1,187 @@
...
+
+/**
+ * This system's interrupt controller
+ *
+ * Populated by uk_intctlr_probe()
+ */
+extern struct uk_intctlr_desc *uk_intctlr;
+
+/** The event payload for the #UK_INTCTLR_EVENT_IRQ event */
+struct uk_intctlr_event_irq_data {
```
jschlumpp: I think it would be easier to understand at a glance if this structure would be next to the `UK_INTCTLR_EVENT_IRQ` define.

lib/ukintctlr/ukintctlr.c
```diff
@@ -29,43 +29,45 @@
...
-#include <uk/plat/common/irq.h>
 #include <uk/intctlr.h>
 #include <uk/assert.h>
 #include <uk/event.h>
 #include <uk/trace.h>
 #include <uk/print.h>
 #include <errno.h>
 #include <uk/bitops.h>
-#ifdef CONFIG_UKPLAT_ISR_ECTX_ASSERTIONS
+#ifdef CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS
```
mogasergiu: With this occasion, probably a good idea to also change this to `#if`

lib/ukintctlr/ukintctlr.c
```diff
@@ -96,13 +99,13 @@ int ukplat_irq_register(unsigned long irq, irq_handler_func_t func, void *arg)
...
-void _ukplat_irq_handle(struct __regs *regs, unsigned long irq)
+void uk_intctlr_irq_handle(struct __regs *regs, unsigned int irq)
 {
 	struct irq_handler *h;
 	int i;
 	int rc;
-	struct ukplat_event_irq_data ctx;
-#ifdef CONFIG_UKPLAT_ISR_ECTX_ASSERTIONS
+	struct uk_intctlr_event_irq_data ctx;
+#ifdef CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS
```
mogasergiu: Ditto

lib/ukintctlr/ukintctlr.c
```diff
@@ -151,19 +154,29 @@ void _ukplat_irq_handle(struct __regs *regs, unsigned long irq)
 	 * devices, and (2) to minimize impact on drivers that share one
 	 * interrupt line that would then stay disabled.
 	 */
-	trace_plat_kvm_unhandled_irq(irq);
+	trace_uk_intctlr_unhandled_irq(irq);
 
 exit_ack:
-#ifdef CONFIG_UKPLAT_ISR_ECTX_ASSERTIONS
+#ifdef CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS
```
mogasergiu: Ditto

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -43,9 +43,9 @@
 #include <uk/plat/common/acpi.h>
 #endif /* CONFIG_UKPLAT_ACPI */
 #include <uk/plat/common/bootinfo.h>
```
mogasergiu, line 45: In `drivers/ukintctlr/gic: Adapt the GIC driver to the new uk_intctlr API` commit you repeated yourself with the sentence
```
Rename ops->set_irq_type to ops->set_irq_trigger, as the former is too vague.
```

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+#include <errno.h>
+#include <uk/config.h>
+#include <uk/assert.h>
+#include <uk/intctlr.h>
+#include <uk/intctlr/gic-v2.h>
+#include <uk/intctlr/gic-v3.h>
+
+struct _gic_dev *gic;
+struct uk_intctlr_desc intctlr;
+struct uk_intctlr_driver_ops ops;
```
mogasergiu, line 15: Reverse xmas tree declarations? :thinking:

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+#define FDT_GIC_IRQ_FLAGS_TL_EDGE_LO	2   /* edge trigerred, high-to-low */
+#define FDT_GIC_IRQ_FLAGS_TL_LEVEL_HI	4   /* level sensitive, active high */
+#define FDT_GIC_IRQ_FLAGS_TL_LEVEL_LO	8   /* level sensitive, active low */
+
+int uk_intctlr_fdt_get_irq(const void *fdt, int nodeoffset, __u32 index,
+			   struct uk_intctlr_irq *irq)
+{
+	int rc, size;
+	fdt32_t *prop;
+	__u32 type, irq_num, flags;
```
mogasergiu, line 33: Reverse xmas tree :parrot:

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+{
+	if (irq->trigger != UK_INTCTLR_IRQ_TRIGGER_NONE)
+		gic->ops.set_irq_trigger(irq->id, irq->trigger);
+}
+
+int uk_intctlr_probe(void)
+{
+	int rc = -ENODEV;
+
+#ifdef CONFIG_LIBUKINTCTLR_GICV2
```
mogasergiu: `#if`

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+	/* First, try GICv2 */
+	rc = gicv2_probe(&gic);
+	if (rc == 0) {
+		intctlr.name = "GICv2";
+		rc = gic->ops.initialize();
+		goto init;
+	}
+#endif /* LIBUKINTCTLR_GICV2 */
+
+#ifdef CONFIG_LIBUKINTCTLR_GICV3
```
mogasergiu: `#if`

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+
+#ifdef CONFIG_LIBUKINTCTLR_GICV2
+	/* First, try GICv2 */
+	rc = gicv2_probe(&gic);
+	if (rc == 0) {
+		intctlr.name = "GICv2";
+		rc = gic->ops.initialize();
+		goto init;
+	}
+#endif /* LIBUKINTCTLR_GICV2 */
```
mogasergiu: Add `CONFIG_` prefix

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+
+#ifdef CONFIG_LIBUKINTCTLR_GICV3
+	/* GICv2 is not present, try GICv3 */
+	rc = gicv3_probe(&gic);
+	if (rc == 0) {
+		intctlr.name = "GICv3";
+		rc = gic->ops.initialize();
+		goto init;
+	}
+#endif /* LIBUKINTCTLR_GICV3 */
```
mogasergiu: Add `CONFIG_` prefix

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -0,0 +1,119 @@
...
+		rc = gic->ops.initialize();
+		goto init;
+	}
+#endif /* LIBUKINTCTLR_GICV2 */
+
+#ifdef CONFIG_LIBUKINTCTLR_GICV3
+	/* GICv2 is not present, try GICv3 */
+	rc = gicv3_probe(&gic);
+	if (rc == 0) {
+		intctlr.name = "GICv3";
```
mogasergiu, line 104: We should probably make this field of the struct optional/configurable, just like `UKPLAT_MEMRNAME`

drivers/ukintctlr/Makefile.uk
```diff
@@ -6,4 +6,5 @@
 
 UK_DRIV_INTCTLR_BASE := $(UK_DRIV_BASE)/ukintctlr
 
+$(eval $(call _import_lib,$(UK_DRIV_INTCTLR_BASE)/xpic))
```
mogasergiu, line 9: Actually I am not sure about this :thinking:. There is also `OMPIC` (Open Multi-Processor Interrupt Controller) and for `GIC` as well there is the `MIPS-GIC` (Global Interrupt Controller). I know we may never support these arch's and won;t end up having clashing names but still does this not sound too vague :thinking:? And I guess we can't really just place them into `arch/` either. Maybe we could use more explicit names instead? E.g. an `i8259` directory for `PIC`, an `apic` directory for `APIC` (`x2apic` will be part of the `apic` driver as well, as it's not a different interrupt controller), a `gic-v2` directory for `GIC-v2` and same for `gic-v3`?

lib/ukofw/Makefile.uk
```diff
@@ -5,4 +5,3 @@ CINCLUDES-$(CONFIG_LIBUKOFW)   += -I$(LIBUKOFW_BASE)/include
 CXXINCLUDES-$(CONFIG_LIBUKOFW) += -I$(LIBUKOFW_BASE)/include
 
 LIBUKOFW_SRCS-y                += $(LIBUKOFW_BASE)/fdt.c
-LIBUKOFW_SRCS-y                += $(LIBUKOFW_BASE)/gic_fdt.c
```
mogasergiu, line 8: Maybe in the future with `libfwif` we could just have a generic firmware node ops method equivalent to that of `of_xlate` from Linux/U-Boot :open_mouth:. The first time I saw this function some long time ago I wrongly assumed it had nothing to do with the GIC and could have been a more generic function instead. I clearly did not read it thoroughly enough. :exploding_head:

lib/ukintctlr/ukintctlr.c
```diff
@@ -40,6 +41,7 @@
 #include <uk/arch/ctx.h>
 #endif
 
+#define MAX_IRQ				UK_INTCTLR_MAX_IRQ
```
mogasergiu, line 52: We should probably add an `#error` here in case it is not defined? :thinking:

plat/kvm/x86/traps.c
```diff
@@ -234,6 +235,218 @@ void traps_table_init(void)
 	FILL_IRQ_GATE(14, 1);
 	FILL_IRQ_GATE(15, 1);
 
+#ifdef CONFIG_KVM_X86_FULL_IDT
```
mogasergiu: `#if` :cockroach:

plat/kvm/x86/traps.c
```diff
@@ -234,6 +235,218 @@ void traps_table_init(void)
...
+	FILL_IRQ_GATE(215, 1);
+	FILL_IRQ_GATE(216, 1);
+	FILL_IRQ_GATE(217, 1);
+	FILL_IRQ_GATE(218, 1);
+	FILL_IRQ_GATE(219, 1);
+	FILL_IRQ_GATE(220, 1);
+	FILL_IRQ_GATE(221, 1);
+	FILL_IRQ_GATE(222, 1);
+	FILL_IRQ_GATE(223, 1);
+#endif
```
mogasergiu: `/* CONFIG_KVM_X86_FULL_IDT */`

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
```diff
@@ -6,6 +6,17 @@
...
 #define __UK_INTCTLR_XPIC_LIMITS_H__
 
-#define UK_INTCTLR_MAX_IRQ                             16
+#define UK_INTCTLR_FIRST_ALLOCABLE_IRQ			16
+
+#if CONFIG_UKINTCTLR_XPIC_FULL_IDT
+#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			223
+#define UK_INTCTLR_MAX_IRQ				(256 - 32)
+#else /* CONFIG_UKINTCTLR_XPIC_FULL_IDT */
+#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			16
```
mogasergiu: Probably should have been `15` here.

lib/ukintctlr/ukintctlr.c
```diff
@@ -44,13 +45,19 @@
 #define MAX_IRQ				UK_INTCTLR_MAX_IRQ
 #define MAX_HANDLERS_PER_IRQ		CONFIG_LIBUKINTCTLR_MAX_HANDLERS_PER_IRQ
 
+#define ALLOCABLE_IRQ_COUNT		UK_INTCTLR_ALLOCABLE_IRQ_COUNT
```
mogasergiu, line 55: Why not use `ALLOCATABLE` instead of `ALLOCABLE` everywhere? :thinking:

lib/ukintctlr/ukintctlr.c
```diff
@@ -203,6 +210,40 @@ void uk_intctlr_irq_handle(struct __regs *regs, unsigned int irq)
...
+					      ALLOCABLE_IRQ_COUNT, 0,
+					      count, 0);
+	if (start == ALLOCABLE_IRQ_COUNT)
+		return -ENOSPC;
+
+	uk_bitmap_set(irqs_allocated, start, count);
+	for (idx = start; idx < (start + count); idx++) {
+		*irqs = idx + FIRST_ALLOCABLE_IRQ;
+		irqs++;
+	}
```
mogasergiu, line 274: Newline here for consistency? As `return 0;` is not part of the same logic block of the loop above, right?

drivers/virtio/mmio/virtio_mmio.c
```diff
@@ -395,45 +395,38 @@ static struct virtio_config_ops virtio_mmio_config_ops = {
...
 static int virtio_mmio_probe(struct pf_device *pfdev)
 {
+	int rc;
 	const fdt32_t *prop;
-	int type, hwirq, prop_len;
-	int fdt_vm = pfdev->fdt_offset;
-	__u64 reg_base;
-	__u64 reg_size;
+	int prop_len, offs;
+	__u64 reg_base, reg_size;
```
mogasergiu, line 401: Reverse xmas tree with this opportunity?

plat/common/arm/time.c
```diff
@@ -110,9 +107,8 @@ __nsec ukplat_time_get_ticks(void)
 /* must be called before interrupts are enabled */
 void ukplat_time_init(void)
 {
-	int rc, irq, fdt_timer;
-	uint32_t irq_type, hwirq;
-	uint32_t trigger_type;
+	int rc, offs;
+	struct uk_intctlr_irq irq;
```
mogasergiu, line 111: Reverse xmas tree?

plat/drivers/rtc/pl031.c
```diff
@@ -155,41 +153,37 @@ void pl031_clear_intr(void)
...
+	return uk_intctlr_irq_register(pl031_irq, handler, NULL);
 }
 
 int pl031_init_rtc(void *dtb)
 {
 	__u64 size;
-	__u32 irq_type, hwirq, trigger_type;
-	int fdt_rtc, rc;
+	int offs, rc;
+	struct uk_intctlr_irq irq;
```
mogasergiu, line 163: Ditto

plat/Config.uk
```diff
@@ -39,7 +39,7 @@ config UKPLAT_LCPU_MAXCOUNT
 config HAVE_SMP
```
mogasergiu, line 39: Commit `plat/common: Adapt common code to the new uk_intctlr API` has no commit message, just something short would suffice, but let's try to not leave it empty. I assume you will remove `Notice: Picking individual commits in this PR will break the build.` before merge, right?

plat/kvm/arm/setup.c
```diff
@@ -231,6 +229,11 @@ void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
 	if (unlikely(rc < 0))
 		UK_CRASH("Failed to get PSCI method: %d.\n", rc);
 
+#ifdef CONFIG_RTC_PL031
```
mogasergiu: `#if`

plat/kvm/Config.uk
```diff
@@ -10,7 +10,9 @@ menuconfig PLAT_KVM
 	select ELF64_TO_32 if ARCH_X86_64
 	select LIBUKRELOC if OPTIMIZE_PIE
 	select LIBUKOFW if ARCH_ARM_64
-	select LIBUKINTCTLR if ARCH_X86_64
+	select HAVE_INTCTLR
```
mogasergiu, line 13: Commit `plat/kvm: Adapt KVM to the new uk_intctlr API` has no commit message.

mogasergiu (changes requested):
Awesome work! About time this happened :tada:! Had a quick look to get a general understanding of the changes as a first round of review. Left a couple of minor comments. I am going to have another more thorough look by the end of today to catch potential bugs and whatnot.

Thank you for bringing these changes :pray:

lib/ukintctlr/include/uk/intctlr.h
```diff
@@ -1,42 +1,187 @@
...
+ * @param irqs pointer to array of irqs
+ * @param sz   number of array elements
+ * @return zero on success, or negative value on error
+ */
+int uk_intctlr_irq_free(unsigned int *irqs, __sz count);
+
+/**
+ * Parse `interrupts` fdt node into IRQ descriptor
+ *
+ * This function must be implemented by the driver
```
jschlumpp: Is fdt relevant for x86? Maybe we can make that function optional.

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
```diff
@@ -86,15 +86,16 @@ struct _gic_operations {
...
+				enum uk_intctlr_irq_trigger trigger);
 	/** Set priority for IRQ */
 	void (*set_irq_prio)(uint32_t irq, uint8_t priority);
 	/** Set IRQ affinity (or "target" for GICv2) */
 	void (*set_irq_affinity)(uint32_t irq, uint32_t affinity);
 	/** Translate to absolute IRQ according to type e.g. PPI SPI SGI */
 	uint32_t (*irq_translate)(uint32_t type, uint32_t irq);
 	/** Handle IRQ */
-	void (*handle_irq)(struct __regs *regs);
+	void (*handle_irq)(struct __regs *regs, uint32_t __unused irq);
```
jschlumpp: Is there a reason to introduce the parameter, if it's not used for both GIC implementations?

drivers/ukintctlr/Makefile.uk
jschlumpp: I think the idea is that the driver handles PIC + x2apic currently.

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
jschlumpp: Hmm. When I wrote the allocation code I had a "past-end" meaning in mind. So the range would be [first, last). That way the IRQ_COUNT can be calculated as `last - first`. Maybe using `end` instead of `last` would have been clearer.

drivers/virtio/mmio/virtio_mmio.c
```diff
@@ -395,45 +395,38 @@ static struct virtio_config_ops virtio_mmio_config_ops = {
...
-			goto error_exit;
-		}
+	rc = uk_intctlr_fdt_get_irq(dtb, offs, 0, &irq);
+	if (unlikely(rc < 0))
+		return -EINVAL;
 
-		/* only care about base addr, ignore the size */
-		fdt_get_address(dtb, fdt_vm, 0, &reg_base, &reg_size);
-	}
+	uk_intctlr->ops->configure_irq(&irq);
```
jschlumpp: I would introduce a new function instead of having to access a global variable. Also I think it's sensible to have that function return an error code in case configuring the interrupt is not possible.

drivers/ukintctlr/gic/Config.uk
```diff
@@ -4,10 +4,10 @@ config LIBUKINTCTLR_GIC
 
 config LIBUKINTCTLR_GICV2
 	bool "Arm Generic Interrupt Controller (GICv2)"
-	depends on ARCH_ARM_64
+	depends on HAVE_INTCTLR
```
jschlumpp: This change allows selecting the GIC on x86, which I think is not very useful.

arch/x86/x86_64/include/uk/asm/arch.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#ifndef __UK_ARCH_H__
```
jschlumpp: I think that header guard is a bit too ambiguous and might cause problems.

arch/x86/x86_64/include/uk/asm/pic.h
jschlumpp, line 1: I would add either a header guard to this file or a check that this file is only included from the `arch.h` header.

drivers/ukintctlr/Makefile.uk
mogasergiu: That I understood, what I am saying is that it looks vague and IMO every irq controller driver should maybe have its own directory with its corresponding C/header source files. This way you could also have more specific directories. I find `xpic` too vague.

drivers/ukintctlr/gic/ukintctlr.c
michpappas: Ho ho ho 🎅🏼 I take this as a bit arbitrary, as we have never discussed this in the scope of the coding conventions. Also I personally find the regular xmas tree easier to read, and I think it's also what is used most of the time in our code base, so I'd leave it as is for now and we can discuss it on the maintainer's call to see what others think.

drivers/ukintctlr/gic/ukintctlr.c
michpappas: Good point. I think the criterion would be why did it make it configurable in `UKPLAT_MEMRENAME`, and whether that applies here too. For context, the reason I placed it there is because I thought it would be useful when printing messages from the generic API, so perhaps that should be taken into account too.

drivers/ukintctlr/Makefile.uk
michpappas: @mogasergiu, @mschlumpp I have no opinion about the name, especially given that I don't maintain this driver. I only picked it because I assumed that PIC, APIC, and x2APIC would have inter-dependencies and would likely live under the same driver. If you can think of something more suitable, I would be more than happy to update 🙃

lib/ukofw/Makefile.uk
michpappas: > Maybe in the future with libfwif we could just have a generic firmware node ops method equivalent to that of of_xlate from Linux/U-Boot

I don't follow, can you please elaborate? Actually I thought for a bit whether there should be a centralized place for such helpers or whether each driver should provide its own when they are not resuable, and concluded on the latter. Notice that the interpretation of the `interrupts` property is interrupt controller specific, ie for GIC it is defined in https://www.kernel.org/doc/Documentation/devicetree/bindings/interrupt-controller/arm%2Cgic.yaml, which means that in theory another architecture that uses fdt may provide a different interpretation using a different number of `interrrupt-cells` and so on.

lib/ukintctlr/ukintctlr.c
michpappas: Good point, will do 👍🏼

drivers/ukintctlr/gic/ukintctlr.c
mogasergiu: > when printing messages from the generic API,

Probably this is why it was made configurable, at least this is what I think. If you want a highly optimized build, every unnecessarily wasted byte matters, right? Probably also something to be discussed in the maintainer's call? :call_me_hand:.

lib/ukofw/Makefile.uk
mogasergiu: Yes, I understood. For such cases U-Boot, for example, allows the registration of an `of_xlate` method that does this device specific translation, e.g. for [clocks](https://github.com/u-boot/u-boot/blob/master/include/clk-uclass.h#L32), as depending on the clock in the clock-tree it may interpret the size/cells differently.

I mentioned this as a potential method to be registered in what would be `libukfwif` :wink: .

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
mogasergiu: Well then why is this not aligned with what is above?
```C
#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			223 /* 224 - 1 */
#define UK_INTCTLR_MAX_IRQ				224
```
Am I missing something? :thinking:

drivers/ukintctlr/gic/ukintctlr.c
michpappas: Well... I find it a bit of an overkill to add a config option just for a single value, I'd rather remove it altogether instead. Let's see what @mschlumpp thinks.

jschlumpp: I also wouldn't make that configurable as there will be usually only one instance of that string in the binary.
In the case of memory regions there can be many more instances so it has some (very small) impact on the memory consumption.

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
jschlumpp: Changing it to `15` would break the subtraction in `IRQ_COUNT`: `LAST - FIRST = 15 - 16 = -1`.

mogasergiu: Is this not the case for
```C
#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			223 /* 224 - 1 */
#define UK_INTCTLR_MAX_IRQ				224
```
as well?

lib/ukofw/Makefile.uk
michpappas: Actually what is `libfwif`? 🤔 I don't remember discussing about it before. Either way, what I understand from the above is the proposal to move the fdt parsing from a function in the ukintctlr API into driver ops because a similar pattern can be established for other types of devices. Is that correct?

Nevertheless, the rest of the GIC ops are interrupt-controller specific, and from what I see similarly for u-boot they add this function to other device-specific ops, at least for clocks and irq that I checked on. So it doesn't seem they follow a generic device API that includes such a function. I would had expected such a thing instead, ie a generic device API with common ops for all types of devices, along with a set of device-specific ops. But all this is out of scope of this PR.

Btw, there was actually a translate function in gic ops that was only taking care of the translation of the interrupt number, and I got rid of it in favor of a more fully-fledged parsing function of the `interrupts` property, perhaps you noticed that already.

lib/ukintctlr/ukintctlr.c
michpappas: Actually hold on. If we do that we will also need to do the same for the definitions of the IRQ allocation limits. Do we really want to add these checks? Because: If the driver does not define `limits.h` the build will fail. If the driver defines `limits.h` without these values, the build will fail again. So do we get any benefit from these checks after all? 🤔

plat/Config.uk
michpappas: @skuenzer wanted to have these when individual commits in a PR break the build for folks who do `git bisect` 🤷🏼‍♂️

lib/ukintctlr/include/uk/intctlr.h
michpappas: Yes, you're right. See also the discussion in https://github.com/unikraft/unikraft/pull/1103#discussion_r1332840305, but either way it will be only applicable to cases that use fdt.

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
michpappas: It's because `uk_intctlr->ops->handle_irq()` has such a parameter because of x86 that requires passing the interrupt number to the handler. An alternative would be to keep this pure and introduce a wrapper in `gic/ukintctlr.c` that ignores the `irq` parameter and calls this function. Perhaps that's more clean, what do you think?

drivers/ukintctlr/gic/Config.uk
michpappas: Yes, I noticed too. I'll fix that thanks 👍🏼

drivers/virtio/mmio/virtio_mmio.c
michpappas: > I would introduce a new function instead of having to access a global variable. 

But doesn't that apply to all `uk_intctlr->ops`? What makes `configuire_irq()` different? Or are you referring to all ops?

> Also I think it's sensible to have that function return an error code in case configuring the interrupt is not possible.

Ok, I can do that 👍🏼

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
jschlumpp: But this is the GIC internal abstraction? And you are ignoring the parameter in the `handle_irq` implementations and then use the proper IRQ number there.

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
jschlumpp: Probably yes.

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
michpappas: Not only we ignore it, but we also populate it with a dummy value. That is because on arm you retrieve (and ack) an interrupt when you read `GICC_IAR`, but at the same time you are expected to loop until you read the spurious IRQ (see the implementation in either gic-v2 or gic-v3). Because of that, we cannot move the interrupt retrieval logic outside `gic->handle_irq`. These are the merits of unified interrupt handing ops 👹

lib/ukintctlr/ukintctlr.c
michpappas: That was from the original PR by @mschlumpp so I kept it as is. I speculate it's for keeping the definition shorter and not have to deal so much with the line limit?

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
michpappas: So should `(256 - 32)` be replaced by 223? Where did `(256 - 32)` come from in the first place?

drivers/virtio/mmio/virtio_mmio.c
jschlumpp: > But doesn't that apply to all uk_intctlr->ops? What makes configuire_irq() different? Or are you referring to all ops?

Yes, it's about all operations. It's a bit inconsistent in the Unikraft codebase, but we do that in vfscore for example. 
It also feels a bit more consistent with the remaining library API (`uk_intctlr_irq_register`)

lib/ukofw/Makefile.uk
mogasergiu: Yes, ok, I was probably not clear enough, I am not asking for any change. I am simply mentioning that in the future it would be good to have a generic `of_xlate`-like method that a driver could implement with the help of a `libukfwif` that would abstract away access to DT/AML, but for now all fine. :+1: I am not requesting any changes.

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
jschlumpp: Oh hmm, I just noticed that you changed the convention of `MAX_IRQ` to include `MAX_IRQ`.

lib/ukintctlr/ukintctlr.c
jschlumpp: Where is allocatable used?

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
jschlumpp: So my understanding is that the callchain looks like this:
`trap_el1_irq(regs)` => `gic->ops.handle_irq(regs, 0)` == `gicv3_handle_irq(regs, 0)` => `uk_intctlr_irq_handle(regs, irq)`

That's why I don't see a need for having the irq parameter before `uk_intctlr_irq_handle`.

lib/ukintctlr/ukintctlr.c
mogasergiu: > benefit from these checks after all

Shouldn't the driver be enforced to define these? If we go without `#error`, I think we should at least have some default values instead? :thinking:

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
mogasergiu: > Where did `(256 - 32)` come from in the first place?

To quote the Intel SDM

> The IA-32 Architecture defines 18 predefined interrupts and exceptions and 224 user defined interrupts, which are
associated with entries in the IDT. Each interrupt and exception in the IDT is identified with a number, called a
vector. Vectors 0
through 8, 10 through 14, and 16 through 19 are the predefined interrupts and exceptions; vectors 32 through 255
are for software-defined interrupts, which are for either software interrupts or maskable hardware interrupts

So basically the first 32 are dedicated to implicit exceptions (page faults, alignment check, division by 0, etc) and the rest to 255 can be used for OS defined IRQ's. That is also why the PIC remap has to happen, because by default it generates IRQ's 0 - 15 which overlap with the exceptions, so we make it change this interval to 32 - 47 instead.

lib/ukintctlr/ukintctlr.c
mogasergiu: I mean it's a 2 letter difference, thought we may as well use the correct word. But I have not strong feelings towards this, proceed however you want.

plat/Config.uk
mogasergiu: Ah, I see, actually a good idea :+1:, especially since commits end up having that `Closes: #<PR>` tag at the end added automatically.

lib/ukintctlr/ukintctlr.c
jschlumpp: Fun fact: https://en.wiktionary.org/wiki/allocable

lib/ukofw/Makefile.uk
michpappas: > I am simply mentioning that in the future it would be good to have a generic of_xlate-like method that a driver could implement with the help of a libukfwif that would abstract away access to DT/AML, but for now all fine.

Gotcha 👍🏼 Yes we can consider such abstractions as we move forward.

drivers/virtio/mmio/virtio_mmio.c
michpappas: Okay no problem. Would you be happy if I add macro wrappers to `uk_intctlr->ops` and leave the `uk_intctlr` global public? Alternative we can implement a stricter scheme that provides all functionality via the API.

arch/x86/x86_64/include/uk/asm/arch.h
michpappas: Good point, I'll change it to `__UK_ARCH_X86_64_H`

arch/x86/x86_64/include/uk/asm/pic.h
michpappas: It was supposed to a have an include guard but I forgot. Thanks 👍🏼

drivers/virtio/mmio/virtio_mmio.c
mogasergiu: > Alternative we can implement a stricter scheme that provides all functionality via the API

That would actually be really cool. I imagine the API being a set of functions whose one of the arguments is a pointer to the intctlr itself. Each driver could have its own device structure containing its correspondent intctlr pointer, which a device driver could use to call intctlr API's.

> Each driver could have its own device structure containing its correspondent intctlr pointer

This maybe for later :D

lib/ukintctlr/ukintctlr.c
mogasergiu: Damn, you caught my english vocabulary lacking :tired_face: :rofl:

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
michpappas: Sorry I see where is the gap now. I was supposed to update `trap_el1_irq` to call the generic `handle_irq` from the `uk_intctlr` ops, as a step towards gradually moving all the calls to the driver via the API and deprecate direct access to driver ops. But I forgot 🤦🏼

lib/ukintctlr/ukintctlr.c
michpappas: Sure, I don't have a strong opinion, either way is fine with me. I'll add a few check then, as defining defaults for alloca(ta)ble interrupt numbers sounds a bit risky to me 👍🏼

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
michpappas: I now updated it to `uk_intctlr->ops->handle_irq()` please let me know if you think about the idea of adding wrapper in `gic/ukintctlr.c` so that we keep `gic->handle_irq()` pure. Personally I am leaning towards that direction.

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
michpappas: Okay, so what's the verdict on this?

drivers/virtio/mmio/virtio_mmio.c
michpappas: > I imagine the API being a set of functions whose one of the arguments is a pointer to the intctlr itself. Each driver could have its own device structure containing its correspondent intctlr pointer, which a device driver could use to call intctlr API's.

Who populates that argument and from where? And what problem does it solve?

michpappas:
@mogasergiu @mschlumpp I pushed an update to address the low hanging fruit 🍒 🍌 🥝 🍍 I also resolved any items I considered we had consensus, but please feel free to check yourselves. FTR, those with a thumbs-up mean that I took action, others (like the 🎄) that I did nothing there's no thumbs up.

For the remaining items I would like further feedback. Thanks again for reviewing! 🚀

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
mogasergiu:
```diff
-#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			223
+#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			224
#define UK_INTCTLR_MAX_IRQ				(256 - 32)
```
Probably this. @mschlumpp

drivers/ukintctlr/Makefile.uk
mogasergiu: I just think each irq controller should have its own directory: `i8259` for `PIC`, `apic` for `APIC`, `gic-v2` for `GICv2`, `gic-v3` for `GICv3`, etc. Each directory having the driver's source code as well as its corresponding private header (containing registers and other specific stuff that should not be accessible unless accessed through the driver's methods). E.g.
```
ukintctlr/
       i8259/
           i8259.c
           i8259.h
       apic/
           apic.c
           apic.h
       gic-v2/
           gic-v2.c
           gic-v2.h
       gic-v3/
           gic-v3.c
           gic-v3.h
```

jschlumpp: @mogasergiu Yes, but: the current state of the x86 interrupt system is that APIC+PIT are bundled together. You have to use both if you want to use the APIC. This PR is mostly "move everything in the new hierarchy" and not "implement proper APIC support".

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
jschlumpp: That might require some changes in the other code. Because the PR also contains some changes where the interpretation of the interval changed. (I might be wrong though)

drivers/virtio/mmio/virtio_mmio.c
jschlumpp: I'm thinking at something like in [uknetdev](https://github.com/unikraft/unikraft/blob/018fe1a4c408a2ca41beac79af95d7f2ef5794bf/lib/uknetdev/netdev.c#L554-L572). The functions might also contain some assertions and then also take care of optional operations.

mogasergiu: > > I imagine the API being a set of functions whose one of the arguments is a pointer to the intctlr itself. Each driver could have its own device structure containing its correspondent intctlr pointer, which a device driver could use to call intctlr API's.
> 
> Who populates that argument and from where? And what problem does it solve?

Well imagine it like this
- Device has irq connected to GPIO controller
- GPIO controller has its IRQ to GIC
- in DT, Device would have a phandle to the GPIO controller in its `interrupts = ` property, so the device structure could have a pointer to the GPIO controller's device structure as its irq controller
- in DT, GPIO controller would not have a phandle to GIC and we can populate that argument with the irq controller at the top of the hierarchy by default, so GPIO controller could have in its device structure a pointer to the GIC as its irq controller
- GIC would have its parent irq controller set to NULL since it is at the top.

When device wants to request an irq, the API just iteratively goes from parent to parent of the API invoking the correspondent driver method. This way you could handle nesting.

michpappas: @mschlumpp agreed, will update as suggested 👍🏼 

@mogasergiu since we don't have a use-case for nesting, let's update the API when this becomes a requirement along with any other associated changes.

drivers/ukintctlr/Makefile.uk
michpappas: > @mogasergiu Yes, but: the current state of the x86 interrupt system is that APIC+PIT are bundled together. You have to use both if you want to use the APIC. This PR is mostly "move everything in the new hierarchy" and not "implement proper APIC support".

Agreed. Regarding `xPIC` if no one comes up with a better name I'll leave as is and let you folks change it if you decide to do so later. Notice however that this would involve a non-backwards compatible change on the Kconfig options.

michpappas:
@mogasergiu, @mschlumpp I pushed an update that hopefully addresses the remaining review items.

In summary:

- Added wrappers for driver ops into the API and removed the `uk_intctlr` global.
- Due to the above change, IRQ vectors can no longer access to driver ops via the `uk_intctlr` API. Now 
  each architecture's IRQ vector invokes an interrupt handling function provided by the driver.
  As before, the driver function calls the unified handler of the `uk_intctlr` API.
- For consistency, fdt parsing is now implemented via driver ops, and a wrapper is provided by the
  `uk_intctlr` API. That adds a `fdt_xlat` function to the driver ops (@moga), but also see below (*).
- `uk_inctlr->ops->configure_irq()` now returns an error code instead of being `void`.
- `UK_INTCTLR_LAST_ALLOCABLE_IRQ` has been updated to 224. I am not aware of any open issuses related
  to code using the values exported by the drivers' `limits.h`.
- Added a few assertions in the implementation of the `uk_inctlr` API.

(*) Overall I don't like this very much, as it implies that there is only one type of node that the driver interprets, which for interrupt controllers is not necessarily true (think of the `interrupt-controller` node for example). Moreover, I don't also like that the fdt parsing function is in the scope of `uk_intctlr` while all other fdt parsing functions are in the scope of `fdt`. With that in mind it would make more sense if that remained part of `libukofw`, but then it would need to either depend on `libukinctlr` so that it populates a `struct uk_intctlr_irq_desc` or alternative we would need to delegate the cost to the user. Anyway, perhaps it's okay for now, we can consider refactoring later once more similar functionality comes.

drivers/ukintctlr/gic/include/uk/intctlr/gic.h
michpappas: Update: I removed it again, see https://github.com/unikraft/unikraft/pull/1103#issuecomment-1735547936

drivers/virtio/mmio/virtio_mmio.c
jschlumpp: @mogasergiu feel free to mark this as resolved

drivers/ukintctlr/xpic/Config.uk
```diff
@@ -0,0 +1,21 @@
...
+config LIBUKINTCTLR_XPIC
+	bool "Intel PIC"
+	depends on ARCH_X86_64
+	select LIBUKINTCTLR
+
+config LIBUKINTCTLR_APIC
+	bool "Intel APIC"
+	depends on HAVE_APIC
+	select LIBUKINTCTLR_PIC
+	imply LIBUKINTCTLR_APIC_FULL_IDT
```
jschlumpp: Is there a reason why FULL_IDT does not use `default y` instead?

lib/ukintctlr/ukintctlr.c
```diff
@@ -0,0 +1,313 @@
...
+#if CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS
+#include <uk/arch/ctx.h>
+#endif /* CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS */
+
+#if !defined(UK_INTCTLR_MAX_IRQ) ||					\
+	!defined(UK_INTCTLR_ALLOCABLE_IRQ_COUNT) ||			\
+	!defined(UK_INTCTLR_FIRST_ALLOCABLE_IRQ) ||			\
+	!defined(UK_INTCTLR_LAST_ALLOCABLE_IRQ)
+#error "Incomplete definition of uk_intctlr driver limits"
+#endif /* !UK_INTCTLR_MAX_IRQ */
```
jschlumpp: What is the policy for these `#endif` comments? 
In this case, the comment is incomplete and having all conditions would be too verbose. 
I would rather leave out the ending comment rather than having an incomplete one.

plat/common/include/uk/plat/common/irq.h
```diff
@@ -33,8 +33,12 @@
...
 
-#include <uk/plat/irq.h>
-
+/* Saving /restoring CPU IRQ flags and enabling / disabling IRQs on the
+ * CPU were previously part of the ukplat_irq API. These are not managed
+ * by the interrupt controller, so they need to be part of the lcpu
+ * low-level API. As some platforms may have their own implementation of
+ * these (linuxu, xen) keep them on a separate header.
+ */
 #if defined(__X86_64__)
```
jschlumpp: Is this message more appropriate for the commit message?
The historic functionality is not really relevant for the current state.

michpappas: Normally yes. The reason I left it there is that as a reminder for whoever migrates `lcpu`, who will unlikely check the commit history. During re-arch it will go away. I'll add a `TODO` to make sure it's removed.

lib/ukintctlr/ukintctlr.c
michpappas: Yeah you're right, we actually don't have a policy for that. I'll remove the comment 👍🏼

drivers/ukintctlr/xpic/Config.uk
michpappas: Based on internal discussion, `FULL_IDT` no longer needs to be configurable.

jschlumpp (approved):

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
```diff
@@ -6,6 +6,17 @@
...
 #define __UK_INTCTLR_XPIC_LIMITS_H__
 
-#define UK_INTCTLR_MAX_IRQ                             16
+#define UK_INTCTLR_FIRST_ALLOCABLE_IRQ			16
+
+#if CONFIG_UKINTCTLR_XPIC_FULL_IDT
+#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			224
+#define UK_INTCTLR_MAX_IRQ				(256 - 32)
+#else /* CONFIG_UKINTCTLR_XPIC_FULL_IDT */
+#define UK_INTCTLR_LAST_ALLOCABLE_IRQ			16
```
mogasergiu: Is this needed anymore?

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -43,15 +43,18 @@
...
 #include <arm/cpu.h>
+#include <uk/intctlr.h>
 #include <uk/intctlr/gic-v2.h>
+#include <uk/intctlr/limits.h>
 #include <uk/ofw/fdt.h>
 
 /* Max CPU interface for GICv2 */
 #define GIC_MAX_CPUIF		8
 
+#define GIC_MAX_IRQ	UK_INTCTLR_MAX_IRQ
```
mogasergiu: This looks misaligned with the above macro (tabs wise).

mogasergiu (approved):
Ok, LGTM so far, some minor comments :). Other than that...

mogasergiu (changes requested):
Dummy req changes

drivers/ukintctlr/xpic/include/uk/intctlr/limits.h
michpappas: Ooops, somehow that slipped. Thanks 👍🏼

drivers/ukintctlr/gic/gic-v2.c
michpappas: Done

michpappas:
@razvand this one is ready to test now

razvand:
@michpappas , @mogasergiu , @mogasergiu , will test with the current staging and this PR.

razvand:
@mschlumpp , @mogasergiu , I tested and it works. @mogasergiu , you still need to click on the approve button as part of the approval. After that, I will merge / integrate this PR.

mogasergiu (approved):

PR #1104, opened 2023-09-18
build: Support replacing internal libraries

skuenzer:
### Description of changes

This PR introduces the build system parameter `E=` which expects a colon separated list of suffixes that are matched against any library path that the build system is including. This mechanism is intended to be used for replacing internal libraries.
For example, to replace `uksched`, the replacing library (that has to match configuration name(s) and APIs) needs to be included with `L=` while at the same time, the Unikraft-internal library has to be excluded with `E=`:

```
 make L=/path/to/external-uksched E=lib/uksched
```


Makefile
```diff
@@ -665,13 +665,6 @@ CC_VER_MAJOR   := $(word 2,$(subst ., ,$(CC_INFO)))
 CC_VER_MINOR   := $(word 3,$(subst ., ,$(CC_INFO)))
 CC_VERSION     := $(CC_VER_MAJOR).$(CC_VER_MINOR)
 
-ifeq ($(call have_clang),y)
```
michpappas, line 680: Oooops, looks like I approved that back then. Sorry about that 🙈

Makefile
```diff
@@ -174,13 +174,13 @@ CONFIG_DIR := $(dir $(C))
...
 # Retrieved from P variable from the command line (paths separated by colon)
 ifeq ("$(origin P)", "command line")
-$(foreach E,$(subst :, ,$(P)), \
-$(if $(filter /%,$(E)),,$(error Path to external platform "$(E)" (P) is not absolute));\
-$(if $(wildcard $(E)), \
-	$(eval EPLAT_DIR += $(E)) \
-, $(if $(wildcard $(CONFIG_UK_BASE)/$(E)),\
-	$(eval EPLAT_DIR += $(CONFIG_UK_BASE)/$(E)), \
-	$(error Cannot find platform library: $(E)) \
+$(foreach ITR,$(subst :, ,$(P)), \
```
michpappas, line 178: I assume `ITR` stands for "iteration". If that's the true, I would say it looks a bit strange to me, as I would have expected something more conventional like `i` or something related to the context, eg `$(foreach D, $(ELIB_DIR))`. I think the origin of the problem is that we use single letter variables for non-temporary variables, which creates the risk of conflicts in such cases. One possible solution would be to use lowercase for temp vars, eg `$(foreach p,$(subst :, , $(P))`, if that is compliant with the coding style, although I have to admit lowercase vars in shell scripts and makefiles look strange too. Another one is using a longer name, eg `$(foreach DIR,$(ELIB_DIR))` but again it looks unorthodox and possibly risky too (think of `PATH`, `LIB` etc). Another one (not sure if possible) is to rename / prefix the external variables to a longer internal ones internally, and use single-letter vars internally as temps.

More importantly this highlights another problem with the single-letter public variable names: Soon we will run out of sensible letters to match concepts (`P` for platform, `L` for library etc), and start assigning arbitrary ones. To some extent that happens with `E` already. `E` (external?) could be applicable to other externally defined things. Saying that, I don't have a counter-proposal right now other than discussing whether to replace the single vars with longer variables, which is of course out of the scope of this PR.

Makefile
```diff
@@ -204,11 +204,24 @@ $(if $(wildcard $(ITR)), \
 endif
 ELIB_DIR := $(realpath $(patsubst %/,%,$(patsubst %.,%,$(ELIB_DIR))))
 
+# IMPORT_EXCLSUFFIX (list of path suffixes to exclude)
```
michpappas: While going through this PR I was confused by the term "path suffix" until I read the example in the commit message. Conceptually speaking paths don't have suffixes, so things like `lib/uksched` are effectively relative paths to unikraft's top-level directory. Perhaps it would be worth to consider also replacing `SUFFIX` with `PATH`, or `RELPATH`, or something else in the code (also applies to the changes in `config-submenu`).

Similarly, I would also rephrase the commit message to something along the lines of:

> This commit introduces the build system parameter `E=` which expects a
> colon separated list of internal library paths that the build system should
> exclude. This mechanism is intended to be used in conjunction with the
> `L=` parameter for replacing internal libraries.

For a moment I was also concerned about potential naming clashes that would exclude more than intended if we had a matching name under a different location (`lib/uksched/` and `drivers/uksched`) but I noticed you gracefully take care of this by using *paths* rather than directory names 👌🏼 Another option would had been to use library names as returned by `make print-libs` instead of paths, which also uniquely identifies libraries, and also would not have the terminology problem, but I don't see any pros / cons between the two other than that the library name would withstand changes to the path. I think that's unlikely to happen though, or at least as likely as the library changing its name, so I see both as two equivalent ways to achieve the same result. Anyway, just thinking out loud, it's good as it is now 👍🏼

support/build/config-submenu.sh
```diff
@@ -16,16 +16,19 @@ usage()
 	echo "  -t [TEXT]                  Generates a submenu with TEXT as title"
 	echo "  -l [PATH]                  Library directory"
 	echo "  -c [PATH:PATH:...]         Colon separated list of library directories"
+	echo "  -e [SUFFIX:SUFFIX:...]     Colon separated list of path (suffixes) of"
```
michpappas: Besides the `SUFFIX` part (see the relevant comment in the top-most commit)  that should be `list of paths` instead of `list of path`. A better phrasing IMO would be `Colon separated list of library paths that shall be excluded`, or even better `Colon separated list of library paths to exclude`. Also looking at the help message only, there is also a similar `-c` option that does not specify what happens to these library paths.

For consistency I would propose:
```
  -c [PATH:PATH:...]    Colon separated list of library paths to include
  -e [PATH:PATH:...]    Colon separated list of library paths to exclude
```

On the other hand, one would argue that the `-e` flag only makes sense in conjunction with `-r`. When used together with `-c` one could simply exclude the paths from `-c`.

*--- 8< --- snip snip --- 8< --- Everything below is irrelevant to this PR --- 8<--- snip snip --- 8< ---*

Some remarks on `config-submenu.sh` from the user's perspective:

I also noticed that, although the `-c` option uses the term `directories`, the `-r` option uses the terms `directories` and `subfolders`. This is inconsistent, as the term `folder` is normally associated with graphical desktop environments, as opposed to `directory` which relates to filesystems. From https://en.wikipedia.org/wiki/Directory_(computing):
> There is a difference between a directory, which is a [file system](https://en.wikipedia.org/wiki/File_system) concept, and the [graphical user interface](https://en.wikipedia.org/wiki/Graphical_user_interface) metaphor that is used to represent it (a folder).

Also irrelevant to this PR but, despite it's name, the `-r` option does not search recursively (I remember that was my initial expectation when I first ran this script when adding the `drivers` subsystem), so a description along the lines of the one below would perhaps be a bit more clear:
```
  -r [PATH:PATH:...]    Colon separated list of paths to search for libraries under.
                        Performs one level directory search, non-recursive.
```
Another alternative is to update the implementation to make it recursive.

Moreover, I think there is some redundancy between the `-c` option and the library paths one can provide directly without a flag, ie `config-submenu.sh PATH_LIB_FOO PATH_LIB_BAR PATH_LIB_BAZ` is the same as `config-submenu.sh -c PATH_LIB_FOO:PATH_LIB_BAR:PATH_LIB_BAZ`. 

Finally, when called without parameters `config-submenu.sh` should display the help message.

support/build/config-submenu.sh
```diff
@@ -92,6 +104,39 @@ else
...
+	if [ ! -f "${CONFIG_UK}" ]; then
+		if [ $OPT_QUIET -ne 0 ]; then
+			echo "Could not find \"Config.uk\" under \"${ARG_PATH}\". Skipping..." 1>&2
+		fi
+		continue
+	fi
+
+	DONT_INCLUDE=1
+	for ARG_EXCLSUFFIX in "${ARG_EXCLSUFFIXES[@]}"; do
+		if [[ "${ARG_PATH}" == *"${ARG_EXCLSUFFIX}" ]]; then
```
michpappas: You need to trim the trailing slash from the excluded path here too, otherwise you risk missing paths.

michpappas: In conjunction with the comment on the help screen: I see here that you compare `INCL` against `*EXCL`, ie expect that the former may or may not be a full path, while the latter is a relative path. It would be more fail-safe to have no expectations on whether the user passes relative or absolute paths, and instead expand both included and excluded paths to absolute ones before comparing.

support/build/config-submenu.sh
```diff
@@ -92,6 +104,39 @@ else
...
+			if [ $OPT_QUIET -ne 0 ]; then
+				echo "Skipping \"${ARG_PATH}\" due to suffix matching with \"${ARG_EXCLSUFFIX}\"..." 1>&2
```
michpappas: Not introduced in this commit only, but IIRC that if you use curly braces in your vars you don't need to additionally quote, ie
```
echo "Skipping ${ARG_PATH} due to suffix..."
```
vs
```
echo "Skipping \"$ARG_PATH\" due to suffix..."
```
I looked it up with no success, so I might be wrong... 🤔

michpappas (changes requested):
Hi @skuenzer besides the comments below, I noticed that the current approach allows to use `E` to disable an internal library without replacing it. Perhaps there's nothing wrong with it though, in fact it may provide some flexibility.

Makefile
skuenzer: :+1: no problem

support/build/config-submenu.sh
craciunoiuc: Yes, you are right, but I think here it is for readability, so people know where the arg path starts and ends

michpappas: I see, so they are meant to be displayed. Thanks for that 👍🏼

Makefile
skuenzer: Yes, `ITR` stands for iteration. I am following the variable declaration style in our build system with capital letters. In fact, I could have used `I` for it, or `DIR`... but as you said, this comes with some risk that we are overwriting something else. So, `ITR` was a bit of a pragmatic choice since it is a bit more descriptive what it is and I have the hope that this reduces the risk of potential clashes, more than a single letter variable.
In fact, the original problem is the single letter parameters that we have. I agree that we should solve this with a different PR. Originally this comes from the `V` parameter that we use to set the verbosity. I chose `E` for `exclude`. In an earlier version I used `H` for `hide`, but I was concerned that `H` is more commonly understood  for `help`. Overall, we would have then the following single-letter build system parameters:
 `V` verbosity
 `L` library
 `P` platform
 `A` application
 `E` exclude (new with this PR)

Makefile
skuenzer: Hum... path suffix is then probably the wrong name. I am open for suggestions. What I wanted to say is that it is a suffix of a path that you are handing over for matching. On purpose it should match to the right part of a path so that it feels a bit like a relative path although it isn't. Because this suffix can be also a fully qualified path that matches from right to left, the idea was to get some convenience while being able to still do full path matching.

In general, I would have preferred a different method, like you say, match on the registered library name. However, we have currently the problem that a `Makefile.uk` is a simple file include. The library name is set by the libraries' `Makefile.uk` and you can in principle even register multiple libraries with the same file. In some rare cases this happens. If we are not following a very strict standard of `Makefile.uk` it can do whatever it wants and change any build system variable and rule. On one hand this gives us great flexibility but if you want to exclude a library from a build, I think the safest way is currently only by avoid including its `Makefile.uk`... and this works only based on the path.

I guess, longer term, we could solve it with per-library sub-make calls to create a per-library Make context, similar to the Linux kernel. Then libraries would have much more separated Make context and would not intefere with each other. Maybe this could solve restrictions that we have at the moment. But I think this is a more dramatic change and a bit of out of context of this PR.

support/build/config-submenu.sh
skuenzer: :+1: Yes, these quotes are for readability and will be printed. Using the curly braces for bash variables is just good practice because you can then better control `bash` where it detects the start and the end of a variable name.

support/build/config-submenu.sh
skuenzer: Most important is that the matching result is identical with what happens in [`Makefile.rules`](https://github.com/unikraft/unikraft/pull/1104/files#diff-f535de85a92c17723a9b46772fd6d8d8252feb9802a5079056b253d1bdd01df3R246). This is the only way to keep `menuconfig` in sync with what the build system includes. This is why I kept the matching itself simple... I just noticed that I need to exclude also library-provided `Makefile.rules`... I'll add this to the PR update.

However, you are right that if you handover a trailing slash, this wont match at all anymore. I could add a comment about this in the help screen... what do you think?

skuenzer: Because the `EXCLSUFFIX` is only a matching expression on the right side of the path, it cannot be expanded to a full path. However, since the build system works with fully qualified paths, I can make sure that a path is always matched in its FQ form within `config-submenu`. So far, I did not see relative paths because `config-submenu` is feeded by the build system but with such a change we are then 100% sure that we achieve the same behavior between Makefile and menu.

support/build/config-submenu.sh
skuenzer: I agree with your comments. Maybe let us come up with a proper name for the currently called "path suffixes" first and then I can give `-e` a better description.  Basically, it is a string used for a suffix match on the library path.

Regarding irrelevant comments:
A very good point to differentiate between folder and directory. Thanks to your good explanation, I will never forget this anymore 😄 . In general, I am be fine to also update the descriptions as you suggested for the `-r` option with this PR. Makes sense, and I think opening another PR for something small like this, does probably not make so much sense.

The redundancy with `-c` and `-l` is true (and `-l` is the same as the additional arguments list) I noticed, too: `config-submenu.sh -l LIBA -l LIBB` is the same as `config-submenu.sh -c LIBA -c LIBB`. In comparison, `-c` parses additionally any colon `:` while `-l` would take a colon as part of a path. In principle, we could clean this up a bit (but better a separate PR) that removes `-c` and upgrades every parameter and the argument list with the colon feature. I am not sure if even `:` is an allowed character in a path... and for sure we are not handling it in the build system.

Makefile
michpappas: Okay, then maybe let's add the vars topic into the backlog of the build system 📝

support/build/config-submenu.sh
michpappas: I think any tool must be self-contained and function correctly on any input. Will something break if we trim trailing slashes similarly to what it does with `ARGPATH` on line 113?

Makefile
michpappas: > Hum... path suffix is then probably the wrong name. I am open for suggestions. What I wanted to say is that it is a suffix of a path that you are handing over for matching. On purpose it should match to the right part of a path so that it feels a bit like a relative path although it isn't. Because this suffix can be also a fully qualified path that matches from right to left, the idea was to get some convenience while being able to still do full path matching.

Sorry I still don't understand what you mean by "suffix of a path". Later you state that "it should match the right part of a path", so in my understanding this means that it has to be either a relative path or an absolute path. But either way it is always a path that if you pass it to the `realpath` command it should result into a fully qualified path. Perhaps you provide some examples to clarify the difference between a relative path, and absolute path and a "suffix of a path"?

support/build/config-submenu.sh
michpappas: > Because the EXCLSUFFIX is only a matching expression on the right side of the path, it cannot be expanded to a full path.

Similarly to https://github.com/unikraft/unikraft/pull/1104#discussion_r1334111702 can you please provide an example? What I understand from this is that you match `*lib/uksched`, so `$(realpath lib/uksched)` would resolve into `/home/user/unikraft/lib/uksched`. Clearly I'm missing something? 🤔

Makefile
skuenzer: :+1:

Makefile
skuenzer: I am sorry that this is confusing, I am really open ears for a better naming. You need to consider it as a string matching operation. This means that even a substring of a folder name can be matched, as long it is the right part of the name.

Examples:

suffix of a path (fully qualified path): `$(UK_BASE)/lib/ukblkdev`
matches: `$(UK_BASE)/lib/ukblkdev`

suffix of a path (relative path): `lib/ukblkdev`
matches: `$(UK_BASE)/lib/ukblkdev`

suffix of a path (relative path): `ukblkdev`
matches: `$(UK_BASE)/lib/ukblkdev`

suffix of a path (substring): `blkdev`
matches: `$(UK_BASE)/lib/ukblkdev`
In detail:
```
path   $(UK_BASE)/lib/ukblkdev
suffix                  blkdev
```

suffix of a path (substring): `dev`
matches: `$(UK_BASE)/lib/uknetdev` `$(UK_BASE)/lib/ukblkdev`

suffix of a path (substring): `ukblk`
matches: none

support/build/config-submenu.sh
skuenzer: It won't break, we just would need to keep the matching operation in sync within the [main makefile](https://github.com/unikraft/unikraft/pull/1104/files#diff-f535de85a92c17723a9b46772fd6d8d8252feb9802a5079056b253d1bdd01df3R246). What I mean is that we need to remove the trailing slash there, too. I'd need to check how this can be achieved. On the other hand, the exclude suffix is handed over by the user and then used by the build system and forwarded to `config-submenu.sh` calls. Because of this I meant the behavior must be in sync at least. At the moment both would not match with anything anymore if a trailing slash was handed over.

skuenzer: This is correct. You could even do something weird like `*ib/uksched` which would still match with `/home/user/unikraft/lib/uksched`. Everything that matches is filtered out => skipped by the build system.

Makefile
michpappas: > You need to consider it as a string matching operation. This means that even a substring of a folder name can be matched, as long it is the right part of the name.

Aha! So `substring match` is the key here 🗝️  So effectively it's a substring match against the right-hand side of a path. Now it's clear, thanks.

My question is then, what is the motivation for a substring match? I am concerned of false positives, eg the user passes `dev` with the intention to match one library and then the system matches both `libukblkdev`, and `libuknetdev`, which makes no sense, as you need to have a one-to-one relationship with libraries that you want to drop and libraries you want to add. If it's there just for convenience (eg user wants to replace both `libukblkdev` and `libuknetdev` using a single pattern) I would find it more fail-safe if we dropped substring match and require that the user explicitly specifies one-by-one the libraries to exclude as paths, either absolute or relative to the unikraft's top-level directory, and internally convert these into fully qualified paths.

PS: The rest of the comments depend on the answer to the above, so maybe we can first clarify/finalize this part first.

skuenzer: The decision for this had to do with how implementable it is while fitting convenience cases. Matching a substring is the easiest method, a one liner operation in `Make` and `bash`. I guess, with some preparation we could get it to a mode where it at least matches a relative path, for example, we could prefix `/` whenever the path suffix does not start with `/`. However, as long as we do not resolve relative paths to fully qualified once, supporting relative paths with substring matching will always has corner cases and weird behavior: e.g., what is the expected base path (e.g., $(UK_ROOT), your current workdir), and what to do with `..` (imagine `lib/../lib/../lib/ukblkdev`).

A cleaner option could be to fix the input to a fully qualified path for now (same requirement as the other build system parameters as well). It would be less convenient to use by hand but then there is nothing unclear about what is filtered-out. If the path does not exist, it is also reasonable to throw a error message.

michpappas: I would say it's a reasonable requirement to express `E` relative to `$(UK_ROOT)`, since the libraries to exclude are always part of unikraft core.

In summary I would expect to see something in the application's Makefile along the lines of:
```
UK_ROOT ?= $(PWD)/../unikraft
UK_LIBS ?= $(PWD)/..

LIBS := $(UK_LIBS)/mynetdev
ELIBS := $(UK_ROOT)/lib/uknetdev

all:
	@$(MAKE) -C $(UK_ROOT) A=$(PWD) L=$(LIBS) E=$(ELIBS)

$(MAKECMDGOALS):
	@$(MAKE) -C $(UK_ROOT) A=$(PWD) L=$(LIBS) E=$(ELIBS) $(MAKECMDGOALS)
```

The reason I would avoid having the user to specify full paths is that you may build the same config in different systems, (ie locally and on the build server, and on a different CI/CD server for testing) each one of which may fetch sources under a different path, similarly to how all other paths in the application's `Makefile` are defined relative to `$(PWD)`.

skuenzer: In your example you are calling Unikraft with fully qualified paths (or actually a path from root (`/`)) (see `ELIBS := $(UK_ROOT)/lib/uknetdev`). Except from what you write, I agree with the example because it is the approach also for the other parameters (`A`, `L`, ...). In the end there would no surprise if we make `E` inline with `L` etc. If you are okay, I'll make this change.

michpappas: Sounds great, thanks! In that case, I guess this also answers the questions on the open topics in `config-submenu.sh`, terminology etc.

skuenzer: Yes, the behavior of both need to be in sync. 👍🏻  I'll submit a version with these changes.

razvand:
@skuenzer , also be aware of the merge conflict.

support/build/config-submenu.sh
skuenzer: I updated the PR to remove a trailing slash from paths

support/build/config-submenu.sh
skuenzer: What do you think we should do with the parameters?

skuenzer:
> @skuenzer , also be aware of the merge conflict.

New version is pushed.

support/build/config-submenu.sh
```diff
@@ -58,7 +61,16 @@ while getopts :hqt:l:r:c:o: OPT; do
 		IFS=':'
 		for P in ${OPTARG}; do
 			IFS=$IFS_ORIG
-			ARG_PATHS+=("${P}")
+			[ ! -z "${P}" ] && ARG_PATHS+=("${P}")
+			IFS=$IFS_NL
```
craciunoiuc: Is `shellcheck` used on unikraft helper scripts?

For example here it suggests to use `-n` instead of `! -z`

support/build/config-submenu.sh
```diff
@@ -98,6 +110,39 @@ else
...
+		if [ $OPT_QUIET -ne 0 ]; then
+			echo "Could not find \"Config.uk\" under \"${ARG_PATH}\". Skipping..." 1>&2
+		fi
+		continue
+	fi
+
+	DONT_INCLUDE=1
+	for ARG_EXCLUDE in "${ARG_EXCLUDES[@]}"; do
+		if [ "${ARG_PATH}" = "${ARG_EXCLUDE}" ]; then
+			if [ $OPT_QUIET -ne 0 ]; then
```
craciunoiuc, line 133: Theoretically all variables should be escaped with `${}` to prevent globbing.

Also I remember reading that upper case variables are usually reserved for env vars and 'global' variables, but it is just a convention.

craciunoiuc (commented):
Hey @skuenzer 

The PR overall looks good. Only two items from my side

support/build/config-submenu.sh
craciunoiuc: If we want to enforce it we can add a new workflow that runs it and outputs stuff on all shell files (as reported by `file`)

skuenzer: Do you want me to change it? Effectively it is the same.

support/build/config-submenu.sh
michpappas: A little bit of rework is required but I think it's out of scope of this PR. Perhaps we can have a follow up PR after this one or sometime later.

Makefile
```diff
@@ -1215,6 +1220,7 @@ endif
 	@echo '                           (note: the name in the configuration file is not overwritten)'
 	@echo '  L=[PATH]:[PATH]:..     - colon-separated list of paths to external libraries'
 	@echo '  P=[PATH]:[PATH]:..     - colon-separated list of paths to external platforms'
+	@echo '  E=[PATHSUFFIX]:..      - colon-separated list of paths (or suffixes) of libraries that shall be excluded'
```
michpappas: With the latest update this message needs to be now `E=[PATH]:[PATH]` and `(or suffixes)` needs to be removed, right?

support/build/config-submenu.sh
craciunoiuc: I'm indifferent, this is just an example.

My question was more: should we enforce style checks on shell scripts in the upcoming future? I know you're the guy usually editing those so it would be good you know in advance/decide if we want to do this

support/build/config-submenu.sh
skuenzer: `shellscript` is actually not bad. We can discuss next maintainer meeting if we want to use it. I'd suggest to update our scripts based on that output.

craciunoiuc (approved):
All good from my side. We'll fix styling problems after the release.

michpappas (changes requested):
Hi @skuenzer thanks for the updates, I think it looks much nicer now. Besides one small comment below, in the commit message:
```
build: Rename `_import_linker` to `import_linker`

Because of similar reasons as in the previous commit, ...
```
In general, since commits can be cherry-picked here and there, it's preferrable to repeat the rationale instead of referring to another commit.

Other than that it looks good to merge from my side 👍🏼

Makefile
skuenzer: sorry, I somehow missed that one :+1:

skuenzer:
Update pushed, thanks for all the reviews and testing!

michpappas (approved):

mariasfiraiala (changes requested):
Thanks for this @skuenzer! I've started testing and I did a rebase locally, however even a normal build fails due to the  very recent changes that were introduced to the `ukintctlr` lib/driver dirs (#1103):

```Makefile
/home/maria/14.1/unikraft/lib/ukintctlr/ukintctlr.c:34:10: fatal error: uk/intctlr/limits.h: No such file or directory
   34 | #include <uk/intctlr/limits.h>
```

Would you mind doing a rebase to this PR and resolve the conflicts (and subsequently this issue)?

skuenzer:
> Thanks for this @skuenzer! I've started testing and I did a rebase locally, however even a normal build fails due to the very recent changes that were introduced to the `ukintctlr` lib/driver dirs (#1103):
> 
> ```makefile
> /home/maria/14.1/unikraft/lib/ukintctlr/ukintctlr.c:34:10: fatal error: uk/intctlr/limits.h: No such file or directory
>    34 | #include <uk/intctlr/limits.h>
> ```
> 
> Would you mind doing a rebase to this PR and resolve the conflicts (and subsequently this issue)?

No problem. I just pushed the rebased version 👍🏻

mariasfiraiala (changes requested):
Thanks again @skuenzer, the normal build works fine now! However, when I try to replace `uksched` with itself (for a lack of a library with the same API I am replacing a lib with itself), I get the following error:

```Makefile
$ make L=~/14.1/unikraft/lib/uksched/ E=~/14.1/unikraft/lib/uksched
In file included from /home/maria/14.1/unikraft/lib/ukboot/boot.c:70:
/home/maria/14.1/unikraft/lib/ukschedcoop/include/uk/schedcoop.h:40:10: fatal error: uk/sched.h: No such file or directory
   40 | #include <uk/sched.h>
      |          ^~~~~~~~~~~~
```

I am not sure whether this is a critical issue, other than that I'm ready to approve this!

craciunoiuc:
> Thanks again @skuenzer, the normal build works fine now! However, when I try to replace `uksched` with itself (for a lack of a library with the same API I am replacing a lib with itself), I get the following error:
> 
> ```Makefile
> $ make L=~/14.1/unikraft/lib/uksched/ E=~/14.1/unikraft/lib/uksched
> In file included from /home/maria/14.1/unikraft/lib/ukboot/boot.c:70:
> /home/maria/14.1/unikraft/lib/ukschedcoop/include/uk/schedcoop.h:40:10: fatal error: uk/sched.h: No such file or directory
>    40 | #include <uk/sched.h>
>       |          ^~~~~~~~~~~~
> ```
> 
> I am not sure whether this is a critical issue, other than that I'm ready to approve this!

I think this happens because you ignore the uksched symbol from the build and then add it again

This means that there are two uksched symbols and both are ignored

It's a good catch. Not sure if the build system should warn about this or not

skuenzer:
> > Thanks again @skuenzer, the normal build works fine now! However, when I try to replace `uksched` with itself (for a lack of a library with the same API I am replacing a lib with itself), I get the following error:
> > ```
> > $ make L=~/14.1/unikraft/lib/uksched/ E=~/14.1/unikraft/lib/uksched
> > In file included from /home/maria/14.1/unikraft/lib/ukboot/boot.c:70:
> > /home/maria/14.1/unikraft/lib/ukschedcoop/include/uk/schedcoop.h:40:10: fatal error: uk/sched.h: No such file or directory
> >    40 | #include <uk/sched.h>
> >       |          ^~~~~~~~~~~~
> > ```
> > 
> > 
> > I am not sure whether this is a critical issue, other than that I'm ready to approve this!
> 
> I think this happens because you ignore the uksched symbol from the build and then add it again
> 
> This means that there are two uksched symbols and both are ignored
> 
> It's a good catch. Not sure if the build system should warn about this or not

Really good point and a nice and clever use case for testing the feature 💡  ... what would have worked is to include `uksched` from a second unikraft repo... hum :thinking:...

At least in verbose mode (`V=1`) you'd see today that a library is skipped. My concern with an extra warning is that I need to do a parameter comparison, like checking every `L=` path, `P=` path and `A=` if there is something matching with `E=`. Because this would be another `O(n²)` operation for the build system, I'd like to avoid it in order not to further slow down calling `make`. And, btw, we are also not checking today if somebody includes an internal library a second or third time by using `L=`. I just checked, `kmenuconfig` does somehow handle it but the build fails terribly.

Nevertheless, I think it makes sense to add a clear description of this special behavior of `E=` parameter with `make help`. I wonder if you agree with me to leave it with such a change as a compromise.

mariasfiraiala (commented):
> Nevertheless, I think it makes sense to add a clear description of this special behavior of E= parameter with make help. I wonder if you agree with me to leave it with such a change as a compromise.

Yes, that's fine by me, @michpappas @craciunoiuc what do you think? 

Also, thank you so much for looking into this @skuenzer. I've tried the trick with including the same library but from another repo and it didn't work :disappointed:

michpappas:
@mariasfiraiala that's a clever test case indeed. It would be nice if the build system could handle these gracefully, but if that adds complexity and overhead to the build system, then I would agree with @skuenzer that it's probably not worth it, as it impacts both maintenance and build time.

PS: While reviewing I tested the feature by replacing `uksched` with an external library, and it worked for me. Unfortunately I don't remember the exact details of what I did it, but perhaps it would be worth to confirm it works as expected before we merge.

craciunoiuc:
I'm fine to leave it like this. At least on the kraftkit side we can handle this case and all is good

skuenzer:
> > Nevertheless, I think it makes sense to add a clear description of this special behavior of E= parameter with make help. I wonder if you agree with me to leave it with such a change as a compromise.
> 
> Yes, that's fine by me, @michpappas @craciunoiuc what do you think?
> 
> Also, thank you so much for looking into this @skuenzer. I've tried the trick with including the same library but from another repo and it didn't work 😞

What was your cmdline, what did `V=1` output? I just tried and it was fine. I did this (within `unikraft` folder):
```
make E=$( readlink -f lib/uksched ) L=$( readlink -f ../unikraft.2/lib/uksched )
```
Within the build directory I confirmed that `libuksched.o` was compiled from source files coming form the `unikraft.2` path. I looked at the `.cmd` files.

mariasfiraiala (commented):
@skuenzer, I don't have an issue when compiling unikraft by itself (without any app, inside the core), but when I try to build `helloworld` with inclusion/exclusion of libraries.

The output is the same as above (the missing `uksched` symbol). I've checked the output of `make V=1` and the library is excluded but not included again, meaning that there is no `cmd` file for it.

My build command is:

```console
maria@spectrum:~/14.1/apps/app-helloworld$ make L=~/uksched/ E=~/14.1/unikraft/lib/uksched
```

skuenzer:
> @skuenzer, I don't have an issue when compiling unikraft by itself (without any app, inside the core), but when I try to build `helloworld` with inclusion/exclusion of libraries.
> 
> The output is the same as above (the missing `uksched` symbol). I've checked the output of `make V=1` and the library is excluded but not included again, meaning that there is no `cmd` file for it.
> 
> My build command is:
> 
> ```
> maria@spectrum:~/14.1/apps/app-helloworld$ make L=~/uksched/ E=~/14.1/unikraft/lib/uksched
> ```

Ouh, is `E=` forwarded by the helloworld `Makefile`?

skuenzer:
Hey @mariasfiraiala, can you try again with the new version? I maybe found the [potential issue](https://github.com/unikraft/unikraft/compare/f9b82d8a2d9602ed870bef8e481fb1a67226fd74..b93840cad2040597f192b8c39396a09e6a008034) in your case. While I was trying with helloworld I run into this as well: The symbolic link to the main Makefile that is placed in the build directory and invoked with a sub-make call might be outdated and point to a different Unikraft repo. I updated the PR to forcefully update this symbolic link always.

craciunoiuc (approved):
All good here. Thanks!

mariasfiraiala (commented):
Unfortunately, still not working on my side. Maybe I am missing something :thinking:?

razvand:
> Unfortunately, still not working on my side. Maybe I am missing something 🤔?

@mariasfiraiala , can you do the setup on a common system, such as `wasp`, so that @skuenzer can quickly check it over there?

mariasfiraiala:
> @mariasfiraiala , can you do the setup on a common system, such as `wasp`, so that @skuenzer can quickly check it over there?

Done.

michpappas (approved):
Regarding the latest updates:

mariasfiraiala (approved):
All good, thanks a lot for the great work @skuenzer!

razvand (approved):

PR #1105, opened 2023-09-19
plat/kvm/efi.c: Obtain MemoryMap again and retry when ExitBootService fails

kha-dinh:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86`
 - Platform(s): kvm
 - Application(s): app-helloworld

### Description of changes
Changed the call to ExitBootService to re-obtain the memory map in case it fails and added a macro for printing warning messages. According to the UEFI spec, an invalid memory map is the only reason ExitBootService could fail.

### DIFF
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -654,9 +654,13 @@ void __uk_efi_api __noreturn uk_efi_main(uk_efi_hndl_t self_hndl,
 {
 	uk_efi_init_vars(self_hndl, sys_tbl);
 	uk_efi_cls();
-	uk_efi_setup_bootinfo();
 	uk_efi_reset_attack_mitigation_enable();
 	uk_efi_exit_bs();
+
+	/* uk_efi_setup_bootinfo must be called last, since it will exit Boot
+	 * Service after obtaining EFI memory map
+	 */
+	uk_efi_setup_bootinfo();
 
 	/* Jump to arch specific post-EFI entry */
 	uk_efi_jmp_to_kern();
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -27,7 +27,6 @@ static struct uk_efi_boot_services *uk_efi_bs;
 static struct uk_efi_sys_tbl *uk_efi_st;
 static uk_efi_hndl_t uk_efi_sh;
 
-static uk_efi_uintn_t uk_efi_map_key;
 static __u8 uk_efi_mat_present;
 
 /* As per UEFI specification, the call to the GetMemoryMap routine following
@@ -217,11 +216,27 @@ static int uk_efi_md_to_bi_mrd(struct uk_efi_mem_desc *const md,
 	return 0;
 }
 
-static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz,
-			    uk_efi_uintn_t *desc_sz)
+static void uk_efi_get_mmap_and_exit_bs(struct uk_efi_mem_desc **map,
+					uk_efi_uintn_t *map_sz,
+					uk_efi_uintn_t *desc_sz)
 {
+	uk_efi_uintn_t map_key;
 	uk_efi_status_t status;
+	__u8 retries = 0;
 	__u32 desc_ver;
+
+uk_efi_get_mmap_retry:
+	if (retries) {
+		if (unlikely(retries > 1))
+			UK_EFI_CRASH("Failed to exit Boot Services second time\n");
+
+		/* Free the memory map previously allocated */
+		status = uk_efi_bs->free_pages((uk_efi_paddr_t)*map,
+					       DIV_ROUND_UP(*map_sz,
+							    PAGE_SIZE));
+		if (unlikely(status != UK_EFI_SUCCESS))
+			UK_EFI_CRASH("Failed to free previous memory map\n");
+	}
 
 	/* As the UEFI Spec says:
 	 * If the MemoryMap buffer is too small, the EFI_BUFFER_TOO_SMALL
@@ -234,7 +249,7 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 	 */
 	*map_sz = 0;  /* force EFI_BUFFER_TOO_SMALL */
 	*map = NULL;
-	status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
+	status = uk_efi_bs->get_memory_map(map_sz, *map, &map_key,
 					   desc_sz, &desc_ver);
 	if (unlikely(status != UK_EFI_BUFFER_TOO_SMALL))
 		UK_EFI_CRASH("Failed to call initial dummy get_memory_map\n");
@@ -250,10 +265,21 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 		UK_EFI_CRASH("Failed to allocate memory for map\n");
 
 	/* Now we call it for real */
-	status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
+	status = uk_efi_bs->get_memory_map(map_sz, *map, &map_key,
 					   desc_sz, &desc_ver);
 	if (unlikely(status != UK_EFI_SUCCESS))
 		UK_EFI_CRASH("Failed to get memory map\n");
+
+	/* We now exit Boot Services since we no longer need it. */
+	/* In case of exit failure, we obtain the memory map again, since the
+	 * memory map may have been changed.
+	 */
+	status = uk_efi_bs->exit_boot_services(uk_efi_sh, map_key);
+	if (unlikely(status != UK_EFI_SUCCESS)) {
+		retries++;
+		uk_efi_pr_debug("ExitBootServices failed, retrying GetMemoryMap\n");
+		goto uk_efi_get_mmap_retry;
+	}
 }
 
 /* Runtime Services memory regions in the Memory Attribute Table have a higher
@@ -357,8 +383,11 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 	if (unlikely(status != UK_EFI_SUCCESS))
 		UK_EFI_CRASH("Failed to free rt_mrds\n");
 
-	/* Get memory map through GetMemoryMap */
-	uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
+	/* Get memory map through GetMemoryMap and also exit Boot service.
+	 * NOTE: after exiting, EFI printing provided by BS is not available
+	 * anymore, so UK_CRASH should be used instead.
+	 */
+	uk_efi_get_mmap_and_exit_bs(&map_start, &map_sz, &desc_sz);
 
 	map_end = (struct uk_efi_mem_desc *)((__u8 *)map_start + map_sz);
 	for (md = map_start; md < map_end;
@@ -368,7 +397,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 
 		rc = ukplat_memregion_list_insert(&bi->mrds,  &mrd);
 		if (unlikely(rc < 0))
-			UK_EFI_CRASH("Failed to insert mrd\n");
+			UK_CRASH("Failed to insert mrd\n");
 	}
 
 	ukplat_memregion_list_coalesce(&bi->mrds);
@@ -376,7 +405,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 #if defined(__X86_64__)
 	rc = ukplat_memregion_alloc_sipi_vect();
 	if (unlikely(rc))
-		UK_EFI_CRASH("Failed to insert SIPI vector region\n");
+		UK_CRASH("Failed to insert SIPI vector region\n");
 #endif
 }
 
@@ -605,15 +634,6 @@ static void uk_efi_setup_bootinfo(void)
 	bi->efi_st = (__u64)uk_efi_st;
 }
 
-static void uk_efi_exit_bs(void)
-{
-	uk_efi_status_t status;
-
-	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
-	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to to exit Boot Services\n");
-}
-
 /* Sect 4. of TCG Platform Reset Attack Mitigation Specification Version 1.10
  * Rev. 17
  */
@@ -655,7 +675,6 @@ void __uk_efi_api __noreturn uk_efi_main(uk_efi_hndl_t self_hndl,
 	uk_efi_init_vars(self_hndl, sys_tbl);
 	uk_efi_cls();
 	uk_efi_reset_attack_mitigation_enable();
-	uk_efi_exit_bs();
 
 	/* uk_efi_setup_bootinfo must be called last, since it will exit Boot
 	 * Service after obtaining EFI memory map
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -3,8 +3,10 @@
  * Licensed under the BSD-3-Clause License (the "License").
  * You may not use this file except in compliance with the License.
  */
+
 #include <kvm/efi.h>
 #include <uk/arch/paging.h>
+#include <uk/libid.h>
 #include <uk/plat/common/bootinfo.h>
 
 extern struct ukplat_memregion_desc bpt_unmap_mrd;
@@ -16,6 +18,9 @@ static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_ANY_PAGES;
 #else  /* !CONFIG_HAVE_PAGING */
 static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_MAX_ADDRESS;
 #endif /* !CONFIG_HAVE_PAGING */
+#if CONFIG_LIBUKDEBUG_PRINTD
+#include <stdio.h>
+#endif /* CONFIG_LIBUKDEBUG_PRINTD */
 
 static struct uk_efi_runtime_services *uk_efi_rs;
 static struct uk_efi_boot_services *uk_efi_bs;
@@ -36,6 +41,39 @@ static __u8 uk_efi_mat_present;
 #define EFI_STUB_CMDLINE_FNAME	CONFIG_KVM_BOOT_PROTO_EFI_STUB_CMDLINE_FNAME
 #define EFI_STUB_INITRD_FNAME	CONFIG_KVM_BOOT_PROTO_EFI_STUB_INITRD_FNAME
 #define EFI_STUB_DTB_FNAME	CONFIG_KVM_BOOT_PROTO_EFI_STUB_DTB_FNAME
+
+#define UK_EFI_MAX_FMT_STR_LEN					256
+
+#if CONFIG_LIBUKDEBUG_PRINTD
+static __sz ascii_to_utf16(const char *str, char *str16, __sz max_len16);
+static void uk_efi_printf(const char *str, ...)
+{
+	char fmt_str[UK_EFI_MAX_FMT_STR_LEN];
+	char str_tmp[UK_EFI_MAX_FMT_STR_LEN];
+	__s16 str16[UK_EFI_MAX_FMT_STR_LEN];
+	va_list ap;
+
+	sprintf(fmt_str, "dbg: [%s] <%s @ %4u> %s\r", uk_libname_self(),
+		STRINGIFY(__BASENAME__), __LINE__, str);
+
+	va_start(ap, str);
+	vsprintf(str_tmp, fmt_str, ap);
+	va_end(ap);
+
+	ascii_to_utf16(str_tmp, (char *)str16, UK_EFI_MAX_FMT_STR_LEN - 1);
+	uk_efi_st->con_out->output_string(uk_efi_st->con_out, str16);
+}
+#define uk_efi_pr_debug					uk_efi_printf
+/* UEFI for proper \n, we must also use CRLF */
+#define UK_EFI_CRASH(...)					\
+	do {							\
+		uk_efi_printf(__VA_ARGS__);			\
+		uk_efi_do_crash();				\
+	} while (0)
+#else /* !CONFIG_LIBUKDEBUG_PRINTD */
+#define uk_efi_pr_debug(...)
+#define UK_EFI_CRASH(str)				uk_efi_do_crash()
+#endif /* !CONFIG_LIBUKDEBUG_PRINTD */
 
 void uk_efi_jmp_to_kern(void) __noreturn;
 
@@ -76,30 +114,13 @@ static __sz utf16_to_ascii(const char *str16, char *str, __sz max_len)
 	return i + 1;
 }
 
-static void _uk_efi_crash(void)
+static void uk_efi_do_crash(void)
 {
 	const char reset_data[] = "UK EFI SYSTEM CRASH";
 
 	uk_efi_rs->reset_system(UK_EFI_RESET_SHUTDOWN, UK_EFI_SUCCESS,
 				sizeof(reset_data), (void *)reset_data);
 }
-
-#ifdef CONFIG_KVM_BOOT_PROTO_EFI_STUB_DEBUG
-#define UK_EFI_MAX_CRASH_STR_LEN				256
-/* UEFI for proper \n, we must also use CRLF */
-#define uk_efi_crash(str)						\
-	do {								\
-		__s16 str16[UK_EFI_MAX_CRASH_STR_LEN];			\
-									\
-		ascii_to_utf16("[uk_efi]: "str"\r", (char *)str16,	\
-			       UK_EFI_MAX_CRASH_STR_LEN - 1);		\
-		uk_efi_st->con_out->output_string(uk_efi_st->con_out,	\
-						  str16);		\
-		_uk_efi_crash();					\
-	} while (0)
-#else
-#define uk_efi_crash(str)					_uk_efi_crash()
-#endif
 
 static void uk_efi_cls(void)
 {
@@ -216,7 +237,7 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 	status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
 					   desc_sz, &desc_ver);
 	if (unlikely(status != UK_EFI_BUFFER_TOO_SMALL))
-		uk_efi_crash("Failed to call initial dummy get_memory_map\n");
+		UK_EFI_CRASH("Failed to call initial dummy get_memory_map\n");
 
 	/* Make sure the actual allocated buffer is bigger */
 	*map_sz += *desc_sz * UK_EFI_SURPLUS_MEM_DESC_COUNT;
@@ -226,13 +247,13 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 					   DIV_ROUND_UP(*map_sz, PAGE_SIZE),
 					   (uk_efi_paddr_t *)map);
 	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to allocate memory for map\n");
+		UK_EFI_CRASH("Failed to allocate memory for map\n");
 
 	/* Now we call it for real */
 	status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
 					   desc_sz, &desc_ver);
-	if unlikely((status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to get memory map\n");
+	if (unlikely(status != UK_EFI_SUCCESS))
+		UK_EFI_CRASH("Failed to get memory map\n");
 }
 
 /* Runtime Services memory regions in the Memory Attribute Table have a higher
@@ -278,7 +299,7 @@ static void uk_efi_rt_md_to_bi_mrds(struct ukplat_memregion_desc **rt_mrds,
 					  *rt_mrds_count * sizeof(**rt_mrds),
 					  (void **)rt_mrds);
 	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to allocate memory for Memory Sub-region Descriptors\n");
+		UK_EFI_CRASH("Failed to allocate memory for Memory Sub-region Descriptors\n");
 
 	/* Convert the EFI Runtime Services Memory descriptors to
 	 * ukplat_memregion_desc's
@@ -320,7 +341,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 #if defined(__X86_64__)
 	rc = ukplat_memregion_list_insert_legacy_hi_mem(&bi->mrds);
 	if (unlikely(rc < 0))
-		uk_efi_crash("Failed to insert legacy high memory region\n");
+		UK_EFI_CRASH("Failed to insert legacy high memory region\n");
 #endif
 
 	/* Fetch the Runtime Services memory regions from the MAT */
@@ -328,13 +349,13 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 	for (i = 0; i < rt_mrds_count; i++) {
 		rc = ukplat_memregion_list_insert(&bi->mrds, &rt_mrds[i]);
 		if (unlikely(rc < 0))
-			uk_efi_crash("Failed to insert rt_mrd\n");
+			UK_EFI_CRASH("Failed to insert rt_mrd\n");
 	}
 
 	/* We no longer need the list of Runtime Services memory regions */
 	status = uk_efi_bs->free_pool(rt_mrds);
 	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to free rt_mrds\n");
+		UK_EFI_CRASH("Failed to free rt_mrds\n");
 
 	/* Get memory map through GetMemoryMap */
 	uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
@@ -347,7 +368,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 
 		rc = ukplat_memregion_list_insert(&bi->mrds,  &mrd);
 		if (unlikely(rc < 0))
-			uk_efi_crash("Failed to insert mrd\n");
+			UK_EFI_CRASH("Failed to insert mrd\n");
 	}
 
 	ukplat_memregion_list_coalesce(&bi->mrds);
@@ -355,7 +376,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 #if defined(__X86_64__)
 	rc = ukplat_memregion_alloc_sipi_vect();
 	if (unlikely(rc))
-		uk_efi_crash("Failed to insert SIPI vector region\n");
+		UK_EFI_CRASH("Failed to insert SIPI vector region\n");
 #endif
 }
 
@@ -372,7 +393,7 @@ static struct uk_efi_ld_img_hndl *uk_efi_get_uk_img_hndl(void)
 					    UK_EFI_LOADED_IMAGE_PROTOCOL_GUID,
 					    (void **)&uk_img_hndl);
 	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to handle loaded image protocol\n");
+		UK_EFI_CRASH("Failed to handle loaded image protocol\n");
 
 	return uk_img_hndl;
 }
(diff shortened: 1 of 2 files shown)


mogasergiu (commented):
Hi, thank you for the PR! Indeed, the memory map key needs to be up to date before calling `ExitBootServices`. However, I don't see how it's possible in the current context that it would not be up to date when calling ExitBootServices. Should something like this happen, the mistake is most likely somewhere else. Having to re-fetch the memory map again and re-process it again denotes to me a logic mistake somewhere else.

Have you encountered a case where GetMemoryMap needs to be called again?

kha-dinh:
Hi, I encountered a case where the memory map changes when porting Unikraft to SEV-SNP (it works fine on the previous versions of SEV). SEV-SNP allows the guest to validate the gPA -> hPA memory mapping with a specific instruction, so I guest OVMF firmware does something that changes the mapping internally. I couldn't find the specific reason for this behavior though.

mogasergiu:
> Hi, I encountered a case where the memory map changes when porting Unikraft to SEV-SNP (it works fine on the previous versions of SEV). SEV-SNP allows the guest to validate the gPA -> hPA memory mapping with a specific instruction, so I guest OVMF firmware does something that changes the mapping internally. I couldn't find the specific reason for this behavior though.

I see, it makes sense. Do you have an idea where in OVMF this happens, or where in your code this happens so we can look at the corresponding OVMF code? I would not consider this current PR as the solution as this ends up looking to me like: oh, somehow, unexpectedly, without our knowledge, firmware might have changed its internel memory map, breaking the specification. I don't see how it would make sense that UEFI would do this without us invoking any UEFI routine that may involve memory manipulation.

Now I know that UEFI does use the system timer to asynchronously do timekeeping stuff as well as execute EFI drivers' routines registered through interval timer or alarm/wakeup API's present in the UEFI Driver Model. But if one of these drivers ends up invoking memory manipulation routines in such a context I'd find it to be a misbehaving driver, don't you think? I doubt UEFI has things such as LK's GFP_ATOMIC, right? It would be weird to need it. It's just some boot firmware not a high performance application in memory constraining environments :smile:.

If it indeed does turn out that to be the case somehow, I would simply still only call the setup of mrds right before ExitBootServices, but only once and there. I would just find it mindblowing that without us even invoking firmware routines, UEFI changes stuff behind the scenes, between the call to GetMemoryMap and ExitBootServices. That would be just breaking the specification IMO.

Let me know your thoughts on this.

kha-dinh:
I made this solution after looking at Linux's `efi_exit_boot_service` implementation [1], but it seems that I missed the part where a user-defined function that processes EFI's memory map is called so the memory map may be changed. So, I agree that since Unikraft does not change the memory map, retrying does not make sense.

For now, I'll find the step in the boot flow where this happens and let you know after.

[1] https://github.com/torvalds/linux/blob/master/drivers/firmware/efi/libstub/efi-stub-helper.c#L377

kha-dinh:
I played around with EFI boot and found the following sequence will exit successfully:
1. get the memory map for the first time
2. call exit boot service, which will surely fail, even when you just obtained the memory map
3. get the memory map again
4. call exit boot service, which will now succeed

The weird thing is I tried omitting steps 2 or 3, but it doesn't work. I think the problem may have to do with OVMF and not Unikraft. Do you know where in OVMF's source I should look at to explain this?

mogasergiu:
> I played around with EFI boot and found the following sequence will exit successfully:
> 
> 1. get the memory map for the first time
> 2. call exit boot service, which will surely fail, even when you just obtained the memory map

This should not happen. Are there any UEFI routines called between GetMemoryMap and ExitBootServices on your end? There have to be no memory related routines between.

> The weird thing is I tried omitting steps 2 or 3, but it doesn't work. I think the problem may have to do with OVMF and not Unikraft.

This only happens with SEV?

> Do you know where in OVMF's source I should look at to explain this?

Whenever I had issues with OVMF I would just debug OVMF itself, I have a setup ready with debug symbols and all. If you would provide me with a Unikraft branch of what you are trying to do as well as the QEMU command you use, I could start debugging OVMF while having your context and see what's going wrong. But, for starters, I would like you to answer my first question. Maybe we can pin down the issue through its answer.

kha-dinh:
> This should not happen. Are there any UEFI routines called between GetMemoryMap and ExitBootServices on your end? There have to be no memory related routines between.

I did not change the boot sequence in `kvm/efi.c`.
What I tried is I insert calls to ExitBootService right after `uk_efi_get_mmap` in `uk_efi_setup_bootinfo_mrds`. The following code would crashes 
```c
	/* Get memory map through GetMemoryMap */
	uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
	if (unlikely(status != UK_EFI_SUCCESS)) {
			uk_efi_crash("CRASH\n");
	}
	uk_pr_info( "Exit BS OK\n");
```
But the following would exit:
```c
	/* Get memory map through GetMemoryMap */
	uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
	uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
        uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
	if (unlikely(status != UK_EFI_SUCCESS)) {
			uk_efi_crash("CRASH\n");
	}
	uk_pr_info( "Exit BS OK\n");
```

> This only happens with SEV?

This happens only when I enable SEV-SNP. In the launch scripts [1], it adds this flag to qemu:
```
-object sev-snp-guest,id=sev0,policy=0x30000,cbitpos=51,reduced-phys-bits=1
```

[1] https://github.com/sev-step/sev-step/blob/release/launch-qemu.sh

mogasergiu:
> What I tried is I insert calls to ExitBootService right after uk_efi_get_mmap in uk_efi_setup_bootinfo_mrds. The following code would crashes

Could you also try to print the return values as well as the value of `uk_efi_map_key` after every call? Actually can you just print every single argument between calls and let's see if anything happens. Maybe also add some useful prints in `uk_efi_get_mmap`. Don't forget to enable `CONFIG_KVM_BOOT_PROTO_EFI_STUB_DEBUG`. 

Otherwise, the [memory map key](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Mem/Page.c#L30) seems to only be modified by [`CoreAddRange`](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Mem/Page.c#L157) which seems to be used by some functions locally the most notable one being [`PromoteMemoryResource`](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Mem/Page.c#L387). Interestingly, this function is used in a [comment](https://github.com/tianocore/edk2/blob/master/OvmfPkg/AmdSevDxe/AmdSevDxe.c#L145) of an `AMD-SEV` related function, [`AcceptAllMemory `](https://github.com/tianocore/edk2/blob/master/OvmfPkg/AmdSevDxe/AmdSevDxe.c#L98), that has this interesting comment: `Get a copy of the memory space map to iterate over while changing the map.` Anyway, this function is only called by `ResolveUnacceptedMemory` that is only called by `AmdSevDxeEntryPoint` as an [argument](https://github.com/tianocore/edk2/blob/master/OvmfPkg/AmdSevDxe/AmdSevDxe.c#L344) to this **SUPER** interesting function: [`CreateEventEx`](https://github.com/tianocore/edk2/blob/master/OvmfPkg/AmdSevDxe/AmdSevDxe.c#L341)!!!! :laughing: 

Remember when I said this?
> Now I know that UEFI does use the system timer to asynchronously do timekeeping stuff as well as execute EFI drivers' routines registered through interval timer or alarm/wakeup API's 

Well it turns out it does exactly that, but instead of doing it asynchronously on a timer event, it registers a function in a notify list to be executed when [`ExitBootServices`](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c#L770). So, since `ExitBootServices` itself modifies the memory map when it is called, it will also obviously modify the key... this is literally `ExitBootServices` ruining its own map key on purpose.... now check out this [commit](https://github.com/tianocore/edk2/commit/1b5420e8071b4f9ba13136f19365542dfe66bf04)...

LOL, I am mindblown... why does it feel like UEFI is breaking its own specification?...

Ok, back to business, what I think the fix should look like:
1. First commit: ensure there are no other UEFI routine calls between GetMemoryMap and ExitBootServices
```diff
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -616,8 +616,8 @@ void __uk_efi_api __noreturn uk_efi_main(uk_efi_hndl_t self_hndl,
 {
 	uk_efi_init_vars(self_hndl, sys_tbl);
 	uk_efi_cls();
-	uk_efi_setup_bootinfo();
 	uk_efi_reset_attack_mitigation_enable();
+	uk_efi_setup_bootinfo();
 	uk_efi_exit_bs();
 
 	/* Jump to arch specific post-EFI entry */
```

2. Second commit: Retry ExitBootServices only once (since the Accept memory event is closed afterwards by the SEV driver) in case of failure (with this opportunity, let's also get rid of the global map key variable
```diff
--- a/plat/kvm/efi.c
+++ b/plat/kvm/efi.c
@@ -12,7 +12,6 @@ static struct uk_efi_boot_services *uk_efi_bs;
 static struct uk_efi_sys_tbl *uk_efi_st;
 static uk_efi_hndl_t uk_efi_sh;
 
-static uk_efi_uintn_t uk_efi_map_key;
 static __u8 uk_efi_mat_present;
 
 /* As per UEFI specification, the call to the GetMemoryMap routine following
@@ -187,6 +186,7 @@ static int uk_efi_md_to_bi_mrd(struct uk_efi_mem_desc *const md,
 static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz,
 			    uk_efi_uintn_t *desc_sz)
 {
+	uk_efi_uintn_t uk_efi_map_key;
 	uk_efi_status_t status;
 	__u32 desc_ver;
 
@@ -216,8 +216,22 @@ static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz
 	/* Now we call it for real */
 	status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
 					   desc_sz, &desc_ver);
-	if unlikely((status != UK_EFI_SUCCESS))     <---------- LOL this got merge without surrounding parantheses!! Ez to miss
+	if (unlikely(status != UK_EFI_SUCCESS))
 		uk_efi_crash("Failed to get memory map\n");
+
+	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
+	if (status != UK_EFI_SUCCESS) {
+		status = uk_efi_bs->get_memory_map(map_sz, *map, &uk_efi_map_key,
+						   desc_sz, &desc_ver);
+		if (unlikely(status != UK_EFI_SUCCESS))
+			uk_efi_crash("Failed to get memory map second time\n");
+	} else {
+		return;
+	}
+
+	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
+	if (unlikely(status != UK_EFI_SUCCESS))
+		uk_efi_crash("Failed to to exit Boot Services again\n");
 }
 
 /* Runtime Services memory regions in the Memory Attribute Table have a higher
@@ -567,15 +581,6 @@ static void uk_efi_setup_bootinfo(void)
 	bi->efi_st = (__u64)uk_efi_st;
 }
 
-static void uk_efi_exit_bs(void)
-{
-	uk_efi_status_t status;
-
-	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
-	if (unlikely(status != UK_EFI_SUCCESS))
-		uk_efi_crash("Failed to to exit Boot Services\n");
-}
-
```

See if this fix works, I don't have an AMD CPU that supports this feature to test stuff but I believe this should work.

Let me know your thoughts! I am very surprised that OVMF does this... I find it stupid to be honest, but maybe there is something else that I am not seeing.

mogasergiu:
`ExitBootServices` be like
exitbsmapkey

kha-dinh:
Thanks for taking a deep look into it and the detailed write-up! Looks like the issue is with the implementation of SEV-SNP. On SNP, the guest has to accept pages mapped by the host explicitly, and by default, all memory is accepted at boot. OVMF chose to do this when serving ExitBootService, which creates this conundrum 😆.
As a side note, when [unaccepted memory](https://www.phoronix.com/news/UEFI-Unaccepted-Memory-CC-6.5) is supported, the OS informs OVMF that it will accept the memory, so I guess this behavior would not happen.

> See if this fix works, I don't have an AMD CPU that supports this feature to test stuff but I believe this should work.

The proposed PR looks good, and I'll incorporate and test it.

> LOL, I am mindblown... why does it feel like UEFI is breaking its own specification?...

Is it mentioned anywhere that ExitBootService should not change the memory mapping? It is indeed stupid that it changes its own mapping and then returns an error, but I guess most OS retry exiting anyway, so no one noticed.

mogasergiu:
> On SNP, the guest has to accept pages mapped by the host explicitly, and by default, all memory is accepted at boot. OVMF chose to do this when serving ExitBootService, which creates this conundrum 😆.
As a side note, when [unaccepted memory](https://www.phoronix.com/news/UEFI-Unaccepted-Memory-CC-6.5) is supported, the OS informs OVMF that it will accept the memory, so I guess this behavior would not happen.

Thank you for the insight and the useful link! :man_dancing: 

> The proposed PR looks good, and I'll incorporate and test it.

Let's hope it works :pray:. It is pretty much what you also did, but without setting up all of the mrd's again and I think we should only retry once. If in the future we are met with a case where we have to retry more times... we'll make that change :crystal_ball:.

> Is it mentioned anywhere that ExitBootService should not change the memory mapping? It is indeed stupid that it changes its own mapping and then returns an error, but I guess most OS retry exiting anyway, so no one noticed.

The UEFI spec does say this though:

> All events from the EFI_EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES
and EFI_EVENT_GROUP_EXIT_BOOT_SERVICES event notification groups as well as events of type
EVT_SIGNAL_EXIT_BOOT_SERVICES must be signaled before ExitBootServices() returns EFI_SUCCESS.

and this

> EFI_EVENT_GROUP_EXIT_BOOT_SERVICES
This event group is notified by the system when ExitBootServices() is invoked after notifying
EFI_EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES event group. This event group is functionally equivalent
to the EVT_SIGNAL_EXIT_BOOT_SERVICES flag for the Type argument of CreateEvent. The notification function
for this event must comply with the following requirements:
• The notification function is not allowed to use the Memory Allocation Services, or call any functions that use
the Memory Allocation Services, because these services modify the current memory map.
Note: Since consumer of the service does not necessarily knows if the service uses memory allocation services,
this requirement is effectively a mandate to reduce usage of any external services (services implemented outside of
the driver owning the notification function) to an absolute minimum required to perform an orderly transition to a
runtime environment. Usage of the external services may yield unexpected results. Since UEFI specification does not
guarantee any given order of notification function invocation, a notification function consuming the service may be
invoked before or after the notification function of the driver providing the service. As a result, a service being called
by the notification function may exhibit boot time behavior or a runtime behavior (which is undefined for a pure boot
services).
• The notification function must not depend on timer events since timer services will be deactivated before any
notification functions are called.

That being said, I believe OVMF's SEV driver is breaking the specification. Maybe their engineers are already aware and are actively trying to find a better workaround.

kha-dinh:
Hey @mogasergiu, I updated the PR. I found that you need to also get the memory map size again, so your changes won't work. I instead call ExitBootService in the caller, so that the sequence for getting the memory map is executed.

mogasergiu:
> I found that you need to also get the memory map size again, so your changes won't work

Ah, yes, of course. On my end it worked because I do not have a SEV environment to make the driver change the size of my memory map and thus make my ExitBootServices fail. It's good that you figured it out :100:!

Overall, the fix looks nice! I may have a few comments (mostly related to styling, functionality-wise all looks fine), but I would like to have a chance to give more complete feedback and invest some time into reading your code. This will definetely be part of 0.15, but at the moment I am caught up with PR's that need to be part of 0.14.1 and I won't have time this week. Next week however I will definetely drop my review comments so we can discuss and have this PR merged!

Thank you a lot for your collaboration and time! :pray:

plat/kvm/efi.c
```diff
@@ -3,10 +3,15 @@
  * Licensed under the BSD-3-Clause License (the "License").
  * You may not use this file except in compliance with the License.
  */
+
 #include <kvm/efi.h>
 #include <uk/arch/paging.h>
 #include <uk/plat/common/bootinfo.h>
 
+#ifdef CONFIG_KVM_BOOT_PROTO_EFI_STUB_DEBUG
```
mogasergiu: I know `#ifdef`'s are correct as well, but we are trying to use `#if`'s instead. The pattern we use is this
```C
#if CONFIG_ABC
...
#else /* !CONFIG_ABC */ <--- negation here because it is like "#elif !CONFIG_ABC"
...
#endif /* !CONFIG_ABC */ <--- repeats the last condition we had
```
and
```C
#if CONFIG_ABC
...
#endif /* CONFIG_ABC */ <--- end of CONFIG_ABC
```

Anyway, let's go one step further :small_airplane:  and follow the `libukdebug`. What do you think about deleting this config option entirely and instead make things depend on `#if CONFIG_LIBUKDEBUG_PRINTD` instead :thinking:.

Let's also add an additional function so that we can throw around debug messages :sunglasses:
```C
#if CONFIG_LIBUKDEBUG_PRINTD
#define uk_efi_pr_debug  uk_efi_printf
#else /* !CONFIG_LIBUKDEBUG_PRINTD */
#define uk_efi_pr_debug
#endif /* !CONFIG_LIBUKDEBUG_PRINTD */
```

plat/kvm/efi.c
```diff
@@ -76,16 +81,29 @@ static void _uk_efi_crash(void)
 
 #ifdef CONFIG_KVM_BOOT_PROTO_EFI_STUB_DEBUG
 #define UK_EFI_MAX_CRASH_STR_LEN				256
```
mogasergiu: Now that the use cases can be more varied, let's rename this to just `UK_EFI_MAX_FMT_STR_LEN`

plat/kvm/efi.c
```diff
@@ -76,16 +81,29 @@ static void _uk_efi_crash(void)
...
+
+static void uk_efi_printf(const char *str, ...)
+{
+	char fmt_str[UK_EFI_MAX_CRASH_STR_LEN];
+	char str_tmp[UK_EFI_MAX_CRASH_STR_LEN];
+	__s16 str16[UK_EFI_MAX_CRASH_STR_LEN];
+	va_list ap;
+
+	sprintf(fmt_str, "[uk_efi]: %s\r", str);
```
mogasergiu: Let's go with
```C
sprintf("dbg: [%s] <%s @ %4u> %s\r", uk_libid_self(), STRINGIFY(__BASENAME__), __LINE__, str);
```
instead :thinking:? (like `_vprint` and `uk_printd`)

plat/kvm/efi.c
```diff
@@ -76,16 +81,29 @@ static void _uk_efi_crash(void)
...
-									\
-		ascii_to_utf16("[uk_efi]: "str"\r", (char *)str16,	\
-			       UK_EFI_MAX_CRASH_STR_LEN - 1);		\
-		uk_efi_st->con_out->output_string(uk_efi_st->con_out,	\
-						  str16);		\
-		_uk_efi_crash();					\
+#define uk_efi_crash(str)					\
+	do {							\
+		uk_efi_printf(str);				\
+		_uk_efi_crash();				\
```
mogasergiu: I know that I initially renamed this with a leading `_`, but now we try to maintain a convention where we do not have such symbols. Let's rename `uk_efi_crash` to `UK_EFI_CRASH` (like `UK_CRASH`) and `_uk_efi_crash` to just `uk_efi_do_crash` with this opportunity :smile:. Also, shouldn't the signature be with `str, ...` and `uk_efi_printf` have two arguments as well now? :thinking:

plat/kvm/efi.c
```diff
@@ -634,10 +634,14 @@ void __uk_efi_api __noreturn uk_efi_main(uk_efi_hndl_t self_hndl,
...
 	uk_efi_init_vars(self_hndl, sys_tbl);
 	uk_efi_cls();
-	uk_efi_setup_bootinfo();
 	uk_efi_reset_attack_mitigation_enable();
 	uk_efi_exit_bs();
 
+	/* uk_efi_setup_bootinfo must be called last, since it will exit Boot
+	 * Service after obtaining EFI memory map
+	 */
+	uk_efi_setup_bootinfo();
```
mogasergiu, line 682: Awesome! :100:

plat/kvm/efi.c
```diff
@@ -319,6 +328,7 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 	__u32 rt_mrds_count = 0, i;
 	uk_efi_status_t status;
 	int rc;
+	uk_efi_uintn_t uk_efi_map_key;
```
mogasergiu: Let's try to keep a reverse xmas tree variable declaration :pray:. Since it no longer is a key global variable, let's also remove the `uk_efi_` prefix.

plat/kvm/efi.c
```diff
@@ -340,7 +350,22 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
...
+
+	/* We now exit Boot Services since we no longer need it */
+	/* In case of exit failure, we obtain the memory map again, since the
+	 * memory map may have been changed
+	 */
+	status = uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
+	if (unlikely(status != UK_EFI_SUCCESS)) {
+		uk_efi_get_mmap(&uk_efi_map_key, &map_start, &map_sz, &desc_sz);
+		status =
+		    uk_efi_bs->exit_boot_services(uk_efi_sh, uk_efi_map_key);
```
mogasergiu: Let's keep the function call on the same line as `status = ` and, instead, place the second argument on the second line (if it is still needed after the rename).
```C
status = uk_efi_bs->exit_boot_services(uk_efi_sh,
                                       map_key);
```

plat/kvm/efi.c
```diff
@@ -340,7 +350,22 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
...
+		if (unlikely(status != UK_EFI_SUCCESS))
+			uk_efi_crash(
+			    "Failed to exit Boot Services second time\n");
```
mogasergiu: Ditto, but keep the string on the same line, even if it is longer than 80 characters so that we can grep for it afterwards if we have it as an error message.
```C
uk_efi_crash("Failed to exit Boot Services second time\n");
```

plat/kvm/efi.c
```diff
@@ -349,18 +374,24 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 			continue;
 
 		rc = ukplat_memregion_list_insert(&bi->mrds,  &mrd);
-		if (unlikely(rc < 0))
-			uk_efi_crash("Failed to insert mrd\n");
+		if (unlikely(rc < 0)) {
+			uk_pr_err("Failed to insert mrd\n");
+			_uk_efi_crash();
```
mogasergiu: Why use `uk_pr_err` :thinking:? It is supposedly not initialized yet. Rely on just the proposed `UK_EFI_CRASH` instead.

plat/kvm/efi.c
```diff
@@ -202,12 +201,22 @@ static int uk_efi_md_to_bi_mrd(struct uk_efi_mem_desc *const md,
...
+			    uk_efi_uintn_t *map_sz, uk_efi_uintn_t *desc_sz)
 {
 	uk_efi_status_t status;
 	__u32 desc_ver;
 
+	/* The MemoryMap buffer may has been allocated by the previous
+	 * invocation
+	 */
+	if (*map != NULL) {
+		status = uk_efi_bs->free_pool(*map);
```
mogasergiu: Can't we do this retry in the same function still? :thinking:
It's just that this `*map != NULL` check feels unnatural and unsafe. How about we use a locally declared variable . How about we do something like this instead
```C
static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz,
			    uk_efi_uintn_t *desc_sz)
{
	uk_efi_uintn_t *map_key;
	uk_efi_status_t status;
	__u32 desc_ver;
	__u8 retries;

	retries = 0;
uk_efi_get_mmap_retry:
	if (retries) {
		if (unlikely(retries > 1))
			UK_EFI_CRASH("Failed ExitBootServices a second time\n");

		status = uk_efi_bs->free_pool(*map);
		if (unlikely(status != UK_EFI_SUCCESS))
			UK_EFI_CRASH("Failed to free previous memory map\n");
	}

	/* As the UEFI Spec says:
	 * If the MemoryMap buffer is too small, the EFI_BUFFER_TOO_SMALL
	 * error code is returned and the MemoryMapSize value contains the
	 * size of the buffer needed to contain the current memory map. The
	 * actual size of the buffer allocated for the consequent call to
	 * GetMemoryMap() should be bigger then the value returned in
	 * MemoryMapSize, since allocation of the new buffer may potentially
	 * increase memory map size.
	 */
	*map_sz = 0;  /* force EFI_BUFFER_TOO_SMALL */
	*map = NULL;
	status = uk_efi_bs->get_memory_map(map_sz, *map, &map_key,
					   desc_sz, &desc_ver);
	if (unlikely(status != UK_EFI_BUFFER_TOO_SMALL))
		UK_EFI_CRASH("Failed to call initial dummy get_memory_map\n");

	/* Make sure the actual allocated buffer is bigger */
	*map_sz += *desc_sz * UK_EFI_SURPLUS_MEM_DESC_COUNT;
	status = uk_efi_bs->allocate_pool(UK_EFI_LOADER_DATA, *map_sz,
					  (void **)map);
	if (unlikely(status != UK_EFI_SUCCESS))
		UK_EFI_CRASH("Failed to allocate memory for map\n");

	/* Now we call it for real */
	status = uk_efi_bs->get_memory_map(map_sz, *map, &map_key,
					   desc_sz, &desc_ver);
	if (unlikely(status != UK_EFI_SUCCESS))
		UK_EFI_CRASH("Failed to get memory map\n");

	status = uk_efi_bs->exit_boot_services(uk_efi_sh, map_key);
	if (status != UK_EFI_SUCCESS) {
		uk_efi_pr_debug("Failed ExitBootServices\n");
		retries++;
		goto uk_efi_get_mmap_retry;
	}
}
```
Wrote this mostly as "pseudocode"<sup>TM</sup>, haven't tested it yet

mogasergiu (changes requested):
Well, I know I said that I would review this after 0.14.1, but it seems to have been delayed even further. Nevertheless, I promised I would review the PR this week and I am going to keep my word :pray:, for I am thankful for contributions to our UEFI stub.

I have some comments, mostly related to styling and how I would love to see things in the codebase :thinking:, let me know your thoughts :pray:.

Also, as a side note, you probably worded the commits messages by taking after the way I worded them initially. Unfortunately I was wrong in the way I worded mine, w.r.t. `[subsystem]: summary` commit message headline. We have to try to avoid having literal source file names in the `[subsystem]`. Instead let's go with something that also implies the component of the subsystem, e.g. `plat/kvm: Call uk_efi_setup_bootinfo last in UEFI stub`. This way we have the subsystem, while the summary lets us know which of its sub-components we made this modification to (also keep in mind that this line has to be below [70 chars](https://github.com/unikraft/unikraft/pull/1077)). I know, it might make writing commit messages a bit more challenging, but I'd say it's worth it.

plat/kvm/efi.c
kha-dinh: > Anyway, let's go one step further 🛩️ and follow the `libukdebug`. What do you think about deleting this config option entirely and instead make things depend on `#if CONFIG_LIBUKDEBUG_PRINTD` instead 🤔.

Sounds good!

plat/kvm/efi.c
kha-dinh: ACK

plat/kvm/efi.c
kha-dinh: After exiting Boot Service, the EFI printing service does not work anymore. What do you think is a good way to handle this?

plat/kvm/efi.c
kha-dinh: I updated `uk_efi_get_mmap`  to follow this. 
 
> Can't we do this retry in the same function still? 🤔

Imo, doing it in the same function works fine in this case, but API design-wise, `uk_efi_get_mmap`   would perform two jobs at a time. The caller might not expect that getting the memory map would exit the Boot Service :smiley_cat:. How about renaming it into `uk_efi_get_mmap_and_exit`?

plat/kvm/efi.c
mogasergiu: Ah, right, this now gets executed after `ExitBootServices`... :man_facepalming:. Thanks for pointing it out :+1:.

Right... ok, well, our printing functionality on x86 is either the VGA Text Mode framebuffer which probably is not really there on UEFI (but its zone gets reserved anyway) or the good ol' "COM" UART which should always be there (I don't know of any chipset that does not initially hardcode it to those I/O ports, though might be different for some embedded x86 SoC's which we don't target anyway). On the other hand, on ARM, we actually do have something of an "early UART" and those console drivers **inefficiently** always check if the console is initialized before trying to print (for every character, ughhh...).

Therefore, I guess it's okay for now, after all, to leave it as just `uk_pr_err` + `do_uk_efi_crash` since nothing should break anyway.

plat/kvm/efi.c
mogasergiu: > I updated `uk_efi_get_mmap` to follow this.

Thank you :pray:.

> How about renaming it into `uk_efi_get_mmap_and_exit`?

Good idea, love it :tiger:.

plat/kvm/efi.c
```diff
@@ -16,6 +17,9 @@ static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_ANY_PAGES;
 #else  /* !CONFIG_HAVE_PAGING */
 static enum uk_efi_alloc_type uk_efi_alloc_type = UK_EFI_ALLOCATE_MAX_ADDRESS;
 #endif /* !CONFIG_HAVE_PAGING */
+#if CONFIG_LIBUKDEBUG_PRINTD
```
mogasergiu, line 21: Let's move this higher up together with the  other `#include`s. Also, if possible, try to have them in an alphabetic order, i.e. `uk/libid.h` would go after `uk/arch/paging.h`

plat/kvm/efi.c
```diff
@@ -216,11 +216,26 @@ static int uk_efi_md_to_bi_mrd(struct uk_efi_mem_desc *const md,
...
-static void uk_efi_get_mmap(struct uk_efi_mem_desc **map, uk_efi_uintn_t *map_sz,
-			    uk_efi_uintn_t *desc_sz)
+static void uk_efi_get_mmap_and_exit_bs(struct uk_efi_mem_desc **map,
+					uk_efi_uintn_t *map_sz,
+					uk_efi_uintn_t *desc_sz)
 {
+	uk_efi_uintn_t map_key;
 	uk_efi_status_t status;
 	__u32 desc_ver;
+	__u8 retries = 0;
```
mogasergiu, line 225: Nitpick :laughing:. I like reverse xmas tree :christmas_tree: :santa:  variable declarations. So I would move this between `status` and `desc_ver`. :nerd_face:

plat/kvm/efi.c
```diff
@@ -216,11 +216,26 @@ static int uk_efi_md_to_bi_mrd(struct uk_efi_mem_desc *const md,
...
+
+uk_efi_get_mmap_retry:
+	if (retries) {
+		if (unlikely(retries > 1))
+			UK_EFI_CRASH("Failed to exit Boot Services second time\n");
+
+		/* Free the memory map previously allocated */
+		status = uk_efi_bs->free_pages(
```
mogasergiu: Nitpick, can we rewrite this as:
```C
		/* Free the memory map previously allocated */
		status = uk_efi_bs->free_pages((uk_efi_paddr_t)*map,
		                               DIV_ROUND_UP(*map_sz,
							    PAGE_SIZE));
```

plat/kvm/efi.c
```diff
@@ -76,30 +80,46 @@ static __sz utf16_to_ascii(const char *str16, char *str, __sz max_len)
...
+#define UK_EFI_CRASH(...)					\
+	do {							\
+		uk_efi_printf(__VA_ARGS__);			\
+		uk_efi_do_crash();				\
 	} while (0)
 #else
-#define uk_efi_crash(str)					_uk_efi_crash()
-#endif
+#define uk_efi_pr_debug(...)
+#define UK_EFI_CRASH(str)					uk_efi_do_crash()
```
mogasergiu: Hmm this line seems to be over 80 characters, maybe let's have one tab less :thinking:.

plat/kvm/efi.c
```diff
@@ -356,8 +382,10 @@ static void uk_efi_setup_bootinfo_mrds(struct ukplat_bootinfo *bi)
 	if (unlikely(status != UK_EFI_SUCCESS))
 		UK_EFI_CRASH("Failed to free rt_mrds\n");
 
-	/* Get memory map through GetMemoryMap */
-	uk_efi_get_mmap(&map_start, &map_sz, &desc_sz);
+	/* Get memory map through GetMemoryMap and also exit Boot service.
+	 * NOTE: after exiting, EFI printing provided by BS is not available
+	 * anymore, so UK_CRASH should be used instead. */
```
mogasergiu: Unless it is a single line comment, we should try to not have trailing `*/` on the same line as the last line of the comment. Instead let's have it on a separate line like you did [here](https://github.com/unikraft/unikraft/pull/1105/commits/d4c51d8f0cf4b9d3a9950a7a4e5924757021cdcd#diff-4aa842463d722246e6374f63ef4cee5b61481d652a2369123dd9483383332daeR275). :pray:

mogasergiu (changes requested):
Thanks a lot for the updates @kha-dinh :pray: and I apologize for taking so long, things have been a bit busy. This should be the last round of review from my side. Everything looks very good and ready to merge.

I only have some nitpicks :smiley: . Nevertheless, I have tested on both x86_64 and ARM64 and things look good :+1: .

plat/kvm/efi.c
kha-dinh: ACK

plat/kvm/efi.c
kha-dinh: I also moved `uk_efi_printf ` out of the ifdef to have a leaner ifdef. Is that fine?

mogasergiu: Yes, great! It's even better, thank you for the attention to detail! :smile:

plat/kvm/efi.c
```diff
@@ -37,6 +41,21 @@ static __u8 uk_efi_mat_present;
...
+
+#if CONFIG_LIBUKDEBUG_PRINTD
+#define uk_efi_pr_debug					uk_efi_printf
+/* UEFI for proper \n, we must also use CRLF */
+#define UK_EFI_CRASH(...)					\
+	do {							\
+		uk_efi_printf(__VA_ARGS__);			\
+		uk_efi_do_crash();				\
+	} while (0)
+#else
```
mogasergiu: By the way we have a sort of styling revamp process going on. The way we do `#if`'s is something like:
```C
#if CONFIG
...
#endif /* CONFIG */
```
and in your case
```C
#if CONFIG
...
#else /* !CONFIG */
...
#endif /* !CONFIG */
```

mogasergiu (approved):
Thank you so much for the responsiveness, cooperation and patience :pray: !

Other than that tiny `#if` related suggestion. It all looks great!

With that being said,

plat/kvm/efi.c
kha-dinh: BTW, what is the correct practice, should function definitions be inside or outside of ifdefs?

mogasergiu: I guess it depends on the case right? :thinking: But I am not sure I understand your question. In this case it is fine to have it outside I would say (for now) as there is nobody calling it directly really and does not have any dependency. But now that you mention it I am starting to have second thoughts :sweat_smile:. Since that function is not really used it occupies unnecessary space in the image. But that is so minor... and printf is such a generically available function as well... maaan I am really not sure, really getting second thoughts now :rofl:. I hate big ifdefs too... oh well let's just put it back for now so that it is implied that printing in efi stub is depndent on PRINTD config? What do you think? I just think that it is best to keep the efi stub as tiny as possible, although that function is not of that much impact either IMO...

kha-dinh: I'm asking because sometimes I see function definitions that are related to a config but are not wrapped between ifdefs, and sometimes they are 🤣. So, I'm unsure what the correct practice here is. About the code side, I think the compiler will optimize unused functions out if there is no caller, but I'm not 100% sure.

mogasergiu: > About the code side, I think the compiler will optimize unused functions out if there is no caller, but I'm not 100% sure.

Oh, you are right, I forgot :sweat_smile:, we do have `CONFIG_OPTIMIZE_DEADELIM` and `CONFIG_OPTIMIZE_LTO` among other things :+1:.

> I'm asking because sometimes I see function definitions that are related to a config but are not wrapped between ifdefs, and sometimes they are 🤣

You are also right about that :smile: (some of those inconsistencies were also introduced by me :grin:). You could say that our guidelines are still a bit of a WIP :wink:. I'd say don't worry about this trivial styling stuff that could be easily solved during review.

Ok, let's keep the printf function. Or I will let you decide here :smile:, I am fine either way.

kha-dinh:
The last push should address the remaining issues.

mogasergiu (approved):
LGTM @razvand let's merge

razvand (approved):

PR #610, opened 2022-10-23
Introduce network support for Firecracker

Krechals:
### Prerequisite checklist
 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.pl`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.pl) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): x86_64
 - Platform(s): kvm
 - Application(s): app-httpreply

### Additional configuration

 - `CONFIG_VIRTIO_MMIO=y`
 - `CONFIG_LIBUKMMIO=y`
 
 Run app-httpreply on virtio-mmio:
```
sudo /home/qemu-6.2.0/build/qemu-system-x86_64 -M type=microvm,rtc=on,pit=on,pic=on -enable-kvm \
	-cpu host -m 512m -kernel ./build/app-httpreply_kvm-x86_64 \
	-nodefaults -no-user-config -nographic -serial stdio \
	-netdev bridge,id=en0,br=virbr0,helper=/home/qemu-bridge-helper \
	-device virtio-net-device,netdev=en0 \
	-no-acpi -bios /home/qemu-6.2.0/build/pc-bios/bios-microvm.bin \
	-append "netdev.ipv4_addr=172.44.0.2 netdev.ipv4_gw_addr=172.44.0.1 netdev.ipv4_subnet_mask=255.255.255.0 --"
```

### Description of changes
Firecracker implements virtio-mmio modern net device (Virtio 1.0+ spec). This changes are for adding network support for Firecracker.

PR #1115, opened 2023-09-27
drivers/virtio: Support modern virtio

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR collects work of various individuals for modern virtio support on Unikraft, along with additional fixes.

Supersedes #610

### DIFF
--- a/drivers/virtio/include/virtio/virtio_config.h
+++ b/drivers/virtio/include/virtio/virtio_config.h
@@ -111,11 +111,71 @@ static inline void virtio_cread_bytes(const void *addr, const __u8 offset,
 		}
 	}
 }
+
+static inline
+void virtio_mmio_cwrite_bytes(const void *addr, const __u8 offset,
+			      const void *buf, int len, int type_len)
+{
+	int i = 0;
+	__u64 io_addr;
+	int count;
+
+	count  = len / type_len;
+	for (i = 0; i < count; i++) {
+		io_addr = ((unsigned long)addr) + offset + (i * type_len);
+		switch (type_len) {
+		case 1:
+			writeb((__u8 *)io_addr, ((__u8 *)buf)[i * type_len]);
+			break;
+		case 2:
+			writew((__u16 *)io_addr, ((__u16 *)buf)[i * type_len]);
+			break;
+		case 4:
+			writel((__u32 *)io_addr, ((__u32 *)buf)[i * type_len]);
+			break;
+		default:
+			UK_CRASH("Unsupported virtio write operation\n");
+		}
+	}
+}
+
+static inline
+void virtio_mmio_cread_bytes(const void *addr, const __u8 offset,
+			     void *buf, int len, int type_len)
+{
+	int i = 0;
+	__u64 io_addr;
+	int count;
+
+	count = len / type_len;
+	for (i = 0; i < count; i++) {
+		io_addr = ((unsigned long)addr) + offset + (i * type_len);
+		switch (type_len) {
+		case 1:
+			((__u8 *)buf)[i * type_len] = readb((__u8 *)io_addr);
+			break;
+		case 2:
+			((__u16 *)buf)[i * type_len] = readw((__u16 *)io_addr);
+			break;
+		case 4:
+			((__u32 *)buf)[i * type_len] = readl((__u32 *)io_addr);
+			break;
+		case 8:
+			((__u64  *)buf)[i * type_len] = readq((__u64  *)io_addr);
+			break;
+		default:
+			UK_CRASH("Unsupported virtio read operation\n");
+		}
+	}
+}
 #else  /* !CONFIG_ARCH_X86_64 */
 
 /* IO barriers */
 #define __iormb()		rmb()
 #define __iowmb()		wmb()
+
+#define virtio_mmio_cwrite_bytes	virtio_cwrite_bytes
+#define virtio_mmio_cread_bytes		virtio_cread_bytes
 
 static inline void virtio_cwrite_bytes(const void *addr, const __u8 offset,
 				       const void *buf, int len, int type_len)
@@ -189,7 +249,7 @@ static inline int virtio_cread_bytes_many(const void *addr, const __u8 offset,
 					  __u8 *buf, __u32 len)
 {
 	__u8 old_buf[len];
-	int check = -1;
+	int check;
 	int cnt = 0;
 	__u32 i = 0;
 
@@ -310,6 +370,114 @@ static inline void virtio_cwrite32(const void *addr, const __u8 offset,
 				   const __u32 data)
 {
 	virtio_cwrite_bytes(addr, offset, &data, sizeof(data), sizeof(data));
+}
+
+/**
+ * Read an 8-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @return __u8
+ *   Returns the value configuration register.
+ */
+static inline __u8 virtio_mmio_cread8(const void *addr, const __u8 offset)
+{
+	__u8 buf = 0;
+
+	virtio_mmio_cread_bytes(addr, offset, &buf, sizeof(buf), sizeof(buf));
+
+	return buf;
+}
+
+/**
+ * Read a 16-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @return __u16
+ *   Returns the value configuration register.
+ */
+static inline __u16 virtio_mmio_cread16(const void *addr, const __u8 offset)
+{
+	__u16 buf = 0;
+
+	virtio_mmio_cread_bytes(addr, offset, &buf, sizeof(buf), sizeof(buf));
+
+	return buf;
+}
+
+/**
+ * Read a 32-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @return __u32
+ *   Returns the value configuration register.
+ */
+static inline __u32 virtio_mmio_cread32(const void *addr, const __u8 offset)
+{
+	__u32 buf = 0;
+
+	virtio_mmio_cread_bytes(addr, offset, &buf, sizeof(buf), sizeof(buf));
+
+	return buf;
+}
+
+/**
+ * Write an 8-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @param data
+ *   The value to write to the configuration.
+ */
+static inline void virtio_mmio_cwrite8(const void *addr, const __u8 offset,
+				       const __u8 data)
+{
+	virtio_mmio_cwrite_bytes(addr, offset, &data, sizeof(data),
+				 sizeof(data));
+}
+
+/**
+ * Write a 16-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @param data
+ *   The value to write to the configuration.
+ */
+static inline void virtio_mmio_cwrite16(const void *addr, const __u8 offset,
+					const __u16 data)
+{
+	virtio_mmio_cwrite_bytes(addr, offset, &data, sizeof(data),
+				 sizeof(data));
+}
+
+/**
+ * Write a 32-bit item from the device's config space
+ *
+ * @param addr
+ *   The base address of the device.
+ * @param offset
+ *   The offset with the device address space.
+ * @param data
+ *   The value to write to the configuration.
+ */
+static inline void virtio_mmio_cwrite32(const void *addr, const __u8 offset,
+					const __u32 data)
+{
+	virtio_mmio_cwrite_bytes(addr, offset, &data, sizeof(data),
+				 sizeof(data));
 }
 
 #ifdef __cplusplus
--- a/drivers/virtio/include/virtio/virtio_config.h
+++ b/drivers/virtio/include/virtio/virtio_config.h
@@ -34,30 +34,31 @@
 #define __VIRTIO_CONFIG_H__
 
 #include <uk/arch/types.h>
+#include <uk/config.h>
 #include <uk/plat/common/cpu.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus __ */
 
-#define MAX_TRY_COUNT    (10)
+#define MAX_TRY_COUNT				10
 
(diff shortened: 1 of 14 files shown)


michpappas:
Hi @Krechals, you will notice that this PR includes commits that are based of your work on #610. I added your SOB first on these commits as you are the main author, along with other contributors' SOB in the cases where they made additional changes. Please review and let me know if you're happy with it. Thanks!

CC: @razvand

razvand:
> Hi @Krechals, you will notice that this PR includes commits that are based of your work on #610. I added your SOB first on these commits as you are the main author, along with other contributors' SOB in the cases where they made additional changes. Please review and let me know if you're happy with it. Thanks!
> 
> CC: @razvand

Perfect. I added @Krechals as a reviewer, together with @mschlumpp , @zhxj9823 and @rares-miculescu , who are also reviewers for PR #1116 .

drivers/virtio/ring/virtio_ring.c
```diff
@@ -233,26 +233,34 @@ __u64 virtqueue_feature_negotiate(__u64 feature_set)
...
 	if (likely(vq->vq_callback))
 		rc = vq->vq_callback(vq, vq->priv);
+
+	/*
+	 * If the device provides a callback to handle ISRs, we let the
+	 * driver decide if the IRQ has been handled or not. Otherwise, we
+	 * expect IRQs to be handled automatically (i.e., the driver just
+	 * ignores them at this point).
+	 */
+	rc = (likely(vq->vq_callback)) ? vq->vq_callback(vq, vq->priv) : 1;
```
Krechals: The first if should be removed now.

drivers/virtio/pci/virtio_pci.c
```diff
@@ -136,17 +135,19 @@ static int virtio_pci_handle(void *arg)
...
+		uk_pr_warn("Unsupported config change interrupt received on virtio-pci device %p\n", d);
 	}
 
 	if (isr_status & VIRTIO_PCI_ISR_HAS_INTR) {
-		UK_TAILQ_FOREACH(vq, &d->vdev.vqs, next) {
+		UK_TAILQ_FOREACH(vq, &d->vdev.vqs, next)
 			rc |= virtqueue_ring_interrupt(vq);
-		}
+
+		rc = 1; /* TODO: Should not be necessary. */
```
Krechals: Indeed, this line should be removed cause it would silence other bugs.

drivers/virtio/net/virtio_net.c
```diff
@@ -922,11 +937,17 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 		VIRTIO_FEATURE_SET(drv_features, VIRTIO_NET_F_GUEST_CSUM);
 	}
 
+	/* VirtIO modern */
+	if (VIRTIO_FEATURE_HAS(host_features, VIRTIO_F_VERSION_1)) {
+		VIRTIO_FEATURE_SET(drv_features, VIRTIO_F_VERSION_1);
+		vndev->modern = 1;
```
Krechals: The only reason I added the `modern` member field was for determining the header size of virtio-net (either 10 or 12 depending on the version). Now that I think about it, we should store the header size instead:
```
vndev->hdr_size = sizeof(virtio_net_hdr_mrg_rxbuf); 
```

drivers/virtio/net/virtio_net.c
```diff
@@ -486,7 +494,7 @@ static int virtio_netdev_rxq_enqueue(struct uk_netdev_rx_queue *rxq,
 	/**
 	 * Retrieve the buffer header length.
 	 */
-	rc = uk_netbuf_header(netbuf, header_sz);
+	rc = uk_netbuf_header(netbuf, sizeof(*rxhdr));
```
Krechals: I think the space for prepending should be either 10 or 12, depending on the virtio version.

drivers/virtio/net/virtio_net.c
```diff
@@ -263,8 +265,7 @@ static void virtio_netdev_xmit_free(struct uk_netdev_tx_queue *txq)
 #define RX_FILLUP_BATCHLEN 64
 
 static int virtio_netdev_rx_fillup(struct uk_netdev_rx_queue *rxq,
-				   __u16 nb_desc,
-				   int notify)
+				   __u16 nb_desc, int notify, __u8 is_modern)
```
Krechals: I think it would me more useful to have the size of the virtio header as an argument, instead of doing so many if-else checks based on the `is_modern` flag.

drivers/virtio/mmio/virtio_mmio.c
```diff
@@ -133,40 +138,40 @@ static int vm_get(struct virtio_dev *vdev, __u16 offset,
 	__u16 w;
 	__u32 l;
 
-	if (vm_dev->version == 1) {
+	if (vm_dev->version <= 2) {
```
Krechals:
```suggestion
	if (vm_dev->version == 1) {
```

When I did this hacky change, the default case branch from the switch-case was not there. Now that we have it, there is no need for having such hack.

FTR, the default case branch is needed since the MAC address has 6 bytes and there is no way apply what the virtio standard defines (4.2.2.2 Driver Requirements: MMIO Device Register Layout)

Krechals (changes requested):
Thank you for the splitting the PRs and for your time and effort you put, @michpappas. It generally looks good to me. Just a few things to consider.

drivers/virtio/include/virtio/virtio_config.h
```diff
@@ -312,6 +372,152 @@ static inline void virtio_cwrite32(const void *addr, const __u8 offset,
...
+ *    The destination buffer to which the value has to be copied.
+ * @param len
+ *    The length of the destination buffer.
+ */
+static inline
+int virtio_mmio_cread_bytes_many(const void *addr, const __u8 offset,
+				 __u8 *buf, __u32 len)
+{
+	__u8 old_buf[len];
+	int check = -1;
```
jschlumpp: This assignment is not necessary. The first statement in the loop will write to it already.

drivers/virtio/ring/virtio_ring.c
michpappas: Indeed. I will remove, thanks 👍🏼

drivers/virtio/pci/virtio_pci.c
michpappas: Agreed, since the purpose of the return value of the interrupt handlers is to let the caller know [whether to call other handlers that may be registered for this interrupt](https://github.com/unikraft/unikraft/blob/staging/plat/kvm/irq.c#L128), ~~we can't distinguish between case where the IRQ was targeting `virtio` but for some legitimate reason there was nothing to handle (eg the virt-queue was empty), and the case where the IRQ was targeting another handler~~ we can use the `VIRTIO_MMIO_INTERRUPT_STATUS` bits to tell whether an IRQ originated from a `virtio` device, and once you know for fact that it's a `virtio` interrupt, you expect that the driver *must* return `handled`. 

Saying that, there are two issues that need to be fixed:
1. We should return `handled` if the IRQ originated from a `virtio` device but the virtqueue is found to be empty, as this is normal behavior due to the different control and data paths.
2. Because it's possible that the virtio IRQ is shared with another device, we should make sure we don't write zero to `VIRTIO_MMIO_INTERRUPT_ACK` if the interrupt status bits are empty (i.e. the interrupt did not originate from a `virtio` device) as that according to the spec resets the device.

I implement both of these changes in the next update.

CC: @marcrittinghaus who is the original author of this patch, and @mschlumpp who AFAIK previously looked at the `Unhandled IRQ` message in the scope of `virtio`.

drivers/virtio/mmio/virtio_mmio.c
michpappas: Yes you're right, thanks for spotting that 👍🏼

drivers/virtio/net/virtio_net.c
michpappas: This actually needs to use the padded header. According to the commit history:
```
    We introduce buffer padding while transmitting and receiving data. We
    explicitly introduce padding to make sure that each network packet
    needs two descriptors (one for virtio header and another for
    network packet).

    According the specification 5.1.6.6, we need to explicitly use 2
    descriptor for each transmit and receive network packet since we do not
    negotiate for VIRTIO_F_ANY_LAYOUT.
    
    In the previous implementation the sg list merged the two descriptor
    into one as the memory region were contiguous.
```

drivers/virtio/net/virtio_net.c
michpappas: ~~I think the `modern` field will be useful for other features later, that are not available in legacy. Also see the comment below on the `is_modern` parameter.~~ Actually it's not necessary, one can check modern support via `vndev->vdev->features`.

drivers/virtio/net/virtio_net.c
michpappas: Yes, I don't like it very much either. In the next update, I provide what I believe to be a cleaner approach where we pass the device as a parameter, and use a macro to derive the header size. It also adds the `num_buffers` field into the header and allows to transparently add the fields that depend on `VIRTIO_NET_F_HASH_REPORT` later.

drivers/virtio/include/virtio/virtio_config.h
michpappas: Fixed thanks. With the opportunity I also fixed the previous occurrence in the same file.

jschlumpp (approved):

drivers/virtio/include/virtio/virtio_config.h
```diff
@@ -308,7 +370,153 @@ static inline void virtio_cwrite16(const void *addr, const __u8 offset,
...
+ * @param offset
+ *   The offset with the device address space.
+ * @param buf
+ *    The destination buffer to which the value has to be copied.
+ * @param len
+ *    The length of the destination buffer.
+ */
+static inline
+int virtio_mmio_cread_bytes_many(const void *addr, const __u8 offset,
+				 __u8 *buf, __u32 len)
```
Krechals: This function is not used anywhere anymore. Let's remove it.

drivers/virtio/net/virtio_net.c
```diff
@@ -462,18 +487,18 @@ static int virtio_netdev_xmit(struct uk_netdev *dev,
...
 }
 
-static int virtio_netdev_rxq_enqueue(struct uk_netdev_rx_queue *rxq,
+static int virtio_netdev_rxq_enqueue(struct virtio_net_device *vndev,
+				     struct uk_netdev_rx_queue *rxq,
 				     struct uk_netbuf *netbuf)
 {
 	int rc = 0;
 	struct virtio_net_hdr_padded *rxhdr;
-	int16_t header_sz = sizeof(*rxhdr);
```
Krechals: `virtio_net_hdr_padded` is not defined anymore. Let's either use `virtio_net_hdr` (I think it's a bit more explicit), or remove this declaration and use `netbuf->data`.

drivers/virtio/net/virtio_net.c
```diff
@@ -41,24 +42,26 @@
...
+#define DRIVER_NAME	"virtio-net"
 
-#define DRIVER_NAME           "virtio-net"
+/* VIRTIO_PKT_BUFFER_LEN = VIRTIO_NET_HDR + ETH_HDR + ETH_PKT_PAYLOAD_LEN */
+#define VIRTIO_PKT_BUFFER_LEN(_vndev)				\
+	((UK_ETH_PAYLOAD_MAXLEN) + (UK_ETH_HDR_UNTAGGED_LEN) +	\
+	 (size_t)virtio_net_hdr_size(_vndev))
 
+/**
+ * When mergeable buffers are not negotiated, the virtio_net_hdr_padded struct
```
Krechals, line 53:
```suggestion
 * When mergeable buffers are not negotiated, the virtio_net_hdr struct
```

drivers/virtio/net/virtio_net.c
```diff
@@ -41,24 +42,26 @@
...
+ * below is placed at the beginning of the netbuf data. Use 4 bytes of pad to
+ * both keep the VirtIO header and the data non-contiguous and to keep the
+ * frame's payload 4 byte aligned.
+ */
+#define VTNET_HDR_SIZE_PADDED				\
+	(ALIGN_UP(sizeof(struct virtio_net_hdr), 4) + 4)
```
Krechals: I like this, but could we have `ALIGN_UP( (size_t)virtio_net_hdr_size(_vndev), 4)` instead? With our current supported virtio features, we only need 10-12 bytes for the virtio header. Meanwhile `struct virtio_net_hdr` has 20 bytes. Feel free to discard this comment if it's too much of a hassle for 8 bytes.

Krechals (commented):
It looks good. Thank you for making this happen, @michpappas! Only a few comments from my end.

drivers/virtio/net/virtio_net.c
michpappas: Good catch, thanks 👍🏼

drivers/virtio/net/virtio_net.c
michpappas: I thought about it at some point too, but in the end didn't do it. But since you also brought it up, I'll change it 👍🏼

drivers/virtio/include/virtio/virtio_config.h
michpappas: Okay, although it may be useful later. I guess we can add it again, if that happens.

drivers/virtio/net/virtio_net.c
```diff
@@ -354,10 +378,11 @@ static int virtio_netdev_xmit(struct uk_netdev *dev,
 
 	buf_start = pkt->data;
 	buf_len = pkt->len;
+
 	/**
 	 * Use the preallocated header space for the virtio header.
 	 */
-	rc = uk_netbuf_header(pkt, header_sz);
+	rc = uk_netbuf_header(pkt, VTNET_HDR_SIZE_PADDED(vndev));
```
Krechals, line 385: Can we somehow store `VTNET_HDR_SIZE_PADDED(vndev)`/`virtio_net_hdr_size(vndev)` in a separate variable? We are calling at least 3 times (on the base case) `virtio_net_hdr_size()` in `virtio_netdev_xmit()`.

Krechals: Maybe the compiler is smart enough to determine vndev is const inside `virtio_net_hdr_size` and not compute everything multiple times. But I haven't check...

michpappas: I'm almost sure the compiler will optimize that out :)

Krechals (approved):

razvand (approved):

PR #1088, opened 2023-09-12
{lib/uknetdev,drivers/virtio}: Add support for IPv4/TCP segmentation offloading

jschlumpp:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

This adds support for TCP segmentation offload with IPv4 packets. This allows the driver sent packets larger than the MTU and let the network device split up the TCP packets into MTU-sized IPv4 packets.

This PR also includes an implementation for virtio. Additionally, it also introduces an header for some basic network packet structures, which have been useful when implementing support for TSO in other drivers.

PR #1116, opened 2023-09-27
drivers/virtio: Pass virtio-mmio devices via the cmdline

michpappas:
### Prerequisite checklist

### Base target

### Additional configuration

A newly introduced config parameter `VIRTIO_MMIO_LINUX_COMPAT_CMDLINE` controls compatibility with Linux. For more details see the description below.

### Description of changes

This series provides the changes required to pass virtio-mmio devices via the command line. A newly introduced config parameter controls compatibility with Linux. Specifically, when `VIRTIO_MMIO_LINUX_COMPAT_CMDLINE` is selected, the MMIO driver is configured to accept devices in the Linux format, ie:
```    
virtio_mmio.device = <size>@<base>:<irq>[:<id>]
```
    
If this option is not enabled, use the default libukparam format for this library, ie:
```    
libvirtio_mmio.device = <size>@<base>:<irq>[:<id>]
```    
This is useful for VMMs like Firecracker, that automatically inject device descriptors the command line.

To additionally support `x86_64`, various components are updated to make fdt-based discovery optional. These are the platform bus, `virtio-mmio`, and PCI ECAM.

Finally, a generic function `uk_bus_pf_devmap()` to (re)map devices introduced to the platform bus. This allows removing the dependency of declaring devices at the boot pagetables for platforms that support paging (with the exception of early devices), and serves as a placeholder for additional device configuration in security related implementations like the CCA, or SEV. Notice that because of the current boot flow that initializes the `vmmap` API after the probing the interrupt controller - which is required for bringing up secondary cores - it is not possible for `uk_bus_pf_devmap()` to use `uk_vma_map_dma()`, so we need to resort to the low level paging API and map devices 1:1.

GitHub-Depends-On: #1076
GitHub-Depends-On: #1088
GitHub-Depends-On: #1089
GitHub-Depends-On: #1095
GitHub-Depends-On: #1103
GitHub-Depends-On: #1115


drivers/ukbus/platform/platform_bus.c
```diff
@@ -230,6 +235,47 @@ static int pf_init(struct uk_alloc *a)
 	return 0;
 }
 
+#if CONFIG_PAGING
+#include <uk/errptr.h>
+#include <uk/plat/paging.h>
```
jschlumpp: Move these includes to the other includes at the top of the file.

drivers/ukbus/platform/include/uk/bus/platform.h
```diff
@@ -97,6 +97,17 @@ struct pf_device {
...
+ * (Re)map device memory
+ *
+ * Maps a region with attributes suitable for device memory access.
+ * If the region is already mapped, it will be remapped unconditionally.
+ *
+ * @param base Base address of the region to map.
+ * @param size Region size. Must be page-aligned.
+ * @return virtual address
+ */
+__vaddr_t uk_bus_pf_devmap(__u64 base, __sz size);
```
jschlumpp, line 111: The function definition is only available when paging is enabled. Either provide a stub that also works without paging or make this declaration also dependent on paging.

drivers/virtio/mmio/Config.uk
```diff
@@ -15,4 +15,27 @@ config LIBVIRTIO_MMIO_FDT
...
+		format, ie:
+
+		virtio_mmio.device = <size>@<base>:<irq>[:<id>]
+
+		If this option is not enabled, use the default libukparam
+		format for this library, ie:
+
+		libvirtio_mmio.device = <size>@<base>:<irq>[:<id>]
+
+
```
jschlumpp: Very minor thing: there are two empty lines

drivers/virtio/mmio/virtio_mmio_cmdl.c
```diff
@@ -0,0 +1,135 @@
...
+
+UK_LIBPARAM_PARAM_ARR_ALIAS(device, uk_libparam_devices, charp,
+			    MAX_DEV_CMDLINE, "virtio-mmio devices");
+
+/* Helper to parse numeric string with size suffix (eg "4K") into numeric */
+static char *parse_size(const char *size_str, size_t *sz)
+{
+	char *endptr;
+
+	*sz = strtoull(size_str, &endptr, 0);
```
jschlumpp, line 31: The question is: error checking, yes or no?

drivers/virtio/mmio/virtio_mmio_cmdl.c
```diff
@@ -0,0 +1,135 @@
...
+		++endptr;
+		break;
+	case 'g':
+	case 'G':
+		*sz *= 1024 * 1024 * 1024;
+		++endptr;
+		break;
+	default:
+		break;
+	}
```
jschlumpp:
```suggestion
	switch (*endptr) {
	case 'g':
	case 'G':
		*sz *= 1024;
		/* fall through */
	case 'm':
	case 'M':
		*sz *= 1024;
		/* fall through */
	case 'k':
	case 'K':
		*sz *= 1024;
		++endptr;
		break;
	default:
		break;
	}
```
But I guess this is a matter of preference.

drivers/ukbus/platform/platform_bus.c
michpappas: These were not meant to be there, thanks.

drivers/ukbus/platform/include/uk/bus/platform.h
michpappas: Thanks 👍🏼

drivers/virtio/mmio/Config.uk
michpappas: ACK

drivers/virtio/mmio/virtio_mmio_cmdl.c
michpappas: Yes, I have seen this pattern before. We can do it that way too, either way looks fine to me

drivers/virtio/mmio/virtio_mmio_cmdl.c
michpappas: That's a good question given that proper error checking for `strtoull` is a nightmare. One possible compromise is to check against some arbitrary min / max. Any opinion?

drivers/virtio/mmio/virtio_mmio_cmdl.c
michpappas: And by the way, any objections on adding a macro for `__fallthrough` in `essentials.h`? From what I see it is supported by our minimum compiler version of both GCC (`7.x`) and clang (`9.x`).

michpappas:
@mschlumpp I updated the PR according to your feedback. I added a few commits to support the `freestanding` attribute and replace the comment-based markers. The item about whether we should do error handling in `strtoull` is still pending.

PS: I didn't rebase yet to make the changes easier to review

drivers/virtio/mmio/virtio_mmio_cmdl.c
jschlumpp: The fields are not usually set by a user manually, so checking them is not that important.

michpappas:
@mschlumpp I pushed a fix for the `LIBUKLIBPARAM` config option.

jschlumpp (approved):

rares-miculescu (approved):
Reviewed-by: Rares Miculescu [miculescur@gmail.com](mailto:miculescur@gmail.com)

razvand (approved):

PR #1070, opened 2023-08-28
lib/vfscore: Embedded initrd

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

-`LIBVFSCORE_AUTOMOUNT_ROOTFS=y`
-`LIBVFSCORE_ROOTFS_EINITRD=y`
-`LIBVFSCORE_ROOTFS_EINITRD_PATH=<path/to/initrd.cpio>`

### Description of changes

This PR introduces the ability to embed a CPIO initrd in a unikernel image. This can be useful to hard link an initrd content to an unikernel. An additionally provided initrd can then be used freely with the fstab function.

PR #1117, opened 2023-09-27
lib/uklibid: Library metadata

skuenzer:
### Additional configuration

 - `CONFIG_LIBUKLIBID_INFO=y`

### Description of changes

This PR introduces the `.uk_libinfo` section which is used to store additional library information in library object files.
For each library the build system will generate a single header and each header can contain an individual number of records to store values. Additionally, storing of information related to the image as a whole and not to a specific library is also supported.

The format of the `.uk_libinfo` section was directed by the following design principles:

- The format should be able to get changed in the future while preserving binary backwards compatibility.
- Older tools/parsers should be able to handle newer metadata layout versions somehow, e.g. by skipping unsupported metadata layout when parsing.
- Support mixing metadata layout versions within the same unikernel image.
- Adding a new record type (= metadata field) should be as easy as reserving a new record type number.
  Binary backward compatibility must be preserved.
- Externally created libraries can contain different information.
  The use of pre-built libraries with different sets of information must be supported and any information must be preserved in the final image.
 
When a final Unikraft image is linked, the individual sections of each library are merged into one section.

By default, a minimal set of library and global information is enabled. The motivation behind this is that library objects are instrumented with uk_libinfo from now on. In a future commit, the build system could provide an option to remove this section from (non-debug) images if it is not used by Unikraft code itself.

This PR depends on features that are introduced with PR #1070 .


lib/uklibid/Config.uk
```diff
@@ -31,6 +31,14 @@ config LIBUKLIBID_INFO
 				If enabled, the Unikraft release codename is
 				stored as part of the global metadata
 				information.
+
+		config LIBUKLIBID_INFO_UKCONFIGGZ
+			bool "Embed gzipped configuration (.config)"
+			default n
```
michpappas: Since our latest discussion on Kconfig we discourage using `default n`, given that it's the default behavior, and therefore explicitly stating it is redundant.

lib/uklibid/Config.uk
```diff
@@ -31,6 +31,14 @@ config LIBUKLIBID_INFO
...
+			help
+				If enabled, the current configuration `.config`
+				of the build is 'gz'ipped and embedded to the
+				global metadata infomration space.
```
michpappas, line 53: s/informration/information/

lib/uklibid/Config.uk
```diff
@@ -31,6 +31,14 @@ config LIBUKLIBID_INFO
 				If enabled, the Unikraft release codename is
 				stored as part of the global metadata
 				information.
+
+		config LIBUKLIBID_INFO_UKCONFIGGZ
+			bool "Embed gzipped configuration (.config)"
+			default n
+			help
+				If enabled, the current configuration `.config`
+				of the build is 'gz'ipped and embedded to the
```
michpappas: I think gzipped (without the quotes) is pretty common. I also did a Google search to confirm it's not just me and it returned plenty of results :)

Edit: I also noticed that you use it on this config option's title.

lib/uklibid/Config.uk
```diff
@@ -31,6 +31,14 @@ config LIBUKLIBID_INFO
 				If enabled, the Unikraft release codename is
 				stored as part of the global metadata
 				information.
+
+		config LIBUKLIBID_INFO_UKCONFIGGZ
```
michpappas, line 48: Perhaps `LIBUKLIBID_INFO_UKCONFIG_GZ` in case we later provide a different compression algorithm and also so that it doesn't sound like Gorillaz 🦍? Actually I notice other config options in this PR concatenate words (eg `COPMPILEDBY`), I think also in general we normally break these down? Anyway, if you like it as is, please feel free to ignore,  as maybe it's only a matter of personal preference.

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+			bool "Full unikraft version"
+			default y
+			help
+				If enabled, the full Unikraft version including
+				Git SHA is embedded. Otherwise, only the short
+				Unikraft version is stored.
+
+		config LIBUKLIBID_INFO_UKCODENAME
+			bool "Unikraft codename"
+			default n
```
michpappas: `default n` redundant here too

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+			This option stores the compile date and time for each
+			library and the unikernel image.
+
+	config LIBUKLIBID_INFO_COMPILEDBY_TEXT
+		string "Compiled by"
+		help
+			Individual text that is added to each compiled library
+			and to the global metadata information of a unikernel.
+			If used, the text should name the person/company/robot
+			that compiled the libraries or unikernel.
```
michpappas: Newline missing between these two config options

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+	config LIBUKLIBID_INFO_COMPILEDBY
+		bool
+		default y if LIBUKLIBID_INFO_COMPILEDBY_TEXT != ""
+
+	config LIBUKLIBID_INFO_COMPILEDBYASSOC_TEXT
+		string "Compiled by (association)"
+		depends on LIBUKLIBID_INFO_COMPILEDBY
```
michpappas: Newline missing here too :)

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+		bool "Compiler (CC) information"
+		default y
+		help
+			This option enables storing of the C compiler ($(CC))
+			that was in use while a library and unikernel was
+			compiled.
+
+	config LIBUKLIBID_INFO_COMPILEDATE
+		bool "Compile date"
+		default n
```
michpappas: `default n` redundant bla bla 🦜

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+				If enabled, the Unikraft release codename is
+				stored as part of the global metadata
+				information.
+	endmenu
+
+	config LIBUKLIBID_INFO_COMPILER
+		bool "Compiler (CC) information"
+		default y
+		help
+			This option enables storing of the C compiler ($(CC))
```
michpappas: "enables storing the C compiler that was used" or "enables the storing of the C compiler that was used" 🤓

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
...
+
+	config LIBUKLIBID_INFO_COMPILEDATE
+		bool "Compile date"
+		default n
+		help
+			This option stores the compile date and time for each
+			library and the unikernel image.
+
+	config LIBUKLIBID_INFO_COMPILEDBY_TEXT
+		string "Compiled by"
```
michpappas: It's customary to automatically derive the username (and hostname in the form of `user@host`) that invoked the build, instead of a using a free text, so this could be changed instead to a boolean option that enables fetching that info from the Makefile similarly to the compiler?

Besides, having it as free text somewhat defeats the purpose of traceable builds, as it requires that every time the user remember to update this, which realistically speaking is unlikely to happen. Think for example the case where a someone creates a build and later decides to distribute to a 3rd party. Even on automated builds, it would be required that the build system needs to update the configuration, which in theory shouldn't do.

Saying that, what would be actually useful to have as free text is an additional release / version identifier. That can be used to describe, among others, release variants (eg `+musl` vs `+newlib`) or a global release version that applies to the entire release artifact (i.e. `my_xyz` version `1.0` which consists of `unikraft` `RELEASE-0.14.0`, `app-mine` `v1.2.0`, `lib-musl` `RELEASE-0.14.0`, `lib-mine` `v2.0`, ...).

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,87 @@
+#ifndef __LIBUKLIBID_INFO_H__
```
michpappas: 1. Copyright header is needed here.
2. The header is not the main header of the library, and therefore it should be under `uklibid/include/uk/libid/libifo.h`
3. According to the conventions the include guards should be  `__UK_LIBID_INFO_H__`.

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,70 @@
-config LIBUKLIBID
+menuconfig LIBUKLIBID
 	bool "uklibid: Library identifier"
-	default n
+	default y
 	# FIXME: We actually depend on <uk/assert.h> but lib/ukdebug also
```
michpappas, line 4: TBD, not necessarily in the scope of this PR: Should `libukdebug` be a hard requirement for production builds? Is the dependency temporary until we implement the always-available definitions in `libukdebug`?

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,87 @@
...
+#include <uk/arch/types.h>
+#include <uk/essentials.h>
+
+/*
+ * Library information is stored in the special section " .uk_libinfo". For
+ * each library there is one `struct uk_libid_info_hdr` generated. Each header
+ * can contain multiple `struct uk_libid_info_rec` records. A library name
+ * record (`UKLI_REC_LIBNAME`) is the only mandatory record type for each
+ * library.
+ * Additonally, a single header is created for global information, which differs
```
michpappas: "Additionally"

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,87 @@
...
+	unsigned short type;    /* record type */
+	unsigned int   len;     /* including record header */
+	char           data[];  /* raw data */
+} __packed __align(1);
+
+/*
+ * NOTE: Due to binary cross-compatibility, the sizes, offsets, data types and
+ *       meanings must remain the same regardless of the layout version:
+ *       `len` and `version`.
+ */
```
michpappas: In that case, is there any reason for not using fixed-width integer types and endianess for all fields both here and in `struct uk_libid_info_rec`? Especially given that one may be reading the metadata from a cross-compiled build.

**Edit:** Fixed endianess may require some more work on big-endian architectures, but maybe we can define it as little-endian and cross that bridge when we get there? (ie have whoever introduces the architecture add endianess conversion). If a big-endian architecture comes into Unikraft ever, that is.

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,87 @@
...
+#define UKLI_REC_COMPILEDATE      0x000C /* C-string */
+#define UKLI_REC_COMPILEDBY       0x000D /* C-string */
+#define UKLI_REC_COMPILEDBYASSOC  0x000E /* C-string */
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_libid_info_rec {
```
michpappas: The good old TLV 👴🏼 👍🏼

support/scripts/uk-libinfo.py
```diff
@@ -0,0 +1,252 @@
...
+	(opt, rem_args) = parser.parse_known_args()
+
+	try:
+		uklibinfo = uklibinfo_elf_load(opt.file)
+	except Exception as e:
+		print("{}: Failed to load uklibinfo data: {}".format(opt.file, str(e)))
+		sys.exit(1)
+
+	try:
+		#
```
michpappas, line 256: I see in various places (`README.md`, this commit message) mentioned that the metadata are available in debug images. I wonder whether it would be useful for some adopters to be able to optionally keep metadata also on the final binary image. Besides adding a `magic` to identify the beginning of the metadata, would it take something more to support this feature?

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,102 @@
...
+ #define UKLI_LAYOUT              0x0001
+
+/*
+ * NOTE: In order to keep backwards compatibility with pre-compiled
+ *       libraries, this list should only be extended with new entries.
+ */
+#define UKLI_REC_LIBNAME          0x0001 /* C-string, absent for global hdr */
+#define UKLI_REC_COMMENT          0x0002 /* C-string */
+#define UKLI_REC_VERSION          0x0003 /* C-string */
+#define UKLI_REC_LICENSE          0x0004 /* C-string */
```
michpappas: License info is actually a **very** good idea 👍🏼 Since there is no config option to enable this, is this a placeholder for something we will be deriving via automation later?

CC: @kubanrob

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,102 @@
...
+ */
+ #define UKLI_LAYOUT              0x0001
+
+/*
+ * NOTE: In order to keep backwards compatibility with pre-compiled
+ *       libraries, this list should only be extended with new entries.
+ */
+#define UKLI_REC_LIBNAME          0x0001 /* C-string, absent for global hdr */
+#define UKLI_REC_COMMENT          0x0002 /* C-string */
+#define UKLI_REC_VERSION          0x0003 /* C-string */
```
michpappas: Perhaps it would be worth adding short comments to clarify the difference between this `VERSION`, `UK_VERSION`, and `UK_FULLVERSION`.

lib/uklibid/include/uk/libinfo.h
```diff
@@ -0,0 +1,102 @@
...
+#define UKLI_REC_LICENSE          0x0004 /* C-string */
+#define UKLI_REC_GITDESC          0x0005 /* C-string */
```
michpappas: Any reason we are not implementing this?

lib/uklibid/Makefile.uk
```diff
@@ -44,5 +44,11 @@ ASFLAGS += -D__LIBUKLIBID_COMPILER__="$(CC_INFO)"
 LIBUKLIBID_SRCS-$(CONFIG_LIBUKLIBID_INFO) +=  $(LIBUKLIBID_BASE)/infosec.ld
 LIBUKLIBID_SRCS-$(CONFIG_LIBUKLIBID_INFO) += $(LIBUKLIBID_BASE)/libinfo.S|global
 LIBUKLIBID_LIBINFO_GLOBAL_FLAGS-y += -D__GLOBALINFO__
+LIBUKLIBID_LIBINFO_GLOBAL_CDEPS-$(CONFIG_LIBUKLIBID_INFO_UKCONFIGGZ) += $(LIBUKLIBID_BUILD)/config.gz
+LIBUKLIBID_LIBINFO_GLOBAL_FLAGS-$(CONFIG_LIBUKLIBID_INFO_UKCONFIGGZ) += -DUK_CONFIGGZINC="$(LIBUKLIBID_BUILD)/config.gz"
+
+$(LIBUKLIBID_BUILD)/config.gz: $(UK_CONFIG_OUT)
```
michpappas, line 58: As is this doesn't build because nothing calls this rule:
```
/home/mpp/devel/unikraft/uk_helloworld/unikraft/lib/uklibid/libinfo.S: Assembler messages:
/home/mpp/devel/unikraft/uk_helloworld/unikraft/lib/uklibid/libinfo.S:28: Error: file not found: /home/mpp/devel/unikraft/uk_helloworld/app-helloworld/build/libuklibid/config.gz
make[3]: *** [/home/mpp/devel/unikraft/uk_helloworld/unikraft/support/build/Makefile.build:27: /home/mpp/devel/unikraft/uk_helloworld/app-helloworld/build/libuklibid/libinfo.global.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:1135: sub-make] Error 2
make[1]: *** [Makefile:41: _all] Error 2
make[1]: Leaving directory '/home/mpp/devel/unikraft/uk_helloworld/unikraft'
make: *** [Makefile:6: all] Error 2
```
Perhaps it needs something like:
```
UK_PREPARE-$(CONFIG_LIBUKLIBID_INFO_UKCONFIGGZ) += $(LIBUKLIBID_BUILD)/config.gz
```

support/scripts/uk-libinfo.py
```diff
@@ -0,0 +1,252 @@
...
+				self._tdesc = "Unikraft full version"
+				self._data  = self._data_decode(data, 'ascii')
+			case 0x0008:
+				self._tname = "UKCODENAME"
+				self._tdesc = "Unikraft codename"
+				self._data  = self._data_decode(data, 'ascii')
+			case 0x0009:
+				self._tname = "UKCONFIG"
+				self._tdesc = "Unikraft .config"
+				self._data  = (True, data)
```
michpappas, line 59: It would be useful to add an option to extract `config.gz` into a local file. Similarly for other binary data.

lib/uklibid/infodump.c
```diff
@@ -0,0 +1,50 @@
+#include <uk/libinfo.h>
+#include <uk/ctors.h>
+#include <uk/print.h>
+#include <uk/assert.h>
+
+static void libinfo_bootdump(void)
```
michpappas, line 11: Besides this full info, it would be useful to provide the ability to the user to specify the format of a single line version banner that can take values from the global data.

That could be something along the lines of:
```
 "mysomething %LI_REC_VERSION %LI_REC_USER@%LI_REC_HOST %LI_REC_COMPILE_DATE\n"
```

@kubanrob I take it that you need this feature?

michpappas (changes requested):
Hi @skuenzer that's a masterpiece of engineering 👷🏼 I really like it 😃 Please find some items to consider below.

michpappas:
@kubanrob I am bringing this into your attention as you have immediate interest on this feature

lib/uklibid/Config.uk
skuenzer: Right, no problem - I'll remove this.

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/Makefile.uk
skuenzer: You need PR #1070 , it provides a rule which resolves this dependency. Basically it makes this line work:
```
LIBUKLIBID_LIBINFO_GLOBAL_CDEPS-$(CONFIG_LIBUKLIBID_INFO_UKCONFIGGZ) += ...
```
But your suggestion would probably also work, a very good point :thinking:. `CDEPS` can be used to directly express additional file dependencies while `UK_PREPARE` is a bit more indirect and handled by the build system.

support/scripts/uk-libinfo.py
skuenzer: I agree, it was also the plan. My goal was to have this early version as part of this upcoming release so that we have the infrastructure and can add these missing items to the tool or build system afterwards.

lib/uklibid/infodump.c
skuenzer: Sounds good, also here I thought we can improve after the release because the infrastructure is there by then. Another option that we could do is to dump the library list on a crash screen.

lib/uklibid/include/uk/libinfo.h
skuenzer: The idea was to provide the framework with this PR and extend it with functionality subsequently. The PR misses currently a feature to execute a script or command for each library to retrieve its Git version. Similarly are the licenses. Do we want to detect SPDX headers or should we use a defined place per library to look this information up? My idea was that these fields could be manually already be used with custom code and would have a well-defined place to store. Implementing these fields with an automated way could be done afterwards. What do you think?

lib/uklibid/Config.uk
skuenzer: Good point, I am going to change this. Thanks 👍🏻

lib/uklibid/include/uk/libinfo.h
skuenzer: Sure, I'll add something. It is basically a 1-to-1 mapping of the provided default defines from `/Makefile.uk` at line 82 and 83.

lib/uklibid/Config.uk
skuenzer: 🦍 ... nice association 😄 . You mean only the config option or also the record type names: `UKLI_REC_*` and within `support/scripts/uk-libinfo.py`? I kept all in sync and did not use any `_` for the typenames. It is probably personal preference, I don't mind if we do one or the other way. My concern was only that with `_`, some names get really long 🍎 🐍 . For consistency, shouldn't we then also introduce a `_` between `UK` and the type, like: `UK_CONFIG_GZ` (-> `CONFIG_LIBUKLIBID_INFO_UK_CONFIG_GZ`), `UK_FULLVERSION`?

michpappas: I think `UK` is kind of an exception because in many places we use `UKXYZ`. But anyway, it you prefer it that way it's also fine, I just find it a bit unusual to the rest of our code.

lib/uklibid/include/uk/libinfo.h
michpappas: The git part is much easier than the licensing which can be quite tricky for libraries that involve multiple licenses, and may require generating a license report using automated tooling like FOSSology or BlackDuck.

But okay, if the intention was to have it as a placeholder, then as soon as someone needs it will open a PR 👍🏼

support/scripts/uk-libinfo.py
michpappas: Ok, we can do it when it's needed then 👍🏼

lib/uklibid/infodump.c
michpappas: Ok 👍🏼

lib/uklibid/Makefile.uk
michpappas: > You need PR https://github.com/unikraft/unikraft/pull/1070 , it provides a rule which resolves this dependency

Okay I see, I totally missed that. In that case, I guess this is unnecessary, `CDEPS` sounds like a better way to go.

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/Config.uk
skuenzer: 👍🏻

lib/uklibid/include/uk/libinfo.h
skuenzer: Sorry, you are right 👍🏻

lib/uklibid/Config.uk
skuenzer: At the moment, `ukdebug` is already a dependency that is always on (e.g., selected by every platform). As part of plat re-arch we should get rid of this hard dependency by providing stubs when the library is off. This would then also apply for `uklibid` and we can build it without `ukdebug`.

lib/uklibid/include/uk/libinfo.h
skuenzer: 👍🏻

lib/uklibid/include/uk/libinfo.h
skuenzer: I thought about this, too, both actually. My problem (but it isn't an excuse) was that we do not have fix width integer definitions in assembler, we would need to provide them with `<uk/arch/types.h>`. Because of the additional requirement of endianess, I thought the way to go about it could be by detecting the image architecture (e.g., parsing ELF headers). Both information (integer width and endianess) can be derived from this for external tools and internally we have the types matching anyways. What do you think? I am okay with adding assembler equivalent types for `__u16`, `__u8`, in fact this may make dealing with binary image formats easier, when there is no simple header for detecting the architecture.

support/scripts/uk-libinfo.py
skuenzer: At the moment, the section is always part of the image. I agree this should be configurable and yes a magic is needed for binary images where we do not have the ELF section headers anymore. I thought that this work we can do iteratively later.

lib/uklibid/include/uk/libinfo.h
skuenzer: Yes, exactly my thought. Something obvious we will need, could be usable already manually, but automation I'd add later.

kubanrob: I assume that licensing can not 100% figured out by automated scanning, so this probably will need human oversight.

A feasible solution would be to let each library set this field in its Makefile, maybe depending on the config options.

lib/uklibid/libinfo.S
```diff
@@ -0,0 +1,55 @@
...
+
+.pushsection .uk_libinfo, "a"
+	.align 1
+3:	.int	4f - 3b			/* len */
+	.short	UKLI_LAYOUT		/* version */
+
+#if __GLOBALINFO__
+	/* Only one global header per unikernel is generated */
+	/* NOTE: Global metadata do not have a libname record */
+#if CONFIG_LIBUKLIBID_INFO_UKFULLVERSION
```
kubanrob, line 25: There is the good chance that you also want to know the version of Unikraft a library is compiled with, in order to spot inconsistencies.

kubanrob (commented):
I left some initial comments.

Apart from that, I think it would be helpful to have some attribute in the metadata that would group libraries together.

For example, if I want to get a quick overview, I do not need the list of all Unikraft internal libraries. But I need to know that they are consistent (for example same git description, ...). Similar if a library/application adds multiple libraries.

However, I am not aware of a variable in the Unikraft build system that would be immediately suited for that. Maybe the name of the directory the `Makefile.uk` of the library is in, or some canonical name that needs to be added by the library itself?

lib/uklibid/include/uk/libinfo.h
skuenzer: Yes, I had something along these lines in my head, too. Maybe this could be done with a `Library.uk`mechanism, alternatively, we can also scan for SPDX headers. I left it open as future work.

lib/uklibid/libinfo.S
skuenzer: This is a good point. I am going to add this as another config option.

lib/uklibid/Makefile.uk
skuenzer: 👍🏻

lib/uklibid/include/uk/libinfo.h
skuenzer: 👍🏻

lib/uklibid/include/uk/libinfo.h
skuenzer: I updated the code to introduce assembler equivalents for `__u8`, `_u16`, `__u32` and so on...

lib/uklibid/Config.uk
skuenzer: Overall, yes, we were saying the same thing 👍🏻

LucaSeri:
image
Currently working on https://github.com/unikraft/kraftkit/issues/857. Please let me know if there will be more changes to this format.

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,113 @@
...
+	config LIBUKLIBID_INFO_COMPILEDATE
+		bool "Compile date"
+		help
+			This option stores the compile date and time for each
+			library and the unikernel image.
+
+	config LIBUKLIBID_INFO_COMPILEDBY
+		bool "Compiled by"
+		help
+			Name of the person/entity that did the compilation.
```
michpappas, line 95: This is now automatically set to `user@host`, so maybe the help string can be updated accordingly.

Also when looking at the generated menu via `menuconfig` it looks more like the `Compiled by` is a field that enables the ones right below that set it to a custom value, rather than a field that automatically generates something. That is because the `LIBUKLIBID_INFO_COMPILEDBY` option is overloaded with two semantics: 1. It enables `LIBUKLIBID_INFO_COMPILEDBY_TEXT`, which `depends` on it, and 2. Sets the default value. I think these cases are handled with a `menu` construct, ie:

```
menu "Compiled by"

config LIBUKLIBID_INFO_COMPILEDBY
    bool "user@hostname"
    help

config LIBUKLIBUD_INFO_COMPILEDBY_TEXT
    string "custom"
    help

endmenu
```

PS: Notice in the above example I only used a single option for the custom field, as one field is enough for the user to provide the full string of `user@host` or even just an alternative format they may prefer to use (eg a single organization name).

lib/uklibid/Config.uk
```diff
@@ -1,10 +1,113 @@
...
+
+		config LIBUKLIBID_INFO_LIB_UKCODENAME
+			bool "Unikraft codename"
+			help
+				If enabled, the Unikraft release codename is
+				stored as part of each library metadata
+				information.
+	endmenu
+
+	config LIBUKLIBID_INFO_COMPILER
```
michpappas: These options are neither global nor per library, both here and in `libinfo.S`. What is the rationale for that? Moreover, with the possibility of external libraries provided as prebuilts, it is possible that a library provided as a binary all the compiler info may be different than the rest of the build. Therefore, should these be considered as per-library data too?

support/scripts/uk-libinfo.py
```diff
@@ -0,0 +1,265 @@
...
+	def get_recs(self):
+		return self._recs
+
+	def libname(self):
+		return self._libname
+
+def parse_uklibinfo(data, byteorder=sys.byteorder):
+	if not data:
+		raise Exception("No uklibinfo data found for parsing")
+	hdr_hdr_len = (4 + 2)
```
michpappas: Some explanatory comments, similarly to the ones provided in similar cases below (eg line 181), would be useful here.

michpappas (changes requested):
Hi @skuenzer, thanks for the updates. Besides a few minor comments, it looks great.

**Update:** Looks like the changes in `types.h` break the Xen build 🤔 
```
/home/mpp/devel/unikraft/uk_helloworld/unikraft/plat/xen/x86/entry64.S: Assembler messages:
/home/mpp/devel/unikraft/uk_helloworld/unikraft/plat/xen/x86/entry64.S:51: Error: no such instruction: `_word hypercall_page'
```
Seems that `uk/asm/types.h` needs to be moved back to where it was, ie outside the `__ASSEMBLY__` directive, as it defines `_WORD` for the assembly files.

lib/uklibid/Config.uk
skuenzer: 👍🏻  I agree, I am going to change it. Please check the update: I will add the "Compiled by" option as a `menuconfig` submenu.

lib/uklibid/Config.uk
skuenzer: This info was stored autoamtically stored in both places: per-library and globally. However, to reduce confsion, I am going to change it and add the configuration items under both, the global configuration and per-library. I think this gives you more flexibility and it is more clear where the information ends up.

support/scripts/uk-libinfo.py
skuenzer: 👍🏻  Sure, I'll add.

michpappas (approved):
Looks good, thanks.

@razvand please be aware this PR has dependencies that need to be merged first.

skuenzer:
> image
> 
> Currently working on https://github.com/unikraft/kraftkit/issues/857. Please let me know if there will be more changes to this format.

Really nice work. Format will stay, we will just add more items. Latest addition was compile options (LTO, DCE, PIE). Could be useful to print as well.

lib/uklibid/Makefile.uk
```diff
@@ -38,4 +38,10 @@ LIBUKLIBID_SRCS-y += $(LIBUKLIBID_BASE)/exportsyms.awk>.uk
 LIBUKLIBID_EXPORTSYMS_AWKINCLUDES-y += $(LIBUKLIBID_BUILD)/libraries.in
 LIBUKLIBID_EXPORTS-y += $(LIBUKLIBID_BUILD)/exportsyms.uk
 
+EACHOLIB_SRCS-$(CONFIG_LIBUKLIBID_INFO) += $(LIBUKLIBID_BASE)/libinfo.S|libuklibid
```
kubanrob, line 42: This will not be recompiled with if any other file of the library is altered. So the information embedded will only be correct for a clean build. Probably that is currently only relevant for the date/time?

```
$ make
  GEN     libuklibid: libraries.in.new
  CP      libuklibid: libraries.in
  CP      config
  CC      libukdebug: outf.o
  LD      libukdebug.ld.o
  OBJCOPY libukdebug.o
  LD      unikraft_qemu-arm64.dbg
  SCSTRIP unikraft_qemu-arm64
  UKBI    unikraft_qemu-arm64.bootinfo
  OBJCOPY unikraft_qemu-arm64.bin
  MKLINUX unikraft_qemu-arm64.img
  GZ      unikraft_qemu-arm64.gz
```

skuenzer: Yeah, at the moment a weak point regarding DATE and TIME. We'd need to add another build system option to retrigger builds if something else in a lib changes. For fresh builds this is not an issue and PR #1119 is at least retriggering a rebuild on config changes (not code changes). I'd move this to a improvement at a later step later, because under certain circumstances the DATE and TIME information is correct if used (e.g., by a CI/CD doing a fresh build).

kubanrob: Fair enough. It might be good to document this limitation, because it can be surprising for somebody just trying to use this feature.

kubanrob (approved):
Looks like a solid foundation, thank you Simon!

lib/uklibid/Makefile.uk
skuenzer: Nevermind, I just had the idea to handover the current time and date as compiler argument. I changed this in the PR. As soon as we properly detect changes of compile flags, the date and time should be always up to date.

lib/uklibid/Config.uk
skuenzer: I change it to auto-detected username (user@host) but keeping a configuration option to overwrite this with a custom string. The idea is that this could be useful for CI/CD system or for the case user@host should not be used as this information.

Regarding your second point: We can add extra free later with another PR by using the `UKLI_REC_COMMENT` record. I had the same thoughts already. We could even have multiple `UKLI_REC_COMMENT` records per library if needed.

support/scripts/uk-libinfo.py
```diff
@@ -0,0 +1,272 @@
...
+##
+if __name__ == '__main__':
+	parser = argparse.ArgumentParser(description='Unikraft library metadata tool')
+	parser.add_argument("file", help="Kernel or library to investigate")
+	(opt, rem_args) = parser.parse_known_args()
+
+	try:
+		uklibinfo = uklibinfo_elf_load(opt.file)
+	except Exception as e:
+		print("{}: Failed to load uklibinfo data: {}".format(opt.file, str(e)))
```
kubanrob, line 252: This fails for me when looking at an arm64 binary: `build/unikraft_qemu-arm64: Failed to load uklibinfo data: Magic number does not match`

I think this is a limitation of the elf loader ([User guide](https://github.com/eliben/pyelftools/wiki/User's-guide)), not the rest of the implementation:
> The current focus of the library is on Intel's x86 and x64 architectures, with some ARM support added recently.

kubanrob (approved):

support/scripts/uk-libinfo.py
skuenzer: Hum, sounds like this. For sure it has to do with [pyelftools](https://github.com/eliben/pyelftools). Idea was to add later own magic numbers that mark beginning and end of the uklibinfo section. This way we can even parse the information without parsing ELF which is good for some machine images that aren't ELF.
Additionally, `kraftkit` tooling will get embedded support for this, too. So, I think the current limitation of the python script is not a big deal longer term.

razvand (approved):

Issue #47, opened 2020-06-24
Track changes of compile flags


skuenzer:
The build system is currently not taking changes of compile flags or build commands into account for detecting which objects need to be recompiled. We have already `fixdep` taken from Linux so we should be able to adopt the build rules.

kubanrob:
@skuenzer

If I remember correctly, fixdep has a parameter that is the whole command line[1]. Unfortunately this can be very long, so in #663 we decided to replace it with the path to the command file [2].

If this problem is a regression since the command file was adopted, this might be it.

Previously I proposed that we might pass a hash of the command argument to fixdep, that would change whenever a parameter changes [3], but I'm not sure if that is sufficient to fix the problem.

[1] https://github.com/unikraft/unikraft/pull/663#discussion_r1036064275
[2] https://github.com/unikraft/unikraft/pull/663#issuecomment-1332316757
[3] https://github.com/unikraft/unikraft/pull/663#discussion_r1034040130

PR #1119, opened 2023-09-30
build: Recompile everything on `.config` changes (workaround)

skuenzer:
### Description of changes

Unfortunately, our build system is still not able to re-trigger the compilation if only the command line of a compilation unit changed due to a configuration change. This can result in broken builds if the user did not call `make clean` in between.

This commit introduces a workaround that attempts to improve the user experience by triggering a recompile when the configuration is changed. This makes this cleaning step unecessary.

It is intended that commit "build: Recompile on `.config` changes (workaround)" gets reverted as soon as the build system can properly detect changes to the compiler command line.

This PR depends on PR #1070.
This is a workaround for issue #47.

### DIFF
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -819,7 +819,12 @@ $(error buildrule_$(call fileext,$(strip $(2))) is not defined: Failed to instal
 $(call buildrule_$(call fileext,$(strip $(2))),$(strip $(1)),$(strip $(2)),$(strip $(3)),$(strip $(4)),$(strip $(5)))
 
 # Default buildrule_* dependencies
-$(4): $(call qstrip,$($(call vprefix_src,$(1),$(2),$(3),CDEPS)) $($(call vprefix_src,$(1),$(2),$(3),CDEPS-y))) | preprocess
+# FIXME: We add $(KCONFIG_AUTOHEADER) in order to forcefully retrigger compiling
+#        whenever there is a config change. This is a workaround until we can
+#        handle changes that affect the compile command lines properly.
+$(4): $(call qstrip,$($(call vprefix_src,$(1),$(2),$(3),CDEPS)) \
+      $($(call vprefix_src,$(1),$(2),$(3),CDEPS-y))) $(KCONFIG_AUTOHEADER) \
+      | preprocess
 endef
 
 #################################################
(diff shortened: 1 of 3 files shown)


StefanJum (approved):
This works fine, thank you. I'll add the tag after #1070 gets in and this gets rebased.

support/build/Makefile.rules
```diff
@@ -813,7 +813,7 @@ $(error buildrule_$(call fileext,$(strip $(2))) is not defined: Failed to instal
 $(call buildrule_$(call fileext,$(strip $(2))),$(strip $(1)),$(strip $(2)),$(strip $(3)),$(strip $(4)),$(strip $(5)))
 
 # Default buildrule_* dependencies
-$(4): $(call qstrip,$($(call vprefix_src,$(1),$(2),$(3),CDEPS)) $($(call vprefix_src,$(1),$(2),$(3),CDEPS-y))) | preprocess
+$(4): $(call qstrip,$($(call vprefix_src,$(1),$(2),$(3),CDEPS)) $($(call vprefix_src,$(1),$(2),$(3),CDEPS-y))) $(UK_CONFIG_OUT) | preprocess
```
jschlumpp: I think this should also include a comment about the workaround (Same for the next hunk)

support/scripts/configupdate
```diff
@@ -25,18 +25,12 @@ NL=$'\n'
 # Change of architecture?
 _comparesym "CONFIG_ARCH_" "${UK_CONFIG}" "${UK_CONFIG_OLD}" || UPDATES+="*** - CPU architecture changed${NL}"
 
-# Change of CPU options?
```
michpappas, line 34: Since the full rebuild is only a transient measure, won't these be required once we move to a proper fix?

support/build/Makefile.rules
michpappas: +1

michpappas (changes requested):

support/build/Makefile.rules
michpappas: +1

skuenzer: I added 👍🏻

support/scripts/configupdate
skuenzer: Even after transiting, we should be able to properly handle re-compilation on updated compile command line. This means that CPU options and optimization should be covered. Actually, also the architecture should be covered (since we call a different compiler). I thought of even removing the architecture line but I don't know if we should keep it due external libraries. I am currently imagining a library that downloads different sources depending on the architecture. 🤔

michpappas (approved):

razvand (approved):

PR #1120, opened 2023-09-30
plat/common/acpi: Check for offline `GICC`'s and fix `VGIC` `IRQ Mode` flag

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

    
When we probe for available, valid, GICC's we only look for those
that are already online and initialized, totally forgetting about
those that are usable but offline. Therefore, check for powered off
GICC's as well, since, during probe, we re-initialize GICC's anyway.

As a side fix,  the bitfield `ACPI_MADT_GICC_FLAGS_VGIC_IRQ_MODE` of `GICC` had
the wrong value: it usedvalue `0x3` instead of the actual
value `(1 << 2)`. Therefore, fix it.

### DIFF
--- a/plat/common/arm/lcpu.c
+++ b/plat/common/arm/lcpu.c
@@ -118,7 +118,8 @@ static int do_arch_mp_init(void *arg __unused)
 		m.h = (struct acpi_subsdt_hdr *)(madt->entries + off);
 
 		if (m.h->type != ACPI_MADT_GICC ||
-		    !(m.gicc->flags & ACPI_MADT_GICC_FLAGS_EN))
+		    (!(m.gicc->flags & ACPI_MADT_GICC_FLAGS_EN) &&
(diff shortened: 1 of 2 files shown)


razvanvirtan:
Hey @mogasergiu, thank you for this!

razvanvirtan (commented):
Hey @mogasergiu, thank you for this!
On my side, only a couple of suggestions regarding the commit messages:

- for the first commit, the old wrong value was 0x3, so I think in the message we should have something like `...its bitfield offset (0x3)...`
- for the second commit, the order of the commas seems a bit strange to me, WDYT about `...available, valid GICC's, we only look...`?

razvanvirtan (approved):
All good on my side.

michpappas (approved):
LGTM 👍🏼 

razvand (approved):

PR #1127, opened 2023-10-11
build: Use ISO 8601 format for `$(HOSTUTC)`

skuenzer:
### Description of changes

This PR makes the time format consistent for all host localizations and systems with the ISO 8601 format for UTC.

PR #1131, opened 2023-10-12
lib/uklibid: Clang-compatible compile option flags

### LINKED ISSUE #1128: `lib/uklibid`: Clang issue when assembling `lib/uklibid/libinfo.S`
### Describe the bug

Using Clang to build [`app-helloworld`](https://github.com/unikraft/app-helloworld) results in error after PR #1117 . The PR introduces `lib/uklibid`.

This is caused by the assembling of the `lib/uklibid/libinfo.S` file, namely [lines `67-70`](https://github.com/unikraft/unikraft/blob/staging/lib/uklibid/libinfo.S#L67).

### Steps to reproduce

Use the steps below to reproduce the issue:

```console
git clone https://github.com/unikraft/app-helloworld helloworld
cd helloworld/
./scripts/setup.sh
UK_DEFCONFIG=$(pwd)/defconfigs/qemu-x86_64 make defconfig
make prepare
make CC=clang
```

### Expected behavior

The Clang-based build should be successful.

x86_64

kvm

### Relevant log output

```shell
/tmp/libinfo-1e83aa.s:51:52: error: expected relocatable expression
 .align 1 ; 1: .short 0x000F ; .int 2f - 1b ; .int ( (!(!1) * (0x01 << 0)) | (!(!__LIBUKLIBID_COMPILEOPT_DCE__) * (0x01 << 1)) | (!(!__LIBUKLIBID_COMPILEOPT_LTO__) * (0x01 << 2))) ; 2: .align 1
                                                   ^
make[3]: *** [/home/unikraft/maintainer-tools/workdir/unikraft/support/build/Makefile.build:49: /home/unikraft/maintainer-tools/workdir/apps/helloworld/workdir/build/libkvmpl031/libinfo.libuklibid.o] Error 1
```

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_LIBUKLIBID=y`
 - `CONFIG_LIBUKLIBID_INFO=y`

Use Clang as compiler: `make CC=clang`

### Description of changes

Clang's preprocessor seems to behave differently when computing an integer based on undefined macros. For example, the following expression:
```
  (!(!__LIBUKLIBID_COMPILEOPT_PIE__) * (0x01 << 0))
```
resolves to 0x0 on GCC's GAS if `__LIBUKLIBID_COMPILEOPT_PIE__` is not defined. Clang is throwing an compile error because of the missing definition. This PR adopts `Makefile.uk` of `lib/uklibid` to defined the macros with `0` when the requested feature is not enabled.

This PR fixes the issue #1128 .


StefanJum (commented):
This fixes the issue if #1127 is also used.

However, is `Makefile.uk`  the right place to do this? I feel like if this will occur with any other symbols, then it could get quite ugly.
Maybe adding an `ifndef define 0` in the source code, where the symbols are actually declared / used coud work better?

skuenzer:
> This fixes the issue if #1127 is also used.
> 
> However, is `Makefile.uk` the right place to do this? I feel like if this will occur with any other symbols, then it could get quite ugly. Maybe adding an `ifndef define 0` in the source code, where the symbols are actually declared / used coud work better?

I pushed an update and moved the logic to `libinfo.S`

kubanrob (approved):
The change seems good to me. I agree that it makes sense to move the defines into the makefile, especially given that the config variable already have all the information that is needed.

StefanJum (approved):
All good, thanks.

razvand (approved):

PR #1138, opened 2023-10-15
plat/kvm: Adapt EFI boot to new APIC driver

kha-dinh:
Fixed the header include, and also added config guard for APIC-related operations in `efi_post.c`.

### Prerequisite checklist

### Base target

### Description of changes

The re-arched upstream APIC driver breaks EFI compilation. This PR adds minor changes to EFI boot to make it work.

### DIFF
--- a/plat/kvm/x86/efi_post.c
+++ b/plat/kvm/x86/efi_post.c
@@ -7,7 +7,6 @@
 #include <uk/arch/paging.h>
 #include <uk/plat/common/bootinfo.h>
 #include <uk/plat/lcpu.h>
-#include <x86/apic_defs.h>
 #include <x86/cpu.h>
 
 /* Slave Controller Edge/Level Triggered Register */
@@ -17,6 +16,11 @@
 
 /* Initial Count Register (for Timer) */
 #define LAPIC_TMICT					0xFEE00380
+
+/* APIC MSR Register */
+/* TODO: This should be removed once our APIC driver is able to handle LAPIC */
+#define LAPIC_MSR_BASE					0x01B
+#define LAPIC_BASE_EN					(1 << 11)
 
 /* Master/Slave PIC Data Registers */
 #define PIC1_DATA					0x21
@@ -51,8 +55,8 @@ static inline void lapic_timer_disable(void)
 	__u32 eax, edx;
 
 	/* Check if APIC is active */
-	rdmsr(APIC_MSR_BASE, &eax, &edx);
-	if (unlikely(!(eax & APIC_BASE_EN)))
+	rdmsr(LAPIC_MSR_BASE, &eax, &edx);
+	if (unlikely(!(eax & LAPIC_BASE_EN)))
 		return;
 
 	/* Zero-ing out this register disables the LAPIC Timer */


jschlumpp (approved):
Thanks

plat/kvm/x86/efi_post.c
```diff
@@ -47,6 +49,7 @@ static inline void unmask_8259_pic(void)
 /* UEFI enables the LAPIC Timer to run periodic routines, usually at 10KHz */
 static inline void lapic_timer_disable(void)
 {
+#if CONFIG_LIBUKINTCTLR_APIC
```
mogasergiu: Hmm, weird how I did not bump into this error. But it makes sense now :+1:. Problem is, this means that the LAPIC Timer would only get disabled if the APIC driver is enabled (which seems to happen implicitly for SMP builds only), which is not ok. Let's follow the PIC example (see the PIC macros defined above) and declare duplicate macros for `APIC_MSR_BASE` and `APIC_BASE_EN` in this file and add a comment with a TODO to let the reader know this should be removed once we handle the LAPIC Timer in our APIC driver.

mogasergiu: > which seems to happen implicitly for SMP builds only

Ahh.. now I know I did not bump into this... I always have SMP enabled... :laughing:

plat/kvm/x86/efi_post.c
```diff
@@ -7,7 +7,9 @@
 #include <uk/arch/paging.h>
 #include <uk/plat/common/bootinfo.h>
 #include <uk/plat/lcpu.h>
-#include <x86/apic_defs.h>
+#if CONFIG_LIBUKINTCTLR_APIC
+#include <uk/intctlr/apic.h>
```
mogasergiu: Do we still need this header? We probably only need the one for the `rdmsr` instead :thinking:, unless it is already included.

plat/kvm/x86/efi_post.c
```diff
@@ -44,6 +46,10 @@ static inline void unmask_8259_pic(void)
 	outb(PIC2_DATA, PIC2_DATA_DEFAULT_MASK);
 }
 
+/* Redeclaration of APIC MSR registers */
+/* TODO: This should be removed once our APIC driver is able to handle LAPIC */
+#define APIC_MSR_BASE					0x01b
+#define APIC_BASE_EN					(1 << 11)
```
mogasergiu: Awesome :+1:. Thank you :pray:.

Let's move these, as well as the comments, below `LAPIC_TMICT` and also add the `L` letter so that they look somewhat more uniform, i.e. `LAPIC_MSR_BASE`/`LAPIC_BASE_EN`.

mogasergiu (changes requested):
Thank you for the quick update :camera_flash:! Two very minor comments. Also keep in mind that the commit message may need re-adjustment :smile:, if you agree with the suggested changes.

plat/kvm/x86/efi_post.c
kha-dinh: Done :sunglasses:.

plat/kvm/x86/efi_post.c
mogasergiu: What about this? :thinking: 
```diff
diff --git a/plat/kvm/x86/efi_post.c b/plat/kvm/x86/efi_post.c
index 63cb82a6..81d261d9 100644
--- a/plat/kvm/x86/efi_post.c
+++ b/plat/kvm/x86/efi_post.c
@@ -7,9 +7,6 @@
 #include <uk/arch/paging.h>
 #include <uk/plat/common/bootinfo.h>
 #include <uk/plat/lcpu.h>
-#if CONFIG_LIBUKINTCTLR_APIC
-#include <uk/intctlr/apic.h>
-#endif /* CONFIG_LIBUKINTCTLR_APIC */
 #include <x86/cpu.h>
 
 /* Slave Controller Edge/Level Triggered Register */
 ```
 Seems to work for me with both SMP enabled and disabled
 
 EDIT: Btw I noticed you also added the `/* CONFIG... */ which is what we are trying to have for `#if` guards. Thanks a lot for the careful thought :pray:

File: plat/kvm/x86/efi_post.c [hunk 4]
```diff
@@ -18,6 +20,11 @@
 /* Initial Count Register (for Timer) */
 #define LAPIC_TMICT					0xFEE00380
 
+/* APIC MSR Register */
+/* TODO: This should be removed once our APIC driver is able to handle LAPIC */
+#define LAPIC_MSR_BASE					0x01b
```
mogasergiu: You're going to hate me for this nitpick :grin:, but I really like **muh** consistency<sup>TM</sup>: can we do this with a capital :laughing: `0x01B` like the other macros in this file? Sorry :sweat_smile:, but better now during review. Other than that, everything looks good. Thank you again for catching this :muscle:.

mogasergiu (changes requested):
Sorry, I am probably being annoying here :sweat_smile:, but two more minor comments.

plat/kvm/x86/efi_post.c
kha-dinh: I left the include there because there might be dependency in the future. Removed it.

mogasergiu (approved):
Awesome :love_you_gesture:! Thanks a lot for your work and cooperation!

razvand (approved):

PR #1145, opened 2023-10-23
lib/vfscore: Introduce `ukopts`, a new `vfs.fstab` option and `mkmp`, a `ukopts` option

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Allow the `vfs.fstab` option arguments to receive an additional                                                                                                       
option that is Unikraft specific and that enables one to change                                                                                                       
a volume's mount behavior depending on the passed option.                                                                                                             
                                          
Thus, now the `vfs.fstab` syntax is as follows:                                                                                                                       
```
vfs.fstab=[                                                              
      "<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"                                                                                           
       "<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"                                                                                           
      ...                                                              
]                             
```
                                          
The `ukopts` options will have the same separator as `fstab`, that                                                                                                    
being `:`. 

Furthermore, in case someone may want to mount a filesystem to a directory                                                                                                         
that does not currently exist in the rootfs, allow them to do so                                                                                                      
without having to rebuild the rootfs by exposing a `ukopts`                                                                                                           
option called `mkpath`. This option will let `fstab` know that                                                                                                        
it must ensure that the mount path exists before attempting to                                                                                                        
mount.

### DIFF
--- a/lib/vfscore/automount.c
+++ b/lib/vfscore/automount.c
@@ -56,6 +56,11 @@
 #include <sys/stat.h>
 #endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS */
 
+#if CONFIG_LIBVFSCORE_FSTAB
+#define LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP			':'
+#define LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP			','
+#endif /* CONFIG_LIBVFSCORE_FSTAB */
+
 struct vfscore_volume {
 	/* Volume source device */
 	const char *sdev;
@@ -67,16 +72,52 @@ struct vfscore_volume {
 	unsigned long flags;
 	/* Mount options */
 	const char *opts;
+#if CONFIG_LIBVFSCORE_FSTAB
+	/* Unikraft Mount options, see vfscore_mount_volume() */
+	char *ukopts;
+#endif /* CONFIG_LIBVFSCORE_FSTAB */
 };
+
+#if CONFIG_LIBVFSCORE_FSTAB
+/**
+ * vv->ukopts must follow the pattern below, each option separated by
+ * the character defined through LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP (e.g. with
+ * LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP = ','):
+ *	[<ukopt1>,<ukopt2>,<ukopt3>,...,<ukoptN>]
+ *
+ */
+static int vfscore_volume_process_ukopts(struct vfscore_volume *vv)
+{
+	char *o_curr, *o_next;
+	int rc;
+
+	UK_ASSERT(vv);
+
+	o_curr = vv->ukopts;
+	while (o_curr) {
+		o_next = strchr(o_curr, LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP);
+		if (o_next) {
+			*o_next = '\0';
+			o_next++;
+		}
+
+		o_curr = o_next;
+	}
+
+	return 0;
+}
+#endif /* CONFIG_LIBVFSCORE_FSTAB */
 
 static inline int vfscore_mount_volume(struct vfscore_volume *vv)
 {
+#if CONFIG_LIBVFSCORE_FSTAB
+	vfscore_volume_process_ukopts(vv);
+#endif /* CONFIG_LIBVFSCORE_FSTAB */
+
 	return mount(vv->sdev, vv->path, vv->drv, vv->flags, vv->opts);
 }
 
 #ifdef CONFIG_LIBVFSCORE_FSTAB
-
-#define LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP			':'
 
 static char *vfscore_fstab[CONFIG_LIBVFSCORE_FSTAB_SIZE];
 
@@ -87,12 +128,12 @@ UK_LIBPARAM_PARAM_ARR_ALIAS(fstab, &vfscore_fstab, charp,
 /**
  * Expected command-line argument format:
  *	vfs.fstab=[
- *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
- *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
+ *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"
+ *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"
  *		...
  *	]
  * These list elements are expected to be separated by whitespaces.
- * Mount options and flags are optional.
+ * Mount options, flags and Unikraft mount options are optional.
  */
 static char *vfscore_fstab_next_volume_arg(char **pos)
 {
@@ -132,14 +173,21 @@ static void vfscore_fstab_fetch_volume_args(char *v, struct vfscore_volume *vv)
 
 	vv->path = vfscore_fstab_next_volume_arg(&pos);
 	UK_ASSERT(vv->path);
+	if (unlikely(vv->path[0] != '/')) {
+		uk_pr_err("Mountpoint \"%s\" is not absolute\n", vv->path);
+		vv->sdev = NULL;
+		vv->path = NULL;
+		vv->drv = NULL;
+		vv->opts = NULL;
+		vv->ukopts = NULL;
+		vv->flags = 0;
+		return;
+	}
 
 	vv->drv = vfscore_fstab_next_volume_arg(&pos);
 	UK_ASSERT(vv->drv);
 
-	if (!pos)
-		return;
-
-	/* Allow empty flags: "<src_dev>:<mntpoint>:<fsdriver>::<opts>" */
+	/* Allow empty flags: "<src_dev>:<mntpoint>:<fsdriver>::<opts>:" */
 	flags = vfscore_fstab_next_volume_arg(&pos);
 	if (flags && flags[0] != '\0')
 		vv->flags = strtol(flags, NULL, 0);
@@ -149,6 +197,18 @@ static void vfscore_fstab_fetch_volume_args(char *v, struct vfscore_volume *vv)
 	vv->opts = vfscore_fstab_next_volume_arg(&pos);
 	if (vv->opts && vv->opts[0] == '\0')
 		vv->opts = NULL;
+
+	vv->ukopts = vfscore_fstab_next_volume_arg(&pos);
+	if (vv->ukopts && vv->ukopts[0] == '\0')
+		vv->ukopts = NULL;
+
+	uk_pr_info("Fetched VFSCore Volume args: %s:%s:%s:%lx:%s:%s\n",
+		   vv->sdev == NULL ? "none" : vv->sdev,
+		   vv->path == NULL ? "none" : vv->path,
+		   vv->drv == NULL ? "none" : vv->drv,
+		   vv->flags,
+		   vv->opts == NULL ? "none" : vv->opts,
+		   vv->ukopts == NULL ? "none" : vv->ukopts);
 }
 #endif /* CONFIG_LIBVFSCORE_FSTAB */
 
--- a/lib/vfscore/automount.c
+++ b/lib/vfscore/automount.c
@@ -48,12 +48,12 @@
 #include <uk/init.h>
 #include <uk/libparam.h>
 #include <uk/plat/memory.h>
+#include <sys/stat.h>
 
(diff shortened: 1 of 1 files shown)


nderjung (approved):
Tested my side, this works. :) Thanks!

lib/vfscore/automount.c
```diff
@@ -67,10 +71,47 @@ struct vfscore_volume {
...
+static int vfscore_volume_process_ukopts(struct vfscore_volume *vv)
+{
+	char *o_curr, *o_next;
+	int rc;
+
+	UK_ASSERT(vv);
+
+	o_curr = vv->ukopts;
+	while (o_curr) {
+		o_next = strchr(o_curr, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
```
mogasergiu: TODO (for me before merge, so I don't forget): use another separator, different from `LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP` so that they don't conflict.

lib/vfscore/automount.c
```diff
@@ -99,6 +145,16 @@ static int vfscore_volume_process_ukopts(struct vfscore_volume *vv)
 			o_next++;
 		}
 
+		/* First check is so we do not run `mkpath` on `/` */
+		if (vv->path[1] != '\0' && !strcmp(o_curr, "mkpath")) {
```
skuenzer: I think it makes sense to add also a check that a mountpoint has always be specified with `vv->path[0] == '/'`. This should be independently checked of `mkpath`. If the path string does not start with `/` we should fail booting. Additionally, if the `/` check passed, we know it is safe to access `vv->path[1]`.

lib/vfscore/automount.c
```diff
@@ -78,11 +78,57 @@ struct vfscore_volume {
...
+}
+
 /**
  * vv->ukopts must follow the pattern below, each option separated by
  * the character defined through LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP (e.g. with
  * LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP = ':'):
  *	[<ukopt1>:<ukopt2>:<ukopt3>:...:<ukoptN>]
+ *
+ * Currently implemented, Unikraft Mount options:
+ * - mkpath	A custom mount path may be specified. If it does not exist
```
skuenzer: I tend to name it `mkmp` for `make mountpoint` but `mkpath` is also fine. The description I'd change to:
```
Ensures that the specified mount point exists. If it does not exist in the current vfs, the directory structure is created.
```

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+		return rc;
+
+	return 0;
+}
+
+/**
+ * vv->ukopts must follow the pattern below, each option separated by
+ * the character defined through LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP (e.g. with
+ * LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP = ':'):
+ *	[<ukopt1>:<ukopt2>:<ukopt3>:...:<ukoptN>]
```
skuenzer: Somehow, I think we should comma-separate `ukopts` so that this option list looks similar to the mount options.

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+static int vfscore_volume_process_ukopts_do_mkpath(const char *path)
+{
+	const char *curri;
+	char *pos;
+	int rc;
+
+	UK_ASSERT(path);
+
+	curri = path;
+	pos = strchr(path + 1, '/');
```
skuenzer: Is the compiler not warning here? `strchr()` should return `const char *`

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+
+	while (pos) {
+		/* If we reached last directory, delete trailing `/` and stop */
+		if (pos[1] == '\0') {
+			*pos = '\0';
+			break;
+		}
+
+		/* Zero out the next '/' */
+		*pos = '\0';
```
skuenzer: Because `path` is `const char *`, we should not be allowed to simply do this operation. But probably we can just remove `const` from the `path` argument of this function.

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+		/* mkdir() with S_IRWXU */
+		rc = mkdir(curri, 0700);
+		if (rc && errno != EEXIST)
+			return rc;
+
+		/* Restore current '/' */
+		*pos = '/';
+
+		/* Now prepare for the next '/' in the path, if there */
+		pos = strchr(pos + 1, '/');
```
skuenzer: Maybe we could do a `do {} while` loop or even `for (;;)` to deduplicate the `strchr` operation?

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+		if (pos[1] == '\0') {
+			*pos = '\0';
+			break;
+		}
+
+		/* Zero out the next '/' */
+		*pos = '\0';
+
+		/* mkdir() with S_IRWXU */
+		rc = mkdir(curri, 0700);
```
skuenzer: `curri` is equal to `path`... 🤔   I guess, I am missing something because `curri` would always point to the first cut out substring of `path`. A subdirectory structure would not be correctly created?

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+	int rc;
+
+	UK_ASSERT(path);
+
+	curri = path;
+	pos = strchr(path + 1, '/');
+
+	while (pos) {
+		/* If we reached last directory, delete trailing `/` and stop */
+		if (pos[1] == '\0') {
```
skuenzer: I guess we also need to catch `pos[0] == '\0'` before we look into `pos[1]`. I am also thinking of inputs like `///path//with/too//many/slashes///`

skuenzer: I think we should also test and forbid `.` and `..` in the mountpoint path. For example:
```
/./subdir/../../subdir/again///
```
The function should return `-EINVAL` if we see something like this.

lib/vfscore/automount.c
```diff
@@ -67,10 +71,103 @@ struct vfscore_volume {
...
+
+		/* Restore current '/' */
+		*pos = '/';
+
+		/* Now prepare for the next '/' in the path, if there */
+		pos = strchr(pos + 1, '/');
+	}
+
+	/* We have reached the last directory in the path */
+	rc = mkdir(curri, 0700);
```
skuenzer: Can we get this also into the loop?

lib/vfscore/automount.c
skuenzer: We also assume here that path is longer than 1 Byte (`\0`).
I think we should add another assertion for this function:
```diff
 UK_ASSERT(path);
+UK_ASSERT(path[0] == '/');
```

lib/vfscore/automount.c
```diff
@@ -87,8 +184,8 @@ UK_LIBPARAM_PARAM_ARR_ALIAS(fstab, &vfscore_fstab, charp,
 /**
  * Expected command-line argument format:
  *	vfs.fstab=[
- *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
- *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>]"
+ *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"
+ *		"<src_dev>:<mntpoint>:<fsdriver>[:<flags>:<opts>:<ukopts>]"
```
skuenzer, line 200: We can also briefly mention that `flags`, `opts`, and `ukopts` can be empty as well.

skuenzer (changes requested):

lib/vfscore/automount.c
mogasergiu: Agreed, I am also going to add some more asserts in other places/functions.

lib/vfscore/automount.c
mogasergiu: I do mention this below

lib/vfscore/automount.c
mogasergiu: Ok, I did a quick workaround for having multiple `/`'s in the path. But our `mkdir`/`lookup` implmenetation does not handle such cases, which I find worth fixing sometime.

lib/vfscore/automount.c
mogasergiu: > Is the compiler not warning here? strchr() should return const char *

We generally have so many warnings that I missed this one.

mogasergiu:
Sorry, I know, lots of pushes, I messed up some rebasing :upside_down_face:

lib/vfscore/automount.c
```diff
@@ -48,36 +48,145 @@
...
+		 */
+		if (unlikely(prev_pos[1] == '.' &&
+			     /* /../ and /.. */
+			     ((prev_pos[2] == '.' &&
+			       (prev_pos[3] == '/' || prev_pos[3] == '\0')) ||
+				/* OR /./ and /. */
+			      (prev_pos[2] == '/' || prev_pos[2] == '\0')
+			     )))
+			uk_pr_err("Do not allow `/./` or `/../` in the path. "
+				  "Also do not allow overwriting .. or . "
```
skuenzer: I think this extra message is not needed, I find it a bit confusing.
Just keep something like `"'.' or '..' are not supported in mount paths.\n"`

lib/vfscore/automount.c
```diff
@@ -48,36 +48,145 @@
...
+				  "files\n");
```
skuenzer: The function should be left here with an error code, e.g. `-EINVAL`

lib/vfscore/automount.c
```diff
@@ -48,36 +48,145 @@
...
+
+		/* mkdir() with S_IRWXU */
+		rc = mkdir(path, 0700);
+		if (rc && errno != EEXIST)
+			return rc;
```
skuenzer: I think we could do `return -errno;`

lib/vfscore/automount.c
```diff
@@ -48,36 +48,145 @@
...
+
+		/* Restore current '/' */
+		if (pos)
+			*pos = '/';
+
+		/* Handle paths with multiple `/` */
+		while (pos && pos[1] == '/')
+			pos++;
+	} while (pos);
```
skuenzer, line 130: I think `pos` is never `NULL` here, correct? It can be `\0` here due to the double-slash `//` skip-loop but line 98 should already catch `pos == NULL` case.
I think you can use a `for (;;)` loop instead.

lib/vfscore/automount.c
skuenzer: 👍🏻

lib/vfscore/automount.c
```diff
@@ -132,14 +241,22 @@ static void vfscore_fstab_fetch_volume_args(char *v, struct vfscore_volume *vv)
 
 	vv->path = vfscore_fstab_next_volume_arg(&pos);
 	UK_ASSERT(vv->path);
+	if (unlikely(vv->path[0] != '/')) {
+		uk_pr_err("Invalid absolute path provided. Please have your "
+			  "mount point path be absolute.\n");
```
skuenzer: What about? `uk_pr_err("Mountpoint \"%s\" is not absolute\n", vv->path)`

skuenzer (changes requested):
Hi @mogasergiu, thanks a lot for your work. I just noticed some little things with the update but overall looks very good already!

lib/vfscore/automount.c
mogasergiu: It actually is for the last node of the path if it does not end with a `/`, which results in `strchr` returning a `NULL`.
E.g. say you have just entered the loop with the following:
`path = /path/to/file`, `pos = /file`, `prev_pos = /to/file`
In that case, `prev_pos` becomes `/file`, `strchr(pos + 1, '/');` returns `NULL` so `pos` becomes `NULL`. You now have: `path = /path/to/file`, `pos = NULL`, `prev_pos = /file`. I need the last `while` to still check for `pos != NULL` in case we have something like `/path/to/file///////` and then the last check on `pos != NULL` (the one of the `do { } while`) catches that case.

This is why in the previous version of this PR I had another `mkdir` outside the loop, to avoid having to do this step inside the loop.

skuenzer: You are right. All good then.

skuenzer (approved):
Thanks a lot for your work and updates!

PR #1148, opened 2023-10-26
lib/ukboot: Shutdown feature

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR introduces support for shutting down the system. The change consists of an extension of inittab. Each entry is extended to two function pointers: an initialization function pointer and a termination function pointer, which are handled in pairs and share the same priority. While initialization means iterating over the inittab and calling the initialization functions, termination is done by iterating over the inittab in reverse order.
Initialization and termination are handled by the "init" thread of `lib/ukboot`. With scheduling, PR introduces the option to start the `main()` of the application in a separate "main" thread. This allows shutdown requests to be handled independently of the state of `main()`. A shutdown request driver (e.g., ACPI button) or any other application thread can send a request to the "init" thread to initiate a shutdown.


include/uk/init.h
```diff
@@ -55,88 +79,117 @@ typedef int (*uk_init_func_t)(void);
...
-	__uk_inittab ## base ## prio ## _ ## fn = (fn)
+#define __UK_INITTAB_ENTRY(init_fn, term_fn, base, prio)		\
+	static const struct uk_inittab_entry				\
+	__used __section(".uk_inittab" #base #prio) __align(8)		\
+		__uk_inittab ## base ## prio ## _ ## init_fn ## _ ## term_fn = {\
+		.init = (init_fn),					\
+		.term = (term_fn)					\
+	}
 
-#define _UK_INITTAB(fn, base, prio)					\
```
jschlumpp, line 63: I would prefer to not change the existing interface and instead introduce a new macro that allows specifying the `term_fn`. Then it wouldn't break all existing (external) clients of the API.

lib/ukboot/boot.c
```diff
@@ -419,38 +488,49 @@ void ukplat_entry(int argc, char *argv[])
...
 
-	/**
-	 * Call termination functions from init table in reverse order
-	 */
-	/* NOTE: The init loop left `init_entry` at the position that is one
-	 *       step further after the last successfully initialized  entry
+#if CONFIG_LIBUKBOOT_MAINTHREAD
+static __noreturn void main_thread(void *a_argc, void *a_argv)
+{
+	int argc    = (int)((__uptr)a_argc);
```
mogasergiu: I know it looks nice, but I think we should maintain consistency from now on and have `=` separated by only one whitespeace end-to-end.

lib/uklock/include/uk/isr/semaphore.h
```diff
@@ -1,8 +1,34 @@
...
+
+static inline void uk_semaphore_up_isr(struct uk_semaphore *s)
+{
+	unsigned long irqf;
+
+	UK_ASSERT(s);
+
+	uk_spin_lock_irqsave(&(s->sl), irqf);
+	++s->count;
+#ifdef UK_SEMAPHORE_DEBUG
```
mogasergiu: why not `#if` (and closing `/* UK_SEMAPHORE_DEBUG */`).

lib/ukboot/shutdown_req.c
```diff
@@ -54,3 +55,17 @@ int uk_boot_shutdown_req_isr(enum ukplat_gstate target)
...
+{
+	enum ukplat_gstate request = (enum ukplat_gstate)data;
+	int rc;
+
+	rc = uk_boot_shutdown_req_isr(request);
+	if (unlikely(rc < 0))
+		return UK_EVENT_NOT_HANDLED;
+	return UK_EVENT_HANDLED_CONT;
+}
+UK_EVENT_HANDLER(UKPLAT_SHUTDOWN_EVENT, shutdown_req_handler);
```
mogasergiu, line 70: One question. What happens if there are multiple shutdown handlers registered? Are they executed based on a priority, like on Linux? E.g. UEFI shutdown routine > ACPI shutdown routine

mogasergiu (commented):
Hi Simon :wave:! Had a quick look, very clean PR! I like the semaphore idea! The only downside of this PR I can see is the shutdown event not being handled with highest priority (i.e. it has to wait to be scheduled in the round-robin fashion) which can be risky in the case of high emergency power management events, especially in a cooperative scheduler environment (maybe something like a deferred handler could solve this :pleading_face:)

Anyway, I left some minor comments. I will probably have another look in the near future :cowboy_hat_face:.

include/uk/init.h
skuenzer: The update requires a macro that sets both at the same time because this is the way how the table is constructed, the same priority value is used for initialization and termination (reverse order). This simplifies cleaning up in case of errors of one of the init function.
Of course we could argue, another macro could be provided as wrapper that sets only init but I wanted to have this explicitly. The fact that the system can shutdown should also make developers think about a termination. As soon this is upstreamed, we have a number of PR across libs that fix existing code so that disruption should be minimal.

lib/ukboot/boot.c
skuenzer: Is fine, I don't mind. I will send an update.

lib/uklock/include/uk/isr/semaphore.h
skuenzer: I'll update 👍🏻

lib/ukboot/shutdown_req.c
skuenzer: UEFI and ACPI would be triggers of the shutdown event queue. Whatever pulls the trigger first, made it actually.

mogasergiu: UEFI shutdown services are at the top of the hierarchy in the list of shutdown execution. An ACPI power button event may trigger it but even then, it's still the UEFI shutdown that should be done to shutdown the machine. If that fails or is somehow absent, then yes fallback to ACPI shutdown.

skuenzer:
> Hi Simon 👋! Had a quick look, very clean PR! I like the semaphore idea! The only downside of this PR I can see is the shutdown event not being handled with highest priority (i.e. it has to wait to be scheduled in the round-robin fashion) which can be risky in the case of high emergency power management events, especially in a cooperative scheduler environment (maybe something like a deferred handler could solve this 🥺)
> 
> Anyway, I left some minor comments. I will probably have another look in the near future 🤠.

Hi Sergiu, thanks a lot for your review. It is in fact true that a shutdown request can be ignored as long as a thread never releases the CPU which lets the init thread not being scheduled to forward a shutdown request to the application. However, I think this behavior is fine since we are speaking here about a shutdown request which the application could even ignore. So, my assumption is that this is a cooperative action. With preempt scheduling this might be even less of a problem and the hypervisor can still terminate non-cooperating virtual machines.

mogasergiu:
> can be ignored as long as a thread never releases the CPU

Yes you are right. I was however referring to the situation where this is not the next thread that will be executed on yield, because we do not have a priority scheduling schema, right?

>  hypervisor can still terminate non-cooperating virtual machines.

Yes, agreed, not something worth bothering with now :smiley:. I just thought it would be worth a mention :wink:.

lib/ukboot/boot.c
```diff
@@ -339,6 +341,9 @@ void ukplat_entry(int argc, char *argv[])
 	uk_sched_start(s);
 #endif /* !CONFIG_LIBUKBOOT_NOSCHED */
 
+	ictx.cmdline.argc = argc - kern_args;
```
mogasergiu: Hmm, isn't `kern_args` always `0`? Ditto for line below.

include/uk/init.h
```diff
@@ -126,26 +144,46 @@ typedef int (*uk_init_func_t)(struct uk_init_ctx *);
...
 	     (itr)++)
 
+/*
+ * Reverse iteration
+ * NOTE: The reverse iteration is intended for calling the termination functions
+ *       according to their priorities.
+ */
+#define uk_inittab_foreach_reverse(itr, inittab_start, inittab_end)	\
+	for ((itr) = (DECONST(struct uk_inittab_entry *,		\
+			 (&(inittab_end)))) - 1);			\
```
mogasergiu: Perhaps we could align this to the `DECONST` open paranthesis above?

include/uk/ctors.h
```diff
@@ -42,7 +42,7 @@
 extern "C" {
 #endif
 
-typedef void (*uk_ctor_func_t)(void);
+typedef void (*uk_ctor_func_t)();
```
mogasergiu, line 45: Why no arguments? Shouldn't it be the types of `argc`/`argv`?

lib/ukboot/boot.c
```diff
@@ -377,12 +392,65 @@ void ukplat_entry(int argc, char *argv[])
...
+		(*init_entry->term)(&tctx);
+	}
+
+	ukplat_terminate(tctx.target); /* does not return */
+}
+
+static inline int do_main(int argc, char *argv[])
+{
+#if CONFIG_LIBPOSIX_ENVIRON
+	char **envp;
```
mogasergiu, line 442: Aw man, it was almost an xmas tree style local variable declaration :cry:, maybe we can make it like that and slowly turn the codebase to follow that aesthetic pattern :pray:.

lib/ukboot/include/uk/boot.h
```diff
@@ -0,0 +1,44 @@
...
+#else /* !CONFIG_LIBUKBOOT_MAINTHREAD */
+#define uk_boot_shutdown_req(target) \
+	({ -ENOTSUP; })
+#endif /* !CONFIG_LIBUKBOOT_MAINTHREAD */
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* !__ASSEMBLY__ */
+#endif /* !__UK_BOOT_BOOT_H__ */
```
mogasergiu: Why `!__UK_BOOT_BOOT_H__`?

lib/uksched/include/uk/isr/sched.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef __UK_SCHED_SCHED_ISR_H__
+#define __UK_SCHED_SCHED_ISR_H__
+
+#include <uk/isr/thread.h>
+#include <uk/sched.h>
+
+/* ISR variant of `uk_sched_thread_woken()` */
+static inline void uk_sched_thread_woken_isr(struct uk_thread *t)
+{
+
```
mogasergiu: Redundant blank line?

lib/uksched/include/uk/isr/wait.h
```diff
@@ -0,0 +1,23 @@
...
+#include <uk/isr/thread.h>
+#include <uk/wait.h>
+
+/* ISR variant of `uk_waitq_wake_up()` */
+static inline void uk_waitq_wake_up_isr(struct uk_waitq *wq)
+{
+	struct uk_waitq_entry *curr, *tmp;
+	unsigned long flags;
+
+	ukplat_spin_lock_irqsave(&(wq->sl), flags);
```
mogasergiu: Are paranthesis required around `wq->sl`?

lib/uksched/include/uk/sched_impl.h
```diff
@@ -55,15 +55,17 @@ int uk_sched_register(struct uk_sched *s);
...
 	do { \
-		(s)->sched_start     = start_func; \
-		(s)->yield           = yield_func; \
-		(s)->thread_add      = thread_add_func; \
-		(s)->thread_remove   = thread_remove_func; \
-		(s)->thread_blocked  = thread_blocked_func; \
-		(s)->thread_woken    = thread_woken_func; \
-		(s)->idle_thread     = idle_thread_func; \
+		(s)->sched_start      = start_func; \
+		(s)->yield            = yield_func; \
```
mogasergiu, line 62: Could we align these backslashes?

lib/ukboot/include/uk/isr/boot.h
```diff
@@ -0,0 +1,45 @@
...
+#else /* !CONFIG_LIBUKBOOT_MAINTHREAD */
+#define uk_boot_shutdown_req_isr(target) \
+	({ -ENOTSUP; })
+#endif /* !CONFIG_LIBUKBOOT_MAINTHREAD */
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* !__ASSEMBLY__ */
+#endif /* !__UK_BOOT_BOOT_H__ */
```
mogasergiu: `__UK_BOOT_BOOT_ISR_H__`?

lib/ukboot/shutdown_req.c
mogasergiu, line 70: `checkpatch.pl --strict` throws
```
CHECK: Please use a blank line after function/struct/union/enum declarations
#62: FILE: lib/ukboot/shutdown_req.c:70:
+}
+UK_EVENT_HANDLER(UKPLAT_SHUTDOWN_EVENT, shutdown_req_handler);
```

mogasergiu (changes requested):
Ok had my last look. After this I should be able to give my tag. Everything still looks very good :raised_hands:, mostly just some (perhaps subjective?) styling comments.

include/uk/ctors.h
skuenzer: To make it compatible with any number of function arguments even though nothing is handed over.

lib/ukboot/include/uk/boot.h
skuenzer: Because of libukboot and boot.h:
`__UK_LIBNAME_HEADERNAME_H__`

lib/uksched/include/uk/isr/sched.h
skuenzer: Yes 👍🏻

lib/uksched/include/uk/sched_impl.h
skuenzer: This would require aligning all backslahses of this Macro defintion. I am never a fan of too intrusive alignment changes so I adopted to what was there already.

lib/ukboot/include/uk/isr/boot.h
skuenzer: uups, yes 👍🏻

lib/ukboot/shutdown_req.c
skuenzer: This is a registration call and not a (function prototype) declaration. I think checkpatch is too strict here, however, I can add a line in between.

lib/ukboot/boot.c
skuenzer: Ouh, this seems to come from a leftover before we updated `libuklibparam`. I can fix it with this commit. 👍🏻

include/uk/init.h
skuenzer: Sure 👍🏻

lib/ukboot/boot.c
skuenzer: Since when are we following christmas tree order of variable declarations, I usually group them into logical connection and then apply trees per group... I can update this one just for you 😄

lib/uksched/include/uk/isr/wait.h
skuenzer: Probably not, the precedence of C operators does not require it. Sometimes I am paranoid and want to be sure that the order is correct. I'll remove it. 👍🏻

lib/ukboot/include/uk/boot.h
mogasergiu: Yes I meant why the `!`?

skuenzer: :magic_wand: You did not see this 😄

mogasergiu (approved):

razvand (approved):

PR #1149, opened 2023-10-27
arch/arm64: Fix `ukarch_tlb_flush_entry`

kubanrob:
Fix a bug in `ukarch_tlb_flush_entry` that leads to the TLB entry not being invalidated on arm64.

The argument `X` of `tlbi vaae1is, X` is not the virtual address, but a field that contains the bits 55:12 of the virtual address in the bits 43:0. The remaining bits of this field (RES0,TTL) are valid being set to 0.

Store the virtual address at the proper location in the argument.

### Prerequisite checklist

### Base target

 - Architecture(s): arm64
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Map some memory at a virtual address, access it and unmap it. Due to the stale TLB entry, the memory can be still accessed.

### Reference

Description of the instruction: https://developer.arm.com/documentation/ddi0487/ja/?lang=en, C5.5.57, p. C5-1136

### DIFF
--- a/arch/arm/arm64/include/uk/asm/paging.h
+++ b/arch/arm/arm64/include/uk/asm/paging.h
@@ -285,12 +285,13 @@ static inline int ukarch_pt_write_base(__paddr_t pt_paddr)
 
 static inline void ukarch_tlb_flush_entry(__vaddr_t vaddr)
 {
+	__u64 page_number = vaddr >> PAGE_SHIFT;
 	__asm__ __volatile__(
 		"	dsb	ishst\n"        /* wait for write complete */
 		"	tlbi	vaae1is, %x0\n" /* invalidate by vaddr */
 		"	dsb	ish\n"          /* wait for invalidate compl */
 		"	isb\n"                  /* sync context */
-		:: "r" (vaddr) : "memory");
+		:: "r" (page_number) : "memory");
 }
 
 static inline void ukarch_tlb_flush(void)


arch/arm/arm64/include/uk/asm/paging.h
```diff
@@ -285,12 +285,13 @@ static inline int ukarch_pt_write_base(__paddr_t pt_paddr)
 
 static inline void ukarch_tlb_flush_entry(__vaddr_t vaddr)
 {
+	__u64 page_number = vaddr >> PAGE_SHIFT;
```
alexhoppus, line 288: I would prefer term vpn (virtual page number). page_number usually is something associated with the physical page

kubanrob: The names that I am used to are _pages for the virtual memory_ and _frames for physical memory_. Anyways, renaming it to `virtual_page_number` is also fine to me.

Maybe there is some canonical name for that number in Unikraft?

michpappas: I agree with @kubanrob, "virtual" seems redundant since we are referring to a page already. I'm okay to keep it as is. What caught my eye instead is shifting by `PAGE_SIZE`, which in theory is configurable between the standard 4KiB to 64KiB. However, the architecture takes care of that. From the Arm ARM:

> The treatment of the low-order bits of this field depends on the translation granule size, as follows:
> • Where a 4KB translation granule is being used, all bits are valid and used for the invalidation.
> • Where a 16KB translation granule is being used, bits [1:0] of this field are RES0 and ignored
when the instruction is executed, because VA[13:12] have no effect on the operation of the
instruction.
> • Where a 64KB translation granule is being used, bits [3:0] of this field are RES0 and ignored
when the instruction is executed, because VA[15:12] have no effect on the operation of the
instruction.

So I think it's all good 👍🏼

michpappas (approved):

razvand:
@alexhoppus , are you OK with the current state of the PR? Can we upstream it?

alexhoppus:
I'm okay, thanks.

skuenzer (approved):
An approval tag is still missing 😄 .

PR #1152, opened 2023-10-31
build: Import .config for noconfig targets

kubanrob:
Most of the noconfig_targets still need to import parts of the .config if it is available:

- `COMMON_CONFIG_ENV` imports values from the .config
- `savedefconfig` uses `UK_ARCH`

This fixes multiple problems with the noconfig targets for me, especially:

- savedefconfig saves wrong architecture when for x86_64 host arch and am64 guest arch
- olddefconfig does not use a known `UK_NAME` because it is not imported into `COMMON_CONFIG_ENV` 

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -380,11 +380,13 @@ EACHOLIB_LOCALS :=
 EACHOLIB_LOCALS-y :=
 
 # Pull in the user's configuration file
-ifeq ($(filter $(noconfig_targets),$(MAKECMDGOALS)),)
 ifneq ("$(wildcard $(UK_CONFIG))","")
+# Import the .config even if we do not expect it to be complete:
+# COMMON_CONFIG_ENV imports some values, savedefconfig uses UK_ARCH, ...
 $(call verbose_info,Including $(UK_CONFIG)...)
 -include $(UK_CONFIG)
-UK_HAVE_DOT_CONFIG := y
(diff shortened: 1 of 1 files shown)


StefanJum:
@kubanrob looks fine, but the `UK_HAVE_DOT_CONFIG` should be renamed in the [`syscall-shim/Makefile.rules`](https://github.com/unikraft-upb/unikraft/blob/master/lib/syscall_shim/Makefile.rules) file too.

kubanrob:
> @kubanrob looks fine, but the `UK_HAVE_DOT_CONFIG` should be renamed in the [`syscall-shim/Makefile.rules`](https://github.com/unikraft-upb/unikraft/blob/master/lib/syscall_shim/Makefile.rules) file too.

TBH I am not sure if this variable needs to be renamed in general, on the other hand `UK_HAVE_DOT_CONFIG` not being set is misleading if we are using values from `.config`. I would be happy to get feedback on that.

mariasfiraiala (commented):
Thanks @kubanrob, this makes sense and it works just fine. I think it's alright to have `UK_HAVE_DOT_CONFIG` renamed. @skuenzer, @StefanJum any opinions on this?

StefanJum:
I don't have any strong opinions on the renaming, maybe @skuenzer has some insights.

PR #1160, opened 2023-11-16
include: Add strong/weak reference counting

andreittr:
### Description of changes

This change adds a utility header implementing a reference counting scheme that supports regular (strong) as well as weak references.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

Checkpatch is ignorant of docstring formatting (or context); false-positive warnings.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A

PR #1126, opened 2023-10-11
{include/, lib/uksched, lib/uklock}: Add static initializers

andreittr:
### Description of changes

This change set adds static initializer macros for various internal Unikraft data structures:
- Refcount fields
- Thread wait queues
- Read-Write locks

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

The one checkpatch warning is a false error and the code is correct: casts must bind to their target without a space.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A

PR #1125, opened 2023-10-11
include/arch: Add atomic fetch & AND operation

andreittr:
### Description of changes

This change adds a Unikraft macro for an architecture's atomic fetch & AND operation, similar to the fetch & OR we already have.
It also fixes the order of the fetch & OR operations to be truthful to the docstring. All internal uses of fetch & OR discard the fetched result, so no behavior will change.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A

PR #1165, opened 2023-11-20
lib/ukfile: Introduce the ukfile library

andreittr:
### Description of changes

This change introduces the ukfile library to provide a new fundamental abstraction for "files" in Unikraft. This forms the base for a larger rearchitecting of vfscore.

This PR does not introduce any new syscalls or take over existing ones. All functionality is provided through Unikraft-internal APIs alone.

Depends on changes made in the following PRs:
- https://github.com/unikraft/unikraft/pull/1125
- https://github.com/unikraft/unikraft/pull/1126
- https://github.com/unikraft/unikraft/pull/1160

Merged branch for testing @ https://github.com/andreittr/unikraft/tree/ttr/feat-vfs/ukfile.

Part of the vfs rearch. Questions and comments regarding design, implementation, and documentation are more than welcome and will help make the new vfs layer more clear, intuitive, and performant.

### Prerequisite checklist

### Base target

 - Architecture(s): all
 - Platform(s): all
 - Application(s): N/A

### Additional configuration

Since no userspace interface is available to uk_file, you can test basic functionality with a dummy file implementation:
```c
#include <stdio.h>
#include <uk/file.h>
#include <uk/file/nops.h>
#include <uk/assert.h>

static uk_file_refcnt refcnt = UK_FILE_REFCNT_INITIALIZER;
static struct uk_file_state state = UK_FILE_STATE_INITIALIZER(state);

void myfile_release(const struct uk_file *f, int what)
{
	if (what & UK_FILE_RELEASE_RES)
		puts("Releasing resources...");
	if (what & UK_FILE_RELEASE_OBJ)
		puts("Releasing file object...");
}

static struct uk_file myfile = {
	.vol = NULL,
	.node = NULL,
	.ops = &uk_file_nops,
	.refcnt = &refcnt,
	.state = &state,
	._release = myfile_release
};

int main(void)
{
	int r;
	uk_pollevent ev;

	/* NOP syscalls return -ENOSYS */
	UK_ASSERT(uk_file_read(&myfile, NULL, 0, 0, 0) == -ENOSYS);

	/** EVENTS **/
	/* No events are set */
	r = uk_file_poll_immediate(&myfile, 0xff);
	UK_ASSERT(!r);

	/* Test timeout */
	puts("Tick");
	r = uk_file_poll_until(&myfile, 0xff, 1000000000);
	puts("Tock. (should've taken 1 sec)");
	UK_ASSERT(!r);

	/* Setting events, expecing to get them back immediately */
	ev = uk_file_event_set(&myfile, 0x5a);
	UK_ASSERT(!ev);
	r = uk_file_poll(&myfile, 0xff);
	UK_ASSERT(r == 0x5a);
	/* Test event filtering */
	r = uk_file_poll(&myfile, 0x45);
	UK_ASSERT(r == 0x40);
	/* Test event clear */
	ev = uk_file_event_clear(&myfile, 0xff);
	UK_ASSERT(ev == 0x5a);
	r = uk_file_poll_immediate(&myfile, 0xff);
	UK_ASSERT(!r);

	/** Refcounting **/
	uk_file_acquire(&myfile);
	uk_file_acquire_weak(&myfile);
	uk_file_release(&myfile);
	puts("Nothing should have been released yet");
	uk_file_release(&myfile);
	puts("Resources should have been released");
	uk_file_release_weak(&myfile);
	puts("File object should have been released");

	return 0;
}
```
Test code for multi-threaded event update/polling is left as an exercise for the reader. Future PRs that build on `uk_file` will, however, excercise polling code plenty.


lib/ukfile/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## What is a _file_?
+
+To overuse a classical *NIX idiom, "everything is a file".
+More concretely however, a file is an abstraction for any resource that offers a combination of input, output, and/or control operations.
+A file in Unikraft is a combination of an _immutable identity_ coupled with _mutable state_.
+
+Files are represented in Unikraft by the `struct uk_file` type, referenced in APIs as `const struct uk_file *` to enforce immutability.
+Identity consists of:
+- A volume identifier: driver-specific field, used to identify the file type as well as its originating volume
```
skuenzer: I am wondering if there is even a better name than `volume` for describing where a file comes from?

lib/ukfile/include/uk/file-nops.h
```diff
@@ -0,0 +1,33 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Convenience uk_file stub operations */
+
+#ifndef __UKFILE_FILE_NOPS_H__
+#define __UKFILE_FILE_NOPS_H__ 1
```
skuenzer: We usually have the following header guards layout (no defintion to `1` and closing `#endif` repeats the guard name with a comment):
```c
#ifndef __UKFILE_FILE_NOPS_H__
#define __UKFILE_FILE_NOPS_H__
```
...
```c
#endif /* __UKFILE_FILE_NOPS_H__ */
```

This should be adopted at multiple header files.

lib/ukfile/include/uk/statx.h
```diff
@@ -0,0 +1,80 @@
...
+	uint32_t stx_dev_major;
+	uint32_t stx_dev_minor;
+	uint64_t stx_mnt_id;
+	uint32_t stx_dio_mem_align;
+	uint32_t stx_dio_offset_align;
+	uint64_t _spare[12];
+};
+
+/* Bits used in stx_mask */
+#define STATX_TYPE        0x00000001U /* File type in stx_mode */
```
skuenzer: If these are Unikraft definitions, they should be all prefixed with `UK_` to avoid any potential name conflict (including indirect appearing name clashes). If they are Linux-common, the way we usually do it is by placing them at the same header path where they are also defined for Linux. If it is a `sys/*.h` header, you can place them under `include/sys` of this library. Otherwise, it is pretty likley that a libc provides these definitions, so you must place them in an according header in `nolibc`. At the same time, the code base has still to compile with musl (as feature-rich libc) that does provide its own header with these defintions.

lib/ukfile/include/uk/pollqueue.h
```diff
@@ -0,0 +1,584 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Multi-event poll/wait queue with update chaining support */
+
+#ifndef __UKFILE_POLLQUEUE_H__
+#define __UKFILE_POLLQUEUE_H__ 1
```
skuenzer: I wonder if we should place the pollqueue into an own library and make `lib/ukfile` depend on it: `/lib/ukpollqueue/`

lib/ukfile/include/uk/file.h
```diff
@@ -0,0 +1,261 @@
...
+
+/* I/O functions are non-blocking & return -EAGAIN when unable to perform.
+ * The behavior of concurrent calls to these functions is driver-dependent and
+ * no general assumptions can be made about their ordering and/or interleaving.
+ * Callers should themselves use the state->iolock (and/or other locks)
+ * as appropriate in order to provide the desired concurrency guarantees.
+ */
+
+/* I/O */
+typedef
```
skuenzer: So far we did not let `typedef` in an own line.

lib/ukfile/include/uk/pollqueue.h
```diff
@@ -0,0 +1,593 @@
...
+#include <uk/config.h>
+
+#include <uk/assert.h>
+#include <uk/arch/atomic.h>
+#include <uk/rwlock.h>
+#include <uk/plat/time.h>
+#include <uk/thread.h>
+
+/* Bitmask of event flags */
+typedef unsigned int uk_pollevent;
```
skuenzer: Do you really want to do a `typedef` for the flags?

lib/ukfile/include/uk/file.h
```diff
@@ -0,0 +1,263 @@
...
+	/* Voluntary locks (flock) */
+	/* TODO */
+};
+
+#define UK_FILE_STATE_INITIALIZER(name) ((struct uk_file_state){ \
+	.iolock = UK_RWLOCK_INITIALIZER((name).iolock, 0), \
+	.pollq = UK_POLLQ_INITIALIZER((name).pollq) \
+})
+
+typedef struct uk_swrefcount uk_file_refcnt;
```
skuenzer, line 119: Can you add a comment about this `typedef` as we discussed?

lib/ukfile/Config.uk
```diff
@@ -0,0 +1,11 @@
+config LIBUKFILE
+	bool "ukfile: Common support for files"
+	select LIBUKLOCK
+	select LIBUKLOCK_MUTEX
+	select LIBUKLOCK_RWLOCK
+	default n
```
skuenzer: `default n` is not needed to be specified, it is the default.

lib/ukfile/Config.uk
```diff
@@ -0,0 +1,11 @@
...
+
+# Hidden, selected by core components when required
+config LIBUKFILE_CHAINUPDATE
+	bool
+	default n
```
skuenzer: ...here, too.

lib/ukfile/include/uk/statx.h
```diff
@@ -0,0 +1,80 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UKFILE_STATX_H__
+#define __UKFILE_STATX_H__ 1
```
skuenzer: Can we move this header under `#include <uk/file/statx.h>`?

lib/ukfile/include/uk/file-nops.h
skuenzer: Can we move this header under `#include <uk/file/nops.h>`?

lib/ukfile/Config.uk
```diff
@@ -0,0 +1,11 @@
+config LIBUKFILE
+	bool "ukfile: Common support for files"
+	select LIBUKLOCK
+	select LIBUKLOCK_MUTEX
+	select LIBUKLOCK_RWLOCK
```
skuenzer, line 5: Because of pollqueue, we have a dependency to `LIBUKSCHED` as well.

lib/ukfile/include/uk/file.h
```diff
@@ -0,0 +1,263 @@
...
+
+/* I/O */
+typedef
+ssize_t (*uk_file_io_fn)(const struct uk_file *f,
+			 const struct iovec *iov, int iovcnt,
+			 off_t off, long flags);
+
+/* Info (stat-like & chXXX-like) */
+typedef
+int (*uk_file_getstat_fn)(const struct uk_file *f,
```
skuenzer: According to style guide, we should use `_func` for the function types: `uk_file_getstat_func`.

lib/ukfile/include/uk/pollqueue.h
skuenzer: Alternatively, I think, would be to move the header into a subfolder: `#include <uk/file/pollqueue.h>`.With plat re-arch we started to leave the mode where our libs flood headers directly under `uk/`. The mode is more like: `uk/mainheader.h` and `uk/libname/other_headers.h`

lib/ukfile/include/uk/file.h
```diff
@@ -0,0 +1,263 @@
...
+			  unsigned int mask, struct uk_statx *arg);
+typedef
+int (*uk_file_setstat_fn)(const struct uk_file *f,
+			  unsigned int mask, const struct uk_statx *arg);
+
+/* Control */
```
skuenzer, line 45: Is this parameter `fam` (family)?

lib/ukfile/README.md
andreittr: I've given some thought to this name, and "volume", even though it's a bit FS-centric, is the best one I could come up with. Here are my runners-up:
- `driver`: misleading name because it should identify a driver _instance_, not the driver itself (i.e., same driver may have multiple, independent, orthogonal instances)
- `driver_instance`: too long and unwieldy, implies some actual structure to the field (it has none, driver is free to put anything there)
- `tag`: carries little meaning beyond "here's a field in this struct"
- `type`: risks being confused with the file type bits (subset of posix open file mode bits)
- `magic`: as uninformative as `tag` but more playful

Feel free to suggest others, but I don't know any better ones.

lib/ukfile/include/uk/file-nops.h
andreittr: ~~Is it common/ok to subdivide the `uk/` header pathspace by lib? I don't mind doing so, but I hadn't seen it done much in the core codebase and assumed a flat hierarchy is preferable.~~

Nvm, saw the explanation below that it's a new policy for rearches. Will subdivide.

lib/ukfile/include/uk/statx.h
andreittr: To answer the first question, these are unikraft definitions that should equal Linux kernel definitions, however these are not available in libc's.

statx is a syscall interface and is not usually exposed to userspace directly. There is no `<sys/statx.h>` and it's not usually present in other `<sys/*.h>` headers either:
- glibc under `_GNU_SOURCE` will pull statx defintions into `<sys/stat.h>` by including a header from `bits/` (which are copied from kernel headers at configure-time)
- musl does not include any headers and defines its own `struct statx` in `fstatat.c`; it defines no mask bits and uses the magic value `0x7ff` in syscalls

Given these circumstances, what's the best call?

lib/ukfile/include/uk/pollqueue.h
andreittr: `#include <uk/file/pollqueue.h>` makes most sense IMO.
Moving pollqueue into its own lib poses some subtle Kconfig issues, but more importantly would only be valuable if it can be used independently from files. At the moment I don't see such an application.
Moving under `<uk/file/...>` makes sense for direct consumers outside of `ukfile` (mostly epoll) because it expresses intent more clearly: "Epoll wants to directly interface with the API of a file's pollqueue", and not just "of any pollqueue".

lib/ukfile/include/uk/file.h
andreittr: Fair enough, I didn't know (and checkpatch didn't complain). I can fix if you think it looks ugly.

lib/ukfile/include/uk/pollqueue.h
andreittr: It will come in use if we ever want to have unikraft/pollqueue-internal flags that are guaranteed to never overlap with posix event flags (I actually had some of these in early implementations but managed to engineer them out).
In such a case we'd change it to `unsigned long`, and any necessary type conversion between unikraft event type and posix event type remains isolated in `posix-fdio`.

lib/ukfile/include/uk/file.h
andreittr: Yes, I'll add a comment making it clear.

lib/ukfile/include/uk/statx.h
skuenzer: This sounds to me then that we should prefix the definitions with `UK_` to avoid any clashes but then we can keep them in this file. Since you wrote you got them from Linux: Under which license are these defintions? I hope non-GPL or covered by the syscall note. Is it possible that  because of licensing, the bits are not defined when using musl (if by chance you know)?

lib/ukfile/include/uk/pollqueue.h
skuenzer: Can you add a comment to the typedef with the justification?

lib/ukfile/include/uk/statx.h
andreittr: I don't know whether musl doesn't include them because of licensing issues, or because they're Linux extensions (i.e., not posix) and musl can't (yet) be bothered.
The header file itself (`include/uapi/linux/stat.h` in the kernel source tree) is under GPL2 with the syscall note:
```
/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
```

lib/ukfile/README.md
skuenzer: 🤔  Maybe a compromise is to add `driver instance` in brackets behind the first occurrence of volume. What do you think? But, I agree the other suggestion fit less and I also do not have a better one atm.

lib/ukfile/include/uk/statx.h
skuenzer: Can you also write down the file path where you note that the defines come from Linux. Maybe in brackets the license with syscall note.

skuenzer (approved):
Thanks a lot for your work. I will take it in.

PR #1168, opened 2023-11-22
lib/*: Introduce posix-fdtab

andreittr:
### Description of changes

This changeset introduces `posix-fdtab`, a library implementing a file descriptor table, moving this functionality out of `vfscore`.
At its core, posix-fdtab is designed to work with the new-style files introduced with the vfs rearch. Support for legacy vfscore files is provided as a stop-gap measure, allowing the use of the same file descriptor table to reference both types of files in one binary.

posix-fdtab provides several APIs:
- `uk/posix-fdtab.h`: general Unikraft-internal API
- `uk/posix-fdtab-legacy.h`: vfscore-specific API, allowing providers of vfscore files to delegate fd handling
- file descriptor specific syscalls

This PR moves the following syscalls out of `vfscore`:
- `close`
- `dup`, `dup2`, `dup3`

Depends on the following PR
- https://github.com/unikraft/unikraft/pull/1165
as well as all of its dependencies.

Merged branch for testing @ https://github.com/andreittr/unikraft/tree/ttr/feat-vfs/fdtab.

Part of the vfs rearch. Questions and comments regarding design, implementation, and documentation are more than welcome and will help make the new vfs layer more clear, intuitive, and performant.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

All applications that use file descriptors (filesystem, pipes, sockets) should continue working as normal.


nderjung:
Hi @andreittr, thanks a lot for this. This has been a long-time in the works and an awesome contribution. Please just check out the failing [`intergration/libc-test`](https://github.com/unikraft/unikraft/actions/runs/6951220080/job/18912816632?pr=1168#step:4:1317):
```
 i    CC      libposix_fdtab: fdtab.o
 W  In file included from /github/workspace/lib/posix-fdtab/fdtab.c:16:
 W  /github/workspace/lib/posix-fdtab/include/uk/posix-fdtab.h:13:10: fatal error: uk/ofile.h: No such file or directory
 W     13 | #include <uk/ofile.h>
 i    AS      libposix_fdtab: libinfo.libuklibid.o
 W        |          ^~~~~~~~~~~~
 W  compilation terminated.
 ```

File: lib/posix-fdtab/fdtab.c [hunk 1]
```diff
@@ -0,0 +1,493 @@
...
+static char init_bmap[UK_BMAP_SZ(UK_FDTAB_SIZE)];
+static void *init_fdmap[UK_FDTAB_SIZE];
+
+struct uk_fdtab {
+	struct uk_alloc *alloc;
+	struct uk_fmap fmap;
+};
+
+static struct uk_fdtab init_fdtab = {
+	.fmap = {
```
skuenzer, line 38: I am a bit confused by the identation. Shouldn't it be?:
```c
static struct uk_fdtab init_fdtab = {
	.fmap = {
		.bmap = {
			.size = UK_FDTAB_SIZE,
			.bitmap = (unsigned long *)init_bmap
		},
		.map = init_fdmap
	}
};
```

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -0,0 +1,93 @@
...
+/* POSIX compatible file descriptor table */
+
+#ifndef __UK_POSIX_FDTAB_H__
+#define __UK_POSIX_FDTAB_H__ 1
+
+#include <uk/config.h>
+#include <uk/ofile.h>
+
+/* Internal fd type; should safely cast to what userspace expects an fd to be */
+typedef int uk_fd;
```
skuenzer: hum... since we are doing a `posix-fdtab`, we can go directly with `int`, right? I am thinking this way because `int` is the type for a POSIX fd anyways.

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Close all files that have the O_CLOEXEC flag set.
+ *
+ * Assumes no other threads are touching the active fdtab during the call.
+ */
+void uk_fdtab_cloexec(void);
+
+#if CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM
+union uk_shim_file {
```
skuenzer, line 72: I assume the shim file will disappear at some point? If so, can you put a `FIXME`, `TODO`, or similar comment explaingin this before we introduce the declarations here?

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+#include <uk/arch/atomic.h>
+#include <uk/assert.h>
+#include <uk/bitops.h>
+#include <uk/bitmap.h>
+#include <uk/thread.h>
+
+/**
+ * Lock-free bitmap, with ones representing a free index.
+ */
+struct uk_bmap {
```
skuenzer, line 24: As discussed offline, we will move the lock-free bitmap operations to `include/uk/bitmap.h` and make use of them here in a PR later. This is to have the updated VFS earlier available.

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+ */
+struct uk_fmap {
+	struct uk_bmap bmap;
+	void *volatile *map;
+};
+
+#define UK_FMAP_SZ(s) ((s) * sizeof(void *))
+
+
+#define UK_FMAP_INRANGE(m, i) ((i) < (m)->bmap.size)
```
skuenzer: Just for the case, we have some RANGE helper macros also in `uk/essentials.h`. Your macro assume `i` is a non-negative number.

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+	uk_bitmap_fill((void *)bm->bitmap, bm->size);
+}
+
+/**
+ * Mark `idx` as used, and return whether we were the ones to do so.
+ *
+ * @return
+ * 	0 if we marked `idx` as used, non-zero otherwise
+ */
+static inline int uk_bmap_reserve(const struct uk_bmap *bm, size_t idx)
```
skuenzer: `size_t` for an index which is at the end a posix file descriptor number? Could we just use `int` or at least `unsigned int`?

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+static inline void uk_fmap_init(const struct uk_fmap *m)
+{
+	memset((void *)m->map, 0, m->bmap.size * sizeof(void *));
+	uk_bmap_init(&m->bmap);
+}
+
+/**
+ * Lookup the entry at `idx`.
+ *
+ * WARNING: Use of this function is vulerable to use-after-free race conditions
```
skuenzer: `vulerable` -> `vulnerable`

lib/posix-fdtab/include/uk/posix-fdtab-legacy.h
```diff
@@ -0,0 +1,18 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Legacy file interface into posix-fdtab for use by vfscore */
+
+#ifndef __UK_POSIX_FDTAB_LEGACY_H__
```
skuenzer, line 9: Please also update the header guards in this PR.

lib/posix-fdtab/fdtab.c
skuenzer: Btw, I like that you have the init(ial) table statically initialized as much as possible 👍🏻 .

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,493 @@
...
+}
+static inline void file_rel(struct uk_fdtab *tab, void *p, int flags)
+{
+	if (flags & UK_FDTAB_VFSCORE)
+		fdrop((struct vfscore_file *)p);
+	else
+		ofile_rel(tab, (struct uk_ofile *)p);
+}
+
+#else
```
skuenzer:
```c
#else /* !CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
```

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,493 @@
...
+
+#define file_acq(p, f) ofile_acq((struct uk_ofile *)(p))
+#define file_rel(t, p, f) ofile_rel((t), (struct uk_ofile *)(p))
+
+#endif
```
skuenzer:
```c
#endif /* !CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
```

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+ * for entry objects whose lifetime depends on their membership in this data
```
skuenzer, line 133: So far we indented with the message tag:
```c
/*
 * WARNING: Use of this function is vulnerable to use-after-free race conditions
 *          for entry objects whose lifetime depends on their membership in [...]
 */
```
However, this is only minor and I am fine if you keep the indentation as is.

lib/posix-fdtab/fmap.h
```diff
@@ -0,0 +1,302 @@
...
+ * structure (e.g., refcounting on put, take, and after lookups).
+ * For that case please use the `uk_fmap_critical_*` functions.
+ *
+ * @return
+ * 	The entry at `idx`, or NULL if out of range.
+ */
+static inline void *uk_fmap_lookup(const struct uk_fmap *m, size_t idx)
```
skuenzer: Can't we handle the entries with an actual type instead of `void *`?

lib/posix-fdtab/Config.uk
```diff
@@ -0,0 +1,16 @@
...
+
+if LIBPOSIX_FDTAB
+	config LIBPOSIX_FDTAB_MAXFDS
+	int "Maximum number of file descriptors"
+	default 1024
+
+	# Hidden, selected by core components when needed
+	config LIBPOSIX_FDTAB_LEGACY_SHIM
+	bool
+	default n
```
skuenzer, line 13: `default n` is not required. `default n` is the default behavior.

lib/posix-fdtab/fdtab.c
andreittr: You're right, I've no clue how this slipped through.

andreittr: Runtime inits and dynamic memory allocation are the devil :upside_down_face: .

lib/posix-fdtab/include/uk/posix-fdtab.h
andreittr: Something like
```c
/*
 * This shim interface exists to support cohabitation with vfscore until its eventual removal.
 * Unless you are implementing a vfscore shim interface, please do not use this API.
 */
```
?

lib/posix-fdtab/fmap.h
andreittr: They indeed do (because size_t is unsigned), and since fmap is internal to fdtab, we control all callsites and can ensure this invariant (which is done in `_fdtab_get`), preventing redundant checks. That said, I did find places where fd validation was missing, so maybe moving all validation into fmap is the better call.

Also, I should probably rename these macros to internal ones; names are left-over from an earlier design where I had intended fmap to have a more generic and perhaps public API.

lib/posix-fdtab/fmap.h
andreittr: Leftover from a more generic design; can convert into int. Although in this particular case is makes no difference since the idx is not stored anywhere in memory.

lib/posix-fdtab/fmap.h
andreittr: whoops

lib/posix-fdtab/fmap.h
andreittr: I tried, but getting C to let me
- store a union type into a pointer-sized thing with atomic ops
- set/check bit flags present in the least-significant bits guaranteed to be unused due to the alignment requirements of both `struct uk_file` and `struct vfscore_file`
- get a union type back out of that mess

without resorting to UB was beyond my C-fu.
The only two options I could think of were `void *` or `uintptr_t`.

Mind you, once the vfscore shim stuff is out, we will be able to have actual types in there (I'll find a way to store flags in typed pointers without UB _somehow_).

lib/posix-fdtab/include/uk/posix-fdtab.h
andreittr: That's fair; maybe there is a conceptual split to be made between `ukfdtab` - what implements most internal API & ofile management and whatnot, and `posix-fdtab` which provides syscalls, which would warrant this typedef. But then again, this split is so tenuous that I don't think it's worth bringing about right now.
Let's hope we won't ever have to either, because unlike this fix it's not a simple search&replace.

lib/posix-fdtab/include/uk/posix-fdtab.h
skuenzer: Hum, I was actually thinking of an additionally one liner comment, but you are right, this comment is good enough. I think we can just add here a `TODO` tag here. 👍🏻

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -0,0 +1,96 @@
...
+	struct uk_ofile *ofile;
+	struct vfscore_file *vfile;
+};
+
+/**
+ * Get the open file description associated with descriptor `fd`, shimming
+ * between vfscore_file and uk_ofile, writing the value in `*out`.
+ *
+ * @return
+ *   0, if successful and `*out` is a uk_ofile
```
skuenzer: Now seeing code using this interface, maybe we want to provide `#define` for `0` and `1` so that code gets even quicker readable, something along these lines:
```c
#define UK_FSHIM_OFILE 0
#define UK_FSHIM_VFSCORE 1
```
What do you think?

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,481 @@
...
+
+static int init_posix_fdtab(void)
+{
+	init_fdtab.alloc = uk_alloc_get_default();
+	/* Consider skipping init for .map (static vars are inited to 0) */
+	uk_fmap_init(&init_fdtab.fmap);
+	return 0;
+}
+
+uk_early_initcall_prio(init_posix_fdtab, UK_PRIO_EARLIEST);
```
skuenzer, line 56: Can we add short one-liner comment, something like
```
/* Initialize fdtab as early as possible so that libc functionality is becoming available that rely on fds */
```

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,481 @@
...
+
+/* STUB: Adapt when multiple processes are supported */
```
skuenzer: We typically use `TODO:` or `FIXME:` instead.
https://unikraft.org/docs/contributing/coding-conventions#comment-keywords

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,481 @@
...
+#endif
+
+#define _FLAG_MASK (((uintptr_t)_MAX_FLAG << 1) - 1)
+
+UK_CTASSERT(__alignof__(struct uk_ofile) > _MAX_FLAG);
+#if CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM
+UK_CTASSERT(__alignof__(struct vfscore_file) > _MAX_FLAG);
+#endif
+
+static inline const void *fdtab_encode(const void *f, int flags)
```
skuenzer, line 87: I wonder if the code is slightly more easy to understand if we rename the two functions to something like:
```c
fdtab_shim_pack()
fdtab_shim_unpack()
```
Not sure how you feel about this, we can also keep the naming.

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,481 @@
...
+	void *p;
+	int flags;
+};
+
+#define UK_FDTAB_CLOEXEC 1
+
+#if CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM
+#define UK_FDTAB_VFSCORE 2
+#define _MAX_FLAG 2
+#else
```
skuenzer:
```c
#else /* !CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
```

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,481 @@
...
+#define _MAX_FLAG 1
+#endif
```
skuenzer:
```c
#endif /* !CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
```

lib/posix-fdtab/fdtab.c
skuenzer: I also wonder if we should define a `union` instead of using `void *` to get at least some type checking. Would something along these lines make sense?:
```c
union fdtab_shim_ofile {
    struct vfscore_file *vfsfile;
    struct uk_ofile *ofile;
};
```
Btw, I don't know how well such a flag encoding works with pointer authentication in place. I can imagine we need to replace the implementation. Because we want to integrate this updates very soon, I think we should do this later in a new PR.

skuenzer: Am I right that this encoding/packing will not be needed anymore when vfscore is completely replaces?

lib/posix-fdtab/include/uk/posix-fdtab.h
skuenzer: Makes sense from where the idea is coming from. I am naively assume atm that we will not need to replace it, just because of the posix interface. My stronger argument is that `int` simplifies the code base because we mentally hop less with data types that are effectively the same.

lib/posix-fdtab/fmap.h
skuenzer: Okay, fair enough if they become somewhat internal.

lib/posix-fdtab/fdtab.c
andreittr: Packing _might_ be needed for O_CLOEXEC, unless I can find a clever way to atomically store the flag along with the ofile without using locks. But if pressed by smth like ptr auth we can always give up on packing and adopt locks (perhaps selectable via config).

lib/posix-fdtab/include/uk/posix-fdtab.h
andreittr: Useful defines, even for code meant to be thrown out eventually; will add.

lib/posix-fdtab/fmap.h
skuenzer: hum... I understand. Since the function is internal anyways, it is less critical, we can leave the `void *` as is and I am looking forward when vfscore is out and we can give it a proper type. Maybe you could put a very short comment regarding `void *` in the code so that we don't loose this piece of valuable information (especially that we will be able to give a proper type later)?

lib/posix-fdtab/fdtab.c
skuenzer: I just noticed that these comments are related to https://github.com/unikraft/unikraft/pull/1168#discussion_r1408521095 . Same conclusion applies also here.

lib/posix-fdtab/fmap.h
andreittr: I think the interpretation of fmap's `void *` return is (sensibly) left to the encoding functions in `fdtab.c`, and bringing in any of the shim shenanigans into fmap would just pollute the code base.
In a future version where fmap will always store the same typed pointers we can propagate this type into here much cleaner.

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,482 @@
...
+#if CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM
+int uk_fdtab_legacy_open(struct vfscore_file *vf)
+{
+	struct uk_fdtab *tab = _active_tab();
+	const void *entry;
+	int fd;
+
+	fhold(vf);
+	entry = fdtab_encode(vf, UK_FDTAB_VFSCORE);
+	fd = _fmap_put(&tab->fmap, entry, 0);
```
skuenzer, line 241: I think this was forgotten on the last update to `uk_fmap_put`. I'll do this for you while upstreaming 👍🏻

lib/posix-fdtab/fdtab.c
```diff
@@ -0,0 +1,482 @@
...
+		if (v.flags & UK_FDTAB_VFSCORE) {
+			vf = (struct vfscore_file *)v.p;
+			fhold(vf);
+		}
+		uk_fmap_critical_put(fmap, fd, p);
+	}
+	return vf;
+}
+
+int uk_fdtab_shim_get(uk_fd fd, union uk_shim_file *out)
```
skuenzer, line 270: I'll fix this while upstreaming: `uk_fd` -> `int`

lib/posix-fdtab/include/uk/posix-fdtab.h
skuenzer: I'll add the tag while upstreaming.

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Get the open file description associated with descriptor `fd`, shimming
+ * between vfscore_file and uk_ofile, writing the value in `*out`.
+ *
+ * @return
+ *   UK_SHIM_OFILE, if successful and `*out` is a uk_ofile
+ *   UK_SHIM_LEGACY, if successful and `*out` is a vfscore_file
+ *   < 0, if `fd` is not mapped
+ */
+int uk_fdtab_shim_get(uk_fd fd, union uk_shim_file *out);
```
skuenzer, line 86: I am fixing while upstreaming: `uk_fd` -> `int`

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -0,0 +1,100 @@
...
+
+#define UK_SHIM_OFILE  0
+#define UK_SHIM_LEGACY 1
+
+#endif /* CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
+
+/* Internal syscalls */
+int uk_sys_close(int fd);
+int uk_sys_dup(int oldfd);
+int uk_sys_dup_min(int oldfd, size_t min, int flags);
```
skuenzer, line 96: Also here: `size_t` -> `int`

skuenzer (approved):

PR #1169, opened 2023-11-22
build: Introduce `UK_IMAGE_NAME_OVERWRITE` make var

nderjung:
### Prerequisite checklist

### Base target

 - Platform(s): all

### Description of changes

This commit introduces a new make variable, `UK_IMAGE_NAME_OVERWRITE`, which is used to customize the output binary name of the kernel image.  This is a special, and experimental, variable which can be helpful in certain use cases, for example where targets with the same architecture and platform are present.  This variable should be used in conjunction with a different build directory, `O=`, as the resulting binary will always have the name from this option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,6 +396,20 @@ else
 CONFIG_UK_NAME ?= $(notdir $(APP_DIR))
 endif
 
+# Option to overwrite file name of generated images by supported platforms. The
+# target platform has to support this feature.  When unset or unsupported by the
+# platform, the default semantic of the target platform (typically
+# `$(UK_NAME)_$(PLAT)_$(ARCH)`) will be used.
+#
+# NOTE: Please note that multi-platform builds do not work anymore when this
+#       option is used.  The reason is that the build system will generate the
+#       same file by each selected platform.  Make won't be able to proceed
+#       building.
+# NOTE: This feature is currently used by kraftkit.
+ifneq ($(call qstrip,$(UK_IMAGE_NAME_OVERWRITE)),)
+UK_IMAGE_NAME_OVERWRITE := $(call qstrip,$(UK_IMAGE_NAME_OVERWRITE))
+endif
+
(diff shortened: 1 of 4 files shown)


Config.uk
```diff
@@ -236,3 +236,18 @@ config UK_NAME
 	default UK_DEFNAME
 	help
 	  Name to be used for final image
+
+# Hidden config option due to caveat of overwriting existing
+# binaries in multi-target builds
+config UK_BIN_NAME
```
skuenzer: Can we use `UK_IMAGE_NAME_OVERWRITE`?

Config.uk
```diff
@@ -236,3 +236,18 @@ config UK_NAME
...
+	string ""
+	help
+		Set the final output binary name for the kernel.
```
skuenzer:
```diff
-Set the final output binary name for the kernel.
+Provides a way to overwrite the file name for the final image that is generated by a platform.
+The target platform has to support this feature.
```

Config.uk
```diff
@@ -236,3 +236,18 @@ config UK_NAME
...
+		When unset, the default semantic of
+		`$(UK_NAME)_$(VMM)_$(ARCH)` will be used.
```
skuenzer:
```diff
When unset or unsupported by the platform, the default semantic of the target platform (typically `$(UK_NAME)_$(PLAT)_$(ARCH)`) will be used.
```

Config.uk
```diff
@@ -236,3 +236,18 @@ config UK_NAME
...
+		Note that when building for different platform/
```
skuenzer: Instead of this description:
Please note that multi-platform builds do not work anymore when this option is used. The reason is that the build system will generate the same file by each selected platform. Make won't be able to proceed building.

skuenzer (approved):
Thanks for the quick updates. The PR looks fine.

PR #1173, opened 2023-11-24
lib/syscall_shim: Fix `arch_prctl`'s `ARCH_SET_FS` handling

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

NOTE: This is interdependent on the `app-elfloader` commit that changes `arch_prctl`'s `ARCH_SET_FS` behavior, [PR #63](https://github.com/unikraft/app-elfloader/pull/63)

The code may store and restore userland TLS pointer properly in the system call handler and it does check if it was changes by syscalls such as `arch_prctl`'s `ARCH_SET_FS` command. However, one aspect was missed: `__UK_SYSCALL_RETADDR_CLEAR`. The macro `__UK_SYSCALL_RETADDR_CLEAR` accesses what is meant to be Unikraft's TLS and is invoked in the time window between TLS store and restore. In the `ARCH_SET_FS` case, `arch_prctl` sets the the `fs` register regardless and thus we end up having the userland desired TLS pointer during a time window where keeping our own TLS pointer is mandatory.

Thus, to avoid this, do not set any `fs` register value during `arch_prctl` (see top NOTE) and, instead, simply modify the `_uk_syscall_ultlsp` value which will be used instead to restore the userland TLS as the last operation.

### DIFF
--- a/lib/syscall_shim/uk_syscall_binary.c
+++ b/lib/syscall_shim/uk_syscall_binary.c
@@ -51,7 +51,7 @@ void ukplat_syscall_handler(struct __regs *r)
 {
 #if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
 	struct uk_thread *self;
-	__uptr orig_tlsp;
+	__uptr ultlsp;
 #endif /* CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS */
 	/* Place backup of extended register state on stack */
 	__sz ectx_align = ukarch_ectx_align();
@@ -75,11 +75,11 @@ void ukplat_syscall_handler(struct __regs *r)
 
 #if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
 	/* Activate Unikraft TLS */
-	orig_tlsp = ukplat_tlsp_get();
+	ultlsp = ukplat_tlsp_get();
 	self = uk_thread_current();
 	UK_ASSERT(self);
 	ukplat_tlsp_set(self->uktlsp);
-	_uk_syscall_ultlsp = orig_tlsp;
+	_uk_syscall_ultlsp = ultlsp;
 #endif /* CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS */
 
 	/* uk_syscall6_r() will clear _uk_syscall_return_addr on return */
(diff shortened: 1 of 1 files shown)


lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -112,17 +112,20 @@ void ukplat_syscall_handler(struct __regs *r)
...
-	} else {
+#if CONFIG_LIBUKDEBUG_PRINTD
+	if (_uk_syscall_ultlsp != ultlsp) {
 		uk_pr_debug("System call updated userland TLS pointer register to %p (before: %p)\n",
-			    (void *) orig_tlsp, (void *) ukplat_tlsp_get());
+			    (void *)ultlsp, (void *)ukplat_tlsp_get());
 	}
+#endif
+
+	ultlsp = _uk_syscall_ultlsp;
```
michpappas: And as `ultlsp` is not used beyond this point, this would be more readable as:
```diff
-	ultlsp = _uk_syscall_ultlsp;
-	uk_thread_uktls_var(self, _uk_syscall_ultlsp) = 0x0;
-
-	ukplat_tlsp_set(ultlsp);
+	ukplat_tlsp_set(_uk_syscall_ultlsp);
+ 	uk_thread_uktls_var(self, _uk_syscall_ultlsp) = 0x0;
```

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -75,11 +75,11 @@ void ukplat_syscall_handler(struct __regs *r)
 
 #if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
 	/* Activate Unikraft TLS */
-	orig_tlsp = ukplat_tlsp_get();
+	ultlsp = ukplat_tlsp_get();
 	self = uk_thread_current();
 	UK_ASSERT(self);
 	ukplat_tlsp_set(self->uktlsp);
-	_uk_syscall_ultlsp = orig_tlsp;
+	_uk_syscall_ultlsp = ultlsp;
```
michpappas, line 82: With the current PR the purpose of `ultlsp` is reduced to printing that message below. Personally I wouldn't mind getting rid of that message in favor of less redundancy.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -112,17 +112,20 @@ void ukplat_syscall_handler(struct __regs *r)
...
-	if (likely(ukplat_tlsp_get() == self->uktlsp)) {
-		ukplat_tlsp_set(orig_tlsp);
-	} else {
+#if CONFIG_LIBUKDEBUG_PRINTD
+	if (_uk_syscall_ultlsp != ultlsp) {
 		uk_pr_debug("System call updated userland TLS pointer register to %p (before: %p)\n",
-			    (void *) orig_tlsp, (void *) ukplat_tlsp_get());
+			    (void *)ultlsp, (void *)ukplat_tlsp_get());
 	}
+#endif
```
michpappas: If we keep that message nevertheless, maybe add `/* !CONFIG_LIBUKDEBUG_PRINTD */` here.

michpappas (changes requested):
Dear @mogasergiu, besides the inline comments:

1. Please consider https://github.com/unikraft/app-elfloader/pull/63#discussion_r1406478235
2. Some nits on the commit message:
```
NOTE: This is interdependent on the `app-elfloader` commit that
```
s/interdependent on/interdependent with/

```
changes `arch_prctl`'s `ARCH_SET_FS` behavior, entitled:
```
s/entitled/titled/ (entitled means something different)

```
Do not change real `fs_base` register in `arch_prctl`

The code may store and restore userland TLS pointer properly in the
system call handler and it does check if it was changes by syscalls
```
s/if it was changes/if it was changed/ or "if there were changes"

```
such as `arch_prctl`'s `ARCH_SET_FS` command. However, one aspect
```
s/command/macro/

Also perhaps for improved context:

such as the `ARCH_SET_FS` macro defined in `app-elfloader/arch_prctl.c` (although you also specify it in the `NOTE` above).

Saying that, the problem is more generic, ie it applies to the syscall implementation updating the userland TLS regardless of the means (eg `ARCH_SET_FS` or otherwise), so perhaps the commit message could be simplified.

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: It's worth keeping, that message saved me a lot of time, I owe that to it :pleading_face:

michpappas: Ok, fair enough

michpappas (approved):
Looks good, thanks!

Issue #16, opened 2019-10-15
Fix xen ukplat_monotonic_clock


felipehuici:
n Xen, ukplat_monotonic_clock returns the time elapsed (in nanoseconds) since the xen hypervisor booted, not since ukplat_time_init (as the function docs for ukplat_monotonic_clock say).

PR #1174, opened 2023-11-24
Implement stack switching on system call entries and on ARM traps/IRQ's

mogasergiu:
### Prerequisite checklist

### Base target

 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

- assign the current LCPU's `struct lcpu *` to a system register (`gs_base`/`k_gs_base` on `x86_64`, `tpidr_el1` on `ARM64`)
- use `sp_el0` on `ARM64` as the register to hold the middle address of a double-sized buffer meant to represent the two stacks to switch to in case of an IRQ/trap:
```
          STACK_SIZE               STACK_SIZE
     <---------------------><--------------------->
     |============================================|
     |                     |                      |
     |       trap stack    |        IRQ stack     |
     |                     |                      |
     |=============================================
                           ^
                         SP_EL0
```
- define a per-thread and per-LCPU field `auxsp` which represents the auxiliary stack that both architectures can use to switch to in case of a system call
- the stack switch on `x86_64` is achieved with the help of `swapgs` and `%gs:<AUXSP_OFFSET>`, while on `ARM64` it is achieved by using `tpidrro_el0` as a scratch register. The latter is assumed to be fine to do so because it will always hold a value the application can't modify and we will always be able to restore it to its desired known value anytime we want

Depends on #1173

NOTE: This should not be tested on its own. Use #1175 instead as it contains everything.


plat/common/include/x86/cpu_defs.h
```diff
@@ -91,6 +91,8 @@
 /* FS and GS Registers in 64-bit Mode */
 #define X86_MSR_FS_BASE         0xc0000100
 #define X86_MSR_GS_BASE         0xc0000101
+/* Used in conjunction with swapgs instruction */
+#define X86_MSR_KERNEL_GS_BASE	0xc0000102
```
michpappas, line 95: Maybe this and the next commit can be squashed, no need for one commit for every new line added 😅

plat/common/include/x86/gsbase.h
```diff
@@ -0,0 +1,117 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <x86/cpu.h>
+#include <uk/essentials.h>
+
+/* TODO: Use wrgsbase/rdgsbase instead, as they are faster */
```
michpappas, line 13: `wrgsbase` / `rdgsbase` are implemented right below, so is this TODO left on purpose or is it forgotten?

plat/common/x86/lcpu.c
```diff
@@ -76,6 +77,9 @@ int lcpu_arch_init(struct lcpu *this_lcpu)
 
 	traps_lcpu_init(this_lcpu);
 
+	wrkgsbase((__uptr)this_lcpu);
```
michpappas, line 80: Please confirm my understanding, as the commit message specifies that they both need to be initialized but does not specify the reason. Are you also initializing KGS so that we get unikraft's GS restored upon syscall entry (`swapgs`)?

include/uk/plat/memory.h
```diff
@@ -252,6 +257,61 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags);
...
+ *   accesses to the auxiliary stack do not generate page faults in more
+ *   fragile system states.
+ * @param auxsp_len
+ *   The custom length of the auxiliary stack. If 0, then
+ *   CONFIG_UKPLAT_AUXSP_SIZE is used instead as the default length.
+ *
+ * @return
+ *   Pointer to the allocated auxiliary stack
+ */
+static inline __uptr ukplat_auxsp_alloc(struct uk_alloc *a,
```
michpappas: So for the main stack and tls in `uk_thread` we simply call `memalign()`. Is there a reason that `auxsp` requires its own stack allocation function? If there is enough justification for `uk_vma_advise`, then perhaps the `stack` and `tls` should use that too?

CC: @skuenzer

plat/common/x86/syscall.S
```diff
@@ -33,12 +33,65 @@
...
+	 * `auxsp` field.
+	 * We thus achieve a complete switch to another stack while preserving
+	 * the context of the application.
+	 */
+	/* Temporarily store current stack pointer in scratch register */
+	movq	%rsp, %r11
+
+	/* Switch to the auxiliary stack so that we do not contaminate the
+	 * application's stack, as this could either be too small and result
+	 * in corrupted memory or we could modify unwanted variables stored
```
michpappas: > 	 * in corrupted memory or we could modify unwanted variables stored

Nitpick: Are the variables unwanted? Or do you mean to unwantedly modify variables in the red zone? In that case this statement is redundant as you mention corrupting memory right before that.

plat/common/x86/syscall.S
```diff
@@ -63,6 +65,8 @@ ENTRY(_ukplat_syscall)
 	 */
 	movq	%gs:LCPU_AUXSP_OFFSET, %rsp
 
+	pushq_cfi	$(GDT_DESC_OFFSET(GDT_DESC_DATA))
```
michpappas, line 72: Unless there is a reason to have these in separate commits, please squash this commit along with the previous one. The process taken to reach a certain state during development does not need to be reflected in the commits merged.

plat/common/lcpu.c
```diff
@@ -59,7 +59,7 @@
  * for non-present CPUs and does not force us to configure the maximum number
  * of CPUs beforehand.
  */
-struct lcpu lcpus[CONFIG_UKPLAT_LCPU_MAXCOUNT];
+UKPLAT_PER_LCPU_DEFINE(struct lcpu, lcpus);
```
michpappas, line 62: That's good, yet I noticed that the `lcpus` global is referenced directly in https://github.com/unikraft/unikraft/blob/staging/plat/kvm/x86/lcpu_start.S#L312. Since we are abstracting the implementation away here, it would be proper to also get a reference to the current `lcpu` via an accessor there too.

michpappas (changes requested):
Nice work! Besides the inline comments, would it be okay to open a separate PR for the Arm part, as I would like to give a bit more thought on the implementation of the IRQ / trap stack without blocking the `x86_64` part? I am planning to provide a review asap, so that we get it merged by early next week. Thanks a lot 🙏🏼

include/uk/plat/lcpu.h
```diff
@@ -231,8 +241,13 @@ int ukplat_lcpu_run(const __lcpuidx lcpuidx[], unsigned int *num,
 int ukplat_lcpu_wakeup(const __lcpuidx lcpuidx[], unsigned int *num);
 
 #else /* CONFIG_HAVE_SMP */
-#define ukplat_lcpu_idx()	(0)
-#define ukplat_lcpu_count()	(1)
+#define ukplat_lcpu_count()		(1)
+#if !CONFIG_PLAT_KVM
```
michpappas: Please please please, no platform dependencies in generic code 🙏🏼

lib/uksched/include/uk/sched.h
```diff
@@ -212,6 +217,7 @@ struct uk_thread *uk_sched_thread_create_fn1(struct uk_sched *s,
 					     uk_thread_fn1_t fn1,
 					     void *argp,
 					     size_t stack_len,
+					     size_t auxstack_len,
```
michpappas, line 221: Since the public functions of `uk_thread` are changed, please make sure no regressions are introduced in external libs. Example from `lib-lwip`:
```
/home/mpp/devel/unikraft/uk_vfork/app-elfloader/.unikraft/libs/lwip/threads.c: In function ‘sys_thread_new’:
/home/mpp/devel/unikraft/uk_vfork/app-elfloader/.unikraft/libs/lwip/threads.c:54:13: error: too few arguments to function ‘uk_sched_thread_create_fn1’
   54 |         t = uk_sched_thread_create_fn1(s,
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/mpp/devel/unikraft/uk_vfork/app-elfloader/.unikraft/libs/lwip/threads.c:33:
/home/mpp/devel/unikraft/uk_vfork/app-elfloader/.unikraft/unikraft/lib/uksched/include/uk/sched.h:216:19: note: declared here
  216 | struct uk_thread *uk_sched_thread_create_fn1(struct uk_sched *s,
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~
  CC      liblwip: ip.o
```

plat/common/include/x86/gsbase.h
mogasergiu: I was referring to the instructions themselves. There are also dedicated instructions to read/write from/to GS/FS, because MSR read/write is slow. But we don't have those yet (you need to set a bit in CR4 AFAIK) so I thought I would simply add a TODO. Also #862

plat/common/x86/lcpu.c
mogasergiu: Native builds will end up having `gs_base == k_gs_base == lcpu`. If a bincompat application ends up modifying `gs_base`, ok fine no problem we will know. The reason I still initialize both is because the `clone` code `uk_syscall_regs_popall` introduced in #1175 has a `swapgs` and is used by both native and bincompat, since they have to be able coexist.

plat/common/x86/syscall.S
mogasergiu: Oops, I meant `unwantedly`, correct, thanks :ok_hand: :sweat_smile:

plat/common/x86/syscall.S
mogasergiu: > The process taken to reach a certain state during development does not need to be reflected in the commits merged

I have a different opinion here actually. I believe that it is good for the commits and their messages to attempt to tell the story of the author's train of thought and how he has reached this feature: first, there came the stack switch implementation and **then** came the bonus implementation that also fills in these registers to their actual value. They are entirely unrelated IMO, so I would prefer I keep the two commits separate. In fact, this commit does not help that much to be honest, it's really a bonus so that we don't patch a `clone` child's `iretq` required registers to some known values and have them instead be inherited by the parent. That being said, this commit should've probably been part of #1175 instead.

plat/common/lcpu.c
mogasergiu: Agreed, I will add assembly macro's in the header :ok_hand:

include/uk/plat/lcpu.h
mogasergiu: This came as an afterthought. When I first submitted the PR, I was notified that builds were failing on `linuxu`/`xen` with undefined references because they don't have their own definitions of these. So I just quickly hacked in this `#if` to make the CI shut up. :sweat:

So, I would really appreciate some suggestions here as to how it should be handled instead.

lib/uksched/include/uk/sched.h
mogasergiu: The diff is simple, indeed, I had the modification locally and forgot about it. Fix is as simple as
```diff
diff --git a/threads.c b/threads.c
index ea7f008..f431256 100644
--- a/threads.c
+++ b/threads.c
@@ -53,7 +53,7 @@ sys_thread_t sys_thread_new(const char *name, lwip_thread_fn thread, void *arg,
 
     t = uk_sched_thread_create_fn1(s,
                        (uk_thread_fn1_t) thread, arg,
-                       (size_t) stacksize,
+                       (size_t) stacksize, 0,
                        false,
                        false,
                        name,
```


When we get close to merging this stuff I will submit a PR for this :+1:

include/uk/plat/memory.h
mogasergiu: I don't know about the TLS/stack. But the auxiliary stack is definetely something you would use in emergency cases or more fragile situations when a page fault should be avoided. I would say they are different in that regard IMO. As long as you don't have TLS accesses in a page fault handler that is...

mogasergiu: But now that you bring this up, you reminded me of something that I thought off some time ago, but I forgot. Can we guarantee that there will be no TLS accessses in IRQ/trap handlers? Because if not then, especially for binompat, we will have to be able to store/restore TLS pointer so as not to contaminate the application's TLS.

This needs to be kept in mind when reviewing additions to the IRQ/trap handling code.

michpappas: > I don't know about the TLS/stack. But the auxiliary stack is definetely something you would use in emergency cases or more fragile situations when a page fault should be avoided. I would say they are different in that regard IMO. As long as you don't have TLS accesses in a page fault handler that is...

To clarify after internal discussion with @mogasergiu: This is regarding uses of the aux stack for cases other than syscalls. In particular in native mode where the aux stack is not used for syscalls, madvise ensures that we avoid the case where the moment the aux stack is needed it has not been mapped to an actual page yet and will generate a page fault we are unable to handle.

> But now that you bring this up, you reminded me of something that I thought off some time ago, but I forgot. Can we guarantee that there will be no TLS accessses in IRQ/trap handlers? Because if not then, especially for binompat, we will have to be able to store/restore TLS pointer so as not to contaminate the application's TLS.

Given that IRQ handling is decoupled from the currently running thread, it sounds like a reasonable assumption to me that the handler won't need to access thread-specific information.

plat/common/x86/syscall.S
jschlumpp, line 1: The changes currently break unwinding over the syscall handler.

mogasergiu: Hmm, could it be due to wrong usage of the CFI directives :thinking:

include/uk/plat/memory.h
michpappas: Perhaps this function can be generalized for allocating stacks in general and add guards pages to avoid stack overflow corrupting memory. Also `ukvmem` provides `vmem_stack` that uses guards.

Besides `auxsp_alloc` we probably also need a `free` function for `uk_thread_release`.

plat/Config.uk
```diff
@@ -115,3 +115,7 @@ config FPSIMD
 	depends on ARCH_ARM_64
 	help
 		Enable support FPU usage in application
+
+config UKPLAT_AUXSP_SIZE
```
michpappas: In other cases we use order instead of size in bytes:
```
CONFIG_STACK_SIZE_PAGE_ORDER=8
CONFIG_LIBPOSIX_PIPE_SIZE_ORDER=16
CONFIG_LIBUKFALLOCBUDDY_MAX_ALLOC_ORDER=30
CONFIG_LIBVFSCORE_PIPE_SIZE_ORDER=16
```

mogasergiu: Oh, I forgot about this :man_facepalming:... I will address it in the next iteration... sorry :sweat:

mogasergiu:
Sorry, pushed the branch for #1175 instead... to compare do a git diff [e3ff15c](https://github.com/unikraft/unikraft/commit/e3ff15cfbe02e328b11593d9856ee5fb9e95e8e1) [c4c3abf](https://github.com/unikraft/unikraft/commit/c4c3abfad58421ae945a38bc77d851291922b4e1) instead :man_facepalming:

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -216,8 +214,8 @@ struct __regs {
 	/* Stack Pointer */
 	uint64_t sp;
 
-	/* Stack Pointer from el0 */
-	uint64_t sp_el0;
+	/* Padding to make sure this structure is 16-byte aligned */
+	__u8 pad[8];
```
michpappas: It would be useful to add compile-time sanity checks for `__TRAP_STACK_SIZE` and `__SP_OFFSET`, e.g. `sizeof(struct __regs) == __TRAP_STACK_SIZE` etc.

**Edit:** While reviewing #1175 I noticed you add a `CT_ASSERT` for `sizeof(regs)` in e493bd78177e39fc05654bb283a4c46c0ee8f933. Perhaps you can add all these `CT_ASSERTS` here and get rid of that patch in #1175.

Also you can use `__u8 pad[__REGS_PAD_SIZE]` instead of `__u8 pad[8]`.

plat/common/arm/lcpu.c
```diff
@@ -42,6 +45,20 @@
...
 #define CPU_ID_MASK 0xff00ffffffUL
 
+/*
+ *       STACK_SIZE               STACK_SIZE
+ * <---------------------><--------------------->
+ * |============================================|
+ * |                     |                      |
+ * |       trap stack    |        IRQ stack     |
+ * |                     |                      |
+ * |=============================================
```
michpappas, line 55: Some points worth to document:
1. **Why do we need separate stacks for IRQ and traps?** We could use a single stack for exceptions, but in that case when we had a nested case (eg an IRQ triggers a synchronous exception) we would need to check at the interrupt vector entry whether we are coming from an exception (ie keep using the exception stack) or not (switch to the exception stack).
2. **Why can one stack not corrupt the other?** During a trap IRQs are disabled. So only the trap stack could potentially corrupt the IRQ stack. The ordering of the stacks is made so that this cannot happen.

plat/kvm/arm/exceptions.S
```diff
@@ -66,34 +68,69 @@
 	EXCHANGE_SP_WITH_X0
 .endm
 
-.macro SAVE_REGS
-	sub sp, sp, #__TRAP_STACK_SIZE
+.macro SAVE_REGS, is_irq
+	/* Use TPIDRRO_EL0 as scratch register. It is fine to do so because
```
michpappas, line 72: Why can't the application modify `TPIDRRO_EL0` if it's also running in `EL1`?

plat/common/arm/lcpu.c
```diff
@@ -90,6 +108,10 @@ int lcpu_arch_init(struct lcpu *this_lcpu)
 
 	SYSREG_WRITE64(tpidr_el1, (__uptr)this_lcpu);
 
+	irqntrap_sp = (__uptr)&ukplat_per_lcpu_array_current(lcpu_irqntrap_sp,
+							     STACK_SIZE / 2);
```
michpappas: `STACK_SIZE / 2` or `STACK_SIZE`?

arch/arm/arm64/include/uk/asm/ctx.h
```diff
@@ -35,6 +35,10 @@
 #error Do not include this header directly
 #endif
 
+/* 32 * 2 * 8 bytes + 4 bytes (FPSR) + 4 bytes (FPCR) */
```
michpappas: Please provide a more clear comment on what `32 * 2 * 8` stands for (eg 32x 128-bit NEON registers).

include/uk/plat/memory.h
```diff
@@ -252,6 +257,63 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags);
...
+ *   accesses to the auxiliary stack do not generate page faults in more
+ *   fragile system states.
+ * @param auxsp_len
+ *   The custom length of the auxiliary stack. If 0, then
+ *   CONFIG_UKPLAT_AUXSP_SIZE is used instead as the default length.
+ *
+ * @return
+ *   Pointer to the allocated auxiliary stack
+ */
+static inline __uptr ukplat_auxsp_alloc(struct uk_alloc __maybe_unused *a,
```
michpappas, line 279: It would be worth considering making this function generic and guard other dynamically allocated stacks.

lib/ukboot/Makefile.uk
```diff
@@ -2,6 +2,8 @@ $(eval $(call addlib_s,libukboot,$(CONFIG_LIBUKBOOT)))
 
 CINCLUDES-$(CONFIG_LIBUKBOOT)		+= -I$(LIBUKBOOT_BASE)/include
 CXXINCLUDES-$(CONFIG_LIBUKBOOT)	+= -I$(LIBUKBOOT_BASE)/include
+LIBUKBOOT_ASINCLUDES-y        += -I$(UK_PLAT_COMMON_BASE)/include
```
michpappas: C*INCLUDES are so horribly aligned 😿 Maybe with the opportunity of adding the private includes maybe you can align these too? 🍦
```
CINCLUDES-$(CONFIG_LIBUKBOOT)   += -I$(LIBUKBOOT_BASE)/include
CXXINCLUDES-$(CONFIG_LIBUKBOOT) += -I$(LIBUKBOOT_BASE)/include

LIBUKBOOT_ASINCLUDES-y          += -I$(UK_PLAT_COMMON_BASE)/include
LIBUKBOOT_CINCLUDES-y           += -I$(UK_PLAT_COMMON_BASE)/include
```

plat/kvm/arm/exceptions.S
```diff
@@ -100,6 +100,25 @@
 	/* Store old SP previously saved into x0 */
 	str     x0, [sp, #-16]
 
+	/* Is this trap a system call? */
+	mrs	x0, esr_el1
+	and	x0, x0, #0x00000000fc000000
```
michpappas: I don't like these raw values 🥩 You should be using `ESR_EC_MASK`, `ESR_EC_SHIFT`, and `ESR_EL1_EC_SVC64`  from `arch/arm/arm64/include/uk/asm/arch.h` instead.

plat/common/x86/syscall.S
```diff
@@ -33,12 +33,69 @@
...
+	 * rely on the scratch register being maintained between thread switches
+	 */
+	pushq_reg_cfi r11
+	.cfi_rel_offset rsp, 0
+
+	/* We are now in a state where the stack looks like this:
+	 *	--------------- <-- auxsp (i.e. lcpu_get_current()->auxsp OR
+	 *	| app's saved |	           uk_thread_current()->auxsp)
+	 *	|     %rsp    |
+	 *	--------------- <-- (auxsp - 8) OR (**current %rsp**)
```
michpappas: Maybe use `i.e.` instead of `OR`.

plat/common/x86/syscall.S
```diff
@@ -33,12 +33,69 @@
...
+	/* Store application's stack pointer at the top of current thread's
+	 * auxiliary stack. We have to do this because we obviously can't
+	 * rely on the scratch register being maintained between thread switches
+	 */
+	pushq_reg_cfi r11
+	.cfi_rel_offset rsp, 0
+
+	/* We are now in a state where the stack looks like this:
+	 *	--------------- <-- auxsp (i.e. lcpu_get_current()->auxsp OR
+	 *	| app's saved |	           uk_thread_current()->auxsp)
```
michpappas: Maybe use `i.e` instead of `OR` to avoid confusion.

michpappas (changes requested):
@mogasergiu just a few minor comments, I think this is the last iteration 🙂

plat/Config.uk
michpappas: Bumping for the next update.

michpappas:
One more. In the following commit message:
```
    arch: Make context specific macro's visible in assembly code
    
    Simple macro's that do not involve C code may as well be visible to code
    written in assembly. This may come in handy when not wanting to use
    magic values.
    
    Therefore, make it so that architecture specific macro's from context
    related headers are also visible to assembly written code by having
    them defined outside the `!__ASSEMBLY` guards.
```
A `s/macro's/macros/` will keep the Apostrophe Protection Society happy.

arch/x86/x86_64/include/uk/asm/ctx.h
```diff
@@ -31,6 +31,15 @@
 #error Do not include this header directly
 #endif
 
+/* 512 for the actual structure + 64 bytes for the XSAVE header */
+#define UKARCH_ECTX_SAVE_MAX_SIZE			576
```
michpappas: Although ectx size can depend on cpu features in x86, that's not necessarily true for other architectures, so perhaps it would be nicer to use a unified name i.e. `UKARCH_ECTX_SIZE` (which aligns nicely with `UKARCH_STACK_SIZE`) and leave a comment in the case of x86_64 simiar to:
```
#define UKARCH_ECTX_SIZE    576 /* Max possible size */
```

plat/common/arm/lcpu.c
michpappas: Another perhaps simpler approach would be to define a single exception stack that uses `SP_EL0`. Then we can do `msr SPSel, #0` and solve the problem of whether the exception is nested or not. What do you think?

mogasergiu: I like the idea.

> and solve the problem of whether the exception is nested or not

I am not sure I understand how this would solve this, but what it definetely help us with is getting rid of the need for a scratch register. Good idea! Will try it out.

plat/kvm/arm/exceptions.S
mogasergiu: It can, but it won't **try** it :point_up:! I will `s/can't/won't try` :+1:

plat/common/arm/lcpu.c
mogasergiu: The latter :sweat: oops :laughing:

include/uk/plat/memory.h
mogasergiu: Agreed. But for now maybe outside the scope of this PR. I will definetely note it down as something to keep in mind :smile:. Perhaps we could do it statically somehow for those we use in `.bss`

plat/common/arm/lcpu.c
mogasergiu: Oh, man, unfortuantely, there are drawbacks. It does still work but it has its drawbacks. I am attaching here a diff for reference in case we ever have to come back to this (tested on Redis)
```diff
diff --git a/plat/kvm/arm/exceptions.S b/plat/kvm/arm/exceptions.S
index b02be2a9..e9c92032 100644
--- a/plat/kvm/arm/exceptions.S
+++ b/plat/kvm/arm/exceptions.S
@@ -88,10 +88,10 @@
 	 *                       ^
 	 *                     SP_EL0
 	 */
-	mrs     x0, sp_el0
+	mov     x0, sp
 
-	/* Make it so that SP contains SP_EL0 and x0 contains old SP */
-	EXCHANGE_SP_WITH_X0
+	msr	spsel, #0
+	mrs	x0, sp_el1
 
 .if \is_irq != 0
 	/* If we are an IRQ, use the IRQ stack instead. */
@@ -162,7 +162,7 @@
 	stp	x22, x23, [sp, #16 * 16]
 .endm
 
-.macro RESTORE_REGS
+.macro RESTORE_REGS, is_irq
 	/* Mask IRQ to make sure restore would not be interrupted by IRQ */
 	msr daifset, #2
 
@@ -181,7 +181,7 @@
 	ldp x24, x25, [sp, #16 * 12]
 	ldp x22, x23, [sp, #16 * 11]
 	ldp x20, x21, [sp, #16 * 10]
-	/* Skip x18, x19 */
+	ldp x18, x19, [sp, #16 * 9]
 	ldp x16, x17, [sp, #16 * 8]
 	ldp x14, x15, [sp, #16 * 7]
 	ldp x12, x13, [sp, #16 * 6]
@@ -192,13 +192,13 @@
 	ldp x2, x3, [sp, #16 * 1]
 	ldp x0, x1, [sp, #16 * 0]
 
-	/* Restore stack pointer for exception from EL1 */
-	ldr x18, [sp, #__SP_OFFSET]
-	mov x19, sp
-	mov sp, x18
+.if \is_irq != 0
+	/* If we are an IRQ, use the IRQ stack instead. */
+	add     sp, sp, #CPU_EXCEPT_STACK_SIZE
+.endif
 
-	/* Restore x18, x19 */
-	ldp x18, x19, [x19, #16 * 9]
+	/* Restore stack pointer for exception from EL1 */
+	msr	spsel, #0x1
 
 	eret
 .endm
@@ -214,7 +214,7 @@ el1_sync:
 	mov x0, sp
 	mrs x1, far_el1
 	bl trap_el1_sync
-	RESTORE_REGS
+	RESTORE_REGS 0
 
 .align 6
 el1_irq:
@@ -222,7 +222,7 @@ el1_irq:
 	msr daifclr, #(8 | 4 | 1)
 	mov x0, sp
 	bl trap_el1_irq
-	RESTORE_REGS
+	RESTORE_REGS 1
 
 /* Bad Abort numbers */
 #define BAD_SYNC  0
 ```
 
 But there are mainly two problems:
 - Once we write to `SPSel`, we also modify `SPSR_EL1` and the one that ends up being saved on the frame is not really the original one... although, yes, you can patch it afterwards knowing that this modification happened, but it is not nice looking at all
 - When we reach the point of return we must restore the stack to its previous value before restoring `SPSel` to `1` on normal execution, but in the case of a nested situation (e.g. trap happening during IRQ handling) we would need to check on return whether we really should restore SPSel or not, which ends up being not so nice looking either :cry: 
 
 All in all, this ends up resulting in having to do a tiny bit more instructions, however the biggest problem is design wise as we end up doing some ugly explicit patching and checking.

mogasergiu:
Changelog

765f4155890bafd6d3db2d673634b9c744200275
plat/kvm/arm: Obsolete `EL0` logic from trap/IRQ handling
- Squashed e493bd78177e39fc05654bb283a4c46c0ee8f933 of #1175 into it
- used `__REGS_PAD_SIZE` instead of `8` for `pad`
- define ARM64 __REGS_PAD_SIZE/__REGS_SIZEOF

51820f0c908fdc5b62b626a7fa06b9d8611d4848
arch/arm64: Implement stack switching on IRQ/trap
- `sp_el0` = `lcpu_irqntrap_sp` + `STACK_SIZE`
  not
  `sp_el0`= `lcpu_irqntrap_sp` + `STACK_SIZE / 2`
- Added `Some points worth to document` comments
! `s/can't/won't try` corrupt TPIDRRO_EL0
! SPSel0

cd25d18247dde2c800c161baa5d08e362b3043ad
arch/arm/arm64: Add ECTX save maximum alignment/size macro's
- Add NEON registers clarification comment from `32 * 2 * 8`
- `UKARCH_ECTX_SAVE_MAX_SIZE` -> `UKARCH_ECTX_SIZE`
- `UKARCH_ECTX_ALIGN` -> `UKARCH_ECTX_ALIGN`

c84dfbf41d9e1ef5dc9de56b24f96d2f67f93072
arch/x86/x86_64: Add ECTX save maximum alignment/size macro's
- `UKARCH_ECTX_SAVE_MAX_SIZE` -> `UKARCH_ECTX_SIZE`
- `UKARCH_ECTX_ALIGN` -> `UKARCH_ECTX_ALIGN`
- + Max possible/needed size/alignment comment

e97f51a46b20a7456ac0cfa6abe64e4ab05881f5
plat: Implement a method for allocating the auxiliary stack
- Use PAGE_ORDER instead of SIZE

e6f8ffbd409b8f3db5fc4296cbe5da0b8819cae1
lib/uksched: Introduce the per-thread auxiliary stack
- Use PAGE_ORDER instead of SIZE (auxsp config rename)

29054aab2dcd7d96c9c091e5245bedd1286be215
lib/ukboot: Allocate bootstrap LCPU auxiliary stack
- Align *INCLUDES in lib/ukboot/Makefile.uk

679d79c65a39c03156d54ba487b5217b4e9c9e15
plat/common/x86: Switch to auxiliary stack on syscall entry
- use `i.e.` instead of `OR`
- Use PAGE_ORDER instead of SIZE (auxsp config rename)

c4c3abfad58421ae945a38bc77d851291922b4e1
plat/kvm/arm: Switch to auxiliary stack on syscall entry
- use #ESR_EC_MASK/#ESR_EC_SHIFT/#ESR_EL1_EC_SVC64 instead of raw values

mogasergiu:
Also pulled 371d7030739d268c36bae14d09f8fd9f7461edea from #1175 here so that 1174 builds individually.

plat/common/arm/lcpu.c
michpappas: Agreed, adding checks on the exit sequence defeats the purpose of that change, as the idea was to simplify the design with `SPSel`. Thanks a lot for looking into it 🙏🏼

michpappas (approved):
@mogasergiu many thanks! 🥂 

razvand (approved):

PR #1175, opened 2023-11-24
Introduce `UK_LLSYSCALL_R_U_DEFINE` and register `clone` with it

mogasergiu:
### Prerequisite checklist

### Base target

 - Platform(s): `kvm`
 - Application(s): N/A

### Description of changes

Some runtimes assume that the child of `clone` inherits full context of the parent (except for scratch registers).
Introduce `struct uk_syscall_regs`:
```C
struct uk_syscall_regs {                                                                                                                                                      
       struct __regs regs;                                                                                                                                                    
       struct ukarch_ulctx ulctx;                                                                                                                                             
       __u8 ectx[UKARCH_ECTX_SAVE_MAX_SIZE];                                                                                                                                  
       __u8 pad[UK_SYSCALL_PAD_SIZE];                                                                                                                                         
};
```
A structure made up of architecture specific structures:
- `struct __regs` for general purpose registers
- new `struct ukarch_ulctx`, meant to help the kernel keep track of certain special registers an application may change and which we may want to know of in case we have to swap them with ours when entering a system call. E.g. TLS pointer in bincompat, or the `x86_64` `gs_base` register which an application may change through `arch_prctl` system call.
- `ectx` mean to represent the architecture specific slot where we may want to save/restore to/from when enterin/exiting a system call

Introduce `UK_LLSYSCALL_R_U_DEFINE`, a system call registration macro alternative that offers a system call access to the caller/parent's full `struct uk_syscall_regs` context. Two system calls shall be registered: `clone` and [`arch_prctl`](https://github.com/unikraft/app-elfloader/pull/65). Additionally, `uk_syscall6_r` and `ukplat_syscall_handler` shall now make use of this generic structure.

Thus, now `clone` is able to make the child inherit full context of the parent if desired and satisfy picky runtimes.

Depends on #1173 and #1174


lib/syscall_shim/arch/arm64/syscall_regs.S
```diff
@@ -0,0 +1,59 @@
...
+
+	/* Restore stack pointer */
+	ldr	x18, [sp, #__SP_OFFSET]
+	mov	x19, sp
+	mov	sp, x18
+
+	/* Restore x18, x19 */
+	ldp	x18, x19, [x19, #16 * 9]
+
+	msr	daifclr, #2
```
mogasergiu: Remembered about this while working on smth else, but basically this is not needed because `eret` will restore corresponding `PSTATE`.

TODO: Delete this line

lib/syscall_shim/arch/arm64/include/uk/ulctx.h
```diff
@@ -0,0 +1,20 @@
...
+
+#if !__ASSEMBLY__
+
+#include <uk/essentials.h>
+
+/* Architecture specific userland context */
+struct ukarch_ulctx {
+	__uptr tpidr_el0;
+
+	__u8 pad[8];	/* Make sure we are a multiple of 16 bytes */
```
michpappas: Perhaps instead of padding you can do:
```c
struct ukarch_ulctx {
 ...
} __align(__STACK_ALIGN_SIZE);
```

include/uk/essentials.h
```diff
@@ -105,6 +105,9 @@ extern "C" {
 #ifndef __fallthrough
 #define __fallthrough          __attribute__((fallthrough))
 #endif
+#ifndef __naked
+#define __naked                __attribute__((naked))
```
michpappas, line 112: If `__attribute__((naked))` is not supported by `arm64`, then perhaps this should be conditionally defined to `x86_64` or to `!arm64`, or at least add a comment for people who are not aware?

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -399,6 +399,107 @@ typedef long uk_syscall_arg_t;
...
+ * lots of build errors.
+ * TODO: Plat re-architecting will help with this, but for now, simply
+ * re-define this macro to not waste time on this trivial matter.
+ */
+#ifndef LCPU_AUXSP_OFFSET
+#define LCPU_AUXSP_OFFSET		0x20
+#endif /* !LCPU_AUXSP_OFFSET */
+
+#if defined(__X86_64__)
+#define UK_SYSCALL_USR_PROLOGUE_DEFINE(pname, fname, x, ...)		\
```
michpappas: This should be defined in an architecture-specific header. Also, since this looks a bit strange, could you please provide some context on why this defined in a header as a macro and not in a separate assembly file?

include/uk/essentials.h
```diff
@@ -357,6 +357,10 @@ extern "C" {
 #define UK_NARGS(...) \
 	__UK_NARGS_X(, ##__VA_ARGS__, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, \
 		     15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+#define UK_NACTUAL_ARGS(...) \
```
michpappas: This is a bit too tailored to `syscall_shim` to belong to `essentials.h` IMO. A comment block that provides some context would also be helpful.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -213,6 +213,73 @@ typedef long uk_syscall_arg_t;
 #define __UK_NAME2SCALLE_FN(name) UK_CONCAT(uk_syscall_e_, name)
 #define __UK_NAME2SCALLR_FN(name) UK_CONCAT(uk_syscall_r_, name)
 
+#define UK_USR_MAP0(...)
+/* Taken from regmap_linuxabi.h, but defined differently so as not to
+ * needlessly contaminate the namespace of source files including this header
+ */
+#if (defined __X86_64__)
```
michpappas: Similarly to other cases, please use architecture-specific headers.

lib/syscall_shim/arch/x86_64/ulctx.c
```diff
@@ -0,0 +1,32 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#include <uk/arch/ctx.h>
```
michpappas: These look like a lot of headers for these two functions, I would expect you'd only need around half of them.

lib/syscall_shim/arch/x86_64/include/uk/ulctx.h
```diff
@@ -25,6 +25,10 @@ UK_CTASSERT(sizeof(struct ukarch_ulctx) == UKARCH_ULCTX_SIZE);
 __uptr ukarch_ulctx_get_tlsp(struct ukarch_ulctx *ulctx);
 
 void ukarch_ulctx_set_tlsp(struct ukarch_ulctx *ulctx, __uptr tlsp);
+
+void ukarch_ulctx_switchoff_tls(struct ukarch_ulctx *ulctx);
```
michpappas: I'm not 100% keen on these names, as they imply that the TLS is something you toggle on and off 💡, while what we actually do is switch between user / kernel TLS. Besides, it's easy to forget after a while which one is `on` and which one is `off`. Something along the lines of `ukarch_ulctx_switch_ultls()` / `ukarch_ulctx_switch_uktls()` would be more clear IMO.

Edit: After seeing `ukarch_ulctx_switchon()` / `ukarch_ulctx_switchoff()` I get the idea. Still, perhaps it would be a bit more clear to use `_ultls` instead of `_tls`? If everyone thinks they're clear, I don't have a strong opinion though.

lib/syscall_shim/arch/x86_64/ulctx.c
```diff
@@ -29,4 +29,27 @@ void ukarch_ulctx_set_tlsp(struct ukarch_ulctx *ulctx, __uptr tlsp)
 
 	ulctx->fs_base = tlsp;
 }
+
+void ukarch_ulctx_switchoff_tls(struct ukarch_ulctx *ulctx)
+{
+	struct uk_thread *t = uk_thread_current();
+
+	UK_ASSERT(ulctx);
```
michpappas: It would be useful if we somehow could assert that we are currently in `switchon` state. I don't have a good idea on how though 🤔

lib/syscall_shim/arch/x86_64/ulctx.c
```diff
@@ -10,8 +10,42 @@
...
 #include <uk/thread.h>
+#include <x86/gsbase.h>
 #include <uk/syscall.h>
 
+void ukarch_ulctx_switchoff(struct ukarch_ulctx *ulctx)
+{
+	UK_ASSERT(ulctx);
+	UK_ASSERT(lcpu_get_current());
+
+	/* This can only be called from Unikraft ctx in bincompat mode.
```
michpappas: Similarly to `ukarch_ulctx_switchon_tls()`, an assertion that we're operating in Unikraft context would be useful. Similarly to `ukarch_ulctx_switchon_tls()`, I have no good ideas to propose 🤔 

I would also add that it's not enough to be in Unikraft context, but also in syscall / switchon context.

lib/syscall_shim/arch/x86_64/ulctx.c
```diff
@@ -10,8 +10,42 @@
...
+
+#if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
+	ukarch_ulctx_switchoff_tls(ulctx);
+#endif /* CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS */
+}
+
+void ukarch_ulctx_switchon(struct ukarch_ulctx *ulctx)
+{
+	UK_ASSERT(ulctx);
+	UK_ASSERT(lcpu_get_current());
```
michpappas: Maybe leave a comment on what does this protect from? (also use the `IS_LCPU_PTR`)

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -35,18 +35,66 @@
 #ifndef __UK_SYSCALL_H__
 #define __UK_SYSCALL_H__
 
+#include <uk/ulctx.h>
+
+#if defined(__X86_64__)
```
michpappas: These should be in architecture-specific headers (eg `lib/syscall-shim/arch/x86_64/include/uk/arch/syscall.h`). That way you also don't have to deal with the "Undefined architecture" problem.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -35,18 +35,66 @@
...
+/* 512 for the actual structure + 64 bytes for the XSAVE header */
+#define UKARCH_ECTX_SAVE_MAX_SIZE			578
```
michpappas: 512 + 64 = 576, not 578 🤔

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -35,18 +35,66 @@
...
 #include <stdarg.h>
 #include <uk/print.h>
 #include "legacy_syscall.h"
+#include <uk/thread.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+struct uk_syscall_regs {
```
michpappas: The naming here is a bit messed up 😵‍💫 
- This struct contains the entire cpu context, ie more than regs, yet it's named `regs`. Since context normally refers to the entire CPU state, so that would be a more appropriate name for this structure.
- For what we call context in this PS (which is not really context - only `FS`/`GS`), we use `ulctx`, yet for the extended context we call it just `ectx`.

~~But if you guys are happy with it, I don't have a strong opinion at all :)~~ Edit: Actually the more I see it, the more I don't like it, because it makes calling code unintuitive.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -399,6 +399,107 @@ typedef long uk_syscall_arg_t;
...
+ * TODO: Plat re-architecting will help with this, but for now, simply
+ * re-define this macro to not waste time on this trivial matter.
+ */
+#ifndef LCPU_AUXSP_OFFSET
+#define LCPU_AUXSP_OFFSET		0x20
+#endif /* !LCPU_AUXSP_OFFSET */
+
+#if defined(__X86_64__)
+#define UK_SYSCALL_USR_PROLOGUE_DEFINE(pname, fname, x, ...)		\
+	long __used __naked __attribute__((optimize("O3")))		\
```
michpappas: Shouldn't that be `O0` instead?

michpappas (changes requested):
@mogasergiu amazing as always 😎 I am leaving some first comments. I might provide additional feedback after porting my work. Thanks!

PS: As with #1174 perhaps we can move the Arm stuff to a separate PR 🙏🏼

lib/posix-process/clone.c
```diff
@@ -305,6 +307,101 @@ static void _clone_child_gc(struct uk_thread *t)
 	}
 }
 
+void uk_syscall_regs_popall(void);
+
+#if defined(__X86_64__)
```
michpappas: Please move these into `lib/posix-process/arch/...`

lib/syscall_shim/include/uk/syscall.h
mogasergiu: This needs to be expanded to a very specific function signature whose inline assembly calls a very specific function w.r.t. to the arguments passed to the macro. Each system call registered through this macro will have its own personalized defined prologue.

lib/syscall_shim/arch/x86_64/ulctx.c
mogasergiu: I used to have these assertions but they are not quite compatible with the native use-case where the GS register pair is equal :(. But I could add the `IS_LCPU_PTR` :+1:

lib/syscall_shim/include/uk/syscall.h
mogasergiu: :nerd_face: :computer: :brain: :mage_man:

lib/syscall_shim/include/uk/syscall.h
mogasergiu: I shall introduce a nerw structure `struct ukarch_actx`, which will include `(__regs, s/ulctx/{fs/gs + tpidr_el0/pad}, ectx)` and all `*ctx*` stuff shall go intro `include/uk/arch/actx.h`

lib/syscall_shim/include/uk/syscall.h
mogasergiu: Will keep `O3` on ARM and naked only on x86 :+1:. Also add a comment. All I need is for the ABI compliant prologue/epilogue to not be generated and, unfortuantely, ARM does not support `naked`

lib/posix-process/clone.c
```diff
@@ -376,20 +472,20 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
+		uk_pr_debug(" parent_tid: %p\n", (void *)cl_args->parent_tid);
 	if (flags & (CLONE_CHILD_CLEARTID | CLONE_CHILD_SETTID))
-		uk_pr_debug(" child_tid: %p\n", (void *) cl_args->child_tid);
-	uk_pr_debug(" stack: %p\n", (void *) cl_args->stack);
-	uk_pr_debug(" tls: %p\n", (void *) cl_args->tls);
-	uk_pr_debug(" <return>: %p\n", (void *) return_addr);
+		uk_pr_debug(" child_tid: %p\n", (void *)cl_args->child_tid);
+	uk_pr_debug(" stack: %p\n", (void *)cl_args->stack);
+	uk_pr_debug(" tls: %p\n", (void *)cl_args->tls);
+	uk_pr_debug(" <return>: %p\n", (void *)usr->rip);
```
michpappas: Should be: usr->regs.rip

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -351,6 +367,250 @@ typedef long uk_syscall_arg_t;
...
+		long ret;						\
+									\
+		ret = rname(_usr);					\
+		if (ret < 0 && PTRISERR(ret)) {				\
+			errno = -(int) PTR2ERR(ret);			\
+			return -1;					\
+		}							\
+		return ret;						\
+	}								\
+	static inline rtype __##rname(struct uk_syscall_regs *usr,	\
```
michpappas: This one is problematic in syscalls with no args like `vfork`, as it evaluates to:
```
__##rname(struct uk_syscall_regs *usr, )
```
To reproduce:
```
UK_LLSYSCALL_R_U_DEFINE(pid_t, vfork)
```

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -61,110 +109,79 @@ extern "C" {
...
+#define usr_arg1		x[1]
+#define usr_arg2		x[2]
+#define usr_arg3		x[3]
+#define usr_arg4		x[4]
+#define usr_arg5		x[5]
+#else
+#error "Missing register mappings for selected target architecture"
+#endif  /* !__X86_64__ && !__ARM_64__ */
+
+#define UK_USR_MAP2_1(m, type, arg) (long)usr->regs.usr_arg0
```
michpappas: `UK_LLSYSCALL_R_U_DEFINE` with 1 param don't work because of `UK_USR_MAP2_7` missing.

To reproduce:
```
UK_LLSYSCALL_R_U_DEFINE(int, mysyscall, unsigned long, arg)
```

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -399,6 +399,107 @@ typedef long uk_syscall_arg_t;
...
+				     __REGS_SIZEOF)"), %%rsp\n\t"	\
+		"/* Now build stack frame beginning with 5 pointers\n\t"\
+		" * in the classical iretq/`struct __regs` format\n\t"	\
+		" */\n\t"						\
+		"/* Push stack segment, GDT data segment selector:\n\t"	\
+		" * [15: 3]: Selector Index - second GDT entry\n\t"	\
+		" * [ 2: 2]: Table Indicator - GDT, table 0\n\t"	\
+		" * [ 1: 0]: Requestor Privilege Level - ring 0\n\t"	\
+		" */\n\t"						\
+		"pushq	$(0x10)\n\t"					\
```
michpappas: This could be a `#define` instead of magic. Same for line 448.

mogasergiu: Sadly, just like other magic values for ARM, the headers defining the macro's for these magic values would introduce even more dependencies for `syscall_shim`, which, in turn, more dependencies for others that include `syscall.h` (which are many). Therefore, I chose to add very descriptive comments instead :disappointed:

mogasergiu:
@michpappas @skuenzer Ok, all comments should've been addressed.
One thing left to do: renamings?
- `struct uk_syscall_regs` -> `struct uk_syscall_ctx`? Keep in mind this will mean also changing all abbreviations from `USR`/`usr` to `USC`/`usc`.
- maybe `struct ukarch_ulctx`? idk

mogasergiu:
Another force-push to fix an implcit declaration warning of `clone_setup_child_ctx`

include/uk/essentials.h
```diff
@@ -360,6 +360,10 @@ extern "C" {
 #define UK_NARGS(...) \
 	__UK_NARGS_X(, ##__VA_ARGS__, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, \
 		     15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+#define UK_NARGS_HALF(...) \
```
michpappas: It would be useful to write a comment to explain why is that macro required and how it's used.

Also IIRC in the last review round it was agreed that this should be in `libsyscall_shim` as it's specific to that library. Does the change in name change that? Are there any other use-cases? If the answer is no then perhaps both `UK_NARGS` and `UK_NARGS_HALF` belong to `libyscall_shim`? @skuenzer any thoughts?

plat/common/arm/traps_arm64.c
```diff
@@ -224,13 +225,13 @@ void trap_el1_irq(struct __regs *regs)
...
 
-extern void ukplat_syscall_handler(struct __regs *r);
+extern void ukplat_syscall_handler(struct uk_syscall_regs *usr);
 
 static int arm64_syscall_adapter(void *data)
 {
 	struct ukarch_trap_ctx *ctx = (struct ukarch_trap_ctx *)data;
 
-	ukplat_syscall_handler(ctx->regs);
+	ukplat_syscall_handler((struct uk_syscall_regs *)ctx->regs);
```
michpappas: Although not introduced in this PR, I'm not entirely happy with `ukarch_trap_ctx`. We have too many definitions of context (also `ukarch_ctx`). But let it be for now, we can look at it later 🙂

michpappas (changes requested):
@mogasergiu thanks for the updates! Besides the few minor comments below I think there are only a couple of naming issues left to go.

For `switchon` / `switchoff` I would propose something less ambiguous. One idea that comes to mind is `ukarch_ulctx_switch_ul` / `ukarch_ulctx_switch_uk` (userland / unikraft) or similar.

Also less important:
- `struct uk_syscall_regs`: This is actually a context, as by context one normally refers to the collective CPU state (yes _technically speaking_ it does consist of registers so _technically speaking_ the current name is also correct). One idea is to name it `struct uk_syscall_context` if it stays in the syscall shim, or perhaps more correctly make it generic, call it `uk_lcpu_context` and move it into `lcpu`. Or leave it as is.
- `struct __regs` -> `struct uk_lcpu_regs`. A bit of a misnomer as the rest of the struct is also registers, but we can keep it. Alternatively `struct uk_lcpu_gpregs` (general purpose regs). Or leave it as is and we rename it along with everything else once we migrate into `libuklcpu`.
- `struct ukarch_ulctx`: Should probably be `struct uk_lcpu_sysregs` or something. I think this one should actually be renamed in this PR, as `ctx` is definitely misused here.

include/uk/essentials.h
skuenzer: `UK_NARGS` is generic enough to keep it here. It is just a way for using the pre-processor to count the number of arguments that a variadic macro got handed over. This information can be used to build more complex macros on top. `UK_NARGS_HALF` looks to me syscall_shim specific because I remember we had a macro like this originally in syscall_shim but I was able to rewrite the macros in syscall_shim to base them also on `UK_NARGS`. Is it really needed for your case?

plat/common/arm/traps_arm64.c
skuenzer: Hum... I agree with this point that we have too many `ctx`es, but we probably need to clean this up later. What we also should avoid is to introduce contexts that are basically the same thing and we do need to do typecasting of the same pointer.
I wonder if `struct ukarch_trap_ctx` could be handed over directly to `ukplat_syscall_handler` and if this makes sense also for x86. :thinking: What do we have more on `ukarch_trap_ctx`?

include/uk/essentials.h
mogasergiu: I use it for `UK_USR_CALLMAPx`. Thought I would pass the actual number of arguments.. but I guess I could just use the double of that and just pretend I am using the actual number of args but I will have to have a look again because I don't remember much.

plat/common/arm/traps_arm64.c
mogasergiu: **What do we have less**:
```C
struct ukarch_trap_ctx {
	struct __regs *regs;
	int trapnr;
	int error_code;

	unsigned long fault_address; /* for page faults */
};
```
```C
struct uk_syscall_regs {
	struct __regs regs;
	struct ukarch_ulctx ulctx;
	__u8 ectx[UKARCH_ECTX_SAVE_MAX_SIZE];
	__u8 pad[UK_SYSCALL_PAD_SIZE];
};
```

mogasergiu:
Changelog

e493bd78177e39fc05654bb283a4c46c0ee8f933
arch/arm/arm64: Add constant assertions to register structures
- squashed into 765f4155890bafd6d3db2d673634b9c744200275 of #1174

9c940ab4bb91a1f904fc1045ffb9c7633945b5ef
lib/syscall_shim/arch/x86_64: Add fs_base getter/setter for ulctx
fdd52f147d98434b6ceb9fdadaaada7c8b1f013b
lib/syscall_shim/arch/arm64: Add tpidr_el0 getter/setter for ulctx
cc4a48aa37260bf03936856ff7ddda1a4c95a7c2
lib/syscall_shim/arch/x86_64: Add gs_base getter/setter
- s/ulctx/sysregs
- s/ULCTX/SYSREGS

44b13c828dbfa18606b15c94405d273b145e7b25
lib/syscall_shim/arch/x86_64: Implement TLS switchon/switchoff
e90361bfe78d00c097ecd3655b55b79aa2124f41
lib/syscall_shim/arch/arm64: Implement TLS switchon/switchoff
390cf661e698ce729be93c96c4af19bb9334eab4
lib/syscall_shim/arch/x86_64: Add ulctx switchon/switchoff
4fcbd9c73eca35ad7b97cc7789c310be02291664
lib/syscall_shim/arch/arm64: Add ulctx switchon/switchoff
- s/ulctx/sysregs
- s/ULCTX/SYSREGS
- switchon/switchoff -> switch_ul/switch_uk

006a5db79d1303a8cc2b55fd3be0c568eb94c914
lib/syscall_shim: Introduce struct uk_syscall_regs
- s/UK_SYSCALL_REGS*/UK_SYSCALL_CTX
- s/uk_syscall_regs/uk_syscall_ctx
- `UKARCH_ECTX_SAVE_MAX_SIZE` -> `UKARCH_ECTX_SIZE`
- `UKARCH_ECTX_ALIGN` -> `UKARCH_ECTX_ALIGN`
- s/ulctx/sysregs
- s/ULCTX/SYSREGS

3da0f6bc95c531e5ace6324eef1f50111f6a7ce3
lib/syscall_shim/arch/x86: Add uk_syscall_regs saving prologue
a3490253d5e1f6bf367b6bfd8c4bc89b2020c4cd
lib/syscall_shim/arch/arm64: Add uk_syscall_regs saving prologue
- s/UK_SYSCALL_REGS*/UK_SYSCALL_CTX
- s/uk_syscall_regs/uk_syscall_ctx
- s/ulctx/sysregs
- s/ULCTX/SYSREGS
- s/USR/USC
- switchon/switchoff -> switch_ul/switch_uk

962b429f6c4765691149d53d86e90c994d04fd7c
lib/syscall_shim/arch/x86_64: Add struct uk_syscall_regs restore
e60f0a96828e6b46ab3fbbe099a91948d4c5014f
lib/syscall_shim/arch/arm64: Add struct uk_syscall_regs restore
- s/UK_SYSCALL_REGS*/UK_SYSCALL_CTX
- s/uk_syscall_regs/uk_syscall_ctx
- s/ulctx/sysregs
- s/ULCTX/SYSREGS
- switchon/switchoff -> switch_ul/switch_uk

1352cfc6621837601f80fd81a1528dab02c3b330
lib/syscall_shim: Define UK_LLSYSCALL_R_U_DEFINE
20881b33784986d9d40fa2260b81f1b991d8ce75
lib/syscall_shim: Save struct uk_syscall_regs on binary syscalls
47da6d8ef5fd831de6d1efb6dea254ec25f7fd24
lib/syscall_shim: Save struct uk_syscall_regs in uk_syscall6_r
be9245a7a6673bca4cf81159e48eeccc195fa933
lib/posix-process: Copy full parent registers context on clone
- s/UK_SYSCALL_REGS*/UK_SYSCALL_CTX
- s/uk_syscall_regs/uk_syscall_ctx
- s/ulctx/sysregs
- s/ULCTX/SYSREGS
- switchon/switchoff -> switch_ul/switch_uk
- s/USR/USC

Deleted
bd70beecbf8362d8c6353dc8d33e351694cf669b
include/essentials: Define UK_NARGS_HALF macro
and doubled suffixes of UK_USC_EMAP/UK_USC_CALLMAP instead

mogasergiu:
Also moved https://github.com/unikraft/unikraft/commit/371d7030739d268c36bae14d09f8fd9f7461edea from https://github.com/unikraft/unikraft/pull/1175 to #1174 so that #1174 builds individually.

michpappas (approved):

@mogasergiu once again thanks a lot for going through the length process for this merge!
@razvand this needs to be rebased and tested together with dependencies before it can be merged.

razvand (approved):

plat/kvm/arm/exceptions.S
```diff
@@ -112,18 +113,27 @@
...
 
 	/* We now have in SP the trap stack and in x0 the auxiliary stack */
 	EXCHANGE_SP_WITH_X0  /* Switch them */
 	/* Restore old SP we stored before system call check */
 	ldr	x0, [x0, #-16]
-	str	x0, [sp, #-16]  /* Store old SP in auxiliary stack */
+	str	x0, [sp, #__SP_OFFSET]  /* Store old SP in auxiliary stack */
+	b	1f
 0:
 	/* Restore x0 */
```
Socratesa, line 132: Why not restore x0 in the case of a system call

mogasergiu: Yeah you're right, x0 should have been restored on syscalls as well of course. I must have somehow wrongly changed this during the review process and not notice. The restoration of x0 should have been placed after the `1:` label instead.

razvand: @mogasergiu , would this require an update (PR) from your side?

mogasergiu: Yes I could look into it after I return from my vacation, unless someone else would like to have a go at it as well. Though the fix should be rather simple: moving the restoration of x0 two lines below should suffice if I see this right.

mogasergiu: Done #1256

Issue #37, opened 2019-10-15
Recompile an object whenever compiler flags have changed


felipehuici:
Make is not taking compile flags into account for recompilation. We have some weird cases where we have to `make clean` and `make` again to solve some problems. Examples:
* Whenever we switch between `newlib` and internal `nolibc`
* A library is configured by passing MACRO definition as compiler flag (`-D`) instead of the gnerated KConfig _config.h header
Since we introduced fixdep from Kbuild, we have almost all pieces for doing this stuff. We just need to port some rules from kbuild.rules (Linux) to make it work. The .d files store already the build command line.

PR #1199, opened 2023-12-11
Make Unikraft compatible with newlib again

skuenzer:
This PR makes the Unikraft compatible with newlib again. The hope is that in general libc compatibility is improved. Please note that still not all Unikraft libraries are compatible with newlib at the moment (e.g., posix-mmap) but the minimum configuration works. This PR depends on PR [#37](https://github.com/unikraft/lib-newlib/pull/37) of newlib.

### DIFF
--- a/include/uk/essentials.h
+++ b/include/uk/essentials.h
@@ -39,6 +39,7 @@
 #include <uk/config.h>
 
 #if CONFIG_LIBNEWLIBC
+#ifndef __ASSEMBLY__
 /*
  * Needed for __used, __unused, __packed, __section,
  *   __nonnull, __offsetof, __containerof
@@ -48,6 +49,7 @@
 #include <sys/param.h>
 /* Needed for MIN, MAX */
 #include <inttypes.h>
+#endif /* !__ASSEMBLY__ */
 #endif
 
 #ifdef __cplusplus
--- a/lib/posix-fdio/fdio.c
+++ b/lib/posix-fdio/fdio.c
@@ -6,6 +6,7 @@
 
 /* Internal syscalls for file I/O */
 
+#include <stdio.h>
 #include <poll.h>
 
 #include <uk/posix-fdio.h>
--- a/lib/posix-fdtab/fdtab.c
+++ b/lib/posix-fdtab/fdtab.c
@@ -5,6 +5,7 @@
  */
 
 #include <errno.h>
+#include <limits.h>
 #include <fcntl.h>
 
 #include <uk/alloc.h>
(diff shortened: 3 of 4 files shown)


mariapana (commented):
When attempting to use `newlib` for the `helloworld` app, the following error occurs:

```
 LD      helloworld_qemu-x86_64.dbg
/usr/bin/ld: /home/mia/Unikraft/workdir/apps/helloworld/workdir/build/libnewlibc.o: in function `__collate_err':
/home/mia/Unikraft/workdir/apps/helloworld/workdir/build/libnewlibc/origin/newlib-2.5.0.20170922/newlib/libc/posix/collate.c:192: undefined reference to `exit'
make[3]: *** [/home/mia/Unikraft/workdir/apps/helloworld/workdir/unikraft/plat/kvm/Linker.uk:54: /home/mia/Unikraft/workdir/apps/helloworld/workdir/build/helloworld_qemu-x86_64.dbg] Error 1
make[2]: *** [Makefile:1170: sub-make] Error 2
make[1]: *** [Makefile:41: _all] Error 2
make[1]: Leaving directory '/home/mia/Unikraft/workdir/apps/helloworld/workdir/unikraft'
make: *** [Makefile:7: all] Error 2
```

Similar errors occur for different apps as well (for example not being able to find `sys/cdefs.h`).

eduardvintila (changes requested):
Hey, @skuenzer, thanks for the good work. Please rebase the PR onto the latest staging branch and solve the conflicts. Otherwise, there are several problems here which prevent complex applications from successfully compiling:

1. The assembly guards needs to be added to the `compiler.h` source file, as well.
2. `uklibid` has problems with handling libraries which have a dash inside their name. Namely, `pthread-embedded` cannot be compiled with it (line 16 from the `selfids.awk` file in `lib/uklibid`).
3. Other problems with vfscore integration, such as incompatibility with the prototype of the `getcwd` syscall

skuenzer:
> Hey, @skuenzer, thanks for the good work. Please rebase the PR onto the latest staging branch and solve the conflicts. Otherwise, there are several problems here which prevent complex applications from successfully compiling:
> 
>     1. The assembly guards needs to be added to the `compiler.h` source file, as well.
> 
>     2. `uklibid` has problems with handling libraries which have a dash inside their name. Namely, `pthread-embedded` cannot be compiled with it (line 16 from the `selfids.awk` file in `lib/uklibid`).
> 
>     3. Other problems with vfscore integration, such as incompatibility with the prototype of the `getcwd` syscall

Thanks, yes, I'll do a rebase. I need to check if some further breaking things have been introduced since I pushed this version so I need to adopt the PR. The focus is on getting the minimal Unikraft configuration (no extra libs, no external app) working again so that we can move it from there to get full support back. `pthread-embedded` isn't even supported since a while, actually since scheduler rework. Tests need to be done without `pthread-embedded`.

However, I'd like to integrate newlib back to some pipeline tests because it makes the Unikraft base much more resistent to specifc libc characteristics, a more solid code base of the core repository.

PR #1201, opened 2023-12-12
lib/vfscore: initrd: `extract` option

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_LIBVFSCORE=y`
 - `CONFIG_LIBVFSCORE_FSTAB=y`

### Description of changes

This PR introcues the mount option "extract" to the virtual `initrd` filesystem. This option will only perform an extraction of a CPIO archive over an already mounted filesystem. Its intention is to use it as sort of emulating an overlay filesystem.

### DIFF
--- a/lib/ukcpio/cpio.c
+++ b/lib/ukcpio/cpio.c
@@ -36,6 +36,7 @@
 #include <string.h>
 #include <stdint.h>
 #include <limits.h>
+#include <errno.h>
 
 #include <uk/assert.h>
 #include <uk/print.h>
@@ -141,6 +142,7 @@ int uk_syscall_r_chmod(const char *, mode_t);
 int uk_syscall_r_utime(const char *, const struct utimbuf *);
 int uk_syscall_r_mkdir(const char *, mode_t);
 int uk_syscall_r_symlink(const char *, const char *);
+int uk_syscall_r_stat(const char *, struct stat *);
 
 static enum ukcpio_error
 extract_file(const char *path, const char *contents, size_t len,
@@ -153,7 +155,7 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT|O_RDWR, 0);
+	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
 	if (fd < 0) {
 		uk_pr_err("%s: Failed to create file: %s (%d)\n",
 			  path, strerror(-fd), -fd);
@@ -201,12 +203,35 @@ extract_dir(const char *path, mode_t mode)
 	int r;
 
 	uk_pr_info("Creating directory %s\n", path);
-	if ((r = uk_syscall_r_mkdir(path, mode))) {
-		uk_pr_err("%s: Failed to create directory: %s (%d)\n",
-			  path, strerror(-r), -r);
-		return -UKCPIO_MKDIR_FAILED;
+	r = uk_syscall_r_mkdir(path, mode);
+	if (r == -EEXIST) {
+		struct stat pstat;
+
+		r = uk_syscall_r_stat(path, &pstat);
+		if (r < 0) {
+			r = -EEXIST;
+			goto err_out;
+		}
+		if (!S_ISDIR(pstat.st_mode)) {
+			r = -EEXIST;
+			goto err_out;
+		}
+
+		/* Directory already exists, set mode only */
+		r = uk_syscall_r_chmod(path, mode);
+		if (r < 0) {
+			uk_pr_warn("%s: Failed to chmod: %s (%d)\n",
+				   path, strerror(-r), -r);
+		}
+	} else if (r < 0) {
+		goto err_out;
 	}
 	return UKCPIO_SUCCESS;
+
+err_out:
+	uk_pr_err("%s: Failed to create directory: %s (%d)\n",
+		  path, strerror(-r), -r);
+	return -UKCPIO_MKDIR_FAILED;
 }
 
 static enum ukcpio_error
--- a/lib/vfscore/automount.c
+++ b/lib/vfscore/automount.c
@@ -60,6 +60,7 @@
 #define LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP			':'
 #define LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP			','
 #endif /* CONFIG_LIBVFSCORE_FSTAB */
+#define LIBVFSCORE_INITRD_OPT_EXTRACT				"extract"
 
 struct vfscore_volume {
 	/* Volume source device */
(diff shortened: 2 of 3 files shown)


StefanJum (approved):
Works fine, thanks.

lib/vfscore/automount.c
```diff
@@ -79,17 +80,26 @@ struct vfscore_volume {
...
-	rc = mount("", path, "ramfs", 0x0, NULL);
-	if (unlikely(rc)) {
-		uk_pr_crit("Failed to mount ramfs to \"%s\": %d\n",
-			   path, errno);
-		return -1;
+	/* 'extract' is an (e)initrd-specific mount option that allows
+	 * extracting an archive to an existing filesystem. If this option
+	 * is chosen, no other mount option is accepted.
+	 */
+	if (!opts || (strcmp(opts, LIBVFSCORE_INITRD_OPT_EXTRACT) != 0)) {
```
mogasergiu, line 94: Would it make sense to introduce an option macro for `mkmp` as well?

lib/vfscore/automount.c
```diff
@@ -60,6 +60,7 @@
 #define LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP			':'
 #define LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP			','
 #endif /* CONFIG_LIBVFSCORE_FSTAB */
+#define LIBVFSCORE_INITRD_OPT_EXTRACT				"extract"
```
mogasergiu, line 63: Maybe let's go with `UKOPT` instead of just `OPT`, to align it with the `ukopts` struct field member.

lib/ukcpio/cpio.c
```diff
@@ -201,12 +203,35 @@ extract_dir(const char *path, mode_t mode)
 	int r;
 
 	uk_pr_info("Creating directory %s\n", path);
-	if ((r = uk_syscall_r_mkdir(path, mode))) {
-		uk_pr_err("%s: Failed to create directory: %s (%d)\n",
-			  path, strerror(-r), -r);
-		return -UKCPIO_MKDIR_FAILED;
+	r = uk_syscall_r_mkdir(path, mode);
+	if (r == -EEXIST) {
```
mogasergiu, line 207: Would it make sense to have `unlikely`'s here and in the following error if cases?

lib/ukcpio/cpio.c
```diff
@@ -153,7 +155,7 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT|O_RDWR, 0);
+	fd = uk_syscall_r_open(path, O_CREAT|O_WRONLY|O_TRUNC, 0);
```
mogasergiu: `checkpatch.pl --strict` complains:
```
CHECK: spaces preferred around that '|' (ctx:VxV)
```

lib/vfscore/automount.c
```diff
@@ -79,17 +80,26 @@ struct vfscore_volume {
...
+	 * extracting an archive to an existing filesystem. If this option
+	 * is chosen, no other mount option is accepted.
+	 */
+	if (!opts || (strcmp(opts, LIBVFSCORE_INITRD_OPT_EXTRACT) != 0)) {
+		/* mount a ramfs volume as base only in non-'extract' mode */
+		uk_pr_info("Mounting 'ramfs' to %s...\n", path);
+		rc = mount("", path, "ramfs", flags, opts);
+		if (unlikely(rc)) {
+			uk_pr_crit("Failed to mount ramfs to \"%s\": %d\n",
+					path, errno);
```
mogasergiu: Misaligned to open parantheses.

lib/vfscore/automount.c
```diff
@@ -412,6 +422,28 @@ static int vfscore_automount_fstab_volumes(void)
...
 		vfscore_fstab_fetch_volume_args(vfscore_fstab[i], &vv);
 
+#if CONFIG_LIBVFSCORE_INITRD_EXTRACT_WORKAROUND && \
+    !CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS
+		/* WORKAROUND: In case the image was built without predefined
+		 *             rootfs and we are asked to `extract` an initrd
+		 *             to `/` as first filesystem, we mount an empty
+		 *             ramfs as base.
+		 */
+		if ((i == 0) &&
```
mogasergiu: Why so many parantheses:
- `(i == 0)`
- ` (strcmp("initrd", vv.drv) == 0)`
- `(strcmp(LIBVFSCORE_INITRD_OPT_EXTRACT, vv.opts) == 0)`
- `(vv.path[0] == '/') && (vv.path[1] == '\0')`

lib/ukcpio/cpio.c
```diff
@@ -141,6 +142,7 @@ int uk_syscall_r_chmod(const char *, mode_t);
 int uk_syscall_r_utime(const char *, const struct utimbuf *);
 int uk_syscall_r_mkdir(const char *, mode_t);
 int uk_syscall_r_symlink(const char *, const char *);
+int uk_syscall_r_stat(const char *, struct stat *stat);
```
mogasergiu: Probably should remove the last `stat` to align with the other declarations?

mogasergiu (changes requested):
Hey @skuenzer :wave: ! Thanks for your work! A couple of comments from my side.

razvand (approved):

lib/vfscore/automount.c
skuenzer: I thought about doing this with ukopt but it did overall make less sense to me. The thing is that the initrd fs is already not real (based on ramfs) and then you move an option which is only specific to it (extract) to ukopt feels like distributing the options all over the place. The thing is that extract is only specific to initrd whereas mkmp is in general applicable to any filesystem. If we keep the generality of ukopt (which I thought is the idea), the implementation of any future ukopt is simpler and we do not bloat the code with condition handling.

lib/ukcpio/cpio.c
razvand: As the other checks don't use unlikely, I would aim for consistency first and not add it.

lib/vfscore/automount.c
razvand: I added an issue for that, in order to keep this PR simple: #1229

lib/ukcpio/cpio.c
mogasergiu: That does not mean that it is good. But ok fine. We should at some point have a look at most source files and add `unlikely`'s on error cases where it makes sense (e.g. it makes sense not to use `unlikely` if `newfstatat` is run on an inexistent file, as it is common for apps to test whether a file exists or not through such syscalls but it makes sense to add `unlikely`'s on initrd extraction since it is done by us and if a function fails here then there is not much point in continuing booting).

mogasergiu (approved):

mogasergiu (approved):
Thank you for your work @razvand @skuenzer :christmas_tree: :santa: 

lib/ukcpio/cpio.c
razvand: @mogasergiu , fully agree this should be added. My concern is that adding it **now, here** will cause confusion. Why here and not elsewhere?

We would need to be consistent and add it to an entire file / compilation unit.

To keep track of this, I created issue #1233 for this.

PR #1203, opened 2023-12-12
lib/posix-{fdio,socket}: Add stat support to socket files

andreittr:
### Description of changes

This changeset adds support for `stat`ing socket files, along with more general improvements to `stat` compatibility & performance. Specifically:
- Have `fstat` zero out the output buffer before writing in values -- an undocumented behavior present in Linux that (bincompat) applications may depend on; we should replicate this for best compatibility
- Remove synchronization from `stat`-like functions -- `stat(2)` documentation specifically permits stat calls to not be synchronized with other syscalls that change file attributes (e.g. `(f)chmod` or I/O); thus we can safely remove locking for better performance
- Actual support for `stat` on socket files -- these are the same for all sockets, as there is no relevant data to be supplied by the driver.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Basic snippet to test against Linux (no struct stat pretty-printing, check `stat(3type)` for the struct layout and eyeball it):
```c
#include <stdio.h>
#include <assert.h>
#include <stdint.h>
#include <string.h>

#include <sys/stat.h>
#include <sys/socket.h>


int main(void)
{
	int s = socket(AF_UNIX, SOCK_STREAM, 0);
	assert(s >= 0);

	struct stat sb;
	memset(&sb, -1, sizeof(sb));
	int r = fstat(s, &sb);
	assert(!r);

	printf("%zu\n\n", sizeof(sb));
	for (int i = 0; i < sizeof(sb)/4; i++)
		printf("%x\n", ((uint32_t *)&sb)[i]);

	return 0;
}
```
(needs `CONFIG_LIBPOSX_UNIXSOCKET`, but should work similarly with any sockets)

### DIFF
--- a/lib/posix-socket/socket.c
+++ b/lib/posix-socket/socket.c
@@ -39,6 +39,8 @@
 #define _GNU_SOURCE
 
 #include <sys/types.h>
+#include <sys/stat.h>
+
 #include <uk/socket.h>
 #include <uk/file.h>
 #include <uk/file/nops.h>
@@ -163,10 +165,33 @@ socket_ctl(const struct uk_file *sock, int fam, int req,
 	}
 }
 
+#define SOCKET_STATX_MASK \
+	(UK_STATX_TYPE|UK_STATX_MODE|UK_STATX_NLINK|UK_STATX_INO|UK_STATX_SIZE)
+
+static int
+socket_getstat(const struct uk_file *sock,
+	       unsigned int mask __unused, struct uk_statx *arg)
+{
+	/* Since all information is immediately available, ignore mask arg */
+	arg->stx_mask = SOCKET_STATX_MASK;
+	arg->stx_mode = S_IFSOCK|0777;
+	arg->stx_nlink = 1;
+	arg->stx_ino = (uintptr_t)sock; /* Must be unique per-socket */
+	arg->stx_size = 0;
+
+	/* Following fields are always filled in, not in stx_mask */
+	arg->stx_dev_major = 0;
+	arg->stx_dev_minor = 8; /* Same value Linux returns for sockets */
+	arg->stx_rdev_major = 0;
+	arg->stx_rdev_minor = 0;
+	arg->stx_blksize = 0x1000;
+	return 0;
+}
+
 static const struct uk_file_ops socket_file_ops = {
 	.read = socket_read,
 	.write = socket_write,
-	.getstat = uk_file_nop_getstat,
+	.getstat = socket_getstat,
 	.setstat = uk_file_nop_setstat,
 	.ctl = socket_ctl
 };
--- a/lib/posix-fdio/fdstat.c
+++ b/lib/posix-fdio/fdstat.c
@@ -6,6 +6,8 @@
 
 /* Internal syscalls for manipulating file metadata */
 
+#include <string.h>
+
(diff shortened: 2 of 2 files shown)


lib/posix-socket/socket.c
```diff
@@ -163,10 +165,33 @@ socket_ctl(const struct uk_file *sock, int fam, int req,
...
 
+#define SOCKET_STATX_MASK \
+	(UK_STATX_TYPE|UK_STATX_MODE|UK_STATX_NLINK|UK_STATX_INO|UK_STATX_SIZE)
+
+static int
+socket_getstat(const struct uk_file *sock,
+	       unsigned int mask __unused, struct uk_statx *arg)
+{
+	/* Since all information is immediately available, ignore mask arg */
+	arg->stx_mask = SOCKET_STATX_MASK;
```
skuenzer, line 176: What do you think of initializing `arg` with a struct copy instead?
```c
*arg = { .stx_mask = SOCKET_STATX_MASK,
         .stx_mode = S_IFSOCK|0777, 
         .stx_nlink = 1, 
         /* .... */
       };
```
My argument would be that this would make the code future proof, any later added field to `struct uk_statx` would be automatically zero'ed out.

andreittr: I had considered using designated initializers, however, since `struct statx` is pretty big (256 bytes at the time of writing) I'd rather explicitly _not_ have code fill in the entire thing if not needed, something explicitly allowed by the documentation (and is precisely what the mask arg/field is for).
Then again, Linux seems to zero out the output buffer regardless, an undocumented behavior apps may rely on, much like `stat`; plus stat(x) is usually/hopefully not on the hotpath, so it'd make sense to be conservative here and zero out everything.

Question is: how do we ensure this for any and all stat calls, to maintain consistency? Designated initializers in the driver ensure zeroing out _only_ for that driver, but adding explicit `memset(0)` to `posix-fdio` may mean redundant zeroing out. We either
- (1) have `getstat` to use designated initializers in all drivers, enforced via code review (+ optional reminder comment in code) -- depends on human action, somewhat brittle, but should be fine once implemented and reviewers remember what's what
- (2) enforce zeroing out in `posix-fdio` -> drivers explicitly should _not_ use designated initializers to prevent doing double work -- more robust, worst case you lose performance but maintain compat

One other small caveat re: designated initializers: if we ever want this code to pass through a C++ compiler (super unlikely, since it's not a header and we're not crazy) we need to make sure the field order is _exactly_ the same as in the struct declaration (which atm it is not). C++ is dumb in that way.

skuenzer: Hey, sorry, you're right, I somehow missed the mask and I agree with your chosen approach. Thinking about it again, I also agree that - as kernel - we should zero out the stat buffer for the `SYS_fstat` case (as you also patched in this PR), because with `fstat` you do not have a way to communicate which fields as application you are interested in and the kernel cannot indicate which components are supported by the subsystem. `statx` is of course a different consideration and it looks that we are actually close providing that system call, too. But overall, everything is fine for this PR. 👍🏻

skuenzer (approved):
Thanks, I will take the PR in.

PR #1205, opened 2023-12-13
Add `lib/posix-fdtab` and `lib/vfscore` destructors

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Add `lib/posix-fdtab` destructor:
When all strong references to an open file structure are gone, `fdrop`
calls `vfs_close` which invokes the `close` method of the underlying
filesystem driver. The latter may not necessarily flush cached open
file contents back to the storage device.

Thus, fix this by also invoking the `sync` method of the aforementioned
filesystem driver right before invoking the `close` method.

Add `lib/vfscore/automount` destructor:
Since the constructor of `vfscore` simply mounts the `rootfs` and any
additional volumes in `vfs.fstab`, the destructor shall iterate in
reverse upon these mount points and perform an unmount operation on
them.

Furthermore, and most importantly, a special adjustment must be taken
into consideration w.r.t. the registration of this destructor: priority
of termination functions. Since the vfscore destructor unmounts active
volumes, it makes sense to be preceded by the termination of
`lib/posix-fdtab` which syncs and closes remaining open file structures.
However, right now the aforemention library has the highest priority and
would thus end up being executed after the unmounting termination method
introduced in this commit.

Therefore, fix this by declaring an explicit `lib/posix-fdtab` priority
of `1`, enough to make room for one more preceding priority to be taken
by the unmounting method so that its termination function succeeds that
of `lib/posix-fdtab`. Additionally, sync the other libraries that depend
on `lib/posix-fdtab` to the newly introduced priority value.

### DIFF
--- a/lib/vfscore/automount.c
+++ b/lib/vfscore/automount.c
@@ -49,6 +49,8 @@
 #include <uk/libparam.h>
 #include <uk/plat/memory.h>
 #include <sys/stat.h>
+#include <vfscore/mount.h>
+#include <uk/posix-fdtab.h>
 
 #ifdef CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS
 #include <errno.h>
@@ -441,4 +443,21 @@ static int vfscore_automount(struct uk_init_ctx *ictx __unused)
 	return vfscore_automount_fstab_volumes();
 }
 
-uk_rootfs_initcall_prio(vfscore_automount, 0x0, 4);
+extern struct uk_list_head mount_list;
+
+static void vfscore_autoumount(const struct uk_term_ctx *tctx __unused)
+{
+	struct mount *mp;
+	int rc;
+
+	uk_list_for_each_entry_reverse(mp, &mount_list, mnt_list) {
+		/* For now, flags = 0 is enough. */
+		rc = VFS_UNMOUNT(mp, 0);
+		if (unlikely(rc))
+			uk_pr_err("Failed to unmount %s: error %d.\n",
+				  mp->m_path, rc);
+	}
+}
+
+uk_rootfs_initcall_prio(vfscore_automount, vfscore_autoumount,
+			UK_PRIO_BEFORE(UK_LIBPOSIX_FDTAB_PRIO));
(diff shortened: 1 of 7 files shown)


lib/vfscore/automount.c
```diff
@@ -441,4 +445,26 @@ static int vfscore_automount(struct uk_init_ctx *ictx __unused)
...
+		if (unlikely(rc))
+			uk_pr_err("Failed to unmount %s: error %d.\n",
+				  mp->m_path, rc);
+	}
+}
+
+uk_rootfs_initcall_prio(vfscore_automount, vfscore_autoumount,
+#if CONFIG_LIBPOSIX_FDTAB
+			UK_PRIO_BEFORE(UK_LIBPOSIX_FDTAB_PRIO)
+#else /* !CONFIG_LIBPOSIX_FDTAB */
```
skuenzer: Because we cannot have `vfscore` without `posix-fdtab`, the `#if` condition is not required.

lib/vfscore/automount.c
```diff
@@ -49,6 +49,10 @@
 #include <uk/libparam.h>
 #include <uk/plat/memory.h>
 #include <sys/stat.h>
+#include <vfscore/mount.h>
+#if CONFIG_LIBPOSIX_FDTAB
```
skuenzer: `posix-fdtab` is a dependency for `vfscore`, the header should be available.

lib/vfscore/fops.c
```diff
@@ -45,6 +45,13 @@ int vfs_close(struct vfscore_file *fp)
 	int error;
 
 	vn_lock(vp);
+	/* This is called when there are no more strong references to this file
```
skuenzer, line 48: Maybe put also a `FIXME` comment. As we discussed, the sync operation should actually happen on the fs driver during unmount. This is a workaround for now.

mogasergiu: This should actually be fine and make sense. If there are no more references to an open file structure, it makes sense to sync, right? CC @andreittr

skuenzer: I am referring to this: https://man7.org/linux/man-pages/man2/close.2.html (see notes)

skuenzer (approved):
Thanks a lot!

Issue #1150, opened 2023-10-29
x86 SMP startup


tkf2019:
### Describe the bug

I tried to enable SMP and start up APs according to this [example](https://unikraft.org/blog/2022-08-20-unikraft-releases-phoebe#testing) on the latest release, 0.15.0, but the kernel seemed to crash and failed to read from stack after [lcpu_start.S](https://github.com/unikraft/unikraft/blob/staging/plat/kvm/x86/lcpu_start.S#L248)

### Steps to reproduce

Run the test [code](https://unikraft.org/blog/2022-08-20-unikraft-releases-phoebe#testing) on KVM and QEMU 8.1.2.

### Expected behavior

I expected the unikernel to boot and print "Hello World!".

### Which architectures were you using or does this bug affect?

x86_64

### Which platforms were you using or does this bug affect?

kvm

PR #1211, opened 2023-12-15
lib/uknetdev: New `netdev.ip` kernel parameter

skuenzer:
### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

 - `CONFIG_LIBUKNETDEV=y`
 - `CONFIG_LIBUKNETDEV_EINFO_LIBPARAM=y`

### Description of changes

This commit reintroduces the ability to override netdev einfo fields via kernel command line parameters. The `netdev.ip` parameter can be used to override IPv4 address information for multiple devices. For each device the following colon-separated format is introduced:

  `cidr[:gw[:dns0[:dns1[:hostname[:domain]]]]]`

Any of the fields can be left blank and fewer fields can be passed by omitting fields from the right. This also ensures that more fields can be appended in the future while maintaining backwards compatibility.

Example 1:
 - IPv4 address: `192.168.0.123`
 - IPv4 mask: `255.255.255.0`
 - IPv4 default gateway: `192.168.0.1`
 - Primary DNS server (IPv4): `192.168.0.1`
...converts to:
 `netdev.ip=192.168.0.123/24:192.168.0.1:192.168.0.1`

Example 2:
 Additionally to example 1, we add the following fields:
 - We do not hand-over a secondary DNS server
 - Hostname: `example`
 - Domain suffix: `unikraft.org`
...this converts to:
 `netdev.ip=192.168.0.123/24:192.168.0.1:192.168.0.1::example:unikraft.org`

> [!CAUTION]
> This is a breaking change and this PR should be checked in conjunction with [unikraft/kraftkit#1150][0] and merged simultaneously with [#1150][0] to avoid incompatibility issues. Simultaneously, all [catalog](https://github.com/unikraft/catalog) entries should be retriggered against Unikraft's `staging` branch following the merge. Since Unikraft is pre-v1.0, no backwards compatibility is included in this refactor.

[0]: https://github.com/unikraft/kraftkit/pull/1150

### DIFF
--- a/lib/uknetdev/Config.uk
+++ b/lib/uknetdev/Config.uk
@@ -31,6 +31,24 @@ config LIBUKNETDEV_DISPATCHERTHREADS
 		allocated for each configured receive queue.
 		libuksched is required for this option.
 
+config LIBUKNETDEV_EINFO_LIBPARAM
+	bool "Netdev einfo with kernel parameters"
+	select LIBUKLIBPARAM
+	help
+		Extended information field are used to pass IPv4 addresses
+		and network settings to network stacks. This option enables to
+		overwrite the values that are provided by underlying drivers
+		with the `netdev.ip` kernel parameter.
+
+config LIBUKNETDEV_EINFO_LIBPARAM_MAXCOUNT
+	int "Maximum number of configurable netdevs"
+	depends on LIBUKNETDEV_EINFO_LIBPARAM
+	range 1 512
+	default 4
+	help
+		This number defines an upper limit of netdev devices that can
+		be configured via the kernel command line.
+
 config LIBUKNETDEV_STATS
 	bool "Collect network statistics"
 	default n
--- a/lib/uknetdev/netdev.c
+++ b/lib/uknetdev/netdev.c
@@ -46,72 +46,7 @@
 struct uk_netdev_list uk_netdev_list =
 	UK_TAILQ_HEAD_INITIALIZER(uk_netdev_list);
 static uint16_t netdev_count;
-/**
- * TODO: Define Network argument format when multiple driver device need to
- * coexist. For example like:
- * Driver Name:IP Address:Net Mask
- */
-static char *ipv4_addr;
-static char *ipv4_subnet_mask;
-static char *ipv4_gw_addr;
-static char *ipv4_dns0_addr;
 
-UK_LIB_PARAM_STR(ipv4_addr);
-UK_LIB_PARAM_STR(ipv4_subnet_mask);
-UK_LIB_PARAM_STR(ipv4_gw_addr);
-UK_LIB_PARAM_STR(ipv4_dns0_addr);
-
-static const char *_parse_ipv4_addr(void)
-{
-	/** Remember the reference to the ip address for successive calls*/
-	static char *ip_addr;
-
-	if (ip_addr)
-		return strtok_r(NULL, " ", &ip_addr);
-	else if (ipv4_addr)
-		return strtok_r(ipv4_addr, " ", &ip_addr);
-
-	return NULL;
-}
-
-static const char *_parse_ipv4_net_mask(void)
-{
-	/** Remember the reference to the netmask for successive calls*/
-	static char *net_mask;
-
-	if (net_mask)
-		return strtok_r(NULL, " ", &net_mask);
-	else if (ipv4_subnet_mask)
-		return strtok_r(ipv4_subnet_mask, " ", &net_mask);
-
-	return NULL;
-}
-
-static const char *_parse_ipv4_gw_addr(void)
-{
-	/** Remember the reference to the gateway address for successive calls*/
-	static char *gw;
-
-	if (gw)
-		return strtok_r(NULL, " ", &gw);
-	else if (ipv4_gw_addr)
-		return strtok_r(ipv4_gw_addr, " ", &gw);
-
-	return NULL;
-}
-
-static const char *_parse_ipv4_dns0_addr(void)
-{
-	/** Remember the reference to the dns server for successive calls */
-	static char *dns0;
-
-	if (dns0)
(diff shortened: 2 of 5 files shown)


nderjung:
Will this syntax work for IPv6? The colon delimiter I think would be best replaced with a comma to support this future usecase.

skuenzer:
> Will this syntax work for IPv6? The colon delimiter I think would be best replaced with a comma to support this future usecase.

I chose the colon separator to be close to what people know from Linux: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ip_networking_from_the_kernel_command_line
It is simply a familiarity reason and the `vfs.fstab` parameter of `lib/vfscore` uses the same separator already. According to what I found is that Fedora Linux accepts also IPv6 addresses with the same `ip` parameter. Here, an IPv6 address is encapsulated into square brackets. In principle, this should work for us, too. For example:
`netdev.ip6=[ "[2001:db8::b0b0]/64:[2001:db8::f00d]" ]`
Might be even more readable to have an IPv6 address within brackets 🤔

lib/uknetdev/Config.uk
```diff
@@ -31,6 +31,24 @@ config LIBUKNETDEV_DISPATCHERTHREADS
 		allocated for each configured receive queue.
 		libuksched is required for this option.
 
+config LIBUKNETDEV_EINFO_LIBPARAM
```
StefanJum, line 34: Should we add a `default y` here, since currently we run all applications using kernel params, and this will need to be  enabled for that to work?

lib/uknetdev/netdev.c
```diff
@@ -274,44 +301,74 @@ void uk_netdev_info_get(struct uk_netdev *dev,
...
+		case UK_NETDEV_IPV4_DOMAIN:
+			if (dev->_einfo->ip4.domain)
+				return dev->_einfo->ip4.domain;
+			break;
+		default:
+		}
+	}
+#endif /* CONFIG_LIBUKNETDEV_EINFO_LIBPARAM */
+
+	if (!ret && dev->ops->einfo_get) {
```
StefanJum: Isn't `ret` always `NULL` here?

StefanJum (commented):
Works fine @skuenzer, just few minor comments

nderjung:
A PR has been introduced to KraftKit which incorporates this change and these two PRs should be checked in conjunction with one another: https://github.com/unikraft/kraftkit/pull/1150

lib/uknetdev/netdev.c
skuenzer: Actually, true. Probably a left-over. I'll remove it.

lib/uknetdev/Config.uk
skuenzer: No, because Unikraft options should build a minimal kernel as default with all non-required features off. The idea is that application configurations (application's `Config.uk` or `Kraftfile`) should take care of enabling needed options. For example, we could put an `imply` option to `app-elfloader`. We should not additionally need to take extra care to switch an option off that is unwanted.

lib/uknetdev/netdev.c
skuenzer: Hey, in the new version I got rid of `ret` variable completely. Please, double-check. I also noticed that `NULL` would be always returned in case of `UK_NETDEV_IPV4_ADDR` and `UK_NETDEV_IPV4_MASK`. However, there should be a value returned by `dev->ops->einfo_get()` if there is no CIDR address with the device's einfo or the user provided kernel parameter.

lib/uknetdev/netdev.c
```diff
@@ -274,43 +301,71 @@ void uk_netdev_info_get(struct uk_netdev *dev,
...
+			break;
+		case UK_NETDEV_IPV4_HOSTNAME:
+			if (dev->_einfo->ip4.hostname)
+				return dev->_einfo->ip4.hostname;
+			break;
+		case UK_NETDEV_IPV4_DOMAIN:
+			if (dev->_einfo->ip4.domain)
+				return dev->_einfo->ip4.domain;
+			break;
+		default:
```
StefanJum, line 350:
```suggestion
		default:
		    break;
```

This will error on compile time.

lib/uknetdev/netdev.c
```diff
@@ -274,43 +301,71 @@ void uk_netdev_info_get(struct uk_netdev *dev,
...
+		switch (einfo) {
+		case UK_NETDEV_IPV4_ADDR:
+			if (dev->ops->einfo_get(dev, UK_NETDEV_IPV4_CIDR))
+				return NULL; /* IPv4 CIDR exists */
+			break;
+		case UK_NETDEV_IPV4_MASK:
+			if (dev->ops->einfo_get(dev, UK_NETDEV_IPV4_CIDR))
+				return NULL; /* IPv4 CIDR exists */
+			break;
+		default:
```
StefanJum, line 366:
```suggestion
		default:
		    break;
```
Same as above.

StefanJum (changes requested):
See the changes @skuenzer, works fine besides that.

lib/uknetdev/netdev.c
skuenzer: What is the error that you are receiving? I am unable to reproduce an error. It compiles fine for me with GCC 13.2.0.

skuenzer: As we just discussed and confirmed, in fact, older GCC versions handle this differently and do not accept an empty `default` case. I'll push an update. Thanks for reporting this. 👍🏻

lib/uknetdev/netdev.c
skuenzer: 👍🏻

StefanJum (approved):
All good now, thanks.

razvand (approved):

PR #1212, opened 2023-12-17
uk/plat/memory: Introduce `pg_off` and `pg_count` memregion fields

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

To make memory region management easier w.r.t. alignment handling,
define two additional fields for `struct ukplat_memregion_desc`:
- `pg_off` to represent the in-page offset from where the actual
resource this memory region is dedicated to starts
- `pg_count` to represent the length of the entire, end-to-end
page-aligned, memory region in number of pages

Thus, the definition of some other fields shall then change:
- `pbase` will be the physical page-aligned base address of the
region. This means that in order to get the actual address of a
resource, one may have to make the following basic addition:
        `pbase` + `pg_off`
- `vbase` same as `pbase` but for virtual base address
- `len` will now represent the length of the resource inside the
region, not the length of the region.

E.g.
For a resource with address `0x1050` and length `0x430` the
corresponding memory region descriptor will have the following
values:
- `pbase` and `vbase` equal to `0x1000` (`PAGE_ALIGN_DOWN(0x1050)`)
- `pg_off` equal to `0x50` (`0x1050 & ~PAGE_MASK`)
- `pg_count` equal to `5` (`PAGE_COUNT(0x1050 + 0x430)`)
- `len` equal to `0x430`

The other fields (`type`, `flags`, `name`) will keep their meaning.

### DIFF
--- a/include/uk/plat/memory.h
+++ b/include/uk/plat/memory.h
@@ -81,12 +81,16 @@ extern "C" {
  * Descriptor of a memory region
  */
 struct ukplat_memregion_desc {
-	/** Physical base address */
+	/** Physical page-aligned base address of the region */
 	__paddr_t pbase;
-	/** Virtual base address */
+	/** Virtual page-aligned base address of the region */
 	__vaddr_t vbase;
-	/** Length in bytes */
+	/** Offset where the resource starts in the region's first page */
+	__off pg_off;
+	/** Length in bytes of the resource inside this region */
 	__sz len;
+	/** Number of pages the end-to-end aligned region occupies */
+	__sz pg_count;
 	/** Memory region type (see UKPLAT_MEMRT_*) */
 	__u16 type;
 	/** Memory region flags (see UKPLAT_MEMRF_*) */
--- a/include/uk/plat/memory.h
+++ b/include/uk/plat/memory.h
@@ -37,6 +37,9 @@
 #include <uk/essentials.h>
 #include <uk/arch/types.h>
 #include <uk/alloc.h>
+#include <uk/arch/paging.h>
+#include <uk/arch/types.h>
+#include <uk/essentials.h>
 #include <uk/config.h>
 #include <uk/arch/ctx.h>
 #include <uk/arch/paging.h>
@@ -100,6 +103,124 @@ struct ukplat_memregion_desc {
 	char name[36];
 #endif /* CONFIG_UKPLAT_MEMRNAME */
 } __packed __align(__SIZEOF_LONG__);
+
+/** UK_ASSERT_VALID_MRD_TYPE(mrd) macro
+ *
+ * Ensure a given memory region descriptor has one of the following defined
+ * types only:
+ *	UKPLAT_MEMRT_FREE		Uninitialized memory
+ *	UKPLAT_MEMRT_RESERVED		In use by platform
+ *	UKPLAT_MEMRT_KERNEL		Kernel binary segment
+ *	UKPLAT_MEMRT_INITRD		Initramdisk
+ *	UKPLAT_MEMRT_CMDLINE		Command line
+ *	UKPLAT_MEMRT_DEVICETREE		Device tree
+ *	UKPLAT_MEMRT_STACK		Thread stack
+ * @param mrd pointer to the memory region descriptor whose type to validate
+ */
+#define UK_ASSERT_VALID_MRD_TYPE(mrd)					\
+	do {								\
+		switch ((mrd)->type) {					\
+		case UKPLAT_MEMRT_FREE:					\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_RESERVED:				\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_KERNEL:				\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_INITRD:				\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_CMDLINE:				\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_DEVICETREE:				\
+			__fallthrough;					\
+		case UKPLAT_MEMRT_STACK:				\
+			break;						\
+		default:						\
+			UK_CRASH("Invalid mrd type: %hu\n",		\
+				 (mrd)->type);				\
+		}							\
+	} while (0)
+
+/** UK_ASSERT_VALID_MRD_FLAGS(mrd) macro
+ *
+ * Ensure a given memory region descriptor has one of the following defined
+ * flags only:
+ *	UKPLAT_MEMRF_READ		Region is readable
+ *	UKPLAT_MEMRF_WRITE		Region is writable
+ *	UKPLAT_MEMRF_EXECUTE		Region is executable
+ *	UKPLAT_MEMRF_UNMAP		Must be unmapped at boot
+ *	UKPLAT_MEMRF_MAP		Must be mapped at boot
+ *
+ * @param mrd pointer to the memory region descriptor whose type to validate
+ */
+#define UK_ASSERT_VALID_MRD_FLAGS(mrd)					\
+	do {								\
+		__u16 flags_all = UKPLAT_MEMRF_READ    |		\
+				  UKPLAT_MEMRF_WRITE   |		\
+				  UKPLAT_MEMRF_EXECUTE |		\
+				  UKPLAT_MEMRF_UNMAP   |		\
+				  UKPLAT_MEMRF_MAP;			\
+									\
+		UK_ASSERT(((mrd)->flags & flags_all) == (mrd)->flags);	\
+	} while (0)
+
+/** UK_ASSERT_VALID_MRD(mrd) macro
+ *
+ * Ensure memory region descriptor general correctness:
+ * - must be of only one valid type as per UK_ASSERT_VALID_MRD_TYPE
+ * - must only have valid flags as per UK_ASSERT_VALID_MRD_FLAGS
+ * - memory region is not empty or of length 0
+ * - virtual/physical base addresses are page-aligned
+ * - resource in-page offset must be in the range [0, PAGE_SIZE)
+ *
+ * @param mrd pointer to the free memory region descriptor to validate
+ */
+#define UK_ASSERT_VALID_MRD(mrd)					\
+	do {								\
+		UK_ASSERT_VALID_MRD_TYPE((mrd));			\
+		UK_ASSERT_VALID_MRD_FLAGS((mrd));			\
+		UK_ASSERT(PAGE_ALIGNED((mrd)->vbase));			\
+		UK_ASSERT(PAGE_ALIGNED((mrd)->pbase));			\
+		UK_ASSERT((mrd)->pg_off >= 0 &&				\
+			  (mrd)->pg_off < (__off)PAGE_SIZE);		\
+	} while (0)
+
+/** UK_ASSERT_VALID_FREE_MRD(mrd) macro
+ *
+ * Ensure free memory region descriptor particular correctness:
+ * - must meet the criteria of a general valid memory region descriptor
+ * - virtual/physical base addresses are equal
+ * - region is aligned end-to-end, therefore length is multiple of
+ * PAGE_SIZE times region's page count and the resource's
+ * in-page offset must be 0
+ *
+ * @param mrd pointer to the free memory region descriptor to validate
+ */
+#define UK_ASSERT_VALID_FREE_MRD(mrd)					\
+	do {								\
+		UK_ASSERT_VALID_MRD((mrd));				\
+		UK_ASSERT((mrd)->type == UKPLAT_MEMRT_FREE);		\
+		UK_ASSERT((mrd)->vbase == (mrd)->pbase);		\
+		UK_ASSERT((mrd)->pg_count * PAGE_SIZE == (mrd)->len);	\
+		UK_ASSERT(!(mrd)->pg_off);				\
+	} while (0)
+
+/** UK_ASSERT_VALID_KERNEL_MRD(mrd) macro
+ *
+ * Ensure kernel memory region descriptor particular correctness:
+ * - must meet the criteria of a general valid memory region descriptor
+ * - region is aligned end-to-end, therefore length is multiple of
+ * PAGE_SIZE times region's page count and the resource's
+ * in-page offset must be 0
+ *
+ * @param mrd pointer to the kernel memory region descriptor to validate
+ */
+#define UK_ASSERT_VALID_KERNEL_MRD(mrd)					\
+	do {								\
+		UK_ASSERT_VALID_MRD((mrd));				\
+		UK_ASSERT((mrd)->type == UKPLAT_MEMRT_KERNEL);		\
+		UK_ASSERT((mrd)->pg_count * PAGE_SIZE == (mrd)->len);	\
+		UK_ASSERT(!(mrd)->pg_off);				\
+	} while (0)
 
 /**
  * Check whether the memory region descriptor overlaps with [pstart, pend) in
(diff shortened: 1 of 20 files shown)


mogasergiu:
This proposed improvement has been discussed during the Arm CCA GSoC track to make debugging memory region setup on newer, not yet supported, platforms easier.

plat/xen/x86/setup.c
```diff
@@ -173,14 +177,14 @@ static int _init_mem(struct ukplat_bootinfo *const bi)
...
+		if (HYPERVISOR_start_info->flags & SIF_MOD_START_PFN)
 			mrd.vbase = (__vaddr_t)to_virt(mrd.pbase);
-		} else {
-			mrd.pbase = HYPERVISOR_start_info->mod_start;
+		else
 			mrd.vbase = mrd.pbase;
-		}
+		mrd.pg_off = HYPERVISOR_start_info->mod_start - mrd.pbase;
 		mrd.len = (size_t)HYPERVISOR_start_info->mod_len;
+		mrd.pg_count = PAGE_COUNT(mrd.len);
```
michpappas: Shouldn't this take into account the offset?

plat/common/memory.c
```diff
@@ -194,35 +199,54 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
 		/* If the right region is contained within the left region,
 		 * drop it entirely
 		 */
-		if (RANGE_CONTAIN(ml->pbase, ml->len, mr->pbase, mr->len)) {
+		/* This can only happen if mr is a free mrd or if it has
+		 * the same type as ml
+		 */
+		UK_ASSERT(mr->type == UKPLAT_MEMRT_FREE ||
+			  mr->type == mr->type);
```
michpappas: Must be `mr->type == ml->type`, but also in the caller there is an assertion that requires at least one of the regions has to be `UKPLAT_MEMRT_FREE`. That guarantees that if `mr->type` is not `UKPLAT_MEMRT_FREE`, `ml->type` has to be `UKPLAT_MEMRT_FREE`, so `mr->type == ml->type` always evaluates to true which makes it redundant. Of course the function must make no assumptions about the caller, but it may be sensible to collect all assertions in one place which is usually the callee. In other words the assertion on the caller seems to also be redundant if we get rid of `mr->type == ml->type`.

plat/common/memory.c
```diff
@@ -232,57 +256,23 @@ static inline void overlapping_mrd_fixup(struct ukplat_memregion_list *list,
...
+						.pg_off = 0,
+						.len = len,
+						.pg_count = PAGE_COUNT(len),
+						.type = UKPLAT_MEMRT_FREE,
 						.flags = ml->flags
 					}, ridx + 1);
 
 			/* Drop the fraction of ml that overlaps with mr */
-			ml->len = mr->pbase - ml->pbase;
+			ml->len = mr->pbase + mr->pg_off -
```
michpappas: Something looks off here. For example if `mr->pg_off` is zero we trim too much off `ml`. Also if `mr->pg_off` >0, `ml->len` cannot exceed it's last page anyway so it can't be part of `mr`.

plat/common/memory.c
```diff
@@ -374,41 +356,54 @@ void ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
 
 			/* If they have the same priority, merge them. If they
 			 * are contained within each other, drop the contained
-			 * one.
+			 * one. Do not allow merging of kernel resources, as
+			 * their resource page offset into the region is
+			 * important!
 			 */
-			} else {
+			} else if (ml_prio != MRD_PRIO_KRNL_RSRC &&
```
michpappas: Generally speaking this condition should not only apply to kernel regions but any region with a non-zero offset. For `MRD_PRIO_FREE` and `MRD_PRIO_RESVD` that should never be the case, and we catch that with assertions below. If, however, there is an overlap in a kernel regions, this condition lets it silently slip through. Maybe it's better to leave the `else` as it was and add an assertion that checks that neither region has kernel priority.

plat/common/memory.c
```diff
@@ -422,60 +417,42 @@ void ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
-				del = 1;
 			}
 
 		/* If they do not overlap but they are contiguous and have the
-		 * same flags and priority.
+		 * same flags and priority. Do not merge Kernel type memregions,
+		 * as we have to preserve pg_off's and len's.
 		 */
-		} else if (ml->pbase + ml->len == mr->pbase &&
+		} else if (ml_prio != MRD_PRIO_KRNL_RSRC &&
```
michpappas: For readability I would place the main condition (i.e. whether contiguous) first and move the kernel priority check next to `ml_prio == mr_prio`

plat/common/memory.c
```diff
@@ -374,41 +356,54 @@ void ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
 				 */
 				UK_ASSERT(ml->flags == mr->flags);
 
+				/* We do not allow overlaps of memory regions
+				 * whose resource page offset into their region
+				 * is not equal to 0. Regions don't that meet
+				 * this condition are hand-inserted by us and
+				 * should not overlap.
+				 */
+				UK_ASSERT(!ml->pg_off);
```
michpappas, line 386: At first I thought of some cases where these assertions can lead into false positives, along with some cases that may require offset adjustments, but after all we should hopefully expect kernel, reserved, or free regions to be page aligned. Until we come with an unusual case we can probably leave them as is.

plat/common/memory.c
```diff
@@ -374,41 +356,54 @@ void ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
...
+							 PAGE_SIZE)) {
 					uk_pr_debug("Deleting right mrd!\n");
 					ukplat_memregion_list_delete(list,
 								     i + 1);
-					del = 1;
-
-					goto restore_mrds;
 				}
 
 				uk_pr_debug("Merging two overlapping mrd's.\n");
```
michpappas: Not part of this PR, but with the opportunity: `s/mrd's/mrds/` (plural doesn't take apostrophe).

plat/kvm/x86/lxboot.c
```diff
@@ -110,14 +114,20 @@ lxboot_init_mem(struct ukplat_bootinfo *bi, struct lxboot_params *bp)
...
+		mrd.pg_count = PAGE_COUNT(mrd.pg_off + mrd.len);
 
 		if (entry->type == LXBOOT_E820_TYPE_RAM) {
 			mrd.type = UKPLAT_MEMRT_FREE;
 			mrd.flags = UKPLAT_MEMRF_READ | UKPLAT_MEMRF_WRITE;
 
+			/* Free memory regions have
+			 * mrd.len == mrd.pg_count * PAGE_SIZE
+			 */
+			mrd.len = PAGE_ALIGN_UP(mrd.len);
```
michpappas, line 130: I'm not sure I follow this. Does it also imply a zero offset? If not, should `len` be set to `pg_count`? The same question applies to the relevant change in multiboot.

plat/common/paging.c
```diff
@@ -1506,22 +1500,19 @@ int ukplat_paging_init(void)
 	/* Perform mappings */
 	ukplat_memregion_foreach(&mrd, 0, UKPLAT_MEMRF_MAP,
 				 UKPLAT_MEMRF_MAP) {
+		UK_ASSERT(!(mrd->vbase & ~PAGE_MASK));
```
michpappas: This looks incorrect.

include/uk/plat/memory.h
```diff
@@ -101,6 +104,124 @@ struct ukplat_memregion_desc {
 #endif /* CONFIG_UKPLAT_MEMRNAME */
 } __packed __align(__SIZEOF_LONG__);
 
+/** UK_ASSERT_VALID_MRD_TYPE(mrd) macro
+ *
+ * Ensure a given memory region descriptor has one of the following defined
+ * types only:
+ *	UKPLAT_MEMRT_FREE		0x0001	Uninitialized memory
```
michpappas: Although unlikely to change, better avoid duplicating the values of these definitions just in case. Similarly on other macros below.

include/uk/plat/memory.h
```diff
@@ -101,6 +104,124 @@ struct ukplat_memregion_desc {
...
+ *	UKPLAT_MEMRT_KERNEL		0x0004	Kernel binary segment
+ *	UKPLAT_MEMRT_INITRD		0x0008	Initramdisk
+ *	UKPLAT_MEMRT_CMDLINE		0x0010	Command line
+ *	UKPLAT_MEMRT_DEVICETREE		0x0020	Device tree
+ *	UKPLAT_MEMRT_STACK		0x0040	Thread stack
+ * @param mrd pointer to the memory region descriptor whose type to validate
+ */
+#define UK_ASSERT_VALID_MRD_TYPE(mrd)					\
+	do {								\
+		switch ((mrd)->type & UKPLAT_MEMRT_ANY) {		\
```
michpappas: Since `type` is governed by a controlled vocabulary I don't see what is the purpose of the logic AND. Also it makes an assumption about the value of `UKPLAT_MEMRT_ANY`.

include/uk/plat/memory.h
```diff
@@ -101,6 +104,124 @@ struct ukplat_memregion_desc {
...
+		UK_ASSERT(PAGE_ALIGNED((mrd)->vbase));			\
+		UK_ASSERT(PAGE_ALIGNED((mrd)->pbase));			\
+		UK_ASSERT((mrd)->pg_off >= 0 &&				\
+			  (mrd)->pg_off < (__off)PAGE_SIZE);		\
+	} while (0)
+
+/** UK_ASSERT_VALID_FREE_MRD(mrd) macro
+ *
+ * Ensure free memory region descriptor particular correctness:
+ * - must meed the criterias of a general valid memory region descriptor
```
michpappas: s/criterias/criteria/

include/uk/plat/memory.h
```diff
@@ -101,6 +104,124 @@ struct ukplat_memregion_desc {
...
+		UK_ASSERT((mrd)->type == UKPLAT_MEMRT_FREE);		\
+		UK_ASSERT((mrd)->vbase == (mrd)->pbase);		\
+		UK_ASSERT((mrd)->pg_count * PAGE_SIZE == (mrd)->len);	\
+		UK_ASSERT(!(mrd)->pg_off);				\
+	} while (0)
+
+/** UK_ASSERT_VALID_KERNEL_MRD(mrd) macro
+ *
+ * Ensure kernel memory region descriptor particular correctness:
+ * - must meed the criterias of a general valid memory region descriptor
```
michpappas: s/meed/meet/
s/criterias/criteria/

plat/common/memory.c
```diff
@@ -391,6 +394,7 @@ void ukplat_memregion_list_coalesce(struct ukplat_memregion_list *list)
 						  ml->pg_count * PAGE_SIZE)) {
 					uk_pr_debug("Deleting left mrd!\n");
 					ukplat_memregion_list_delete(list, i);
+					continue;
```
michpappas, line 400: I think this fix does not belong to this commit. If the logic error was present in the code already it would be worth adding a new commit that also provides some details on the fixed issue.

plat/common/paging.c
```diff
@@ -1488,6 +1486,15 @@ int ukplat_paging_init(void)
 	/* Perform unmappings */
 	ukplat_memregion_foreach(&mrd, 0, UKPLAT_MEMRF_UNMAP,
 				 UKPLAT_MEMRF_UNMAP) {
+		/* Ensure unmap memory region descriptors correctness */
```
michpappas: s/descriptors/descriptors'/

plat/common/paging.c
```diff
@@ -1488,6 +1486,15 @@ int ukplat_paging_init(void)
...
+		/* Must be non-empty and aligned end-to-end */
+		UK_ASSERT(mrd->len);
+		UK_ASSERT(mrd->pg_count * PAGE_SIZE == mrd->len);
+		UK_ASSERT(!(mrd->vbase & ~PAGE_MASK));
```
michpappas: Is that right? 👴🏼

michpappas (changes requested):
@mogasergiu this is a nice changeset. Besides the incline comments, a few grammar issues I spotted in the commit message of the 2nd commit.

> As a basis, MRD's must

MRDs (or mrds to be consistent with the rest)

> validation macro's were added.

macros

> addition of these macro's,

macros

> criteria's

critieria

> we use meets the criterias that all free memory regions must have

criteria

> of `ukplat_memregion_list_coalesce` against the general criterias

criteria

> criterias that must be met by all memory region descriptors.

criteria

> Properly validate mrd's

mrds

plat/common/memory.c
mogasergiu: I would like to keep the assertion in the caller. I want to make it clear in the caller that that branch is to be taken only if one of them is of type FREE. The callee here through this `mr->type == ml->type` assertion ensure that it has only be called in such a scenario, while the type of the mrd assertion in the caller is there to ensure that that branch of the caller can only happen if mr (or ml for the second branch) is the one that is free. I generally don't see assertions as redundant: the caller branch ensures it has been reached only by such scenarios without making assumption about callee, while callee ensures it has been called only in its specific scenarios, without making assumptions about caller.

plat/kvm/x86/lxboot.c
mogasergiu: > Length in bytes of the resource inside this region

Oops should have been ` + mrd.pg_off`

mogasergiu: Ah, wait, nevermind, I remembered. Ok I actually extend free memory regions to always be multiple of PAGE_SIZE. This is to extend to page aligned margins both start/end of free mrd's. Some memory maps give only page-sized free memory regions and the next memory region starts at an in-page offset. This means we may lose some bytes of free memory that may have otherwise been used by allocators that don't require page granularity like bbuddy, e.g. TLSF.

plat/kvm/x86/multiboot.c
```diff
@@ -134,14 +139,21 @@ void multiboot_entry(struct lcpu *lcpu, struct multiboot_info *mi)
...
 
 			if (m->type == MULTIBOOT_MEMORY_AVAILABLE) {
 				mrd.type  = UKPLAT_MEMRT_FREE;
 				mrd.flags = UKPLAT_MEMRF_READ |
 					    UKPLAT_MEMRF_WRITE;
+
+				/* Free memory regions have
+				 * mrd.len == mrd.pg_count * PAGE_SIZE
+				 */
+				mrd.len = PAGE_ALIGN_UP(mrd.len);
```
mogasergiu: Therefore, ditto here. I guess I will add a comment.

plat/common/memory.c
mogasergiu: This is actually not a fix. Before it was handled correctly through the ugly `goto restore_mrds;`

plat/common/memory.c
mogasergiu: Is that right :older_man: ?

plat/common/memory.c
michpappas: Okay sure, let's keep it then 👍🏼

plat/common/memory.c
michpappas: It was when I wrote it, now that I look at the code again it seems to be fine 😉

plat/common/memory.c
michpappas: Okay then perhaps at least move to the commit that removes that `goto`. Currently it is part of the commit that adds the assertions :)

michpappas (approved):
Thanks!

razvanvirtan (approved):
Thank you, @mogasergiu!

PR #1214, opened 2023-12-17
arch: Introduce dedicated CPU Exception stack configuration

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Before, traps and IRQ's would share the same stack size as user
created threads through the `STACK_SIZE` macro. This would end up
unnecessarily increasing the Unikraft in-memory footprint by a very
large amount.

Application stacks can be very large (MB in size, even bigger than the
Unikernel itself) and CPU exception handling should not need such
massive sized stacks. Therefore, introduce the
`CONFIG_CPU_EXCEPT_STACK_SIZE_PAGE_ORDER` option having a default
value of `4`, just like its user stack counterpart.

The difference will show itself in practice, as it is not a rare case
that we need to increase stack sizes to allow applications to work
properly, especially with filesystem hungry ones that demand a lot
of stack for their system call handling.


plat/common/arm/lcpu.c
```diff
@@ -58,7 +58,7 @@
  *                     SP_EL0
  */
 static __align(UKARCH_SP_ALIGN)
-UKPLAT_PER_LCPU_ARRAY_DEFINE(__u8, lcpu_irqntrap_sp, STACK_SIZE * 2);
+UKPLAT_PER_LCPU_ARRAY_DEFINE(__u8, lcpu_irqntrap_sp, CPU_EXCEPT_STACK_SIZE * 2);
```
mogasergiu, line 74: Must also update `exceptions.S` for IRQ entries to add `CPU_EXCEPT_STACK_SIZE` instead of `STACK_SIZE`

michpappas:
One more. In the following commit message:
```
    arch: Make context specific macro's visible in assembly code
    
    Simple macro's that do not involve C code may as well be visible to code
    written in assembly. This may come in handy when not wanting to use
    magic values.
    
    Therefore, make it so that architecture specific macro's from context
    related headers are also visible to assembly written code by having
    them defined outside the `!__ASSEMBLY` guards.
```
A `s/macro's/macros/` will keep the Apostrophe Protection Society happy.

michpappas (approved):
Looks good to me, subject to https://github.com/unikraft/unikraft/pull/1214#discussion_r1431171065

razvanvirtan (approved):
This one looks good to me, thank you @mogasergiu!

razvand (approved):

PR #1215, opened 2023-12-17
.github/workflows: Move integration tests to upb runners

craciunoiuc:
### Prerequisite checklist

### Base target

### Description of changes

You can see all usable labels here: <https://github.com/organizations/unikraft/settings/actions/runners>

If you need direct access to the systems for whatever reason, i can hook you up :pray: 

The list of labels uses a logical AND between them, which means `[upb, qemu, X64]` will run on x86_64 systems that are UPB and have qemu installed

Edit:

~~needs some github-action fixes~~

~~not urgent anyways~~

Fixed, works now on 2 systems. Might be a bit slow if we start using it everywhere.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,8 @@
+self-hosted-runner:
+  labels:
+    - upb
+    - qemu
+    - xen
+    - run-qemu
+    - build-any
+config-variables: null
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -13,7 +13,7 @@ on:
 jobs:
   libc-test:
     name: libc-test
-    runs-on: ubuntu-latest
+    runs-on: [upb, qemu, X64, run-qemu]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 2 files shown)


nderjung:
Is it possible to still include `ubuntu-latest` in the list of runners? (Just in case any of the UPB machines go down?)

Also a technical question if it's possible to mix self-hosted with GH provided runners?

craciunoiuc:
@nderjung only place that mentions something related to this would be here <https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#choosing-runners-in-a-group>

Basically you can create a group of runners together with the self-hosted ones, **but** this only works with their premium, larger, instances.
So, at a glance I would say no to both questions, or at least not easily.
They would need to offer access to their scheduling policy to do something interesting with it.
Basically we could use GitHub-Runners only in the case where there are more than 50 queued jobs, etc.

Interesting part to read as there were questions https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#usage-limits

craciunoiuc:
If you think this could disrupt work, then we can have the upb runners only crunch on nightly builds of the catalog as a compromise

craciunoiuc:
This is nice and makes sense, but, self-hosted runners are not needed in OSS repositories, as runners are already free.

We will revisit this when costs apply and it makes sense @nderjung

PR #1226, opened 2023-12-22
lib/{vfscore, posix-tty}: Introduce `posix-tty` & move stdio/console files out of vfscore

### LINKED ISSUE #41: vfscore: Support closing STD{IN,OUT,ERR} file descriptors
The initial release of `vfscore` library did not introduce closing of standard file descriptors. This is also mentioned in the code [here](https://github.com/unikraft/unikraft/blob/fc022dfa/lib/vfscore/fd.c#L100). The underlying structure for standard file descriptors is [statically allocated](https://github.com/unikraft/unikraft/blob/fc022dfa/lib/vfscore/stdio.c#L189), unlike ordinary file descriptors. A proper fix should allocate these dynamically as well.

andreittr:
### Description of changes

This changeset introduces a new library `posix-tty` that provides newvfs files to use as the system console (stdin/out/err), moving this functionality out of legacy vfscore and finally making file descriptors `0`, `1`, and `2` normal files like any other, with no need for special handling in unikraft core or apps.

Provides the following file types:
- pseudo-files: NULL, void, zero -- reads return constant pre-determined values; writes are discarded
- serial: platform-specific serial console, similar to what we used before

In addition, console files are now pollable, enhancing application support.
The `stat` family of syscalls is supported for both serial and pseudo-files, and return meaningful values comparable to Linux.
Serial files support a core set of tty ioctls.

For compatibility with older configurations, vfscore now selects `posix-tty`.

Potential conflict w/ https://github.com/unikraft/unikraft/pull/1341; whichever merges first will require a patch to the other.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A


andreittr:
~~This PR is **ON HOLD** pending https://github.com/unikraft/unikraft/pull/1245; converted to draft.
Once that is merged, I'll rebase and reopen.~~

andreittr:
Improved, rebased, reopened & updated PR description.

lib/posix-tty/serial.c
```diff
@@ -0,0 +1,259 @@
...
+	for (int i = 0; i < iovcnt; i++) {
+		char *buf = iov[i].iov_base;
+		size_t len = iov[i].iov_len;
+		char *last;
+		int bytes_read;
+
+		if (unlikely(!buf && len))
+			return -EFAULT;
+
+		bytes_read = ukplat_cink(buf, len);
```
jschlumpp, line 76: IIRC this operation will block until a character has been read. So the whole thing won't be async currently?

jschlumpp (commented):
Just one small question but looks good otherwise.

lib/posix-tty/serial.c
andreittr: If `ukplat_cink` blocks then indeed serial read will also block & won't ever return `-EAGAIN`. The rest of ukfile/posix-fd* code should handle this gracefully though -- it expects files to always be nonblocking, but doesn't _demand_ them to be so; it can handle always-blocking files just fine.

Where this could pose compat issues is if an application opens/fcntls the serial file `O_NONBLOCK` and then _absolutely relies_ on nonblocking behavior. That might be unlikely, since `O_NONBLOCK` is merely a hint and doesn't AFAIK make `open`/`fcntl` error out if the file cannot be opened nonblockingly, but worth considering.
In that case we may need to change `ukplat_cink` to only probe the serial input and never block.

Admittedly the current implementation is suboptimal for such a probing `ukplat_cink`, as the readiness flags are set very conservatively: POLLOUT is always set, POLLIN remains set until we get an EOF => polling code would immediately return, busy-waiting on console input. We'd much rather want to signal POLLIN only when input is available.

jschlumpp: Okay blocking was maybe a bit imprecise. The problem is that the serial read will busy-poll on the port. So it's not really "blocking" in the classical unix I/O sense, but worse.
(Note: this is not directly a problem with this PR, but I wanted to make clear how the underlying serial code behaves currently)

andreittr: Oh I see, so it's spinning/busy-waiting on the serial input. Indeed much worse :sweat_smile: .
Let's hope actual platform serial input is never performance-critical, but definitely something worth improving in the long run.

jschlumpp: I think it's less a performance problem and more a function problem. AFAIK it soft-breaks the quake server, because you have repeatedly type stuff on the serial to make it continue.
EDIT: I have to add that you can at least redirect stdin/stdout to somewhere else with this PR. So you could workaround that.

andreittr: By default AFAICT (Kconfig picks the 1st choice unless told otherwise) we have null as stdin instead of serial console, which I believe should fix daemons like the quake server, or does a null stdin also pose an issue?

I agree the current serial implementation is not quite fit for serious workloads, and a better one would be welcome.

andreittr:
Force-push: rebased on current staging + integrated improvements from newly-merged https://github.com/unikraft/unikraft/pull/1397.

lib/posix-tty/serial.c
jschlumpp: null as stdin should be fine I guess. Might be a bit surprising for interactive workloads, but I don't expect people to do that much (and that would be likely broken anyway).

jschlumpp (approved):
Nice improvement.
(Let's fix the busy-poll serial port somewhere else.)

lib/posix-tty/serial.c
```diff
@@ -0,0 +1,155 @@
...
+
+static int serial_ctl(const struct uk_file *f, int fam, int req, uintptr_t arg1,
+		      uintptr_t arg2 __unused, uintptr_t arg3 __unused)
+{
+	UK_ASSERT(f->vol == SERIAL_VOLID);
+
+	switch (fam) {
+	case UKFILE_CTL_IOCTL:
+		switch (req) {
+		case TIOCGWINSZ:
```
michpappas, line 156: Not sure why this particular ioctl was selected to be implemented, who is the consumer of this info? If there is a use-case, should the `TIOCSWINSZ` counterpart be there as well? Moreover, since there is a commit that introduces more ioctls, perhaps this one belongs there too.

lib/posix-tty/pseudo.c
```diff
@@ -0,0 +1,155 @@
...
+/* Pseudo-files: files with constant input and discarding output */
+
+#include <string.h>
+
+#include <uk/assert.h>
+#include <uk/file.h>
+#include <uk/file/nops.h>
+#include <uk/posix-fd.h>
+
+/* Null file: input is always at EOF */
```
michpappas, line 17: Maybe I'm missing something, are these standard (POSIX)? From what I see these can be selected in config to implement various use-cases, could you provide some examples?

lib/posix-tty/Config.uk
michpappas, line 1: I would expect that `Config.uk`, `Makefile`, `tty.c` belong to the commit that introduces the library, and this commit only does what the title states. Is there a reason for this split?

michpappas (changes requested):
Hi @andreittr, only some comments on the structure and a couple of clarifications. Also the changes introduced in the commit that fixes warnings in `libuktty` should probably be squashed to the commit(s) that introduce the affected code in this series. Thanks.

lib/posix-tty/serial.c
andreittr: TIOCGWINSZ is commonly used to check whether stdin/out is a terminal; tbh I don't remember the exact workload (quake? dunno), but I have seen it called during development. The values themselves probably don't matter except for fancy paged output, but some software relies on sane values here to accept that stdin is a terminal.
The 'set' counterpart (TIOCSWINSZ) IMO is ok to error out from unless there is a benefit to keeping the size as mutable file state. According to the manpages for `tty_ioctl`, the window size is kept by but not directly used by the kernel.

As for which commit this is in? Development & testing order; I have not tested tty files w/o this ioctl and I cannot guarantee they won't break if I do move things around. But this seems innocent enough to squash into the ioctl commit; will do in the pass.

lib/posix-tty/pseudo.c
andreittr: By "these" do you mean the pseudo-file types? I'm not sure if POSIX standardises devices that a system would have available.

What examples beyond the general description of pseudofiles ("files with constant input and discarding output") and the individual clarifications ("input is always at EOF", etc.) would you think would be genuinely useful to describe them better? You could say null and zero files roughly correspond to `/dev/null` and `/dev/zero` respectively (hence the name similarity), but AFAIK the "void" file has no equivalent (I guess it's most similar to a non-broken read end of a pipe whose write end will never be touched).

lib/posix-tty/Config.uk
andreittr: ~~I'm not sure what commit "this commit" is referring to, and github's review interface isn't helpful; could you be more specific?~~

edit: ok, I think I figured out what you mean, please correct me if I'm wrong: you are arguing that commit "Move stdio out of vfscore [...]" is redundant and should be folded into the previous one "Introduce posix-tty library" (using messages instead of hashes b/c force-pushes). I consider that the two commits provide quite different features:
- 1st provides `ukfile` drivers for character pseudo-files and the platform serial console; these files can be used independently of other posix APIs
- 2nd provides initialization of fds 0, 1, and 2 as configured for stdin/out/err during unikraft boot and is strictly dependent on an fdtab

Although now that you mention it, this separation should be more apparent in Config, to allow selecting the 1st bit of functionality w/o the latter. This will also work well as a broader separation of mechanism from API within the ukfile ecosystem.
I'm going to make this change in the review pass.

andreittr:
Force-push: rebase to staging & update to ukfile API changes introduced by https://github.com/unikraft/unikraft/pull/1341

DeliaPavel (approved):

razvand (approved):

PR #1146, opened 2023-10-23
lib/devfs: Remove dependency on `LIBVFSCORE_AUTOMOUNT_ROOTFS`

mogasergiu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Now that we have `fstab`, we no longer depend on something mounting a rootfs being preconfigured within the kernel. A user may simply choose to specify rootfs through the `vfs.fstab` command-line argument.

PR #1240, opened 2024-01-05
lib/devfs: select rootfs automount on devfs automount

lukaslihotzki:
### Prerequisite checklist

 - [x] Updated relevant documentation. (no documentation to update)

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

devfs automount fails without rootfs automount:
```
[    0.102003] ERR:  [libdevfs] <devfs_vnops.c @  315> Failed to create /dev: 20
[    0.107465] ERR:  [libukboot] <boot.c @  393> Init function at 0x116ea0 returned error -1
```

When `LIBVFSCORE_AUTOMOUNT_ROOTFS` is selected, `ramfs` is chosen by default, which is a reasonable choice if you just need `/dev`.

### DIFF
--- a/lib/devfs/Config.uk
+++ b/lib/devfs/Config.uk
@@ -6,6 +6,7 @@ menuconfig LIBDEVFS
 if LIBDEVFS
 	config LIBDEVFS_AUTOMOUNT
 	bool "Mount /dev during boot"
+	select LIBVFSCORE_AUTOMOUNT_ROOTFS
 	default n
 
 	# hidden


razvand:
@lukaslihotzki , note that `LIBVFSCORE_AUTOMOUNT_ROOTFS` may not be required if you mount the root filesystem via `vfs.fstab`. For example, [the `base` image](https://github.com/unikraft/catalog/tree/main/library/base) doesn't have the `LIBVFSCORE_AUTOMOUNT_ROOTFS` enabled, and it relies on the an explicit mount via `vfs.fstab` as is the case with the [`http-tokio` example](https://github.com/unikraft/catalog/pull/33).

So I wouldn't force the selection of the `LIBVFSCORE_AUTOMOUNT_ROOTFS` when `LIBDEVFS_AUTOMOUNT` is selected. As you may simply choose your own way of mounting the root filesystem.

skuenzer:
> @lukaslihotzki , note that `LIBVFSCORE_AUTOMOUNT_ROOTFS` may not be required if you mount the root filesystem via `vfs.fstab`. For example, [the `base` image](https://github.com/unikraft/catalog/tree/main/library/base) doesn't have the `LIBVFSCORE_AUTOMOUNT_ROOTFS` enabled, and it relies on the an explicit mount via `vfs.fstab` as is the case with the [`http-tokio` example](https://github.com/unikraft/catalog/pull/33).
> 
> So I wouldn't force the selection of the `LIBVFSCORE_AUTOMOUNT_ROOTFS` when `LIBDEVFS_AUTOMOUNT` is selected. As you may simply choose your own way of mounting the root filesystem.

In fact, PR #1146 removed this dependency because of these reasons.

skuenzer:
I am closing this PR because it would re-introduce the option that we decided to take out with PR #1146.

PR #1244, opened 2024-01-08
lib/nolibc: Update musl-imported headers for signals

michpappas:
### Prerequisite checklist

### Base target

 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series updates the `musl-imported` headers for signals:
* Import `feature.h` from musl to simplify management of FTMs (https://lwn.net/Articles/590381/)
* Replace `signal.h` with a clean copy from musl to simplify manitenance
* Import architecture-specifi headers for `arm64` and `x86_64`
* Import additional definitions to resolve dependencies


lib/nolibc/musl-imported/include/signal.h
```diff
@@ -83,63 +178,124 @@ struct sigaction {
...
-int killpg(int pgrp, int sig);
-int raise(int sig);
-int siginterrupt(int sig, int flag);
-void psignal(int sig, const char *s);
+#define SYS_SECCOMP 1
+#define SYS_USER_DISPATCH 2
+#endif
 
+#ifdef _GNU_SOURCE
 typedef void (*sighandler_t)(int);
```
csvancea, line 270: `lib/uksignal/signal.c` does not define `_GNU_SOURCE`, which leads to `sighandler_t` being undefined here: https://github.com/michpappas/unikraft/blob/1d52b24e408c0918d967f643494668d1991492bc/lib/uksignal/signal.c#L87

Of course this is tackled in #1248, but is it expected to have build failures in-between this PR and #1248?

John-Ted:
Can you rebase this on the latest main?

lib/nolibc/musl-imported/include/signal.h
michpappas: Thanks, fixed

michpappas:
Rebased from `staging`. Notice that the checkpatch errors are due to:
* Tags for tracability that checkpatch confuses for invalid SPDX tags
* Non-confirming syntax of musl (we don't adapt to our convention)

lib/nolibc/musl-imported/src/string/strsignal.c
```diff
@@ -2,6 +2,7 @@
 
 #include <uk/essentials.h>
 #include <string.h>
+#define _GNU_SOURCE
```
andreittr, line 3: Hate to be nitpicky, but I think we should define any feature test macros before including any header files. The `feature_test_macros(7)` manpage says as much in the first note: "In order to be effective, a feature test macro must be defined before including any header files.". Or is there a very compelling reason why we'd want only `<signal.h>` to see this macro?

(We know `<uk/essentials.h>` doesn't use any feature macros, so we can maybe ignore that, but `<string.h>` is outside of our control (technically it isn't since we're inside nolibc, but fewer assumptions make for cleaner code)).

michpappas: Not nitpicky at all, as you pointed out FTMs must be defined first. No idea why I did it that way back then, but I testing the proper way didn't show any problems, so I will push an update. Thanks for catching that.

andreittr (approved):
Looks good, thanks for the work!

andreittr:

PR #1246, opened 2024-01-08
support/scripts: Implement python and bash suggestions

craciunoiuc:
### Prerequisite checklist

### Base target

 - Architecture(s): all
 - Platform(s): `kvm` && `xen`
 - Application(s): all

### Description of changes

I went through and fixed all `bash` files to respect the shellcheck guidelines. The only style suggestion that I did not fix was removing `$?`, which seemed unnecessary.

@skuenzer if you do not agree with a change, revert it and run `shellcheck` on the file and see what it says. We can ignore that warning/error/style if you want. note: Only warnings/errors cause the pipeline to fail.

While doing this, I saw that `qemu-guest` had one duplicate argument. Let me know if I changed it correctly.

@skuenzer please slowly go through everything to make sure that functionality was not altered while linting.

### DIFF
--- a/support/scripts/mkefi.py
+++ b/support/scripts/mkefi.py
@@ -2,151 +2,176 @@
 
 import argparse
 import subprocess
-import os
 import re
-from struct import pack
 
-ELF_MACHINE = {
-    'x86_64': 62,               # EM_X86_64
-    'arm64': 183                # EM_AARCH64
-}
+ELF_MACHINE = {"x86_64": 62, "arm64": 183}  # EM_X86_64  # EM_AARCH64
 
 PE_MACHINE = {
-    'x86_64': 0x8664,           # IMAGE_FILE_MACHINE_AMD64
-    'arm64' : 0xaa64            # IMAGE_FILE_MACHINE_ARM64
+    "x86_64": 0x8664,  # IMAGE_FILE_MACHINE_AMD64
+    "arm64": 0xAA64,  # IMAGE_FILE_MACHINE_ARM64
 }
 
 PE_SEC_SZ = 40
 
 # COFF Characteristics
 COFF_HDR_CHRS = (
-                 0x0002     |  # IMAGE_FILE_EXECUTABLE_IMAGE
-                 0x0004     |  # IMAGE_FILE_LINE_NUMS_STRIPPED
-                 0x0200        # IMAGE_FILE_DEBUG_STRIPPED
+    0x0002
+    | 0x0004  # IMAGE_FILE_EXECUTABLE_IMAGE
+    | 0x0200  # IMAGE_FILE_LINE_NUMS_STRIPPED  # IMAGE_FILE_DEBUG_STRIPPED
 )
 
 # All sections that we obtain from the PT_LOAD Program Headers must be
 # writable, since we have to apply initial relocations. We are going to
 # change permissions ourselves afterwards anyway.
 PE_SEC_CHRS = (
-                0x00000040 |  # IMAGE_SCN_CNT_INITIALIZED_DATA
-                0x00d00000 |  # IMAGE_SCN_ALIGN_4096BYTES
-                0x20000000 |  # IMAGE_SCN_MEM_EXECUTE
-                0x40000000 |  # IMAGE_SCN_MEM_READ
-                0x80000000    # IMAGE_SCN_MEM_WRITE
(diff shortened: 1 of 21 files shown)


craciunoiuc:
@skuenzer we should get this in sooner than later, it easily can break other PRs/be broken by them

support/scripts/mkefi.py
```diff
@@ -2,183 +2,208 @@
...
-    "SizeOfHeapReserve"          : (0x0000000000000000, 0x08),  # Offset 176
-    "SizeOfHeapCommit"           : (0x0000000000000000, 0x08),  # Offset 184
-    "LoaderFlags"                : (0x0000000000000000, 0x04),  # Offset 192
-    "NumberOfRvaAndSizes"        : (0x0000000000000000, 0x04),  # Offset 196
-
+    "MajorImageVersion": (0x0000000000000000, 0x02),  # Offset 132
+    "MinorImageVersion": (0x0000000000000000, 0x02),  # Offset 134
+    "MajorSubsystemVersion": (0x0000000000000000, 0x02),  # Offset 136
+    "MinorSubsystemVersion": (0x0000000000000000, 0x02),  # Offset 138
+    "Win32VersionValue": (0x0000000000000000, 0x04),  # Offset 140
```
skuenzer, line 68: I have to say that sometimes the tool is too brutal with alignment. I find the previous block more readable.

support/scripts/xen-guest
```diff
@@ -550,10 +555,10 @@ if [ ${OPT_GDBPORT} -eq 0 ]; then
 	printf '   GDB server:            %s\n' "${ARG_GDBPORT}"
 fi
 printf '\n'
-if [ ${OPT_GDBPORT} -eq 0 -a ${OPT_KERNEL} -eq 0 ]; then
+if [ ${OPT_GDBPORT} -eq 0 ] && [ ${OPT_KERNEL} -eq 0 ]; then
```
skuenzer, line 558: Do you know why double quoting is not added here, as well?

craciunoiuc: No idea, it did not complain about this line.

My hunch is that it sees it is checked as a number, so globbing/splitting would not make sense in this case.

Theoretically you can add quotes everywhere, just that in not all places it's necessary

skuenzer: Seems to be less intelligent with `NICID` which is also a number only. 🤔

support/scripts/mkefi.py
craciunoiuc: Robots show no compromise, unless you specify to them. If you can, look for any way to ignore lines, but not sure if there's something for this.

Me I see it as this: If as simple as possible (no formatting), everyone can style however they like when reading.
For example I did not agree with the spaces before the `:`, they should have been after it

craciunoiuc: I would say take the hit an leave like this and look to add line ignores only if someone complains about readability.

support/scripts/xen-guest
craciunoiuc: We could just add quotes everywhere and be done with it :D

though I would prefer to do changes only to shut it up, cause things worked before anyways (apart from a few bugs that the check actually fixed)

support/scripts/mkefi.py
skuenzer: Actually, I think these robots are there to support us but not to replace. There are situations (like this alignment) where a human should be able to overwrite robot behavior. We do this with checkpatch on the other unikraft sources: in some cases we accept a wrong formatting (in the view of the robot) during review/approval because it still improves the overall code base.
I understand that we have a limitation here with the script format tool and how we integrate it into a github workflow. We can only check the whole file and not just the diff that a commit/PR introduces. This is why we have to live with it and understand the we cannot to a mixed approach for scripts.

support/scripts/xen-guest
skuenzer: Is fine, I am testing some scripts and we merge this PR then.

skuenzer (approved):

support/scripts/mkefi.py
craciunoiuc: I think we could make it ignore those lines, but we would have to write on every line an ignore sequence, which would make it look worse than just letting it format.

I get what you mean. We can look into more in-tune formatting after we clean up all files

PR #1259, opened 2024-01-11
lib/posix-poll: Fix epoll crash on NULL event arg

andreittr:
### Description of changes

Previously epoll_ctl would crash on `EPOLL_CTL_ADD` and `EPOLL_CTL_MOD` if its `event` arg was NULL.
No global checking is done since `event == NULL` is a valid case for `EPOLL_CTL_DEL`.
This change fixes this oversight, safely returning -EFAULT on a NULL arg for the add and mod operations.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

`CONFIG_LIBPOSIX_POLL=y`

Test snippet:
```c
int e = epoll_create1(0);
int r = epoll_ctl(e, EPOLL_CTL_ADD, 0, NULL);
assert(r == -EFAULT); /* No segfault should happen */
```

PR #1260, opened 2024-01-11
lib/vfscore: Stub `F_GETLK` to always return `F_UNLCK`

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Since we always run in an exclusive virtual filesystem mode (no multiple processes trying to acquire the lock on the same file), it is fine for now to stub `F_GETLK` to always return an unlocked file through the `F_UNLCK` flag.

This way, applications that do assume non-exclusive virtual filesystem mechanisms can still proceed as if the file is free to be acquired.

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -1967,7 +1967,20 @@ int vfscore_fcntl(struct vfscore_file *fp, unsigned int cmd, unsigned long arg)
 		uk_pr_warn_once("fcntl(F_SETLK) stubbed\n");
 		break;
 	case F_GETLK:
-		uk_pr_warn_once("fcntl(F_GETLK) stubbed\n");
+		uk_pr_warn_once("fcntl(F_GETLK) stubbed. Always unlocked\n");
+
+		struct flock *flk = (struct flock *)arg;
+
+		if (unlikely(!flk)) {
+			error = EFAULT;
+			goto out_errno;
+		}
+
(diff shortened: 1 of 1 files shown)


lib/vfscore/main.c
```diff
@@ -1967,7 +1967,17 @@ int vfscore_fcntl(struct vfscore_file *fp, unsigned int cmd, unsigned long arg)
 		uk_pr_warn_once("fcntl(F_SETLK) stubbed\n");
 		break;
 	case F_GETLK:
-		uk_pr_warn_once("fcntl(F_GETLK) stubbed\n");
+		uk_pr_warn_once("fcntl(F_GETLK) stubbed. Always unlocked\n");
+
+		struct flock *flk = (struct flock *)arg;
+
+		UK_ASSERT(flk);
```
andreittr: Should we perhaps return -EFAULT here if `!flk`, instead of an assert (and a possible null deref with asserts disabled)?
(Usually asserts are the right call, but here (1) `arg` is received straight and unsanitized from userspace, and (2) its meaning is not defined until `cmd` is interpreted, therefore we have no other place to validate it)

A similar silly issue stemming from poorly validated syscall args: https://github.com/unikraft/unikraft/pull/1259.

andreittr (approved):
Looks good, thanks!

skuenzer (approved):

PR #1261, opened 2024-01-11
lib/nolibc: Import `strstr` from `musl`

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Import from musl the basic implementation of the widely available string searching method `strstr`.

### DIFF
--- a/lib/nolibc/musl-imported/src/string/strstr.c
+++ b/lib/nolibc/musl-imported/src/string/strstr.c
@@ -1,0 +1,208 @@
+/* taken from musl */
+
+#include <stdint.h>
+#include <string.h>
+
+static char *twobyte_strstr(const unsigned char *h, const unsigned char *n)
+{
+	__u16 nw = n[0] << 8 | n[1], hw = h[0] << 8 | h[1];
+
+	for (h++; *h && hw != nw; hw = hw << 8 | *++h)
+		;
+
+	return *h ? (char *)h - 1 : 0;
+}
+
+static char *threebyte_strstr(const unsigned char *h, const unsigned char *n)
+{
+	__u32 nw = (__u32)n[0] << 24 | n[1] << 16 | n[2] << 8;
+	__u32 hw = (__u32)h[0] << 24 | h[1] << 16 | h[2] << 8;
+
+	for (h += 2; *h && hw != nw; hw = (hw | *++h) << 8)
+		;
+
+	return *h ? (char *)h - 2 : 0;
+}
+
+static char *fourbyte_strstr(const unsigned char *h, const unsigned char *n)
+{
+	__u32 nw = (__u32)n[0] << 24 | n[1] << 16 | n[2] << 8  | n[3];
+	__u32 hw = (__u32)h[0] << 24 | h[1] << 16 | h[2] << 8 | h[3];
+
+	for (h += 3; *h && hw != nw; hw = hw << 8 | *++h)
+		;
+
+	return *h ? (char *)h - 3 : 0;
+}
+
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
+
+#define BITOP(a, b, op)						\
+	((a)[(__sz)(b) / (8 * sizeof *(a))] op			\
+	 (__sz)1 << ((__sz)(b) % (8 * sizeof *(a))))
+
+static char *twoway_strstr(const unsigned char *h, const unsigned char *n)
+{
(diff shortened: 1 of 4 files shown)


lib/nolibc/musl-imported/src/string/strstr.c
```diff
@@ -0,0 +1,156 @@
+/* taken from musl */
+
+#include <string.h>
+#include <stdint.h>
+
+static char *twobyte_strstr(const unsigned char *h, const unsigned char *n)
+{
+	uint16_t nw = n[0]<<8 | n[1], hw = h[0]<<8 | h[1];
+	for (h++; *h && hw != nw; hw = hw<<8 | *++h);
```
RaduNichita: Hi @mogasergiu ,

Can you please fix the linting problem?

lib/nolibc/musl-imported/src/string/strstr.c
```diff
@@ -0,0 +1,156 @@
+/* taken from musl */
+
+#include <string.h>
+#include <stdint.h>
+
+static char *twobyte_strstr(const unsigned char *h, const unsigned char *n)
```
RaduNichita, line 6: Can we refactor those function to make something more generic?

lib/nolibc/musl-imported/src/string/strstr.c
```diff
@@ -0,0 +1,156 @@
+/* taken from musl */
```
RaduNichita, line 1: Could you provide a link from where did you take this?

lib/nolibc/musl-imported/src/string/strstr.c
```diff
@@ -0,0 +1,156 @@
...
+
+static char *fourbyte_strstr(const unsigned char *h, const unsigned char *n)
+{
+	uint32_t nw = (uint32_t)n[0]<<24 | n[1]<<16 | n[2]<<8 | n[3];
+	uint32_t hw = (uint32_t)h[0]<<24 | h[1]<<16 | h[2]<<8 | h[3];
+	for (h+=3; *h && hw != nw; hw = hw<<8 | *++h);
+	return *h ? (char *)h-3 : 0;
+}
+
+#define MAX(a,b) ((a)>(b)?(a):(b))
```
RaduNichita: Should we define this? I don't think it is a good idea to put this here.

RaduNichita (commented):
Thanks @mogasergiu 💯  

Left some questions!

lib/nolibc/musl-imported/src/string/strstr.c
mogasergiu: I simply aligned this with `strsignal.c` also taken from musl.  I don't think it is needed, check our `strsignal.c` file.

lib/nolibc/musl-imported/src/string/strstr.c
mogasergiu: This is directly taken from musl's [strstr.c](https://github.com/bminor/musl/blob/master/src/string/strstr.c).

lib/nolibc/musl-imported/src/string/strstr.c
mogasergiu: I generally agree with your points here. And although yes, musl is under MIT license, allowing a great amound of freedom, including publishing modified versions, I would still keep it as is :thinking: 

EDIT: I will probably re-write some more parts of this file. It is after all a copy-paste of an MIT licensed file.

lib/nolibc/musl-imported/src/string/strstr.c
mogasergiu: Yeah this does look like something worth aligning to our coding style standards :+1:.

lib/nolibc/musl-imported/src/string/strstr.c
mogasergiu: Hmm, now that I take a second look at `strsignal.c` it does seem to be quite modified compared to the original. I will come back to this PR and re-write some more parts to look better.

mogasergiu:
Hey @RaduNichita! Thanks for not letting these slip by :muscle:! Did some checkpatch fixes, but I don't think we should change more than that as I want to be able to easily update this source file w.r.t. to musl upstream if it ever comes to it, so I wouldn't change too much of it functionality wise. But, in general, I agree with you, a lot could be refactored to be better here.

flpostolache (approved):
Looks good to me!

RaduNichita (approved):
All good.

razvand (approved):

PR #1154, opened 2023-11-07
lib/posix-time: Allow NULL arg in clock_getres

andreittr:
### Description of changes

This change fixes the behavior of `clock_getres` to not return an error when receiving a NULL as its output argument, in accordance with docs.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

`clock_getres(id, NULL)` should return 0 when `id` is valid.

PR #1262, opened 2024-01-16
lib/vfscore: Auto-mounting scheme with 3 tables

skuenzer:
### Description of changes

This PR introduces a fairly flexible mounting mechanism by
using three filesystem tables:
1) compiled-in, configured via menuconfig
2) user-provided, read entries from kernel parameters ("vfs.fstab")
3) fall-back, similar to "compiled-in"

If auto-mounting is activated, entries of the comiled-in table (1) are mounted first. Afterwards, compiled-in fall-back entries (2) are only processed if the user did not provide any mount entries via the kernel command line (3). The compiled-in tables can be empty and can keep up to four entries.

As part of this PR, The `extract` option of `initrd` replaces the virtual `initrd` filesystem driver. This is done in order to disconnect archive extraction from filesystem mounting and provides greater flexibility for setting up a VFS layout. The traditional initrd mode can be achieved with two entries: First, a RAMFS volume needs to be mounted to `/`. The second entry is used to `extract` virtual fs driver to unpack a CPIO archive to the root volume `/`.

> [!CAUTION]
> This is a breaking change and this PR should be checked in conjunction with [unikraft/kraftkit#1154][0] and merged simultaneously with [#1154][0] to avoid incompatibility issues. Simultaneously, all [catalog](https://github.com/unikraft/catalog) entries should be retriggered against Unikraft's `staging` branch following the merge. Since Unikraft is pre-v1.0, no backwards compatibility is included in this refactor.

[0]: https://github.com/unikraft/unikraft/pull/1154


StefanJum:
@skuenzer just to make sure I'm testing it right, selecting `User-provided filesystem table` -> `Fall-back entries` -> `9pfs` in the config and passing no kernel parameters at runtime should be equivalent to the old vfscore automount feature, right?

skuenzer:
> @skuenzer just to make sure I'm testing it right, selecting `User-provided filesystem table` -> `Fall-back entries` -> `9pfs` in the config and passing no kernel parameters at runtime should be equivalent to the old vfscore automount feature, right?

The full equivalent is configuring 9pfs in the compiled-in table. Then this will be mounted always, irrespective of what the user provides on the kernel command line. The fall-back table is only taken into account if there is nothing provided by the user on the kernel command line.

nderjung (approved):
Adding my review, I have tested this and it works as expected for me. Thanks!

lib/vfscore/automount.c
```diff
@@ -50,18 +50,20 @@
...
-#ifdef CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS
 #include <errno.h>
 #include <uk/config.h>
 #include <uk/arch/types.h>
-#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_ROOTFS */
 
-#if CONFIG_LIBVFSCORE_FSTAB
+#if CONFIG_LIBVFSCORE_AUTOMOUNT_UP
 #define LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP			':'
 #define LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP			','
```
StefanJum: If we do this, compiling with `Compiled-in filesystem table` and without `User-provided filesystem table` config options will fail with `vfscore/automount.c:466:41: error: LIBVFSCORE_FSTAB_UKOPTS_ARGS_SEP’ undeclared`.
Is there any reason to add these in a guard and not just always define them?

StefanJum (changes requested):
See the one comment, all good besides that.

razvand (approved):

lib/vfscore/automount.c
skuenzer: Ouh, good point, there is one tiny mistake then. But you have actually good point, they don't harm if they are unused, no memory will be used. I'll remove this guard. Thanks for spotting this.

skuenzer:
> Reviewed-by: Razvan Deaconescu [razvand@unikraft.io](mailto:razvand@unikraft.io) Approved-by: Razvan Deaconescu [razvand@unikraft.io](mailto:razvand@unikraft.io)

Please note that I added another commit on top because I noticed that ukopts where implicitly deconst'd while parsing. There is a bit more described in the commit subject, however, this change has effect of configuring Unikraft mount options, so I assume it affects maybe KraftKit. please have a look @razvand, @nderjung .

StefanJum (approved):
All good on my side now, thanks.

nderjung (approved):
> Please note that I added another commit on top because I noticed that ukopts where implicitly deconst'd while parsing. There is a bit more described in the commit subject, however, this change has effect of configuring Unikraft mount options, so I assume it affects maybe KraftKit. please have a look @razvand, @nderjung .

I have re-tested the changes with [the accompanying changes to KraftKit](https://github.com/unikraft/kraftkit/pull/1154) and everything still works as expected.

razvand:
I'm testing it myself before the final approval.

razvand (approved):

PR #1277, opened 2024-01-24
lib/syscall_shim: Introduce syscall enter/exit routines tables

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Define two tables containing routines: sysenter and sysexit. These
tables shall be iterated upon on sytem call entry and exit respectively.
By registering into one of these tables, one may be able to have a custom
function called during the entering/exiting of the first called system
call. The order these routines are exected in is dictated by their
priority: lower priority means earlier.


lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -111,3 +86,51 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+	int prsyscalllen;
+
+	UK_ASSERT(sysret_ctx);
+
+	/* Do we have any system call context to base our strace on? If not,
+	 * we were probably called by a native system call instead of a binary
+	 * one. We have separate dedicated debug messages for those, therefore
+	 * ignore.
+	 */
+	if (!sysret_ctx->usc)
```
skuenzer: Can you make this configurable? We need then only one function for native and bincompat but someone should be able to choose if strace should happen just on one of them or both. You could introduce two boolean options, like `CONFIG_LIBSYSCALL_SHIM_STRACE_NATIVE`
`CONFIG_LIBSYSCALL_SHIM_STRACE_BINARY`

skuenzer: The separate debug messages are printed just before a call and are not "pretty printed". In some debugging scenarios, you prefer even to have both messages (one on syscall enter, and another one on syscall exit).

mogasergiu: That's a really good idea! So basically move `__UK_SYSCALL_PRINTD` to `sysrettab`?

michpappas (changes requested):
Adding a note here on the issue on the non-returning nature of `vfork()` / `execve()` that causes an assertion on the refcounts.

lib/syscall_shim/uk_syscall_binary.c
skuenzer: What I meant is that strace on native syscall is something you'd maybe also want to have, so exiting this function earlier under the condition `!sysret_ctx->usc` or `sysret_ctx->usc`, should be configurable. The debug message is only on syscall entrance and not exit and not pretty printed: I see this as an independent configuration.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -111,3 +86,51 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+	 * FIXME:
+	 * We directly use `ukplat_coutk()` until lib/ukdebug printing
+	 * allows us to generate shortened output (avoiding list of details).
+	 */
+	ukplat_coutk(prsyscallbuf, (__sz)prsyscalllen);
+
+	return 0;
+}
+
+uk_sysretcall_prio(uk_binsyscall_strace, 0);
```
skuenzer: I think the prio should be late instead of early, right? As soon as you see the message you assume you return from a syscall.

lib/ukboot/boot.c
michpappas, line 1: Unless I'm missing something this change is obsoleted by #1364.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -90,6 +96,16 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
 	ukplat_coutk(prsyscallbuf, (__sz) prsyscalllen);
 #endif /* CONFIG_LIBSYSCALL_SHIM_STRACE */
 
+	/* Binary system call must always be first in the syscall callchain */
+	UK_ASSERT(uk_syscallchain_count == 1);
+
+	/**
+	 * Run sysret table
+	 */
+	uk_sysrettab_run(&(struct uk_sysrettab_ctx){ .usc = usc, });
```
michpappas: I see a potential issue here with signal handlers. Since these can also make syscalls, we will hit the `uk_syscallchain_count` assertion. On the other hand, moving the decrement before `uk_sysrettab_run()` seems to defeat the purpose of that assertion altogether, as it allows syscalls from every registered callback.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -62,6 +63,11 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
 
 	ukarch_sysregs_switch_uk(&usc->sysregs);
 
+	uk_syscallchain_count++;
+
+	/* Binary system call must always be first in the syscall callchain */
+	UK_ASSERT(uk_syscallchain_count == 1);
```
michpappas: Although we should never make syscalls from uk code, AFAIK right now we have some here and there. So until we remove these, these assertion (also line 100) will break on nested syscalls.

lib/syscall_shim/uk_syscall_binary.c
michpappas: Shouldn't this be:
```c
if (uk_syscallchain_count == 1)
    uk_sysrettab_run();
```

I also see a potential issue here with signal handlers. Since these can also make syscalls, we will hit the `uk_syscallchain_count` assertion. If the assertion goes away, and the above conditional is added, I think it should be okay. If we keep the assertion, one possibility to work around the issue is to move the decrement before `uk_sysrettab_run()`, but that needs to provide clear semantics to the callbacks, that `uk_sysrettab_run()` is executed "after the syscall (chain) is complete", as opposed to "at the end of the syscall".

michpappas (changes requested):
> Adding a note here on the issue on the non-returning nature of vfork() / execve() that causes an assertion on the refcounts.

* clone(): In the a case where [the child uses cl_args->tls as a uktls](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/clone.c#L395) we should reset the syscall count, so sysrettab should register a callback into clonetab. The other case where also `vfork()` falls into allocates a new TLS, so new processes should start with a clear counter (no action needed).
* execve(): Now that `execve()` raises an event, the syscall shim can register a handler to clear the current thread's syscall count.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -339,9 +355,22 @@ typedef long uk_syscall_arg_t;
...
+		uk_syscallchain_count++;				\
+		uk_sysentertab_run(&(struct uk_sysentertab_ctx){	\
+					.execenv = NULL,		\
+				   });					\
+									\
 		__UK_SYSCALL_PRINTD(x, rtype, rname, __VA_ARGS__);	\
 		ret = (long) __##rname(					\
 			UK_ARG_MAPx(x, UK_S_ARG_CAST_ACTUAL, __VA_ARGS__)); \
+									\
+		/* Mark system call exit */				\
```
razvanvirtan: What do we understand by "marking" a system call entry / exit? In my view, running the entry / exit hooks is not marking, but changing the value of `uk_syscallchain_count` is. If this is correct, I think we should move this comment (and the similar ones related to exit marking) right before `uk_syscallchain_count--;`.

mogasergiu:
Rebase force pushed

mogasergiu:
Pushed the commit that optionally throws exception on nested syscalls.

Also pushed three intermediary fixups (in this order):
- s/sysrettab/syscall_entertab + s/sysexittab/syscall_exittab + s/uk_syscallchain_count/uk_syscall_nested_depth (all of whom should show better namespacing)
- uk_syscall_enter_handler + uk_syscall_exit_handler wrappers over the calls to the the syscall hooks tables iterators and the nest depth counter
- one that moves the depth checks to the aforementioned wrappers

It seems that the CI is failing all of the simple helloworld builds, likely because they use syscall.h definitions for syscall registrations, without actually enabling CONFIG_LIBSYSCALL_SHIM, hence the section + TLS variable ending up undefined references.

lib/syscall_shim/Config.uk
```diff
@@ -46,6 +46,17 @@ if LIBSYSCALL_SHIM
 			call and restores it afterwards. This enables the use
 			of different TLS pointers of userland code.
 
+	config LIBSYSCALL_SHIM_ASSERT_NESTED_SYSCALLS
```
michpappas: The discrepancy between the config name and the action made me think, should that be a crash or an assertion?

I expect that the current default is a transient state and eventually we will be changing that to `default y`. While we're there, `default n` is redundant: https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults.

Also wondering if reasoning should go to the commit message, and help should remain more factual (limit the description to what this does), so that we don't need to keep in sync with policy.

lib/syscall_shim/Makefile.uk
```diff
@@ -90,3 +90,5 @@ LIBSYSCALL_SHIM_SRCS-$(CONFIG_LIBSYSCALL_SHIM_HANDLER) += $(LIBSYSCALL_SHIM_BASE
 
 LIBSYSCALL_SHIM_SRCS-y += $(LIBSYSCALL_SHIM_BASE)/uk_prsyscall.c
 LIBSYSCALL_SHIM_SRCS-y += $(LIBSYSCALL_SHIM_BASE)/vars.c
+LIBSYSCALL_SHIM_SRCS-y += $(LIBSYSCALL_SHIM_BASE)/syscall_exittab.lds.S
```
michpappas, line 105: I'm thinking whether these should be two separate files (also for the headers) or whether they should go under a common `syscalltab` umbrella. If you see this as two ways to achieve the same thing let's leave it as is.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -116,6 +116,12 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
 		    execenv->regs.__syscall_rarg1);
 #endif /* CONFIG_LIBSYSCALL_SHIM_DEBUG_HANDLER */
 
+	/* Mark system call entry */
```
michpappas: I'm clearly missing something here... What is the reason we don't call the wrapper functions (and hence the check for nested syscalls) in binary syscalls?

lib/syscall_shim/syscall_entertab.lds.S
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/arch/limits.h> /* for __PAGE_SIZE */
+#include <uk/plat/common/common.lds.h> /* for UK_SEGMENT_DATA */
+
+SECTIONS
+{
+	. = ALIGN(__PAGE_SIZE);
```
michpappas: Should that be `__PAGE_SIZE` or would 8 suffice? (e.g. `uklibparam`, `earlytab`). Same question for `exittab`.

lib/syscall_shim/syscall_entertab.lds.S
```diff
@@ -0,0 +1,21 @@
...
+
+	.uk_syscall_entertab :
+	{
+		PROVIDE(uk_syscall_entertab_start = .);
+		KEEP(*(SORT_BY_NAME(.uk_syscall_entertab[0-9])))
+		PROVIDE(uk_syscall_entertab_end = .);
+	} UK_SEGMENT_DATA
```
michpappas: Interestingly, except from `libukreloc`, we don't annotate the rest of the library-defined sections with `SEGMENT_DATA`. Same question for `exittab`.

lib/syscall_shim/include/uk/syscall_exittab.h
```diff
@@ -0,0 +1,104 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_SYSEXITTAB_H__
+#define __UK_SYSEXITTAB_H__
+
+#include <uk/assert.h>
+
+struct uk_syscall_exit_ctx {
```
michpappas, line 13: If Simon was reviewing he would had said that we should pass `struct ukarch_execenv *` directly. I don't have a strong opinion.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -147,6 +153,12 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
 #endif /* !CONFIG_LIBUKCONSOLE */
 #endif /* CONFIG_LIBSYSCALL_SHIM_STRACE */
 
+	/* Mark system call exit */
+	uk_syscall_exittab_run(&(struct uk_syscall_exit_ctx){
+				.execenv = execenv,
+			  });
+	uk_syscall_nested_depth--;
```
michpappas: Since the counter is in the TLS, any syscalls that clear the TLS of the current thread, (right now only `execve()` comes to mind) will end up with a negative value here. You need to register a handler for `POSIX_PROCESS_EXECVE_EVENT` to set the counter to the right value.

Other syscalls that may wipe the TLS as a result of `clone()` would do that on the child, so I think they're out of context.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -44,14 +44,53 @@
...
+{
+	/* Mark system call entry */
+	uk_syscall_nested_depth++;
+
+#if CONFIG_LIBSYSCALL_SHIM_ASSERT_NESTED_SYSCALLS
+	if (unlikely(uk_syscall_nested_depth != 1))
+		UK_CRASH("Nested system call detected. Depth: 0x%02x\n",
+			 uk_syscall_nested_depth);
+#else /* !CONFIG_LIBSYSCALL_SHIM_ASSERT_NESTED_SYSCALLS */
+	/* Only run this if the first system call in the callchain */
```
michpappas: We should provide some reasoning on why we do that, in a comment or in the commit message (or both).

michpappas (changes requested):

michpappas:
Besides the comments above, IIRC we had also moved strace logs to syscalltab. Was this removed from this PR, or is it something we didn't do in the end?

michpappas:
> Pushed the commit that optionally throws exception on nested syscalls.
> 
> Also pushed three intermediary fixups (in this order):
> 
>     * s/sysrettab/syscall_entertab + s/sysexittab/syscall_exittab + s/uk_syscallchain_count/uk_syscall_nested_depth (all of whom should show better namespacing)
> 
>     * uk_syscall_enter_handler + uk_syscall_exit_handler wrappers over the calls to the the syscall hooks tables iterators and the nest depth counter
> 
>     * one that moves the depth checks to the aforementioned wrappers
> 
> 
> It seems that the CI is failing all of the simple helloworld builds, likely because they use syscall.h definitions for syscall registrations, without actually enabling CONFIG_LIBSYSCALL_SHIM, hence the section + TLS variable ending up undefined references.

Regarding the broken tests: Maybe you can open a PR that adds the missing option, that @razvand helps us fasttrack so that we can see these pass?

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -267,9 +270,22 @@ typedef long uk_syscall_arg_t;
 	{								\
 		long ret;						\
 									\
+		/* Mark system call entry */				\
+		uk_syscallchain_count++;				\
+		uk_sysentertab_run(&(struct uk_sysentertab_ctx){	\
```
michpappas: Consider if you prefer `uk_sysentrytab_*` / `uk_sysexittab_*` or `syscalltab_entry_*` /  `sycalltab_exit_*`. Very subjective though, and I'm not the maintainer, so feel free to leave as is if you prefer the current name.

lib/syscall_shim/Config.uk
mogasergiu: > I expect that the current default is a transient state and eventually we will be changing that to default y

I myself am expecting it to stay as `default n` so that there isn't an unnecessary check on every system call entry.

> Also wondering if reasoning should go to the commit message, and help should remain more factual (limit the description to what this does), so that we don't need to keep in sync with policy.

Fair, I will move the last sentence to the commit message

lib/syscall_shim/Makefile.uk
mogasergiu: I don't have a preference either. Let's see what the others say.

lib/syscall_shim/include/uk/syscall_exittab.h
mogasergiu: My intention was to provide a structure that would allow for easy addition of other fields/flags in the future.

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: hmm looks like i forgot to replace this place with the new wrapper. Thanks for catching that!

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: The wrappers decrement before running the table, so this misplacement here disappears.

EDIT: Nevermind, I forgot that vfork/execve got merged. So it's on me to add that handler now. :+1:

lib/syscall_shim/include/uk/syscall.h
mogasergiu: Yes this is the old version, I renamed in the first fixup to uk_syscall_entertab/exittab

mogasergiu:
> > Pushed the commit that optionally throws exception on nested syscalls.
> > Also pushed three intermediary fixups (in this order):
> > ```
> > * s/sysrettab/syscall_entertab + s/sysexittab/syscall_exittab + s/uk_syscallchain_count/uk_syscall_nested_depth (all of whom should show better namespacing)
> > 
> > * uk_syscall_enter_handler + uk_syscall_exit_handler wrappers over the calls to the the syscall hooks tables iterators and the nest depth counter
> > 
> > * one that moves the depth checks to the aforementioned wrappers
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems that the CI is failing all of the simple helloworld builds, likely because they use syscall.h definitions for syscall registrations, without actually enabling CONFIG_LIBSYSCALL_SHIM, hence the section + TLS variable ending up undefined references.
> 
> Regarding the broken tests: Maybe you can open a PR that adds the missing option, that @razvand helps us fasttrack so that we can see these pass?

Question is do we want to have syscall shim enabled everywhere, even on a hellworld?

EDIT: Nevermind, figured out a simpler way :sweat_smile: :
```diff
diff --git a/lib/syscall_shim/include/uk/syscall.h b/lib/syscall_shim/include/uk/syscall.h
index b1cbe46e..b5ad3ec5 100644
--- a/lib/syscall_shim/include/uk/syscall.h
+++ b/lib/syscall_shim/include/uk/syscall.h
@@ -55,6 +55,7 @@
 extern "C" {
 #endif
 
+#if CONFIG_LIBSYSCALL_SHIM
 static inline void uk_syscall_enter_handler(struct uk_syscall_enter_ctx *sctx)
 {
     /* Mark system call entry */
@@ -100,6 +101,12 @@ static inline void uk_syscall_exit_handler(struct uk_syscall_exit_ctx *sctx)
 
     uk_syscall_exittab_run(sctx);
 }
+#else /* !CONFIG_LIBSYSCALL_SHIM */
+static inline
+void uk_syscall_enter_handler(struct uk_syscall_enter_ctx *sctx __unused) {}
+static inline
+void uk_syscall_exit_handler(struct uk_syscall_exit_ctx *sctx __unused) {}
+#endif /* !CONFIG_LIBSYSCALL_SHIM */
 
 /*
  * Whenever the hidden Config.uk option LIBSYSCALL_SHIM_NOWRAPPER
  ```

**Comment by mogasergiu** [comment]
> Besides the comments above, IIRC we had also moved strace logs to syscalltab. Was this removed from this PR, or is it something we didn't do in the end?

The first pushed version of this PR had it. (Mentioned [here](<https://github.com/unikraft/unikraft/pull/1277#discussion_r1466424590>) as well)
But then we decided to not have it anymore as it introduced a check on every syscall to make sure it is not called on native system calls as well. For native system calls to be able to use this strace handler we would need to do the equivalent of the SYSCALL_PRINTD macro's in the strace handler which i am not sure how doable is since it is a macro that relies on on things like variadic arguments and token pasting.

However, in retrospective, this is going to be used for debugging only, so if it is fine with everyone, I think I can re-add it together with the check and have it enabled only if the binary syscall handler is enabled.

File: lib/syscall_shim/syscall_entertab.lds.S (hunk 10 above)
Comment (mogasergiu) [reply lib/syscall_shim/syscall_entertab.lds.S]: Yes, right, this one does not need that. .ukreloc for some reason needs that though. I had to fight some linker script shenanigans related to section order and their mapping to segments.... .ukreloc and the placement of the other relocation sections is not really conventional.

(That being said, I copy pasted this from there initially :laughing:)

**Comment by mogasergiu** [comment]
> Besides the comments above, IIRC we had also moved strace logs to syscalltab. Was this removed from this PR, or is it something we didn't do in the end?

I also moved the debug handler to the enter tab. There is an important NOTE comment in both this and the strace handler commit messages though. Might have been the reason why I removed it in the first place. However I don't know any place that would directly call clone or execve in the kernel or other syscall registered with `UK_LLSYSCALL_R_E_DEFINE`? Though this might be a problem (aesthetically speaking) if elfloader or any such native program calls these directly as they will appear in the strace... on the bright side most straces begin with execve anyway :-).

File: lib/syscall_shim/include/uk/syscall_exittab.h (hunk 11 above)
Comment (michpappas) [reply lib/syscall_shim/include/uk/syscall_exittab.h:13]: Agreed, I was referring to this discussion: https://github.com/unikraft/unikraft/pull/1472#discussion_r1700353006. We can keep it like that.

File: lib/syscall_shim/Config.uk [hunk 15]
```diff
@@ -46,6 +46,15 @@ if LIBSYSCALL_SHIM
 			call and restores it afterwards. This enables the use
 			of different TLS pointers of userland code.
 
+	config LIBSYSCALL_SHIM_CRASH_NESTED_SYSCALLS
```
michpappas: I am now having doubts on whether we should be adding this, perhaps there is a better way to catch such cases through the build, e.g. by adding compiler instrumentation to detecting calls of  `__uk_syscall_` symbols from kernel libraries.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -122,7 +114,33 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+					.execenv = execenv,
+				});
+
+#if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
+	t->tlsp = ukarch_sysctx_get_tlsp(&execenv->sysctx);
+#endif /* CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS */
+}
+
 #if CONFIG_LIBSYSCALL_SHIM_STRACE
+static void binary_syscall_strace(struct uk_syscall_exit_ctx *exit_ctx)
```
michpappas, line 148: For both this and the strace callback: Maybe it's better to move them to a separate file, `debug.c`? The whole point of moving them to callbacks is cleaning up the code, and while it is cleaner to remove them from the function, moving them away from this file will reduce visual noise even more.

lib/syscall_shim/vars.c
```diff
@@ -1,2 +1,21 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
 #include <uk/syscall.h>
 
+__thread unsigned long uk_syscall_nested_depth;
```
michpappas, line 9: You're making a bold assumption here that the thread that handles the event is the same one that issued the event. While given the current implementation that may be true, for forward compatibility it's safer that instead you take the thread from the parameters passed to the handler. Also the return value is not correct, it must be one of the values defined in `event.h` (in this case `UK_EVENT_HANDLED_CONT` to allow other handles of that event execute). Right now it returns zero, which is assigned to `UK_EVENT_NOT_HANDLED`. Last but not least, this was missing including `event.h` and `prio.h` the handler was not being registered at all. Fixed version below. I also added a comment to explain the semantics of this handler as  it's a bit tricky.

```diff
diff --git a/lib/syscall_shim/vars.c b/lib/syscall_shim/vars.c
index a45226926..dc0a1eb26 100644
--- a/lib/syscall_shim/vars.c
+++ b/lib/syscall_shim/vars.c
@@ -4,15 +4,42 @@
  * You may not use this file except in compliance with the License.
  */
 
-#include <uk/syscall.h>
+#include <uk/event.h>
+#include <uk/prio.h>
+#include <uk/process.h>
+#include <uk/syscall.h>
+#include <uk/thread.h>
 
 __thread unsigned long uk_syscall_nested_depth;
 
 #if CONFIG_LIBPOSIX_PROCESS_EXECVE
-static int syscall_nested_depth_reset(void *data __unused)
+static int syscall_nested_depth_reset(void *data)
 {
-       uk_syscall_nested_depth = 0;
-       return 0;
+       struct posix_process_execve_event_data *event_data;
+
+       UK_ASSERT(data);
+
+       event_data = (struct posix_process_execve_event_data *)data;
+
+       UK_ASSERT(event_data->thread);
+
+       /* execve() will always enter ukplat_syscall_handler() with a
+        * sane depth value, that is:
+        *
+        * - when preceded by a vfork() we start off with a new uktls,
+        *   and hence enter ukplat_sycall_handler() with zero depth.
+        * - when NOT preceded by a vfork(), the depth should have a
+        *   sane value from the previous syscall, that is we also enter
+        *   uk_syscall_handler() with a depth of zero.
+        *
+        * Nevertheless, since execve() does not go through the syscall return
+        * path we must reset its counter back to zero here so that the
+        * next syscall starts off with a sane value.
+        */
+       uk_thread_uktls_var(event_data->thread, uk_syscall_nested_depth) = 0;
+
+       return UK_EVENT_HANDLED_CONT;
 }
 
 UK_EVENT_HANDLER_PRIO(POSIX_PROCESS_EXECVE_EVENT,
```

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -380,20 +432,30 @@ typedef long uk_syscall_arg_t;
...
 						      x, UK_S_ARG_ACTUAL,\
 						      __VA_ARGS__));	\
-	long __used rname(long _execenv)					\
+	long __used rname(long _execenv)				\
 	{								\
 		struct ukarch_execenv *execenv;				\
 		long ret;						\
 									\
 		execenv = (struct ukarch_execenv *)_execenv;		\
+									\
```
michpappas, line 479: There's something buggy in the R_E variant, as the handlers are executed twice: Once upon entry to `ukplat_syscall_handler()` and once again upon execution of the actual syscall.

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -44,14 +44,48 @@
...
 #endif
 
+#if CONFIG_LIBSYSCALL_SHIM
+static inline void uk_syscall_enter_handler(struct uk_syscall_enter_ctx *sctx)
+{
+	/* Mark system call entry */
+	uk_syscall_nested_depth++;
+
+	/* Only run this if the first system call in the callchain */
+	if (uk_syscall_nested_depth != 1)
```
michpappas: I'm thinking whether it's more flexible to pass the nested depth as a parameter to the handlers and let them decide whether they should be running on nested syscalls or not.

lib/syscall_shim/include/uk/syscall_exittab.h
```diff
@@ -0,0 +1,104 @@
...
+
+	uk_pr_debug("Sycall exit table @ %p - %p\n",
+		    &uk_syscall_exittab_start[0], &uk_syscall_exittab_end);
+
+	uk_syscall_exittab_foreach(entry,
+				   uk_syscall_exittab_start,
+				   uk_syscall_exittab_end) {
+		UK_ASSERT(entry);
+
+		if (!entry->func)
```
michpappas, line 138: I would recommend that you add a comment on why you're returning here, or even better a comment in the declaration of `struct uk_syscall_*_ctx` that `execenv` can be `NULL` (and under which circumstances), as all handlers need to be aware of that.

michpappas (changes requested):
@mogasergiu hello from 37,040 ft! ✈️ 😆 I'm leaving another round of comments, hope I did not forget anything.

michpappas:
> @mogasergiu hello from 37,040 ft! ✈️ 😆 I'm leaving another round of comments, hope I did not forget anything.

I did actually: You need to include `uk/events.h` and `uk/prio.h` where you define the debug / strace callbacks.

lib/syscall_shim/include/uk/syscall_entertab.h
```diff
@@ -0,0 +1,104 @@
...
+#ifndef __UK_SYSENTERTAB_H__
+#define __UK_SYSENTERTAB_H__
+
+#include <uk/assert.h>
+
+struct uk_syscall_enter_ctx {
+	struct ukarch_execenv *execenv;
+};
+
+typedef void (*uk_syscall_entertab_func_t)(struct uk_syscall_enter_ctx *);
```
michpappas, line 62: Should we be returning a value in the entry handler in case we want to catch errors, similar to inittab?

lib/syscall_shim/vars.c
michpappas: Also please consider moving the handler to a different file, `execve.c` will do.

michpappas:
@mogasergiu please notice that I updated **https://github.com/unikraft/unikraft/pull/1277#discussion_r1935918241**

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: I kind of still feel like it makes more sense to have them in uk_syscall_binary.c, as they only exist for binary syscalls. If I were to add them to debug.c I would also have to guard them with #if, or I may as well name it uk_syscall_binary_debug.c instead.

lib/syscall_shim/include/uk/syscall.h
mogasergiu: Yes I am thinking about extending uk_syscall_ctx to include it as well as an additional flag to help handlers tell whether they are running in a native or binary syscall.

lib/syscall_shim/include/uk/syscall.h
mogasergiu: Yes they should be moved to the ename variants.

lib/syscall_shim/include/uk/syscall_entertab.h
```diff
@@ -0,0 +1,151 @@
...
+	     (itr) < &(tab_end);					\
+	     (itr)++)
+
+/**
+ * Register a Unikraft system call enter function.
+ *
+ * @param fn
+ *   System call enter function to be called
+ * @param prio
+ *   Priority level (0 (earliest) to 9 (latest)), must be a constant.
```
michpappas: If you phrase it as below it's more self-explanatory 😉 
```
 *   Priority level (`UK_PRIO_ EARLIEST()` to `UK_PRIO_LATEST()`). Use the `UK_PRIO_AFTER()` ...
```

lib/syscall_shim/vars.c
```diff
@@ -1,2 +1,11 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/event.h>
```
michpappas: `uk/event.h` and `uk/prio.h` are no longer needed here.

lib/syscall_shim/uk_syscall_binary.c
michpappas: Well, hopefully in the future we will re-use them for the native case. also guarding will happen in the `Makefile.uk` so no need for preprocessor guards. The motivation was that moving these away from `uk_syscall_binary.c` makes that file much cleaner and limited to its actual scope. But okay, that's minor, so if you think otherwise we can deal with it once we support native syscalls.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -132,7 +124,31 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+#endif /* !CONFIG_LIBSYSCALL_SHIM_STRACE_ANSI_COLOR */
+	int prsyscalllen __maybe_unused;
+	struct ukarch_execenv *execenv;
+
+	UK_ASSERT(exit_ctx);
+
+	if (!(exit_ctx->flags & UK_SYSCALL_EXIT_CTX_BINARY_SYSCALL))
+		return;
+
+	UK_ASSERT(exit_ctx->nested_depth == 1);
```
michpappas: This one worries me a bit because we have not established that other handlers (e.g. signal delivery) should save / restore the counter upon context switch (IMO such a requirement shall be clearly documented as the contract between the syscalltab and its handlers). Taking such a decision is tricky cause we can no longer distinguish whether a syscall is nested in the hanldler unless you maintain a counter backup in the TLS (even worse, since we can have arbitrarily long nesting, at least on signals, that would require also multiple backup levels). And that is of a problem: if your handlers are not  re-entrant, as they can no longer take a decision to immediately return on nested syscalls by inspecting the counter. Ditto for the other handler.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -132,7 +124,31 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+	char prsyscallbuf[512]; /* ANSI color is pretty hungry */
+#else /* !CONFIG_LIBSYSCALL_SHIM_STRACE_ANSI_COLOR */
+	char prsyscallbuf[256];
+#endif /* !CONFIG_LIBSYSCALL_SHIM_STRACE_ANSI_COLOR */
+	int prsyscalllen __maybe_unused;
+	struct ukarch_execenv *execenv;
+
+	UK_ASSERT(exit_ctx);
+
+	if (!(exit_ctx->flags & UK_SYSCALL_EXIT_CTX_BINARY_SYSCALL))
```
michpappas, line 160: Since we now eliminated all kernel-origninating syscalls, would it make more sense to add an assertion in the binary part of the syscalltab wrappers? Besides, that handler cannot called by anything that doesn't pass an execenv (would be worth to add an assertion for that instead) so, unless I'm missing something, atm it can only be registered by binary syscalls anyway. Ditto for the other handler.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -155,13 +171,11 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
 	uk_pr_info(prsyscallbuf);
 #endif /* !CONFIG_LIBUKCONSOLE */
-#endif /* CONFIG_LIBSYSCALL_SHIM_STRACE */
-
-#if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
-	t->tlsp = ukarch_sysctx_get_tlsp(&execenv->sysctx);
-#endif /* CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS */
 }
+
+uk_syscall_exittab_prio(binary_syscall_strace, UK_PRIO_LATEST);
```
michpappas, line 193: `#include <uk/prio.h>` for `UK_PRIO_LATEST` 👈🏼

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -289,7 +290,7 @@ typedef long uk_syscall_arg_t;
  * UK_LLSYSCALL_DEFINE()
  * Low-level variant, does not provide a libc-style wrapper
  */
-#define __UK_LLSYSCALL_DEFINE(x, rtype, name, ename, rname, ...)	\
+#define __UK_LLSYSCALL_DEFINE(x, rtype, name, ename, rname, doname, ...)\
```
michpappas, line 293: Super nit: Since all other variables use a single letter consider if this should be `dname` and `SYSCALLD` (it quite stands out in all changes in this commit).

michpappas (changes requested):

mogasergiu:
Rebase fpush

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: We haven't yet, sadly :(. There are still those in vfscore and ramfs that @andreittr said he will take care of in his VFS rework. :muscle:

lib/syscall_shim/include/uk/syscall.h
mogasergiu: Yes, this is more of a personal taste, I like `do` as it is more suggestive. Just `d` could mean a lot of things :flushed:. Unless you insist on changing this, I would mark this as resolved.

michpappas (approved):

PR #1278, opened 2024-01-24
lib/posix-fdtab: Support independent (per-process) fdtabs

andreittr:
### Description of changes

This PR changes the way `posix-fdtab` handles tables, from a single, static, OS-wide table, to a thread-local reference to one of possibly many independent tables. This allows for, among others, per-process independent fdtabs. Specifically, the commits in this changeset:
- Make the currently active fdtab a thread-local reference
- Support dynamically allocated (and freed) fdtabs
- Handle the `clone()` syscall to duplicate the fdtable when needed

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

`CONFIG_LIBPOSIX_FDTAB=y`

All single-process code should keep working as-is, with a shared fdtab (albeit through independent, refcounted references).
Once multi-process support is added, every thread spawned with `clone()` without the `CLONE_FILES` flag should have independent tables.

### DIFF
--- a/lib/posix-fdtab/fdtab.c
+++ b/lib/posix-fdtab/fdtab.c
@@ -27,6 +27,10 @@
 #include <uk/posix-fdtab-legacy.h>
 #endif /* CONFIG_LIBPOSIX_FDTAB_LEGACY_SHIM */
 
+#if CONFIG_LIBPOSIX_PROCESS_CLONE
+#include <uk/process.h>
+#endif /* CONFIG_LIBPOSIX_PROCESS_CLONE */
+
 #if CONFIG_LIBPOSIX_PROCESS_EXECVE
 #include <uk/event.h>
 #include <uk/prio.h>
@@ -503,6 +507,38 @@ static void fdtab_thread_term(struct uk_thread *child)
 
 UK_THREAD_INIT(fdtab_thread_init, fdtab_thread_term);
 
+#if CONFIG_LIBPOSIX_PROCESS_CLONE
+static int fdtab_clone(const struct clone_args *cl_args,
+		       size_t cl_args_len __unused,
+		       struct uk_thread *child,
+		       struct uk_thread *parent)
+{
+	struct uk_fdtab *tab = uk_thread_uktls_var(parent, active_fdtab);
+	struct uk_fdtab *newtab;
+
+	UK_ASSERT(tab); /* Do not call clone from raw threads */
+	if (cl_args->flags & CLONE_FILES) {
+		/* Inherit parent's fdtab */
+		/* As a compat stop-gap, the raw thread already inherited the
+		 * parent's fdtab ref; we don't need to do anything.
+		 *
+		 * TODO: move inheritance here once stopgap is removed.
+		 */
+		UK_ASSERT(uk_thread_uktls_var(child, active_fdtab) == tab);
+		return 0;
+	} else {
+		/* Duplicate parent's fdtab */
+		newtab = fdtab_duplicate(tab);
+		if (unlikely(!newtab))
+			return -ENOMEM;
+	}
+	uk_thread_uktls_var(child, active_fdtab) = newtab;
+	return 0;
+}
+
+UK_POSIX_CLONE_HANDLER(CLONE_FILES, 0, fdtab_clone, 0);
+
+#endif /* CONFIG_LIBPOSIX_PROCESS_CLONE */
 #endif /* CONFIG_LIBPOSIX_FDTAB_MULTITAB */
 
 /* Init fdtab as early as possible, to enable functions that rely on fds */
--- a/lib/posix-fdtab/fdtab.c
+++ b/lib/posix-fdtab/fdtab.c
@@ -88,12 +88,6 @@ static int init_posix_fdtab(struct uk_init_ctx *ictx __unused)
 	return 0;
 }
 
-/* TODO: Adapt when multiple processes are supported */
-static inline struct uk_fdtab *_active_tab(void)
-{
-	return active_fdtab;
-}
-
 /* Encode flags in entry pointer using the least significant bits */
 /* made available by the open file structure's alignment */
 struct fdval {
@@ -195,7 +189,6 @@ void file_rel(struct uk_fdtab *tab, void *p, int flags __maybe_unused)
 
 int uk_fdtab_open(const struct uk_file *f, unsigned int mode)
 {
-	struct uk_fdtab *tab;
 	struct uk_ofile *of;
 	int flags;
 	const void *entry;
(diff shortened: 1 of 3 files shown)


andreittr:
Rebased to current staging.

John-Ted (approved):
Looks good to me

csvancea (approved):
Reviewed-by: Cosmin Vancea csvancea@gmail.com

andreittr:
Rebased to current staging

lib/posix-fdtab/fdtab.c
```diff
@@ -432,6 +437,63 @@ static void fdtab_thread_term(struct uk_thread *child)
...
+		       size_t cl_args_len __unused,
+		       struct uk_thread *child,
+		       struct uk_thread *parent __unused)
+{
+	if (!(cl_args->flags & CLONE_FILES)) {
+		struct uk_fdtab *tab = uk_thread_uktls_var(child, active_fdtab);
+		struct uk_fdtab *newtab = fdtab_duplicate(tab);
+
+		if (unlikely(!newtab))
+			return -ENOMEM;
```
michpappas, line 517: Very minor: In case `fdtab_duplicate()` fails for additional reasons in the future, instead of making assumptions on  the caller's side, consider if it would be preferable to update `fdtab_duplicate()` to return error pointer that the caller can directly decode and return. Alternatively you can consider adding a line in the description "Returns pointer on success or NULL if out of memory".

lib/posix-fdtab/fdtab.c
```diff
@@ -432,6 +437,63 @@ static void fdtab_thread_term(struct uk_thread *child)
...
+ *
+ * NOTE: This is a basic implementation that does not guarantee atomicity of the
+ * clone operation; we optimistically assume this will not break anything.
+ * Please revisit if this turns out to be false.
+ */
+static struct uk_fdtab *fdtab_duplicate(struct uk_fdtab *tab)
+{
+	struct uk_fdtab *ret = uk_malloc(tab->alloc, sizeof(*ret));
+
+	if (likely(ret)) {
```
michpappas: Generally speaking, the convention in the codebase is to wrap checks for errors with `unlikely` and return immediately, and not use `likely` in general. The reason is more than style: If a block wrapped around `likely()` calls an inline function that involves error handling, that function will almost certainly use `unlikely`, and that results into multiple likelihood attributes in the same execution path, which can result into the [opposite than the desired effect](https://blog.aaronballman.com/2020/08/dont-use-the-likely-or-unlikely-attributes/).

lib/posix-fdtab/fdtab.c
```diff
@@ -432,6 +437,63 @@ static void fdtab_thread_term(struct uk_thread *child)
...
+		if (unlikely(!newtab))
+			return -ENOMEM;
+		if (uk_refcount_release(&tab->refcnt))
+			fdtab_free(tab);
+		uk_thread_uktls_var(child, active_fdtab) = newtab;
+	}
+	return 0;
+}
+
+UK_POSIX_CLONE_HANDLER(CLONE_FILES, 0, fdtab_clone, 0);
```
michpappas, line 523: Not necessarily in the scope of this PR, but from the documentation for `CLONE_FILES` in the man page, it looks like we need to update the `fdtab_handle_execve()` to handle the case where the fdtab is shared between the parent and the child:
> If a process sharing a file descriptor table calls execve(2), its file descriptor table is duplicated (unshared).

lib/posix-fdtab/fmap.h
```diff
@@ -184,6 +184,23 @@ static inline void uk_fmap_init(const struct uk_fmap *m)
...
 }
 
+/**
+ * Sets the value at `idx` to `p`.
+ *
+ * WARNING: This function is not thread-safe, take care when calling.
+ */
+static inline void uk_fmap_set(const struct uk_fmap *m, int idx, const void *p)
+{
+	if (!_FMAP_INRANGE(m, idx))
```
michpappas, line 196: Just curious cause I notice that you make the same check in other functions in this header: is the out-of-range an expected case that we can safely continue execution from, and therefore not qualifying for an assertion?

michpappas (changes requested):

lib/posix-fdtab/fdtab.c
andreittr: I also tend to use `unlikely` in general and keep `likely` for only this exact pattern: "call, likely init, ret", as that's "safe" from unintended effects in isolation. But I hadn't considered the effects of inlining or LTO before :thinking: good point; will replace with unlikely.

lib/posix-fdtab/fdtab.c
andreittr: I agree we have to update fdtab for `execve()`, and more generally make "fdtab unsharing" a singular well-defined event that can be triggered either by `clone()` or `execve()`, and can in turn trigger further actions.
I believe we need this in order to correctly support "magic" files that are unshared implicitly along with the fdtab, without explicit action from userspace, such as `epollfd` and `signalfd`.

lib/posix-fdtab/fmap.h
andreittr: :thinking: fdtab ops/syscalls _should_ return -EBADF if called with an out-of-bounds fd, so that's explicitly a use-case we need to safely continue from, and from looking at the code it seems fdtab relies on fmap to signal errors instead of bounds-checking at higher levels.

I see your point that a cleaner design would do bounds checking & `return -EBADF` in the fdtab ops/syscalls themselves and have the raw internal data structure at most assert bounds -- separation of mechanism and policy and all that. While such a refactor is not something I'm comfortable tacking on to this PR, I think the broader VFS work (i.e., now-ish) is a good time to do this.

lib/posix-fdtab/fdtab.c
andreittr: I see your point, though I prefer that functions who have a single possible failure state by design do not return error codes, but let the caller in higher levels interpret their failure. Returning errptrs is particularly tricky because then the caller must worry _both_ about errptrs _and_ NULLs (because NULL is _not_ an error pointer as it doesn't map to a negative errno).

`fdtab_duplicate()` falls into this single fail state group, as it's a glorified malloc + memcpy by design (and the latter cannot fail recoverably); `fdtab_clone` is the higher-level "public-facing" function that is expected to conform to the errno API, and there is IMO the best place to encode and return errors.
I'll update description to make its return values clear.

lib/posix-fdtab/fmap.h
michpappas: Yes of course it's out of scope. Just a general observation as I was going through the code 👍🏼

lib/posix-fdtab/fdtab.c
michpappas: Alright, thanks for confirming. Let's treat it as a follow up item then 👍🏼

lib/posix-fdtab/fdtab.c
michpappas: Up to you, especially as its not a public API it's not that critical. Just sharing my 2c.

andreittr:
Rebased to current staging

andreittr:
Addressed comments

andreittr:
Rebased to current staging

andreittr:
Pushed a minor rework; in short:
- single fdtab or multi-fdtab is now a hidden Kconfig option, autoselected if `clone()` support is enabled
  - this neatly decouples the mechanism of "having multiple fdtabs" from the policy of "what fdtab each thread has"
  - also: if no `clone()`, no indirection => less overhead for single-thread apps
- newly-created threads no longer inherit their parent's fdtab reference; raw `uk_thread`s get NULL (kernel threads shouldn't be mucking about with the fdtab anyway), `clone` handler implements the semantics of sharing/unsharing for posix threads
- the init thread still starts with a reference to the init fdtab; this is in lieu of a "thread posixification" callback, and once such a mechanism exists, we should probably move the init fdtab there

michpappas (approved):

andreittr:
Pushed a compat stop-gap change to have all raw threads inherit the fdtab ref of their parent (instead of NULL).

andreittr:
Pushed a tiny bonus fix for the wrong value being returned by the `execve` event handler.

andreittr:
Pushed a fix for a type mismatch that resulted in misaligned accesses on arm64.

michpappas (approved):

michpappas:

PR #1279, opened 2024-01-24
lib/posix-socket: Extend socket events

marcrittinghaus:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

- `CONFIG_LIBPOSIX_SOCKET_EVENTS`

### Description of changes

This commit extends the socket events so that each socket keeps track of its current state. This allows event handlers to determine when a socket function is called multiple times in a row (e.g., nginx calls listen twice) and, for what type of socket a close is performed (e.g., listen socket).

As part of this extension, the single socket event is split up into dedicated event objects per event. This avoids calling the hook for events that are not of interest.

### DIFF
--- a/lib/posix-socket/events.h
+++ b/lib/posix-socket/events.h
@@ -14,11 +14,14 @@
 #include <uk/assert.h>
 #include <uk/essentials.h>
 
+enum uk_socket_state;
+
 struct uk_socket_event_data {
 	uintptr_t id;
 	int family;
 	int type;
 	int protocol;
+	enum uk_socket_state state;
 	/* NOTE: We use `struct sockaddr_storage` instead of `struct sockaddr *`
 	 *       to have statically allocated memory of enough space to store
 	 *       any socket address without calling malloc.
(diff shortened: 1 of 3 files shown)


lib/posix-socket/events.h
```diff
@@ -126,12 +129,13 @@ static inline void uk_socket_evd_raddr_set_from(struct uk_socket_event_data
...
-static inline void uk_socket_event_raise(struct uk_socket_event_data *evd,
-					 enum uk_socket_event_type event)
+static inline void socket_event_raise(struct uk_socket_event_data *evd,
+				      enum uk_socket_state new_state,
+				      struct uk_event *evt)
 {
 	struct uk_socket_event emsg = {
 		.id        = evd->id,
-		.event     = event,
+		.state     = evd->state,
```
skuenzer, line 140: This new field and the new way how we use it, requires initialization in `uk_socket_evd_init()` **and** `uk_socket_evd_init_from()`
```diff
+       evd->state      = UK_SOCKET_STATE_CLOSE;
```

Can you add this line to both functions? Without it, I saw some garbage happening for freshly opened/created sockets when they emit an state change event the first time.

marcrittinghaus: Good catch 🙏🏻 ! Thanks, I'll fix it

skuenzer (approved):
Great, works now as we discussed. Thanks a lot!

PR #1282, opened 2024-01-25
lib/vfscore: Additional mount options

skuenzer:
This PR consists of two additions regarding automatically mounting. One addition is a `ramfs` mount option to the simulated `extract` filesystem driver. It can be set to one of the following values:
* 0: Do not mount an underlying ramfs
     (same behavior as if `ramfs` option is absent)
* 1: Mount an underlying ramfs volume before extraction
     (default if present without value)
* 2: Only mount underlying ramfs volume if nothing else is mounted

The second addition extends the Unikraft mount options with `ifinitrd0` and `ifnoinitrd0`.These options make an fstab entry conditional to the case if an initrd module exists and was loaded by the bootloader or virtual machine monitor (VMM).
This can be used to hide or only activate certain fstab entries on the presence or absence of an initrd.


lib/ukargparse/argparse.c
```diff
@@ -139,3 +140,56 @@ int uk_argnparse(char *argb, __sz maxlen, char *argv[], int maxcount)
...
+
+char *uk_nextarg(char **argptr, int separator)
+{
+	__sz arglen;
+	char *arg;
+
+	UK_ASSERT(argptr);
+
+	arg = *argptr;
+	arglen = uk_nextarg_r((const char **) argptr, separator);
```
mogasergiu: As per guide we should refrain using spaces after casts.

lib/ukargparse/argparse.c
```diff
@@ -139,3 +140,56 @@ int uk_argnparse(char *argb, __sz maxlen, char *argv[], int maxcount)
...
+
+	/* Return NULL if we are at the end of parsing */
+	if ((arglen == 0) && (*argptr == __NULL))
```
mogasergiu: Why so many parantheses again :stuck_out_tongue_closed_eyes:.

lib/ukargparse/include/uk/argparse.h
```diff
@@ -50,4 +50,60 @@ int uk_argnparse(char *argb, __sz maxlen, char *argv[], int maxcount);
 }
 #endif
 
+/**
+ * Argument extraction function that can be used to iterates over a C-string
```
mogasergiu: s/iterates/iterate

lib/ukargparse/include/uk/argparse.h
```diff
@@ -50,4 +50,60 @@ int uk_argnparse(char *argb, __sz maxlen, char *argv[], int maxcount);
...
+ * buffer ('\0'-terminated) that contains multiple C-string sub-arguments
+ * separated by a special character. One example is a colon-separated argument
+ * list. On each invocation, the next argument is returned.
+ * NOTE: This function modifies the input buffer by replacing a found separator
+ *       in the buffer with a terminating '\0'-character.
+ *
+ * @param argptr Reference to a pointer pointing to the current parsing offset
```
mogasergiu: s/pointing //

lib/uknetdev/netdev.c
```diff
@@ -139,40 +107,63 @@ static struct uk_netdev_einfo_overwrites *_alloc_einfo(struct uk_alloc *a,
...
-	if (_einfo->ip4.hostname)
-		uk_pr_debug("netdev%d: Overwrite einfo ip4.host: \"%s\"\n",
-			    netdev_id, _einfo->ip4.hostname);
-	if (_einfo->ip4.domain)
-		uk_pr_debug("netdev%d: Overwrite einfo ip4.domain: \"%s\"\n",
-			    netdev_id, _einfo->ip4.domain);
-
+	if (_einfo->ip4.cidr) {
+		if (_einfo->ip4.cidr[0] == '\0')
+			_einfo->ip4.cidr = NULL;
```
mogasergiu: Why not `__NULL` here as well?

lib/uknetdev/netdev.c
```diff
@@ -139,40 +107,63 @@ static struct uk_netdev_einfo_overwrites *_alloc_einfo(struct uk_alloc *a,
...
+		else
```
mogasergiu: These all look like unnecessarily taken branches if PRINTD is not enabled :thinking:. Maybe we could surround this `else` with `#if`'s? Or, better yet, just place everything into one single `uk_pr_debug`? I see that `__uk_vsnprintf` can handle null strings. And then the `if`'s would just be
```C
if (_einfo->* && _einfo->ip4.*[0] == '\0')
```

Although it probably does not matter right? Perhaps DCE removes them anyways.

lib/uknetdev/netdev.c
```diff
@@ -139,40 +107,63 @@ static struct uk_netdev_einfo_overwrites *_alloc_einfo(struct uk_alloc *a,
...
+	 * NOTE: `uk_nextarg` automatically returns NULL if an arguments
+	 *       does not exists.
 	 */
-	_einfo->ip4.cidr     = pick_arg(&ipv4_conf[netdev_id], ':');
-	_einfo->ip4.gw       = pick_arg(&ipv4_conf[netdev_id], ':');
-	_einfo->ip4.dns0     = pick_arg(&ipv4_conf[netdev_id], ':');
-	_einfo->ip4.dns1     = pick_arg(&ipv4_conf[netdev_id], ':');
-	_einfo->ip4.hostname = pick_arg(&ipv4_conf[netdev_id], ':');
-	_einfo->ip4.domain   = pick_arg(&ipv4_conf[netdev_id], ':');
+	_einfo->ip4.cidr     = uk_nextarg(&ipv4_conf[netdev_id], ':');
```
mogasergiu: Let's maintain consistency with the spacing around `=`? :thinking: This way, if someone in the future adds another longer named parameter they wouldn't have to re-align. Could be just my subjective opinion though. :man_shrugging:

lib/vfscore/automount.c
```diff
@@ -591,19 +557,19 @@ static int vfscore_parse_volume(char *v, struct vfscore_volume *vv)
 	UK_ASSERT(vv);
 
 	pos = v;
-	vv->sdev   = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
-	vv->path   = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
-	vv->drv    = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
-	strflags   = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
-	vv->opts   = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
-	ukopts     = next_arg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
+	vv->sdev   = uk_nextarg(&pos, LIBVFSCORE_FSTAB_VOLUME_ARGS_SEP);
```
mogasergiu: Ditto

lib/ukargparse/argparse.c
```diff
@@ -193,3 +193,49 @@ char *uk_nextarg(char **argptr, int separator)
 	arg[arglen] = '\0';
 	return arg;
 }
+
+__ssz uk_strnkeycmp(const char *str, __sz strlen, const char *key,
+		    const char separators[])
+{
+	const char *s;
+	size_t pos;
```
mogasergiu: s/size_t/__sz

lib/ukargparse/argparse.c
```diff
@@ -193,3 +193,49 @@ char *uk_nextarg(char **argptr, int separator)
 	arg[arglen] = '\0';
 	return arg;
 }
+
+__ssz uk_strnkeycmp(const char *str, __sz strlen, const char *key,
```
mogasergiu, line 197: I know it's not a problem here, but can we pick another variable name, something other than `strlen`. :grin:

lib/ukargparse/include/uk/argparse.h
```diff
@@ -106,4 +106,36 @@ char *uk_nextarg(char **argptr, int separator);
...
+ *
+ * The following example demonstrates when the function returns with a match.
+ * C string to check is "ramfs=2", this is checked against the keywords
+ * "ram", "ramfs", and "format" (separators is set to "="):
+ *
+ *   pos | 0   1   2   3   4   5   6   7 |
+ *   ====+===+===+===+===+===+===+===+===+=========================
+ *   str | r   a   m   f   s   =   2  \0 | input to compare against
+ *   ----+---+---+---+---+---+---+---+---+-------------------------
+ *   key | r   a   m  \0                 | no match
```
mogasergiu, line 125: Wow you put in the effort for this one :rocket: :guitar: :muscle: Cool table

lib/vfscore/automount.c
```diff
@@ -346,10 +350,86 @@ extern const char vfscore_einitrd_start[];
 extern const char vfscore_einitrd_end;
 #endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_EINITRD */
 
+#define LIBVFSCORE_EXTRACTOPT_MOUNT_RAMFS		(0x1 << 0)
```
mogasergiu, line 416: Perhaps we could use `UK_BIT` macro here?

lib/vfscore/automount.c
```diff
@@ -346,10 +350,86 @@ extern const char vfscore_einitrd_start[];
...
+#define LIBVFSCORE_EXTRACTOPT_MOUNT_IFNOMOUNT		(0x1 << 1)
+#define LIBVFSCORE_EXTRACTOPT_FORMAT_AUTO		(0x1 << 2)
+
+static inline int vfscore_extract_parse_opts(const char *opts)
+{
+	char strbuf[64];
```
mogasergiu, line 424: You know I am going to say it... :triumph:  **Reverse Xmas Tree**? :pleading_face: Also s/size_t/__sz, s/ssize_t/__ssz.?

lib/vfscore/automount.c
```diff
@@ -346,10 +350,86 @@ extern const char vfscore_einitrd_start[];
...
+		 *        (default if present without value)
+		 *    2 - Only mount underlying ramfs if no mount point exists
+		 */
+		match = uk_strnkeycmp(arg, arglen, "ramfs", "=");
+		if (match == 0) { /* only "ramfs" */
+			uk_pr_debug("fstab: Option for \"extract\": \"ramfs=1\" (always mount)\n");
+			optflags |= LIBVFSCORE_EXTRACTOPT_MOUNT_RAMFS;
+			optflags &= ~LIBVFSCORE_EXTRACTOPT_MOUNT_IFNOMOUNT;
+		} else if (match > 0) { /* "ramfs=N" */
+			int param;
```
mogasergiu, line 450: Even though it is not required in the above scope, maybe still declare it in the beginning with the others?

lib/vfscore/automount.c
```diff
@@ -346,10 +350,86 @@ extern const char vfscore_einitrd_start[];
...
+static inline int vfscore_extract_parse_opts(const char *opts)
+{
+	char strbuf[64];
+	size_t cpylen;
+	const char *arg, *iter;
+	size_t arglen;
+	ssize_t match;
+	int optflags = 0x0;
+
+	for (iter = opts; iter != NULL; ) {
```
mogasergiu: Can we just go with s/iter != NULL/iter/

lib/vfscore/automount.c
```diff
@@ -385,6 +465,26 @@ static int vfscore_extract_volume(const struct vfscore_volume *vv)
...
+	 */
+	if ((opts & LIBVFSCORE_EXTRACTOPT_MOUNT_RAMFS) &&
+	    (!(opts & LIBVFSCORE_EXTRACTOPT_MOUNT_IFNOMOUNT) ||
+	     uk_list_empty_careful(&mount_list))) {
+		uk_pr_info("Mounting 'ramfs' to %s...\n", vv->path);
+		rc = mount("", vv->path, "ramfs", 0x0, "");
+		if (unlikely(rc)) {
+			uk_pr_crit("Failed to mount ramfs to \"%s\": %d\n",
+				   vv->path, errno);
+			return -1;
```
mogasergiu, line 547: Why not just return `rc`?

lib/vfscore/Config.uk
```diff
@@ -104,6 +104,10 @@ if LIBVFSCORE_AUTOMOUNT_CI
 		menu "Entry options"
 		config LIBVFSCORE_AUTOMOUNT_CI0_UKOPTS_MKMP
 		bool "mkmp - Create mountpoint"
+		config LIBVFSCORE_AUTOMOUNT_CI0_UKOPTS_IFINITRD0
```
mogasergiu, line 107: I think `KConfig` styling involves some indentation still but, alas, no config here seems to really respect that anyway so ok..

lib/vfscore/automount.c
```diff
@@ -138,6 +144,12 @@ const struct vfscore_volume __fstab_ci1 = {
 #if CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_MKMP_ARG
 		  | LIBVFSCORE_UKOPT_MKMP
 #endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_MKMP_ARG */
+#if CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_IFINITRD0_ARG
+		  | LIBVFSCORE_UKOPT_IFINITRD0
+#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_IFINITRD0_ARG */
+#if CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_IFNOINITRD0_ARG
+		  | LIBVFSCORE_UKOPT_IFNOINITRD0
+#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_CI1_UKOPTS_IFNOINITRD0_ARG */
```
mogasergiu, line 153: Why not add the `| LIBVFSCORE_UKOPT_IFNOINITRD0` macro as well in this comment?

lib/vfscore/automount.c
```diff
@@ -344,6 +392,20 @@ UK_LIBPARAM_PARAM_ARR_ALIAS(fstab, &fstab_up, charp,
 
 #endif /* CONFIG_LIBVFSCORE_AUTOMOUNT_UP */
 
+int vfscore_get_initrd0(struct ukplat_memregion_desc **out)
```
mogasergiu: This seems to only be called in this file. Maybe we should make it `static?

lib/vfscore/automount.c
```diff
@@ -344,6 +392,20 @@ UK_LIBPARAM_PARAM_ARR_ALIAS(fstab, &fstab_up, charp,
...
+{
+	struct ukplat_memregion_desc *initrd0;
+	struct ukplat_memregion_desc **itr;
+	int rc;
+
+	itr = out ? out : &initrd0;
+	rc = ukplat_memregion_find_initrd0(itr);
+	if (unlikely((rc < 0) || ((*itr)->len == 0)))
+		return -1;
```
mogasergiu, line 405: Why not return the `rc`?

EDIT: Ah, it's because `rc` may be `0` but `((*itr)->len == 0)`.... also why so many parantheses again :grimacing:

lib/vfscore/automount.c
```diff
@@ -440,8 +502,8 @@ static int vfscore_extract_volume(const struct vfscore_volume *vv)
 	if (!strcmp(vv->sdev, LIBVFSCORE_EXTRACT_DEV_INITRD0)) {
 		struct ukplat_memregion_desc *initrd;
 
-		rc = ukplat_memregion_find_initrd0(&initrd);
-		if (unlikely(rc < 0 || initrd->len == 0)) {
+		rc = vfscore_get_initrd0(&initrd);
+		if (unlikely(rc < 0)) {
 			uk_pr_crit("Could not find an initrd!\n");
 			return -1;
```
mogasergiu, line 509: Ditto (I know it is not part of the PR)

lib/vfscore/automount.c
```diff
@@ -576,6 +642,25 @@ static inline int vfscore_mount_volume(const struct vfscore_volume *vv)
 		return -EINVAL;
 	}
 
+	if (vv->ukopts & LIBVFSCORE_UKOPT_MKMP)
+		uk_pr_debug("fstab: Entry with ukopt: mkmp\n");
```
mogasergiu, line 647: Once again, I don't think it's worth adding these `if`'s here right? They would probably be optimized out anyway by DCE though... so ok.

lib/vfscore/Config.uk
```diff
@@ -785,4 +786,8 @@ if LIBVFSCORE_AUTOMOUNT_EINITRD
 		image and mounted under '/'.
 endif
 
+config LIBVFSCORE_AUTOUNMOUNT
```
mogasergiu, line 789: > Only in the case that automatic mounting is enabled, the unmounting option is activated.

But it looks like it could be selected individually? Also, I guess that if a user disables automounting in favor of having his app do the mounting, they would have to know to also enable this config right?

lib/vfscore/automount.c
```diff
@@ -900,5 +906,10 @@ err_umount: __maybe_unused
 	vfscore_autoumount(NULL);
 	return rc;
 }
+#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT */
 
+#if CONFIG_LIBVFSCORE_AUTOMOUNT && CONFIG_LIBVFSCORE_AUTOUNMOUNT
 uk_rootfs_initcall_prio(vfscore_automount, vfscore_autoumount, 4);
+#elif CONFIG_LIBVFSCORE_AUTOUNMOUNT
+uk_rootfs_initcall_prio(0x0, vfscore_autoumount, 4);
+#endif
```
mogasergiu: Forgot the `/* !CONFIG_LIBVFSCORE_AUTOMOUNT && !CONFIG_LIBVFSCORE_AUTOUNMOUNT */` :stuck_out_tongue_closed_eyes:

mogasergiu (changes requested):
Hi :wave: @skuenzer! Thanks for this work! High quality as always. I took a quick initial look mostly nitpicking things. I will take a closer look in the next round to maybe try and catch some edge cases :thinking:.

And, don't forget... **Reverse Xmas Tree is sacred!!** :pray: :pray: :christmas_tree: :christmas_tree:

lib/ukargparse/argparse.c
skuenzer: Sorry...

lib/uknetdev/netdev.c
skuenzer: Yes, my assumption is that the compiler will optimize these things out, even without DCE. So, I would not change it.

lib/uknetdev/netdev.c
skuenzer: You want a single space, nothing aligned? I don't mind actually.

lib/ukargparse/argparse.c
skuenzer: a right, yes. 👍🏻

lib/ukargparse/argparse.c
skuenzer: I named it this way to point out that `strlen` is the len of the `str` parameter. The function name starts also with `strn*` to give analogy to `strncpy()` and co.

lib/vfscore/automount.c
skuenzer: We could, but... I find this more readable and direct. 😄  The thing is actually that I would cause a dependency to `lib/ukbitopts` just for this file-local flag definition. There is nothing else happening than checking for a flag. It is probably a style question. 🤔

lib/vfscore/automount.c
skuenzer: Data types, I agree, will adopt. Reverse xmas..., I find this optional. but I'll change it for you.

lib/vfscore/automount.c
skuenzer: I had a discussion with @andreittr regarding this and I think makes more sense to move such declaration to a subscopes if possible. It is still declared at the beginning of a scope.

lib/vfscore/automount.c
skuenzer: of course, fine.

lib/vfscore/automount.c
skuenzer: I think this is a detail. `mount` will return -1 in error cases. If I want to forward an meaningful error, I should forward `-errno`. THis is to make clear that I am not blindly taking whatever mount returns.

lib/ukargparse/argparse.c
skuenzer: 👍🏻

lib/vfscore/automount.c
skuenzer: We did not use any Unikraft internal data types in this file, I'll keep `size_t` and co.

lib/uknetdev/netdev.c
skuenzer: I did now split it into two if-blocks. This may make it more readable because one if-nesting is removed (which is what you wanted to achieve). I assume the same here: the compiler will opt-out empty if statements.

lib/vfscore/automount.c
skuenzer: true, can be `static`, will adopt

lib/vfscore/automount.c
skuenzer: I see this similar in order to be consistent what the automount function returns (always -1 for failure), even if we would update `vfscore_get_initrd()` return codes.

lib/vfscore/automount.c
skuenzer: The message should be printed only with this if-condition. If debug is off, the if-body becomes empty and the compiler should optimize the complete condition out.

lib/vfscore/Config.uk
skuenzer: The idea is that a user could program its own mounting logic while just take this option for letting handle the unmount automatically. If you slect automount in vfscore, it will enable also unmounting.

lib/ukargparse/argparse.c
```diff
@@ -139,3 +140,56 @@ int uk_argnparse(char *argb, __sz maxlen, char *argv[], int maxcount)
 out:
 	return argc;
 }
+
+__sz uk_nextarg_r(const char **argptr, int separator)
+{
+	const char *nsep;
+	const char *arg;
```
mogasergiu, line 147: Would it make sense to add an assert for `separator` checking if it is out of `char` bounds?

lib/vfscore/automount.c
```diff
@@ -900,5 +906,10 @@ err_umount: __maybe_unused
 	vfscore_autoumount(NULL);
 	return rc;
 }
+#endif /* CONFIG_LIBVFSCORE_AUTOMOUNT */
 
+#if CONFIG_LIBVFSCORE_AUTOMOUNT && CONFIG_LIBVFSCORE_AUTOUNMOUNT
 uk_rootfs_initcall_prio(vfscore_automount, vfscore_autoumount, 4);
+#elif CONFIG_LIBVFSCORE_AUTOUNMOUNT
+uk_rootfs_initcall_prio(0x0, vfscore_autoumount, 4);
+#endif /* CONFIG_LIBVFSCORE_AUTOUNMOUNT */
```
mogasergiu, line 915: `/* !CONFIG_LIBVFSCORE_AUTOMOUNT && CONFIG_LIBVFSCORE_AUTOUNMOUNT */`?

mogasergiu (approved):
Two more minor questions. Looks very good! :1st_place_medal:. I will approve anyway.

lib/ukargparse/argparse.c
skuenzer: Hum, ... since it is just forwarded to `strchr`, I'd offload that check to that function.

lib/vfscore/automount.c
skuenzer: It is this way because of the `#elif` case.
We had agreed on:
```c
#if COND_A

#else /* !COND_A - expression that is covering all previously non-catched cases */

#endif /* !COND_A - repeat last condition */
```

and
```c
#if COND_A

#elif COND_B

#endif /* !COND_B - repeat last condition */
```

razvand (approved):

PR #1296, opened 2024-01-31
lib/posix-socket: Introduce CREATE, BIND events

skuenzer:
The current set of events does not allow UDP connections to be tracked as they do not require a 'listen' call. UNIX domain sockets cannot be tracked either, as they can work even without a `bind` operation. Because of all this, this commit introduces two more state change events that can be observed:
 - CREATE: A socket got created
 - BIND: An address got bind to a socket


StefanJum (approved):
All good, thanks.

mariasfiraiala (approved):
Works just fine, thanks!

razvand (approved):

PR #1297, opened 2024-01-31
lib/posix-poll: Fix epoll missing events

andreittr:
### Description of changes

Previously epoll would miss events when configured for edge-polling and unable to return all captured events in a single call to epoll_wait, by erroneously marking the epoll fd as not ready, even though it may still have pending events.
This change fixes this oversight, restoring expected behavior to epoll_wait.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Trigger/test snippet:
```c
int main(void)
{
	struct epoll_event ev;
	int p[2], q[2];
	int r;

	r = pipe(p);
	assert(!r);
	r = pipe(q);
	assert(!r);

	int e = epoll_create(2);
	assert(e >= 0);

	ev.events = EPOLLIN|EPOLLET;
	ev.data.fd = p[0];
	r = epoll_ctl(e, EPOLL_CTL_ADD, p[0], &ev);
	assert(!r);

	ev.events = EPOLLIN|EPOLLET;
	ev.data.fd = q[0];
	r = epoll_ctl(e, EPOLL_CTL_ADD, q[0], &ev);
	assert(!r);

	write(p[1], "a", 1);
	write(q[1], "b", 1);

	do {
		char buf[2];
		r = epoll_wait(e, &ev, 1, 0);
		printf("%d\n", r);
		if (r) {
			read(ev.data.fd, buf, 1);
			buf[1] = 0;
			puts(buf);
		}
	} while (r);
	return 0;
}
```
Output should be
```
1
a
1
b
0
```
On both Linux and Unikraft; previously unikraft would miss one of the messages (most likely `b`).

### DIFF
--- a/lib/posix-poll/epoll.c
+++ b/lib/posix-poll/epoll.c
@@ -562,8 +562,8 @@ int uk_sys_epoll_pwait2(const struct uk_file *epf, struct epoll_event *events,
 		}
 		uk_file_runlock(epf);
 
-		/* If lvlev, update pollin back in */
-		if (lvlev)
+		/* If lvlev or limited by maxevents, update pollin back in */
+		if (lvlev || nout == maxevents)
 			uk_file_event_set(epf, UKFD_POLLIN);
 
 		if (nout)


StefanJum (approved):
All good, thanks.

lib/posix-poll/epoll.c
```diff
@@ -562,8 +562,8 @@ int uk_sys_epoll_pwait2(const struct uk_file *epf, struct epoll_event *events,
 		}
 		uk_file_runlock(epf);
 
-		/* If lvlev, update pollin back in */
-		if (lvlev)
+		/* If lvlev or limited by maxevents, update pollin back in */
+		if (lvlev || nout == maxevents)
 			uk_file_event_set(epf, UKFD_POLLIN);
```
mogasergiu, line 567: Doesn't this function return something? Is there a reasion we are not checking for return values?

mogasergiu: In fact I don't see its return values being checked at any of its callsite :thinking:

andreittr: `uk_file_event_*` functions return the previous events, as a convenience due to the fact that event setting is done atomically so there's no other way to get that previous value. This is not needed often, but is needed _sometimes_, thus the need to return it.

In this case we don't care what the previous events were. I used to cast such invocations to void, but this turned out to be _everywhere_, visually polluting the code, not to mention increasing the pressure on keeping everything within 80 cols with size 8 tabs :stuck_out_tongue: ; so I stopped that.

mogasergiu: Oh well I guess we also call `printf` while ignoring return value...

mogasergiu (approved):

razvand (approved):

PR #1315, opened 2024-02-07
arch/x86: Use function pointers instead of switch cases for ECTX ops

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

During the initialization of the extended context we would, depending on the result of CPUID, set alignment, size and corresponding store/restore methods of this context. The latter, we would set to the value of an enum which we would afterwards use to decide which of the methods to use when storing/restoring. Change this so that we directly invoke such methods instead doing a switch case.

`NOTE`: I could not use the builtins because of `ectx.c` being marked as `|isr`

### DIFF
--- a/arch/x86/ectx.c
+++ b/arch/x86/ectx.c
@@ -43,19 +43,100 @@
 #include <uk/hexdump.h>
 #include <uk/isr/string.h> /* memset_isr */
 
-enum x86_save_method {
-	X86_SAVE_NONE = 0,
-	X86_SAVE_FSAVE,
-	X86_SAVE_FXSAVE,
-	X86_SAVE_XSAVE,
-	X86_SAVE_XSAVEOPT
-};
-
-static enum x86_save_method ectx_method;
-static __sz ectx_size;
 static __sz ectx_align = 0x0;
+static __sz ectx_size;
 
-static void _init_ectx_store(void)
+static void ectx_store_none(struct ukarch_ectx *state __unused) { }
+
+static void ectx_load_none(struct ukarch_ectx *state __unused) { }
+
+static void ectx_sanitize_none(struct ukarch_ectx *state __unused) { }
+
+static void ectx_store_fsave(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+
+	__asm__ __volatile__("fsave (%0)" :: "r"(state) : "memory");
+}
+
+static void ectx_load_frstor(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+
+	__asm__ __volatile__("frstor (%0)" :: "r"(state));
+}
+
+static void ectx_store_fxsave(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, 16));
+
+	__asm__ __volatile__("fxsave (%0)" :: "r"(state) : "memory");
+}
+
+static void ectx_load_fxrstor(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, 16));
+
+	__asm__ __volatile__("fxrstor (%0)" :: "r"(state));
+}
+
+/* The specific state components restored correspond to the bits set in the
+ * requested-feature bitmap (RFBM), which is the logical-AND of EDX:EAX and
+ * XCR0. So, save everything and use mask 0xffffffff:0xffffffff.
+ */
+static void ectx_store_xsave(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, 64));
+
+	__asm__ __volatile__("xsave (%0)" :: "r"(state),
+			     "a"(0xffffffff), "d"(0xffffffff) : "memory");
+}
+
+static void ectx_store_xsaveopt(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, 64));
+
+	__asm__ __volatile__("xsaveopt (%0)" :: "r"(state),
+			     "a"(0xffffffff), "d"(0xffffffff) : "memory");
+}
+
+static void ectx_load_xrstor(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr) state, 64));
+
+	__asm__ __volatile__("xrstor (%0)" :: "r"(state),
+			     "a"(0xffffffff), "d"(0xffffffff) : "memory");
+}
+
+static void ectx_sanitize_xsave(struct ukarch_ectx *state)
+{
+	UK_ASSERT(state);
+	UK_ASSERT(IS_ALIGNED((__uptr)state, 64));
+
+	/* XSAVE* & XRSTOR rely on sane values in the XSAVE header
+	 * (64 bytes starting at offset 512 from the base address)
+	 * and will raise #GP on garbage data. We must zero them out.
+	 */
+	((__u64 *)state)[64] = 0;
+	((__u64 *)state)[65] = 0;
+	((__u64 *)state)[66] = 0;
+	((__u64 *)state)[67] = 0;
+	((__u64 *)state)[68] = 0;
+	((__u64 *)state)[69] = 0;
+	((__u64 *)state)[70] = 0;
+	((__u64 *)state)[71] = 0;
+}
+
+void (*ukarch_ectx_sanitize)(struct ukarch_ectx *state) = ectx_sanitize_none;
+void (*ukarch_ectx_load)(struct ukarch_ectx *state) = ectx_load_none;
+void (*ukarch_ectx_store)(struct ukarch_ectx *state) = ectx_store_none;
+
+static void init_ectx_store(void)
 {
 	__u32 eax, ebx, ecx, edx;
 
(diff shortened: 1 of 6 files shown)


arch/x86/ectx.c
```diff
@@ -43,22 +43,89 @@
...
+	((__u64 *)state)[65] = 0;
+	((__u64 *)state)[66] = 0;
+	((__u64 *)state)[67] = 0;
+	((__u64 *)state)[68] = 0;
+	((__u64 *)state)[69] = 0;
+	((__u64 *)state)[70] = 0;
+	((__u64 *)state)[71] = 0;
+}
+
+typedef void (*ukarch_ectx_store_fn_t)(struct ukarch_ectx *);
```
andreittr: Function typedefs should use use the `_func` suffix as per [coding conventions](https://unikraft.org/docs/contributing/coding-conventions#typedefs). Largely because `_t` suffix is reserved by posix.

arch/x86/ectx.c
```diff
@@ -43,22 +43,89 @@
...
+typedef void (*ukarch_ectx_load_fn_t)(struct ukarch_ectx *);
+typedef void (*ukarch_ectx_sanitize_fn_t)(struct ukarch_ectx *);
+
+ukarch_ectx_store_fn_t ectx_store_fn;
```
andreittr: These should be static, as they are not exported outside this compilation unit as the code is right now.

That said, I think there's room here for a better implementation. AFAICT (pls correct me if I'm wrong) this PR changes ectx handling overhead from
- direct func call (`uk_ectx_*`) + global mem read (`ectx_method`) + switch branching
to
- direct func call (`uk_ectx_*`) + global mem read (`ectx_*_fn`) + indirect func call

which is not obviously faster. Rather I think we can do better and manage only 1 indirect func call total thusly:
1. remove the `ukarch_ectx_*` functions and make `ukarch_ectx_*` the function pointers instead of `ectx_*_fn` (this time they should be non-static)
2. change `ukarch_ectx_*` func declarations in `uk/arch/ctx.h` into exported function pointers
```c
void (*ukarch_ectx_store)(struct ukarch_ectx *state);
```
3. init these pointers as usual in `init_ectx_store`
4. calling code doesn't need to change! Through the magic of C syntax, calling via function linkage or function pointer looks the same!
5. profit!

Also, perhaps consider something similar for size & align, exporting the global vars; although that might be more involved, needing inlines defined in `uk/arch/ctx.h` to keep API compatibility intact.

arch/x86/ectx.c
```diff
@@ -43,22 +43,89 @@
...
+ukarch_ectx_load_fn_t ectx_load_fn;
+ukarch_ectx_sanitize_fn_t ectx_sanitize_fn;
+
+static void init_ectx_store(void)
 {
 	__u32 eax, ebx, ecx, edx;
 
+	ectx_load_fn = ectx_load_none;
```
andreittr: You can omit these if you initialize the global func pointers at compile time. This also prevents crashes in the unlikely event they get called before initialization, and removes the need for some asserts since you always are guaranteed _something_ sane in those pointers.

andreittr (changes requested):
A few small nitpicks, and one larger improvement proposal that would extract the most performance from this new approach. See comments.

arch/x86/ectx.c
mogasergiu: > AFAICT (pls correct me if I'm wrong) this PR changes ectx handling overhead from
>
>direct func call (uk_ectx_*) + global mem read (ectx_method) + switch branching
>to
>direct func call (uk_ectx_*) + global mem read (ectx_*_fn) + indirect func call

yes. I want to be able to assert if assertions are enabled without copy-pasting them everywhere.

Is it wrong to assume that a fully optimized build without assertions would do just that (what you proposed)?

arch/x86/ectx.c
mogasergiu: Very good observation. I did do this in the newer [fsgsbase refactor](<https://github.com/unikraft/unikraft/pull/1316/files#diff-147aedf7938a29560bd11ce90a24feebcd33d445a5c38d16ded99b77489f5ed6R89>). I suppose I lost sight of making this for this ECTX PR as well. Nice! :+1:

arch/x86/ectx.c
andreittr: > Is it wrong to assume that a fully optimized build without assertions would do just that (what you proposed)?

There's some chance that a good LTO pass _might_ decide to inline `uk_ectx_` calls (assuming the `|isr` thing doesn't negatively impact LTO, you need to check) so ectx calls end up 1 indirect call in the final binary, but that's not guaranteed. If we make it explicit in code it will be certain.

> yes. I want to be able to assert if assertions are enabled without copy-pasting them everywhere.

I think you would need far fewer asserts than right now if you directly export func ptrs, especially if you implement the other comment regarding init values:
- `ectx_align` (i.e., checking whether inited): should not be necessary since (1) a "real" ectx func can get called only after init (bc they're static and accessible only via func ptrs inited by you) and (2) the pre-init value of `ectx_*_none` is a no-op that doesn't care
- `state`: not strictly needed since it only crashes on null ptrs that would trigger a crash anyway; plenty of other invalid ptrs out there that will crash but this won't catch :shrug: 
- `ectx_*_fn`: you _are_ that function, obviously it exists :smile: 
- `IS_ALIGNED(state, ectx_align)`: this one is the only that really is useful, and IMO it would be faster if it had a constant to check against instead of a runtime-assigned global var (LTO can't help you here I'm afraid). So there would be a tangible benefit to moving this assert into individual `uk_ectx_*` funcs that know their own required alignment at compile time.

mogasergiu: > ectx_align (i.e., checking whether inited): should not be necessary since (1) a "real" ectx func can get called only after init (bc they're static and accessible only via func ptrs inited by you) and (2) the pre-init value of ectx_*_none is a no-op that doesn't care

Actually, since there are no 64-bit machines that do not have at least SSE2, those none variants for store/load are useless I guess, I may as well initialize statically to FSAVE/FRSTOR.

> There's some chance that a good LTO pass might decide to inline uk_ectx_ calls (assuming the |isr thing doesn't negatively impact LTO, you need to check) so ectx calls end up 1 indirect call in the final binary, but that's not guaranteed. If we make it explicit in code it will be certain.

```
# compiler flags to prevent use of extended (FP, SSE, AVX) registers.
# This is for files that contain trap/exception/interrupt handlers
ISR_ARCHFLAGS += -mno-80387 -mno-mmx -mno-sse -mno-avx
ISR_ARCHFLAGS-$(call have_gcc)	+= -mno-fp-ret-in-387
ISR_ARCHFLAGS-$(call gcc_version_ge,7,1) += -mgeneral-regs-only
```
Hmm, I don't think it would be bad for LTO.

> state: not strictly needed since it only crashes on null ptrs that would trigger a crash anyway 🤷

That would be worse. I would rather go through the exception hoops of an explicit crash than the hoops of an unhandled trap which would cause  more confusion to a user for example that all they would see is a runtime crash instead of an assertion.

> I think you would need far fewer asserts 

I agree with this and the rest. The only reason I have this indirection is so that we keep assertions in one place as well as have one centralized function where we can place prints. However, I am still skeptical on whether it is worth making this function always a function pointer as I am pretty sure it will be optimized always. I am waiting for other's opinions as well here.

andreittr: > That would be worse. I would rather go through the exception hoops of an explicit crash than the hoops of an unhandled trap which would cause more confusion to a user for example that all they would see is a runtime crash instead of an assertion.

You'd need to do that anyway for all non-NULL invalid ptrs, so you're not saving yourself _that_ much fuss.

> The only reason I have this indirection is so that we keep assertions in one place as well as have one centralized function where we can place prints

~~Then perhaps defining `uk_ectx_*` as inlines (+ exporting the func ptrs & size & align) in `ctx.h` would net us the best of both worlds: centralised assertions & prints as well as inlining without relying on LTO.~~ Nvm that's dumb since you want arch-independence :unamused: .

Let's ask @skuenzer what he thinks. Should we:
- keep `uk_ectx_*` as regular linkage functions and trust good LTO that they'll be inlined; without LTO the performance gain of this PR might be marginal (see [here](https://github.com/unikraft/unikraft/pull/1315#discussion_r1567447824))
- export `uk_ectx_*` as actual function pointers, which guarantees low overhead with the same API, but will require some duplication of asserts; it'll probably also require small changes to ARM ectx code to bring in line

mogasergiu: > You'd need to do that anyway for all non-NULL invalid ptrs, so you're not saving yourself that much fuss.

In the code that I submit I assert every single argument as well as their fields (if structs) that I use. I find it good practice, has 0 runtime cost on production build and can save you time. Just my style/belief. :man_shrugging:

mogasergiu: > it'll probably also require small changes to ARM ectx code to bring in line

oh yes, that 100% will need to happen as ARM only has one functio for each... just NEON registers. Unlike the variable alignment/size/layout of x86 :zany_face: :dumpling:

arch/x86/ectx.c
```diff
@@ -43,22 +43,89 @@
...
+}
+
+static void ectx_load_fxrstor(struct ukarch_ectx *state)
+{
+	__asm__ __volatile__("fxrstor (%0)" :: "r"(state));
+}
+
+/* The specific state components restored correspond to the bits set in the
+ * requested-feature bitmap (RFBM), which is the logical-AND of EDX:EAX and
+ * XCR0. So, save everything ans use mask 0xffffffff:0xffffffff.
```
razvanvirtan:
```suggestion
 * XCR0. So, save everything and use mask 0xffffffff:0xffffffff.
```

razvanvirtan (changes requested):
Thanks for this!
The current code looks good to me, just a couple of small notes. I can review again if you decide to implement the larger changes suggested by Andrei.

NIT regarding the commit message: `...instead **of** doing a switch case.`

arch/x86/ectx.c
skuenzer: Hum, it is actually an interesting  idea to provide the interface with function pointers. In fact we are otherwise only replacing one jump of the switch case with an indirect function call with this pr.

Assertions could also be handled with surrounding function macros or static inline functions 🤔. Actually moving ukarch_ectx_load and co. to the header as static inline   might achieve the same thing and is still well readable source. Overall I would make sure that the individual implementations (xsave , ...) stay static.

mogasergiu: Ok then how about making `ukarch_ectx_load/store` as static inlines with the asserts and those static inlines call the function pointer? I will have to adapt ARM code as well.

skuenzer: Yes, this is what I thought as a good variant to achieve better compiler optimization while still easy readable.

mogasergiu:
Applied feedback. Also requested @michpappas's review as I am now also touching some ARM assembly as well. I tried testing on ARM but for some reason, even without this patch, staging is crashing on Redis on my end. Not sure what the cause could be as I have not invested time into investigating it yet. It is hopefully a misconfiguration on my end though.

I don't really like how this turned out with using function pointers. It feels awkward API wise from assembly calls. Maybe we should use some macro wrappers for the assembly cases :thinking:.

PR #1320, opened 2024-02-10
lib/vfscore: Ensure we return symlink target on `namei_resolve`

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

The `namei_resolve` method is different from its `*_no_follow` variants as in it is supposed to follow the symbolic links and eventually return the dentry of the final symlink target. However, if the first dentry hit in the cache is a symlink, we no longer check if it is a symlink or not and simply return it. To fix this, simply add a check: if current dentry pointer is a symlink, jump to the symlink logic.

To reproduce this may be hard, as it requires parallelism (just concurrency on a single CPU system won't work):
- thread 1 is following symlink/readlink or any other function that may involve holding a reference to the dentry of a symlink
- at the same time, thread 2 does an openat() on the symlink and because a reference is still being held on the symlink it will be found as a dentry in the cache (lib/vfscore/lookup.c:143) and thus be returned immediately, without checking whether it is a symlink or not. The normal behavior of openat would have been to follow the symlink until the final target is reached, however this way the openat() ends up being done on the dentry of the symlink instead.

### DIFF
--- a/lib/vfscore/lookup.c
+++ b/lib/vfscore/lookup.c
@@ -142,6 +142,15 @@ namei_resolve(const char *path, struct dentry **dpp, char *realpath)
 		dp = dentry_lookup(mp, node);
 		if (dp) {
 			/* vnode is already active. */
+
+			/* If this dentry is a symlink then we should try to
+			 * return the final target. Otherwise the *_no_follow
+			 * variants of this function should've been called
+			 * instead.
+			 */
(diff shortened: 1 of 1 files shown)


michpappas (approved):
Picking up this change has fixed the issue for me.

razvand (approved):

PR #1322, opened 2024-02-13
lib: Introduce generic (page guarded) stack allocator library

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Implement a generic stack allocation wrapper over a given Unikraft
memory allocator. The library receives as arguments an allocator and,
if libukvmem is enabled, a virtual address space and an initial size.
The library works in three ways:
1. If libukvmem is disabled, the library will make use of the allocator
received as argument to allocate the initial allocator structure and
Unikraft compliant stacks.
2. If libukvmem is enabled, the library will make use of the allocator
received as argument to allocate the initial allocator structure and
Unikraft compliant stacks. The difference from 1. is that the virtual
address space and the initial size arguments are now available. The
two arguments will hint how much of the stack to premap whenever they
are allocated first, i.e. what is the initial stack size before
on-demand paging kicks in.
3. If libukvmem is enabled AND the configuration option of the library
integrating page guards is enabled, the allocator given as argument
will only be used once when allocating the initial allocator structure,
while the virtual address space and initial size arguments will be used
to create stack VMA's, specific to libukvmem.

With this addition, now libukboot is made aware of this stack allocator
and therefore a new configuration option is introduced to allow the
customization of the aforementioned initial size argument for when
libukvmem is enabled. Consequently, libukschedcoop's initialization
allows now for three additional allocator arguments for regular
stacks, for auxiliary stacks and for the TLS. This is done so that
libukboot can now initialize and pass to libukschedcoop separate
stack allocators, differentiated by the differently configured size,
alignment as well as the fact that, unlike regular stacks,
the auxiiliary stacks must always be fully mapped.

Also a quick basic commit sprinkling some error messages in the frame allocator :man_shrugging:.


lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+	addr = stack_malloc(a, total);
+	if (unlikely(!addr))
+		return NULL;
+
+	memset(addr, 0, total);
+
+	return addr;
+}
+
+struct uk_alloc *uk_allocstack_init(struct uk_alloc *a,
```
mogasergiu: Missed a spot, no trailing comma here.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+				     __sz initial_size
+#endif /* CONFIG_LIBUKVMEM */
+				     )
+{
+	struct uk_stackalloc *sa;
+
+	UK_ASSERT(a);
+#if CONFIG_LIBUKVMEM
+	UK_ASSERT(vas);
+#endif /* CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS */
```
mogasergiu: Note: just noticed I have lots of `CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS` instead of `CONFIG_LIBUKVMEM` comments.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+#include <uk/alloc_impl.h>
+#include <uk/arch/ctx.h>
+#include <uk/arch/paging.h>
+#include <uk/assert.h>
+#include <uk/essentials.h>
+#if CONFIG_LIBUKVMEM
+#include <uk/vmem.h>
+#endif /* CONFIG_LIBUKVMEM */
+
+struct uk_stackalloc {
```
mogasergiu: TODO: s/stackalloc/allocstack

lib/ukallocstack/include/uk/allocstack.h
```diff
@@ -0,0 +1,27 @@
...
+#ifndef __UKALLOCSTACK_H__
+#define __UKALLOCSTACK_H__
+
+#include <uk/alloc.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_alloc *uk_allocstack_init(struct uk_alloc *a
```
michpappas, line 53: As this is the public API, please consider documenting to avoid confusion (e.g. what is the purpose of `initial_size`), incl. all provided ops and the constraints in their parameters. This is especially important for allocator-specific variations from the usual constraints, such as the alignment restrictions in `[posix_]memalign()`.

lib/ukallocstack/include/uk/allocstack.h
```diff
@@ -0,0 +1,27 @@
...
+#if CONFIG_LIBUKVMEM
+				     , struct uk_vas __maybe_unused *vas,
+				     __sz initial_size
+#endif /* CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS */
```
michpappas: Should be `#endif /* CONFIG_LIBUKVMEM */`

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+
+	sa = uk_malloc(a, sizeof(*sa));
+	if (unlikely(!sa)) {
+		uk_pr_err("Failed to allocate stack allocator\n");
+		return NULL;
+	}
+#if CONFIG_LIBUKVMEM
+	sa->vas = vas;
+	sa->initial_size = initial_size;
+#endif /* !CONFIG_LIBUKVMEM */
```
michpappas: Should be `#endif /* CONFIG_LIBUKVMEM */`.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+
+void uk_allocstack_term(struct uk_alloc *a)
+{
+	struct uk_allocstack *sa;
+
+	UK_ASSERT(a);
+
+	sa = to_allocstack(a);
+
+	uk_free(sa->og_a, sa);
```
michpappas: ~~Also free(sa)~~ Sorry I didn't read this correctly.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+{
+	__vaddr_t sp __maybe_unused;
+	struct uk_allocstack *sa;
+	int rc __maybe_unused;
+	__vaddr_t vaddr;
+	__sz real_size;
+
+	UK_ASSERT(a);
+
+	if (!size)
```
michpappas: `unlikely`?

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+#include <uk/arch/ctx.h>
+#include <uk/arch/paging.h>
+#include <uk/assert.h>
+#include <uk/essentials.h>
+#if CONFIG_LIBUKVMEM
+#include <uk/vmem.h>
+#endif /* CONFIG_LIBUKVMEM */
+
+struct uk_allocstack {
+	struct uk_alloc a;
```
michpappas, line 20: I would had called these `sa->sa` and `sa->a` instead of `sa->a` and `sa->og_a` mostly because I don't like the `og_a` vary much, but maybe that's a matter of personal taste, so feel free to ignore if you prefer the current convention.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+	return __containerof(a, struct uk_allocstack, a);
+}
+
+static void *stack_malloc(struct uk_alloc *a, __sz size)
+{
+	__vaddr_t sp __maybe_unused;
+	struct uk_allocstack *sa;
+	int rc __maybe_unused;
+	__vaddr_t vaddr;
+	__sz real_size;
```
michpappas: Heheh, at least you didn't use [`desired_size`](https://github.com/unikraft/unikraft/pull/1212/files#diff-628d5113b18ee6253ddbcc848991834d8493a6ad8684eea4e944be9252f73340L79) 🤤 this time. `alloc_size` could be a more elegant name.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+
+	UK_ASSERT(a);
+
+	if (!size)
+		return NULL;
+
+	sa = to_allocstack(a);
+
+#if CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS
+	real_size = PAGE_ALIGN_UP(size) + PAGE_SIZE;
```
michpappas: Perhaps it would be useful to add a comment about the purpose of that extra page.

lib/ukallocstack/Config.uk
```diff
@@ -0,0 +1,22 @@
+menuconfig LIBUKALLOCSTACK
+	bool "ukallocstack: Generic stack allocator wrapper"
```
michpappas: Not sure if the "wrapper" part is accurate here, given that if guards are enabled it is no longer just a wrapper. See also the comment below.

lib/ukallocstack/Config.uk
```diff
@@ -0,0 +1,22 @@
...
+	default n
+	select LIBNOLIBC if !HAVE_LIBC
+	select LIBUKDEBUG
+	select LIBUKALLOC
+	help
+		This is a generic stack allocating library that is registered
```
michpappas, line 8: So the main value of this library is the guards. And being able to use the library without vmem is just a convenience that allows the user to make a single call without conditionals.

Given that, I would probably reverse the logic and state here that without stack guards enabled this acts as a simple wrapper around the allocator. Similarly, in the description of the guards, I wouldn't describe it as an exception of the wrapper case (i.e. "instead of being a wrapper").

PS: There is also a school of thought that says that you should not describe the implementation at all, but only what the feature provides.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+#endif /* CONFIG_LIBUKVMEM */
+
+	sa = uk_malloc(a, sizeof(*sa));
+	if (unlikely(!sa)) {
+		uk_pr_err("Failed to allocate stack allocator\n");
+		return NULL;
+	}
+#if CONFIG_LIBUKVMEM
+	sa->vas = vas;
+	sa->initial_size = initial_size;
```
michpappas: You may consider making it a requirement that the user passes an aligned `initial_size`, so that you don't need to align up in every individual op.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+	vaddr = (__vaddr_t)uk_memalign(sa->og_a, UKARCH_SP_ALIGN, real_size);
+	if (unlikely(!vaddr)) {
+		uk_pr_err("Failed to allocate stack of sizegth 0x%lx from "
+			  "allocator %p\n", real_size, sa->og_a);
+		return NULL;
+	}
+
+#if CONFIG_LIBUKVMEM
+	sp = ukarch_gen_sp(vaddr, real_size);
+	rc = uk_vma_advise(sa->vas,
```
michpappas, line 78: You may want to take into account the possibility that `size` < `initial_size`, or require that the user passes `size` >= `initial_size`.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+	/* Crash if pointer given to free is not part of any VMA. This would
+	 * normally lead to a heap corruption.
+	 */
+	UK_ASSERT(vma);
+
+	rc = uk_vma_unmap(sa->vas,
+			  vaddr,
+			  vma->end - vma->start,
+			  0);
+	UK_ASSERT(!rc);
```
michpappas: Same here, `UK_BUGON(rc)` if you choose to adopt the previous suggestion.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+
+	sa = to_allocstack(a);
+
+#if CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS
+	vaddr = (__vaddr_t)ptr;
+	vma = uk_vma_find(sa->vas, vaddr);
+	/* Crash if pointer given to free is not part of any VMA. This would
+	 * normally lead to a heap corruption.
+	 */
+	UK_ASSERT(vma);
```
michpappas: Maybe it's a matter of personal taste but as far as semantics go `UK_BUGON(!vma)` would be more suitable here.

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+}
+
+static void *stack_calloc(struct uk_alloc *a, __sz nmemb, __sz size)
+{
+	__sz total;
+	void *addr;
+
+	UK_ASSERT(a);
+
+	if (nmemb <= 1 || !size)
```
michpappas: `unlikely`?

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+	struct uk_allocstack *sa;
+	int rc __maybe_unused;
+	__vaddr_t vaddr;
+	__sz real_size;
+
+	UK_ASSERT(a);
+	UK_ASSERT(align >= UKARCH_SP_ALIGN);
+	UK_ASSERT(IS_ALIGNED(align,  UKARCH_SP_ALIGN));
+
+	if (!size)
```
michpappas: `unlikely`?

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+static int stack_posix_memalign(struct uk_alloc *a, void **memptr, __sz align,
+				__sz size)
+{
+	void *addr;
+
+	UK_ASSERT(a);
+	UK_ASSERT(memptr);
+	UK_ASSERT(align >= UKARCH_SP_ALIGN);
+	UK_ASSERT(IS_ALIGNED(align,  UKARCH_SP_ALIGN));
+#if CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS
```
michpappas: So on the one hand you make this check here, but on the other hand you delegate the `size` check to `stack_memalign()`. Maybe choose one scheme to be consistent?

lib/ukallocstack/stack.c
```diff
@@ -0,0 +1,277 @@
...
+		return NULL;
+
+	sa = to_allocstack(a);
+
+#if CONFIG_LIBUKALLOCSTACK_PAGE_GUARDS
+	UK_ASSERT(align <= PAGE_SIZE);
+
+	real_size = PAGE_ALIGN_UP(size) + PAGE_SIZE;
+	vaddr = __VADDR_ANY;
+	rc = uk_vma_map_stack(sa->vas,
```
michpappas, line 173: Shouldn't you adjust the return vaddr to the requested alignment?

michpappas (changes requested):
@mogasergiu nice 😎  Some review comments below.

lib/ukallocstack/stack.c
mogasergiu: I think I like `sa->a` and `sa->parent_a` :+1:

lib/ukallocstack/stack.c
mogasergiu: Regarding this and the other `unlikely` comments. I meant to align with the default `memalign`/`posix-memalign` checks which don't have these. But now, looking back, it's better to have them here in this case, since this lib is only meant to be used by our own kernel internal allocators :+1:

arch/Config.uk
```diff
@@ -44,6 +44,15 @@ config CPU_EXCEPT_STACK_SIZE_PAGE_ORDER
 		2 ^ order * page size (e.g. 4KB).
 		Only change this if you know what you're doing.
 
+config AUXSTACK_SIZE_PAGE_ORDER
```
michpappas, line 47: I am posting this here as GitHub does not provide a way to comment on the commit message: 

`STACK_SIZE` should not had been under `arch/` in the first place. As discussed before, we should be moving everything not defined by the architecture away from `arch/`. Therefore this change does not improve something, both the current and the new place will be changed again shortly.

An additional problem is that this change propagates to an additional change in Kconfig. In principle Kconfig directives is something we should treat as a stable ABI and limit changes on them to the absolutely necessary, as such changes have an impact both on the catalog (they must be propagated to Kraftfiles) and potentially to systems maintained by adopters of Unikraft - that's even worse as we have no control over these. Of course under re-arch these are expected to change a lot, but still we should avoid them when they are not necessary.

Moreover, various stack sizes it makes more sense for the various stack sizes to be defined by the platform rather than the architecture. My take is that they should live in the upcoming `platform.h`. So I would propose that we drop this change, and move them to `platform.h` as soon as we have consensus on that file with the rest of the maintainers (I'll bring it up on the next call). Alternatively, instead of dropping - and assuming there will be consensus on `platform.h` - we can defer merging and update this commit to move them to that place. What do you think?

lib/ukvmem/Config.uk
```diff
@@ -71,6 +71,18 @@ config LIBUKVMEM_STACK_BASE
 	default 0x0000004000000000
 	depends on LIBUKVMEM_DEDICATED_STACK_BASE
 
+config LIBUKVMEM_STACK_GUARD_PAGES_TOP
+	int "Top guard pages count"
+	default 4
```
michpappas, line 76: Although these are not costly, similarly to timeouts there are no good default values. So I would be curious to know where the selected default come from, are 4 pages found to be good enough for most cases? Also, how common are accesses beyond 1 page?

lib/ukvmem/include/uk/vma_types.h
```diff
@@ -109,32 +109,41 @@ extern const struct uk_vma_ops uk_vma_stack_ops;
...
  */
 static inline int uk_vma_map_stack(struct uk_vas *vas, __vaddr_t *vaddr,
 				   __sz len, unsigned long flags,
 				   const char *name, __sz initial_len)
 {
-	int rc;
 	__vaddr_t va;
+	int rc;
 
-	UK_ASSERT(len >= 2 * PAGE_SIZE); /* one page + guard page */
```
michpappas, line 127: Same here. If there is a requirement for increasing the number of guard pages to >1, then it would be worth providing the reasoning in the commit message.

Moreover, should any of the tests that involve guards need to be updated? I did a quick run and they all pass, but didn't look at the implementation to determine whether they are still valid or they just produce false positives.

lib/ukvmem/Config.uk
```diff
@@ -71,6 +71,18 @@ config LIBUKVMEM_STACK_BASE
...
+	help
+		Number of guard pages for the top of the stack (high address)
+
+config LIBUKVMEM_STACK_GUARD_PAGES_BOTTOM
+	int "Bottom guard pages count"
+	default 4
```
michpappas: For the stack base, is it meaningful to have more than 1 page? The only case I can think of is a case where a bug (benign or not) causes an overwrite beyond the stack base.

lib/ukvmem/include/uk/vma_types.h
```diff
@@ -109,32 +109,41 @@ extern const struct uk_vma_ops uk_vma_stack_ops;
...
- *   reserved as guard page.
+ *   Number of bytes pre-allocated for the stack.
  */
 static inline int uk_vma_map_stack(struct uk_vas *vas, __vaddr_t *vaddr,
 				   __sz len, unsigned long flags,
 				   const char *name, __sz initial_len)
 {
-	int rc;
 	__vaddr_t va;
+	int rc;
```
michpappas, line 136: I see what you did there 🎄

michpappas (changes requested):
@mogasergiu a few more

lib/ukvmem/Config.uk
mogasergiu: It's based on intuition after noticing that 2 pages are not enough for a Wordpress based application.

lib/ukvmem/include/uk/vma_types.h
mogasergiu: > Moreover, should any of the tests that involve guards need to be updated?

They are still fine, the new guard pages also include the old guard page.

lib/ukvmem/Config.uk
mogasergiu: > The only case I can think of is a case where a bug (benign or not) causes an overwrite beyond the stack base.

Yep, following a discussion with @skuenzer it was decided to make the guard pages end-to-end.

lib/ukvmem/include/uk/vma_types.h
mogasergiu: World domination next :christmas_tree: :santa:

arch/Config.uk
michpappas: **Update:** During an offline discussion with @skuenzer it was agreed that we merge this as is, given that migration involves additional code that existed previously in `arch`, and do a migration once all depending PRs are merged.

lib/ukallocstack/stack.c
mogasergiu: hmm i think it's actually not needed anymore as I [removed](https://github.com/unikraft/unikraft/pull/1322/commits/3bf2bc5de8824ae55bd812ec011d9be27914a2c6#diff-af0f86d6a09f05c4a44e859d58dfe38ab95d442f01f60e26f92b345aae167f8fR124) this constraint.

lib/ukallocstack/stack.c
mogasergiu: It's fine actually, that would mean the whole stack would be paged-in. I will document this.

lib/ukallocstack/stack.c
mogasergiu: If using page guards then we allocate stack VMA's instead. So, the return address will always be page-aligned and the maximum allowed alignment is PAGE_SIZE because we cannot satisfy higher alignments and smaller than PAGE_SIZE alignments must be a divisor of PAGE_SIZE.

I wil add a comment

mogasergiu: The only way we could do that is, I guess, if we make the following adjustment, and instead allocating bigger VMA's of size `alloc_size + align` and returning the `vaddr` as `vaddr = ALIGN_UP(vaddr, align)`, but at that point it becomes wasteful I would say as this is a specialized allocator we use in very specific cases, hence why I agreed to the `unlikely`'s as well (I also like them cus they make me look smart). I will add a comment in case this will be needed in the future to know what to do, but I highly doubt it. (yes i am also too lazy to make the change now :cowboy_hat_face: )

lib/ukvmem/Config.uk
michpappas: Random application (e.g. worpress) sounds like an arbitrary criterion though. Similarly to timeouts, there's no such thing as a good default, as there will always either too much or too little. Personally I would choose 1 and let the user adjust to the application requirements. Unless we have strong evidence of an average.

@skuenzer what are your thoughts on this?

lib/ukvmem/include/uk/vma_types.h
michpappas: Yes, but that sounds that they are now incomplete as they only test one guard page?

lib/ukvmem/Config.uk
michpappas: Okay, but what is the rationale for that? @skuenzer

lib/ukallocstack/stack.c
michpappas: Okay since we have a fundamental limitation on alignment by guards, perhaps we should at least document this constraint, given that it deviates from the expected behavior, and also add a check.

lib/ukallocstack/Config.uk
```diff
@@ -0,0 +1,22 @@
...
+		aligned to UKARCH_SP_ALIGN).
+
+if LIBUKALLOCSTACK
+config LIBUKALLOCSTACK_PAGE_GUARDS
+	bool "Use page guards for the allocated stacks"
+	default y
+	depends on LIBUKVMEM
+	help
+		Add end-to-end guard pages to the stacks allocated by the
+		allocator. Usdeful when wanting to catch stack overflows or
```
michpappas: "Useful"

lib/ukallocstack/include/uk/allocstack.h
```diff
@@ -0,0 +1,71 @@
...
+ * Given a parent allocator for the allocation of internal structures, return
+ * an allocator that allocates consistent stacks of a given configuration.
+ * Notably, if CONFIG_LIBUKVMEM is enabled, the Virtual Address Space that will
+ * contain the stacks is required as an argument and, additionaly, the initial,
+ * pre-paged-in, length of the stack may be given (may be 0). The latter may
+ * be used to reduce page faults on initial stack access.
+ *
+ * NOTE: When using the page guards configuration, the maximum allowed
+ * alignment for functions such as memalign/posix_memalign is PAGE_SIZE.
+ * See comment from `stack_memalign()`.
```
michpappas: Nit: "See comment in". 

Also consider whether it would be useful to additionally remind the reader that this is in addition to the usual requirement that alignment must be a power of 2.

lib/ukallocstack/include/uk/allocstack.h
```diff
@@ -0,0 +1,71 @@
...
+ * @initial_size
+ *    The initial paged-in size from top to bottom, i.e. last
+ *    initial_size / PAGE_SIZE pages will be paged-in.
+ *     E.g.:
+ *     - initial_size == PAGE_SIZE * 2 => the top two pages (stacks grow top to
+ *     bottom) of returned stacks will be paged-in and thus no page faults on
+ *     them
+ *     - initial_size == 0 => no page of any of the allocated stacks is
+ *     pre-mapped
+ *     - initial_size == SOME_STACK_SIZE then a uk_malloc(SOME_STACK_SIZE) would
```
michpappas: I don't know why `SOME` initially confused me, `STACK_SIZE` would had been immediately clear to me. But this may be subjective, so if you think it's good as is please feel free to ignore.

lib/ukallocstack/stack.c
michpappas: What I actually meant by `stack_size < initial_size` was under that condition we ask madvise to page in more memory than the actual stack. Therefore we need a check somewhere that `initial_size <= size`.

michpappas (changes requested):
@mogasergiu a few comments on the last update.

lib/ukallocstack/include/uk/allocstack.h
mogasergiu: Will surround it with `< >` to further suggest it's just some placeholder.

lib/ukschedcoop/include/uk/schedcoop.h
```diff
@@ -44,7 +44,10 @@
 extern "C" {
 #endif
 
-struct uk_sched *uk_schedcoop_create(struct uk_alloc *a);
+struct uk_sched *uk_schedcoop_create(struct uk_alloc *a,
```
michpappas, line 47: Leaving this as a note for @skuenzer. Not sure if it's something to do in the scope of this PR, but perhaps it would be worth to consider passing a struct with allocators similarly to `ops`. Throughout these PRs I have noticed that many functions in `thread.h` now take 10 or more parameters.

michpappas (changes requested):
From my side it's good. Remaining work before RB:
* Update tests / justify that they don't need to be updated.
* Open questions that require input from @skuenzer.
*  Break down into per-library commits before we merge.

@skuenzer this one's on you now.

lib/ukvmem/Config.uk
skuenzer: In fact the issue was related to on-stack array access. Since stacks are growing downwards and arrays upwards, you easily get outside of a single 4k page when accessing index 0 of a bigger array. On one hand I agree with you @michpappas and on the other hand this should be a protective feature. I found that there is a [POSIX interface](https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_attr_getguardsize.html) for it, so it should be ultimately a per-thread property to make it right. I can imagine that the rational behind this coulb probably be that an application can estimate typical access distances and size the guard pages accordingly.
At this point, I would accept the 4 pages for user experience but longer term we need to change the integration so that it becomes configurable per thread -> I think this requires some API discussion first.

mogasergiu:
TODO: Split last commit into three
- introduce the lib
- extend schedcoop function for the other three allocators
- adjust ukboot to use stack allocator

mogasergiu:
TODO: Following a discussion it's been decided:
- use only 1 guard page for bottom guards
- instead of `initial_len`/`initial_size` use `premapped_len`
- remove the termination function: "Since we generally cannot do it right, do not do it for now"

michpappas (approved):

lib/ukschedcoop/include/uk/schedcoop.h
skuenzer: I understand what you mean. For now, the creation functions for scheduler initialization are custom and defined by the scheduler implementation, likewise the allocator initialization. There is no common init interface layout defined yet for schedulers and allocators. I am also not sure if we could define one because schedulers and allocators could handle very specific cases with very custom parameters.

skuenzer: Concluding: I think it is okay to expand this definition with this PR. The only place where the scheduler is initialized is in ukboot atm, impact should be minimal for this particular case.

lib/ukboot/boot.c
```diff
@@ -64,6 +65,11 @@
 #include <uk/tinyalloc.h>
 #define uk_alloc_init uk_tinyalloc_init
 #endif
+#include <uk/allocstack.h>
```
skuenzer, line 68:
```c
#if CONFIG_LIBUKBOOT_ALLOCSTACK
```
etc. is missing.

skuenzer, line 69: The following patch makes ukboot still compile even without allocator initialzation:
```diff
--- a/lib/ukboot/boot.c
+++ b/lib/ukboot/boot.c
@@ -65,11 +65,13 @@
 #include <uk/tinyalloc.h>
 #define uk_alloc_init uk_tinyalloc_init
 #endif
+#if CONFIG_LIBUKBOOT_ALLOCSTACK
 #include <uk/allocstack.h>
 #if CONFIG_LIBUKBOOT_ALLOCSTACK_PREMAP_ORDER
 #define ALLOCSTACK_INITIAL_SIZE                                \
        (PAGE_SIZE * (1 << CONFIG_LIBUKBOOT_ALLOCSTACK_PREMAP_ORDER))
 #endif /* CONFIG_LIBUKBOOT_ALLOCSTACK_PREMAP_ORDER */
+#endif /* CONFIG_LIBUKBOOT_ALLOCSTACK */
 #if CONFIG_LIBUKSCHED
 #include <uk/sched.h>
 #endif /* CONFIG_LIBUKSCHED */
```

include/uk/plat/config.h
```diff
@@ -42,5 +42,8 @@
 	__CPU_EXCEPT_STACK_SIZE_PAGE_ORDER
 #define CPU_EXCEPT_STACK_SIZE			__CPU_EXCEPT_STACK_SIZE
 
+#define AUXSTACK_SIZE_PAGE_ORDER		\
+	__AUXSTACK_SIZE_PAGE_ORDER
+#define AUXSTACK_SIZE				__AUXSTACK_SIZE
```
skuenzer, line 47: With re-arch we definitely have to fix the namespacing of the definitions in this file. For now, okay since the newly introduced parameters follows the given examples.

lib/ukvmem/include/uk/vma_types.h
```diff
@@ -86,66 +86,83 @@ static inline int uk_vma_map_anon(struct uk_vas *vas, __vaddr_t *vaddr,
...
  *
- * Stack VMAs are never merged or split to keep the guard page intact.
+ * Stack VMAs are never merged or split to keep the guard pages intact.
  */
 extern const struct uk_vma_ops uk_vma_stack_ops;
 
 /* Stack flags */
 #define UK_VMA_STACK_GROWS_UP		(0x1UL << UK_VMA_MAP_EXTF_SHIFT)
 
+#define STACK_TOP_GUARD_SIZE				\
```
skuenzer: Could we at least namespace those, like `UK_VMA_STACK_TOP_GUARD_SIZE`? We have previous definitions in this file already namespaced.

lib/ukallocstack/include/uk/allocstack.h
```diff
@@ -0,0 +1,63 @@
...
+ *     uk_malloc(<SOME_STACK_SIZE> + PAGE_SIZE) would return a stack whose all
+ *     pages, except the first one, are paged-in. Consequently if allocating
+ *     a stack whose size is smaller than premapped_len, it means the whole
+ *     stack will be paged-in.
+ * @return
+ *     Returns the stack allocator or NULL on error
+ */
+struct uk_alloc *uk_allocstack_init(struct uk_alloc *a
+#if CONFIG_LIBUKVMEM
+				     , struct uk_vas __maybe_unused *vas,
```
skuenzer, line 55: Can this `maybe_unused` really happen? With `CONFIG_LIBUKVMEM` on, it should be used always, right? I would also prefer if we add this attribute to the final function declaration instead.

skuenzer (approved):
Thanks a lot for this work!

PR #1328, opened 2024-02-19
drivers/virtio/net: Ensure half-page alignment for netbuf

mogasergiu:
Currently there is a bug where if the net buffer (of size 2048) is split into more than one descriptor the VMM reports the vring as full and leading to us returning an error and wasting IRQs. The buffer would be split in multiple virtio descriptors if the scatter-gather list notices its span crosses more than one page (i.e. the buffer is not entirely contained in one page) since it is not guaranteed that the two virtually contiguous pages are also contiguous physically. To avoid having this happen make sure that the buffer is always contained in one page by having it aligned to half a page: if the buffer is half a page in size and it is also half a page aligned it is mathematically guaranteed to be entirely contained in one single page.

### Prerequisite checklist

### Base target

### DIFF
--- a/drivers/virtio/net/virtio_net.c
+++ b/drivers/virtio/net/virtio_net.c
@@ -50,6 +50,21 @@
 	 (size_t)virtio_net_hdr_size(_vndev))
 
 /**
+ * Currently there is a bug where if the net buffer (of size 2048) is split
+ * into more than one descriptor the VMM reports the vring as full and leading
+ * to us returning an error and wasting IRQs. The buffer would be split in
+ * multiple virtio descriptors if the scatter-gather list notices its span
+ * crosses more than one page (i.e. the buffer is not entirely contained in
+ * one page) since it is not guaranteed that the two virtually contiguous pages
+ * are also contiguous physically. To avoid having this happen make sure that
+ * the buffer is always contained in one page by having it aligned to half a
+ * page: if the buffer is half a page in size and it is also half a page aligned
+ * it is mathematically guaranteed to be entirely contained in one single
+ * page.
+ */
+#define VIRTIO_PKT_BUFFER_ALIGN			2048
+
+/**
  * When mergeable buffers are not negotiated, the virtio_net_hdr_padded struct
  * below is placed at the beginning of the netbuf data. Use 4 bytes of pad to
  * both keep the VirtIO header and the data non-contiguous and to keep the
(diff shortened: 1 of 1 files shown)


jschlumpp:
An alternative to this would be to implementing indirect descriptors (We would need an allocation for that...)

drivers/virtio/net/virtio_net.c
```diff
@@ -49,6 +50,21 @@
...
+ * multiple virtio descriptors if the scatter-gather list notices its span
+ * crosses more than one page (i.e. the buffer is not entirely contained in
+ * one page) since it is not guaranteed that the two virtually contiguous pages
+ * are also contiguous physically. To avoid having this happen make sure that
+ * the buffer is always contained in one page by having it aligned to half a
+ * page: if the buffer is half a page in size and it is also half a page aligned
+ * it is mathematically guaranteed to be entirely contained in one single
+ * page.
+ */
+#define VIRTIO_PKT_BUFFER_ALIGN			(PAGE_SIZE / 2)
```
jschlumpp: I'm not sure why this is dependent on the PAGE_SIZE. I'm pretty sure the chosen buffer size is just the next power of two for ~1400byte (the default MTU for most networks)

jschlumpp (approved):
Just one really minor thing. Other than that I guess we have to live with this workaround for now.

drivers/virtio/net/virtio_net.c
mogasergiu: Yes, correct. In fact, it is related to PAGE_SIZE  because the buffer must be contained in one page and the fact that this next power of two is the size of that buffer. So the problem here is actually the fact that we require the start address of the buffer would be so that its span does not cross into another page. So this macro should be defined as the smallest divisor of PAGE_SIZE, that is bigger than the buffer size (in this case the 2048 buffer). This happens to be 2048.

How about the following macro:
```C
#define VIRTIO_PKT_BUFFER_ALIGN(bufsz)                               \
               (bufsz < PAGE_SIZE ? (1 << uk_ffs(bufsz & (~1))) : PAGE_SIZE)
```

Assigning `PAGE_SIZE` if bigger than `PAGE_SIZE` because that is the best we can get w.r.t. amount of pages used.

jschlumpp: You don't know the buffer size at that point. It's network stack specified. So we have to hard-code something already there.
If you can't set this to 2048 as a constant because you might run over the actual page size, you already have a big problem. Then you will run into the situation where you split packets because the usual MTU will cause the split already.

mogasergiu: But it's also here `VIRTIO_PKT_BUFFER_LEN` in `virtio-net`. And at that point the features are negotiated, right? So we will know the size right?

jschlumpp: But that is just a very arbitrary size with MTU=1500 and also has no hard relationship to what the network stack will allocate. So you can also just go for 2048 directly.

mogasergiu:
Hmm hold on now I can also remove the paging header

eduardvintila (approved):
Looks good, thanks!

razvand (approved):

PR #1339, opened 2024-02-26
drivers/virtio: Implement LRO and RX buffer merging

jschlumpp:
### Prerequisite checklist

### Base target

### Additional configuration

To test this PR you can use the corresponding LWIP PR: https://github.com/unikraft/lib-lwip/pull/53
When receiving large enough payloads via TCP, QEMU/Linux should then pass on packets size much larger than the configured MTU to the VM.

### Description of changes

This PR adds LRO support to the virtio net driver and the necessary abstraction in uknetdev. To avoid having to allocate huge buffers it also implements support for the RX buffer merge feature. This feature allows the device to join multiple receive buffers together for packets larger than the buffer size.
Because this feature triggered a bug in the event notification code, it also includes a fix for that problem.

### DIFF
--- a/drivers/virtio/net/virtio_net.c
+++ b/drivers/virtio/net/virtio_net.c
@@ -178,7 +178,9 @@ static int virtio_netdev_configure(struct uk_netdev *n,
 				   const struct uk_netdev_conf *conf);
 static int virtio_netdev_rxtx_alloc(struct virtio_net_device *vndev,
 				    const struct uk_netdev_conf *conf);
-static int virtio_netdev_feature_negotiate(struct uk_netdev *n);
+static int virtio_netdev_probe(struct uk_netdev *n);
+static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
+					   const struct uk_netdev_conf *conf);
 static struct uk_netdev_tx_queue *virtio_netdev_tx_queue_setup(
 					struct uk_netdev *n, __u16 queue_id,
 					__u16 nb_desc,
@@ -908,12 +910,12 @@ static __u16 virtio_net_mtu_get(struct uk_netdev *n)
 	return d->mtu;
 }
 
-static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
+static int virtio_netdev_probe(struct uk_netdev *n)
 {
-	__u64 host_features = 0;
-	__u64 drv_features  = 0;
-	int rc = 0;
 	struct virtio_net_device *vndev;
+	__u64 drv_features = 0;
+	__u64 host_features;
+	int rc;
 
 	UK_ASSERT(n);
 	vndev = to_virtionetdev(n);
@@ -998,10 +1000,35 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 	if (VIRTIO_FEATURE_HAS(host_features, VIRTIO_F_EVENT_IDX))
 		VIRTIO_FEATURE_SET(drv_features, VIRTIO_F_EVENT_IDX);
 
+	/* Store the preliminary result in the features field.
+	 * virtio_netdev_feature_negotiate will take care of sending the result
+	 * to the host depending on how the uknetdev client configured the
+	 * device.
+	 */
+	vndev->vdev->features = drv_features;
+
+	return 0;
+err_negotiate_feature:
+	virtio_dev_status_update(vndev->vdev, VIRTIO_CONFIG_STATUS_FAIL);
+	return rc;
+}
+
+static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
+					   const struct uk_netdev_conf *conf)
+{
+	struct virtio_net_device *vndev;
+	__u64 host_features;
+	int rc;
+
+	UK_ASSERT(n);
+	UK_ASSERT(conf);
+	vndev = to_virtionetdev(n);
+
+	host_features = virtio_feature_get(vndev->vdev);
+
 	/**
 	 * Announce our enabled driver features back to the backend device
 	 */
-	vndev->vdev->features = drv_features;
 	virtio_feature_set(vndev->vdev);
 
 	/**
@@ -1013,11 +1040,11 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
 	 * Currently, unaligned read is supported in the underlying function.
 	 */
 	virtio_config_get(vndev->vdev,
-				   __offsetof(struct virtio_net_config, mac),
-				   &vndev->hw_addr.addr_bytes[0],
-				   UK_NETDEV_HWADDR_LEN, 1);
+			  __offsetof(struct virtio_net_config, mac),
+			  &vndev->hw_addr.addr_bytes[0],
+			  UK_NETDEV_HWADDR_LEN, 1);
 
-	if (VIRTIO_FEATURE_HAS(drv_features, VIRTIO_NET_F_MTU)) {
+	if (VIRTIO_FEATURE_HAS(vndev->vdev->features, VIRTIO_NET_F_MTU)) {
 		virtio_config_get(vndev->vdev,
 				  __offsetof(struct virtio_net_config, mac),
 				  &vndev->mtu, sizeof(vndev->mtu), 1);
@@ -1136,8 +1163,14 @@ static int virtio_netdev_configure(struct uk_netdev *n,
 	UK_ASSERT(conf);
 	vndev = to_virtionetdev(n);
 
+	rc = virtio_netdev_feature_negotiate(n, conf);
+	if (unlikely(rc < 0)) {
+		uk_pr_err("%p: Failed to negotiate features: %d\n", n, rc);
+		return rc;
+	}
+
 	rc = virtio_netdev_rxtx_alloc(vndev, conf);
-	if (rc < 0) {
+	if (unlikely(rc < 0)) {
 		uk_pr_err("%p: Failed to initialize rx and tx rings: %d\n",
 			  n, rc);
 	}
@@ -1247,7 +1280,7 @@ static int virtio_net_start(struct uk_netdev *n)
 }
 
 static const struct uk_netdev_ops virtio_netdev_ops = {
-	.probe = virtio_netdev_feature_negotiate,
+	.probe = virtio_netdev_probe,
 	.configure = virtio_netdev_configure,
 	.rxq_configure = virtio_netdev_rx_queue_setup,
 	.txq_configure = virtio_netdev_tx_queue_setup,
--- a/drivers/virtio/net/virtio_net.c
+++ b/drivers/virtio/net/virtio_net.c
@@ -165,6 +165,10 @@ struct virtio_net_device {
 	__u8 state;
 	/* RX promiscuous mode. */
 	__u8 promisc : 1;
+	/* Expected amount of descriptors per buffer */
+#define VIRTIO_NET_BUF_DESCR_COUNT_INLINE		1
+#define VIRTIO_NET_BUF_DESCR_COUNT_SEPARATE		2
+	__u8 buf_descr_count: 2;
 };
 
 /**
@@ -317,16 +321,17 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
 	__u16 filled = 0;
 
 	/**
-	 * Fixed amount of memory is allocated to each received buffer. In
-	 * our case since we don't support jumbo frame or LRO yet we require
-	 * that the buffer feed to the ring descriptor is atleast
-	 * ethernet MTU + virtio net header.
-	 * Because we using 2 descriptor for a single netbuf, our effective
-	 * queue size is just the half.
+	 * The effective queue size depends on how we have to add the buffers
+	 * into the ring. For example modern virtio devices only need a single
+	 * descriptor for the virtio header + payload. However, depending on the
+	 * physical layout of the buffer, we can end up using more descriptors.
+	 * In this case we batch allocate too many buffers and have to free the
+	 * superfluous ones.
 	 */
-	nb_desc = ALIGN_DOWN(nb_desc, 2);
+	UK_ASSERT(POWER_OF_2(vndev->buf_descr_count));
+	nb_desc = ALIGN_DOWN(nb_desc, vndev->buf_descr_count);
 	while (filled < nb_desc) {
-		req = MIN(nb_desc / 2, RX_FILLUP_BATCHLEN);
+		req = MIN(nb_desc / vndev->buf_descr_count, RX_FILLUP_BATCHLEN);
 		cnt = rxq->alloc_rxpkts(rxq->alloc_rxpkts_argp, netbuf, req);
 		for (i = 0; i < cnt; i++) {
 			uk_pr_debug("Enqueue netbuf %"PRIu16"/%"PRIu16" (%p) to virtqueue %p...\n",
@@ -345,7 +350,13 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
 				status |= UK_NETDEV_STATUS_UNDERRUN;
 				goto out;
 			}
-			filled += 2;
+			/* This count is not necessarily correct in the
+			 * presence of paging, but should be close enough as an
+			 * approximation.
+			 * TODO: Use indirect descriptors to ensure we are
+			 *       always below the specified count.
+			 */
+			filled += vndev->buf_descr_count;
 		}
 
 		if (unlikely(cnt < req)) {
@@ -358,7 +369,7 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
 
 out:
 	uk_pr_debug("Programmed %"PRIu16" receive netbufs to receive virtqueue %p (status %x)\n",
-		    filled / 2, rxq, status);
+		    filled / vndev->buf_descr_count, rxq, status);
 
 	/**
 	 * Notify the host, when we submit new descriptor(s).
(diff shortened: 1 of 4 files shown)


drivers/virtio/net/virtio_net.c
```diff
@@ -893,12 +895,12 @@ static __u16 virtio_net_mtu_get(struct uk_netdev *n)
...
-static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
+static int virtio_netdev_probe(struct uk_netdev *n)
 {
-	__u64 host_features = 0;
-	__u64 drv_features  = 0;
-	int rc = 0;
 	struct virtio_net_device *vndev;
+	__u64 drv_features = 0;
+	__u64 host_features;
+	int rc = 0;
```
mogasergiu: Hmm, I see in case of success you `return 0;` regardless. So this early initialization is probably not needed.

drivers/virtio/net/virtio_net.c
```diff
@@ -983,10 +985,34 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
...
+	virtio_dev_status_update(vndev->vdev, VIRTIO_CONFIG_STATUS_FAIL);
+	return rc;
+}
+
+static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
+					   const struct uk_netdev_conf *conf)
+{
+	struct virtio_net_device *vndev;
+	__u64 host_features;
+	int rc = 0;
```
mogasergiu: Ditto? `rc` does not seem to be set anywhere.

drivers/virtio/net/virtio_net.c
```diff
@@ -983,10 +985,34 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n)
...
+	/* Store the preliminary result in the features field.
+	 * virtio_netdev_feature_negotiate will take care of sending the result
+	 * to the host depending on how the uknetdev client configured the
+	 * device.
+	 */
+	vndev->vdev->features = drv_features;
+
+	return 0;
+err_negotiate_feature:
+	virtio_dev_status_update(vndev->vdev, VIRTIO_CONFIG_STATUS_FAIL);
```
mogasergiu, line 1086: Why do we not check for this function's return value? :thinking: 
Actually, it seems that the whole file is doing this???? Am I missing something?

drivers/virtio/net/virtio_net.c
```diff
@@ -150,6 +150,10 @@ struct virtio_net_device {
 	__u8 state;
 	/* RX promiscuous mode. */
 	__u8 promisc : 1;
+	/* Expected amount of descriptors per buffer */
+	__u8 buf_descr_count : 2;
+#define VIRTIO_NET_BUF_DESCR_COUNT_INLINE		1
```
mogasergiu, line 169: Probably subjective to my personal liking :nerd_face: :point_up:. But I find it more conventional that the bitfield def's are above the declaration of the actual field.

drivers/virtio/net/virtio_net.c
```diff
@@ -330,7 +334,13 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
 				status |= UK_NETDEV_STATUS_UNDERRUN;
 				goto out;
 			}
-			filled += 2;
+			/* This count is not necessarily correct in the
+			 * presence of paging, but should be close enough as an
+			 * approximation.
+			 * TODO: Use indirect descriptors to ensure we are
```
mogasergiu, line 356: We should probably also add an assertion here?

drivers/virtio/net/virtio_net.c
```diff
@@ -510,30 +520,41 @@ static int virtio_netdev_rxq_enqueue(struct virtio_net_device *vndev,
...
 
-	/* Appending the data buffer to the sglist */
-	uk_sglist_append(sg, buf_start, buf_len);
+		/**
+		 * Retrieve the buffer header length.
+		 */
+		rc = uk_netbuf_header(netbuf, VTNET_HDR_SIZE_PADDED(vndev));
+		if (unlikely(rc != 1)) {
+			uk_pr_err("Failed to allocate space to prepend virtio header\n");
+			return -EINVAL;
```
mogasergiu, line 564: Why not forward the actual `rc`? :thinking:

drivers/virtio/net/virtio_net.c
```diff
@@ -1078,6 +1082,34 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
...
+	 * NOTE: This allows the host to send packets larger than MTU. The
+	 *       network stack needs to be able to handle such packets.
+	 *       We only enable this if we have also support for mergeable RX
+	 *       buffers, otherwise we would have to either allocate huge
+	 *       buffers (wasting space for smaller buffers) or use many small
+	 *       descriptors (needing indirect descriptors and putting a large
+	 *       burden on the allocator).
+	 */
+	if (conf->lro) {
+		if (VIRTIO_FEATURE_HAS(vndev->vdev->features,
```
mogasergiu: Would it be worth adding a `likely` here? :thinking:

mogasergiu (changes requested):
Hey @mschlumpp :cactus: :wave: ! Thank you for the super useful feature :muscle:! I had quick diagonal look trying to spot what I would describe as nitpicks. I will try to have a deeper look in the next round.

drivers/virtio/net/virtio_net.c
jschlumpp: No, as the comment hints, having too many buffers is not fatal. Also the problem is that we don't have an precise number before trying to add the buffer. So adding an assertion would just cause random panics.

drivers/virtio/net/virtio_net.c
jschlumpp: I haven't written the code. I guess the intention is that if you give it a buffer that returns an error code here, it is an invalid argument error.

drivers/virtio/net/virtio_net.c
jschlumpp: Given the documentation, it either returns success or ENOTSUP. In the case of ENOTSUP, we wouldn't do anything. So we would ignore the return code anway.

drivers/virtio/net/virtio_net.c
jschlumpp: Actually let's also reorder the whole section ^^
Then we can remove one level of indentation on the success case.

drivers/virtio/net/virtio_net.c
mogasergiu: Sigh... the state of these virtio drivers man...

jschlumpp: That's not really one of the bad parts. Not supporting the status update is not a problem, so ignoring having no support is fine.

mogasergiu: Yes, agreed. I meant our virtio drivers in general feel a bit all over the place and there are other functions called whose returned values are not consistently checked for, like `uk_sglist_append` or `virtio_netdev_rx_fillup` or `virtio_dev_status_update`... but I digress.

mariapana (approved):
Apart from @mogasergiu 's remarks, it looks good. Moreover, I haven't found any problems during testing. Seems to work as expected :+1: 

drivers/virtio/net/virtio_net.c
```diff
@@ -322,9 +328,9 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
 	 * Because we using 2 descriptor for a single netbuf, our effective
```
Krechals: We should update the comment from above.

drivers/virtio/net/virtio_net.c
```diff
@@ -322,9 +328,9 @@ static int virtio_netdev_rx_fillup(struct virtio_net_device *vndev,
...
 	 * queue size is just the half.
 	 */
-	nb_desc = ALIGN_DOWN(nb_desc, 2);
+	nb_desc = ALIGN_DOWN(nb_desc, vndev->buf_descr_count);
```
Krechals, line 332: Is it guaranteed that `vndev->buf_descr_count` is a power of 2?

drivers/virtio/net/virtio_net.c
```diff
@@ -523,30 +535,41 @@ static int virtio_netdev_rxq_enqueue(struct virtio_net_device *vndev,
...
-	rxhdr = netbuf->data;
-
+	/* Reset the scatter gather list */
 	sg = &rxq->sg;
 	uk_sglist_reset(sg);
 
-	/* Appending the header buffer to the sglist */
-	uk_sglist_append(sg, rxhdr, virtio_net_hdr_size(vndev));
+	if (vndev->buf_descr_count == VIRTIO_NET_BUF_DESCR_COUNT_INLINE) {
+		rc = uk_netbuf_header(netbuf, virtio_net_hdr_size(vndev));
```
Krechals, line 544: Could you please help me understand what this branch covers?
I'm a bit lost in understanding the use case and why we are not using the padding from `VTNET_HDR_SIZE_PADDED`.

Krechals (commented):
Greetings, Marco! Thank you for adding this feature! I only managed to go over it briefly, but I'll look into it more thoroughly in the coming days.

I expect this feature will boost network throughput by a bit. Though I haven't had a chance to test this yet, have you noticed any improvements?

drivers/virtio/net/virtio_net.c
jschlumpp: Modern virtio devices (or legacy devices with negotiated ANY_LAYOUT, but we don't consider those) do not need to have the header in a separate descriptor. And if we put header and payload into the same descriptor then there must not be any padding between them.
Additionally, if we use mergeable RX buffers the resulting buffers would contain non-continuous payloads for the additional buffers (fixing that up would require `memmove`s)
Also this usually wastes one less descriptor.

drivers/virtio/net/virtio_net.c
jschlumpp: For now yes, but I'll insert an assert in front of it.

jschlumpp:
Updated and rebased the PR

drivers/virtio/net/virtio_net.c
```diff
@@ -1007,6 +1043,16 @@ static int virtio_netdev_probe(struct uk_netdev *n)
 	 */
 	vndev->vdev->features = drv_features;
 
+	if (vndev->vdev->features & (1ULL << VIRTIO_F_VERSION_1)
+	    || vndev->vdev->features & (1ULL << VIRTIO_NET_F_MRG_RXBUF)) {
+		/* Do not use a separate (padded) header descriptor for modern
+		 * devices or when we can use mergeable buffers
+		 */
+		vndev->buf_descr_count = 1;
```
mogasergiu, line 1079: Probably don't need the brackets anymore.

drivers/virtio/net/virtio_net.c
```diff
@@ -588,20 +610,34 @@ static int virtio_netdev_rxq_dequeue(struct virtio_net_device *vndev,
...
+			buf, -((__s16)virtio_net_hdr_size(vndev)));
+	} else {
+		/* The length tracked by the virtio driver does include the
+		 * header but *not the padding*, therefore adjust the netbuf
+		 * size and add the padding.
+		 */
+		buf->len = len + (VTNET_HDR_SIZE_PADDED(vndev) -
+			   virtio_net_hdr_size(vndev));
+		/* Shift the position forward to remove the header */
+		rc = uk_netbuf_header(
```
mogasergiu: paranthesis alignment: buf on this line and the second argument on the next.

drivers/virtio/net/virtio_net.c
```diff
@@ -588,20 +610,34 @@ static int virtio_netdev_rxq_dequeue(struct virtio_net_device *vndev,
...
 	 * Removing the virtio header from the buffer and adjusting length.
-	 * We pad the rx buffer while enqueuing for alignment of the packet
-	 * data. We compensate for this by subtracting the padding to the
-	 * length on dequeue.
 	 */
-	buf->len = len + VTNET_HDR_SIZE_PADDED(vndev);
-	rc = uk_netbuf_header(buf, -((__s16)VTNET_HDR_SIZE_PADDED(vndev)));
+	if (vndev->buf_descr_count == VIRTIO_NET_BUF_DESCR_COUNT_INLINE) {
+		buf->len = len;
+		rc = uk_netbuf_header(
```
mogasergiu: ditto

drivers/virtio/net/virtio_net.c
```diff
@@ -1007,6 +1043,16 @@ static int virtio_netdev_probe(struct uk_netdev *n)
 	 */
 	vndev->vdev->features = drv_features;
 
+	if (vndev->vdev->features & (1ULL << VIRTIO_F_VERSION_1)
```
mogasergiu: I would place paranthesis around both these bitwise `&`'s just to be double sure :man_shrugging:.

drivers/virtio/net/virtio_net.c
```diff
@@ -639,6 +642,20 @@ static int virtio_netdev_rxq_dequeue(struct virtio_net_device *vndev,
 			buf, -((__s16)VTNET_HDR_SIZE_PADDED(vndev)));
 	}
 	UK_ASSERT(rc == 1);
+
+	while (num_buffers > 1) {
+		ret = virtqueue_buffer_dequeue(rxq->vq, (void **) &chain, &len);
+		if (ret < 0) {
```
mogasergiu: `unlikely`? :thinking:

drivers/virtio/net/virtio_net.c
```diff
@@ -525,30 +536,41 @@ static int virtio_netdev_rxq_enqueue(struct virtio_net_device *vndev,
...
 
-	/* Appending the header buffer to the sglist */
-	uk_sglist_append(sg, rxhdr, virtio_net_hdr_size(vndev));
+	if (vndev->buf_descr_count == VIRTIO_NET_BUF_DESCR_COUNT_INLINE) {
+		rc = uk_netbuf_header(netbuf, virtio_net_hdr_size(vndev));
+		if (unlikely(rc != 1)) {
+			uk_pr_err("Failed to allocate space to prepend virtio header\n");
+			return -EINVAL;
+		}
+		uk_sglist_append(sg, netbuf->data, netbuf->len);
```
mogasergiu, line 549: A lot of sglist appends whose return value is not checked around this driver... :thinking:

drivers/virtio/net/virtio_net.c
```diff
@@ -639,6 +642,20 @@ static int virtio_netdev_rxq_dequeue(struct virtio_net_device *vndev,
 			buf, -((__s16)VTNET_HDR_SIZE_PADDED(vndev)));
 	}
 	UK_ASSERT(rc == 1);
+
+	while (num_buffers > 1) {
+		ret = virtqueue_buffer_dequeue(rxq->vq, (void **) &chain, &len);
```
mogasergiu: no whitespace after cast.

drivers/virtio/net/virtio_net.c
```diff
@@ -596,8 +596,12 @@ static int virtio_netdev_rxq_dequeue(struct virtio_net_device *vndev,
 		*netbuf = NULL;
 		return rxq->nb_desc;
 	}
-	if (unlikely((len < (__u32)virtio_net_hdr_size(vndev) + UK_ETH_HDR_UNTAGGED_LEN) ||
-		     len > VIRTIO_PKT_BUFFER_LEN(vndev))) {
+	if (unlikely(
+		(len < (__u32)virtio_net_hdr_size(vndev) +
```
mogasergiu: Why not keep this on the same line as the `unlikely`?

drivers/virtio/net/virtio_net.c
```diff
@@ -1094,6 +1098,33 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
...
+	 *       network stack needs to be able to handle such packets.
+	 *       We only enable this if we have also support for mergeable RX
+	 *       buffers, otherwise we would have to either allocate huge
+	 *       buffers (wasting space for smaller buffers) or use many small
+	 *       descriptors (needing indirect descriptors and putting a large
+	 *       burden on the allocator).
+	 */
+	if (conf->lro) {
+		if (unlikely(
+			!VIRTIO_FEATURE_HAS(vndev->vdev->features,
```
mogasergiu: Ditto

drivers/virtio/net/virtio_net.c
```diff
@@ -1094,6 +1098,33 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
...
+				VIRTIO_NET_F_GUEST_CSUM) ||
+			!VIRTIO_FEATURE_HAS(vndev->vdev->features,
+				VIRTIO_NET_F_MRG_RXBUF))) {
+			rc = -EINVAL;
+			goto err_negotiate_feature;
+		}
+		if (VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_GUEST_TSO4))
+			VIRTIO_FEATURE_SET(vndev->vdev->features,
+				   VIRTIO_NET_F_GUEST_TSO4);
```
mogasergiu: Paranthesis alignment?

drivers/virtio/net/virtio_net.c
```diff
@@ -1094,6 +1098,33 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
...
+		if (VIRTIO_FEATURE_HAS(host_features, VIRTIO_NET_F_GUEST_TSO6))
+			VIRTIO_FEATURE_SET(vndev->vdev->features,
+				   VIRTIO_NET_F_GUEST_TSO6);
```
mogasergiu: Ditto

drivers/virtio/net/virtio_net.c
```diff
@@ -1094,6 +1098,33 @@ static int virtio_netdev_feature_negotiate(struct uk_netdev *n,
...
+	 * Large Receive Offload
+	 * NOTE: This allows the host to send packets larger than MTU. The
+	 *       network stack needs to be able to handle such packets.
+	 *       We only enable this if we have also support for mergeable RX
+	 *       buffers, otherwise we would have to either allocate huge
+	 *       buffers (wasting space for smaller buffers) or use many small
+	 *       descriptors (needing indirect descriptors and putting a large
+	 *       burden on the allocator).
+	 */
+	if (conf->lro) {
```
mogasergiu, line 1113: Assert for conf?

drivers/virtio/net/virtio_net.c
```diff
@@ -1291,10 +1322,13 @@ static void virtio_net_info_get(struct uk_netdev *dev,
 				struct uk_netdev_info *dev_info)
 {
 	struct virtio_net_device *vndev;
+	__u64 host_features;
 
 	UK_ASSERT(dev && dev_info);
 	vndev = to_virtionetdev(dev);
 
+	host_features = virtio_feature_get(vndev->vdev);
```
mogasergiu, line 1331: Why not use the cached features from `vndev->vdev->features`?

mogasergiu (changes requested):
Hey Marco, had another look :wave:

drivers/virtio/net/virtio_net.c
jschlumpp: I think I was adapting the new code a bit too much to the existing one. But yeah that is indeed a bit dangerous. I think it's because the driver predates paging and without paging you never have to worry about running out of sglist elements.

drivers/virtio/net/virtio_net.c
jschlumpp: That field contains the already negotiated features, not the features that are available. For `info_get` the client wants to know what features it could enable (some features such as LRO might be not supported by a specific network stack). Afterwards, it will call the configure function and in there the "client optional" features are set in `vndev->vdev->features`.

drivers/virtio/net/virtio_net.c
mogasergiu: Yeah maybe not worth bothering with this detail now. I am fine whatever you choose, I just hope we do not end up at some point debugging a difficult bug that we cannot pinpoint because we did not have an error check here. The way this whole driver looks is burning my eyes.

drivers/virtio/net/virtio_net.c
jschlumpp: What do you mean? The curly braces? I think there are various opinions about whether a comment counts as multi-line or not (visually it would, the compiler obviously ignores the comment). 
*sigh* this whole single-statement thing without curly braces is just stupid because of all these corner cases...

drivers/virtio/net/virtio_net.c
jschlumpp: Because it feels much more easier to parse on a quick look.... But yeah I know coding conventions are more important 🙄

drivers/virtio/net/virtio_net.c
mogasergiu: Hmm, you are right, it would probably look more visually pleasing with curly braces here. I just hope checkpatch does not ever complain.

mogasergiu (approved):

michpappas:

PR #1341, opened 2024-02-26
lib/ukfile: Add opt-in support for file finalizers

andreittr:
### Description of changes

This change adds optional support for file finalizers -- custom functions registered to run when the last strong reference to a file is released. These can be useful for e.g., automatically removing a closed file from a polling pool.

Since this feature adds some overhead and may not be always required, it is gated behind the LIBUKFILE_FINALIZERS config option. With this option in its default disabled state, behavior and mem usage is as before.

This commit changes the driver API of ukfile, specifically its refcount initializers. Affected consumers of ukfile have also been patched.

Potential conflict w/ https://github.com/unikraft/unikraft/pull/1226; whichever merges first will require a patch to the other.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

N/A

### DIFF
--- a/lib/posix-eventfd/eventfd.c
+++ b/lib/posix-eventfd/eventfd.c
@@ -137,7 +137,7 @@ struct uk_file *uk_eventfile_create(unsigned int count, int flags)
 	al->alloc = a;
 	al->counter = count;
 	al->fstate = UK_FILE_STATE_INIT_VALUE(al->fstate);
-	al->frefcnt = UK_FILE_REFCNT_INIT_VALUE;
+	al->frefcnt = UK_FILE_REFCNT_INIT_VALUE(al->frefcnt);
 	al->f = (struct uk_file){
 		.vol = vol,
 		.node = &al->counter,
--- a/lib/posix-pipe/pipe.c
+++ b/lib/posix-pipe/pipe.c
@@ -292,8 +292,8 @@ int uk_pipefile_create(struct uk_file *pipes[2], int flags)
 	al->node.whead = 0;
 	memset(al->node.buf, 0, sizeof(al->node.buf));
 	al->fstate = UK_FILE_STATE_INIT_VALUE(al->fstate);
-	al->rref = UK_FILE_REFCNT_INIT_VALUE;
-	al->wref = UK_FILE_REFCNT_INIT_VALUE;
+	al->rref = UK_FILE_REFCNT_INIT_VALUE(al->rref);
+	al->wref = UK_FILE_REFCNT_INIT_VALUE(al->wref);
 	al->rf = (struct uk_file){
 		.vol = PIPE_VOLID,
 		.node = &al->node,
--- a/lib/posix-poll/epoll.c
+++ b/lib/posix-poll/epoll.c
@@ -378,7 +378,7 @@ struct uk_file *uk_epollfile_create(void)
 	al->alloc = a;
 	al->list = NULL;
 	al->fstate = UK_FILE_STATE_INIT_VALUE(al->fstate);
-	al->frefcnt = UK_FILE_REFCNT_INIT_VALUE;
+	al->frefcnt = UK_FILE_REFCNT_INIT_VALUE(al->frefcnt);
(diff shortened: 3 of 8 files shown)


John-Ted:
On my end, it did not change anything, but I cannot see the LIBUKFILE_FINALIZERS config option in menuconfig, I had to manually add it in the .config file. Is there something else that needs to be enabled for it to show up (it does not have any dependencies listed) ?

mariapana:
> On my end, it did not change anything, but I cannot see the LIBUKFILE_FINALIZERS config option in menuconfig, I had to manually add it in the .config file. Is there something else that needs to be enabled for it to show up (it does not have any dependencies listed) ?

If I remember correctly (and please correct me if I am wrong), the `kraft.yaml` file needs to be updated in order for the option to show up.

andreittr:
> On my end, it did not change anything, but I cannot see the LIBUKFILE_FINALIZERS config option in menuconfig, I had to manually add it in the .config file. Is there something else that needs to be enabled for it to show up (it does not have any dependencies listed) ?

This new option is [hidden](https://github.com/unikraft/unikraft/pull/1341/files#diff-909f03394f780911be2e3127172356873f60edd6e8d07ca3e4361913feff90beR8) since it doesn't expose any directly user-visible behavior, only an extension to the `uk_file` API. This is similar to the config option right above: `LIBUKFILE_CHAINUPDATE`.
Rather than enabling manually, it is meant to be selected by other Unikraft core components (or libs, or apps) when these API extensions are used; ~~I'm working on a PR for such a component atm and I'll post a link to a branch as soon as it's up~~ [example branch](https://github.com/unikraft/unikraft/commit/5ab7a29fb37ab7bb28c09f4930b847a06ab4ec43) is up.
In the meantime you can check out [an example of `LIBUKFILE_CHAINUPDATE` being used in `posix-poll`](https://github.com/unikraft/unikraft/blob/staging/lib/posix-poll/Config.uk#L6).

> If I remember correctly (and please correct me if I am wrong), the kraft.yaml file needs to be updated in order for the option to show up.

I believe `kraft.yaml` only impacts the behavior of `kraft` and not the raw `make` config system (although I don't use `kraft menu` often enough to know for sure when it updates :shrug: ). Either way, it's normal for the option to not appear.

Edit: add link to future PR code using finalizers.

mariapana:
Looks good. Just one curiosity for my personal understanding, I see that in lib/ukfile/include/uk/file/final.h, line 80, you use a mutex. Wouldn't it be needed to be locked before UK_ASSERT to ensure proper verification, instead of afterwards? Thank you in advance!

andreittr:
> I see that in lib/ukfile/include/uk/file/final.h, line 80, you use a mutex. Wouldn't it be needed to be locked before UK_ASSERT to ensure proper verification, instead of afterwards?

Good question; in short it is indeed OK to assert before locking, and to see why let's review a couple of facts:
- the mutex exists to synchronize access to the finalizer registration list (`(struct uk_file_finref).fins`)
- reference counting is done independently from locking using atomic operations; it's why `uk_file_finref_acquire/release()` do not take the mutex either

Having the lock held would thus not prevent another thread from changing the refcount, so asserting before or after locking shouldn't make a difference. In that case, we want to keep the critical section as short as possible => assert outside the locks.

(In a bit of a tangent) That assert isn't perfect however, consider:
1. the caller has a reference held => it is valid for the lifetime of the call => refcount will always read positive => OK
2. the caller does not have a reference held => refcount _might_ become 0 at any time, but not guaranteed => assert may or may not catch it

As it is now we can only check whether there is _a_ reference held, but ideally we'd want to check whether _the caller_ holds a reference.

mariapana:
I see now, makes sense :) Thank you for the insight!

mariapana (approved):
Looks good.

John-Ted (approved):
Looks good to me

StefanJum (approved):
All good, thanks.

razvand (approved):

PR #1346, opened 2024-03-04
{lib,arch,plat}: Redo syscall ctx's and `swapgs` logic

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

To make git bisecting and rebasing significantly easier and avoid
builds breaking across commits, this whole set of changes shall be
introduced under one single all encompassing commit.

Following the introduction of the concept of auxiliary stack pointers,
swapgs, `struct uk_syscall_ctx` and `struct ukarch_sysregs`, a number
of things have emerged:
- the aforemenetioned structs are very generic so they should be moved
under libcontext (arch/)
- swapgs introduces a significant inconsistency between ARM64 and x86_64
as we never know during an exception the state of
MSR_GS_BASE/MSR_KERNEL_GS_BASE
- auxiliary stack pointers  have increased flexibility as every thread
and LCPU can have one and have private data stored in there than may
be accessed anytime, dependency free

Thus, this commit does the following:
1. Move/rename aforementioned structured to libcontext and document them
- lib/syscall_shim/arch/x86_64/sysregs.c -> arch/x86/sysctx.c
- lib/syscall_shim/arch/x86_64include/arch/sysregs.h -> arch/x86/x86_64include/uk/asm/sysctx.h
- s/struct ukarch_sysregs/struct ukarch_sysctx/ (and all related defs)
- struct uk_syscall_ctx from lib/syscall_shim/include/uk/syscall.h to
include/uk/arch/ctx.h as struct ukarch_execenv
- s/struct uk_syscall_ctx/struct ukarch_execenv/ (and all related defs)
- actually comment these functions
- re-adjust all places that make use of such definitions

2. Get rid of the `swapgs`, architecture specific holdback by exploiting
the flexibility of auxiliary stacks through the introduction of a new
always existing contrl block at their top end:
- introduce `struct ukarch_auxspcb` under libcontext
- add Unikraft system context as field to it so that we always have and
know Unikraft TLS (and LCPU in case ox x86_64) in a dependency free
and assumption free manner
- add a current frame pointer field: since the auxspscb will be part of
the auxiliary stack, we need to know the safe place where we can start
using the auxiliary stack area as a stack (this is also helpful in cases
where we need to nest on the auxstack)
-for the aforementione fields/structs, init/getter/setter functions have
been added and documented
- now the `swapgs` pair will only be done very early during system call
entry (and only there, not on clone child exit anymore either) just
enough so that we, first things first, switch to auxstack and push auxsp
so that on entry to C handler we will know that we must do a call to
`ukarch_sysctx_load` on the Unikraft sysctx we can get from the pushed
auxsp (another benefit of this is we get rid of MSR read/writes)


michpappas (changes requested):
@mogasergiu I had a brief look, this is really great. Similarly to the dependencies of this PR, I would kindly ask that what conceptually comprises of `libukctx` is moved out of `arch` to a concrete library. Besides reasons stated in the other PRs that will also make reviewing easier as now the implementation is scattered in various files around `arch`. I hope now with all in a single commit this won't be as much trouble. Thanks a lot! 😃 🙏🏼 

**Update:** After offline discussion also involving @skuenzer consensus was to merge this as is, and migrate into the `uklcpu` library as soon as the API gets stable.

include/uk/arch/ctx.h
```diff
@@ -56,12 +57,76 @@
 #define UKARCH_CTX_OFFSETOF_SP 8
 #endif
 
+#define	UKARCH_AUXSPCB_SIZE					\
```
michpappas: Since this is the public API, and effectively the first point of reference, please add some comments when introducing new terms, as things like `CB` and `UKSC` are not necessarily obvious. Even better if the code is self-explanatory and variables are named more intuitively. I bet I'm not the only one who thinks  "callback" when they see `CB`. As these are part of the API, they should use more descritive names: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#naming. See also the comment below about variables that use initials.

Also IMO it would be easier to read if `AUXSPCB_SIZE` -> `AUXSP_CB_SIZE` (and `auxspcb` -> `auxsp_cb` in general), but as always these things are a quite subjective (something to consider when moving to a library).

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * sets up the current frame pointer inside the auxstack area.
+ *
+ * @param auxsp
+ *   The auxiliary stack pointer to initialize. Must point to the high end of
+ *  the auxiliary stack.
+ *
+ */
+static inline void ukarch_auxsp_init(__uptr auxsp)
+{
+	struct ukarch_auxspcb *cauxspcb;
```
michpappas: What is a "cauxspcb"?! These variable names make no sense, please use meaningful names.

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
  */
 void ukarch_ctx_switch(struct ukarch_ctx *store, struct ukarch_ctx *load);
 
+/**
+ * Initialize and auxiliary stack pointer. At the moment, this function only
+ * sets up the current frame pointer inside the auxstack area.
+ *
+ * @param auxsp
```
michpappas, line 346: Please document the contstraints, i.e. that `auxsp` must be aligned to `AUXSP_ALIGN`.

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ *   The auxiliary stack pointer to initialize. Must point to the high end of
+ *  the auxiliary stack.
+ *
+ */
+static inline void ukarch_auxsp_init(__uptr auxsp)
```
michpappas, line 353: Just an idea, I wonder if adding a `length` parameter would be useful for forwards compatibility, although we don't use it now. If you want to be pedantic you can even add an assertion that `length > sizeof(struct ukarch_auxspcb)`. But if you think it's unnecessary just resolve this one.

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
  */
 void ukarch_ctx_switch(struct ukarch_ctx *store, struct ukarch_ctx *load);
 
+/**
+ * Initialize and auxiliary stack pointer. At the moment, this function only
```
michpappas: "Initialize and ..." something is missing here :) Also documenting the implementation here does not provide any value to the user (and over time you have to keep it in sync with the implementation).

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+	UK_ASSERT(auxsp);
+	UK_ASSERT(IS_ALIGNED(auxsp, UKARCH_AUXSP_ALIGN));
+
+	cauxspcb = (struct ukarch_auxspcb *)(auxsp - sizeof(*cauxspcb));
+	cauxspcb->curr_frame = auxsp - sizeof(*cauxspcb);
+	UK_ASSERT(IS_ALIGNED(cauxspcb->curr_frame, UKARCH_AUXSP_ALIGN));
+}
+
+/**
+ * Set the control block of the auxiliary stack pointer. At the moment,
```
michpappas: The same comment with the previous function on documenting the implementation applies here too. The title says it all already.

include/uk/arch/ctx.h
```diff
@@ -56,12 +57,76 @@
...
+};
+
+#define SP_IN_AUXSP(sp, auxsp)					\
+	(IN_RANGE((sp), (auxsp) - AUXSTACK_SIZE, AUXSTACK_SIZE))
+
+struct ukarch_auxspcb {
+	/* Current safe frame pointer inside the auxiliary stack area */
+	__uptr curr_frame;
+	/* Unikraft system registers (e.g. TLS pointer) */
+	struct ukarch_sysctx uksc;
```
michpappas: Please do not invent variable names that use initials, these should be generally be avoided unless they correspond to established conventions like `sp` and `fp`. What I would do is use `fp` for `current_frame` ("safe" IMO is redundant) and something more descriptive for `usc` like `uk_sysctx`.

include/uk/arch/ctx.h
```diff
@@ -194,6 +259,174 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ *
+ */
+static inline struct ukarch_auxspcb *ukarch_auxsp_get_cb(__uptr auxsp)
+{
+	struct ukarch_auxspcb *cauxspcb;
+
+	UK_ASSERT(auxsp);
+	UK_ASSERT(IS_ALIGNED(auxsp, UKARCH_AUXSP_ALIGN));
+
+	cauxspcb = (struct ukarch_auxspcb *)(auxsp - sizeof(*cauxspcb));
```
michpappas: Since you use this in more or less every function, you may consider using a macro.

arch/arm/arm64/include/uk/asm/sysctx.h
```diff
@@ -0,0 +1,67 @@
...
+ * given as argument. Stores TLS pointer (TPIDR_EL0).
+ *
+ * @param sysctx
+ *   The system register context to store state in.
+ */
+void ukarch_sysctx_store(struct ukarch_sysctx *sysctx);
+
+/**
+ * Load the system context register state from the system context
+ * given as argument. Loads TLS pointer (MSR_FS_BASE) into its respective
```
michpappas, line 59: ©️ 🍝 ❓

arch/arm/arm64/include/uk/asm/sysctx.h
```diff
@@ -0,0 +1,67 @@
...
+struct ukarch_sysctx {
+	/* AAPCS's TLS pointer register */
+	__uptr tpidr_el0;
+
+	__u8 pad[8];	/* Make sure we are a multiple of 16 bytes */
+};
+
+UK_CTASSERT(sizeof(struct ukarch_sysctx) == UKARCH_SYSCTX_SIZE);
+
+/**
```
michpappas, line 27: These declarations are duplicated on both architectures. They should be defined once in `include/uk/arch/ctx.h` along with the rest of the `arch` API. I understand that the current approach is taken because x86_64 defines additional accessors for FS / GS. So the alternative would be to define a header only for x86, but would break the build because of `#include uk/asm/sysctx.h` so we would need to define an empty header for arm64. Given that none both approaches are ugly, and the problem will be solved once we migrate all this into a library (which should happen shortly), it doesn't matter which one we choose now. So we might as well keep it as it is 🤷🏼‍♂️

arch/arm/sysctx.c
```diff
@@ -0,0 +1,39 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
michpappas, line 1: I'm 99% sure you're aware already, but just in case: 0fcf7ea1b26b54bc247488d32a6b21fdfb624396 updates checkpatch to supress this error, so when you rebase to `staging` these will go away.

include/uk/arch/ctx.h
```diff
@@ -249,6 +482,14 @@ void ukarch_ectx_store(struct ukarch_ectx *state);
  */
 void ukarch_ectx_load(struct ukarch_ectx *state);
 
+/**
+ * Restores a given extended context on the currently executing CPU.
```
michpappas: "extended" here and below should be "execution".

lib/ukboot/Makefile.uk
```diff
@@ -5,6 +5,11 @@ CXXINCLUDES-$(CONFIG_LIBUKBOOT)		+= -I$(LIBUKBOOT_BASE)/include
 LIBUKBOOT_ASINCLUDES-y			+= -I$(UK_PLAT_COMMON_BASE)/include
 LIBUKBOOT_CINCLUDES-y			+= -I$(UK_PLAT_COMMON_BASE)/include
 
+# FIXME: Remove this and make the library self contained
```
michpappas: How will we remember to remove these when we move to `libuklcpu`? Same for the changes in the Makefile of `libuksched` and `libukschedcoop`. IIRC we have some similar ones in `drivers`. But maybe this requires no action for now given that someone sooner or later will notice this comment once things have moved out of plat anyway.

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -278,6 +279,11 @@ static inline int lcpu_current_is_bsp(void)
 	return lcpu_is_bsp(lcpu_get_current());
 }
 
+/**
+ * Return the auxstack pointer of the current LCPU.
+ */
+__uptr lcpu_get_auxsp(void);
```
michpappas: I understand that the new functions are named as `lcpu_` and not `uk_lcpu` to remain consistent with the current state, which is of course the right decision. But we need to fix them once we move to `libuklcpu`.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -53,24 +59,44 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
+	ukarch_ectx_sanitize((struct ukarch_ectx *)&execenv->ectx);
+	ukarch_ectx_store((struct ukarch_ectx *)&execenv->ectx);
+
+	ukarch_sysctx_store(&execenv->sysctx);
 
-	ukarch_sysregs_switch_uk(&usc->sysregs);
+	auxspcb = ukarch_auxsp_get_cb(usc->auxsp);
+	ukarch_sysctx_load(&auxspcb->uksc);
+
+#if CONFIG_LIBSYSCALL_SHIM_HANDLER_ULTLS
```
michpappas, line 98: Switching between UK / UL TLS could also be abstracted into an API.

plat/common/x86/syscall.S
```diff
@@ -65,20 +67,32 @@ ENTRY(_ukplat_syscall)
...
+
 	/* Describing the rsp relative to GS would make it necessary to emit
 	 * raw CFI. Instead of doing so, mark rsp as undefined temporarily
 	 */
 	.cfi_undefined rsp
 
+	subq	$(UKARCH_AUXSPCB_SIZE), %rsp
+	movq	UKARCH_AUXSPCB_OFFSETOF_CURR_FRAME(%rsp), %rsp
+
+	subq	$(UKARCH_EXECENV_END_ALIGN - 8), %rsp
```
michpappas, line 136: Please comment asm as much as possible, as it's not obvious to everyone why we subtract 8 here.

plat/common/x86/syscall.S
```diff
@@ -65,20 +67,32 @@ ENTRY(_ukplat_syscall)
 	 * in the Red Zone.
 	 */
 	movq	%gs:LCPU_AUXSP_OFFSET, %rsp
+
 	/* Describing the rsp relative to GS would make it necessary to emit
 	 * raw CFI. Instead of doing so, mark rsp as undefined temporarily
 	 */
 	.cfi_undefined rsp
 
+	subq	$(UKARCH_AUXSPCB_SIZE), %rsp
```
michpappas, line 76: I think you also need to update the documentation and graphic at line 99.

plat/common/x86/syscall.S
```diff
@@ -65,20 +67,32 @@ ENTRY(_ukplat_syscall)
...
+
+	subq	$(UKARCH_EXECENV_END_ALIGN - 8), %rsp
+	.cfi_adjust_cfa_offset (UKARCH_EXECENV_END_ALIGN - 8)
+
+	pushq	%gs:LCPU_AUXSP_OFFSET
+	.cfi_adjust_cfa_offset 8
+
+	swapgs
+
 	/* NOTE: We should normally align the stack before doing this
```
michpappas, line 160: Is this comment still valid?

plat/common/x86/syscall.S
```diff
@@ -65,20 +67,32 @@ ENTRY(_ukplat_syscall)
...
+	subq	$(UKARCH_AUXSPCB_SIZE), %rsp
+	movq	UKARCH_AUXSPCB_OFFSETOF_CURR_FRAME(%rsp), %rsp
+
+	subq	$(UKARCH_EXECENV_END_ALIGN - 8), %rsp
+	.cfi_adjust_cfa_offset (UKARCH_EXECENV_END_ALIGN - 8)
+
+	pushq	%gs:LCPU_AUXSP_OFFSET
+	.cfi_adjust_cfa_offset 8
+
+	swapgs
```
michpappas, line 158: Please leave a comment about why we are switching back to user GS.

plat/common/x86/syscall.S
```diff
@@ -65,20 +67,32 @@ ENTRY(_ukplat_syscall)
...
 	 */
 	.cfi_undefined rsp
 
+	subq	$(UKARCH_AUXSPCB_SIZE), %rsp
+	movq	UKARCH_AUXSPCB_OFFSETOF_CURR_FRAME(%rsp), %rsp
+
+	subq	$(UKARCH_EXECENV_END_ALIGN - 8), %rsp
+	.cfi_adjust_cfa_offset (UKARCH_EXECENV_END_ALIGN - 8)
+
+	pushq	%gs:LCPU_AUXSP_OFFSET
```
michpappas, line 140: A comment here on why we push the auxsp (we don't document the parameters that this function passes to `ukplat_syscall()` anywhere here).

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -53,24 +59,44 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
...
 
 	UK_ASSERT(usc);
 
+	execenv = &usc->execenv;
+	UK_ASSERT(execenv);
+
 	/* Save extended register state */
-	ukarch_ectx_sanitize((struct ukarch_ectx *)&usc->ectx);
-	ukarch_ectx_store((struct ukarch_ectx *)&usc->ectx);
+	ukarch_ectx_sanitize((struct ukarch_ectx *)&execenv->ectx);
```
michpappas: So now we make room for the control block and the application's context in `syscall.S`. Then we populate parts of them there, and parts of them in this function. It would be nice to enter this function in a clean state, but I also understand that moving it all to assembly leads to increased complexity. Perhaps we should consider a follow up PR that cleans up the handler in `arm64` and moves all context logic into `syscall.S`.

For now what I would propose is adding a comment that describes the contract between this function and its caller, given that unless someone is already familiar with `syscall.S` works, it is not clear what happens until we invoke the syscall.

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -38,11 +38,17 @@
...
 #include <uk/essentials.h>
+#include <uk/thread.h>
 #include "arch/regmap_linuxabi.h"
 #if CONFIG_LIBSYSCALL_SHIM_STRACE
 #include <uk/plat/console.h> /* ukplat_coutk */
 #endif /* CONFIG_LIBSYSCALL_SHIM_STRACE */
 
+struct uk_syscall_ctx {
+	struct ukarch_execenv execenv;
+	__uptr auxsp;
```
michpappas, line 64: Why do we need to push the `auxsp` as an argument? Shouldn't the API provide a call to retrieve the auxsp of the current thread?

plat/kvm/arm/exceptions.S
```diff
@@ -113,20 +113,27 @@
 	 */
 	mrs	x0, tpidr_el1
 	ldr	x0, [x0, #LCPU_AUXSP_OFFSET]
+	sub	x0, x0, #UKARCH_AUXSPCB_SIZE
+	ldr	x0, [x0, #UKARCH_AUXSPCB_OFFSETOF_CURR_FRAME]
+	sub	x0, x0, #UKARCH_EXECENV_END_ALIGN
+
 	/* NOTE: We should normally align the stack before doing this
```
michpappas, line 164: Is this comment still valid given that we now do align?

plat/kvm/arm/exceptions.S
```diff
@@ -113,20 +113,27 @@
...
 	 * is aligned to the corresponding ECTX alignment.
 	 * However, this is guaranteed to already be the case for the
 	 * auxiliary stack because it is allocated with this exact alignment
 	 * in mind.
 	 */
-	sub	x0, x0, #UK_SYSCALL_CTX_SIZE
+	sub	x0, x0, #UKARCH_EXECENV_SIZE
 
 	/* We now have in SP the trap stack and in x0 the auxiliary stack */
 	EXCHANGE_SP_WITH_X0  /* Switch them */
```
michpappas, line 174: Please add a blank line here if possible, to avoid having the stack switch operation being burried when reading the code.

arch/x86/x86_64/execenv.S
```diff
@@ -4,24 +4,25 @@
...
  */
 
+#include <uk/arch/ctx.h>
 #include <uk/arch/lcpu.h>
 #include <uk/asm.h>
-#include <uk/syscall.h>
 
-/* Used to pop the contents of a `struct uk_syscall_ctx` off the stack */
-ENTRY(uk_syscall_ctx_popall)
+/* Used to pop the contents of a `struct ukarch_execenv` off the stack */
```
michpappas: And what does it do at present? 😛 "Pop the contents of...". 

Moreover, I would also add the signature here to be clear about the parameters as well as that it doesn't return. So I would probably rephrase to something along the lines of:
```
/* Perform a context switch by restoring the contents of the passed struct.
 * void ukarch_execenv_load(struct ukarch_execenv *) __noreturn
 */
ENTRY(ukarch_execenv_load) 
```

arch/arm/arm64/execenv.S
```diff
@@ -4,21 +4,22 @@
  * You may not use this file except in compliance with the License.
  */
 
+#include <uk/arch/ctx.h>
 #include <uk/arch/lcpu.h>
 #include <uk/asm.h>
-#include <uk/syscall.h>
 
-ENTRY(uk_syscall_ctx_popall)
+ENTRY(ukarch_execenv_load)
```
michpappas, line 20: Please add a block that describes the purpose of this function and its prototype. See also the relevant comment in the implementation for `x86_64`.

arch/arm/arm64/execenv.S
```diff
@@ -4,21 +4,22 @@
...
 	/* Mask IRQ to make sure restore would not be interrupted by IRQ */
 	msr	daifset, #2
 
-	mov	x0, sp
-	add	x0, x0, #(__REGS_SIZEOF + UKARCH_SYSREGS_SIZE)
-	bl	ukarch_ectx_store
+	mov	sp, x0
 
+	add	x0, x0, #(__REGS_SIZEOF + UKARCH_SYSCTX_SIZE)
```
michpappas, line 37: Here also:
```
/* Restore ectx */
add    x0, x0, $(__REGS_SIZEOF  + UKARCH_SYSCTX_STATE)
```

arch/arm/arm64/execenv.S
```diff
@@ -4,21 +4,22 @@
...
+	bl	ukarch_ectx_load
 	mov	x0, sp
+
 	add	x0, x0, #(__REGS_SIZEOF)
```
michpappas, line 49: And here:
```
/* Restore sysctx */
add    x0, x0, $(__REGS_SIZEOF)
```

arch/x86/sysctx.c
```diff
@@ -0,0 +1,55 @@
...
+	UK_ASSERT(sysctx);
+
+	return sysctx->fs_base;
+}
+
+void ukarch_sysctx_set_tlsp(struct ukarch_sysctx *sysctx, __uptr tlsp)
+{
+	UK_ASSERT(sysctx);
+
+	uk_pr_debug("System call updated userland TLS pointer register to %p "
```
michpappas: And this checkpatch error...

arch/arm/sysctx.c
```diff
@@ -0,0 +1,39 @@
...
+	UK_ASSERT(sysctx);
+
+	return sysctx->tpidr_el0;
+}
+
+void ukarch_sysctx_set_tlsp(struct ukarch_sysctx *sysctx, __uptr tlsp)
+{
+	UK_ASSERT(sysctx);
+
+	uk_pr_debug("System call updated userland TLS pointer register to %p "
```
michpappas: This checkpatch error.

arch/x86/x86_64/execenv.S
```diff
@@ -4,24 +4,25 @@
...
+ENTRY(ukarch_execenv_load)
 	/* Do all this with IRQ's disabled, as the final iretq should pop off
 	 * a proper rflags anyway.
 	 */
 	cli
 
-	movq	%rsp, %rdi
-	addq	$(__REGS_SIZEOF + UKARCH_SYSREGS_SIZE), %rdi
-	call	ukarch_ectx_load
+	movq	%rdi, %rsp
```
michpappas: Please document assembly as much as possible, e.g
```
movq    %rdi, %rsp /* ... */
```

michpappas (changes requested):
@mogasergiu nice, the revised implementation is much cleaner. Also great to see context management gradually moving into an API 💪🏼 All comments are minor, plus some items to consider in follow up PRs.

include/uk/arch/ctx.h
mogasergiu: Actually I would rather not use `_CB` as it would feel inconsistent with `AUXSP` which is joined. But regarding comments, will do. I will also rename `uksc`->`uk_sysctx`.

include/uk/arch/ctx.h
mogasergiu: Hahah. Lazy naming: "Casted auxspcb" :rofl: Had no idea what to name it then and not sure what to name it now.. maybe `auxspcb_ptr`?

include/uk/arch/ctx.h
mogasergiu: I am hoping that the caller would use `UKARCH_AUXSPCB_SIZE` :grimacing:

include/uk/arch/ctx.h
mogasergiu: What about `curr_fp`? I want to emphasize that it is not static. The point of this pointer is to make nesting easier.

lib/ukboot/Makefile.uk
mogasergiu: There are many like this unfortunately (and more to come):
```
$ git grep -n "FIXME: Remove this and make the library self contained"                                   
lib/ukboot/Makefile.uk:8:# FIXME: Remove this and make the library self contained
lib/ukreloc/Makefile.uk:11:# FIXME: Remove this and make the library self contained
lib/uksched/Makefile.uk:7:# FIXME: Remove this and make the library self contained
lib/ukschedcoop/Makefile.uk:3:# FIXME: Remove this and make the library self contained
```

All added by me :man_dancing:

plat/common/include/uk/plat/common/lcpu.h
mogasergiu: I would really love to see this as I don't like this inconsistency either.

plat/common/x86/syscall.S
mogasergiu: **Puts on GNU glasses**
The code is the documentation.

I will update :+1:

lib/syscall_shim/uk_syscall_binary.c
mogasergiu: On syscall C handler entry we are no longer given guarantee that registers have been switched to those of Unikraft's, so therefore we have no direct access to `auxsp`. However we did have it for a brief time on `swapgs` which is when we also push this. I will thoroughly comment this as it's basically an optimization to avoid having to do any `swapgs`/`rdgsbase`/etc instructions more than needed as they cost lots of CPU cycles.

plat/kvm/arm/exceptions.S
mogasergiu: Yes, because we do not actually align we just assume it is aligned. And we also should not explicitly align in assembly in the future either as it will be hard to convey this information through CFI directives to a debugger. As Marco also suggested once, it would be an amazing addition if we had some macro's use raw DWARF bytecode instead... but that requires time investment.

arch/x86/x86_64/execenv.S
mogasergiu: Maaaaan, I clearly ain't becomin a book author any time soon :smile:

plat/common/x86/syscall.S
```diff
@@ -65,48 +67,96 @@ ENTRY(_ukplat_syscall)
...
-	 *	|             |
-	 *	--------------- <-- (auxsp -
-	 *			     (PAGE_SIZE *
-	 *			      (1 << CONFIG_UKPLAT_AUXSP_PAGE_ORDER)))
-	 *	  END OF AUXSP
-	 */
-
 	/*
 	 * Push arguments in the order of 'struct __regs' to the stack.
 	 * We are going to handover a refernce to this stack area as
```
mogasergiu: TODO: set IRQ enable flag in saved eflags so that stored ctx is reflects the reality more honestly - an application will never disable IRQ's.

arch/x86/x86_64/execenv.S
```diff
@@ -0,0 +1,78 @@
...
+ * Loads a given execution environment on the currently executing CPU.
+ *
+ * NOTE: This function cannot be returned from, it overwrites the entire current
+ *       context.
+ *
+ * @RDI execenv
+ *   Reference to execution environment to load
+ */
+ENTRY(ukarch_execenv_load)
+	/**
```
mogasergiu, line 24: TODO: Add CFI directives

lib/posix-process/arch/x86_64/clone.c
```diff
@@ -7,54 +7,57 @@
...
+		cexecenv->sysctx.fs_base = child->tlsp;
 
-	ukarch_ctx_init(&child->ctx,
-			auxsp_pos,
-			0,
-			(__uptr)&uk_syscall_ctx_popall);
+	ukarch_ctx_init_entry1(&child->ctx,
+			       auxsp_pos,
+			       1,
+			       (ukarch_ctx_entry1)&ukarch_execenv_load,
```
michpappas, line 60: This causes a warning:
```
unikraft/lib/posix-process/arch/x86_64/clone.c: In function ‘clone_setup_child_ctx’:
unikraft/lib/posix-process/arch/x86_64/clone.c:61:32: warning: cast between incompatible function types from ‘__attribute__((noreturn)) void (*)(struct ukarch_execenv *)’ to ‘__attribute__((noreturn)) void (*)(long int)’ [-Wcast-function-type]
   61 |                                (ukarch_ctx_entry1)&ukarch_execenv_load,
      |                                ^
```

mogasergiu:
Apologies, I pushed again with the whole rebae :man_facepalming:. Here is the actual diff
```diff
diff --git a/lib/syscall_shim/include/uk/syscall.h b/lib/syscall_shim/include/uk/syscall.h
index b7d96d7d..77f2de99 100644
--- a/lib/syscall_shim/include/uk/syscall.h
+++ b/lib/syscall_shim/include/uk/syscall.h
@@ -514,7 +514,7 @@ long uk_syscall6(long nr, long arg1, long arg2, long arg3,
 /* Raw system call, returns negative codes on errors */
 long uk_syscall_r(long nr, ...);
 long uk_vsyscall_r(long nr, va_list arg);
-long uk_syscall6_r_u(struct ukarch_execenv *execenv);
+long uk_syscall6_r_e(struct ukarch_execenv *execenv);
 long uk_syscall6_r(long nr, long arg1, long arg2, long arg3,
 		   long arg4, long arg5, long arg6);
 
diff --git a/lib/syscall_shim/uk_syscall6_r.awk b/lib/syscall_shim/uk_syscall6_r.awk
index 5b6fc5cd..8810d0f5 100644
--- a/lib/syscall_shim/uk_syscall6_r.awk
+++ b/lib/syscall_shim/uk_syscall6_r.awk
@@ -6,9 +6,9 @@ BEGIN {
 	print "#include <uk/syscall.h>"
 	print "#include \"arch/regmap_linuxabi.h\"\n"
 
-	print "UK_SYSCALL_EXECENV_PROLOGUE_DEFINE(uk_syscall6_r, uk_syscall6_r_u,"
+	print "UK_SYSCALL_EXECENV_PROLOGUE_DEFINE(uk_syscall6_r, uk_syscall6_r_e,"
 	print "\t\t\t\t14, long, nr, long, arg1, long, arg2, long, arg3, long, arg4, long, arg5, long, arg6)";
-	print "\nlong __used uk_syscall6_r_u(struct ukarch_execenv *execenv)"
+	print "\nlong __used uk_syscall6_r_e(struct ukarch_execenv *execenv)"
 	print "{"
 	print "\tlong ret;"
 	print "\n\tswitch (execenv->regs.rsyscall) {"
@@ -18,12 +18,12 @@ BEGIN {
 	name = $1
 	sys_name = "SYS_" name
 	uk_syscall_r = "uk_syscall_r_" name
-	uk_syscall_r_u = "uk_syscall_r_e_" name
+	uk_syscall_r_e = "uk_syscall_r_e_" name
 	args_nr = $2 + 0
 	printf "\n#ifdef HAVE_uk_syscall_%s\n", name;
 	printf "\tcase %s:\n", sys_name;
 	printf "\n#ifdef HAVE_uk_syscall_e_%s\n", name;
-	printf "\t\tret = %s((long)execenv);\n", uk_syscall_r_u;
+	printf "\t\tret = %s((long)execenv);\n", uk_syscall_r_e;
 	printf "#else /* !HAVE_uk_syscall_e_%s */\n", name;
 	printf "\t\tret = %s(\n\t\t\t\t\t", uk_syscall_r;
 	for (i = 0; i < args_nr - 1; i++)
diff --git a/lib/syscall_shim/uk_syscall_binary.c b/lib/syscall_shim/uk_syscall_binary.c
index d0f4f180..a8f78202 100644
--- a/lib/syscall_shim/uk_syscall_binary.c
+++ b/lib/syscall_shim/uk_syscall_binary.c
@@ -113,7 +113,7 @@ void ukplat_syscall_handler(struct uk_syscall_ctx *usc)
 		    execenv->regs.rarg1);
 #endif /* CONFIG_LIBSYSCALL_SHIM_DEBUG_HANDLER */
 
-	execenv->regs.rret0 = uk_syscall6_r_u(execenv);
+	execenv->regs.rret0 = uk_syscall6_r_e(execenv);
 
 #if CONFIG_LIBSYSCALL_SHIM_STRACE
 	prsyscalllen = uk_snprsyscall(prsyscallbuf, ARRAY_SIZE(prsyscallbuf),
diff --git a/lib/uksched/sched.c b/lib/uksched/sched.c
index 920bb847..249f6eea 100644
--- a/lib/uksched/sched.c
+++ b/lib/uksched/sched.c
@@ -35,7 +35,6 @@
 
 #include <stdlib.h>
 #include <string.h>
-#include <sys/membarrier.h>
 #include <uk/plat/common/lcpu.h>
 #include <uk/plat/config.h>
 #include <uk/plat/time.h>
 ```

File: arch/x86/sysctx.c [hunk 35]
```diff
@@ -0,0 +1,54 @@
...
+	UK_ASSERT(sysctx);
+
+	return sysctx->fs_base;
+}
+
+void ukarch_sysctx_set_tlsp(struct ukarch_sysctx *sysctx, __uptr tlsp)
+{
+	UK_ASSERT(sysctx);
+
+	uk_pr_debug("System call updated userland TLS pointer register to %p (before: %p)\n",
```
skuenzer, line 20: I find the debug message misleading by just seeing the library functions here. I know it is mostly called from system calls but the API allows to call it also from other places. I'd either remove the message or generalize it more (not mention system call). The other question that pops up if we should put then a message to gs_base, as well.

skuenzer: I find the debug message misleading by just seeing the library functions here. I know it is mostly called from system calls but the API allows to call it also from other places. I'd either remove the message or generalize it more (not mention system call). The other question that pops up if we should put then a message to gs_base, as well.

arch/x86/sysctx.c
```diff
@@ -0,0 +1,54 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+#include <uk/assert.h>
+#include <uk/arch/ctx.h>
+
+__uptr ukarch_sysctx_get_tlsp(struct ukarch_sysctx *sysctx)
```
skuenzer, line 9: Out of curiosity: Is there a reason why we compile a C file with these functions instead of providing `static inline` helpers in the according header (uk/asm/)? The functions are just accessing some struct fields and we could benefit from compiler optimizations even without LTO.

arch/x86/x86_64/execenv.S
skuenzer: Agree and saw that you did it: CFI here is a very good idea for improving debugging experience.

arch/x86/x86_64/include/uk/asm/sysctx.h
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Get the TLS pointer from system register context given as argument
+ *
+ * @param sysctx
+ *   The system register context whose TLS pointer to get.
+ * @return
+ *   The TLS pointer stored in the system register context.
+ */
+__uptr ukarch_sysctx_get_tlsp(struct ukarch_sysctx *sysctx);
```
skuenzer, line 40: We could have static inline functions here instead but I am mostly only wondering.

include/uk/arch/ctx.h
```diff
@@ -56,12 +59,110 @@
...
+	(IN_RANGE((sp), (auxsp) - AUXSTACK_SIZE, AUXSTACK_SIZE))
+
+struct ukarch_auxspcb {
+	/* Current safe frame pointer inside the auxiliary stack area */
+	__uptr curr_fp;
+	/* Unikraft system registers (e.g. TLS pointer) */
+	struct ukarch_sysctx uksysctx;
+	/* Padding for end alignment, the auxiliary stack area begins after */
+	__u8 pad[UKARCH_AUXSPCB_PAD];
+};
```
skuenzer, line 201: We do not want to have this struct `packed`, right?

include/uk/arch/ctx.h
```diff
@@ -194,6 +295,168 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ */
+static inline void ukarch_auxsp_set_cb(__uptr auxsp,
+				       struct ukarch_auxspcb *auxspcb)
+{
+	struct ukarch_auxspcb *auxspcb_ptr;
+
+	UK_ASSERT(auxsp);
+	UK_ASSERT(IS_ALIGNED(auxsp, UKARCH_AUXSP_ALIGN));
+
+	auxspcb_ptr = (struct ukarch_auxspcb *)(auxsp - sizeof(*auxspcb_ptr));
```
skuenzer: I wonder if `ukarch_auxsp_get_cb` is already covering these lines should be used here instead. I am also thinking if we should imply a call order, meaning that only `ukarch_auxsp_init` and `ukarch_auxsp_get_cb` should only accept a `__uptr` and then both return the pointer to the struct. All other functions would operaate based on a `struct auxspcb_ptr *` argument (instead of `__uptr` to the stack). As a side effect, I assume this would also reduce complexity of these functions and we have code deduplicated.

include/uk/arch/ctx.h
skuenzer: I think the length parameter would only help for cases to keep binary backwards compatibility. Any native code will be recompiled so that the latest size is in use in case `UKARCH_AUXSPCB_SIZE` isn't used.

skuenzer: However, I have a different question: Should we zero out that struct during init?

include/uk/arch/ctx.h
```diff
@@ -56,12 +59,110 @@
...
+	/* Extended context (e.g. SIMD etc.) */
+	__u8 ectx[UKARCH_ECTX_SIZE];
+	/* Padding for end alignment */
+	__u8 pad[UKARCH_EXECENV_PAD_SIZE];
+};
+
+#define SP_IN_AUXSP(sp, auxsp)					\
+	(IN_RANGE((sp), (auxsp) - AUXSTACK_SIZE, AUXSTACK_SIZE))
+
+struct ukarch_auxspcb {
```
skuenzer, line 194: I think a small ascii image for the intended stack layout to show where this struct relies would help.

include/uk/arch/ctx.h
```diff
@@ -194,6 +295,168 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * and the Unikraft system context.
+ *
+ * @param auxsp
+ *   The auxiliary stack pointer whose control block to set.
+ *  Must point to the high end of the auxiliary stack.
+ * @param auxspcb
+ *   The pointer to the control block to set for the auxstack pointer.
+ *
+ */
+static inline void ukarch_auxsp_set_cb(__uptr auxsp,
```
skuenzer: I find the naming confusing in comparison to `get_cb`. By the name I was expecting that this allows you to set a different auxsp control block for a given stack, however, what it does is copying values. So, could we call it `cpy_cb` instead, and what would you think of the following signature:
```
static inline void ukarch_auxsp_set_cb(struct ukarch_auxspcb *dest, struct ukarch_auxspcb *src);
```

skuenzer: Regarding this, please read also my comment at line 341.

include/uk/arch/ctx.h
skuenzer: Also the following functions (set_uktlsp, etc.) would have to be adopted.

include/uk/arch/ctx.h
```diff
@@ -194,6 +295,168 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ */
+static inline void ukarch_auxsp_init(__uptr auxsp)
+{
+	struct ukarch_auxspcb *auxspcb_ptr;
+
+	UK_ASSERT(auxsp);
+	UK_ASSERT(IS_ALIGNED(auxsp, UKARCH_AUXSP_ALIGN));
+
+	auxspcb_ptr = (struct ukarch_auxspcb *)(auxsp - sizeof(*auxspcb_ptr));
+	auxspcb_ptr->curr_fp = auxsp - sizeof(*auxspcb_ptr);
```
skuenzer, line 361: 🤔 In principle, this function is also covered by the functionality of  `ukarch_auxsp_set_curr_fp`, right?

include/uk/arch/ctx.h
```diff
@@ -194,6 +295,168 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ *
+ */
+static inline void ukarch_auxsp_set_uktlsp(__uptr auxsp, __uptr uktlsp)
+{
+	struct ukarch_auxspcb *auxspcb_ptr;
+
+	UK_ASSERT(auxsp);
+	UK_ASSERT(IS_ALIGNED(auxsp, UKARCH_AUXSP_ALIGN));
+
+	auxspcb_ptr = ukarch_auxsp_get_cb(auxsp);
```
skuenzer: I think we can assume that `ukarch_auxsp_get_cb` is already covering the assertions.

plat/common/include/uk/plat/common/lcpu.h
```diff
@@ -62,10 +63,10 @@ struct lcpu_arch { };
 #endif /* !__ASSEMBLY__ */
 #endif /* !LCPU_ARCH_SIZE */
 
-#define IS_LCPU_PTR(ptr)                                               \
-	(IN_RANGE((__uptr)(ptr),                                        \
-		  (__uptr)lcpu_get(0),                                  \
-		  (__uptr)CONFIG_UKPLAT_LCPU_MAXCOUNT *                 \
+#define IS_LCPU_PTR(ptr)						\
```
skuenzer, line 64: Try to avoid doing whitespace/formatting changes that aren't strictly necessary.

plat/common/arm/traps_arm64.c
```diff
@@ -225,13 +226,28 @@ void trap_el1_irq(struct __regs *regs)
...
+	struct ukarch_execenv *execenv = (struct ukarch_execenv *)ctx->regs;
+
+	/* Save extended register state */
+	ukarch_ectx_sanitize((struct ukarch_ectx *)&execenv->ectx);
+	ukarch_ectx_store((struct ukarch_ectx *)&execenv->ectx);
+
+	/* Save system context state */
+	ukarch_sysctx_store(&execenv->sysctx);
 
 	ukplat_syscall_handler((struct uk_syscall_ctx *)ctx->regs);
```
skuenzer: Would this be the same?:
```c
ukplat_syscall_handler((struct uk_syscall_ctx *)execenv);
```
It might be less confusing.

lib/uksched/thread.c
```diff
@@ -262,6 +265,12 @@ static void _uk_thread_struct_init(struct uk_thread *t,
 		t->flags |= UK_THREADF_ECTX;
 	}
 
+	if (t->flags & UK_THREADF_UKTLS && t->flags & UK_THREADF_AUXSP) {
```
skuenzer, line 268: I am wondering if the compiler is intelligent enough to get to this:
```c
(t->flags & (UK_THREADF_UKTLS | UK_THREADF_AUXSP) == (UK_THREADF_UKTLS | UK_THREADF_AUXSP))
```
This isn't a comment to change anything.

lib/uksched/sched.c
```diff
@@ -215,7 +216,7 @@ int uk_sched_start(struct uk_sched *s)
 	 *       an TLS that is derived from the Unikraft TLS template.
 	 */
 	tlsp = ukplat_tlsp_get();
-	auxsp = ukplat_lcpu_get_auxsp();
+	auxsp = lcpu_get_auxsp();
```
skuenzer: This makes me wonder why this rename in this PR? Weren't we thinking to adopt it anyways again with a proper rearch, shouldn't we stick with the scheme we have today?

lib/uksched/Makefile.uk
```diff
@@ -4,6 +4,11 @@ $(eval $(call addlib_s,libuksched,$(CONFIG_LIBUKSCHED)))
 #       `LIBUKSCHED_THREAD_FLAGS-y`, `CFLAGS-y`, and `CXXFLAGS-y`
 #       by `Makefiles.rules`
 
+# FIXME: Remove this and make the library self contained
```
skuenzer: Hum... this should not happen and is even voiolating our current architectural structure because these headers are platform-internal and implementation and headers can vary accross platforms. So, this won't work with how we link the final images using the platform libraries. I would expect the APIs to be declared within `/include/uk/arch/lcpu.h` and `/include/uk/plat/lcpu.h` instead. What do you think?

arch/x86/sysctx.c
mogasergiu: No reason, I just tend to rely on the fact that LTO would inline these if needed.

arch/x86/sysctx.c
mogasergiu: Ok sure, I just kept this from back when this was in lib/syscall_shim. Makes sense to remove now. I will add for gs_base too sure.

include/uk/arch/ctx.h
mogasergiu: Should not be needed, same for sysctx/execenv. This should be always a multiple of 16 in size and, ideally, all fields should be aligned. I am expecting the `UK_CTASSERT`'s to catch misbehavior at build time.

include/uk/arch/ctx.h
mogasergiu: True, but `ukarch_auxsp_set_curr_fp` already calls `ukarch_auxsp_get_cb` which would not make sense to call during its very initialization IMO.

include/uk/arch/ctx.h
mogasergiu: Sure, makes sense.

include/uk/arch/ctx.h
mogasergiu: The problem is uk_thread, struct lcpu and co. use the auxsp pointer `__uptr`, not the auxspcb pointer. The way I viewed this is that through `ukarch_auxsp_*` whoever may hold an auxsp pointer does not need to know of the auxspcb structure, unless they really want to.

include/uk/arch/ctx.h
mogasergiu: True, but do we really want to make assumptions that callee's assert stuff in general? I tend to use `UK_ASSERT` for every argument of a function I use.

lib/uksched/thread.c
mogasergiu: I hope it is and I think that the way it is now is more readable. It also occupies less space :).

plat/common/arm/traps_arm64.c
mogasergiu: Agreed

plat/common/include/uk/plat/common/lcpu.h
mogasergiu: I actually fixed it because some of those lines had whitespaces instead of tabs :(.

arch/x86/sysctx.c
mogasergiu: Actually, another reason would be the fact that I want them to keep the `|isr` attribute. At the moment we have non-ISR code that runs in exception handlers. This is fine if this is included in already ISR files, but If a non ISR function in a non ISR file were to call this then for sure it would not keep the ISR attribute.

include/uk/arch/ctx.h
mogasergiu: hmm that being said, maybe I should add some `__offsetof` constant assertions as well. Similar to the way `struct lcpu` has.

mogasergiu:
> branch from [11cde88 ](https://github.com/unikraft/unikraft/commit/11cde888ef36e1d1cec8ea96bec684f38bde72e5)to [ebf0f40 ](https://github.com/unikraft/unikraft/commit/ebf0f40ce312850594f4b2ca95abc87766a293b7)
[now](https://github.com/unikraft/unikraft/pull/1346#event-12905455701)

This push is a simple rebase. I will now soon push with comment feedback.

arch/x86/sysctx.c
mogasergiu: Following an offline discussion, we decided it is fine to keep as is, especially considering the above and the fact that they may be called from assembly code.

arch/x86/x86_64/include/uk/asm/sysctx.h
mogasergiu: [Related](https://github.com/unikraft/unikraft/pull/1346#discussion_r1611290046)

include/uk/arch/ctx.h
mogasergiu: Following an offline discussion, it's been decided to move forward like so:
```C
static inline void ukarch_auxsp_init(__uptr auxsp);
static inline struct ukarch_auxspcb *ukarch_auxsp_get_cb(__uptr auxsp);
static inline void ukarch_auxspcb_set_uktlsp(struct ukarch_auxspcb *auxspcb, __uptr uktlsp);
static inline __uptr ukarch_auxspcb_get_uktlsp(struct ukarch_auxspcb *auxspcb);
static inline void ukarch_auxspcb_set_curr_fp(struct ukarch_auxspcb *auxspcb, __uptr curr_fp);
static inline __uptr ukarch_auxspcb_get_curr_fp(struct ukarch_auxspcb *auxspcb);
```
And no more getter or control block copy, as it would not be used at the moment anyway.

include/uk/arch/ctx.h
mogasergiu: [Resolved](https://github.com/unikraft/unikraft/pull/1346#discussion_r1611912800)

include/uk/arch/ctx.h
mogasergiu: [Resolved](https://github.com/unikraft/unikraft/pull/1346#discussion_r1611912800)

include/uk/arch/ctx.h
mogasergiu: [Resolved](https://github.com/unikraft/unikraft/pull/1346#discussion_r1611912800)

include/uk/arch/ctx.h
```diff
@@ -56,12 +59,154 @@
 #define UKARCH_CTX_OFFSETOF_SP 8
 #endif
 
-#ifndef __ASSEMBLY__
+/* We must make sure that ECTX is aligned, so we make use of some padding,
+ * whose size is equal to what we need to add to UKARCH_ECTX_SIZE
+ * to make it aligned with UKARCH_ECTX_ALIGN
+ */
+#define UKARCH_EXECENV_PAD_SIZE					\
+	(ALIGN_UP(UKARCH_ECTX_SIZE,				\
```
skuenzer, line 67: `<uk/essentials.h>` is only included for non-__ASSEMBLY__ files. I assume for assembly, ALIGN_UP and ALIGN_DOWN macros do not work. Correct?

include/uk/arch/ctx.h
```diff
@@ -56,12 +59,154 @@
...
+UK_CTASSERT(__offsetof(struct ukarch_execenv, regs) ==
+	    UKARCH_EXECENV_OFFSETOF_REGS);
+UK_CTASSERT(__offsetof(struct ukarch_execenv, sysctx) ==
+	    UKARCH_EXECENV_OFFSETOF_SYSCTX);
+UK_CTASSERT(__offsetof(struct ukarch_execenv, ectx) ==
+	    UKARCH_EXECENV_OFFSETOF_ECTX);
+
+/**
+ * Layout of the auxiliary stack and its embedded control block located at its
+ * end (towards higher address space).
```
skuenzer, line 161: Maybe some arrows that indicate the stack growing direction makes it slightly easier in the image.

include/uk/arch/ctx.h
```diff
@@ -249,6 +515,17 @@ void ukarch_ectx_store(struct ukarch_ectx *state);
  */
 void ukarch_ectx_load(struct ukarch_ectx *state);
 
+/**
+ * Loads a given execution environment on the currently executing CPU.
+ *
+ * NOTE: This function cannot be returned from, it overwrites the entire current
```
skuenzer: NOTE: This function does not return, it overwrites ...

lib/posix-process/arch/arm64/clone.c
```diff
@@ -7,51 +7,53 @@
...
+	/* Make room for child's `struct ukarch_execenv` and copy them */
+	auxsp_pos = ALIGN_DOWN(auxsp_pos, UKARCH_EXECENV_END_ALIGN);
+	auxsp_pos -= UKARCH_EXECENV_SIZE;
 
 	/* Now patch the child's return registers */
-	cusc = (struct uk_syscall_ctx *)auxsp_pos;
+	cexecenv = (struct ukarch_execenv *)auxsp_pos;
+	*cexecenv = *pexecenv;
 
 	/* Child must see x0 as 0 */
```
skuenzer: Maybe mention that x0 is the return value for the child

plat/common/x86/syscall.S
```diff
@@ -65,54 +67,110 @@ ENTRY(_ukplat_syscall)
...
+
 	/* Describing the rsp relative to GS would make it necessary to emit
 	 * raw CFI. Instead of doing so, mark rsp as undefined temporarily
 	 */
 	.cfi_undefined rsp
 
+	subq	$(UKARCH_AUXSPCB_SIZE), %rsp
+	movq	UKARCH_AUXSPCB_OFFSETOF_CURR_FP(%rsp), %rsp
+
+	/* We subtract 8 less bytes here so that we have room for the pushed
```
skuenzer: -less

plat/common/x86/syscall.S
```diff
@@ -65,54 +67,110 @@ ENTRY(_ukplat_syscall)
...
+	 * auxsp.
+	 * Current stack pointer now points to the current frame pointer
```
skuenzer: Afterwards, the current stack pointer is pointing to the current frame...

plat/common/x86/syscall.S
```diff
@@ -65,54 +67,110 @@ ENTRY(_ukplat_syscall)
...
+	 * within the auxiliary stack. Subtract UKARCH_EXECENV_END_ALIGN
+	 * to make room for the 8-byte auxsp pointer, since the layout is
+	 * struct uk_syscall_ctx {
+	 *	struct ukarch_execenv execenv;
+	 *	__uptr auxsp;  <-- make room for these
+	 *     ..... leftover room because of alignment subtraction ....
```
skuenzer:
```diff
-..... leftover room because of alignment subtraction ....
+... padding to ensure alignment ...
```

plat/common/x86/syscall.S
```diff
@@ -65,54 +67,110 @@ ENTRY(_ukplat_syscall)
...
+	 *                | |  struct     | |             |
+	 *                | | ukarch_ectx | |ECTX_SIZE    |
+	 *                | |             | |             |
+	 *                | |             | |             |
+	 *                | |             | |             |
+	 *                v +-------------+ v             v
+	 *                       stack
+	 *                       pointer
+	 *
+	 * Where ukarch_sysctx/ukarch_ectx/__regs we fill in below.
```
skuenzer: Where ukarch_sysctx/ukarch_ectx/__regs is filled in the following:

plat/kvm/arm/exceptions.S
```diff
@@ -113,20 +113,72 @@
...
+	ldr	x0, [x0, #UKARCH_AUXSPCB_OFFSETOF_CURR_FP]
+
+	/**
+	 * Current stack pointer now points to the current frame pointer
+	 * within the auxiliary stack. Subtract UKARCH_EXECENV_END_ALIGN
+	 * to make room for the 8-byte auxsp pointer, since the layout is
+	 * struct uk_syscall_ctx {
+	 *	struct ukarch_execenv execenv;
+	 *	__uptr auxsp;  <-- make room for these
+	 *     ..... leftover room because of alignment subtraction ....
```
skuenzer: See similar comments as given for `syscall.S`

plat/kvm/arm/lcpu.c
```diff
@@ -79,3 +79,13 @@ void ukplat_lcpu_irqs_handle_pending(void)
 {
 	// TODO
 }
+
+void ukplat_lcpu_set_auxsp(__uptr auxsp)
```
skuenzer, line 83: This looks like an `ukarch_lcpu_*()` interface...

lib/syscall_shim/include/uk/syscall.h
```diff
@@ -69,22 +46,12 @@
 #include <stdarg.h>
 #include <uk/print.h>
 #include "legacy_syscall.h"
+#include "../../arch/regmap_linuxabi.h"
```
skuenzer: if we need the linuxabi header here, better to move (officially export) the header to something like
`include/uk/syscall_shim/regmap_linuxabu.h`

skuenzer: or actually, place it under the same path as the other syscall related sub-headers:
`<uk/bits/syscall_linuxabi.h>`
In fact `"legacy_syscall.h"` should have also been included as `<uk/legacy_syscall.h>` or `<uk/syscall_legacy.h>`.

include/uk/arch/ctx.h
mogasergiu:
```C
/* Note: a has to be a power of 2 */
#ifndef ALIGN_UP
#define ALIGN_UP(v, a) (((v) + (a)-1) & ~((a)-1))
#endif

/* Note: a has to be a power of 2 */
#ifndef ALIGN_DOWN
#define ALIGN_DOWN(v, a) ((v) & ~((a)-1))
#endif
```

No reason to not work. The assembler resolves all of that macro mess to a constant. I can confirm it works as these are used for example in `plat/common/x86/syscall.S`

plat/common/x86/syscall.S
mogasergiu: That would be correct in fact.
```asm
	subq	$(UKARCH_EXECENV_END_ALIGN - 8), %rsp
	                                    ^^
	.cfi_adjust_cfa_offset (UKARCH_EXECENV_END_ALIGN - 8)

	pushq	%gs:LCPU_AUXSP_OFFSET <------------------------ the reason
	.cfi_adjust_cfa_offset 8

```

plat/common/x86/syscall.S
mogasergiu: Hmm, I intentionally said "leftover" because it is an unfortunate consequence of having to create a 64 byte slot only to store an 8 byte value just because of alignment constraints.  Padding sounds like it's something that we intentionally add? I don't know :thinking:

plat/kvm/arm/exceptions.S
mogasergiu: > Hmm, I intentionally said "leftover" because it is an unfortunate consequence of having to create a 64 byte slot only to store an 8 byte value just because of alignment constraints. Padding sounds like it's something that we intentionally add? I don't know 🤔

Ditto

plat/kvm/arm/lcpu.c
mogasergiu: We do not have such API's. Not sure I understand? This is the function that you previously asked me to rename:
`lcpu_arh_get/set_auxsp` -> `ukplat_lcpu_get/set_auxsp`.

include/uk/arch/ctx.h
skuenzer: Then this is resolved somewhere implicitly by the preprocessor. What I meant is that the essentials header is not included for assembly build because it is guarded at the top of syscall.h .

plat/common/x86/syscall.S
skuenzer: Hum... is it then "we substract 8 bytes less"?

plat/common/x86/syscall.S
skuenzer: Okay, understand. How about: "space left unused"... somehow trying to avoid "leftover" because it sounds like a gift for somebody - not sure if only I am thinking that way 😄

plat/kvm/arm/lcpu.c
skuenzer: Okay, then leave it here. We will touch this code anyways again with re-arch in mind.

skuenzer (approved):
Thanks a lot for your work. As discussed, we will later restructure libcontext and your additions to it with plat rearch and the lcpu library in particular.

Reviewed-by: Simon Kuenzer <simon.kuenzer@unikraft.io>

@michpappas I am handing over to you.

michpappas:
With all @razvand's tests passing,

PR #1347, opened 2024-03-04
arch: Add exception context escape trampolines ctx creation

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Implement the ability to jump from exception handling context to an
I/O-able context from which one can return to the context that the
exception handler was supposed to return to.

Essentially by creating a return-chain into a function that creates
room on a given stack to store execution environment
(struct ukarch_execenv) which would then return to a custom entry which
would then return to a routine that would restore the aforementioned
execution environment, one is able to freely do I/O, use any registers.

The caller of ukarch_ctx_init_ehtrampoX API's will be given a context
that is guaranteed to run with IRQ's enabled and from which, upon
return, execution from the point the exception handler was supposed
to return to is resumed.
```
        ┌───────────────────┐
▲       │                   │
│       │      struct       │
│       │      execenv      │
│       │                   │
│       │                   │
│       ├───────────────────┤
│       │                   │
│       │  execenv_load(ee) │
│       │                   │
│       ├───────────────────┤
│       │ entry(ee, arg0,   │
│       │       arg1…) for  │
│       │    ehtrampo0,1,2… │
│       │    variants       │
│       │                   │
│       │                   │
│       ├───────────────────┤
│       │                   │
│       │ ehtrampo_.._store │
exec    │                   │
order   │                   │
        └───────────────────┘
```

### DIFF
--- a/arch/x86/ctx.c
+++ b/arch/x86/ctx.c
@@ -47,6 +47,7 @@ void _ctx_x86_clearregs(void);
 void _ctx_x86_call0(void);
 void _ctx_x86_call1(void);
 void _ctx_x86_call2(void);
+void _ctx_x86_call3(void);
 
 void ukarch_ctx_init(struct ukarch_ctx *ctx,
 		     __uptr sp, int keep_regs,
@@ -159,4 +160,32 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
 
 	uk_pr_debug("ukarch_ctx %p: entry:%p(%lx, %lx), sp:%p\n",
 		    ctx, entry, arg0, arg1, (void *)sp);
+}
+
+void ukarch_ctx_init_entry3(struct ukarch_ctx *ctx,
+			    __uptr sp, int keep_regs,
+			    ukarch_ctx_entry3 entry,
+			    long arg0, long arg1, long arg2)
+{
+	__uptr _sp;
+
+	UK_ASSERT(ctx);
+	UK_ASSERT(sp);			/* a stack is needed */
+	UK_ASSERT(entry);		/* NULL as func will cause a crash */
+	UK_ASSERT(!(sp & UKARCH_SP_ALIGN_MASK)); /* sp properly aligned? */
+
+	sp  = ukarch_rstack_push(sp, (__u64)0x0); /* SystemV call convention */
+	_sp = ukarch_rstack_push(sp, (long)entry);
+	_sp = ukarch_rstack_push(_sp, arg0);
+	_sp = ukarch_rstack_push(_sp, arg1);
+	_sp = ukarch_rstack_push(_sp, arg2);
+	if (keep_regs) {
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_call3);
+	} else {
+		_sp = ukarch_rstack_push(_sp, (long)_ctx_x86_call3);
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_clearregs);
+	}
+
+	uk_pr_debug("ukarch_ctx %p: entry:%p(%lx, %lx, %lx), sp:%p\n",
+		    ctx, entry, arg0, arg1, arg2, (void *)sp);
 }
--- a/arch/x86/ctx.c
+++ b/arch/x86/ctx.c
@@ -59,16 +59,16 @@ void ukarch_ctx_init(struct ukarch_ctx *ctx,
 	UK_ASSERT(ip);			/* NULL as IP will cause a crash */
 	UK_ASSERT(!keep_regs && sp);	/* a stack is needed when clearing */
 
-	_sp = ukarch_rstack_push(sp, (long) ip);
+	_sp = ukarch_rstack_push(sp, (long)ip);
 	if (keep_regs) {
-		ukarch_ctx_init_bare(ctx, _sp, (long) _ctx_x86_call0);
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_call0);
 	} else {
-		_sp = ukarch_rstack_push(_sp, (long) _ctx_x86_call0);
-		ukarch_ctx_init_bare(ctx, _sp, (long) _ctx_x86_clearregs);
+		_sp = ukarch_rstack_push(_sp, (long)_ctx_x86_call0);
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_clearregs);
 	}
 
 	uk_pr_debug("ukarch_ctx %p: start:%p sp:%p\n",
-		    ctx, (void *) ip, (void *) sp);
+		    ctx, (void *)ip, (void *)sp);
 }
 
 void ukarch_ctx_init_entry0(struct ukarch_ctx *ctx,
@@ -97,17 +97,17 @@ void ukarch_ctx_init_entry0(struct ukarch_ctx *ctx,
 	 *       do not require any alignments and will pass any sp offsets
 	 *       through.
 	 */
-	sp  = ukarch_rstack_push(sp, (__u64) 0x0);
-	_sp = ukarch_rstack_push(sp, (long) entry);
+	sp  = ukarch_rstack_push(sp, (__u64)0x0);
+	_sp = ukarch_rstack_push(sp, (long)entry);
 	if (keep_regs) {
-		ukarch_ctx_init_bare(ctx, _sp, (long) _ctx_x86_call0);
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_call0);
 	} else {
-		_sp = ukarch_rstack_push(_sp, (long) _ctx_x86_call0);
-		ukarch_ctx_init_bare(ctx, _sp, (long) _ctx_x86_clearregs);
+		_sp = ukarch_rstack_push(_sp, (long)_ctx_x86_call0);
+		ukarch_ctx_init_bare(ctx, _sp, (long)_ctx_x86_clearregs);
 	}
 
 	uk_pr_debug("ukarch_ctx %p: entry:%p(), sp:%p\n",
-		    ctx, entry, (void *) sp);
+		    ctx, entry, (void *)sp);
 }
 
 void ukarch_ctx_init_entry1(struct ukarch_ctx *ctx,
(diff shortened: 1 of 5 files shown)


arch/x86/ctx.c
```diff
@@ -286,3 +286,299 @@ void ukarch_ctx_init_entry6(struct ukarch_ctx *ctx,
...
+	sp = ukarch_rstack_push(sp, (long)entry);
+	sp = ukarch_rstack_push(sp, (long)ee);
+	sp = ukarch_rstack_push(sp, (long)_ctx_x86_call1);
+
+	/* We were preceeded by even number of pushes so in order to enter
+	 * ehtrampo_ectx_and_sysctx_store() with rsp + 8 as 16-byte aligned,
+	 * we have to push something here that returns into the above return
+	 * chain.
+	 */
+	sp = ukarch_rstack_push(sp, (long)_ctx_x86_call0);
```
mogasergiu: Note to self: Write `ehtrampo_ectx_and_sysctx_store` in assembly instead so that we don't bother with this ABI alignment stuff.

include/uk/arch/ctx.h
```diff
@@ -340,6 +340,50 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
 void ukarch_ctx_switch(struct ukarch_ctx *store, struct ukarch_ctx *load);
 
+/*
+ * (Full) Execution environment functions must return into environment restoring
+ * function.
+ */
+typedef void (*ukarch_ehtrampo_entry)(struct ukarch_execenv *, long);
+
+/**
+ * Initializes a context struct with stack pointer and
```
michpappas: > Initializes a context struct with stack pointer and entrance function to trampoline to from an exception handler to an I/O-able context.

I had to read the implementation to understand the documentation. It should be the other way around 🙃 

Consider moving the sentence in line 356 to the top, as it's generally more useful to explain what it is before you explain how it works. I would even go as far as moving the "how" to the comments in the code.

Another thing that is missing, and again more imortant on how it works, is how to use it. So documentation should also state that you are expected to call `ukarch_ectx_jmp()`. Saying that, it would be worth considering whether the init and jump operations should be merged to a single call.

PS: "Entry" is more commonly used than "entrance" in this context.

arch/x86/ctx.c
```diff
@@ -189,3 +190,54 @@ void ukarch_ctx_init_entry3(struct ukarch_ctx *ctx,
...
+	sp -= ALIGN_UP(sizeof(*ee), UKARCH_EXECENV_END_ALIGN);
+	ee = (struct ukarch_execenv *)sp;
+
+	ukarch_ectx_sanitize((struct ukarch_ectx *)&ee->ectx);
+	ukarch_ectx_store((struct ukarch_ectx *)&ee->ectx);
+
+	ukarch_sysctx_store(&ee->sysctx);
+
+	/*
+	 * NOTE: This is very risky to do in ISR context as the compiler
```
michpappas: At first I thought you were trying to prevent the compiler from using `memcpy()`. Proposed phrasing, short and to the point:

> NOTE: Order is important. We save the registers after ectx . As the compiler could implement this using memcpy(), we run the risk of corrupting the ectx before we had a chance to save it.

include/uk/arch/ctx.h
```diff
@@ -340,6 +340,50 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * are saved in the slot of the execution environment passed as
+ * argument to the entrance function.
+ *
+ * This function is most useful when wanting to go from an exception
+ * handler to an I/O-able context while being able to return to the
+ * place that the exception handler was supposed to return to.
+ *
+ * @param ctx
+ *   Reference to context to initialize
+ * @param sp
```
michpappas, line 379: Perhaps you should clarify the purpose of that stack. Perhaps something along the lines of:
> Should point to the stack the target context will execute on. Unless you have a reason, most of the time that would be the auxilliary stack.

On second thought, one idea that could make the caller's life easier is to have el trampo use the aux stack by default, unless the caller passes a non-NULL pointer. Since both el trampo and the auxiliary stack will be part of the same library, that should not be too obtrusive.

Another thing I noticed is that el trampo pushes into the stack. Perhaps it's worth to mention that, in case the caller provides a pointer to space-confined stack.
 
In addition to the above, you might want to switch the order of `sp` under that of `r` in the docstring, to match the function's order of  parameters.

include/uk/arch/ctx.h
```diff
@@ -340,6 +340,50 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
  */
 void ukarch_ctx_switch(struct ukarch_ctx *store, struct ukarch_ctx *load);
 
+/*
+ * (Full) Execution environment functions must return into environment restoring
```
michpappas: With no parameter names and no documentation, this is quite cryptic. Would be worth moving the description from `ukarch_ctx_init_ehtrampo()` here. Also giving names to the parameters makes the reader's life easier, and IMO in a public API it's a must. Plus you can also use these names in docstrings.

arch/x86/ctx.c
```diff
@@ -189,3 +190,54 @@ void ukarch_ctx_init_entry3(struct ukarch_ctx *ctx,
...
+	struct ukarch_execenv *ee;
+
+	UK_ASSERT(ctx);
+	UK_ASSERT(r);
+	UK_ASSERT(sp);			/* a stack is needed */
+	UK_ASSERT(entry);		/* NULL as func will cause a crash */
+	UK_ASSERT(!(sp & UKARCH_SP_ALIGN_MASK)); /* sp properly aligned? */
+
+	/* We re-align the stack anyway just to be sure */
+	sp = ALIGN_DOWN(sp, UKARCH_EXECENV_END_ALIGN);
```
michpappas: If it's stated as requirement already, then pehaps an assert is more suitable. Either that or not make it a requirement and do the align internally.

michpappas (changes requested):

michpappas:
In addition to the above, it would be useful to provide in the cover letter and the commit message that implements the trampoline logic, what is the motivation for this, i.e. problem does it solve.

include/uk/arch/ctx.h
mogasergiu: > it would be worth considering whether the init and jump operations should be merged to a single call.

Thought about this as well. I did it like this the first time to align it with the other `ukarch_ctx_init_*` functions. I am trying to come up with a fitting name for something that implies it does both.

include/uk/arch/ctx.h
mogasergiu: > On second thought, one idea that could make the caller's life easier is to have el trampo use the aux stack by default, unless the caller passes a non-NULL pointer. Since both el trampo and the auxiliary stack will be part of the same library, that should not be too obtrusive.

That would be nice. The problem is the way it is used may differ drastically depending on what the target context may be (e.g. user app handler or nested exception handler)

mogasergiu:
Pushed rebase

mogasergiu:
Pushed updated addressing comments

include/uk/arch/ctx.h
michpappas: Okay I see, let's drop it then 👍🏼

include/uk/arch/ctx.h
```diff
@@ -339,6 +340,59 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * @param arg
+ *   Custom user-passed argument
+ */
+typedef void (*ukarch_ehtrampo_entry)(struct ukarch_execenv *execenv, long arg);
+
+/**
+ * Initializes a context struct with stack pointer and
+ * entry function to trampoline to from an exception handler
+ * into an I/O-able context.
+ * Does not guarantee any register states when entering the user-passed
```
michpappas: This can be also moved to the docstring of `ukarch_ehtrampo_entry()`.

include/uk/arch/ctx.h
```diff
@@ -339,6 +340,59 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * @param execenv
+ *   Pointer to the execution environment that the function will return to
+ * @param arg
+ *   Custom user-passed argument
+ */
+typedef void (*ukarch_ehtrampo_entry)(struct ukarch_execenv *execenv, long arg);
+
+/**
+ * Initializes a context struct with stack pointer and
+ * entry function to trampoline to from an exception handler
```
michpappas: In this line, s/to from/from/

However, as stated on the previous comment, perhaps it's better to focus on the information a caller would need: what it does and how to use it (as opposed to how it works). Notice also how you repeat yourself in two subsequent paragraphs:

>  * Initializes a context struct ... to trampoline to from an exception handler
>  * into an I/O-able context.

and on the very next paragraph again the same statement:

>  * This function is most useful when wanting to go from an exception
> * handler to an I/O-able context while being able to return to the
> * place that the exception handler was supposed to return to.

You also still don't tell the caller how to use it. I would propose something along the lines of:

> Initializes a context that allows  jumping from an exception handling context to a caller-defined function in an  I/O-able context.  Returning from the caller-defined function resumes execution from the place that the exception handler would resume from.
>
> After calling `ukarch_ehtrampo_init()` use `ukarch_ctx_jmp()` to jump to the target context.
> 
> NOTE: This function may return with a tainted extended context.

include/uk/arch/ctx.h
```diff
@@ -339,6 +340,59 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * and jump to instruction pointer stored in the context structure.
+ *
+ * @param ctx
+ *   Reference to context that shall be executed
+ */
+void ukarch_ctx_jump(struct ukarch_ctx *ctx) __noreturn;
+
+/**
+ * Function that can be executed as if outside exception context. After the
+ * function ends, execution will restore the state stored in the execenv
```
michpappas: after the function *returns*, execution will *resume*

include/uk/arch/ctx.h
```diff
@@ -339,6 +340,59 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * argument.
+ *
+ * @param execenv
+ *   Pointer to the execution environment that the function will return to
+ * @param arg
+ *   Custom user-passed argument
```
michpappas: User-defined argument

include/uk/arch/ctx.h
```diff
@@ -339,6 +340,59 @@ void ukarch_ctx_init_entry2(struct ukarch_ctx *ctx,
...
+ * Without saving or restoring anything, directly switch to stack pointer
+ * and jump to instruction pointer stored in the context structure.
+ *
+ * @param ctx
+ *   Reference to context that shall be executed
+ */
+void ukarch_ctx_jump(struct ukarch_ctx *ctx) __noreturn;
+
+/**
+ * Function that can be executed as if outside exception context. After the
```
michpappas: To keep consistent terminology:
> executed in the I/O-able context

michpappas (changes requested):
A few more improvements on the API documentation.

mogasergiu:
Pushed addressing comments
+
Removed the "I/O-able" term and replaced it with "a context where one can take exceptions or yield the current thread"
In earlier versions, the API would always implicitly enable interrupts, making it true that one can do even interrupt-based I/O. But, in current versions, the caller or the called function would have to enable interrupts manually, if there weren't already and if they wanted to.
The reason the API no longer implicitly enables interrupts is because in a scenario where the caller is already in an interrupt-enabled context, the API would needlessly re-enable them.
Therefore now, the API "teleports" the caller to a context where it can taint whatever register context it wants, but the state of interrupt flags is the state execution had before jumping to aforementioned context.

michpappas (approved):
Thanks!

andreittr (approved):

michpappas:

PR #1348, opened 2024-03-04
{plat,arch}: Align exception handling stacks among architectures

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Make it so that both ARM64 and x86_64 have their exception stacks
unified:
- three exception stacks: IRQ, trap, crit
- all three contiguous part of the same huge array

### DIFF
--- a/plat/kvm/arm/exceptions.S
+++ b/plat/kvm/arm/exceptions.S
@@ -32,6 +32,10 @@
 #include <uk/plat/config.h>
 #include <uk/arch/ctx.h>
 
+#define TYPE_CRIT				0
+#define TYPE_TRAP				1
+#define TYPE_IRQ				2
+
 .macro EXCHANGE_SP_WITH_X0
 	add sp, sp, x0	// new_sp = sp + x0
 	sub x0, sp, x0	// new_x0 = new_sp - x0 = sp + x0 - x0 = sp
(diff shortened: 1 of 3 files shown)


plat/kvm/arm/exceptions.S
```diff
@@ -69,34 +73,42 @@
...
 	 */
 	mrs     x0, sp_el0
 
 	/* Make it so that SP contains SP_EL0 and x0 contains old SP */
 	EXCHANGE_SP_WITH_X0
 
-.if \is_irq != 0
-	/* If we are an IRQ, use the IRQ stack instead. */
+.if \except_type == TYPE_CRIT
+	/* If we are a crit, use the crit stack instead. */
```
michpappas: More like "stay on the crit stack"?

michpappas (approved):
I have some comments on the overall choice of enforcing the 3-stack design universally, but that's out of this PR's scope so I won't elaborate.

Besides the small comment on wording, it looks good to me, so I'm leaving my RB already.

Some ideas for improvement:
* Right now it's possible that the trap stack corrupts the IRQ stack: Consider the case that an IRQ handler has heavy stack usage and overflows onto the trap stack, and somehow it causes a recoverable exception like a page fault (although it shouldn't). By the time execution resumes back to the IRQ context, it's state is corrupted. I'm not saying that we should cater for cases that are clearly wrong, but at least make it easier to debug, so perhaps we should at some point enforce that recoverable exceptions from IRQ context result into a panic.
* We should consider adding stack guards at some point.

andreittr (approved):

michpappas:

PR #1351, opened 2024-03-08
scripts/checkpatch: Do not warn about symbolic permissions

skuenzer:
This PR switches the checkpatch warning for symbolic permissions off (Examples are `S_IXUSR`, `S_IFREG`). Such a warning makes sense for a Linux kernel but less for us: We normally compile our kernel code together with a fully featured libc and even bind application logic directly to kernel code. We do not have the classical user-/kernel- code separation and rather prefer using symbolic values over numeric values.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -6,4 +6,5 @@
 --ignore NEW_TYPEDEFS
 --ignore FILE_PATH_CHANGES
 --ignore OBSOLETE
+--ignore SYMBOLIC_PERMS
 --max-line-length=80


michpappas (commented):
I'm a bit confused about the justification on kernel / user separation in the commit message. According to 
https://docs.kernel.org/dev-tools/checkpatch.html and 
and https://lore.kernel.org/lkml/CA+55aFw5v23T-zvDZp-MmD_EYxF8WbafwwB59934FV7g21uMGQ@mail.gmail.com/
the warning on `SYMBOLIC_PERMS` was added for readability:
> Everyone knows what 0644 is, but noone can read S_IRUSR | S_IWUSR |
> S_IRCRP | S_IROTH (*). Please don't do this.

What is the use-case from our side?

craciunoiuc (approved):
All good here. Thanks!

Approved-by: Cezar Craciunoiu <cezar.craciunoiu@unikraft.io>

skuenzer:
> I'm a bit confused about the justification on kernel / user separation in the commit message. According to https://docs.kernel.org/dev-tools/checkpatch.html and and https://lore.kernel.org/lkml/CA+55aFw5v23T-zvDZp-MmD_EYxF8WbafwwB59934FV7g21uMGQ@mail.gmail.com/ the warning on `SYMBOLIC_PERMS` was added for readability:
> 
> > Everyone knows what 0644 is, but noone can read S_IRUSR | S_IWUSR |
> > S_IRCRP | S_IROTH (*). Please don't do this.
> 
> What is the use-case from our side?

Nice references 😄. I think we have here opinions and preferences facing each other. My thinking is mostly compatibility motivated: We have the luxury that we can have one libc that defines the octal translation once and apply it for our kernel and user application. However, we can use this also to simplify switching between BSD and Linux encodings. This avoids completely the situation that there could be a misinterpretation of encodings. How realistic such scenario is for permission bits, is a different question. As far as I know, BSD systems use the same octal representation.
I think the macro indirection is always a good way to have ABIs compatibility configurable. If we apply such indirection as general rule, I think we should not do an exception with permission bits. Maybe this is something to get an agreement in the next maintainer meeting.

jschlumpp:
> However, we can use this also to simplify switching between BSD and Linux encodings. This avoids completely the situation that there could be a misinterpretation of encodings. How realistic such scenario is for permission bits, is a different question. As far as I know, BSD systems use the same octal representation. I think the macro indirection is always a good way to have ABIs compatibility configurable. 

The values seem to be defined next to the macros. So I think they are pretty much equivalent. 
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html

michpappas (approved):
Okay @skuenzer it's fine, just wanted to understand the rationale a bit better.

razvand (approved):

Issue #75, opened 2020-12-04
Add starting and debugging scripts


razvand:
Add scripts to start / debug Unikraft instances: such as starting in KVM with debugging support, starting in KVM with networking (bridge, `dnsmasq` for DNS etc.)

PR #1352, opened 2024-03-08
lib/posix-environ: Deduplicate env variables during boot

skuenzer:
This PR includes fixes for properly support PR [#75](https://github.com/unikraft/app-elfloader/pull/75) at app-elfloader. Most importantly, the PR introduces uniquifying environment variables during boot which finally enables overwriting compiled-in environment variables with the kernel parameter `env.vars`.

### DIFF
--- a/lib/posix-environ/environ.c
+++ b/lib/posix-environ/environ.c
@@ -3,8 +3,11 @@
  * Licensed under the BSD-3-Clause License (the "License").
  * You may not use this file except in compliance with the License.
  */
+#define _GNU_SOURCE
 #include <uk/libparam.h>
 #include <uk/essentials.h>
+#include <uk/init.h>
+#include <string.h>
 #include "environ.h"
 
 #define LIBPOSIX_ENVIRON_ENV_VAR(x)		e##x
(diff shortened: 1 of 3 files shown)


lib/posix-environ/environ.c
```diff
@@ -183,3 +186,44 @@ char **__environ = __init_env;
...
+}
+
+/* Eliminate duplicates from __init_env */
+static int uniquify_env(struct uk_init_ctx *ictx __unused)
+{
+	char **rpos; /*< read position */
+	char **wpos; /*< write position */
+
+	for (rpos = __init_env, wpos = __init_env; *rpos; rpos++) {
+		size_t namelen;
```
mogasergiu, line 210: Should we use `__sz` here?

lib/posix-environ/environ.c
```diff
@@ -183,3 +186,44 @@ char **__environ = __init_env;
 __weak_alias(__environ, ___environ);
 __weak_alias(__environ, _environ);
 __weak_alias(__environ, environ);
+
+/* Find the first occurrence of a variable within __init_env */
+static char **find_vnpos(const char *varname, size_t varname_len, char **end)
```
mogasergiu, line 191: Ditto `__sz`

lib/posix-environ/environ.c
```diff
@@ -183,3 +186,44 @@ char **__environ = __init_env;
...
+static int uniquify_env(struct uk_init_ctx *ictx __unused)
+{
+	char **rpos; /*< read position */
+	char **wpos; /*< write position */
+
+	for (rpos = __init_env, wpos = __init_env; *rpos; rpos++) {
+		size_t namelen;
+		char **opos; /*< overwrite position */
+
+		namelen = strchrnul(*rpos, '=') - *rpos;
```
mogasergiu, line 213: Since environment variables are `key - value` pairs, would it be a good idea to instead throw an assert here in case `strchrnul` did return a char pointer to `\0`? In that case maybe `strchr` and checking for `NULL` would suffice.

mogasergiu (commented):
Hey @skuenzer :ocean: :wave: . Thanks for the commits. Left some minor comments.

StefanJum:
Tested it and it works fine, will approve after @mogasergiu 's comments are solved.

lib/posix-environ/environ.c
skuenzer: Because the other functions (`setenv`, `getenv`, ...) in this lib (that were ported from musl) use libc types, I'd like to keep it here as well, so that we are consistent within this library.

lib/posix-environ/environ.c
skuenzer: I am matching the behavior of musl here (actually the ported functions, like `getenv`). You can find the same length computation in `getenv`, so I think this code should do the same assumption: an environment variable could be just the `key` without value.
I think an assertion is even worse because variables can be set from user input (e.g., `env.vars=` kernel parameter). If we would compile production builds, the assertion would be optimized out...

mogasergiu (approved):
Thanks for clarifications!

StefanJum (approved):

razvand (approved):

PR #1356, opened 2024-03-11
include/uk: Import tree implementations from FreeBSD + improvements

andreittr:
### Description of changes

This changeset imports the splay and rank-balanced tree implementations from [FreeBSD 13.3.0](https://cgit.freebsd.org/src/tree/sys/sys/tree.h?h=release/13.3.0) with the following changes & improvements:
- Replaced FreeBSD conventions with Unikraft:
  - namespaced `#define`s with `UK_` & use Unikraft header guards
  - use Unikraft-defined internal types
  - use Unikraft assertions
- Included fixes for type punning from previous work by @mschlumpp 
- Compiler warning fixes
- Added rb-tree variant that supports key-based lookup, as opposed to node-based; useful when full nodes (or similes thereof) are difficult or expensive to create

No existing Unikraft component makes use of these trees yet, but they serve as basis for ongoing work.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Standalone test `main.c`:
- traditional (node-based lookup)
```c
#include <uk/tree.h>

#include <stdio.h>

struct node {
	UK_RB_ENTRY(node) entry;
	int v;
};

static int node_cmp(struct node *n1, struct node *n2)
{
	return n1->v - n2->v;
}

UK_RB_HEAD(node_tree, node);
UK_RB_GENERATE_STATIC(node_tree, node, entry, node_cmp);

int main(void)
{
	struct node_tree head = UK_RB_INITIALIZER(head);

	struct node targ = { .v = 0 };
	struct node lot[12];

	struct node *r;

	r = UK_RB_FIND(node_tree, &head, &targ);

	printf("%p\n", r);

	for (int i = 0; i < 12; i++) {
		lot[i].v = i;
		r = UK_RB_INSERT(node_tree, &head, &lot[i]);
		UK_ASSERT(!r);
	}

	targ.v = 4;
	r = UK_RB_FIND(node_tree, &head, &targ);
	printf("%p\n", r);

	UK_RB_FOREACH(r, node_tree, &head) {
		printf("%p %i\n", r, r->v);
	}

	return 0;
}
```

- new key-based lookup (w/ structs as keys); `foo` should compile down into very sensible assembly with optimizations on (check w/ `objdump`)
```c
#include <uk/tree.h>

#include <stdio.h>

struct node {
	UK_RB_ENTRY(node) entry;
	size_t v;
	char buf[];
};

struct nkey {
	char *p;
	size_t l;
};

static struct nkey node_key(struct node *n)
{
	return (struct nkey){ n->buf, n->v };
}

static int node_cmp(struct nkey a, struct nkey b)
{
	return a.l - b.l;
}

UK_RB_HEAD(node_tree, node);
UK_RB_KEY_GENERATE_STATIC(node_tree, node, entry, node_cmp, node_key);

struct node_tree head = UK_RB_INITIALIZER(head);

void *foo(size_t x)
{
	return UK_RB_FIND(node_tree, &head, (struct nkey){ .l = x });
}

int main(void)
{
	struct node lot[12];

	struct node *r;

	r = UK_RB_FIND(node_tree, &head, (struct nkey){ .l = 4 });

	printf("%p\n", r);

	for (int i = 0; i < 12; i++) {
		lot[i].v = i;
		r = UK_RB_INSERT(node_tree, &head, &lot[i]);
		UK_ASSERT(!r);
	}

	r = UK_RB_FIND(node_tree, &head, (struct nkey){ .l = 4 });
	printf("%p\n", r);

	UK_RB_FOREACH(r, node_tree, &head) {
		printf("%p %zu\n", r, r->v);
	}

	return 0;
}
```

### DIFF
--- a/include/uk/tree.h
+++ b/include/uk/tree.h
@@ -31,6 +31,7 @@
 #ifndef	__UK_TREE_H__
 #define	__UK_TREE_H__
 
+#include <uk/assert.h>
 #include <uk/essentials.h>
 
 /*
@@ -939,13 +940,7 @@ name##_RB_NEXT(struct type *elm)					\
 	return (elm);							\
 }
 
-#if defined(_KERNEL) && defined(DIAGNOSTIC)
-#define UK__RB_ORDER_CHECK(cmp, lo, hi) do {				\
-	KASSERT((cmp)(lo, hi) < 0, ("out of order insertion"));		\
-} while (0)
-#else
-#define UK__RB_ORDER_CHECK(cmp, lo, hi) do {} while (0)
-#endif
+#define UK__RB_ORDER_CHECK(cmp, lo, hi) UK_ASSERT((cmp)((lo), (hi)) < 0)
 
 #define UK_RB_GENERATE_INSERT_NEXT(name, type, field, cmp, attr)	\
 /* Inserts a node into the next position in the RB tree */		\
--- a/include/uk/tree.h
+++ b/include/uk/tree.h
@@ -1,0 +1,1080 @@
+/*	$NetBSD: tree.h,v 1.8 2004/03/28 19:38:30 provos Exp $	*/
+/*	$OpenBSD: tree.h,v 1.7 2002/10/17 21:51:54 art Exp $	*/
+
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright 2002 Niels Provos <provos@citi.umich.edu>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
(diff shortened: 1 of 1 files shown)


jschlumpp:
I noticed that there are also function macros generated such as `name_RB_INSERT` with name being lower-case by our convention, which looks a bit strange overall. I wonder whether we should keep that or change the casing of the functions (to lower case).

andreittr:
> I noticed that there are also function macros generated such as `name_RB_INSERT` with name being lower-case by our convention, which looks a bit strange overall. I wonder whether we should keep that or change the casing of the functions (to lower case).

Good point, I gave that some thought as well, but I think there is some value in keeping those weird mixed-case generated names, especially considering they never get referenced directly:
- smaller diff versus upstream -- changes can and do happen upstream, and it makes our life easier when backporting
- it's immediately obvious during debug/objdump which functions are generated vs. hand-written; this latter point has helped me a bit when debugging this and another related templating header that I'm working on (which uses the same convention)

But if there's consensus that case consistency trumps these benefits we could go all-lowercase.

EDIT: I realised only very late that perhaps the original API intended for `name` to be all-uppercase when defining these trees. So it'd be:
```c
UK_RB_KEY_GENERATE_STATIC(NODE_TREE, node, entry, node_cmp, node_key);
```
which generates funcs like `NODE_TREE_RB_INSERT`, and callsites like `UK_RB_FIND(NODE_TREE, &head, ...`.
Seems legit :shrug:

jschlumpp:
I just checked the FreeBSD code and they also use lower case for the names. Let's just leave it at the mixed case names. In some cases you will have to use the mixed-case functions though (macro arguments behave different with commas) which doesn't look great.

jschlumpp (approved):

DeliaPavel (approved):

Mihnea0Firoiu (approved):
All good.

razvand (approved):

PR #1358, opened 2024-03-13
lib/posix-futex: Clear thread's objects upon thread exit

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

If a thread creates a futex that is not cleared upon application exit that futex will remain in futex_list after the unmap of the thread's stack causing an abort on the next iteration of the list.

Clear thread's objects upon thread exit.

### DIFF
--- a/lib/posix-futex/futex.c
+++ b/lib/posix-futex/futex.c
@@ -382,14 +382,29 @@ UK_LLSYSCALL_R_DEFINE(pid_t, set_tid_address, pid_t *, tid_ref)
 	return self_tid;
 }
 
-/* Thread exit handler that clears child TID at the stored reference */
-static void pfutex_child_cleartid_term(struct uk_thread *child __unused)
+static void thread_exit_handler(struct uk_thread *child)
 {
+	struct uk_list_head *itr, *tmp;
+	struct uk_futex *f;
+
+	/* Clear child TID at the stored reference */
 	if (child_tid_clear_ref != NULL) {
 		*((pid_t *) child_tid_clear_ref) = 0;
 		futex_wake((uint32_t *) child_tid_clear_ref, 0);
 	}
+
+	/* Clear this thread's entries from the list */
+	uk_spin_lock(&futex_list_lock);
+	uk_list_for_each_safe(itr, tmp, &futex_list) {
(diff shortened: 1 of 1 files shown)


mariapana (approved):
Looks good.

lib/posix-futex/futex.c
```diff
@@ -382,14 +382,25 @@ UK_LLSYSCALL_R_DEFINE(pid_t, set_tid_address, pid_t *, tid_ref)
 	return self_tid;
 }
 
-/* Thread exit handler that clears child TID at the stored reference */
-static void pfutex_child_cleartid_term(struct uk_thread *child __unused)
+static void thread_exit_handler(struct uk_thread *child __unused)
```
mariasfiraiala: Is the `__unused` attribute still needed now? As far as I can see `child` is used at line 399.

mariasfiraiala (commented):
All good on my side, besides a small comment.

flpostolache (commented):
I think a note to mention that this function is not SMP safe should be added so that when the full SMP support is added, this function would be also changed.
From my understanding, the `list_node` will break if two threads will try to delete two adjacent futexes at the same time on two different cores.

lib/posix-futex/futex.c
michpappas: Good catch, thanks! 👍🏼

michpappas:
> I think a note to mention that this function is not SMP safe should be added so that when the full SMP support is added, > this function would be also changed.
> From my understanding, the list_node will break if two threads will try to delete two adjacent futexes at the same time on two different cores.

@flpostolache thanks for pointing this out. I added a calls to acquire and release `futex_llist_lock` to protect the list. 

I also added a `break` statement once the thread's entry is found, as pointed out during a private discussion with @mogasergiu, and removed `__unused` as pointed out by @mariasfiraiala.

flpostolache (approved):
Thanks for the changes! Looks good now.

mariasfiraiala (approved):
All good on my side now too, thanks!

razvand (approved):

PR #1362, opened 2024-03-18
lib/ukcpio: Consistently overwrite existing destination

andreittr:
### Description of changes

Previously ukcpio would inconsistently ovewrite its destination depending on file types; notably, directories could not be overwritten by files or symlinks, and symlinks would be followed instead of overwritten by files.

This PR makes ukcpio consistently overwrite existing files/dirs/symlinks in the destination. Specifically:
- Files & symlinks whose path exists will attempt removal before retrying
- Directories whose path exists and is not a directory will attempt removal
- Directories that already exists will only have their mode bits changed

When attempting to remove a path, the following are tried, in order:
- `unlink()` -- for files & symlinks
- `rmdir()` -- for empty dirs
- `rename()` -- for non-empty dirs; this is chosen instead of recursive removal to minimize boot latency. Ideally one would not overwrite entire directory trees using an initrd anyway.

In addition, and as prerequisites, this PR also brings the following general improvements:
- Export the supported CPIO header format + accessor macros through `<uk/cpio.h>`, making it available for other Unikraft components/apps to use
- Remove special handling of `.`, as it had little benefit and produced several issues

### Prerequisite checklist

### Base target

### Additional configuration

Test with an embedded initrd + initrd that overwrites existing contents.

### DIFF
--- a/lib/ukcpio/cpio.c
+++ b/lib/ukcpio/cpio.c
@@ -58,6 +58,59 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
 int uk_syscall_r_mkdir(const char *, mode_t);
 int uk_syscall_r_symlink(const char *, const char *);
 int uk_syscall_r_stat(const char *, struct stat *);
+int uk_syscall_r_unlinkat(int, const char *, int);
+int uk_syscall_r_rename(const char *, const char *);
+
+static int
+try_rm_nonempty_dir(const char *path)
+{
+	/* rm -rf is nontrivial and costly, so we rename for speed */
+	/* TODO: revisit if efficient recursive dir removal is available */
+	int r;
+	char newpath[PATH_MAX];
+	char *newend = newpath + strlcpy(newpath, path, PATH_MAX);
+
+	if (unlikely(newend - newpath + 2 > PATH_MAX)) {
+		uk_pr_err("Cannot rename %s, path too long\n", path);
+		return -ENAMETOOLONG;
+	}
+	strcpy(newend, ".0");
+	r = uk_syscall_r_rename(path, newpath);
+	uk_pr_info("Rename '%s' to '%s': %d\n", path, newpath, r);
+	return r;
+}
+
+static int
+try_remove(const char *path)
+{
+	int r;
+
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, 0);
+	uk_pr_info("Unlink %s: %d\n", path, r);
+	if (!r || r == -ENOENT)
+		return 0;
+
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, AT_REMOVEDIR);
+	uk_pr_info("Rmdir %s: %d\n", path, r);
+	if (!r || r == -ENOENT)
+		return 0;
+	return try_rm_nonempty_dir(path);
+}
+
+static int
+write_file(int fd, const char *contents, size_t len)
+{
+	while (len) {
+		ssize_t written = uk_syscall_r_write(fd, contents, len);
+
+		if (unlikely(written < 0))
+			return written;
+		UK_ASSERT(len >= (size_t)written);
+		contents += written;
+		len -= written;
+	}
+	return 0;
+}
 
 static enum ukcpio_error
 extract_file(const char *path, const char *contents, size_t len,
@@ -70,39 +123,44 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
-	if (fd < 0) {
+	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_EXCL, 0);
+	if (unlikely(fd == -EEXIST)) {
+		uk_pr_info("Path exists, trying removal\n");
+		err = try_remove(path);
+		if (unlikely(err))
+			/* Not fatal, following open may still succeed */
+			uk_pr_warn("%s: Path cleanup failed: %s (%d)\n",
+				   path, strerror(-err), -err);
+		fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
+	}
+	if (unlikely(fd < 0)) {
 		uk_pr_err("%s: Failed to create file: %s (%d)\n",
 			  path, strerror(-fd), -fd);
 		ret = -UKCPIO_FILE_CREATE_FAILED;
 		goto out;
 	}
 
-	while (len) {
-		ssize_t written = uk_syscall_r_write(fd, contents, len);
-
-		if (written < 0) {
-			err = written;
-			uk_pr_err("%s: Failed to load content: %s (%d)\n",
-				  path, strerror(-err), -err);
-			ret = -UKCPIO_FILE_WRITE_FAILED;
-			goto close_out;
-		}
-		UK_ASSERT(len >= (size_t)written);
-		contents += written;
-		len -= written;
+	err = write_file(fd, contents, len);
+	if (unlikely(err)) {
+		uk_pr_err("%s: Failed to load content: %s (%d)\n",
+			  path, strerror(-err), -err);
+		ret = -UKCPIO_FILE_WRITE_FAILED;
+		goto close_out;
 	}
 
-	if ((err = uk_syscall_r_chmod(path, mode)))
+	err = uk_syscall_r_chmod(path, mode);
+	if (unlikely(err))
 		uk_pr_warn("%s: Failed to chmod: %s (%d)\n",
 			   path, strerror(-err), -err);
 
-	if ((err = uk_syscall_r_utime(path, &times)))
+	err = uk_syscall_r_utime(path, &times);
+	if (unlikely(err))
 		uk_pr_warn("%s: Failed to set modification time: %s (%d)",
 			   path, strerror(-err), -err);
 
 close_out:
-	if ((err = uk_syscall_r_close(fd))) {
+	err = uk_syscall_r_close(fd);
+	if (unlikely(err)) {
 		uk_pr_err("%s: Failed to close file: %s (%d)\n",
 			  path, strerror(-err), -err);
 		if (!ret)
@@ -119,26 +177,40 @@ extract_dir(const char *path, mode_t mode)
 
 	uk_pr_info("Creating directory %s\n", path);
 	r = uk_syscall_r_mkdir(path, mode);
-	if (r == -EEXIST) {
+	if (unlikely(r == -EEXIST)) {
 		struct stat pstat;
 
+		uk_pr_info("Path exists, checking type\n");
 		r = uk_syscall_r_stat(path, &pstat);
-		if (r < 0) {
+		if (unlikely(r < 0)) {
+			uk_pr_warn("%s: Cannot stat path: %s (%d)\n",
+				   path, strerror(-r), -r);
 			r = -EEXIST;
 			goto err_out;
 		}
 		if (!S_ISDIR(pstat.st_mode)) {
-			r = -EEXIST;
-			goto err_out;
+			/* Path exists but is not dir; remove & try again */
+			uk_pr_info("Path exists and not dir, trying removal\n");
+			r = try_remove(path);
+			if (unlikely(r)) {
+				uk_pr_warn("%s: Path cleanup failed: %s (%d)\n",
+					   path, strerror(-r), -r);
+				goto err_out;
+			}
+			r = uk_syscall_r_mkdir(path, mode);
+			if (unlikely(r))
+				goto err_out;
+			else
+				return UKCPIO_SUCCESS;
 		}
 
 		/* Directory already exists, set mode only */
+		uk_pr_info("Path exists and is dir, doing chmod\n");
 		r = uk_syscall_r_chmod(path, mode);
-		if (r < 0) {
+		if (unlikely(r < 0))
 			uk_pr_warn("%s: Failed to chmod: %s (%d)\n",
 				   path, strerror(-r), -r);
-		}
-	} else if (r < 0) {
+	} else if (unlikely(r < 0)) {
 		goto err_out;
 	}
 	return UKCPIO_SUCCESS;
@@ -161,12 +233,64 @@ extract_symlink(const char *path, const char *contents, size_t len)
 	target[len] = 0;
 
 	uk_pr_info("%s: Target is %s\n", path, target);
-	if ((r = uk_syscall_r_symlink(target, path))) {
-		uk_pr_err("%s: Failed to create symlink: %s (%d)\n",
-			  path, strerror(-r), -r);
-		return -UKCPIO_SYMLINK_FAILED;
+	r = uk_syscall_r_symlink(target, path);
+	if (r == -EEXIST) {
+		uk_pr_info("Path exists, trying removal\n");
+		r = try_remove(path);
+		if (!r)
+			r = uk_syscall_r_symlink(target, path);
 	}
-	return UKCPIO_SUCCESS;
+	if (likely(!r))
+		return UKCPIO_SUCCESS;
+
+	uk_pr_err("%s: Failed to create symlink: %s (%d)\n",
+		  path, strerror(-r), -r);
+	return -UKCPIO_SYMLINK_FAILED;
+}
+
+static enum ukcpio_error
+extract_section(const struct uk_cpio_header **headerp, char *fullpath,
+		const char *eof, size_t prefixlen)
+{
+	const struct uk_cpio_header *header = *headerp;
+	uint32_t mode = UKCPIO_U32FIELD(header->mode);
+	uint32_t filesize = UKCPIO_U32FIELD(header->filesize);
+	uint32_t namesize = UKCPIO_U32FIELD(header->namesize);
+	uint32_t mtime = UKCPIO_U32FIELD(header->mtime);
+	const char *fname = UKCPIO_FILENAME(header);
+	const char *data = UKCPIO_DATA(header, namesize);
+	enum ukcpio_error err;
+
+	if (unlikely(fname + namesize > eof)) {
+		uk_pr_err("File name exceeds archive bounds at %p\n", header);
+		return -UKCPIO_MALFORMED_INPUT;
+	}
+	if (unlikely(data + filesize > eof)) {
+		uk_pr_err("File exceeds archive bounds: %s\n", fname);
+		return -UKCPIO_MALFORMED_INPUT;
+	}
+
+	/* namesize includes trailing NUL */
+	if (unlikely(prefixlen + namesize > PATH_MAX)) {
+		uk_pr_err("Resulting path too long: %s\n", fname);
+		return -UKCPIO_MALFORMED_INPUT;
+	}
+	memcpy(fullpath + prefixlen, fname, namesize);
+
+	err = UKCPIO_SUCCESS;
+
+	if (UKCPIO_IS_DIR(mode))
+		err = extract_dir(fullpath, mode & 0777);
+	else if (UKCPIO_IS_FILE(mode))
+		err = extract_file(fullpath, data, filesize,
+				   mode & 0777, mtime);
+	else if (UKCPIO_IS_SYMLINK(mode))
+		err = extract_symlink(fullpath, data, filesize);
+	else
+		uk_pr_warn("File %s unknown mode %o\n", fullpath, mode);
+
+	*headerp = UKCPIO_NEXT(header, namesize, filesize);
+	return err;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/ukcpio/cpio.c
```diff
@@ -269,30 +184,30 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
-	const struct cpio_header *header = *headerp;
+	const struct uk_cpio_header *header = *headerp;
 
-	if ((char *)header >= eof || CPIO_FILENAME(header) > eof) {
+	if ((char *)header >= eof || UKCPIO_FILENAME(header) > eof) {
 		uk_pr_err("Truncated CPIO header at %p", header);
 		return -UKCPIO_INVALID_HEADER;
 	}
-	if (!valid_magic(header)) {
+	if (!ukcpio_valid_magic(header)) {
```
mogasergiu: `unlikely`?

lib/ukcpio/cpio.c
```diff
@@ -269,30 +184,30 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
 static enum ukcpio_error
-process_section(const struct cpio_header **headerp, char *fullpath,
+process_section(const struct uk_cpio_header **headerp, char *fullpath,
 		const char *eof, size_t prefixlen)
 {
-	const struct cpio_header *header = *headerp;
+	const struct uk_cpio_header *header = *headerp;
 
-	if ((char *)header >= eof || CPIO_FILENAME(header) > eof) {
+	if ((char *)header >= eof || UKCPIO_FILENAME(header) > eof) {
```
mogasergiu: `unlikely`?

lib/ukcpio/cpio.c
```diff
@@ -269,30 +184,30 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
+	if (!ukcpio_valid_magic(header)) {
 		uk_pr_err("Bad magic number in CPIO header at %p\n", header);
 		return -UKCPIO_INVALID_HEADER;
 	}
 
-	if (CPIO_ISLAST(header)) {
+	if (UKCPIO_ISLAST(header)) {
 		*headerp = NULL;
 		return UKCPIO_SUCCESS;
 	} else {
```
mogasergiu: Is this `if-else` needed? I think just this `if` is enough as it has a `return`. This way you at least avoid having additional indentation.

lib/ukcpio/cpio.c
```diff
@@ -269,30 +184,30 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
-		const char *data = CPIO_DATA(header, namesize);
+		uint32_t mode = UKCPIO_U32FIELD(header->mode);
+		uint32_t filesize = UKCPIO_U32FIELD(header->filesize);
+		uint32_t namesize = UKCPIO_U32FIELD(header->namesize);
+		uint32_t mtime = UKCPIO_U32FIELD(header->mtime);
+		const char *fname = UKCPIO_FILENAME(header);
+		const char *data = UKCPIO_DATA(header, namesize);
 		enum ukcpio_error err;
 
 		if (fname + namesize > eof) {
```
mogasergiu: Ok so there are more error cases that do not have `unlikely` here, I will just say once that I think it should be added, as writing `unlikely` at each of them is spammy and unproductive.

lib/ukcpio/include/uk/cpio.h
```diff
@@ -35,11 +35,103 @@
...
+		else if (buf[i] >= 'a' && buf[i] <= 'f')
+			val += (buf[i] - 'a') + 10;
+		else
+			return 0;
+	}
+	return val;
+}
+
+#define UKCPIO_ALIGN 4
+#define UKCPIO_ALIGN_4(ptr) ((void *)ALIGN_UP((uintptr_t)(ptr), UKCPIO_ALIGN))
```
mogasergiu: `UKCPIO_ALIGN_UP`?

lib/ukcpio/include/uk/cpio.h
```diff
@@ -35,11 +35,103 @@
...
+
+#define UKCPIO_U32FIELD(buf) \
+	((uint32_t)snhex_to_int((buf), 8))
+#define UKCPIO_FILENAME(header) \
+	((char *)(header) + sizeof(struct uk_cpio_header))
+#define UKCPIO_DATA(header, fnlen) \
+	((char *)UKCPIO_ALIGN_4((char *)(header) + \
+			       sizeof(struct uk_cpio_header) + (fnlen)))
```
mogasergiu: Misaligned with above paranthesis.

lib/ukcpio/include/uk/cpio.h
```diff
@@ -35,11 +35,103 @@
...
+#define UKCPIO_NEXT(header, fnlen, fsize) \
+	((struct uk_cpio_header *)UKCPIO_ALIGN_4(UKCPIO_DATA(header, fnlen) + \
+						(fsize)))
```
mogasergiu: Ditto

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,37 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+int uk_syscall_r_rename(const char *, const char *);
+
+static int
+try_remove(const char *path)
+{
+	int r;
+
+	uk_pr_info("Trying to unlink %s\n", path);
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, 0);
+	if (!r || r == -ENOENT)
```
mogasergiu, line 90: ~~Ditto about `unlikely` here and below on error cases.~~

EDIT: Nvm it actually makes sense to not have `unlikely` here.

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,37 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+		return 0;
+
+	uk_pr_info("Trying to rmdir %s\n", path);
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, AT_REMOVEDIR);
+	if (!r || r == -ENOENT) {
+		return 0;
```
mogasergiu, line 96: `if-else` below here is not necessary, as static code checker also pointed out.

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,37 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+	} else {
+		char newpath[PATH_MAX];
+		char *newend = newpath + strlcpy(newpath, path, PATH_MAX);
+
+		if (newend - newpath + 2 > PATH_MAX) {
+			uk_pr_err("Cannot rename %s, path too long\n", path);
+			return -ENAMETOOLONG;
+		}
+		strcpy(newend, ".0");
```
mogasergiu: I would rather we did
 ```C
 newend[0] = '.';
 newend[1] = '0';
 newend[2] = '\0';
 ```
instead of a function call.

Is there a reason why `strcpy` was used or just a minor oversight.

mogasergiu (changes requested):
Hey @andreittr ! Had a first  go at this. Left some comments (hoping that you will do the same for #1315 :stuck_out_tongue_winking_eye: ). I really think it is worth adding `unlikely` in some places, especially since this is an early mount hot path.

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,37 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+		uk_pr_info("Trying to rename %s to %s\n", path, newpath);
```
mogasergiu: This print will probably have more value after the function `rename` call so that we can also print the return code, I know that the caller will also do this, but I think increased verbosity here helps since RAMFS has lots of messages. I think it goes the same for the other "Trying" messages above. Since they are debug messages, a couple of additional if/else here and there for them should be fine as they would be optimized out in a performance optimized image.
E.g. `uk_pr_info("Rename operation %s to %s: %d\n", path, newpath, rc);`

lib/ukcpio/cpio.c
```diff
@@ -70,7 +101,13 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
+	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_EXCL, 0);
+	if (fd == -EEXIST) {
+		if ((err = try_remove(path)))
+			uk_pr_warn("Path cleanup failed: %s (%d)\n",
+				   strerror(-err), -err);
+		fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
```
mogasergiu, line 134: Let's also add an additional message here, so that we know when a try remove + open happened.

lib/ukcpio/cpio.c
```diff
@@ -161,12 +204,18 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
 
 	uk_pr_info("%s: Target is %s\n", path, target);
-	if ((r = uk_syscall_r_symlink(target, path))) {
-		uk_pr_err("%s: Failed to create symlink: %s (%d)\n",
-			  path, strerror(-r), -r);
-		return -UKCPIO_SYMLINK_FAILED;
+	r = uk_syscall_r_symlink(target, path);
+	if (r == -EEXIST) {
+		r = try_remove(path);
+		if (!r)
```
mogasergiu, line 240: Ditto.

lib/ukcpio/cpio.c
mogasergiu: Can you also give a more detailed description (both as a comment and commit message) of what this helps us with and why it will not break future references to this directory? Something like:
```
The newly processed initrd has a directory that must replace a previously existing same name directory. If the previous directory is not empty, simply rename it so that it does not overlap with the new directory.
```
By the way, isn't memory allocated for the files within the renamed directory lost? Also what if some other thread is holding a reference to this directory or files within this directory? (`EBUSY`)

Are these changes only intended for the early initial mount?

lib/ukcpio/cpio.c
```diff
@@ -70,7 +101,13 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
+	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_EXCL, 0);
+	if (fd == -EEXIST) {
+		if ((err = try_remove(path)))
```
mogasergiu: I am not a fan of this style of programming. I would separate this assignment of `err` outside the `if` and then check with an `unlikely`... Also give a reason (through a comment) saying why a `uk_pr_warn` is enough and it's not a stop reason.

lib/ukcpio/cpio.c
```diff
@@ -128,8 +165,14 @@ extract_dir(const char *path, mode_t mode)
 			goto err_out;
 		}
 		if (!S_ISDIR(pstat.st_mode)) {
-			r = -EEXIST;
-			goto err_out;
+			/* Path exists but is not dir; remove & try again */
+			if (unlikely((r = try_remove(path))))
```
mogasergiu: Ditto, let's separate this assignment and add some message?

mogasergiu (changes requested):
Hey, I had a second look, it's mostly that I feel like more verbosity regarding info/dbg messages as well as comments would help in the long term. I assume that, if disabled, such messages on production images have no impact, so I think it is worth the effort of increasing the verbosity level here.

lib/ukcpio/cpio.c
andreittr: > Is there a reason why strcpy was used or just a minor oversight.

Largely to keep the appended string configurable/of arbitrary length. I doubt you'd argue the same point if the line had been
```c
strcpy(newend, ".cpio_rename~")
```
There's nothing fundamental about `.0`, so I don't think it makes sense to special-case it this hard.

andreittr: We do not rename directories because of any concern for their content, rather because `rm -rf` is annoyingly non-trivial to implement using POSIX APIs, and CPIO extraction is latency-critical for boot.

Ideally you would never get into a situation where the initrd replaces a directory with something else, but in the exceedingly rare case that does happen, we don't want to waste code & time recursively walking a dir tree unlinking things. Thus the rename.

> By the way, isn't memory allocated for the files within the renamed directory lost?

I don't see how that would happen, renamed directories don't lose their contents.

> Also what if some other thread is holding a reference to this directory or files within this directory? (EBUSY)

If you hold an open reference to a file, that file will not `EBUSY` when trying to unlink/rename it (at least not on *nix; Windows behaves like that), the two things are orthogonal. Indeed, an open reference keeps a file "alive" even after unlinking.

lib/ukcpio/cpio.c
andreittr: Seeing that this isn't covered by our coding conventions, nor is it reported by checkpatch, I think it is very much a matter of taste, and I am a fan of this style when the assignments are short and concise.

lib/ukcpio/cpio.c
andreittr: The static code checker unfortunately can't tell that the else block is there to prevent a `PATH_MAX` sized object from coming into scope unless certain conditions are met, you need human reasoning for that. Compiler optimizations _may_ be clever enough to avoid allocating `newpath` unless on the branch making use of it, but it's nowhere near guaranteed.

lib/ukcpio/cpio.c
mogasergiu: Ok but the separation would be worth here so that we could add `unlikely` I would say :). I would not add the function call inside the `unlikely`.

lib/ukcpio/cpio.c
mogasergiu: > I don't see how that would happen, renamed directories don't lose their contents.

I understand. What I meant is they would remain allocated there but never be reachable.

> If you hold an open reference to a file

I meant the path of the file not an open file descriptor. The path would no longer be reachable.

lib/ukcpio/cpio.c
mogasergiu: Ok so the only reason you keep this is for optimizations to potentially and undeterministically be able to avoid allocating a buffer on the stack?

lib/ukcpio/cpio.c
andreittr: I very strongly favour minimizing the scope of variables to where they're actually used, since humans (even programmers) are _absolutely terrible_ at keeping track of a lot of things at once. Thus I think an indentation level is an OK price to pay for the separate scope introduced by a block.

That is to say if this `else` weren't there, I'd most likely just make a naked block around said code, which AFAIK is more frowned upon than redundant elses.

lib/ukcpio/include/uk/cpio.h
andreittr: Eh, as long as it's not used anywhere else yet, I'm agnostic to its name; can be `UKCPIO_ALIGN_UP`.

lib/ukcpio/cpio.c
andreittr: I think you misunderstand, the "undeterministic" optimizations would happen if I were to remove the else and declare `newpath` in the function scope. The way it is now (with "redundant" else) it is guaranteed by the C standard that `newpath` will only be allocated when it comes in scope, regardless of how dumb your compiler is.

lib/ukcpio/cpio.c
andreittr: > I understand. What I meant is they would remain allocated there but never be reachable.

That is (mostly) correct -- they won't be reachable by their old paths, and that is quite unfortunate :( but wasted RAM is a cost we pay for lower extraction latencies. We may reconsider this tradeoff in the future. Or remove it entirely with a ramfs that can natively do recursive directory unlinks (in the works).

> I meant the path of the file not an open file descriptor. The path would no longer be reachable.

Correct.

lib/ukcpio/cpio.c
andreittr: You'd be surprised what I consider pretty :stuck_out_tongue: `if (unlikely((r = try_remove(path))))`
But sure, if it helps readability that much, it's not a huge change. But fair warning for future work of mine.

andreittr:
Force-push: addressed reviews.

lib/ukcpio/cpio.c
mogasergiu: And that is helpful how in our case? I would rather always have the same predictable frame allocation.

> it is guaranteed by the C standard that newpath will only be allocated when it comes in scope

it is guaranteed by the C standard that newpath ~~will only~~ **may or may not be** be allocated when it comes in scope. Is that really a guarantee?

lib/ukcpio/cpio.c
mogasergiu: > are absolutely terrible at keeping track of a lot of things at once

My views on this is that if that is the case, your function is just too long and not modularized enough. Subscope declarations may be nice to have and a good addition, I agree, but I don't think this is the case here for such short functions.

andreittr: Maybe you're right, that latter block does make sense as a separate function `extract_section` since that's precisely what it does. Will split out.

lib/ukcpio/cpio.c
andreittr: Perhaps it's better to split off in its own function then; that also completely isolates the decision to do rename instead of `rm -rf` into a single-purpose function.

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,44 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+	char newpath[PATH_MAX];
+	char *newend = newpath + strlcpy(newpath, path, PATH_MAX);
+
+	if (unlikely(newend - newpath + 2 > PATH_MAX)) {
+		uk_pr_err("Cannot rename %s, path too long\n", path);
+		return -ENAMETOOLONG;
+	}
+	strcpy(newend, ".0");
+	r = uk_syscall_r_rename(path, newpath);
+	uk_pr_info("rename '%s' to '%s': %d\n", path, newpath, r);
```
mogasergiu: Capital letter for "Rename...". Unless you meant it as a "<syscall_name>..." format... in which case I would add a `()` to make it clear, i.e. `"rename() '%s'...` but I would prefer the former as it provides no additional value considering the dbg level messages already say the syscall name and its args

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,44 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+	return r;
+}
+
+static int
+try_remove(const char *path)
+{
+	int r;
+
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, 0);
+	uk_pr_info("unlink %s: %d\n", path, r);
```
mogasergiu: Ditto. Let's keep all debug messages aligned with other existing ones (starting with capital).

There are others below as well (like `rmdir`), won't spam `ditto` for each one though :grin:

lib/ukcpio/cpio.c
```diff
@@ -58,6 +58,44 @@ int uk_syscall_r_utime(const char *, const struct utimbuf *);
...
+	if (!r || r == -ENOENT)
+		return 0;
+
+	r = uk_syscall_r_unlinkat(AT_FDCWD, path, AT_REMOVEDIR);
+	uk_pr_info("rmdir %s: %d\n", path, r);
+	if (!r || r == -ENOENT)
+		return 0;
+	return try_rm_nonempty_dir(path);
```
mogasergiu, line 97: Love it :+1: :smiling_face_with_three_hearts:

lib/ukcpio/cpio.c
```diff
@@ -70,8 +108,17 @@ extract_file(const char *path, const char *contents, size_t len,
 
 	uk_pr_info("Extracting %s (%zu bytes)\n", path, len);
 
-	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_TRUNC, 0);
-	if (fd < 0) {
+	fd = uk_syscall_r_open(path, O_CREAT | O_WRONLY | O_EXCL, 0);
+	if (unlikely(fd == -EEXIST)) {
+		uk_pr_info("Path exists, trying removal\n");
```
mogasergiu, line 128: just "removal" feels dishonest :thinking: maybe "overwrite"?

lib/ukcpio/cpio.c
```diff
@@ -81,7 +128,7 @@ extract_file(const char *path, const char *contents, size_t len,
 	while (len) {
 		ssize_t written = uk_syscall_r_write(fd, contents, len);
```
mogasergiu: See, yet another one :yum: . Could it be that we could replace this loop with a reusable `write_all`-like function?

lib/ukcpio/cpio.c
```diff
@@ -119,26 +169,41 @@ extract_dir(const char *path, mode_t mode)
...
+				goto err_out;
+			else
+				return UKCPIO_SUCCESS;
 		}
 
 		/* Directory already exists, set mode only */
+		uk_pr_info("Path exists and is dir, doing chmod\n");
 		r = uk_syscall_r_chmod(path, mode);
-		if (r < 0) {
+		if (unlikely(r < 0)) {
```
mogasergiu: You can probably drop the brackets.

lib/ukcpio/cpio.c
```diff
@@ -161,12 +226,65 @@ extract_symlink(const char *path, const char *contents, size_t len)
...
+	if (unlikely(prefixlen + namesize > PATH_MAX)) {
+		uk_pr_err("Resulting path too long: %s\n", fname);
+		return -UKCPIO_MALFORMED_INPUT;
+	}
+	memcpy(fullpath + prefixlen, fname, namesize);
+
+	err = UKCPIO_SUCCESS;
+
+	if (UKCPIO_IS_DIR(mode)) {
+		err = extract_dir(fullpath, mode & 0777);
```
mogasergiu, line 283: Ditto

mogasergiu (changes requested):
Had another look. :pray:

lib/ukcpio/cpio.c
andreittr: I feel "overwrite" is misleading w.r.t. the operation being attempted: actual unlink/rmdir/wtvr, followed by extraction (the latter of which is explicitly said by the very previous info msg). This is relevant because "overwriting" a path is ambiguous: if the path is a symlink reasonable minds will disagree as to what "overwriting" that symlink means in casual conversation: whether to replace the symlink or what it points to. Indeed `open` without `O_EXCL` does the latter, and it's explicitly not what we want to happen.
Thus I want to be explicit that we're "removing" path before the extraction happens. That has pretty clear semantics.

lib/ukcpio/cpio.c
andreittr: Yet another what?
Either way, such a reusable function falls outside the scope of this PR (and we only loop write files once here).

mogasergiu: You are right, I understand I may seem like I am overstepping the PR's boundaries. I may have created a misunderstanding. What I meant is: "Can you move this `while` in a separate `file_write` function in `cpio.c` that maybe other functions could reuse as well?"

mogasergiu (approved):
Nice! Thanks a lot for your cooperation and quick updates :pray: ! Enjoyed reviewing this! :muscle: 

eduardvintila (approved):
All good here, thanks!

razvand (approved):

Issue #580, opened 2022-10-07
Unit Tests for Internal Libraries


razvand:
### Feature request summary

Unit testing ensure a good quality of the implementation.
Existence of unit tests allows regression testing: any new feature will be validated with existing unit tests.
Add unit tests (using [uktest](https://unikraft.org/docs/develop/writing-tests/)) for [internal Unikraft libraries](https://github.com/unikraft/unikraft/tree/staging/lib).

Each item represents an issue to be resolved as a PR which introduces the relevant `tests/` folder and tests to the corresponding library.

```[tasklist]
- [ ] Introduce `tests/` for `9pfs`
- [ ] Introduce `tests/` for `devfs`
- [ ] Introduce `tests/` for `fdt`
- [ ] Introduce `tests/` for `isrlib`
- [ ] Introduce `tests/` for `nolibc`
- [ ] Introduce `tests/` for `posix-libdl`
- [ ] Introduce `tests/` for `posix-process`
- [ ] Introduce `tests/` for `posix-sysinfo`
- [ ] Introduce `tests/` for `posix-user`
- [ ] Introduce `tests/` for `ramfs`
- [ ] Introduce `tests/` for `syscall_shim`
- [ ] Introduce `tests/` for `ubsan`
- [ ] Introduce `tests/` for `uk9p`
- [ ] Introduce `tests/` for `ukalloc`
- [ ] Introduce `tests/` for `ukallocbbuddy`
- [ ] Introduce `tests/` for `ukallocpool`
- [ ] Introduce `tests/` for `ukallocregion`
- [x] Introduce `tests/` for `ukargparse`
- [ ] Introduce `tests/` for `ukblkdev`
- [ ] Introduce `tests/` for `ukboot`
- [ ] Introduce `tests/` for `ukbus`
- [ ] Introduce `tests/` for `ukcpio`
- [ ] Introduce `tests/` for `ukdebug`
- [ ] Introduce `tests/` for `ukfalloc`
- [ ] Introduce `tests/` for `ukfallocbuddy`
- [ ] Introduce `tests/` for `uklibparam`
- [ ] Introduce `tests/` for `uklock`
- [ ] Introduce `tests/` for `ukmmap`
- [ ] Introduce `tests/` for `ukmpi`
- [ ] Introduce `tests/` for `uknetdev`
- [ ] Introduce `tests/` for `ukrust`
- [ ] Introduce `tests/` for `uksched`
[...]

PR #1367, opened 2024-03-29
lib/isrlib: Adding unit tests for isrlib

ahmd-kamel:
### Prerequisite checklist

### Base target

### Description of changes

This PR adhere to the opend issue #580 .

I have added a full unit test cases file in order to check the correctness and point of weaknesses in the following library.
While running the tests you will find some unit tests fails you need to check weather this test case is valid or not based on the purpose of the functions that is implemented inside isrlib. Finally i wanna highlight one result approximately the whole functions inside isrlib do not have any checking for false inputs like passing null pointer you will see at the first or at the end of each test suite I commented a line of code this line actually can make the system to crash if the line executed. 

- Changing the config.uk inside the library in order to generate the main config.uk correctly.
-  Adding the unit_test.c for the library after going throw the whole library to understand the purposes of functions inside.
- Editing the makefile to add the unit_test.c with executables in order to perform unit testing to the library.


lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/test.h>
+#include <uk/isr/string.h>
+
+/* 1- Threre is no type checking inside memset_isr
```
RaduNichita:
```suggestion
/* 1 - There is no type checking inside memset_isr
```

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+
+	memcpy_isr(dest_0, src_0, 10);
+	UK_TEST_EXPECT_BYTES_EQ(src_0, dest_0, 10);
+
+	char src_1[14] = "Hello, World!";
+	char dest_1[14];
+	char result_1[5] = "Hello";
+
+	memcpy_isr(dest_1, src_1, 5);
+	UK_TEST_EXPECT_BYTES_EQ(result_1, dest_1, 5);
```
RaduNichita, line 62: I think you can use src_1 instead of the result_1, just as you did in the previous example from the same test

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+	/* memcpy_isr(NULL, src_1, 14); */
+	/* memcpy_isr(result_2, NULL, 14); */
+
+	char src_0[10] = "0123456789";
+	char dest_0[10];
+
+	memcpy_isr(dest_0, src_0, 10);
+	UK_TEST_EXPECT_BYTES_EQ(src_0, dest_0, 10);
+
+	char src_1[14] = "Hello, World!";
```
RaduNichita, line 57: I would mark this as const and do not modify it in the testcase

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+	char dest_1[14];
+	char result_1[5] = "Hello";
+
+	memcpy_isr(dest_1, src_1, 5);
+	UK_TEST_EXPECT_BYTES_EQ(result_1, dest_1, 5);
+
+	char result_2[15] = "HelloHellold!";
+
+	memcpy_isr(src_1 + 5, src_1, 5);
```
RaduNichita, line 66: If you want to do some overlap of memory, I think it would be better to create a new char* instead of modifying the source.

In that way, if there are some more usecases to be able to this test, it would be easier to follow.

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+	UK_TEST_EXPECT_BYTES_EQ(src_1, result_2, 14);
+
+	memcpy_isr(result_2, src_1, 0);
+	UK_TEST_EXPECT_BYTES_EQ(src_1, result_2, 14);
+
+	char src_2[1024] = {'A'};
+	char dest_2[1024];
+
+	memcpy_isr(dest_2, src_2, 1024);
```
RaduNichita, line 75: What do you want to test in this scenario? How is this different from the first scenario from this test.

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+ *  the following line case crash for accessing nullptr
+ */
+
+UK_TESTCASE(isrlib_testsuite, test_strcmp_isr)
+{
+	/* ret = strcmp_isr(NULL, NULL); */
+
+	int ret = strcmp_isr("unikraft", "unikraft");
+
+	UK_TEST_EXPECT_SNUM_EQ(ret, 0);
```
RaduNichita, line 146: I would also add a test in which the result of EQ is different from 0.

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+ */
+
+UK_TESTCASE(isrlib_testsuite, test_memrchr_isr)
+{
+	/* memrchr_isr(NULL, '\0', 2); */
+
+	char array_0[12] = "Hello World";
+	char *ptr = (char *)memrchr_isr(array_0, 'o', 11);
+
+	UK_TEST_EXPECT_PTR_EQ(ptr, array_0 + 7);
```
RaduNichita: I would organize the test in the following way. What do you think?

```suggestion

	char *ptr = (char *)memrchr_isr(array_0, 'o', 11);
	UK_TEST_EXPECT_PTR_EQ(ptr, array_0 + 7);
```

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+
+	ptr = (char *)memrchr_isr(array_0, 'H', 11);
+
+	UK_TEST_EXPECT_PTR_EQ(ptr, array_0);
```
RaduNichita: Same for this

```suggestion

	ptr = (char *)memrchr_isr(array_0, 'H', 11);
	UK_TEST_EXPECT_PTR_EQ(ptr, array_0);
```

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,590 @@
...
+	memmove_isr(src_0, src_0, 10);
+	UK_TEST_EXPECT_BYTES_EQ(expect_0, src_0, 10);
+
+	char src_1[15] = "Hello, World!";
+	char expect_1[15] = "World, World!";
+
+	memmove_isr(src_1, src_1 + 7, 5);
+	UK_TEST_EXPECT_BYTES_EQ(expect_1, src_1, 15);
+
+	/* incase of forward overlapping the method */
```
RaduNichita:
```suggestion
	/* in case of forward overlapping the method */
```

RaduNichita (changes requested):
Thanks for the contribution @ahmd-kamel 

I did a first pass, please see my comments.

lib/isrlib/tests/test_isrlib.c
ahmd-kamel: I know you might see no difference, but here I am testing the case of copying a large string at the first I am just testing copying of small string.

lib/isrlib/tests/test_isrlib.c
ahmd-kamel: I have just add a test case for EQ in line 158 and 159.

lib/isrlib/tests/test_isrlib.c
ahmd-kamel: The purposes of this string is to help me test :
1 - Copying sub string  from the whole string. --> Not the main purpose but I take it on my way.
2 - Test the case of self overlapping. --> the main purpose you can see the following snip shot.

Screenshot from 2024-05-03 03-35-43

I take the memcpy as my reference so since  memcpy_isr doing the same function we must expect the same results.

lib/isrlib/tests/test_isrlib.c
ahmd-kamel: You are right, I will edit once we finish the conversation per this test suite.

lib/isrlib/tests/test_isrlib.c
ahmd-kamel: I have added the snip shot in the above conversation I think it's all linked together.

StefanJum (commented):
Looks good, @ahmd-kamel please also squash the commits.

razvand:
> Looks good, @ahmd-kamel please also squash the commits.

@ahmd-kamel , could you please squash the commits according to feedback from @StefanJum ?

ahmd-kamel:
Thank you, @StefanJum, @razvand. I apologise for the late response.
 I have squashed only the last two commits.
 I wanna hear if there are any notes that can make the PR better.

StefanJum (changes requested):
@ahmd-kamel you need to squash *all* commits into one. That means only one commit should appear in the `commit` tab.
You can follow the instructions [here](https://www.freecodecamp.org/news/git-squash-commits/).

Also, about the nullptr acesses, you can add the proper checks nor nullptrs in the faulty functions, or add a FIXME before the comment so we don't forget about that.

ahmd-kamel:
Hello, @StefanJum as requested in your review I have squashed all commits in one commit, also I add `FIXME:` tag in comments.

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,618 @@
...
+	ret = strncmp_isr("unikraft", "unikraftfft", 9);
+	UK_TEST_EXPECT_SNUM_LT(ret, 0);
+
+	ret = strncmp_isr("\0", "", 1);
+	UK_TEST_EXPECT_SNUM_EQ(ret, 0);
+}
+
+/*
+ *  FIXME: There is no type checking inside memrchr_isr
+ *  the following comment causes crash for accessing nullptr
```
RaduNichita, line 186: What is the following comment? I think it should be place right under this for an easy understanding.

Same for the rest.

lib/isrlib/tests/test_isrlib.c
```diff
@@ -0,0 +1,618 @@
...
+	ret = strlcat_isr("Hello, world!", "", 20);
+	UK_TEST_EXPECT_SNUM_EQ(ret, 13);
+
+	char dist[] = "Hello";
+	char src[] = ", world!";
+
+	ret = strlcat_isr(dist, src, 10);
+	UK_TEST_EXPECT_SNUM_EQ(ret, 13);
+
+	/* take a look at this case and the above they */
```
RaduNichita, line 598: Can you better explain the situation here, please? What is different?

RaduNichita (commented):
It looks better, I still have left some question / suggestions @ahmd-kamel

razvand:
@ahmd-kamel , could you check the comments from @RaduNichita ?

PR #1368, opened 2024-03-30
include/uk: Replace libc types with Unikraft defined

rares-miculescu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Replace libc types with Unikraft defined, to make `include/uk/` independent of libc.

### DIFF
--- a/plat/kvm/x86/time.c
+++ b/plat/kvm/x86/time.c
@@ -84,7 +84,7 @@ void ukplat_time_fini(void)
 {
 }
 
-uint32_t ukplat_time_get_irq(void)
+__u32 ukplat_time_get_irq(void)
 {
 	return 0;
 }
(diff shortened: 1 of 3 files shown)


plat/kvm/x86/time.c
```diff
@@ -38,6 +38,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <stdint.h>
```
michpappas: It makes sense that the change in the declaration (this header) must come together with updating the definition (in `plat/kvm/x86/time.c`). So perhaps you can also add a commit that updates the latter.

Also a general comment on your commit messages; there is no need for a line-break on every sentence. This is applicable only to documents. Instead, you should break at the 70-char limit similarly to:
```
Include `stdint/h` temporarily. Because of the removed include from
`include/uk/plat/time.h`, it is necessary to include it here. It will
be deleted when the libc data types from `plat/` are replaced with
Unikraft defined.
```

mariapana (approved):
Looks good.

andreistan26 (approved):
There is another definition of `ukplat_time_get_irq()` in `plat/common/arm/time.c` shouldn't that be changed as well?

Other than that it looks good!

michpappas (approved):

PR #1370, opened 2024-03-30
Fix calloc uses in vfscore

dorjoy03:
### Prerequisite checklist

### Base target

### Description of changes

I was grepping through the codebase looking into the malloc like function calls and saw some "calloc" calls in lib/vfscore where the arguments were reversed. Per 'man calloc', the first parameter should be the number of elements and the second parameter should be the size of a single element i.e., the "sizeof(...)" argument should be the second argument.

This PR has 2 types of changes:
1. It fixes the places that had calloc arguments reversed (~~the niov ones look like they might cause real bugs~~ probably won't. searching on google seems to suggest they would allocate the same amount of memory so wouldn't cause any bugs.)
2. Also it uses sizeof(variable) instead of sizeof(struct) for consistency with the codebase

### DIFF
--- a/lib/vfscore/dentry.c
+++ b/lib/vfscore/dentry.c
@@ -69,7 +69,7 @@ struct dentry *
 dentry_alloc(struct dentry *parent_dp, struct vnode *vp, const char *path)
 {
 	struct mount *mp = vp->v_mount;
-	struct dentry *dp = (struct dentry*)calloc(sizeof(*dp), 1);
+	struct dentry *dp = calloc(1, sizeof(*dp));
 
 	if (!dp) {
 		return NULL;
--- a/lib/vfscore/mount.c
+++ b/lib/vfscore/mount.c
@@ -161,7 +161,7 @@ UK_SYSCALL_R_DEFINE(int, mount, const char*, dev, const char*, dir,
 	/*
 	 * Create VFS mount entry.
 	 */
-	mp = calloc(1, sizeof(struct mount));
+	mp = calloc(1, sizeof(*mp));
 	if (!mp) {
 		error = ENOMEM;
 		goto err1;
--- a/lib/vfscore/syscalls.c
+++ b/lib/vfscore/syscalls.c
@@ -192,7 +192,7 @@ sys_open(char *path, int flags, mode_t mode, struct vfscore_file **fpp)
 		}
 	}
 
-	fp = calloc(sizeof(struct vfscore_file), 1);
+	fp = calloc(1, sizeof(*fp));
 	if (!fp) {
 		error = ENOMEM;
 		goto out_unlock;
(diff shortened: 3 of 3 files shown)


RaduNichita (approved):
LGTM, many thanks @dorjoy03 

StefanJum (commented):
This looks fine @dorjoy03, but I would go a bit further and replace all `*alloc` calls with the equivalent `uk_*alloc` call. That will mean also replacing the `free` calls with `uk_free` when the structures are freed.
We do not want to call library functions from the unikraft core if not needed.

razvand:
> This looks fine @dorjoy03, but I would go a bit further and replace all `*alloc` calls with the equivalent `uk_*alloc` call. That will mean also replacing the `free` calls with `uk_free` when the structures are freed. We do not want to call library functions from the unikraft core if not needed.

@dorjoy03 , could you please update the PR according to feedback from @StefanJum ?

dorjoy03:
Hey sorry for the delay. I will see if I can make some time to make the necessary changes.

dorjoy03:
> This looks fine @dorjoy03, but I would go a bit further and replace all `*alloc` calls with the equivalent `uk_*alloc` call. That will mean also replacing the `free` calls with `uk_free` when the structures are freed. We do not want to call library functions from the unikraft core if not needed.

@StefanJum @razvand It makes sense to just replace all *alloc calls in one go. I had a look into how many "free" calls are in this repo and seems like there are quite a few. Although the task is mechanical to some degree, I don't think I can make the time to go through all the call sites and replace them. Sorry about that. I suggest we merge this as is if you guys think it adds any value, otherwise feel free to close this one as there is no bug without this patch tbh.

razvand:
> > This looks fine @dorjoy03, but I would go a bit further and replace all `*alloc` calls with the equivalent `uk_*alloc` call. That will mean also replacing the `free` calls with `uk_free` when the structures are freed. We do not want to call library functions from the unikraft core if not needed.
> 
> @StefanJum @razvand It makes sense to just replace all *alloc calls in one go. I had a look into how many "free" calls are in this repo and seems like there are quite a few. Although the task is mechanical to some degree, I don't think I can make the time to go through all the call sites and replace them. Sorry about that. I suggest we merge this as is if you guys think it adds any value, otherwise feel free to close this one as there is no bug without this patch tbh.

@dorjoy03 , let's move this to the next release (0.18), then. Note there are also some formatting / checkpatch issues that need fixing. Let's wait for you to get some time to update it to its full form and then we can merge it.

razvand:
@dorjoy03 , we're aiming to have this PR part of the next release (planned for the end of August). Could you please make updates to the PR by then? Or do we leave it for later?

dorjoy03:
> @dorjoy03 , we're aiming to have this PR part of the next release (planned for the end of August). Could you please make updates to the PR by then? Or do we leave it for later?

@razvand Thanks for the reminder. I don't think I will able to look into this in the near future. Sorry about that. I suggest we close this PR and maybe someone else could pick up if needed. What do you think?

StefanJum:
Will be covered by a more extensive pr

PR #972, opened 2023-07-05
plat: Migrate rtc pl031 to drivers/ukrtc/

rares-miculescu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Description of changes

Base target

  -  Architecture(s): `ARM64`
  -  Platform(s): N/A
  -  Application(s): N/A

RTC PL031 driver moved from `plat/drivers/` to `drivers/ukrtc`.

PR #1373, opened 2024-04-01
plat: Rework paged memory init

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series reworks the initialization of paged memory to provide a more flexible implementation that is capable of handling regions beyond the limits defined in the boot page-tables. The motivation for this change is to allow mapping device regions that are unknown at compile-time, such as _Unprotected IPA Alias_ regions of Arm CCA Realms [1][2], the address of which depends on the executing platform.

This rework has side-effects on the way device MMIO regions are mapped. To address these requirements, this PR performs the following changes:
* Introduce early init boot stage in `kvm/arm64`.
* Add `early_init` to drivers that use MMIO, and have these drivers register bootinfo mrds for these regions during `early_init`.
* Add runtime mapping of MMIO regions to the initialization of relevant drivers.

All the above are described in further detail in the sections that follow.

Additional fixes include:
* lib/ukbus: Use the correct init class and priority.
* plat/drivers/uktty: Clean up driver init (`ns16550` and `pl011`).
* plat/drivers/uktty: Move driver init to `uk_inttab` (`ns16550` and `pl011`).
* plat/kvm/arm64: Remove console registration from platform setup.

#### Rework of paged memory init

Under the new scheme, bootinfo is simplified to only contain valid memory regions. Upon paged memory initialization these regions are either mapped with the protections defined in their mrd, or are already mapped, in which case only their protections are updated. An exception to this is mrds of the `UKPLAT_MEMRT_DEVICE` type, that are ignored to allow MMIO regions mapped during early device initialization to remain mapped.
    
Any memory not defined in bootinfo is unmapped. The theoretical bounds used as a reference are platform-specific and are defined in a newly introduced `platform.h` that must be provided by all platforms.

#### Early device initialization

`early_init` is invoked as the last call of early boot code, as soon as the stack and MMU are enabled.
    
Early devices are initialized before bootinfo setup, as the latter depends on UART for debug messages.

Early devices can use the APIs provided by libfdt / ACPI to obtain any information required from the firmware, such as device regions and the kernel command line.
    
Drivers that register with `early_init` are updated to append mrds of their MMIO regions to `bootinfo`. That is required so that their regions are not unmapped by paged memory init. The mrds added during `early_init` to bootinfo must use the newly introduced `UKPLAT_MEMRT_DEVICE` type.

Be aware that early drivers should not call `ukplat_bootinfo_coalesce`, as mrd coalescing is performed once at the end of `bootinfo_setup`, which is now performed as to the last step of `early_init`.

Notice that:
* `early_init` is currently introduced in `kvm/arm64`, as it is the only platform that involves early initialization of drivers that use MMIO. 
* `early_init` is currently implemented as a simple placeholder from where drivers can add manual calls of their early_init functions. This should be replaced with an early inittab as soon as bootinfo is moved out of `plat`.

#### Late (i.e. non-early) device initialization

Given that all memory that is not backed by an mrd is unmapped during paged memory init, devices that don't register mrds during early init are now required to map their regions during probing. To address this requirement this PR updates relevant drivers to map their MMIO regions dynamically.

#### References

[1] Realm Management Monitor Specification DEN0137 Sect. D2.1
[2] #964

GitHub-Depends-On: #972 
GitHub-Depends-On: #1212
GitHub-Depends-On: #1346


michpappas:
As pointed out by @marcrittinghaus during a private discussion, the proposed way of mapping regions has a potential flaw if a region is partially mapped, as that would cause `ukplat_page_map()` to return `EEXIST`, which will in turn cause the subsequent `ukplat_page_set_attr()` to fail.

The implementation of paged memory init is now reworked to initialize a new page table instead (courtesy of @mogasergiu). Besides being simpler, this approach additionally avoids the expensive TLB invalidations associated with unmap.

drivers/ukbus/platform/platform_bus.c
```diff
@@ -259,17 +260,25 @@ __vaddr_t uk_bus_pf_devmap(__paddr_t base, __sz size)
...
-	if (!rc)
-		goto out;
+	for (__sz i = 0; i < pages; i++) {
+		va = vaddr + i * PAGE_SIZE;
+		pa = paddr + i * PAGE_SIZE;
+		rc = ukplat_page_map(pt, va, pa, 1, attr, 0);
+		if (!rc)
+			continue;
+
+		if (rc == -EEXIST)
```
mogasergiu, line 275: I think we should keep the before behavior and crash instead. Something must be inconsistent in the memory map if we end up with an overlap, right?

drivers/uktty/pl011/pl011.c
```diff
@@ -65,29 +65,27 @@
...
+static __u64 pl011_base;
 #endif /* !CONFIG_LIBUKTTY_PL011_EARLY_CONSOLE_BASE */
 
 /* Macros to access PL011 Registers with base address */
-#define PL011_REG(r)		((__u16 *)(pl011_uart_bas + (r)))
+#define PL011_REG(r)		((__u16 *)(pl011_base + (r)))
 #define PL011_REG_READ(r)	ioreg_read16(PL011_REG(r))
 #define PL011_REG_WRITE(r, v)	ioreg_write16(PL011_REG(r), v)
 
-static void init_pl011(__u64 bas)
```
mogasergiu, line 87: `bas`ed :moyai:

SerbanSo (commented):
The changes for the MRDs affecting kvm-x86 look good from my point of view.
However, I don't have enough code context for the changes to the drivers and the other architectures.

michpappas:
Rebased from `staging`, no functional changes.

plat/common/memory.c
```diff
@@ -64,17 +68,12 @@ struct uk_alloc *ukplat_memallocator_get(void)
 void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 {
 	struct ukplat_memregion_desc *mrd, alloc_mrd = {0};
-	__vaddr_t unmap_start, unmap_end;
-	__sz unmap_len, desired_sz;
+	__sz desired_sz;
```
mogasergiu: I thought you liked reverse xmas :(

plat/kvm/arm/early_init.c
```diff
@@ -0,0 +1,25 @@
...
+/* KVM early init
+ *
+ * Initialize early devices and coalesce mrds.
+ * TODO Replace with init-based callback registration.
+ */
+void ukplat_early_init(void)
+{
+	struct ukplat_bootinfo *bi;
+
+	bi = ukplat_bootinfo_get();
```
mogasergiu, line 41: Why no `bi` check here?

plat/kvm/arm/setup.c
```diff
@@ -155,13 +155,13 @@ static void __noreturn _ukplat_entry2(void)
...
-void __no_pauth _ukplat_entry(struct ukplat_bootinfo *bi)
+void __no_pauth _ukplat_entry(void)
 {
+	struct ukplat_bootinfo *bi;
 	void *bstack;
-	void *fdt;
 	int rc;
 
-	fdt = (void *)bi->dtb;
+	bi = ukplat_bootinfo_get();
```
mogasergiu: Ditto?

lib/ukbus/include/uk/bus.h
```diff
@@ -42,9 +42,13 @@
 extern "C" {
 #endif
 
-#define UK_BUS_INIT_CLASS UK_INIT_CLASS_PLAT
-#define UK_BUS_INIT_PRIO  0
-#define UK_BUS_REGISTER_PRIO 0
+#define UK_BUS_INIT_CLASS UK_INIT_CLASS_SYS
```
mogasergiu: Do we know why it had PLAT prio before?

drivers/uktty/pl011/pl011.c
```diff
@@ -105,39 +103,63 @@ static void init_pl011(__u64 bas)
 
 	/* Just enable UART and data transmit/receive */
 	PL011_REG_WRITE(REG_UARTCR_OFFSET, CR_TXE | CR_UARTEN);
+
+	return 0;
 }
 
-void pl011_console_init(const void *dtb)
+int pl011_console_init(void *dtb)
```
mogasergiu: Why is the `const` removed?

drivers/uktty/pl011/exportsyms.uk
```diff
@@ -1,3 +1,4 @@
+pl011_early_init
```
mogasergiu, line 1: Commit message of
> drivers/uktty/pl011: Add early init

Add early device init. When paging is enabled, this adds an mrd for
the pl011 MMIO region to bootinfo so that the region is **mapped after paged memory init**.

drivers/uktty/ns16550/ns16550.c
```diff
@@ -126,35 +123,49 @@ static void init_ns16550(__u64 base)
 	ns16550_reg_write(NS16550_FCR_OFFSET,
 			  ns16550_reg_read(NS16550_FCR_OFFSET) &
 					 ~(NS16550_FCR_FIFO_EN));
+	return 0;
 }
 
-void ns16550_console_init(const void *dtb)
+int ns16550_console_init(void *dtb)
```
mogasergiu: Ditto const?

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -608,6 +613,29 @@ static int gicv2_do_probe(void)
...
 
+#if CONFIG_PAGING
+static int gicv2_map(void)
+{
+	__vaddr_t vbase;
+
+	vbase = uk_bus_pf_devmap(gicv2_drv.dist_mem_addr,
+				 gicv2_drv.dist_mem_size);
+	if (unlikely(PTRISERR(vbase)))
+		return PTR2ERR(vbase);
```
mogasergiu, line 625: No error messages?

drivers/ukintctlr/gic/gic-v2.c
```diff
@@ -641,6 +669,12 @@ int gicv2_probe(struct _gic_dev **dev)
 		return rc;
 	}
 
+#if CONFIG_PAGING
+	rc = gicv2_map();
+	if (unlikely(rc))
+		return rc;
```
mogasergiu, line 680: Ditto

drivers/ukintctlr/gic/gic-v3.c
```diff
@@ -692,6 +697,29 @@ static int gicv3_do_probe(void)
 }
 #endif /* !CONFIG_UKPLAT_ACPI */
 
+#if CONFIG_PAGING
+static int gicv3_map(void)
```
mogasergiu, line 701: Ditto for GICv3

drivers/ukrtc/pl031/pl031.c
```diff
@@ -180,6 +184,12 @@ int pl031_init_rtc(void *dtb)
 	}
 	uk_pr_info("Found RTC at: 0x%lx\n", pl031_base_addr);
 
+#if CONFIG_PAGING
+	pl031_base_addr = uk_bus_pf_devmap(pl031_base_addr, size);
+	if (unlikely(PTRISERR(pl031_base_addr)))
+		return PTR2ERR(pl031_base_addr);
```
mogasergiu, line 190: Ditto?

drivers/uktty/ns16550/ns16550.c
```diff
@@ -126,35 +128,100 @@ static void init_ns16550(__u64 base)
...
+	struct ukplat_bootinfo *bi;
+	int rc;
+
+	ns16550_uart_base = CONFIG_LIBUKTTY_NS16550_EARLY_CONSOLE_BASE;
+
+	/* From this point we can print */
+	ns16550_uart_initialized = 1;
+
+	bi = ukplat_bootinfo_get();
+	if (unlikely(!bi)) {
```
mogasergiu: Hmm, should probably just have an assertion here, as the getter already checks stuff.

drivers/uktty/pl011/pl011.c
```diff
@@ -105,39 +104,109 @@ static void init_pl011(__u64 bas)
...
 	/* Just enable UART and data transmit/receive */
 	PL011_REG_WRITE(REG_UARTCR_OFFSET, CR_TXE | CR_UARTEN);
+
+	return 0;
+}
+
+int pl011_early_init(void)
+{
+#if CONFIG_LIBUKTTY_PL011_EARLY_CONSOLE
+	struct ukplat_bootinfo *bi = ukplat_bootinfo_get();
```
mogasergiu, line 114: assert `bi`?

plat/common/bootinfo_fdt.c
```diff
@@ -198,14 +198,22 @@ void ukplat_bootinfo_fdt_setup(void *fdtp)
...
+ * `ukplat_memregion_alloc()` which would be unsafe to do so before
+ * knowing that the memory region descriptor list has been coalesced
+ * and sorted.
+ */
+void ukplat_bootinfo_fdt_post_setup(void)
+{
+	struct ukplat_bootinfo *bi;
+
+	bi = ukplat_bootinfo_get();
+	if (unlikely(!bi))
```
mogasergiu, line 215: Ditto

plat/common/memory.c
```diff
@@ -133,7 +120,7 @@ void *ukplat_memregion_alloc(__sz size, int type, __u16 flags)
 		alloc_mrd.len = desired_sz;
 		alloc_mrd.pg_count = PAGE_COUNT(desired_sz);
 		alloc_mrd.type = type;
-		alloc_mrd.flags = flags | UKPLAT_MEMRF_MAP;
+		alloc_mrd.flags = flags;
 
 		bi = ukplat_bootinfo_get();
```
mogasergiu, line 125: I know I wrote this :laughing: but ditto, this is a crash reason tbh.

mogasergiu (changes requested):
Hey, LGTM :+1:. Couple of comments. I will then proceed to have an additional look and test on x86 UEFI. Thanks.

plat/common/memory.c
michpappas: Ooops thanks

lib/ukbus/include/uk/bus.h
michpappas: Good point, I assumed that this was since the platform bus was introduced, but now you mention I had a closer look and see that the change was added in 3092e3ccd1a62406af2bd49d47496a0f9e2581de with some justification on why it cannot be `CLASS_EARLY`, but not much justification on why it's set to `CLASS_PLAT`. My understanding that devices are initialized at `CLASS_SYS` is because of the relevant comment in https://github.com/unikraft/unikraft/blob/staging/include/uk/init.h#L125:
>  Define a stage for device initialization

HOWEVER: I now notice that `CLASS_SYS` states https://github.com/unikraft/unikraft/blob/staging/include/uk/init.h#L104:

> Define a stage for platform initialization. Platform at this point read
> all the device and device are initialized.

What I propose is that I revert this change and move also the initialization of late UART from SYS to PLAT. And let's clarify the scope of these two init levels with @skuenzer when we re-arch ukboot.

Thanks for catching that!

drivers/uktty/pl011/exportsyms.uk
michpappas: Currently the commit message states:
> region is not unmapped during paged memory init.

That is actually accurate. MMIO regions of early devices are already mapped, and the purpose of adding an mrd to bootinfo is so that the device does not get unmapped during paged memory init (i.e. in the mapping makes it to the new pagetable and the region stays mapped).

drivers/ukbus/platform/platform_bus.c
michpappas: So this is a fix regarding correctness as described in the commit message. Of course in pracrtice if that actually happens on device regions something has gone very wrong but even then I would leave handling to the caller instead. Either way that function is temporary, once we rework ukboot we should use be replacing this with `uk_vma_dma_map()`.

drivers/uktty/ns16550/ns16550.c
michpappas: ~~Generally speaking I'd rather not make assumptions on a function's implementation and rather rely on the interface, given that if the implementation changes all calling code breaks. Besides, from a quick look I can see several places where we check the result of `ukplat_bootinfo_get()` rather than asserting.~~

Actually nevermind, although the above is conceptually correct an assertion seems more appropriate in this context 👍🏼

plat/common/bootinfo_fdt.c
michpappas: I follow the pattern used in the rest of this file by the original author:
```
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 187) void ukplat_bootinfo_fdt_setup(void *fdtp)
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 188) {
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 189)      struct ukplat_bootinfo *bi;
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 190) 
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 191)      bi = ukplat_bootinfo_get();
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 192)      if (unlikely(!bi))
1afb423425 (Sergiu Moga     2023-05-14 18:21:55 +0300 193)              ukplat_bootinfo_crash("Invalid bootinfo");
```

plat/common/memory.c
michpappas: In principle I agree, but taking into account earlier comments on the same topic, and the number places in the code base where we actually check the return value of `ukplat_bootinfo_get()`, I would rather if we came up with a common approach that is also expressed on the interface of this function, rather than ad-hoc requests on reviews. Either way, I would leave this as is given that it was not changed in this PR and the error is propagated to the caller anyway.

drivers/uktty/pl011/pl011.c
michpappas: Honestly I don't remember, but the interface of this function is updated on a later commit anyway. If you think it's necessary I can revert this change though.

drivers/uktty/ns16550/ns16550.c
michpappas: See https://github.com/unikraft/unikraft/pull/1373#discussion_r1616815507

drivers/uktty/pl011/pl011.c
mogasergiu: Meh, if it's overwritten anyway I would say don't bother :shrimp: :fried_shrimp: :man_shrugging:

drivers/uktty/ns16550/ns16550.c
mogasergiu: I feel like the problem here may also lie in the fact that it's kinda impossible for this getter to fail (or is it?). Maybe a dtb getter should exist and be used instead. Anyway no matter, too much of a trivial thing that will get overwritten by rearch/refactor anyway. I will let you choose whatever feels more sane here.

plat/common/bootinfo_fdt.c
mogasergiu: The **unbelievably handsome and charismatic** author may have just changed his mind :thinking:, I am joking (about the latter). In all seriousness though, I believe these changes were back when I barely had any familiarity with this code I guess. I am fine with leaving as is and coming back when we rearch ukboot.

plat/common/memory.c
mogasergiu: > The **unbelievably handsome and charismatic** author may have just changed his mind 🤔, I am joking (about the latter). In all seriousness though, I believe these changes were back when I barely had any familiarity with this code I guess. I am fine with leaving as is and coming back when we rearch ukboot.

Ditto. I am fine with leaving for later.

michpappas:
@mogasergiu I addressed all review comments with a thumbs-up.

For the `ukplat_bootinfo_get()` I agree in principle, but in that case we should consider updating the interface to reflect that this function should always succeed and as such any failures should be caught via assertions.

I also tested the latest state with:
* `arm64/qemu`: `app-helloworld` with `lxboot`, `virt`, and `UEFI` OK ✅
* `arm64/fircracker`: ~~compile only, may fire a baremetal instance for a runtime test later~~ OK ✅
* `x86_64/firecracker` with `app-helloworld` OK ✅
* `x86_64/qemu`: `app-helloworld` with lxboot OK ✅
* `x86_64/qemu`: `app-helloworld` with EFI gives me an error: `qemu-system-x86_64: combined size of system firmware exceeds 8388608 bytes` but I'm not very familiar with `qemy-system-x86_64` so it may be that I'm doing something wrong.

mogasergiu (approved):
Me gusta. Tested on x86 UEFI as well.

michpappas:
Pushed previously missed revert for https://github.com/unikraft/unikraft/pull/1373#discussion_r1616084891.

michpappas:
Rebased from `unikraft/staging`

SerbanSo (approved):
Looks good to me.

Reviewed-by: Serban Sorohan <serban.sorohan@gmail.com>

michpappas:
Rebased from `staging` and resolved conflict with #1434

razvand (approved):

PR #1375, opened 2024-04-03
lib/posix-timerfd: Fix update thread double free and add update thread dtor

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Whenever a timer fd is released, the thread associated with its update
function is also released through a combination of `uk_thread_terminate`
and `uk_thread_release`. However, if the released thread is different
from the current thread then `uk_thread_terminate` also calls
`uk_thread_release` on its own, leading to a double free. This is always
the case as the release function is impossible to be called from the
update function.

Therefore, fix this by deleting the explicit call to `uk_thread_release`
and rely on the one implicitly called by `uk_thread_terminate`.

On the other hand, with future introduction of posix-signals, we will
have to compensate for the fact that some subsystems end up terminating
threads without being timerfd aware.

For example, posix-process may try to terminate all threads of a
given process without being aware that a thread may have open fd's.
In this case, a closed thread with an open timerfd results in the
timerfd still holding a reference to the released thread to terminate
it when it is closed, resulting in a double termination or memory
corruption.

Fix this by adding a destructor that, when called, uses the `priv` field
of `struct uk_thread`, initially assigned to the timerfd structure
holding this reference, to mark this reference as NULL so that double
termination is avoided.

### DIFF
--- a/lib/posix-timerfd/timerfd.c
+++ b/lib/posix-timerfd/timerfd.c
@@ -176,8 +176,6 @@ static void timerfd_release(const struct uk_file *f, int what)
 		uk_file_rlock(f);
 		uk_thread_terminate(d->upthread);
 		uk_file_runlock(f);
-		/* Collect thread */
-		uk_thread_release(d->upthread);
 	}
 	if (what & UK_FILE_RELEASE_OBJ) {
 		struct timerfd_alloc *al;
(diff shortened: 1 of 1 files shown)


skuenzer (approved):
Thanks for this fix!

PR #1377, opened 2024-04-04
lib/posix-process: Fix off-by-one error in tid check

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Fix an off-by-one error in tid2pthread() that returns failure when the passed tid is the maximum allowed value.
    
Show error message when CONFIG_LIBPOSIX_PROCESS_MAX_PID is reached.

### DIFF
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -110,6 +110,8 @@ static inline pid_t find_free_tid(void)
 	}
 	if (found == TIDMAP_SIZE) {
 		/* no free PID */
+		uk_pr_err("Could not allocate TID: Out of TIDs (configured max: %d)\n",
+			  CONFIG_LIBPOSIX_PROCESS_MAX_PID);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


DeliaPavel (approved):
Looks good to me.

lib/posix-process/process.c
```diff
@@ -110,6 +110,8 @@ static inline pid_t find_free_tid(void)
 	}
 	if (found == TIDMAP_SIZE) {
 		/* no free PID */
+		uk_pr_err("Could not allocate TID: Out of TIDs (configured max: %d)\n",
```
RaduNichita, line 113: Not much related, but I think we may move the second if that is checking `if (found == TIDMAP_SIZE)` inside the first one.

If we have found something with the first search, there is no need to check the condition `if (found == TIDMAP_SIZE)` two times.

RaduNichita (commented):
It looks good to me, I left some nit. I am not sure if we should also include this one in the current PR or create another one.

lib/posix-process/process.c
michpappas: Hi @RaduNichita, yes, I added that message here while preserving the existing logic, but I think you're right. `git blame` shows that this pattern originates since inception of `process.c`, and until a1c73be910 both conditionals checked against `CONFIG_LIBPOSIX_PROCESS_MAX_PID`. Before making any changes I would like to check on the reasoning with @skuenzer who is the original author, but given that he's away this week I would prefer if we considered that change on a follow-up PR.

RaduNichita: Agree.

RaduNichita (approved):
Let's merge this and discuss with @skuenzer based on our discussion. Thanks @michpappas.

razvand (approved):

PR #1316, opened 2024-02-07
{arch,plat}/x86: Refactor FS_BASE/GS_BASE operations

mogasergiu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): N/A
 - Application(s):N/A

### Description of changes

Depending on the configured CPU, `__FSGSBASE__` and its related built-ins may be defined. Use them when possible/available as it may improve performance when trying to read/write from such registers.

PR #1386, opened 2024-04-16
lib/posix-process: Add vfork & execve

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

* CONFIG_LIBPOSIX_PROCESS_EXECVE: Enables the `execve()` syscall
* CONFIG_LIBPOSIX_PROCESS_VFORK: Enables the `vfork()` syscall
* CONFIG_LIBUKBINFMT: Enables the `ukbinfmt` library

### Description of changes

This PR adds support for `vfork()` / `execve()` to libposix-process. This provides the foundation for running multiprocess applications on Unikraft.

`vfork()` creates a child process that shares the same memory with the calling thread, including the stack. Upon `vfork()` the parent is suspended until the child calls `execve()` or `_exit()` to prevent the parent's state from being corrupted. Due to above, the only permissible actions taken by the child are:
* Update a variable of type `pid_t` to store its return value.
* Calling any function other than one of the functions of the `exec(2)` family, or `_exit(2)`.

Any action besides the above results into undefined behavior.

For a more elaborate description on `vfork()` see [vfork(2)](https://man7.org/linux/man-pages/man2/vfork.2.html).

`execve()` implements the standard behavior described in [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html). To provide flexibility on the supported binary formats the implementation relies on `libukbinfmt`, which is also introducedin this series. Similarly to linux, this provides a minimal framework for registering loaders that handle different binary formats, introducing the ability of loading interpreted scripts. The implementation may be extended with support for linux's [`binfmt_misc`](https://docs.kernel.org/admin-guide/binfmt-misc.html) fs, should such a use-case arise. The implementation of the ELF `ukbinfmt` loader is introduced in `app-elfloader`.

GitHub-Depends-On: #1316
GitHub-Depends-On: #1322
GitHub-Depends-On: #1346
GitHub-Depends-On: https://github.com/unikraft/app-elfloader/pull/79


lib/Makefile.uk
```diff
@@ -70,3 +70,4 @@ $(eval $(call import_lib,$(CONFIG_UK_BASE)/lib/ukrust))
 $(eval $(call import_lib,$(CONFIG_UK_BASE)/lib/ukreloc))
 $(eval $(call import_lib,$(CONFIG_UK_BASE)/lib/ukofw))
 $(eval $(call import_lib,$(CONFIG_UK_BASE)/lib/ukallocstack))
+$(eval $(call import_lib,$(CONFIG_UK_BASE)/lib/ukbinfmt))
```
mogasergiu, line 75: > lib: Introduce libukbinfmt

s/sha-bang/shebang in the commit msg

lib/ukbinfmt/include/uk/binfmt.h
```diff
@@ -0,0 +1,102 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/posix-process/process.h
```diff
@@ -58,6 +67,7 @@ struct posix_thread {
 	struct uk_list_head thread_list_entry;
 	struct uk_thread *thread;
 	struct uk_alloc *_a;
+	enum posix_thread_state state;
```
mogasergiu, line 85: I would also add a comment here that states again the potential state difference between uk_thread and posix_thread

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,29 @@
...
+
+void execve_arch_execenv_init(struct ukarch_execenv *execenv,
+			      __uptr ip, __uptr sp)
+{
+	execenv->regs.rip = ip;
+	execenv->regs.rsp = sp;
+
+	/* Prepare for iret */
+	execenv->regs.eflags = X86_EFLAGS_IF; /* enable IRQs on ctx switch */
+	execenv->regs.cs = 8;
```
mogasergiu: `GDT_DESC_OFFSET(GDT_DESC_CODE)` and `GDT_DESC_OFFSET(GDT_DESC_DATA)`. yes this introduces deps to plat common :zany_face:

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,29 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/arch/ctx.h>
+#include <uk/essentials.h>
+
+void execve_arch_execenv_init(struct ukarch_execenv *execenv,
```
mogasergiu: Why no asserts! :angry:

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/arch/ctx.h>
+#include <uk/essentials.h>
+
+void execve_arch_execenv_init(struct ukarch_execenv *execenv,
```
mogasergiu: Ditto :relieved:

lib/posix-process/execve.c
```diff
@@ -0,0 +1,194 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024 :clock10:

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,20 @@
...
+			      __uptr ip, __uptr sp)
+{
+	execenv->regs.lr = ip;
```
mogasergiu: I think you may want to use `elr_el1` here instead.

lib/posix-process/execve.c
```diff
@@ -0,0 +1,194 @@
...
+	struct uk_thread *this_thread;
+	void *stack_old;
+
+	execenv_new = (struct ukarch_execenv *)arg0;
+	stack_old = (void *)arg1;
+
+	this_thread = uk_thread_current();
+	UK_ASSERT(this_thread);
+
+	pthread = tid2pthread(ukthread2tid(this_thread)); /* FIXME */
```
mogasergiu, line 39: Should we assert `pthread` too? :thinking: Or we assume it's fine based on the above assertion?

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,20 @@
...
+	execenv->regs.sp = sp;
+
```
mogasergiu, line 30: I think we should also set `spsr_el1` as I believe it to have an unsafe value coming from a `malloc`d stack. I would reuse the one of the parent and, additionally ` &= ~PSR_I` so that IRQs are enblaed. Ditto for x86, let's reuse the parents flags but with IRQs forcefully enabled (although I think in both cases it should be safe to assume they are already enabled IMO).

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,29 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: :two: :zero: :two: :four:

lib/posix-process/clone.c
```diff
@@ -310,19 +310,19 @@ static void _clone_child_gc(struct uk_thread *t)
...
  *       The reason is that Unikraft libraries place TLS variables and use
  *       the TLS effectively as TCB.
- *       In case no TLS is handed over (CLONE_SETTLS is not set), _clone will
+ *       In case no TLS is handed over (CLONE_SETTLS is not set), uk_clone will
  *       still allocate an Unikraft TLS but sets the TLS architecture pointer
  *       to zero.
  */
-static int _clone(struct clone_args *cl_args, size_t cl_args_len,
-		  struct ukarch_execenv *execenv)
+int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
```
mogasergiu, line 324: `__sz`?

lib/posix-process/clone.c
```diff
@@ -388,6 +388,25 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
+	* space as the parent.
+	*
+	* Assign the parent's stack. We assign the user TLS to the
+	* parent's in clone_setup_arch_sysregs(). We copy the registers
+	* later below.
+	*/
+	if ((flags & CLONE_VM)) {
+		if (!cl_args->stack && !cl_args->stack_size) {
+			uk_pr_debug("Using parent's sp @ 0x%lx\n", execenv->regs.rsp);
+			cl_args->stack = execenv->regs.rsp; /* FIXME */
```
mogasergiu: I assume FIXME cus x86 specific, right?

lib/posix-process/clone.c
```diff
@@ -476,10 +520,21 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
 	/* Assign the child to the scheduler */
 	uk_sched_thread_add(s, child);
 
+	/* CLONE_VFORK: Block the parent until the child calls execve()
+	 * or exit(). Yield to schedule the child.
+	 */
+	if (flags & CLONE_VFORK) {
+		struct posix_thread *pthread = tid2pthread(ukthread2tid(t));
```
mogasergiu: Subscope declaration? So unlike you :disappointed:

mogasergiu (changes requested):
Hi there :wave: :man_health_worker: . Had a brief go at this. I will have a deeper look once we get the dependency PRs in.

lib/Makefile.uk
michpappas: "sha-bang" is also valid, see https://en.wikipedia.org/wiki/Shebang_(Unix)

lib/posix-process/arch/x86_64/execve.c
michpappas: I'm not sure we want to introduce such a dependency though. I'll add a `FIXME`.

lib/posix-process/clone.c
michpappas: Had I written this from scratch yes, but since the rest of this file uses `size_t` I chose to stick to the existing convention.

lib/posix-process/execve.c
michpappas: I can add an assertion

lib/posix-process/arch/arm64/execve.c
michpappas: Oh yes

lib/posix-process/arch/arm64/execve.c
michpappas: Actually, what is the reasoning for updating `SPSR_EL1` and `ESR_EL1` in `ukarch_execenv_load`?

lib/posix-process/arch/x86_64/execve.c
mogasergiu: ~~Can you also add a comment like a TODO or smth so that someone in the future may do this replacement once we get rid of this `plat/common` stuff?~~
Nvm, yes this is what you meant, I somehow skipped the last sentence :man_facepalming:

lib/Makefile.uk
mogasergiu: Ok fine... but I think you can agree with me that we all would much rather have a "she" than a "sha" :innocent: :stuck_out_tongue:

lib/posix-process/clone.c
michpappas: Correct. I now included your patch with `ukarch_regs_get_sp()` to resolve this.

lib/posix-process/arch/arm64/execve.c
mogasergiu: Following an offline discussion, we decided that we should for now keep these semantics for both `ESR_EL1` and `SPSR_EL1`.

So, initially, the ARM code shared the entries for both IRQ's and regular traps. Since the former would not require `ESR_EL1` and potentially creating confusion we would normally restore `ESR_EL1` to make sure that IRQ code does not somehow think it is in a trap because `ESR_EL1` has a value that a trap would normally set. But I will soon submit a small patch that separates the two entries, thus restoring `ESR_EL1` is no longer required.

On the other hand, we still have to keep `SPSR_EL1` updated as that restores the `PSTATE`. For example, consider a page fault that may happen during a syscall. This is essentially a nested trap so `SPSR_EL1` must remain consistent when returning from the page fault as that could mean confusing the previous trap. Same thing if we nest page faults, which we should soon have a use case for.

So, for now, it is not clear if it is really worth dropping `ESR_EL1`, as we cannot see the use of restoring it anymore but, at the same time, it ensures a little bit of consistency on return from exceptions, although we are not sure if it is a useful consistency to have either, it will probably save us in the future. From what I have seen, neither linux not freebsd restore it on exception return but, I do not have knowledge of how they deal with nested pagefaults, or if they do consider them at all.

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -223,6 +223,11 @@ struct __regs {
 
 UK_CTASSERT(sizeof(struct __regs) == __REGS_SIZEOF);
 
+static inline __uptr ukarch_regs_get_sp(struct __regs *r)
```
mogasergiu: Don't forget to add your sob + cab.

lib/ukbinfmt/binfmt.c
```diff
@@ -0,0 +1,100 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/ukbinfmt/binfmt.c
```diff
@@ -0,0 +1,100 @@
...
+
+#define uk_binfmt_loader_foreach(_loader, _type)			     \
+	uk_list_for_each_entry((_loader), &uk_binfmt_loader_list, list_head) \
+		if ((_loader)->type == (_type))
+
+int uk_binfmt_register(struct uk_binfmt_loader *loader)
+{
+	UK_ASSERT(loader);
+	UK_ASSERT(loader->name);
+	UK_ASSERT(loader->ops.load);
```
mogasergiu, line 23: Is it mandatory that both ops are implemented and that it has a name? I assume this is because `uk_binfmt_load` does not check for such cases?

lib/ukbinfmt/binfmt.c
```diff
@@ -0,0 +1,100 @@
...
+	UK_ASSERT(!args->loader);
+
+	/* Try INTERP types first. These should update args and
+	 * fall back to EXEC types.
+	 */
+	uk_binfmt_loader_foreach(ldr, UK_BINFMT_LOADER_TYPE_INTERP) {
+		UK_ASSERT(ldr->ops.load);
+		rc = ldr->ops.load(args);
+		if (unlikely(rc < 0))
+			return rc;
```
mogasergiu, line 54: Perhaps an err msg?

lib/ukbinfmt/binfmt.c
```diff
@@ -0,0 +1,100 @@
...
+			UK_ASSERT(!args->user);
+			break; /* Assume one interpreter max */
+		}
+		if (rc == UK_BINFMT_HANDLED)
+			return 0;
+	}
+
+	uk_binfmt_loader_foreach(ldr, UK_BINFMT_LOADER_TYPE_EXEC) {
+		rc = ldr->ops.load(args);
+		if (unlikely(rc < 0))
```
mogasergiu: ditto

lib/ukbinfmt/binfmt.c
```diff
@@ -0,0 +1,100 @@
...
+int uk_binfmt_unload(struct uk_binfmt_loader_args *args)
+{
+	int rc;
+
+	UK_ASSERT(args);
+	UK_ASSERT(args->loader);
+
+	uk_pr_debug("Unload handled by %s\n", args->loader->name);
+	rc = args->loader->ops.unload(args);
+	if (unlikely(rc < 0))
```
mogasergiu: ditto

lib/ukbinfmt/include/uk/binfmt.h
```diff
@@ -0,0 +1,102 @@
...
+
+enum uk_binfmt_type {
+	UK_BINFMT_LOADER_TYPE_NONE,
+	UK_BINFMT_LOADER_TYPE_INTERP,
+	UK_BINFMT_LOADER_TYPE_EXEC,
+};
+
+struct uk_binfmt_loader_args {
+	const char *pathname; /* fully qualified */
+	char *progname;
```
mogasergiu, line 32: Would it be worth it documenting each field? Like a 1-line comment, since they are public interfaces.

lib/ukbinfmt/include/uk/binfmt.h
```diff
@@ -0,0 +1,102 @@
...
+/* Loader unload function.
+ *
+ * @param args Loader args.
+ * @return zero on success or negative value on error.
+ */
+typedef	int (*uk_binfmt_unload_fn)(struct uk_binfmt_loader_args *args);
+
+struct uk_binfmt_loader {
+	const char *name;
+	enum uk_binfmt_type type;
```
mogasergiu, line 69: Ditto

lib/posix-process/Makefile.uk
```diff
@@ -24,7 +27,7 @@ LIBPOSIX_PROCESS_ASINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 LIBPOSIX_PROCESS_CXXINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 
 UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS_CLONE) += clone-5e
-UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS) += execve-3
+UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS_EXECVE) += execve-3
```
mogasergiu: `execve-3e`? :thinking:

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,34 @@
...
+void execve_arch_execenv_init(struct ukarch_execenv *execenv_new,
+			      struct ukarch_execenv *execenv,
+			      __uptr ip, __uptr sp)
+{
+	UK_ASSERT(execenv_new);
+	UK_ASSERT(execenv);
+	UK_ASSERT(ip);
+	UK_ASSERT(sp);
+
```
mogasergiu, line 19: Perhaps an assertion for sp alignment?

lib/posix-process/arch/arm64/execve.c
```diff
@@ -0,0 +1,34 @@
...
+	/* Copy SPSR to preserve the application's state at
+	 * syscall time.
+	 */
+	execenv_new->regs.spsr_el1 = execenv->regs.spsr_el1;
+
+	/* Copy ESR to make sure we restore a sane value */
+	execenv_new->regs.esr_el1 = execenv->regs.esr_el1;
+
+	/* Also copy the current sysregs to avoid ending up with undefined
```
mogasergiu: s/sysregs/sysctx

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,37 @@
...
+{
+	UK_ASSERT(execenv_new);
+	UK_ASSERT(execenv);
+	UK_ASSERT(ip);
+	UK_ASSERT(sp);
+
+	execenv_new->regs.rip = ip;
+	execenv_new->regs.rsp = sp;
+
+	/* Prepare for iret
```
mogasergiu: Nitpick: let's be specific: `iretq`. As there is also `iret` and `iretd` whose semantics can get very different. Just to avoid confusion of future readers.

lib/posix-process/execve.c
```diff
@@ -0,0 +1,197 @@
...
+	UK_ASSERT(pthread);
+
+	pthread_parent = pthread->parent;
+
+	/* If there is no parent, then must be calling execve()
+	 * from init without having called vfork().
+	 */
+	if (!pthread_parent) {
+		UK_ASSERT(pthread->process->pid == 0); /* FIXME */
```
mogasergiu: What is this FIXME about?

lib/posix-process/clone.c
```diff
@@ -388,6 +389,27 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
 #endif /* UK_DEBUG */
 
+	/* CLONE_VM requires that the child operates on the same memory
+	 * space as the parent.
+	 *
+	 * Assign the parent's stack. We assign the user TLS to the
+	 * parent's in clone_setup_arch_sysregs(). We copy the registers
+	 * later below.
+	 */
+	if ((flags & CLONE_VM)) {
```
mogasergiu: Why double paranthesis?

lib/posix-process/clone.c
```diff
@@ -388,6 +389,27 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
+		if (!cl_args->stack && !cl_args->stack_size) {
+			cl_args->stack = ukarch_regs_get_sp(&execenv->regs);
```
mogasergiu: This is probably just my hunch, but it feels uncomfortable modifying the passed argument here (ditto for tls). Can we instead do this assignment on a locally defined variable instead? Or is there a reason behind this.

lib/posix-process/clone.c
```diff
@@ -446,16 +468,46 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
+		child->_mem.stack_a = t->_mem.stack_a;
+	} else  {
+		/* CLONE_SETTLS: Instead of just activating the Unikraft TLS
+		 * we activate the passed TLS pointer as soon as the child
+		 * wakes up.
+		 * NOTE: If SETTLS is not set, we do not activate any TLS
+		 * although a Unikraft TLS was allocated.
+		 */
+		if ((flags & CLONE_SETTLS) || (flags & CLONE_VM))
+			child->tlsp = cl_args->tls;
```
mogasergiu: And here we reuse that cl_args argument.

lib/posix-process/clone.c
```diff
@@ -446,16 +468,46 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
...
+		uk_pr_debug("Creating new process for child\n");
+		ret = uk_posix_process_create(uk_alloc_get_default(), child, t);
+		if (unlikely(ret)) {
+			uk_pr_err("Could not create child process\n");
+			goto err_free_child;
+		}
+	}
+
 	/* Call clone handler table but treat CLONE_SETTLS as handled */
 	ret = _uk_posix_clonetab_init(cl_args, cl_args_len,
```
mogasergiu, line 521: Same here. Does clonetab rely on being preceeded by such operations that would adjust the original user `cl_args`?

lib/posix-process/clone.c
```diff
@@ -476,10 +528,21 @@ static int _clone(struct clone_args *cl_args, size_t cl_args_len,
 	/* Assign the child to the scheduler */
 	uk_sched_thread_add(s, child);
 
+	/* CLONE_VFORK: Block the parent until the child calls execve()
```
mogasergiu, line 540: I am not sure this is related to this PR but I have just noticed that `ukthread2tid` is called before `uk_sched_thread_add`. It feels awkward. Do  you happen to know why this order?

lib/posix-process/vfork.c
```diff
@@ -0,0 +1,36 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/posix-process/vfork.c
```diff
@@ -0,0 +1,36 @@
...
+	cl_args.flags       = CLONE_VM | CLONE_VFORK;
+	cl_args.exit_signal = SIGCHLD;
+
+	child_tid = uk_clone(&cl_args, sizeof(cl_args), execenv);
+	if (unlikely(child_tid < 0)) {
+		uk_pr_err("Could not clone thread\n");
+		return child_tid;
+	}
+
+	return tid2pthread(child_tid)->process->pid;
```
mogasergiu: I feel like we should have maybe some assertions or checks here?

mogasergiu (changes requested):

lib/ukbinfmt/binfmt.c
michpappas: Yes, they are all mandatory. The name is useful for messages. For ops, `unload()` is necessary for transient payloads.

lib/posix-process/Makefile.uk
michpappas: I wonder why this worked btw. Unless that `e` was eaten by a confilct during a rebase 🤔

lib/posix-process/vfork.c
michpappas: I'll add an assertion for the child process

lib/posix-process/execve.c
michpappas: IIRC it's because the posix-process API does not provide accessors so I had to resort to that double indirection. I'll add an assertion to make sure `pthread->process` is not NULL.

lib/posix-process/clone.c
michpappas: Yes, I went through that path as well. The problem with using local variables is that it would require making every subsequent access to `cl_args->stack` and `cl_args->tls` conditional to:
```c
if ((flags & CLONE_VM) && (!cl_args->stack))
  /* use local var */
else
 /* use cl_args->stack */

if ((flags & CLONE_VM) && (!cl_args->tls))
  /* use local var */
else
 /* use cl_args->tls */
```
Because these variables are used in several places in `clone()`, that would result in a lot of code duplication which I wanted to avoid as the implementation is complex enough already.

lib/posix-process/clone.c
michpappas: Maybe I don't fully understand the question, but I don't see a problem. `tid` is assigned by `libposixprocess` at thread creation (https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/process.c#L462) and it's not related to whether the thread is added to the scheduler or not.

mogasergiu: Sorry I got confused by that `ret`. I thought that we are calling a function that returns a value and right afterwards we call a different function without even checking if that `ret` is valid or an error.... but in reality it's just a misnamed local variable... probably `tid` would have been more appropriate instead of `ret`. Anyway it's not realted to this PR I guess...

lib/posix-process/process.h
```diff
@@ -41,6 +41,21 @@
 
 #define TIDMAP_SIZE (CONFIG_LIBPOSIX_PROCESS_MAX_PID + 1)
 
+/* Notice: The RUNNING state is not necessarily in sync with the state
+ * of the undelrying uk_thread (may be blocked by the scheduler).
```
mogasergiu: s/undelrying/underlying/
Btw, you also used both `uk_thread` and `uk-thread` in this paragraph.

lib/posix-process/Makefile.uk
```diff
@@ -9,6 +9,9 @@ CXXINCLUDES-$(CONFIG_LIBPOSIX_PROCESS) += $(LIBPOSIX_PROCESS_COMMON_INCLUDES-y)
 
 LIBPOSIX_PROCESS_CFLAGS-$(CONFIG_LIBPOSIX_PROCESS_DEBUG) += -DUK_DEBUG
 
+LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_EXECVE) += $(LIBPOSIX_PROCESS_BASE)/execve.c
+LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_EXECVE) += $(LIBPOSIX_PROCESS_BASE)/arch/$(CONFIG_UK_ARCH)/execve.c|arch
```
mogasergiu, line 13: `|arch`? Maybe `|$(CONFIG_UK_ARCH)` or I am missing smth but I do not see `|arch` used anywhere else.

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,38 @@
...
+	UK_ASSERT(sp);
+	UK_ASSERT(IS_ALIGNED(sp, UKARCH_SP_ALIGN));
+
+	execenv_new->regs.rip = ip;
+	execenv_new->regs.rsp = sp;
+
+	/* Prepare for iretq
+	 * FIXME re-arch: use GDT macros once moved out of plat/common
+	 */
+	execenv_new->regs.eflags = X86_EFLAGS_IF; /* enable IRQs on ctx switch */
```
mogasergiu: I don't think this is necessary anymore: <https://github.com/unikraft/unikraft/blob/staging/plat/common/x86/syscall.S#L188>? Either way, not sure about direct assignment, but maybe a `|=` instead.

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,38 @@
...
+	execenv_new->regs.cs = 8; /* GDT_DESC_OFFSET(GDT_DESC_CODE) */
+	execenv_new->regs.ss = 16; /* GDT_DESC_OFFSET(GDT_DESC_DATA) */
+
+	/* Copy current ectx to inerhit platform-initialized regs like mxcsr */
+	ukarch_ectx_sanitize((struct ukarch_ectx *)&execenv_new->ectx);
```
mogasergiu, line 31: I assume we do not do this on ARM because there we simply copy the Neon regs which are scratch registers as per ABI anyway? If yes we should for sure leave a comment on the ARM implementation as to why we do not do anything ECTX related there.

lib/posix-process/execve.c
```diff
@@ -0,0 +1,198 @@
...
+	loader_args.loader = NULL;
+	loader_args.user = NULL;
+
+	/* Assume that if argv is set the caller follows the convention */
+	loader_args.progname = (argv && argv[0]) ? argv[0] : "<null>";
+
+	/* Assign the default allocator to the loader. This will be used
+	 * to allocate memory for the executable image.
+	 */
+	loader_args.alloc = uk_alloc_get_default();
```
mogasergiu, line 115: I am wondering... should we maybe use current thread's allocator instead? In practice for now this will be the same thing sure but maybe in the future we will have a use case for separate per-thread allocators.

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -0,0 +1,38 @@
...
+			      struct ukarch_execenv *execenv,
+			      __uptr ip, __uptr sp)
+{
+	UK_ASSERT(execenv_new);
+	UK_ASSERT(execenv);
+	UK_ASSERT(ip);
+	UK_ASSERT(sp);
+	UK_ASSERT(IS_ALIGNED(sp, UKARCH_SP_ALIGN));
+
+	execenv_new->regs.rip = ip;
```
mogasergiu, line 20: Where do you init the other gp regs? Ditto for ARM.

lib/posix-process/execve.c
```diff
@@ -0,0 +1,198 @@
...
+	execve_arch_execenv_init(execenv_new, execenv,
+				 loader_args.ctx.ip,
+				 loader_args.ctx.sp);
+
+	/* Prepare process for executing new context. First notify posix
+	 * libraries that may have registered for the event, then do the
+	 * internal cleanup.
+	 */
+	event_data.thread = this_thread;
+	uk_raise_event(POSIX_PROCESS_EXECVE_EVENT, &event_data);
```
mogasergiu: This may return an error code.

lib/posix-process/clone.c
mogasergiu: ok but this still feels awkward and I am thinking of the case where someone may pass something to clone and wake up to clonetab having a different value somehow. Maybe at least there should be a warning/note comment somewhere?

mogasergiu (changes requested):

lib/posix-process/Makefile.uk
michpappas: That's just namespacing to avoid compile errors, as `execve.o` is already generated for `lib/posix-process/execve.c`. There are no semantics in `arch`, just a way to cause the build system generate `execve.arch.o`.

lib/posix-process/arch/x86_64/execve.c
michpappas: The thing is that we need to initialize it to something as currently the new context contains whatever happens to be in the stack, where it's allocated from. And that's why we also can't do `|=`.

lib/posix-process/arch/x86_64/execve.c
michpappas: I think it's okay to leave gpregs into an undefined state, as we jump into a new execution context.

lib/posix-process/arch/x86_64/execve.c
michpappas: Yes, we don't copy them though. We leave them in an undefined state similarly to the gpregs. I'll add a comment.

lib/posix-process/arch/x86_64/execve.c
michpappas: What I will do is copy the state from `execenv` instead.

lib/posix-process/execve.c
michpappas: This is intentional. Notice that below I use the thread's allocator for the new stack. The image uses the default allocator similarly to what we do in the `app-elfloader`, so I see it more as a matter of policy. It also avoids the potential issue of the image occupying a significant amount of the thread's dynamic memory pool.

mogasergiu (approved):

michpappas:
@mogasergiu 
> ok but this still feels awkward and I am thinking of the case where someone may pass something to clone and wake up to clonetab having a different value somehow. Maybe at least there should be a warning/note comment somewhere?

I pushed an update where I shadow elements of `cl_args` into locals.

michpappas:
Rebased from `staging`.

lib/ukbinfmt/include/uk/binfmt.h
```diff
@@ -0,0 +1,106 @@
...
+extern "C" {
+#endif
+
+#define UK_BINFMT_NOT_HANDLED	0  /* Not handled. Try next loader. */
+#define UK_BINFMT_HANDLED	1  /* Handled. Stop passing to loaders. */
+#define UK_BINFMT_HANDLED_CONT	2  /* Handled. Pass to next loader. */
+
+/** Loader type */
+enum uk_binfmt_type {
+	UK_BINFMT_LOADER_TYPE_NONE,
```
andreittr, line 25: Is there any use for having the enum value `UK_BINFMT_LOADER_TYPE_NONE`? If not, we can consider removing it.

The only use I can think of is ensuring calling code doesn't forget to set loader type when populating a `struct uk_binfmt_loader` via designated initializers. And for this purpose we should explicitly set this enum value to 0, so it gets caught by the assert in `uk_binfmt_register`.

I'm fine with either keeping (w/ explicit value of 0) or ditching, don't feel too strongly either way.

lib/posix-process/Config.uk
```diff
@@ -34,7 +34,13 @@ endif
 		the new thread does progress while the clone caller is not
 		context switching.
 
-	config LIBPOSIX_PROCESS_DEBUG
+config LIBPOSIX_PROCESS_EXECVE
+	bool "execve() system call"
+	select LIBPOSIX_PROCESS_PIDS
+	select LIBUKBINFMT
+	select LIBVFSCORE
```
andreittr: I think you can remove the hard-select of vfscore here, as no part of `posix-process` actually includes or calls into vfscore directly. From what I see `execve` takes paths and passes them to the binfmt loaders.
(Yes, you do need a VFS to eventually resolve those paths, but IMO that's best left to the binfmt implementation that actually does the VFS calls to `select`.)

lib/posix-process/Makefile.uk
```diff
@@ -24,7 +27,8 @@ LIBPOSIX_PROCESS_ASINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 LIBPOSIX_PROCESS_CXXINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 
 UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS_CLONE) += clone-5e
-UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS) += execve-3
+UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS_EXECVE) += execve-3e
+UK_PROVIDED_SYSCALLS-$(CONFIG_LIBPOSIX_PROCESS_VFORK) += vfork-2e
```
andreittr, line 33: I think this vfork syscall made it into the wrong commit when parceling :smile: 
(Don't bother rewriting history to fix this, it's perfectly fine)

lib/posix-process/execve.c
```diff
@@ -0,0 +1,202 @@
...
+
+	pthread_parent = pthread->parent;
+
+	/* If there is no parent, then must be calling execve()
+	 * from init without having called vfork().
+	 */
+	if (!pthread_parent) {
+		UK_ASSERT(pthread->process);
+		UK_ASSERT(pthread->process->pid == 0);
+		UK_ASSERT(pthread_parent->state != POSIX_THREAD_BLOCKED_VFORK);
```
andreittr: This 3rd assert might trigger UB, as you deref `pthread_parent->state` but you only reach this branch `if (!pthread_parent)` :thinking:

andreittr (commented):
Nice work, thanks! I've only some very minor comments, then we can get on with testing and merging.

lib/ukbinfmt/include/uk/binfmt.h
michpappas: Yes exactly, the only reason it's there is to make the check in `uk_binfmt_register()` easy. I think the value does not need to be explicitly initialized to zero, as that's guaranteed by the language.  From [ISO/IEC 9899](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) Sect 6.7.2.2:
```
The identifiers in an enumerator list are declared as constants that have type int and
may appear wherever such are permitted.109) An enumerator with = defines its
enumeration constant as the value of the constant expression. If the first enumerator has
no =, the value of its enumeration constant is 0.
```
An alternative would had been to set remove `UK_BINFMT_LOADER_TYPE_NONE`, explicitly set the first item to something other than zero, and remove the check altogether. Then validity will be enforced by the enum at compile time.
```
enum uk_binfmt_type {
    UK_BINFMT_LOADER_TYPE_NONE  = 1,
    UK_BINFMT_LOADER_TYPE_EXEC,
};
```
Overall it's not much different than what is currently there, and it can't protect from a caller that tries hard to do something stupid, so I don't have a strong opinion. Happy to go with either (or some alternative proposal, if you have one).

lib/posix-process/Config.uk
michpappas: Yup, thanks for catching that. The VFS requirement is on the `app-elfloader` side. Not sure why it's there, probably a leftover from early development. Will remove.

lib/posix-process/Makefile.uk
michpappas: Will fix, thanks 👍🏼

lib/posix-process/execve.c
michpappas: Yeah that shouldn't be there, removing.

lib/ukbinfmt/include/uk/binfmt.h
andreittr: I'm aware C guarantees the first enum value to be assigned 0, but I'm not so trusting of future developers remembering not to put any other enum before :smile: . Anyway, it's perfectly fine for now.

I do like the compile-time error idea, that's the best time to catch programming errors! And you don't get away with disabling asserts either :D maybe something for a later change, as it won't change the souce-level API any.

andreittr (approved):
Looks good to me, thanks!

andreittr:
Let's get it merged! :tada: 

PR #1387, opened 2024-04-17
lib/ukfile: Remove padding from struct uk_statx

andreittr:
### Description of changes

This change removes the padding fields from uk_statx, minimizing memory waste when allocating inside the kernel.
Syscalls never reveal these kernel structs, and userspace allocates a struct statx dictated by its libc.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukfile/include/uk/file/statx.h
+++ b/lib/ukfile/include/uk/file/statx.h
@@ -13,7 +13,7 @@
  * Linux-compatible `statx` structure for use in syscalls, along with
  * definitions of bit flags.
  *
- * Layout taken from statx(2) man page with paddings from musl v1.2.3.
+ * Layout taken from statx(2) man page cross-checked with musl v1.2.3.
  * Flag values taken from Linux headers v6.5.6 (include/uapi/linux/stat.h).
  */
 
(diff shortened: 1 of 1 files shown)


Mihnea27:
It fails one of the integration tests (can't even compile it). I tried the test locally and it seems to compile for me, so i am not sure what the issue might be with that.

Also above the definition of the structures there is a comment saying the padding is the same as in musl, I suppose that should be removed as it will not be consistent with that anymore.

andreittr:
> It fails one of the integration tests (can't even compile it). I tried the test locally and it seems to compile for me, so i am not sure what the issue might be with that.

I'm not quite sure why that is; the logs point to a patch failing on `libctest` itself, something this code doesn't remotely touch :shrug: might be a bug in the test lib.

> Also above the definition of the structures there is a comment saying the padding is the same as in musl, I suppose that should be removed as it will not be consistent with that anymore.

Good catch, thanks! Will remove.

StefanJum (approved):

andreistan26 (approved):

razvand (approved):

PR #1389, opened 2024-04-20
.github: Add label generation tool and update labels

### LINKED ISSUE #1283: `.github/labels`: Update list of labels to match current repository
### Feature request summary

Someone needs to go and update automated labels that are used inside unikraft.

First such case is the `area/` type. Someone needs to go and add all new areas, and also update existing ones in case there are description/color mismatches.

Everything else in the directory needs to be brought up to date so it matches current labels so we can use them as part of the CI.

After all this I can go and convince `governctl` to apply more tags than just `area/` ones.

[1] https://github.com/unikraft/unikraft/blob/staging/.github/labels/area.yaml 
[2] https://github.com/unikraft/unikraft/tree/staging/.github/labels

### Describe alternatives

_No response_

### Related architectures

None

### Related platforms

None

### Additional context

_No response_

arvganesh:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

Added `update_labels.py`, a tool to automatically update the files in `.github/labels` with configuration of latest labels from Github.

It only has one external dependency, [ruamel-yaml](https://yaml.readthedocs.io/en/latest/), which is a more powerful derivative of [PyYAML](https://pypi.org/project/PyYAML/#:~:text=YAML%20is%20a%20data%20serialization,parser%20and%20emitter%20for%20Python.) for interacting with YAML files.

### Description of changes

Update labels in `.github/labels` to match the current state of the repo.

To aid with this process in the future, I also wrote a tool (`update_labels.py`) to automate the majority of this process. It queries [this endpoint](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#list-labels-for-a-repository) in the GithubAPI, grabs the relevant data, and updates the YAML files in `.github/labels`.

GitHub-Fixes: #1283

### DIFF
--- a/.github/labels/Epic.yaml
+++ b/.github/labels/Epic.yaml
@@ -1,0 +1,4 @@
+labels:
+  - name: Epic
+    description:
+    color: "#3e4b9e"
--- a/.github/labels/RFC.yaml
+++ b/.github/labels/RFC.yaml
@@ -1,0 +1,4 @@
+labels:
+  - name: RFC
+    description: Request for Comments (do not review yet)
+    color: "#7239ec"
--- a/.github/labels/area.yaml
+++ b/.github/labels/area.yaml
@@ -1,12 +1,12 @@
 labels:
   - name: area/arch
-    description: Issues and PRs which affect architecture code
+    description: Unikraft Architecture
     color: "#7ec2f7"
     apply_on_pr_match_paths:
       - arch/**
 
   - name: area/docs
-    description: Issues and PRs which affect documentation
+    description: Documentation
     color: "#db4eb8"
     apply_on_pr_match_paths:
       - doc/**
@@ -19,40 +19,40 @@ labels:
       - exportsyms.uk
 
   - name: area/include
-    description: Issues and PRs which affect headers
+    description: Part of include/uk
     color: "#e99695"
     apply_on_pr_match_paths:
       - include/**
 
   - name: area/kconfig
-    description: Issues and PRs which affect KConfig code
+    description: Part of the Unikraft KConfig option system
     color: "#ed8ebd"
     apply_on_pr_match_paths:
       - Config.uk
 
   - name: area/lib
-    description: Issues and PRs which affect internal Unikraft libraries
-    color: "#ed8ebd"
+    description: Internal Unikraft Microlibrary
+    color: "#efbf67"
     match_repos:
       - unikraft
     apply_on_pr_match_paths:
       - lib/**
 
   - name: area/makefile
-    description: Issues and PRs which affect Makefile code
+    description: Part of the Unikraft Makefile build system
     color: "#ffe5c9"
     apply_on_pr_match_paths:
       - Makefile.uk
       - Makefile
 
   - name: area/plat
-    description: Issues and PRs which affect platform code
+    description: Unikraft Patform
     color: "#f2daa9"
     apply_on_pr_match_paths:
       - plat/**
 
   - name: area/support
-    description: Issues and PRs which affect support code
+    description: Support scripts, tools, services.
     color: "#bfd4f2"
     apply_on_pr_match_paths:
       - support/**
--- a/.github/labels/arm.yaml
+++ b/.github/labels/arm.yaml
@@ -1,8 +1,12 @@
 labels:
   - name: arm/smcc
-    description: Secure Monitor Call Calling Convention on ARM
-    color: "#0BA2F3"
+    description:
+    color: "#ededed"
     apply_on_pr_match_paths:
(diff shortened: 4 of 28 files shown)


.github/update_labels.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Updates .github/labels with labels that are retrieved via the Github API.
+def update_repo(labels: defaultdict(list)):
+    for file_name in labels:
+        update_repo_for_file(file_name, labels[file_name])
+
+# Usage: python update_labels.py
+if __name__ == "__main__":
+    labels = get_labels()
+    update_repo(labels)
```
craciunoiuc:
```suggestion
    update_repo(labels)

```

Missing newline at the end

.github/update_labels.py
```diff
@@ -0,0 +1,135 @@
...
+    label_path = f"{base_label_dir}/{file_name}.yaml"
+    template_path = "LABEL_TEMPLATE.yaml"
+    label_file_exists = os.path.isfile(label_path)
+    old_yaml_path = label_path if label_file_exists else template_path
+
+    with open(old_yaml_path, "r+") as old_yaml_file: # open exisiting YAML file or template.
+        labels_from_file = yaml_context.load(old_yaml_file)
+
+    output_file = open(label_path, "w")
+
```
craciunoiuc: Some trailing whitespace here

.github/update_labels.py
```diff
@@ -0,0 +1,135 @@
+from collections import defaultdict
```
craciunoiuc: Can you move this script and the template to a separate `scripts/` directory?

craciunoiuc (changes requested):
Hey @arvganesh 

This initial version looks really good, left you some comments.

Can you squash you commits and leave only two:
1. `.github/scripts: Add label generation tool`
2. `.github/labels: Update labels to track latest staging`

You will also need to add commit message bodies for each to describe the changes. (in order for the `checkpatch` check to pass)

craciunoiuc:
In a separate PR, if you want, we can also add a workflow that runs on pushes (or weekly) and checks if labels changed and updates them automatically.

nderjung (changes requested):
Hi @arvganesh! Thanks for this contribution. We use an external tool called [`governctl`](https://github.com/unikraft/governance) to do the heavy lifting in terms of interacting with GitHub's API, performing operations specifically on the Unikraft core repository as well as organization-wide repositories and teams.

In order to prevent too many tools, scripts or programs dotted all over the place, I suggest looking at how label synchronization can be accomplished with this tool under a new subcommand `governctl labels sync`.

One thing to note is that these label config files are in fact meant to work in the opposite direction: their static declaration should dictate GitHub's state of labels (aka "Infrastructure as Code").  This means that labels should be updated from the repo and not the other way around.

I think this particular issue can be kept nearly as-is sans the introduction of the python script, changes to descriptions and changing the hex color values. 

For direct changes to the labels, please:

- Ensure all color attributes are quoted, lowercase and include a pound (`#`) symbol i.e.: `"#1d76db"` is better vs. `1D76DB`.  This prevents the diff from needless changes;
- Only change the description if it is a better representation of what the label is intended;
- Remove `libx/posix-unisocket`;

Thanks!

arvganesh:
Hi @nderjung. Thanks for getting back to me! Had a few follow-up questions:

Re: `governctl labels sync`, are you suggesting that I should look into adding the label generation code to `governctl`? 

Re: The changes in this PR, there were many label files that were "created" (i.e labels were on GitHub but not in `./github/labels`). Since you said the contents of `.github/labels` are the ground truth, should I revert any new label files that were created?

nderjung:
> Re: governctl labels sync, are you suggesting that I should look into adding the label generation code to governctl?

Yes.

> Re: The changes in this PR, there were many label files that were "created" (i.e labels were on GitHub but not in ./github/labels). Since you said the contents of .github/labels are the ground truth, should I revert any new label files that were created?

The new libraries we can keep.  There are also changes which need to be reverted:

- There are additions and deletions which represent re-ordering of libraries.  E.g. `bug/fix`, `lifecycle/stale`, `lifecycle/rotten`, `priority/low`, etc.  Please double check all entries and please keep their order.
- A lot of descriptions are changed, I assume because they were taken what GitHub has.  Having gone through the list, I think the best approach is to take the longer description.  E.g. "Issues and PRs related to Linux Userspace platform" vs "Unikraft for Linux Userspace", the former is better.

Thanks!

razvand:
@arvganesh , could you please see @nderjung 's comments?

PR #1394, opened 2024-04-24
lib/posix-fdio: Add bincompat support for RWF_*

andreittr:
### Description of changes

This change adds values for RWF_* flags in posix-fdio, allowing it to interpret their meaning even without support from our (no)libc. This enhances binary compatibility.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/posix-fdio/fdio.c
+++ b/lib/posix-fdio/fdio.c
@@ -22,6 +22,22 @@
 
 #define _SHOULD_BLOCK(r, m) ((r) == -EAGAIN && _IS_BLOCKING((m)))
 
+/* RWF flags not supported by our libc's yet; defining them for bincompat */
+#ifndef RWF_NOWAIT
+#define RWF_NOWAIT	0x08
+#endif /* RWF_NOWAIT */
+
+#ifndef RWF_SYNC
+#define RWF_SYNC	0x04
+#endif /* RWF_SYNC */
+
+#ifndef RWF_DSYNC
+#define RWF_DSYNC	0x02
+#endif /* RWF_DSYNC */
+
+#ifndef RWF_APPEND
+#define RWF_APPEND	0x10
+#endif /* RWF_APPEND */
 
 static inline
 ssize_t fdio_get_eof(const struct uk_file *f)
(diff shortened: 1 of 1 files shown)


Mihnea0Firoiu (approved):
All good.

robertZamfir601:
Hello! The PR looks good, the functions are a lot clearer this way. I have two questions/observations though: 

1. For `RWF_NOWAIT`, previously the flag was ignored, but the user could still read. Now we will return an error that might break some apps that worked previously. Could this be a problem?
2. Should we use `uk_pr_warn_once` for `RWF_SYNC`, `RWF_DYNC`  and `RWF_APPEND` also? 

Everything else looks good.

andreittr:
Hi @robertZamfir601 , thanks for your review!
> For RWF_NOWAIT, previously the flag was ignored, but the user could still read. Now we will return an error that might break some apps that worked previously. Could this be a problem?

This is a good point. While at first it might seem more compatible to ingore the flag and move ahead with potentially blocking I/O, that poses issues for workloads that rely on the documented semantics of that flag.
The program woud run, yes, but it may not work as expected, and without a clear sign of why or that it's even happening.

A thread blocking or hanging at the wrong time is a much trickier condition to debug than an outright crash, so I chose to err on the side of "fail loud and early", and purposely fix any issues that pop up in the future, rather than risk silent misbehavior. This is especially true since you cannot set any of these flags "by mistake" (i.e. inadvertently through other libc funcs); you have to explicitly call `preadv2` (a glibc-only func) with `RWF_NOWAIT`, in which case I have to assume your code _really wants_ that behavior.

Eventually, once filesystems come into play, I plan to revisit meaningfully implementing `RWF_NOWAIT` to remove that error case.

> Should we use uk_pr_warn_once for RWF_SYNC, RWF_DYNC and RWF_APPEND also?

I don't think it is necessary; according to the manpage, those flags are indeed only _meaningful_ for `pwritev2`, but they are not disallowed in calls to `preadv2` (similar for `RWF_NOWAIT` for write). I think it's valid behavior to treat them as no-ops.

robertZamfir601 (approved):
I understand, sounds good to me.

Reviewed-by: Robert Zamfir <georobi.016@gmail.com>

razvand (approved):

PR #1272, opened 2024-01-22
lib/posix-*: Add support for packet pipes & datagram unix sockets

andreittr:
### Description of changes

This changeset adds support for packet-mode pipes and datagram unix sockets, along with accompanying code to better make use of them. Specifically:
- Have `posix-fdio` honor the `O_DIRECT` flag set on open file decriptions and pass it to the I/O operations
- Replace implementation of `posix-pipe`
- Stream- or packet-mode is selected at runtime by the O_DIRECT flag passed to the `write` operation
- Add support for `AF_UNIX` sockets of type `SOCK_DGRAM` and `SOCK_SEQPACKET` (0-length datagrams are not supported)
- Add warning about not supporting 0-length datagrams

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Additional configuration

```
CONFIG_LIBPOSIX_PIPE=y
CONFIG_LIBPOSIX_PIPE_PACKET=y
CONFIG_LIBPOSIX_UNIXSOCKET=y
```

Pipe test snippet:
```c
int main(void)
{
	int p[2];
	char b[32];

	int r = pipe2(p, O_DIRECT|O_NONBLOCK);
	assert(!r);

	/* These writes are packets */
	printf("%zd\n", write(p[1], "qwert", 5));
	printf("%zd\n", write(p[1], "asdfgh", 6));
	printf("%zd\n", write(p[1], "12345678", 8));
	errno = 0;
	fcntl(p[1], F_SETFL, O_NONBLOCK);
	printf("-%d\n", errno);
	/* Switched to stream now */
	printf("%zd\n", write(p[1], "qwert", 5));
	printf("%zd\n", write(p[1], "asdfgh", 6));
	printf("%zd\n", write(p[1], "12345678", 8));
	errno = 0;
	fcntl(p[1], F_SETFL, O_DIRECT|O_NONBLOCK);
	printf("-%d\n", errno);
	/* Switched back to packet; however, stream still unread, writes append to it */
	printf("%zd\n", write(p[1], "qwert", 5));
[...]

PR #1396, opened 2024-04-24
lib/ukfile: Add utility inlines for iovec I/O

andreittr:
### Description of changes

This change adds a utility header providing convenience inlines for doing I/O on buffers described by struct iovec, namely:
- zero out
- scatter data from buffer to iov
- gather data from iov into buffer

This code is not used by any lib or core component yet, however, the following PRs may benefit:
- https://github.com/unikraft/unikraft/pull/1226
- https://github.com/unikraft/unikraft/pull/1272

If the above are merged before this PR, will append commits w/ patches here.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukfile/include/uk/file/iovutil.h
+++ b/lib/ukfile/include/uk/file/iovutil.h
@@ -1,0 +1,171 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Convenience operations for dealing with iovec */
+
+#ifndef __UKFILE_IOVUTIL_H__
+#define __UKFILE_IOVUTIL_H__
+
+#include <string.h>
+#include <sys/uio.h>
+
+#include <uk/essentials.h>
+
+/**
+ * Zero out at most `len` bytes in memory regions described by `iov[iovcnt]`,
+ * starting at `*curp` offset from the buffer at `iov[*iovip]`.
+ *
+ * If the total remaining space in `iov` is less than `len`, exit early.
+ * After the call, `*iovip` and `*curp` are updated with the new positions.
+ *
+ * @return Number of bytes zeroed
+ */
+static inline
+size_t uk_iov_zero(const struct iovec *iov, int iovcnt, size_t len,
+		   int *iovip, size_t *curp)
+{
+	size_t ret = 0;
+	int i = *iovip;
+	size_t cur = *curp;
+
+	UK_ASSERT(i < iovcnt);
+	UK_ASSERT(cur < iov[i].iov_len);
+	if (cur) {
+		const size_t l = MIN(len, iov[i].iov_len - cur);
+
+		memset((char *)iov[i].iov_base + cur, 0, l);
+		ret += l;
+		len -= l;
+		cur += l;
+		if (cur == iov[i].iov_len) {
+			i++;
+			cur = 0;
+		}
+	}
+	if (len) {
+		UK_ASSERT(!cur);
+		while (i < iovcnt && len) {
+			const size_t l = MIN(iov[i].iov_len, len);
+
+			memset(iov[i].iov_base, 0, l);
+			ret += l;
+			len -= l;
+			if (len)
+				i++;
+			else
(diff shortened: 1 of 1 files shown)


SerbanSo:
Looks good.

I see the only difference between `uk_iov_scatter` and `uk_iov_gather` is the source and destination in the memcpy function. One idea would be to make a more general method that contains the code while making `uk_iov_scatter` and `uk_iov_gather` wrappers around that method. I don't think it will be a big improvement for the moment, but it might be useful in the future.

Reviewed by: Serban Sorohan serban.sorohan@gmail.com

SerbanSo (approved):

lib/ukfile/include/uk/file/iovutil.h
```diff
@@ -0,0 +1,165 @@
...
+ * @return Number of bytes zeroed
+ */
+static inline
+size_t uk_iov_zero(const struct iovec *iov, int iovcnt, size_t len,
+		   int *iovip, size_t *curp)
+{
+	size_t ret = 0;
+	int i = *iovip;
+	size_t cur = *curp;
+
```
Mihnea27, line 33: I think it should be checked that `*iovip < iovcnt` and `*curp < iov[*iovip]`, for all functions in this file.

lib/ukfile/include/uk/file/iovutil.h
```diff
@@ -0,0 +1,165 @@
...
+		ret += l;
+		len -= l;
+		cur += l;
+		if (cur == iov[i].iov_len) {
+			i++;
+			cur = 0;
+		}
+	}
+	if (len) {
+		UK_ASSERT(!cur);
```
Mihnea27, line 49: This check is not needed right? There is no way to reach this point with `cur != 0`, unless in the above `if` it ends up being larger than `iov[i].ion_len` which can happen only if it was originally larger that that (which should be checked initially as said above)

Mihnea27 (commented):
Suggested a couple changes, they apply for all functions not just the first one. Besides those, everything looks correct.

Reviewed by: Mihnea Popeanga popeanga.mihnea@gmail.com

lib/ukfile/include/uk/file/iovutil.h
andreittr: You mean as in bounds checking? Yeah, good catch, will add assertions for that. I'm choosing assertions over defined checks because:
- caller must ensure indexes are placed appropriately before the call => it is illegal to supply out-of-bounds args and we must crash loudly if that happens
- this code is intended to be used in inner loops, and checks should be optimized away for performance

lib/ukfile/include/uk/file/iovutil.h
andreittr: In theory it is indeed not needed, as you said previous code ensures this precondition.
However, I generally find asserts useful not only for checking against undefined inputs, but also to express non-trivial preassumptions of a block of code, both for the computer and programmer.
Obviously simple cases like
```c
int x = 0;
UK_ASSERT(!x);
```
are redundant, as you can trivially tell at a glance that the condition is always true, but the more complex it gets, the more time and context you need to deduce these preassumptions. You could then add comments like `/* We know x to be 0 from now on */` but those are inferior to asserts IMO.

This specific case I agree is borderline trivial, but I consider the added clarity and programmer context this provides to be worth the small execution time cost in a debug build.
That said, I could be persuaded to remove the assert if you think the previous code is trivial enough.

lib/ukfile/include/uk/file/iovutil.h
Mihnea27: I agree assertions are better, in this case out-of-bounds access should crash.

lib/ukfile/include/uk/file/iovutil.h
Mihnea27: Okay that makes sense. I didn't realise asserts with `UK_ASSERT` are optimised out in release builds. In that case I think it is okay to keep this assert. It is not really 100% clear that `cur` is 0 here as the logic in the previous `if` block is not completely trivial.

Mihnea27 (approved):
Everything looks good now!

Reviewed by: Mihnea Popeanga popeanga.mihnea@gmail.com

eduardvintila (approved):
Nice addition, thanks.

jschlumpp (approved):

PR #1407, opened 2024-05-02
lib/ukvmem: Move file VMA implementation into vfscore

andreittr:
### Description of changes

This changeset moves the implementation of the file VMA type from `ukvmem` to `vfscore`, bringing file VMAs under the same roof as file operations more generally. The implementation itself is left minimally changed, and no difference to behavior should be observed. Fixing and/or otherwise improving the existing file VMA code falls outside of scope.

As prerequisites for its main goal, this PR includes several tweaks to the `ukvmem` API, making it easier for libs to implement their own custom VMA types.

Following this change, `ukvmem` remains as before responsible for virtual memory management & the VMA API, but only provides implementations for fundamental VMAs: anonymous, stack, DMA, reserved.
Other, custom types of VMAs are to be implemented by their respective libs.

### Prerequisite checklist

### Base target

### Additional configuration

N/A; behavior and configuration should be unaffected.


lib/ukvmem/include/uk/vma_ops.h
```diff
@@ -0,0 +1,40 @@
...
+
+/* Argument list is empty to accept (and ignore) any number of args.
+ *
+ * Do not convert to `uk_vma_op_deny(void)` no matter how much your static code
+ * checker screams at you.
+ */
+/**
+ * Always fail, denying the operation.
+ */
+int uk_vma_op_deny();
```
mogasergiu: Should this maybe be
```C
int uk_vma_op_deny(void);
```
(I know that it initially did not have it either, but the comment even describes it as such)

lib/ukvmem/include/uk/vma_types.h
```diff
@@ -210,60 +210,8 @@ static inline int uk_vma_map_dma(struct uk_vas *vas, __vaddr_t *vaddr,
...
-	UK_ASSERT(fd >= 0);
-	UK_ASSERT(offset >= 0);
-	UK_ASSERT(PAGE_ALIGNED(offset));
-
-	return uk_vma_map(vas, vaddr, len, attr, flags, __NULL,
-			  &uk_vma_file_ops, &args);
-}
-#endif /* CONFIG_LIBVFSCORE */
+/* Common flags for any file-like VMA implementations */
+#define UK_VMA_FILE_SHARED	(1ULL << UK_VMA_MAP_EXTF_SHIFT)
```
mogasergiu, line 214: I have the feeling that this might now be out of place. My understanding is that vfscore is going to now implement file VMA's as a base and so if anyone would require any base definition of such VMA's they would need to enable vfscore. When vfscore will get retired, where will these go instead? Regardless of vfscore or whatever library these definitions would go to, I think this bitfield is supposed to be part of a flag, flag whose definition has moved to vfscore and I guess so should this bitfield.

Though I may be missing something in the grand scheme of VFS rearch? What do you think? :thinking:

lib/ukvmem/tests/test_vmem.c
```diff
@@ -688,6 +688,8 @@ UK_TESTCASE(ukvmem, test_vma_stack)
 #include <stdlib.h>
 #include <unistd.h>
 
+#include <vfscore/vma.h>
```
mogasergiu: Should maybe the file VMA tests also go to vfscore? It's just that I now see vfscore's VMA file implementation as a derivative/expansion of ukvmem's exposed definitions, which thus makes ukvmem a base that maybe should not be given awareness of its derivatives, but instead keep them contained in the library that does this said expansion.

I see this sort of like: ukvmem becomes parent "class", vfscore's file vmas extend ukvmem and thus any overrides or "unit tests" should also go there :thinking:

mogasergiu (changes requested):
Good changes. It makes sense to me to move these out of ukvmem :muscle:.
I have a couple of comments, likely more of a personal opinion only :man_shrugging:.

lib/ukvmem/include/uk/vma_ops.h
andreittr: Unfortunately not, because they mean different things:
- `int uk_vma_op_deny(void)` takes _exactly_ 0 arguments
- `int uk_vma_op_deny()` takes any number of arguments of any type

We specifically need the latter, because the purpose of `uk_vma_op_deny` is to be a drop-in "always fail" function for _any_ entry in a `struct uk_vma_ops`, each of which is a function pointer with different signatures which an `int (*func)(void)` would not be compatible with (and make most compilers error out).

mogasergiu: Ah, yes, actually this is what the comment itself states. I don't know what I read :sweat_smile: .

lib/ukvmem/include/uk/vma_types.h
andreittr: > My understanding is that vfscore is going to now implement file VMA's as a base

That's not entirely accurate, vfscore is going to implement file-like VMAs for vfscore files only, and offer no "base" functionality whatsoever -- vfscore is an "end-user" if you will.
File stacks other than vfscore (like the new `uk_file` family) will have their own VMA implementation(s) provided likely by `posix-fdio` with only `ukvmem` as dependency.
This is entirely orthogonal and independent from what vfscore does. Once vfscore goes away, so will its file VMA impl.
Any common concepts to what a "file"-like VMA should look like ("base definitions" as you say) stay in ukvmem.

Addressing this bit specifically, I think you misunderstand, the flag field this is used in hasn't moved to vfscore, it's still in `ukvmem` as part of `struct uk_vma` (the `.flags` field) and can be set like any other `UK_VMA_FLAG_*` bit.
The definition of `UK_VMA_FILE_SHARED` IMO needs to be here because all file-like VMAs must support being _asked_ for a shared mapping (even if they error out on it like vfscore). Think of it as part of the public file-like VMA interface.
 This simplifies code that creates VMAs such as `mmap`, so it doesn't have to look up how each provider of file VMAs wants the "shared" flag to be set on vma->flags.

lib/ukvmem/tests/test_vmem.c
andreittr: ooof, yeah, good catch. This test should definitely be moved to vfscore with the VMA impl; will do.

lib/ukvmem/include/uk/vma_types.h
mogasergiu: Following a more detailed discussion we have decided that this flag makes sense to stay here, but the name might need changing for the sake of namespacing. That being said, the idea of a "shared VMA" can be applied to any kind of VMA and, besides, the bitfield, as you said, is associated with the `flags` of `struct uk_vma`. Therefore it makes more sense to rename it to `UK_VMA_FLAG_SHARED` instead.

andreittr: Agree; pushed a new commit with this rename

lib/ukvmem/include/uk/vmem.h
```diff
@@ -129,6 +129,16 @@ struct uk_vma {
 	const char *name;
 };
 
+/**
+ * Returns the length of a VMA in bytes.
```
mogasergiu, line 133: Nitpick, but since we make it public and place it here, let's also align it with the others and add the typical `@param/@return` comments

lib/vfscore/Makefile.uk
```diff
@@ -17,6 +17,14 @@ LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/lookup.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/fops.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/subr_uio.c
 LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/extra.ld
+
+ifeq ($(CONFIG_LIBUKVMEM),y)
+LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/vma_file.c|isr
+ifneq ($(filter y,$(CONFIG_LIBUKVMEM_TEST) $(CONFIG_LIBUKTEST_ALL)),)
+LIBVFSCORE_SRCS-y += $(LIBVFSCORE_BASE)/tests/test_vma.c
```
mogasergiu: Should we maybe include this to the build conditioned by a vfscore test config in addition to it being enabled only in ukvmem is enabled? Something like
```
LIBVFSCORE_SRCS-$(LIBVFSCORE_TEST) += $(LIBVFSCORE_BASE)/tests/test_vma.c
```

lib/vfscore/tests/test_vma.c
```diff
@@ -0,0 +1,119 @@
...
+
+	free(buf);
+
+	/* Create the mapping */
+	va1 = __VADDR_ANY;
+
+	rc = uk_vma_map_file(vas, &va1, 3 * PAGE_SIZE, PROT_R, 0, fd, 0);
+	UK_TEST_EXPECT_ZERO(rc);
+
+	len = probe_r(va1, 3 * PAGE_SIZE);
```
mogasergiu: I think the build is failing because you are using some function definitions from the original teast_vmem.c

mogasergiu (changes requested):

lib/vfscore/Makefile.uk
mogasergiu: Actually, I take that back. Maybe it is better to just do something like:
LIBVFSCORE_SRCS-$(LIBVFSCORE_TEST) += $(LIBVFSCORE_BASE)/tests/test_vfscore.c
and inside that file have the addition `#if vmem` guarded. My rationale is that this would be where vfscore would have its tests and at the moment there is only the test for file VMA's. That being said, what I am thinking is that vma_file.c would become the only file in the makefile confitioned by ukvmem.

andreittr: That would make sense if there were any other vfscore tests to speak of, or if we ever intend to add any (we don't, vfscore is to be deprecated). I chose to stay compatible with previous configs to minimize impact, so file VMA tests run if-and-only-if: (1) vfscore is enabled, (2) ukvmem is enabled, and (3) vmem tests or _all_ tests are enabled.

andreittr:
Rebased to staging

lib/ukvmem/include/uk/vma_ops.h
```diff
@@ -0,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2025 Maybe?

lib/vfscore/include/vfscore/vma.h
```diff
@@ -0,0 +1,61 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu, line 2: Ditto

lib/vfscore/include/vfscore/vma.h
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Creates a new file mapping. See uk_vma_map() for a description of the
+ * parameters not listed here.
+ *
+ * @param fd
+ *   File descriptor of the file to map into memory. The file descriptor must
+ *   have been opened with sufficient permissions to allow for all operations
+ *   permitted by the VMA's attributes.
+ * @param offset
+ *   Offset within the file to map in the VMA. Must be aligned to the page size.
```
mogasergiu, line 42: `@return` or also mention `return/parameters not listed here`?

lib/vfscore/tests/test_vma.c
```diff
@@ -0,0 +1,211 @@
...
+				 STRINGIFY(cond));			\
+	} while (0)
+
+struct vma_entry {
+	__vaddr_t start;
+	__vaddr_t end;
+	unsigned long attr;
+};
+
+static void vmem_print(struct uk_vas *vas)
```
mogasergiu: Ok so in order to fix the tests we had to copy-paste for now and we leave it for later to address this. FIne by me. Can you please add a comment here about it giving the reason why it's duplicated?

mogasergiu (commented):
Some minor nitpicks, otherwise looks good to go in.

lib/ukvmem/include/uk/vma_ops.h
andreittr: Copyright notices in general are meant to include the year of _first publication_, not of last modification (`ctime`, not `mtime` in `stat` terms)(see [37 C.F.R. § 202.2](https://www.copyright.gov/title37/202/37cfr202-2.html) (a)(1) and (b)(6)(i)(A) or [wikipedia summary](https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies)), and since I have made this file public as part of this PR in 2024, the copyright notice should reflect that -- it's _the same work_, regardless if it sits in my branch or the main unikraft repo.

Saying this, I suspect most contributors are unaware of the intricacies of copyright law, and we should probably document this explicitly somewhere.

lib/vfscore/include/vfscore/vma.h
andreittr: This code was moved verbatim from [<uk/vma_types.h>](https://github.com/unikraft/unikraft/pull/1407/files#diff-af0f86d6a09f05c4a44e859d58dfe38ab95d442f01f60e26f92b345aae167f8fL265) and I purposely limited my changes to the minimum needed to maintain behavior. IMO improvements to file VMAs and their docs should fall outside of scope here.

lib/ukvmem/include/uk/vma_ops.h
mogasergiu: You're right, I've always thought it was about the time this gets upstream. Good to know.

mogasergiu:
Hmm, I just now noticed but, shouldn't the file vma configs be migrated as well?

andreittr:
On further consideration, there might not be a clean enough way to implement this PR without broadening its scope to include some more general refactoring that touches `ukvmem` and `posix-mmap`.

Converting to draft for now; will re-open as a more general work that includes:
- refactor of `ukvmem` API to better expose common VMA code + common VMA _test_ code cleanly
- this PR's current moving of vfscore file VMAs to vfscore
- ukfile-specific file VMAs
- changes to posix-mmap to work with the above

andreittr:
> On further consideration, there might not be a clean enough way to implement this PR without broadening its scope to include some more general refactoring that touches `ukvmem` and `posix-mmap`.
> 
> Converting to draft for now; will re-open as a more general work that includes:
> 
>     * refactor of `ukvmem` API to better expose common VMA code + common VMA _test_ code cleanly
> 
>     * this PR's current moving of vfscore file VMAs to vfscore
> 
>     * ukfile-specific file VMAs
> 
>     * changes to posix-mmap to work with the above

Turns out I was wrong on the latter two points; reopening this PR with the same scope: moving file VMAs into vfscore; this time with the addition of a cleanly exported interface for VMA unittest primitives.

mogasergiu (approved):

michpappas:

PR #1413, opened 2024-05-14
lib/ukboot: Fix configuration of no "no scheduler"

skuenzer:
### Description of changes

This PR fixes the KConfig configuration of "no scheduler" with `lib/ukboot`. Previously, scheduler initialization code was still compiled in even if no scheduling was selected. This happened if at the same time no allocator was configured as well.

Please note that this PR fixes only the configuration handling in `lib/ukboot`. I noticed a crash within one of the registered inittab functions if no scheduler was initialized. There is also an assertion failure if `lib/ukboot` is configured without allocator initialization (system does not boot without an allocator assigned to the platform libraries). Both need fixing with corresponding subsystem PRs and taking re-arch into account.

### DIFF
--- a/lib/ukboot/Config.uk
+++ b/lib/ukboot/Config.uk
@@ -81,7 +81,14 @@ if LIBUKBOOT
 	bool
 	default n
 
-	choice LIBUKBOOT_INITALLOC
+	# Hidden option that indicates that allocator is initialized
+	# We do this here in order to allow specifying dependecies to
+	# the allocator choice
+	config LIBUKBOOT_INITALLOC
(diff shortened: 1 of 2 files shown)


michpappas (approved):
Looks good. In the future we need to be a bit careful not to use `LIBUKBOOT_NOALLOC` /`LIBUKBOOT_NOSCHED`, also during reviews. I will wait for one more RB before approving for merge.

StefanJum (approved):
Works fine, thank you.

razvand (approved):

PR #1437, opened 2024-05-30
lib/posix-*: Remove hard dependency on posix-fdtab

andreittr:
### Description of changes

This changeset removes `posix-fdtab` as a hard dependency from all `posix-*` libraries that provide file implementations.
This improves modularity, allowing for files to be used internally without exposing a userspace-visible file descriptor, nor requiring a file descriptor table at all.

Naturally, support for file-specific binary syscalls continues to require fdtab, as file descriptors are the only interface for userspace to pass file references to the kernel.

### Prerequisite checklist

### Base target

### Additional configuration
N/A


lib/posix-eventfd/Config.uk
```diff
@@ -1,5 +1,4 @@
 config LIBPOSIX_EVENTFD
 	bool "posix-eventfd: Support for eventfd files"
 	select LIBPOSIX_FDIO
```
mogasergiu, line 3: > allowing non-fd parts of its functionality to work without the former selected

But posix-fdio will also automatically select posix fdtab, right? :thinking:.
Ditto for the other posix-* in the PR: they either select posix-fdio or posix-pipe (which selects posix-fdio).
Or is it that we will be able to select these without fdtab in a follow up PR?
Either way it is not bad to remove the dependency from the Config.uk since it is implicitly selected by other dependencies ( like posix-fdio).

lib/posix-timerfd/include/uk/posix-timerfd.h
```diff
@@ -9,13 +9,16 @@
...
 
+#include <uk/file.h>
+
 /* File creation */
 
 struct uk_file *uk_timerfile_create(clockid_t id);
 
 /* Internal Syscalls */
-
+#if CONFIG_LIBPOSIX_FDTAB
```
mogasergiu, line 19: I have just noticed this is not guarded in the C file though, is that ok? I see no one else in the library calling this directly other than the syscalls.

mogasergiu (commented):
A couple of questions.

lib/posix-eventfd/Config.uk
andreittr: > But posix-fdio will also automatically select posix fdtab, right? 🤔.

At the moment yes, but as you said, after this PR the dependency becomes _implicit_ => whenever `posix-fdio` becomes decoupled from `posix-fdtab` (coming _very_ soon, next PR in this series) so will all its users.

lib/posix-timerfd/include/uk/posix-timerfd.h
andreittr: Oh yeah, very good catch :pray: it most definitely needs guarding in the C file too. Will do.

mogasergiu (approved):

mogasergiu (approved):
Thanks. Merging this.

PR #1446, opened 2024-06-04
lib/{ukfile, posix-socket, posix-unixsocket}: Add support for explicitly polled files & sockets

andreittr:
### Description of changes

This changeset adds `ukfile` support for explicitly polled files. These files accomodate drivers that, in contrast to the present `ukfile` event polling model, maintain events internally and cannot or will not signal falling edges of events, and thus cannot have `ukfile` maintain event levels for them.
Explicitly polled files provide for this use-case by allowing drivers to implement a custom `poll` operation, which shall be called whenever consumer code requests the instantaneous levels of events. Signaling rising edges of events remains the responsibility of the driver.

Sockets can similarly be explicitly polled, with the API of `posix-socket` changed to accomodate this.
Unix sockets are patched by this PR; LWIP requires [a patch for the new API](https://github.com/unikraft/lib-lwip/pull/58).

### Prerequisite checklist

### Base target

### Additional configuration

N/A


andreittr:
Rebased to current staging

lib/ukfile/include/uk/file/pollqueue.h
```diff
@@ -304,6 +400,21 @@ int _pollq_wait(struct uk_pollq *q, uk_pollevent req, __nsec deadline)
 	return !timeout;
 }
 
+static inline
+uk_pollevent uk_pollq_poll_level(struct uk_pollq *q, uk_pollevent req,
```
michpappas, line 425: Since you've documented all functions in this header, maybe consider if this one should be documented too.

lib/ukfile/Config.uk
```diff
@@ -13,3 +13,6 @@ config LIBUKFILE_CHAINUPDATE
 config LIBUKFILE_FINALIZERS
 	bool
 	select LIBUKLOCK_MUTEX
+
+config LIBUKFILE_POLLED
```
michpappas, line 17: As a general note: Looking at `pollqueue.h` it seems that there is some ambiguity in naming within config, documentation, and the API, when it comes to managed / polled and edge / level (triggered).

Also out of curiosity - and without implying that they should be changed - what makes these options qualify as invisible compared to visible options offered by other libraries?

lib/ukfile/include/uk/file/pollqueue.h
```diff
@@ -26,6 +26,24 @@
...
+ *
+ * Drivers may choose to not provide this callback, in which case they are
+ * responsible for updating the current event levels with `uk_pollq_set`,
+ * `uk_pollq_clear`, and/or `uk_pollq_assign` in-band with I/O operations.
+ *
+ * If drivers do provide this, it will be called every time the instantaneous
+ * level of events is queried. Drivers are then responsible only for notifying
+ * the rising edges of events via `uk_pollq_set`.
+ */
+typedef uk_pollevent (*uk_poll_func)(const struct uk_file *f,
```
michpappas, line 52: Since this needs to implemented by drivers perhaps it would be worth documenting its parameters and return value. More importantly, as I see you don't do any error handling when calling this function, is that a design choice that drivers cannot err here?

michpappas (changes requested):
@andreittr very minor comments here. Also: out-of-scope for this PR, but although the `README.md` provides very good description of the underlying concepts, as someone getting familiar with `libukfile` for the first time, I found that it would be really helpful if the APIs for the driver and user as well as the config options were documented too.

lib/ukfile/include/uk/file/pollqueue.h
andreittr: > worth documenting its parameters and return value

Agreed, I'll expand the docstring.

> is that a design choice that drivers cannot err here?

Yes; `uk_pollevent` is defined as a "set of active events" and has no reserved value to signal "error in retrieving events".
I can't easily think of a situation where such an error can (1) occur, (2) be safely recoverable, and (3) meaningfully different from returning 0 (no events set). But if we do encounter such, I believe there are unused bits in `uk_pollevent` that we can reserve for this purpose.
I chose to not reserve such an error bit unless needed and used, because it would otherwise require input validation in all other funcs (set/clear/assign) that take event sets.

lib/ukfile/Config.uk
andreittr: > that there is some ambiguity in naming within config, documentation, and the API, when it comes to managed / polled and edge / level (triggered).

Yes, the double naming scheme came about as a result of thinking at different abstraction layers when designing&writing (file vs. raw pollqueue); I can't say I'm fully happy with it either, nor with the general naming tbh (pollqueue should really have been called eventqueue :sweat_smile: ). I'll have a closer look and see if I can harmonize these.

> what makes these options qualify as invisible compared to visible options offered by other libraries?

If you mean why are they hidden as opposed to user-visible: these options control extensions to the ukfile API that:
- are _not_ required for general use, but provide special accomodations to other libs:
  - chainupdate & finalizers support complex callers (unixsockets delegating to pipes & epoll autoremove shenanigans)
  - explicitly polled files support drivers that can't do falling-edge notifications
- have a (marginal) runtime cost -- so keeping them always-on when not needed would be wasteful
- have _no_ user-visible behavior by themselves (other than the aforementioned runtime cost)
- are _only_ relevant as part of the API negociation between `ukfile` and its set of consuming libraries
- make no sense to ever be manually enabled; only to be `select`ed by dependent libs

There is ofc an architectural discussion to be had on what should constitute a user-visible configuration knob vs. internal flag; but I feel this use-case is closer in concept to `#define _GNU_SOURCE` before including a libc header than to a true degree of freedom in config.

lib/ukfile/include/uk/file/pollqueue.h
michpappas: Okay fair enough, just wanted to make sure it was a intentional design decision. Thanks for the explanation, sounds reasonable.

lib/ukfile/Config.uk
michpappas: > I'll have a closer look and see if I can harmonize these.

Okay, it was more of a general comment, not necessarily for this PR. Up to you.

> I feel this use-case is closer in concept to #define _GNU_SOURCE before including a libc header than to a true degree of freedom in config

Yeah okay. So far we didn't pay attention to differentiate between options that make or don't make sense to be visible to the user; most of the time if a library *requires* something it just makes a hard select. And an arch-wide definition is a can of worms I don't want to open as I anticipate it will be hard to come up with solid rules.  But I do see your point, though, and it sounds reasonable, so let's keep as is.

michpappas (approved):

michpappas:

PR #1008, opened 2023-07-30
lib/ukrandom: Rename and refactor lib/ukswrand

Alex-deVis:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architectures: x86_64 and arm64
 - Platform: kvm, linuxu
 - Application(s): N/A

### Additional configuration

The library depends of `HAVE_RANDOM`.
`CONFIG_LIBUKRAND_HW=y`
`CONFIG_LIBUKRAND_CHACHA=y`

### Description of changes

Rename `lib/ukswrand` to `lib/ukrand` to better describe its use case. It enables users to customize their entropy source, allowing a choice between hardware-generated randomness or the output of the Chacha20 PRNG.
Both random devices (`/dev/random` and `/dev/urandom`) and `getrandom()` syscall will return random numbers based on the selected source of entropy.

PR #1451, opened 2024-06-05
lib/ukrandom: Alternative options for seeding the CSPRNG

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This PR updates the mechanisms used to seed the CSPRNG of `libukrandom`. The following changes are made:
* **Switch from `rdseed` to `rdrand` when seeding ChaCha:** This is due to the observation that on some x86_64 systems, the loop fails due to exhaustion of the conditioned entropy. In [1] Intel suggests that performance-crtical systems may use `rdrand` instead. Switch to `ukarch_random_u32()` to mitigate this risk. In the future there should be a more flexible mechanism to control seeding.
* **Introduce options to taint Unikraft:** These are options should bet set by any library that provides an option to bypass security. Upon selecting one of these options a message is printed at the main screen of Kconfig. In the future there may be additional hits for tainted builds.
* **Add option to seed the CSPRNG with a fixed value:** Add an option to seed the CSPRNG with a fixed value. The purpose of this option is for development, or running Unikraft in legacy systems that don't provide a TRNG. Enabling this option compromises the system's security and should not be used in production. Moreover, the build system marks Unikraft as tainted and libukrandom prints a banner during init.

GitHub-Depends-On: #1008

### DIFF
--- a/Config.uk
+++ b/Config.uk
@@ -235,3 +235,10 @@ config UK_NAME
 	default UK_DEFNAME
 	help
 	  Name to be used for final image
+
+# The following are selected by libraries that provide options to bypass security.
+config CONFIG_UK_TAINT_INSECURE_CSPRNG
+	bool
+
+comment "Notice: UNIKRAFT IS CONFIGURED WITH INSECURE RANDOM NUMBER GENERATOR"
+	depends on CONFIG_UK_TAINT_INSECURE_CSPRNG
--- a/lib/ukrandom/chacha.c
+++ b/lib/ukrandom/chacha.c
@@ -157,8 +157,13 @@ int uk_swrand_init(void)
 
 	uk_pr_info("Initialize random number generator...\n");
 
+	/* It has been observed that in some x86_64 systems this loop
+	 * fails after a few iterations due to exhastion of conditioned
+	 * entropy (rdseed). Use RDRAND until we provide more flexible
+	 * options for RDSEED.
+	 */
 	for (i = 0; i < seedc; i++) {
-		ret = ukarch_random_seed_u32(&seedv[i]);
+		ret = ukarch_random_u32(&seedv[i]);
 		if (unlikely(ret)) {
 			uk_pr_err("Could not generate random seed\n");
 			return ret;
--- a/lib/ukrandom/chacha.c
+++ b/lib/ukrandom/chacha.c
@@ -47,7 +47,8 @@
  *
  * TODO: bring the nonce-size in-line with the RFC.
  */
-#define CHACHA_SEED_LENGTH 10
+#define CHACHA_SEED_LENGTH		10
+#define CHACHA_SEED_INSECURE		0xdeadb0b0
 
 struct uk_swrand {
 	int k;
@@ -156,6 +157,12 @@ int uk_swrand_init(void)
 	int ret;
 
 	uk_pr_info("Initialize random number generator...\n");
+#if CONFIG_LIBUKRANDOM_SEED_INSECURE
+	uk_pr_err("*******************************************\n");
+	uk_pr_err("* This configuration uses an insecure RNG *\n");
+	uk_pr_err("*         DO NOT USE IN PRODUCTION        *\n");
+	uk_pr_err("*******************************************\n");
+#endif /* CONFIG_LIBUKRANDOM_SEED_INSECURE */
 
 	/* It has been observed that in some x86_64 systems this loop
 	 * fails after a few iterations due to exhastion of conditioned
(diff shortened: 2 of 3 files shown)


Alex-deVis (approved):
Reviewed-by: Alex Apostolescu <alexx.apostolescu@gmail.com>

Config.uk
```diff
@@ -235,3 +235,10 @@ config UK_NAME
 	default UK_DEFNAME
 	help
 	  Name to be used for final image
+
+# The following are selected by libraries that provide options to bypass security.
+config CONFIG_UK_TAINT_INSECURE_CSPRNG
```
mogasergiu, line 240: Sounds like something that would have the `HAVE_*` semantics? :thinking:

lib/ukrandom/chacha.c
```diff
@@ -156,13 +157,28 @@ int uk_swrand_init(void)
 	int ret;
 
 	uk_pr_info("Initialize random number generator...\n");
-
+#if CONFIG_LIBUKRANDOM_SEED_INSECURE
+	uk_pr_err("*******************************************\n");
```
mogasergiu, line 161: Why not `uk_pr_warn`?

lib/ukrandom/chacha.c
```diff
@@ -156,13 +157,28 @@ int uk_swrand_init(void)
...
+	/* It has been observed that in some x86_64 systems this loop
+	 * fails after a few iterations due to exhastion of conditioned
+	 * entropy (rdseed). Use RDRAND until we provide more flexible
+	 * options for RDSEED.
+	 */
 	for (i = 0; i < seedc; i++) {
-		ret = ukarch_random_seed_u32(&seedv[i]);
+#if CONFIG_LIBUKRANDOM_SEED_INSECURE
+		seedv[i] = CHACHA_SEED_INSECURE;
+#else /* CONFIG_LIBUKRANDOM_SEED_INSECURE */
```
mogasergiu: Shouldn't this have been `!CONFIG_LIBUKRANDOM_SEED_INSECURE`?

mogasergiu (changes requested):
Hi Michalis. I don't have much context about these changes. But had a quick look.

lib/ukrandom/chacha.c
michpappas: Because we want these to be printed even if the severity has been set to low. And also to be printed in red if `LIBUKDEBUG_ANSI_COLOR` is set.

mogasergiu: Will the catalog have these shown? I feel like this is still too agressive and misleading as it is not an error.

> we want these to be printed even if the severity has been set to low

Why not `crit` then?

michpappas: It should be aggressive, if you disable the CSRNG the security of the entire system is compromised. I am not using `crit` because according to @skuenzer that level should only used when we crash Unikraft.

Config.uk
michpappas: Yes for sure, that could also be an option. I used `TAINT` here instead to group future options that taint Unikraft.

lib/ukrandom/chacha.c
mogasergiu: Still, error feels wrong as it is not an error and you can still configure Unikraft to not show errors the same way you can configure it to not show warnings :thinking:.

skuenzer: I agree, normally we would have used `warn` for this however I think this particular case is a very special and requires increasing the seriousness of the message with `err`.
The purpose of `lib/ukrand` is to provide cryptographically secure pseudo random numbers. Anything just random without crypotgraphically security can and should be done with the libc (nolibc has the according `randr()` call as well). We introduce only one option to make a system deterministic for debugging purposes (and only for this use case) via a compiled-in seed. For example, this can be useful when we continue implementing ASLR. Any other cryptographically insecure random number generators are not allowed in this lib because it breaks assumptions that application code relies on. So, the error is there to make this dominant, and yes, it is maybe arguable if breaking the assumption is a warning or an error but because we want to provide a secure environment with the Unikraft project and avoid configuration mistake as much as possible (e.g., by mistake an image configured insecurely) we should increase the seriousness. When configuring images, it feels easier to hide warnings, but not errors. And since we consider broken CSRNG to be a serious flaw, it is an error that should not be overlooked. Of course you always can workaround this (e.g., hiding errors), but the measures we implement make it harder to ignore.

skuenzer: Forgot to mention: The "crit" message level is intended to message things that are so serious that the system cannot continue operating, like running into a crash.

mogasergiu: k

Config.uk
mogasergiu: Ok I will approve anyway. I am fine either way with this.

mogasergiu (approved):

razvand (approved):

PR #1454, opened 2024-06-07
lib/nolibc: Compile string.c with -O3

andreittr:
### Description of changes

This change makes nolibc's string.c compile with -O3 optimizations when optimizing for performance. This gives compilers more freedom to better optimize the often-used mem* and str* functions.

### Prerequisite checklist

### Base target

### Additional configuration
Compile with `CONFIG_OPTIMIZE_PERF` with & without this patch. Disassemble the produced `string.o` and examine e.g., `memcpy`.

### DIFF
--- a/lib/nolibc/Makefile.uk
+++ b/lib/nolibc/Makefile.uk
@@ -33,6 +33,7 @@ LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/stdio.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/ctype.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/stdlib.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/string.c
+LIBNOLIBC_STRING_FLAGS-$(CONFIG_OPTIMIZE_PERF) += -O3
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/musl-imported/src/string/strsignal.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/musl-imported/src/string/strstr.c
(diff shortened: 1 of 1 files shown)


StefanJum (approved):

adinamariav (approved):
Reviewed-by: Adina-Maria Vaman adinamariav22@gmail.com

lib/nolibc/Makefile.uk
```diff
@@ -33,6 +33,7 @@ LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/stdio.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/ctype.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/stdlib.c
 LIBNOLIBC_SRCS-y += $(LIBNOLIBC_BASE)/string.c
+LIBNOLIBC_STRING_FLAGS-$(CONFIG_OPTIMIZE_PERF) += -O3
```
skuenzer, line 36: Is gcc picking up `-O3` although `-O2` is set globally by...? https://github.com/unikraft/unikraft/blob/5f534c963744fe66db10dcd4fda8e82feba9c2e7/Makefile.uk#L59

andreittr: As long as the `-O3` flag is set later in the cmdline than `-O2`, it takes precedence and overrides. Quoting from `man gcc` (I'm sure clang behaves similarly):
```
 If you use multiple -O options, with or without level numbers, the last such option is the one that is effective.
```
Looking at the global `Makefile.rules`, it seems that we always place lib-specific & source-specific CFLAGS after global ones.

skuenzer: Sounds good 👍🏻 . Thanks for double-checking.

skuenzer (approved):

PR #1462, opened 2024-06-21
lib/vfscore: Change faccessat with AT_SYMLINK_NOFOLLOW

robertZamfir601:
* Change `faccessat` with `AT_SYMLINK_NOFOLLOW`.
* Move the path creation and logic before the `AT_SYMLINK_NOFOLLOW` (to avoid duplicate code)
* Apply `AT_SYMLINK_NOFOLLOW` only if the file pointed by the path is an actual symlink
* I didn't implement the `don't follow link logic`

### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Change `faccessat` with `AT_SYMLINK_NOFOLLOW`:
* Move the path creation and logic before the `AT_SYMLINK_NOFOLLOW` (to avoid duplicate code)
* Apply `AT_SYMLINK_NOFOLLOW` only if the file pointed by the path is an actual symlink
* I didn't implement the `don't follow link logic`

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -2061,37 +2061,47 @@ UK_SYSCALL_R_DEFINE(int, access, const char*, pathname, int, mode)
 
 UK_SYSCALL_R_DEFINE(int, faccessat, int, dirfd, const char*, pathname, int, mode, int, flags)
 {
-	if (flags & AT_SYMLINK_NOFOLLOW) {
-		UK_CRASH("UNIMPLEMENTED: faccessat() with AT_SYMLINK_NOFOLLOW");
-	}
-
-	if (pathname[0] == '/' || dirfd == AT_FDCWD) {
-		return uk_syscall_r_access((long) pathname, (long) mode);
-	}
-
+	char tmp[PATH_MAX];
+	const char *p;
 	struct vfscore_file *fp;
-	int error = fget(dirfd, &fp);
-	if (error) {
-		goto out_error;
+	int error;
+
+	if (pathname[0] == '/' ||
+	    dirfd == AT_FDCWD) {	/* Full path inside pathname */
+		p = pathname;
+	} else { /* Create Relative Path */
+		error = fget(dirfd, &fp);
+		if (error)
+			goto out_error;
+
+		struct vnode *vp = fp->f_dentry->d_vnode;
+
+		vn_lock(vp);
+
+		/* build absolute path */
+		strlcpy(tmp, fp->f_dentry->d_mount->m_path, PATH_MAX);
+		strlcat(tmp, fp->f_dentry->d_path, PATH_MAX);
+		strlcat(tmp, "/", PATH_MAX);
+		strlcat(tmp, pathname, PATH_MAX);
+
+		vn_unlock(vp);
+		fdrop(fp);
+
+		p = tmp;
 	}
 
-	struct vnode *vp = fp->f_dentry->d_vnode;
-	vn_lock(vp);
+	if (flags & AT_SYMLINK_NOFOLLOW) {
+		struct stat st;
 
-	char p[PATH_MAX];
+		error = uk_syscall_r_lstat((long)p, (long)&st);
+		/* Check if the file is an actual symlink */
+		if (error == 0 && S_ISLNK(st.st_mode))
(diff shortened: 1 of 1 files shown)


StefanJum:
@robertZamfir601 does this work? Is there a reason why it's still a draft?

robertZamfir601:
It is ready for review.

Mihnea0Firoiu (approved):
Looks good!

StefanJum (changes requested):
@robertZamfir601 just to make sure, the actual difference in functionality is that it won't crash if the target is not a symlink, right?

Looks fine, but please remove the `COMMIT_LOG_LONG_LINE` checkpatch ignore and fix the commit description, there is no need to keep the ling lines.

robertZamfir601:
@StefanJum Yes, the only difference is that on a non-symlink target, the `AT_SYMLINK_NOFOLLOW` flag is ignored and the app won't crash.
I changed the description and removed `COMMIT_LOG_LONG_LINE` as well.

StefanJum (changes requested):
One more thing, please use your full name for the author and sign-off message (as in `Robert Zamfir <georobi.016@gmail.com>`). You are currently using `Robert` for both the author field and for the sign-off, and `Robert Zamfir` for the commiter.
Use `git config --global user.name`.

robertZamfir601:
Fixed the names

StefanJum (approved):
All good, thanks.

lib/vfscore/main.c
```diff
@@ -2061,37 +2061,44 @@ UK_SYSCALL_R_DEFINE(int, access, const char*, pathname, int, mode)
...
+	char p[PATH_MAX];
+	struct vfscore_file *fp;
+	int error;
 
-	if (pathname[0] == '/' || dirfd == AT_FDCWD) {
-		return uk_syscall_r_access((long) pathname, (long) mode);
-	}
+	if (pathname[0] == '/' ||
+	    dirfd == AT_FDCWD) {	/* Full path inside pathname */
+		strlcpy(p, pathname, PATH_MAX);
```
andreittr: One minor comment: your change causes `faccessat` to always copy the path to a stack temporary even if it's a full (absolute) path that you could pass on to other functions as-is.
A more economical solution would be something like
```c
char tmp[PATH_MAX];
const char *p;

if (pathname[0] == '/' || dirfd == AT_FDCWD) {
    p = pathname;
} else {
    /* strcat the path together into tmp */
    p = tmp;
}
```
then use `p` as usual.
Or, if you're feeling particularly clever:
```c
const int fullpath = pathname[0] == '/' || dirfd == AT_FDCWD;
char tmp[fullpath ? PATH_MAX : 1];
const char *p;

if (fullpath) {
    p = pathname;
} else {
    /* build path in tmp */
    p = tmp;
}
```
in order to save stack space when not needing `tmp`.

robertZamfir601: First Issue: You are right. I changed the code exactly as you said.
Second Part: I moved the declaration of `tmp` inside the `else` branch. The compiler should be clever with it. I hope it is ok.

andreittr: IIRC it is undefined behavior (or at the very least _very_ non-standard-C) to leak the address of a local variable outside of its scope -- that's explicitly why I didn't propose declaring `tmp` in the else clause, because you need it in the top-level function block.

If the 2nd option looks too hacky declaring `tmp` with a runtime size, you can leave it always at `PATH_MAX` and rely on the compiler being clever enough to not actually do a `sub %rsp` unless `tmp` is used. That at least will be well-defined behavior.

robertZamfir601: I understand. I moved the declaration back to the top-level, with `PATH_MAX` size.

andreittr: Changes look good, thanks! Could you please squash them into the first commit and force-push the PR branch? Then I can give approval and we can merge.

As a policy we try to keep the commit history clean and reflecting self-contained changes, rather than tracking every adjustment during review -- so force-pushes are the common way to address comments.

robertZamfir601: Squashed everything together. Thank you!

andreittr (approved):

PR #1463, opened 2024-06-23
plat/kvm: Add Multiboot2 support

mariapana:
This pull request introduces support for the Multiboot2 protocol, a successor to the original Multiboot variant.

### Prerequisite checklist

### Base target

 - Architecture(s): x86_64
 - Platform(s): kvm
 - Application(s): N/A

### Additional configuration

Since QEMU currently doesn't natively support the Multiboot2 bootloader format, options like `-kernel` are not directly applicable. To test with Multiboot2 images, you'll need to create `.iso` files using the `mkukimg` script. For a detailed example of how to create and boot a Multiboot2 ISO image using QEMU, refer to this [`nginx` sample](https://gist.github.com/mariapana/90ab2e61715e812def0c7582c9482626).

### Description of changes

* Add Multiboot2 header support to ELF64 binaries
* Integrate Multiboot2 support in KVM platform by adding `multiboot2.h`, `multiboot2.c` and modifying `multiboot.S`
* Support building Multiboot2 images by updating the `mkukimg` script
* Remove `ELF64_TO_32` for the Multiboot protocol


plat/kvm/Config.uk
```diff
@@ -35,6 +35,12 @@ config KVM_BOOT_PROTO_MULTIBOOT
 	help
 		Multiboot Boot Protocol Version 1
 
+config KVM_BOOT_PROTO_MULTIBOOT2
+	bool "Multiboot2"
+	depends on KVM_VMM_QEMU && !KVM_VMM_FIRECRACKER && ARCH_X86_64
```
mogasergiu: Drop ` KVM_VMM_QEMU && !KVM_VMM_FIRECRACKER`, it is a generic x86 boot protocol, platform agnostic.

plat/kvm/Linker.uk
```diff
@@ -3,6 +3,9 @@ ifeq ($(CONFIG_KVM_BOOT_PROTO_MULTIBOOT),y)
 KVM_LDFLAGS-y += -Wl,-m,elf_x86_64
 KVM_LDFLAGS-y += -Wl,--entry=_multiboot_entry
 ELF64_TO_32 = y
+else ifeq ($(CONFIG_KVM_BOOT_PROTO_MULTIBOOT2),y)
```
mogasergiu: This looks like it could have been a simple `||` instead of an `else if`.
```C
ifeq ($(findstring y,$(CONFIG_KVM_BOOT_PROTO_MULTIBOOT) $(CONFIG_KVM_BOOT_PROTO_MULTIBOOT2)),y)
```

plat/kvm/Linker.uk
```diff
@@ -81,6 +84,9 @@ $(KVM_IMAGE): $(KVM_IMAGE).dbg
 ifeq ($(ELF64_TO_32),y)
```
mogasergiu: Ughh... this looks so ugly and inconsistent now... I think I am going to ask you to replace this with a `ifeq ($(CONFIG_KVM_BOOT_PROTO_MULTIBOOT),y)`... I know this is out of this PR's scope but at the time when I first added this I thought it would end up being used by other protocols as well... yeah :sweat_smile:

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,381 @@
...
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL ANY
+ *  DEVELOPER OR DISTRIBUTOR BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#ifndef MULTIBOOT_HEADER
```
mogasergiu: Mimic out multiboot header as well, use guards like these
```C
#ifndef __MULTIBOOT2_H__
#define __MULTIBOOT2_H__
```

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,381 @@
...
+#define MULTIBOOT_HEADER 1
+
+/*  How many bytes from the start of the file we search for the header. */
+#define MULTIBOOT_SEARCH 32768
```
mogasergiu: Can you restyle all of the macro's here so that their definition is separated by several tabs, similar to `multiboot.h`. At the moment it is just using one whitespace.

plat/kvm/x86/multiboot.S
```diff
@@ -38,7 +38,11 @@
 #include <uk/asm.h>
 #include <uk/reloc.h>
 
+#ifdef CONFIG_KVM_BOOT_PROTO_MULTIBOOT
```
mogasergiu: We tend to use `#if` instead of `#ifdef` for boolean configurations. So I expect this to be like
```C
#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
#include <kvm-x86/multiboot.h>
#elif  CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
#include <kvm-x86/multiboot2.h>
#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT && !CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
#error "Unknown multiboot version"
#endif
```

plat/kvm/x86/multiboot.S
```diff
@@ -48,9 +52,15 @@
 lcpu_bootstack:
 
 .section .rodata
+#ifdef CONFIG_KVM_BOOT_PROTO_MULTIBOOT
```
mogasergiu: Ditto, but also, notice how you only need to have the entry symbol defined differently, not the boostack or the startup args symbols
```C
lcpu_boot_startup_args:
#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
	ur_data	quad, multiboot_entry, 8
#elif  CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
	ur_data	quad, multiboot_entry2, 8
#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT && !CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
#error "Unknown multiboot version"
#endif
	ur_data	quad, lcpu_bootstack, 8
```

plat/kvm/x86/multiboot.S
```diff
@@ -63,8 +73,11 @@ lcpu_boot_startup_args:
 .code32
 .section .text.32.boot
 ENTRY(_multiboot_entry)
-	/* only multiboot is supported for now */
-	cmpl	$MULTIBOOT_BOOTLOADER_MAGIC, %eax
+	#ifdef CONFIG_KVM_BOOT_PROTO_MULTIBOOT
```
mogasergiu: Ditto let's define a wrapper macro somewhere in the beginning here like so and use that
```C
#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
#define MULTIBOOT_MAGIC MULTIBOOT_BOOTLOADER_MAGIC
#elif  CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
#define MULTIBOOT_MAGIC MULTIBOOT_BOOTLOADER_MAGIC2
#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT && !CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
#error "Unknown multiboot version"
#endif
```
Also, do not indent `#ifdef`'s or `#define`'s

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,208 @@
...
+{
+	struct multiboot_tag_mmap *mmap_tag = (struct multiboot_tag_mmap *)tag;
+
+	struct multiboot_mmap_entry *mmap = mmap_tag->entries;
+	struct ukplat_memregion_desc mrd = {0};
+	__paddr_t start, end;
+	__sz offset;
+
+	__sz mmap_entry_cnt = (mmap_tag->size - sizeof(struct multiboot_tag_mmap)) /
+			    		 mmap_tag->entry_size;
```
mogasergiu: Misaligned with opened paranthesis from the line above (which goes past 80 characters).

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,208 @@
...
+
+	for (offset = 0; offset < mmap_entry_cnt;
+		offset++) {
```
mogasergiu: This is also misaligned but it will be unproductive if I keep spamming for each single coding style issue. I would advise using `support/scripts/checkpatch.pl --strict` on your patches and solve the warnings/errors.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,208 @@
...
+	/* Ensure that the memory map contains the legacy high mem area */
+	rc = ukplat_memregion_list_insert_legacy_hi_mem(&bi->mrds);
+	if (unlikely(rc))
+		multiboot2_crash("Could not insert legacy memory region", rc);
+
+	/* Parse Multiboot2 tags */
+	for (struct multiboot_tag *tag = (struct multiboot_tag *)(addr + 8);
+	     tag->type != MULTIBOOT_TAG_TYPE_END;
+	     tag = (struct multiboot_tag *)((multiboot_uint8_t *)tag
+					    + ((tag->size + 7) & ~7))) {
```
mogasergiu: I will want some comment or macro surrounding this as it does not tell me much about this formula.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,266 @@
...
+            new_shdr[0:ELF64_SHDR_LEN] = shdr[0:ELF64_SHDR_LEN] # keep the rest
+            new_shdr[24:32] = self.prpnd_off(shdr[24:32], mb2_header_end_off) # sh_offset
+
+            return new_shdr
+
+        return [update_shdr(s) for s in self.elf64_shdrs]
+
+def add_tags(tags_opt):
+    tags_hdr = b""
+    for tag_type in tags_opt:
```
mogasergiu: Nice! But instead of `elif`, let's just have an `else` that throws an exception for unknown tag.

mogasergiu (changes requested):
Ok left some comments regarding some styling issues. There are other places I could've left comments at but it would have been unproductive as these can simply be solved by a `checkpatch.pl --strict <your patch files here>`.

So for now I will ask that you update accordingly with these and have the following commit history schema:
- first commit adds the script
- second commit adds the rest

So at the end you should only have two commits for now. The relocation tag and whatever we implement later will come separately.

cristian-vijelie (approved):

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+            return new_shdr
+
+        return [update_shdr(s) for s in self.elf64_shdrs]
+
+
+def add_tags(tags_opt):
+    tags_hdr = b""
+    for tag_type in tags_opt:
+        if (
+            tag_type == MULTIBOOT_HEADER_TAG_END
```
mogasergiu: uhh, this `if` does not look that pythonesque to me :sweat_smile:

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+            or tag_type == MULTIBOOT_HEADER_TAG_MODULE_ALIGN
+        ):
+            size = 8
```
mogasergiu: what is this `size` and does it need to be assigned in each iteration? seems pretty constant to me.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+    mb2_architecture = MULTIBOOT_ARCHITECTURE_I386
+
+    tags_opt = [MULTIBOOT_HEADER_TAG_MODULE_ALIGN, MULTIBOOT_HEADER_TAG_END]
+    tags_hdr = add_tags(tags_opt)
+    tags_size = len(tags_hdr)
+
+    mb2_header_length = MULTIBOOT_HEADER_BASE_SIZE + tags_size
+    # We need to convert checksum to 4-byte signed integer range to avoid overflow
+    mb2_checksum = -(mb2_magic + mb2_architecture + mb2_header_length)
+    mb2_checksum = (mb2_checksum + 2 ** 32) % 2 ** 32
```
mogasergiu: Maybe surround `2 ** 32` with paranthesis as well? Just to be sure.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+
+        return [update_phdr(p) for p in self.elf64_phdrs]
+
+    def get_shdrs(self, mb2_header_end_off):
+        def update_shdr(shdr):
+            new_shdr = bytearray(ELF64_SHDR_LEN)
+
+            new_shdr[0:ELF64_SHDR_LEN] = shdr[0:ELF64_SHDR_LEN]  # keep the rest
+            new_shdr[24:32] = self.prpnd_off(
+                shdr[24:32], mb2_header_end_off
```
mogasergiu, line 138: keep these on the same line as the function call? just like you did for phdrs.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+        return new_off.to_bytes(sz, self.endian)
+
+    def get_ehdr(self, mb2_header_end_off):
+        elf64_ehdr = bytearray(ELF64_EHDR_LEN)
+        elf64_ehdr[0:ELF64_EHDR_LEN] = self.elf64_ehdr[0:ELF64_EHDR_LEN]
+
+        # The offset to the ELF64 Program/Sections Headers are at
+        # the end of the Multiboot Header and the end of the file respectively,
+        # so that we do not mess up the Multiboot2 header, which must exist
+        # in the first 8192 bytes. Although the specification does not enforce
```
mogasergiu: Perhaps mention that the spec mandates that it is in the first 32768 bytes though. This seems to be a GRUB quirk only.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,245 @@
...
+        # Save original ELF content
+        elf64_buf = new_elf.read()
+
+        # Go back to the beginning
+        new_elf.seek(0)
+
+        # Write ELF64 header with updated offsets
+        new_elf.write(elf64_ehdr)
+
+        # Append Multiboot2 header
```
mogasergiu, line 213: > The Multiboot2 header must be contained completely within the first 32768 bytes of the OS image, and must be 64-bit aligned.

The same way `multiboot.py` checks for alignment
```python
        # Append the Multiboot header
        # NOTE: Multiboot header must be 4-byte aligned - luckily that is the
        # case with the size of the ELF32 header (40 bytes)
        if elf32.tell() & 0b11:
            raise Exception("ELF32 header is not a multiple of 4 bytes in size")
```
You should do the same here.

plat/kvm/Linker.uk
```diff
@@ -1,8 +1,7 @@
 ifeq (x86_64,$(CONFIG_UK_ARCH))
-ifeq ($(CONFIG_KVM_BOOT_PROTO_MULTIBOOT),y)
+ifeq ($(findstring y,$(CONFIG_KVM_BOOT_PROTO_MULTIBOOT) $(CONFIG_KVM_BOOT_PROTO_MULTIBOOT2)),y)
 KVM_LDFLAGS-y += -Wl,-m,elf_x86_64
 KVM_LDFLAGS-y += -Wl,--entry=_multiboot_entry
-ELF64_TO_32 = y
```
mogasergiu, line 5: Since you removed this here, you should remove it below as well.

plat/kvm/Linker.uk
```diff
@@ -81,6 +80,9 @@ $(KVM_IMAGE): $(KVM_IMAGE).dbg
 ifeq ($(ELF64_TO_32),y)
```
mogasergiu: As in here :point_up_2: This should now be `CONFIG_KVM_BOOT_PROTO_MULTIBOOT` instead.
Which makes me realize I forgot about this ELF64_TO_32 thingy
Now I think the PR should have 4 commits:
- the commit that removes this ELF64_TO_32  and replaces its usecase with the check for `CONFIG_KVM_BOOT_PROTO_MULTIBOOT` instead
- ... the other 3 you have right now

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,381 @@
...
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL ANY
+ *  DEVELOPER OR DISTRIBUTOR BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
```
mogasergiu: Misalignment?

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,381 @@
...
+};
+
+struct multiboot_tag_load_base_addr {
+	multiboot_uint32_t type;
+	multiboot_uint32_t size;
+	multiboot_uint32_t load_base_addr;
+};
+
+#endif /*  ! __ASSEMBLY__ */
+
```
mogasergiu: Surplus newline

plat/kvm/x86/multiboot.S
```diff
@@ -38,7 +38,23 @@
 #include <uk/asm.h>
 #include <uk/reloc.h>
 
+#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
 #include <kvm-x86/multiboot.h>
+#elif CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
+#include <kvm-x86/multiboot2.h>
+#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT &&
+	!CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
```
mogasergiu: Align the `!`

plat/kvm/x86/multiboot.S
```diff
@@ -38,7 +38,23 @@
...
+#error "Unknown multiboot version"
+#endif
```
mogasergiu: Repeat the
```C
/* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT &&
	!CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
```

plat/kvm/x86/multiboot.S
```diff
@@ -38,7 +38,23 @@
...
+
+#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
+#define MULTIBOOT_MAGIC MULTIBOOT_BOOTLOADER_MAGIC
+#elif  CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
+#define MULTIBOOT_MAGIC MULTIBOOT2_BOOTLOADER_MAGIC
+#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT &&
+	!CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
+#error "Unknown multiboot version"
+#endif
```
mogasergiu: Ditto

plat/kvm/x86/multiboot.S
```diff
@@ -49,7 +65,14 @@ lcpu_bootstack:
...
 lcpu_boot_startup_args:
+#if CONFIG_KVM_BOOT_PROTO_MULTIBOOT
 	ur_data	quad, multiboot_entry, 8
+#elif  CONFIG_KVM_BOOT_PROTO_MULTIBOOT2
+	ur_data	quad, multiboot2_entry, 8
+#else /* !CONFIG_KVM_BOOT_PROTO_MULTIBOOT &&
+	!CONFIG_KVM_BOOT_PROTO_MULTIBOOT2 */
+#error "Unknown multiboot version"
+#endif
 	ur_data	quad, lcpu_bootstack, 8
```
mogasergiu, line 79: Ditto

plat/kvm/x86/multiboot.S
```diff
@@ -63,8 +86,7 @@ lcpu_boot_startup_args:
 .code32
 .section .text.32.boot
 ENTRY(_multiboot_entry)
-	/* only multiboot is supported for now */
-	cmpl	$MULTIBOOT_BOOTLOADER_MAGIC, %eax
+	cmpl $MULTIBOOT_MAGIC, %eax
```
mogasergiu: Tab between instruction and first operand

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Multiboot entry point called after lcpu initialization. We enter with the
+ * 1:1 boot page table set. Physical and virtual addresses thus match for all
+ * regions in the mapped range.
+ */
+void multiboot2_entry(struct lcpu *lcpu, unsigned long addr)
+{
+	struct ukplat_bootinfo *bi;
```
mogasergiu: I have a personal preference when it comes to variable declaration. I will only mention it once, feel free to disregard/apply it if you also dislike/like it. I generally declare these in reverse xmas tree order...
e.g. https://github.com/unikraft/unikraft/blob/staging/plat/kvm/efi.c#L299
Or pretty much how that whole file does it. There are many other files that follow these patterns. Though this is not mandated by our coding style (though I hope it will) and others do not really use this pattern, I personally like it better.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+	/* Ensure that the memory map contains the legacy high mem area */
+	rc = ukplat_memregion_list_insert_legacy_hi_mem(&bi->mrds);
+	if (unlikely(rc))
+		multiboot2_crash("Could not insert legacy memory region", rc);
+
+	/**
+	 * Initialize the first tag address, skipping total_size(4) and
+	 * reserved (4) fields of the boot information structure
+	 */
+	struct multiboot_tag *tag = (struct multiboot_tag *)(addr + 8);
```
mogasergiu: Try to declare things at the beginning of the function. Keep the styling consistent with the codebase. That being said you could make this assignment in the for loop.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+#include <uk/reloc.h>
+#include <kvm-x86/multiboot2.h>
+
+#include <errno.h>
+#include <string.h>
+
+#define multiboot2_crash(msg, rc)	ukplat_crash()
+
+/* Round up 'x' to the nearest multiple of 'MULTIBOOT_TAG_ALIGN' */
+#define ROUND_UP_TO_ALIGNMENT(x) \
```
mogasergiu: For this kind of alignment macro's we use names like `ALIGN_UP` so I would rename this to `MULTIBOOT_TAG_ALIGN_UP` and have it be defined as
```C
#define MULTIBOOT_TAG_ALIGN_UP(x)                      ALIGN_UP((x), MULTIBOOT_TAG_ALIGN)
```
This way you use the already existing macro.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+	/**
+	 * Initialize the first tag address, skipping total_size(4) and
+	 * reserved (4) fields of the boot information structure
+	 */
+	struct multiboot_tag *tag = (struct multiboot_tag *)(addr + 8);
+
+	/* Parse Multiboot2 tags */
+	for (; tag->type != MULTIBOOT_TAG_TYPE_END;
+	     tag = (struct multiboot_tag *)((multiboot_uint8_t *)tag
+					+ ROUND_UP_TO_ALIGNMENT(tag->size))) {
```
mogasergiu: Yeah this will look super ugly as a for loop given how long these definitions are... maybe let's do it as a while loop instead? :)

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+
+			mrd.pbase = PAGE_ALIGN_DOWN(mod_tag->mod_start);
+			mrd.vbase = mrd.pbase;
+			mrd.pg_off = mod_tag->mod_start - mrd.pbase;
+			mrd.len = mod_tag->mod_end - mod_tag->mod_start;
+			mrd.pg_count = PAGE_COUNT(mrd.pg_off + mrd.len);
+			mrd.type = UKPLAT_MEMRT_INITRD;
+			mrd.flags = UKPLAT_MEMRF_READ;
+
+			#ifdef CONFIG_UKPLAT_MEMRNAME
```
mogasergiu: Same commen as last time about not indenting `#ifdef`'s and using `#if`s instead. The `strncoy` looks overindented as well.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,180 @@
...
+				strncpy(mrd.name,
+					(char *)(__uptr)mod_tag->cmdline,
+					sizeof(mrd.name) - 1);
+			#endif /* CONFIG_UKPLAT_MEMRNAME */
+
+			mrd_insert(bi, &mrd);
+			break;
+		case MULTIBOOT_TAG_TYPE_MMAP:
+			struct multiboot_tag_mmap *mmap_tag;
```
mogasergiu: No late declarations, do all declarations at the beginning. And this switch case is kind of fat, I would move it to its own function. This way you get to keep your declarations :P

support/scripts/mkukimg
```diff
@@ -73,6 +73,41 @@ EOF
 	fi
 }
 
+mkukmb2iso()
```
mogasergiu, line 76: All I have to say for this commit is to just name it `support/scripts: Support building multiboot2 images` as far as its commit subject goes. The fact that you did this minor rename for the old multiboot option is a small detail that can be mentioned in the main commit message only as you have already done.

mogasergiu (changes requested):

support/scripts/multiboot2.py
mariapana: The `size` field is the total size the tag will have with all of its fields. In the case of the implemented tags, it is indeed equal to 8. However, there are other tags whose sizes can differ, as stated in the specification  (e.g. `MULTIBOOT_HEADER_TAG_RELOCATABLE`  would have `size` equal to 24). Should I use a macro for now and change it up once more tags are added?

mogasergiu: Understood. There can only be one instance of a tag type correct? I am thinking maybe we should just create an individual function for each. E.g.
```python
def get_multiboot2_hdr():
    mb2_magic = MULTIBOOT_HEADER_MAGIC
    mb2_architecture = MULTIBOOT_ARCHITECTURE_I386

    tags_hdr = multiboot2_tag_align()
    tags_hdr += multiboot2_tag_end()
    tags_size = len(tags_hdr)
    ....
 ```
 instead of the `add_tags` function. So when we would add a certain tag in the future we would create a `multiboot2_tag_<tag_name>(<potential_args>)`, where potential_args would be possible values that may change the behavior of the tag, e.g. `multiboot2_tag_reloc(min_addr, max_addr, align, preference)`.
 
Sorry, when I first made the suggestion, I thought we could like make a very generic class structure or something to generically instantiate tags, but now that I see them properly in the spec they can differ quite a log, some not even having the `flags` field... it may not be worth the future maintenance effort... I think it would be easier to just have a function that returns a tag for type of tag.
What do you think?

Comment (mariapana) [reply support/scripts/multiboot2.py]: Yes, it would be better and cleaner this way. It could get complicated/ugly if complex tags are added down the road. I will update the implementation :smiley:

File: plat/kvm/Linker.uk (hunk 19 above)
Comment (mogasergiu) [inline plat/kvm/Linker.uk:5]: I think you misunderstood me. What I meant was having the following git commit history
- commit that removes this ELF64_TO_32 and replaces its if below with the MULTIBOOT config (mb2 support not in yet
- commit that introduces the script
- commit that adds multiboot2 support (C/asm changes and the addition to Linker.uk of multiboot2.py)
- commit that adds mb2 support and changes the function names in the mkukimg script.

File: support/scripts/multiboot2.py [hunk 35]
```diff
@@ -0,0 +1,247 @@
...
+# Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+# Licensed under the BSD-3-Clause License (the "License").
+# You may not use this file except in compliance with the License.
+
+import os
+import argparse
+from struct import unpack
+
+# Default global values
+ELF_MAGIC = {
```
mogasergiu: I don't see this being used anywhere, so maybe let's remove it.

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,247 @@
...
+    "x86_64": b"\x7fELF\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00",
+}
+
+ELF_MACHINE = {
+    "i386": 3,  # EM_386
+    "x86_64": 62,  # EM_X86_64
+}
+
+ELF64_EHDR_LEN = 64
+
```
mogasergiu: Can you remove these new lines between these declarations?

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,247 @@
...
+MULTIBOOT_HEADER_BASE_SIZE = 16
+
+MULTIBOOT_HEADER_MAGIC = 0xE85250D6
+MULTIBOOT_ARCHITECTURE_I386 = 0
+MULTIBOOT_LOAD_PREFERENCE_LOW = 1
+PAGE_SIZE = 4096
+
+MULTIBOOT_HEADER_TAG_OPTIONAL = 1
+
+MULTIBOOT_HEADER_TAG_END = 0
```
mogasergiu, line 34: Let's only declare the tags that are used. And for those that are used give a brief description/comment for what they are used for and if there is any special treatment from GRUB2 worth mentioning (that is not mandated by the spec).

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,247 @@
...
+            new_shdr[0:ELF64_SHDR_LEN] = shdr[0:ELF64_SHDR_LEN]  # keep the rest
+            new_shdr[24:32] = self.prpnd_off(shdr[24:32], mb2_header_end_off)  # sh_offset
+
+            return new_shdr
+
+        return [update_shdr(s) for s in self.elf64_shdrs]
+
+
+def multiboot2_tag_end():
+    size = 8
```
mogasergiu: Maybe let's provide an endianness argument to these tag function.
So like let's maybe go with
```python
def multiboot2_tag_module_align(endian):
    return (
        MULTIBOOT_HEADER_TAG_MODULE_ALIGN.to_bytes(2, endian)
        + MULTIBOOT_HEADER_TAG_OPTIONAL.to_bytes(2, endian)
        + (8).to_bytes(4, endian)
    )
```
Ditto for `get_multiboot2_hdr` and let's let the caller get the endianness from an ELF editor getter like `elf64.get_endian()`.

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,380 @@
...
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL ANY
+ *  DEVELOPER OR DISTRIBUTOR BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ *  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ *  DEALINGS IN THE SOFTWARE.
+ */
+#ifndef __MULTIBOOT2_H__
+#define __MULTIBOOT2_H__
+
+/*  How many bytes from the start of the file we search for the header. */
```
mogasergiu, line 25: I see all comments have a two whitespace after `/*`. I guess this is from the original file. Let's remove them to comply with our coding style.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,188 @@
...
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, mrd);
+	if (unlikely(rc < 0))
+		multiboot2_crash("Cannot insert bootinfo memory region", rc);
+}
+
+void handle_mmap_tag(struct ukplat_memregion_desc mrd,
+		     struct ukplat_bootinfo *bi, struct multiboot_tag *tag)
+{
+	__sz i, mmap_total_sz, mmap_entry_cnt;
```
mogasergiu, line 44: :muscle: :christmas_tree:

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,188 @@
...
+	 * Initialize the first tag address, skipping total_size(4) and
+	 * reserved (4) fields of the boot information structure
+	 */
+	tag = (struct multiboot_tag *)(addr + 8);
+
+	/* Parse Multiboot2 tags */
+	while (tag->type != MULTIBOOT_TAG_TYPE_END) {
+		switch (tag->type) {
+		case MULTIBOOT_TAG_TYPE_CMDLINE:
+			struct multiboot_tag_string *tag_str;
```
mogasergiu: Can you also declare this at the beginning of the function?

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,188 @@
...
+
+			tag_str = (struct multiboot_tag_string *)tag;
+			cmdline_str = (__u64)tag_str->string;
+			cmdline_len = tag_str->size;
+			break;
+		case MULTIBOOT_TAG_TYPE_MODULE:
+			struct multiboot_tag_module *mod_tag;
```
mogasergiu: Ditto

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,188 @@
...
+
+	if (unlikely(mrd->len == 0))
+		return;
+
+	rc = ukplat_memregion_list_insert(&bi->mrds, mrd);
+	if (unlikely(rc < 0))
+		multiboot2_crash("Cannot insert bootinfo memory region", rc);
+}
+
+void handle_mmap_tag(struct ukplat_memregion_desc mrd,
```
mogasergiu: What is the `mrd` argument for? I think you may as well declare it on this function's frame.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,188 @@
...
+		mrd.pbase = PAGE_ALIGN_DOWN(start);
+		mrd.vbase = mrd.pbase; /* 1:1 mapping */
+		mrd.pg_off = start - mrd.pbase;
+		mrd.len = end - start;
+		mrd.pg_count = PAGE_COUNT(mrd.pg_off + mrd.len);
+
+		if (m->type == MULTIBOOT_MEMORY_AVAILABLE) {
+			mrd.type  = UKPLAT_MEMRT_FREE;
+			mrd.flags = UKPLAT_MEMRF_READ |
+					UKPLAT_MEMRF_WRITE;
```
mogasergiu: Misaligned with `=`

mogasergiu (changes requested):

plat/kvm/x86/multiboot2.c
mariapana: You mean for the function parameters as well or have I misunderstood how it should be done for the declarations starting from line 44?

mogasergiu: Ah no lol just appreciating that you adopted the style :laughing:

mariapana: Yeah, it's much better this way. Looking back on it, I don't know how I didn't do it like this before :)

mogasergiu: Eh, it's just a matter of style :smile:. Unless mandated by the coding style policy, feel free to do things your own way. I just like this style visually. Got it from the Linux Kernel as they have started adopting this on the mailing lists as well.

plat/kvm/Linker.uk
```diff
@@ -103,4 +102,4 @@ UK_IMAGES-$(CONFIG_OPTIMIZE_COMPRESS) += $(KVM_IMAGE).gz
 endif
 
 # ...for cleaning:
-LIBKVMPLAT_CLEAN += $(call build_clean,$(KVM_DEBUG_IMAGE).bootinfo)
+LIBKVMPLAT_CLEAN += $(call build_clean,$(KVM_DEBUG_IMAGE).bootinfo)
```
mogasergiu: Pretty sure this should not appear in the diff :)

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,241 @@
...
+                shdr[24:32], mb2_header_end_off
+            )  # sh_offset
+
+            return new_shdr
+
+        return [update_shdr(s) for s in self.elf64_shdrs]
+
+    def get_endian(self):
+        return self.endian
+
```
mogasergiu, line 147: Double newlines?
(Ditto for below)

support/scripts/multiboot2.py
```diff
@@ -0,0 +1,241 @@
...
+
+def multiboot2_tag_end(endian):
+    size = 8
+    return (
+        MULTIBOOT_HEADER_TAG_END.to_bytes(2, endian)
+        + MULTIBOOT_HEADER_TAG_OPTIONAL.to_bytes(2, endian)
+        + size.to_bytes(4, endian)
```
mogasergiu: Why not just `(8).to_bytes(4, endian)`?
Also just noticed now, but operand should typically come at the end of the line you are continuing on a next line. So I was expecitng something like
```python
return (MULTIBOOT_HEADER_TAG_END.to_bytes(2, endian) + 
        MULTIBOOT_HEADER_TAG_OPTIONAL.to_bytes(2, endian) +
        (8).to_bytes(4, endian)
```

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,380 @@
...
+#define MULTIBOOT2_BOOTLOADER_MAGIC				0x36d76289
+
+/*  Alignment of multiboot modules. */
+#define MULTIBOOT_MOD_ALIGN					0x00001000
+
+/*  Alignment of the multiboot info structure. */
+#define MULTIBOOT_INFO_ALIGN					0x00000008
+
+/*  Flags set in the ’flags’ member of the multiboot header. */
+
```
mogasergiu: Why the newline?

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,380 @@
...
+#define MULTIBOOT_HEADER_TAG_OPTIONAL				1
+
+#define MULTIBOOT_LOAD_PREFERENCE_NONE				0
+#define MULTIBOOT_LOAD_PREFERENCE_LOW				1
+#define MULTIBOOT_LOAD_PREFERENCE_HIGH				2
+
+#define MULTIBOOT_CONSOLE_FLAGS_CONSOLE_REQUIRED		1
+#define MULTIBOOT_CONSOLE_FLAGS_EGA_TEXT_SUPPORTED		2
+
+#ifndef __ASSEMBLY__
```
mogasergiu: `#if !__ASSEMBLY__`

plat/kvm/include/kvm-x86/multiboot2.h
```diff
@@ -0,0 +1,380 @@
...
+	multiboot_uint64_t pointer;
+};
+
+struct multiboot_tag_load_base_addr {
+	multiboot_uint32_t type;
+	multiboot_uint32_t size;
+	multiboot_uint32_t load_base_addr;
+};
+
+#endif /*  ! __ASSEMBLY__ */
```
mogasergiu: Why the two whitespaces before the `!`? :thinking:

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,185 @@
...
+	struct multiboot_tag_mmap *mmap_tag;
+	struct multiboot_mmap_entry *mmap;
+	multiboot_memory_map_t *m;
+	__paddr_t start, end;
+
+	mmap_tag = (struct multiboot_tag_mmap *)tag;
+	mmap = (struct multiboot_mmap_entry *)mmap_tag->entries;
+
+	mmap_total_sz = (mmap_tag->size -
+			 sizeof(struct multiboot_tag_mmap));
```
mogasergiu: Maybe `mmap_total_sz = (mmap_tag->size - sizeof(*mmap_tag))`? Typically it is better to use the actual variable whose type size you are trying to find out. This way, on type rename, you need to update less lines (also shorter).

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,185 @@
...
+
+			mrd_insert(bi, &mrd);
+			break;
+		case MULTIBOOT_TAG_TYPE_MMAP:
+			handle_mmap_tag(bi, tag);
+			break;
+		}
+
+		tag = (struct multiboot_tag *)((multiboot_uint8_t *)tag
+					+ MULTIBOOT_TAG_ALIGN_UP(tag->size));
```
mogasergiu: Misaligned? Also `+` on the first line.

plat/kvm/x86/multiboot2.c
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	memcpy(bi->bootprotocol, "multiboot2", sizeof("multiboot2"));
+
+	ukplat_memregion_list_coalesce(&bi->mrds);
+
+	/* Create cmdline mrd after coalescing */
+	char *cmdline = ukplat_memregion_alloc(cmdline_len,
```
mogasergiu: Late declaration :)

mogasergiu (changes requested):
Some more unimportant nitpicks. Anyway, you will need to rebase now that some related PR's (the one I told you about for the cmdline issue) have been merged.

PR #1465, opened 2024-06-25
drivers/ukbus/pci: Map PCI regions at runtime

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series updates `uk_bus_pf_devmap()` to handle large MMIO regions of devices like PCI in a performant way, and updates the PCI driver to map MMIO regions on runtime when paging is enabled.

### DIFF
--- a/drivers/ukbus/pci/arch/arm64/pci_bus.c
+++ b/drivers/ukbus/pci/arch/arm64/pci_bus.c
@@ -51,6 +51,7 @@
  */
 
 #include <string.h>
+#include <uk/errptr.h>
 #include <uk/print.h>
 #include <uk/plat/common/cpu.h>
 #include <uk/bus/pci.h>
@@ -90,6 +91,15 @@ static int arch_pci_driver_add_device(struct pci_driver *drv,
 	memcpy(&dev->id, devid, sizeof(dev->id));
 	memcpy(&dev->addr, addr,  sizeof(dev->addr));
 	dev->drv = drv;
+
+#if CONFIG_PAGING
+	/* TODO Properly query the BAR size */
+	base = uk_bus_pf_devmap(base, __PAGE_SIZE);
+	if (unlikely(PTRISERR(base))) {
+		uk_pr_err("Could not map device (%d)\n", PTR2ERR(base));
+		return PTR2ERR(base);
+	}
+#endif /* CONFIG_PAGING */
 
 	dev->base = base;
 	dev->irq = irq;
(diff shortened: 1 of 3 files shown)


drivers/ukbus/platform/platform_bus.c
```diff
@@ -261,22 +261,33 @@ __vaddr_t uk_bus_pf_devmap(__paddr_t base, __sz size)
...
+	rc = ukplat_page_map(pt, vaddr, paddr, pages, attr, 0);
+	if (unlikely(rc && rc != -EEXIST)) {
+		uk_pr_err("Could not map MMIO Region at 0x%lx - 0x%lx (%d)\n",
+			  base, base + size, rc);
+		return rc;
+	}
 
-		if (rc == -EEXIST)
-			rc = ukplat_page_set_attr(pt, va, 1, attr, 0);
+	if (rc == -EEXIST) {
```
mogasergiu, line 277: Will this still be needed after the ukboot rework? Who maps this zone so early that results in an EEXIST?

drivers/ukbus/pci/arch/arm64/pci_bus.c
```diff
@@ -91,6 +92,12 @@ static int arch_pci_driver_add_device(struct pci_driver *drv,
 	memcpy(&dev->addr, addr,  sizeof(dev->addr));
 	dev->drv = drv;
 
+#if CONFIG_PAGING
+	base = uk_bus_pf_devmap(base, __PAGE_SIZE);
+	if (unlikely(PTRISERR(base)))
```
mogasergiu: Can we print a message?
Also why `__PAGE_SIZE`?

mogasergiu (changes requested):

drivers/ukbus/platform/platform_bus.c
michpappas: At the current state of things probably nothing, but that should be irrelevant. As the function's scope is similar to that of `ioremap()` it should take that case into account. Additionally it may be called by some driver in the future at a later stage of execution where a given region is indeed mapped.

drivers/ukbus/pci/arch/arm64/pci_bus.c
michpappas: > Can we print a message?

Done

> Also why __PAGE_SIZE?

Because rewriting this anyway, and given this drivers (bad) design I think it's not worth to implement proper size querying. I'll add a TODO.

mogasergiu (approved):

razvand (approved):

PR #1468, opened 2024-06-28
.github: Remove CODEOWNERS

jschlumpp:
The goal of the CODEOWNERS file is to automatically assign PRs to the responsible team. However, this notifies the full team instead of individual persons and we manually assign specific persons anyway via a bot. 

After the team was added as a review once, the notifications will be still sent even after the team removal. This causes an immensive amount of noise and many people disable notifications/mails entirely making it necessary to rely on out-of-band channels to notify people of actual review requests.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A


nderjung (commented):
I wonder if simply turning it into a single catch-all for a defined set of custodians/maintainers may be enough to reduce noise but at the same time provide the [relevant branch protection mechanism which GitHub offers through its declaration](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection) since the GitHub organization has nearly 100 members from the open-source community associated and mistakes happen.  It may be that branch protection mechanisms defined non-statically through GitHub repo/org configuration may be enough and just needs to be double checked.

Otherwise, I am in agreement.  This file produces a lot of noise for all maintainers and reducing that will help increase visibility of more relevant issues/PRs incoming.  Additionally, many changes often result in many of these "sub maintainer" groups to be asked to review, here is a recent screenshot:

image

jschlumpp:
> [relevant branch protection mechanism which GitHub offers through its declaration](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection)

As far as I know we don't even use that feature. Restricting the actual merge to a specific group might be the more practical solution to this. (Effectively Razvan and Simon are already the main people responsible for merging). But this can be also done without CODEOWNERS AFAIK, and then that won't spam individual people for *all* PRs.

> I wonder if simply turning it into a single catch-all for a defined set of custodians/maintainers

Then those few people will still get spammed. It does not solve the problem for those. If some central people still have disable GitHub Notifications completely then we still have the problem of having to rely on external channels to progress PRs.

razvand:
I agree with @mschlumpp here. Since this is something we don't actually use, I would just remove it.

nderjung (approved):
> Restricting the actual merge to a specific group might be the more practical solution to this.

This is the bit I am getting at.  After diving a bit deeper into docs about the compatibility with `CODEOWNERS` and rulesets, it's best approached just by using GitHub's native settings as `CODEOWNERS` _cannot_ provide the "infrastructure-as-code" management I was looking for.

Marking with my approval for merge.


razvand (approved):

PR #1471, opened 2024-07-18
lib/uklibparam: Refactor parameter registration

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Remove UK_CTORS registration to allow libukparam parsing to be executed during early init. Parameters now are registerd into a global section similarly to inittab. This additionally fixes a limitation of the previous registration mechanism where all parameters had to be registered from a single file.

### DIFF
--- a/lib/uklibparam/include/uk/libparam.h
+++ b/lib/uklibparam/include/uk/libparam.h
@@ -66,11 +66,10 @@ extern C {
  * A library parameter descriptor (struct uk_libparam_desc) is referencing to
  * the section's base address for iteration.
  */
-#define UK_LIBPARAM_PARAMSECTION_NAMEPREFIX   uk_libparam_params_
 #define UK_LIBPARAM_PARAM_NAMEPREFIX          __uk_libparam_param_
+#define UK_LIBPARAM_PDATA_NAMEPREFIX          __uk_libparam_pdata_
 
-#define UK_LIBPARAM_PARAMSECTION_NAME       \
-	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PARAMSECTION_NAMEPREFIX, __LIBNAME__)
+#define UK_LIBPARAM_PARAMSECTION_NAME       uk_libparam_params
 #define UK_LIBPARAM_PARAMSECTION_STARTSYM   \
 	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PARAMSECTION_NAME, _start)
 #define UK_LIBPARAM_PARAMSECTION_ENDSYM     \
@@ -109,7 +108,15 @@ enum  uk_libparam_param_type {
 /*
  * Library parameter descriptor
  */
+
+struct uk_libparam_pdata { /* Non-const members of `struct uk_libparam_param` */
+	/* Internally used by parser for array parameters */
+	__sz __widx;
+};
+
 struct uk_libparam_param {
+	/* Library prefix */
+	const char * const prefix;
 	const char * const name;
 	const char * const desc;
 	const enum uk_libparam_param_type type;
@@ -117,67 +124,9 @@ struct uk_libparam_param {
 	const __sz count;
 	/* Reference to corresponding variable */
 	void * const addr;
-
-	/* Internally used by parser for array parameters */
-	__sz __widx;
+	/* Non-const members */
+	struct uk_libparam_pdata *pdata;
 };
-
-/*
- * Library descriptor
- */
-struct uk_libparam_libdesc {
-	/* Library prefix */
-	const char *prefix;
-	/* Section header of parameter reference array */
-	struct uk_libparam_param **params;
-	/* Number of parameters */
-	__sz params_len;
-	/* Next libdesc entry */
-	struct uk_list_head next;
-};
-
-/* -------------------------------------------------------------------------- */
-/* Library registration                                                       */
-
-/* The following code declares a library descriptor as soon as this header
- * is included: This makes an explicit registration to libuklibparam
- * unnecessary.
- * FIXME: This allows using of library parameters only in one source file of a
- *        library. A solution could be to create a file in uklibparam that
- *        declares an array of structs, one struct per library. Ideally the
- *        fields are filled out at compile/link time. This can be done as soon
- *        as a library is available that can export the list of included
- *        libraries of a build.
- * NOTE: Double declaration within one source file is avoided by the header's
- *       include guards.
- */
-#define UK_LIBPARAM_LIBDESC \
-	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PARAMSECTION_NAME, _libdesc)
-#define UK_LIBPARAM_LIBDESC_CTOR \
-	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PARAMSECTION_NAME, _ctor)
-
-static struct uk_libparam_libdesc UK_LIBPARAM_LIBDESC = {
-	.prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX)
-
-	/* Rest of fields are initialized with 0 automatically */
-};
-
-void _uk_libparam_libsec_register(struct uk_libparam_libdesc *ld);
-
-static void UK_LIBPARAM_LIBDESC_CTOR(void)
-{
-	UK_LIBPARAM_LIBDESC.params_len =
-		(__sz)((__uptr) &UK_LIBPARAM_PARAMSECTION_ENDSYM -
-		       (__uptr) UK_LIBPARAM_PARAMSECTION_STARTSYM) /
-		      sizeof(void *);
-
-	if (UK_LIBPARAM_LIBDESC.params_len > 0) {
-		UK_LIBPARAM_LIBDESC.params = (struct uk_libparam_param **)
-					     UK_LIBPARAM_PARAMSECTION_STARTSYM;
-		_uk_libparam_libsec_register(&UK_LIBPARAM_LIBDESC);
-	}
-}
-UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
 
 /* -------------------------------------------------------------------------- */
 /* Parameter registration                                                     */
@@ -185,20 +134,23 @@ UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
 #define __UK_LIBPARAM_PARAM_NAME(varname)				\
 	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PARAM_NAMEPREFIX, varname)
 
-#define __UK_LIBPARAM_PARAM_DEFINE(arg_var, arg_addr, arg_type, arg_count, \
+#define __UK_LIBPARAM_DATA_NAME(varname)				\
+	UK_LIBPARAM_CONCAT(UK_LIBPARAM_PDATA_NAMEPREFIX, varname)
+
+#define __UK_LIBPARAM_PARAM_DEFINE(arg_var, arg_addr, arg_type, arg_count,\
 				   arg_desc)				\
-	static struct uk_libparam_param __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
-		.name  = STRINGIFY(arg_var),				\
-		.desc  = arg_desc,					\
-		.type  = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
-		.count = arg_count,					\
-		.addr  = arg_addr					\
-	};								\
-									\
-	static __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME))	\
-	       __used struct uk_libparam_param * const			\
-	       UK_LIBPARAM_CONCAT(__UK_LIBPARAM_PARAM_NAME(arg_var), _ptr) = \
-		&__UK_LIBPARAM_PARAM_NAME(arg_var)
+	static struct uk_libparam_pdata __UK_LIBPARAM_DATA_NAME(arg_var);\
+	static const struct uk_libparam_param				\
+	__used __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME))	\
+	__align(8) __UK_LIBPARAM_PARAM_NAME(arg_var) = {		\
+		.prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX),		\
+		.name   = STRINGIFY(arg_var),				\
+		.desc   = arg_desc,					\
+		.type   = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
+		.count  = arg_count,					\
+		.addr   = arg_addr,					\
+		.pdata  = &__UK_LIBPARAM_DATA_NAME(arg_var)		\
+	}
 
 #define _UK_LIBPARAM_PARAM_DEFINE(name, var, type, count, desc) \
 	__UK_LIBPARAM_PARAM_DEFINE(name, var, type, count, desc)
--- a/lib/uklibparam/libparam.lds.S
+++ b/lib/uklibparam/libparam.lds.S
@@ -1,12 +1,17 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
 #include <uk/libparam.h>
 
 SECTIONS {
 	. = ALIGN(8);
+	PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
 	.UK_LIBPARAM_PARAMSECTION_NAME : {
-		PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME))
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME.*))
-		PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
+		KEEP(*(SORT_BY_NAME(UK_LIBPARAM_PARAM_NAMEPREFIX*)));
 	}
+	PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
 }
 INSERT AFTER .rodata
--- a/lib/uklibparam/parser.c
+++ b/lib/uklibparam/parser.c
@@ -31,6 +31,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+
 #include <uk/config.h>
 #include <uk/libparam.h>
 #include <ctype.h>
@@ -43,19 +44,12 @@
 #include <uk/version.h>
 #include <uk/essentials.h>
 
-static UK_LIST_HEAD(ld_head);
-
 #define PARSE_PARAM_SEP    '.'
 #define PARSE_VALUE_SEP    '='
 #define PARSE_LIST_START   '['
 #define PARSE_LIST_END     ']'
 #define PARSE_STOP         "--"
 #define PARSE_USAGE        "help"
-
-void _uk_libparam_libsec_register(struct uk_libparam_libdesc *ld)
-{
-	uk_list_add_tail(&ld->next, &ld_head);
-}
 
 static const char *str_param_type(enum uk_libparam_param_type pt)
 {
@@ -88,18 +82,16 @@ static const char *str_param_type(enum uk_libparam_param_type pt)
 	return "<?>";
 }
 
-#define UK_LIBPARAM_FOREACH_LIBDESC(ld_iter) \
-	uk_list_for_each_entry((ld_iter), &ld_head, next)
-#define UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_iter) \
-	for ((p_iter) = 0; (p_iter) < (ld)->params_len; (p_iter)++)
-#define UK_LIBPARAM_PARAM_GET(libdesc, idx) \
-	(((idx) >= (libdesc)->params_len) ? __NULL : ((libdesc)->params)[idx])
+#define UK_LIBPARAM_FOREACH_PARAM(itr)				\
+	for ((itr) = (struct uk_libparam_param *)		\
+			UK_LIBPARAM_PARAMSECTION_STARTSYM;	\
+	     (itr) < (struct uk_libparam_param *)		\
+			&UK_LIBPARAM_PARAMSECTION_ENDSYM;	\
+	     (itr)++)
 
 static void uk_usage(void)
 {
-	struct uk_libparam_libdesc *ld;
 	struct uk_libparam_param *p;
-	__sz p_i;
 
 	/*
 	 * FIXME: Use a console print variant without context prefix
@@ -112,21 +104,16 @@ static void uk_usage(void)
 		   "%12s                   Print this help summary\n\n",
 		   PARSE_USAGE);
 	uk_pr_info("Available parameters:\n");
-	UK_LIBPARAM_FOREACH_LIBDESC(ld) {
-		UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_i) {
-			p = UK_LIBPARAM_PARAM_GET(ld, p_i);
-			UK_ASSERT(p);
 
-			if (p->count == 0)
-				continue;
-			uk_pr_info("%12s.%-18s%s (",
-				   ld->prefix, p->name,
-				   p->desc ? p->desc : p->name);
-			if (p->count > 1)
-				uk_pr_info("array[%"__PRIsz"] of ",
-				p->count);
-			uk_pr_info("%s)\n", str_param_type(p->type));
-		}
+	UK_LIBPARAM_FOREACH_PARAM(p) {
+		UK_ASSERT(p);
+		if (p->count == 0)
+			continue;
+		uk_pr_info("%12s.%-18s%s (", p->prefix, p->name,
+			   p->desc ? p->desc : p->name);
+		if (p->count > 1)
+			uk_pr_info("array[%" __PRIsz "] of ", p->count);
+		uk_pr_info("%s)\n", str_param_type(p->type));
 	}
 	uk_pr_info("\n"
 		   "Numbers can be passed in decimal, octal (\"0\" as prefix), or hexadecimal (\"0x\" as prefix).\n");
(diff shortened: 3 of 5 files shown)


lib/uklibparam/include/uk/libparam.h
```diff
@@ -187,18 +130,16 @@ UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
...
 				   arg_desc)				\
-	static struct uk_libparam_param __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
-		.name  = STRINGIFY(arg_var),				\
-		.desc  = arg_desc,					\
-		.type  = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
-		.count = arg_count,					\
-		.addr  = arg_addr					\
+	static const struct uk_libparam_param				\
+	__used __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME))	\
+	__align(8) __UK_LIBPARAM_PARAM_NAME(arg_var) = {		\
```
mogasergiu, line 145: I wonder if it's worth surrounding every argument with `( )`? :thinking:

lib/uklibparam/parser.c
```diff
@@ -112,22 +104,19 @@ static void uk_usage(void)
 		   "%12s                   Print this help summary\n\n",
 		   PARSE_USAGE);
 	uk_pr_info("Available parameters:\n");
-	UK_LIBPARAM_FOREACH_LIBDESC(ld) {
-		UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_i) {
-			p = UK_LIBPARAM_PARAM_GET(ld, p_i);
-			UK_ASSERT(p);
 
+		UK_LIBPARAM_FOREACH_PARAM(p) {
```
mogasergiu: Surplus indentation?

lib/uklibparam/parser.c
```diff
@@ -597,12 +567,11 @@ static int parse_arg(struct parse_arg_ctx *ctx, char *strbuf, int scan_only)
...
-				if (ctx->ld)
-					ctx->p = find_libparam(ctx->ld,
-							       paramname,
-							       paramname_len);
-				if (!ctx->ld || !ctx->p) {
+				ctx->p = find_libparam(libname,
+						       libname_len,
+						       paramname,
+						       paramname_len);
+				if (!ctx->p) {
```
mogasergiu: I wonder, if this is followed by an `-ENOENT` and a warning, if it would be ok to use `unlikely` here.

lib/uklibparam/parser.c
```diff
@@ -152,40 +141,22 @@ struct parse_arg_ctx {
...
-	__sz i;
-
-	UK_ASSERT(ld);
 
-	UK_LIBPARAM_FOREACH_PARAMIDX(ld, i) {
-		p = UK_LIBPARAM_PARAM_GET(ld, i);
+	UK_ASSERT(libname);
+	UK_ASSERT(paramname);
 
+	UK_LIBPARAM_FOREACH_PARAM(p) {
```
mogasergiu, line 155: Perhaps a `UK_ASSERT(p);` here as well, since the other FOREACH's have it as well?

plat/common/include/uk/plat/common/common.lds.h
```diff
@@ -136,6 +136,14 @@
 		KEEP(*(SORT_BY_NAME(.uk_event_*)))			\
 	}
 
+#define LIBPARAM_SECTION						\
+	uk_libparam_params_start = .;					\
```
mogasergiu: Following some weird paranormal events with `ld` misplacing some of these ld script defined syms in the case of empty sections, I think we should start keeping these defined within the scope of the section they are meant for:
```C
#define LIBPARAM_SECTION						\
	.uk_libparam_params :						\
	{								\
		uk_libparam_params_start = .;				\
		KEEP(*(SORT_BY_NAME(.uk_libparam_param*)))		\
		uk_libparam_params_end = .;				\
	}
```

mogasergiu (changes requested):
Hi @michpappas :wave: :ocean: 
Had a first look. Really nice cleanup! :triumph:

plat/common/include/uk/plat/common/common.lds.h
michpappas: Hmmm, I'm not much of a believer of paranormal phenomena. Can you elaborate on the problem? Either way, if we make such a change, we should probably make it to all sections in that file on a follow-up PR.

lib/uklibparam/include/uk/libparam.h
michpappas: I think it should be okay as is as we don't perform any operations.

plat/common/include/uk/plat/common/common.lds.h
michpappas: Update: I moved the section into a library-defined linker script, and moved the variables into the section. This also fixes a bug where the libukparam section would be mapped as RO. My proposal for a global update on `common.lds.h` holds.

lib/uklibparam/include/uk/libparam.h
```diff
@@ -187,18 +130,16 @@ UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
...
+		.prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX),		\
+		.name   = STRINGIFY(arg_var),				\
+		.desc   = arg_desc,					\
+		.type   = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
+		.count  = arg_count,					\
+		.addr   = arg_addr					\
```
mogasergiu: Nitpick: maybe let's end this like with a `,` as well to keep the convention.

lib/uklibparam/parser.c
```diff
@@ -88,18 +82,16 @@ static const char *str_param_type(enum uk_libparam_param_type pt)
...
 
-#define UK_LIBPARAM_FOREACH_LIBDESC(ld_iter) \
-	uk_list_for_each_entry((ld_iter), &ld_head, next)
-#define UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_iter) \
-	for ((p_iter) = 0; (p_iter) < (ld)->params_len; (p_iter)++)
-#define UK_LIBPARAM_PARAM_GET(libdesc, idx) \
-	(((idx) >= (libdesc)->params_len) ? __NULL : ((libdesc)->params)[idx])
+#define UK_LIBPARAM_FOREACH_PARAM(itr)					\
+	for ((itr) = DECONST(struct uk_libparam_param *,		\
+			UK_LIBPARAM_PARAMSECTION_STARTSYM);		\
```
mogasergiu: This looks misaligned with the opened paranthesis above.

lib/uklibparam/libparam.lds.S
```diff
@@ -4,9 +4,8 @@ SECTIONS {
 	. = ALIGN(8);
 	.UK_LIBPARAM_PARAMSECTION_NAME : {
 		PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME))
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME.*))
+		KEEP(*(SORT_BY_NAME(UK_LIBPARAM_PARAM_NAMEPREFIX*)));
```
mogasergiu, line 13: Strange, even though this has `KEEP` and I have environ, netdev and fstab enabled, which should have registered something in this section, it seems to be empty and non-existent:
```
$ readelf -S --wide build/elfloader_qemu-x86_64.dbg           
There are 33 section headers, starting at offset 0x55ccb8:                                
                                                                                          
Section Headers:                                                                          
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al  
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0  
  [ 1] .text             PROGBITS        0000000000100000 001000 0c65ae 00  AX  0   0 4096
  [ 2] .rodata           PROGBITS        00000000001c7000 0c8000 016440 00   A  0   0 32  
  [ 3] .uk_fs_list       PROGBITS        00000000001dd440 0de440 000010 00  WA  0   0  8  
  [ 4] .uk_thread_inittab PROGBITS        00000000001dd450 0de450 000048 00   A  0   0  8 
  [ 5] .uk_excpttab      PROGBITS        00000000001dd498 0de498 000090 00   A  0   0  1  
  [ 6] .uk_libinfo       PROGBITS        00000000001dd528 0de528 000de5 00   A  0   0  1  
  [ 7] .uk_posix_clonetab PROGBITS        00000000001de310 0df310 000140 00   A  0   0  8 
  [ 8] .eh_frame         PROGBITS        00000000001df000 0e0000 00a4d8 00   A  0   0  8  
  [ 9] .eh_frame_hdr     PROGBITS        00000000001e94d8 0ea4d8 0007b4 00   A  0   0  4  
  [10] .uk_ctortab       PROGBITS        00000000001ea000 0eb000 000060 00   A  0   0  8  
  [11] .uk_inittab       PROGBITS        00000000001ea060 0eb060 0000c0 00   A  0   0  8  
  [12] .uk_eventtab      PROGBITS        00000000001ea120 0eb120 0000a0 00  WA  0   0  8  
  [13] .init_array       INIT_ARRAY      00000000001ea1c0 0eb1c0 000008 08  WA  0   0  8  
  [14] .tdata            PROGBITS        00000000001eb000 0ec000 000004 00 WAT  0   0  8  
  [15] .tbss             NOBITS          00000000001eb008 0ec004 000028 00 WAT  0   0  8  
  [16] .data             PROGBITS        00000000001ec000 0ed000 001bc8 00  WA  0   0 32  
  [17] .posix_socket_driver_list PROGBITS        00000000001edbc8 0eebc8 000050 00  WA  0 
  0  8                                                                                    
  [18] .uk_bootinfo      PROGBITS        00000000001edc18 0eec18 000c50 00  WA  0   0  8  
  [19] .bss              NOBITS          00000000001ef000 0ef868 464000 00  WA  0   0 64  
  [20] .comment          PROGBITS        0000000000000000 0ef868 000ba0 01  MS  0   0  1  
  [21] .debug_aranges    PROGBITS        0000000000000000 0f0410 011390 00      0   0 16  
  [22] .debug_info       PROGBITS        0000000000000000 1017a0 13c48b 00      0   0  1  
  [23] .debug_abbrev     PROGBITS        0000000000000000 23dc2b 025148 00      0   0  1  
  [24] .debug_line       PROGBITS        0000000000000000 262d73 07794e 00      0   0  1  
  [25] .debug_frame      PROGBITS        0000000000000000 2da6c8 02a2b0 00      0   0  8  
  [26] .debug_str        PROGBITS        0000000000000000 304978 11fa9f 01  MS  0   0  1  
  [27] .debug_macro      PROGBITS        0000000000000000 424417 0fa3b0 00      0   0  1  
  [28] .debug_line_str   PROGBITS        0000000000000000 51e7c7 009fa6 01  MS  0   0  1  
  [29] .debug_rnglists   PROGBITS        0000000000000000 52876d 00b480 00      0   0  1  
  [30] .symtab           SYMTAB          0000000000000000 533bf0 019e90 18     31 3005  8 
  [31] .strtab           STRTAB          0000000000000000 54da80 00f0b5 00      0   0  1  
  [32] .shstrtab         STRTAB          0000000000000000 55cb35 00017f 00      0   0  1  
Key to Flags:                                                                             
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),                    
  L (link order), O (extra OS processing required), G (group), T (TLS),                   
  C (compressed), x (unknown), o (OS specific), E (exclude),                              
  D (mbind), l (large), p (processor specific)                                            
```
```
$ nm build/elfloader_qemu-x86_64.dbg | grep uk_libparam_param
00000000001edbc8 D uk_libparam_params_end                                                
00000000001edbc8 D uk_libparam_params_start                                              
```
Am I missing something?

mogasergiu (changes requested):

lib/uklibparam/include/uk/libparam.h
michpappas: I followed the existing convention, see the diff. Generally speaking I'm okay with either way, but unless I missed something I didn't find anything related in https://unikraft.org/docs/contributing/coding-conventions. Until we have such a convention I would consider this a matter of personal style and stick to the original library author's preferred convention.

```diff
-       static struct uk_libparam_param __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
-               .name  = STRINGIFY(arg_var),                            \
-               .desc  = arg_desc,                                      \
-               .type  = _UK_LIBPARAM_PARAM_TYPE_##arg_type,            \
-               .count = arg_count,                                     \
-               .addr  = arg_addr                                       \
+       static const struct uk_libparam_param                           \
+       __used __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME))  \
+       __align(8) __UK_LIBPARAM_PARAM_NAME(arg_var) = {                \
+               .prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX),             \
+               .name   = STRINGIFY(arg_var),                           \
+               .desc   = arg_desc,                                     \
+               .type   = _UK_LIBPARAM_PARAM_TYPE_##arg_type,           \
+               .count  = arg_count,                                    \
+               .addr   = arg_addr                                      \
```

lib/uklibparam/libparam.lds.S
michpappas: Did you enable `CONFIG_LIBUKNETDEV_EINFO_LIBPARAM`?

Btw I just noticed that moving the `start` / `end` vars inside the section results into both vars pointing to the same address:
```
bash$ readelf -sW build/app-helloworld_qemu-x86_64.dbg | grep param_param
   332: 000000000017b690    56 OBJECT  LOCAL  DEFAULT   14 __uk_libparam_param_ip
   696: 000000000017b6c8    56 OBJECT  LOCAL  DEFAULT   14 __uk_libparam_param_device
  1030: 000000000017b690     0 NOTYPE  GLOBAL DEFAULT   14 uk_libparam_params_start
  1055: 000000000017b690     0 NOTYPE  GLOBAL DEFAULT   14 uk_libparam_params_end
```
Moving `start` / `end` outside the section results into the expected behavior:
```diff
bash$ git diff
diff --git a/lib/uklibparam/libparam.lds.S b/lib/uklibparam/libparam.lds.S
index 7c5de3962..195f93487 100644
--- a/lib/uklibparam/libparam.lds.S
+++ b/lib/uklibparam/libparam.lds.S
@@ -2,10 +2,10 @@
 
 SECTIONS {
        . = ALIGN(8);
+       PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
        .UK_LIBPARAM_PARAMSECTION_NAME : {
-               PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
                KEEP(*(SORT_BY_NAME(UK_LIBPARAM_PARAM_NAMEPREFIX*)));
-               PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
        }
+       PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
 }
 INSERT AFTER .data
```

```
bash$ readelf -sW build/app-helloworld_qemu-x86_64.dbg | grep param_param
   332: 000000000017b690    56 OBJECT  LOCAL  DEFAULT   28 __uk_libparam_param_ip
   696: 000000000017b6c8    56 OBJECT  LOCAL  DEFAULT   28 __uk_libparam_param_device
  1030: 000000000017b690     0 NOTYPE  GLOBAL DEFAULT   28 uk_libparam_params_start
  1055: 000000000017b700     0 NOTYPE  GLOBAL DEFAULT   28 uk_libparam_params_end
```
Looks like when the section is defined as part of the library, the linker assigns addresses to the symbols of `libuklibparam` before those of other libraries, while when these vars are defined outside the section, it first collects the symbols from all libraries and then calculates the section boundaries (as expected).

Tested with `GCC-13` and `GCC-14` with no issues when `.uk_libparam_params` is empty, so I think we need to investigate the issure discuseed in https://github.com/unikraft/unikraft/pull/1471#discussion_r1686647738 separately, and most importantly after being able to reproduce.

lib/uklibparam/include/uk/libparam.h
mogasergiu: That comma is a common pattern I've seen in most open source C-based projects. There are a few benefits such as easily being able to reorder lines later or easily add new statically initialized fields. I am not going to insist on a trailing comma, but I wanted to take a moment to explain why I am thinking that it is also objectively better, since I do not see any benefit to not adding it.

lib/uklibparam/libparam.lds.S
mogasergiu: Ok then this is very weird. Honestly, it feels like a static linker bug. I always thought these are equivalent.

lib/uklibparam/include/uk/libparam.h
michpappas: Yes, as I said I do not object to that, in fact I agree. But I would like that we either formalize such things,  or refrain from ad-hoc reviewer's requests on unspecified things that are often a matter of personal taste (e.g. xmas tree). Saying that, if you feel it's important enough to open a PR for `coding-conventions` on the above, I'll vouch for this change.

plat/common/include/uk/plat/common/common.lds.h
michpappas: Update2: See discussion in https://github.com/unikraft/unikraft/pull/1471#discussion_r1694429633.

mogasergiu (approved):

lib/uklibparam/include/uk/libparam.h
```diff
@@ -187,18 +130,16 @@ UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
...
-	static struct uk_libparam_param __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
-		.name  = STRINGIFY(arg_var),				\
-		.desc  = arg_desc,					\
-		.type  = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
-		.count = arg_count,					\
-		.addr  = arg_addr					\
+	static const struct uk_libparam_param				\
+	__used __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME))	\
+	__align(8) __UK_LIBPARAM_PARAM_NAME(arg_var) = {		\
+		.prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX),		\
```
skuenzer, line 146: Out of curiosity, is `gcc` clever enough to declare the library prefix only once and use the same reference for each parameter of the same library? I think it is, because of `const char *` and the actual string ending up in `rodata`, correct?

lib/uklibparam/include/uk/libparam.h
```diff
@@ -187,18 +130,16 @@ UK_CTOR_PRIO(UK_LIBPARAM_LIBDESC_CTOR, 0);
 
 #define __UK_LIBPARAM_PARAM_DEFINE(arg_var, arg_addr, arg_type, arg_count, \
 				   arg_desc)				\
-	static struct uk_libparam_param __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
-		.name  = STRINGIFY(arg_var),				\
-		.desc  = arg_desc,					\
-		.type  = _UK_LIBPARAM_PARAM_TYPE_##arg_type,		\
-		.count = arg_count,					\
-		.addr  = arg_addr					\
+	static const struct uk_libparam_param				\
```
skuenzer, line 143: Hum... we are touching `__widx` later for parsing... should we really declare the struct as `const`? Doing later a DECONST feels wrong... In principle we could also move any `rw` fields to another static `struct` and reference to that one from the parameter struct. This extra struct does not need to be in a special section.

lib/uklibparam/libparam.lds.S
```diff
@@ -1,12 +1,17 @@
...
-		PROVIDE(UK_LIBPARAM_PARAMSECTION_STARTSYM = .);
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME))
-		KEEP(*(.UK_LIBPARAM_PARAMSECTION_NAME.*))
-		PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
+		KEEP(*(SORT_BY_NAME(UK_LIBPARAM_PARAM_NAMEPREFIX*)));
 	}
+	PROVIDE(UK_LIBPARAM_PARAMSECTION_ENDSYM = .);
 }
-INSERT AFTER .rodata
+INSERT AFTER .data
```
skuenzer: With comment https://github.com/unikraft/unikraft/pull/1471/commits/6b37f210a493c9275206404e1e780dc64809f1cb#diff-dba471d15bca2731fbc43ddfc22a50d7662c080713d8f2184325d3ae92bc5554R133 we probably can make the section read-only.

lib/uklibparam/parser.c
```diff
@@ -88,18 +82,16 @@ static const char *str_param_type(enum uk_libparam_param_type pt)
...
 }
 
-#define UK_LIBPARAM_FOREACH_LIBDESC(ld_iter) \
-	uk_list_for_each_entry((ld_iter), &ld_head, next)
-#define UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_iter) \
-	for ((p_iter) = 0; (p_iter) < (ld)->params_len; (p_iter)++)
-#define UK_LIBPARAM_PARAM_GET(libdesc, idx) \
-	(((idx) >= (libdesc)->params_len) ? __NULL : ((libdesc)->params)[idx])
+#define UK_LIBPARAM_FOREACH_PARAM(itr)					\
+	for ((itr) = DECONST(struct uk_libparam_param *,		\
```
skuenzer: See also here: https://github.com/unikraft/unikraft/pull/1471/commits/6b37f210a493c9275206404e1e780dc64809f1cb#diff-dba471d15bca2731fbc43ddfc22a50d7662c080713d8f2184325d3ae92bc5554R133

lib/uklibparam/parser.c
```diff
@@ -597,12 +566,11 @@ static int parse_arg(struct parse_arg_ctx *ctx, char *strbuf, int scan_only)
 			 *       not allowed characters, or empty strings).
 			 */
 			if (!scan_only) {
-				ctx->ld = find_libdesc(libname, libname_len);
-				if (ctx->ld)
-					ctx->p = find_libparam(ctx->ld,
-							       paramname,
-							       paramname_len);
-				if (!ctx->ld || !ctx->p) {
+				ctx->p = find_libparam(libname,
```
skuenzer, line 577: Hum, this changes the potential number of required comparisons. A library with many parameters will cause that the library name is compared much more often. It will be costly for arguments that belong to a library with a big list of parameters. However, because we need to see first how much this in reality impacts the runtime performance, I would add a comment for now that briefly mentions that we could optimize this in the future with an adopted layout if needed.

lib/uklibparam/parser.c
```diff
@@ -664,16 +632,11 @@ static int parse_arg(struct parse_arg_ctx *ctx, char *strbuf, int scan_only)
...
 
-	UK_LIBPARAM_FOREACH_LIBDESC(ld) {
-		UK_LIBPARAM_FOREACH_PARAMIDX(ld, p_i) {
-			p = UK_LIBPARAM_PARAM_GET(ld, p_i);
-			UK_ASSERT(p);
-			p->__widx = 0;
-		}
+	UK_LIBPARAM_FOREACH_PARAM(p) {
+		UK_ASSERT(p);
+		p->__widx = 0;
```
skuenzer: Yes, it is probably cleaner to move `__widx` to a different declaration so that we can keep the parameter struct as `const`.

lib/uklibparam/include/uk/libparam.h
michpappas: Exactly, when adding a 2nd parameter in `libuknetdev`, its `prefix` element will point the same string in `.rodata` (`0x0015d12d` at `0x0016f650` for the 1st parameter, and `0x0015d12d` at `0x0016f688` for the 2nd parameter):
```
Hex dump of section '.uk_libparam_params':
  0x0016f650 2dd11500 00000000 6cda1500 00000000 -.......l.......
  0x0016f660 08dd1500 00000000 0b000000 00000000 ................
  0x0016f670 04000000 00000000 a0c41a00 00000000 ................
  0x0016f680 00000000 00000000 2dd11500 00000000 ........-.......
  0x0016f690 b0eb1500 00000000 08dd1500 00000000 ................
  0x0016f6a0 0b000000 00000000 04000000 00000000 ................
  0x0016f6b0 c0c41a00 00000000 00000000 00000000 ................
  0x0016f6c0 82d11500 00000000 f3bb1500 00000000 ................
  0x0016f6d0 a6101600 00000000 0b000000 00000000 ................
  0x0016f6e0 04000000 00000000 a0fb1a00 00000000 ................
  0x0016f6f0 00000000 00000000                   ........
```
In fact, GCC is so smart that for the `netdev` prefix specified in `addlibparamprefix`, it will reuse part of the existing `libuknetdev` string in `rodata` at `0x0015d128`:
```
  0x0015d120 6962756b 6d706900 6c696275 6b6e6574 ibukmpi.libuknet
  0x0015d130 64657600 6c696275 6b72616e 646f6d00 dev.libukrandom.
```

lib/uklibparam/parser.c
michpappas: Yes I noticed that too, but I believe that in practice the possibility to come across such a case is slim to be worth implementing a more sophisticated scheme. I'll add a comment.

lib/uklibparam/include/uk/libparam.h
michpappas: You're right, the only reason I didn't do it is because it's only `_widx` that is non-const in that struct, so we would need to declare a struct just for that, similarly to:
```c
struct uk_libparam_data {
        /* Internally used by the parser for array parameters */
        __sz __widx;
};

#define __UK_LIBPARAM_PARAM_DEFINE(arg_var, arg_addr, arg_type, arg_count,
                                   arg_desc) \
        static const struct uk_libparam_data __UK_LIBPARAM_DATA_NAME(arg_var); \
        static const struct uk_libparam_param \
        __used __section("." STRINGIFY(UK_LIBPARAM_PARAMSECTION_NAME)) \
        __align(8) __UK_LIBPARAM_PARAM_NAME(arg_var) = { \
                .prefix = STRINGIFY(UK_LIBPARAM_LIBPREFIX), \
                .name   = STRINGIFY(arg_var), \
                .desc   = arg_desc, \
                .type   = _UK_LIBPARAM_PARAM_TYPE_##arg_type, \
                .count  = arg_count, \
                .addr   = arg_addr \
                .data   = &__UK_LIBPARAM_DATA_NAME(arg_var); \
        };
```
Is that what you essentially propose?

PS: Yes `data` is a very uninspired name.

lib/uklibparam/include/uk/libparam.h
skuenzer: Thanks for confirming 👍🏻

lib/uklibparam/include/uk/libparam.h
skuenzer: Yes, I think we should do this. This would clean-up the declarations and is clearer about what is changeable and what not. Should `const` be removed from `data`?

lib/uklibparam/parser.c
skuenzer: Thanks, and yes, I agree with you.

michpappas:
Changes in this update:
* Move non-const parameters into own struct in `.data` and update parser.
* Move section to `.rodata`
* Remove `DECONST()`

CC: @skuenzer

lib/uklibparam/parser.c
skuenzer: I am thinking of a `TODO` comment instead of a `FIXME` because functionality-wise, everything is correct. There is also no shortcut/simplification that ignores corner cases.

I thought along these lines:
```c
/*
 * TODO: This can become expensive if the number of parameters for a
 *       library becomes high: The library name is repeatedly checked
 *       for a match for each library parameter (O(n2)). However, the
 *       impact in practice has yet to be assessed.
 */

skuenzer (approved):
Great work. Thanks for the updates!

PR #1472, opened 2024-07-18
lib/ukboot: Introduce early init bootstage

michpappas:
### Prerequisite checklist

### Base target

 - Platform(s): [`kvm`, `xen`]
 - Application(s): [N/A]

### Description of changes

This series introduces an `early_init` bootstage in `libukboot` executed as soon as the boot protocol passes control to the platform. Libraries can execute in `early_init` by registering functions to `uk_boot_earlytab`, a minimal variant of `uk_inittab`. 

The kernel command line parsing is moved to the earliest priority of `early_init` to provide command line parameters to the rest of the functions registered with earlytab. As all parsing is handled internally by `uk_boot_early_init()`, `ukplat_entry_argp()` is deprecated. Platforms are required to execute `uk_boot_early_init()` before jumping to the entry point of `libukboot`, now renamed to `uk_boot_entry()`.

Bootinfo mrd coalescing is moved to the penultimate priority of the earlytab. Coalescing at the boot protocol is deprecated. The latest priority of earlytab is reserved for post-coalesce operations.

As command line parsing happens before coalescing, it is not possible to copy the command line into a dynamically allocated mrd. `UKPLAT_MEMRT_CMDLINE` and its registration by the boot protocol is deprecated. The boot protocols are updated to provide a direct reference to the command line via `bi->cmdline`. Similarly, copying the command line into a mutable buffer by the platform is deprecated, as that is now internally managed by `uk_boot_early_init()`.

Platforms  (`plat/kvm` & `plat/xen`) are adapted and cleaned up as needed.

GitHub-Depends-On: #1471


lib/ukboot/include/uk/boot/earlytab.h
```diff
@@ -0,0 +1,48 @@
...
+	uk_boot_earlytab_func_t init;
+};
+
+extern const struct uk_boot_earlytab_entry uk_boot_earlytab_start[];
+extern const struct uk_boot_earlytab_entry uk_boot_earlytab_end;
+
+#define __UK_EARLYTAB_ENTRY(init_fn, prio)				\
+	static const struct uk_boot_earlytab_entry			\
+	__used __section(".uk_boot_earlytab" #prio) __align(8)		\
+		__uk_boot_earlytab ## prio ## _ ## init_fn = {		\
```
mogasergiu: is this overindented?

plat/common/include/uk/plat/common/common.lds.h
```diff
@@ -129,6 +129,14 @@
 	}								\
 	uk_inittab_end = .;
 
+#define EARLYTAB_SECTION						\
```
mogasergiu: Ditto as in #1471. Additionally, any reason this is not named something like `BOOT_EARLYTAB_SECTION`? As in, the others seem to be defined in such a manner that just the "UK" :british_indian_ocean_territory: is missing.
```C
#define EARLYTAB_SECTION						\
	.uk_boot_earlytab :						\
	{								\
		uk_boot_earlytab_start = .;				\
		KEEP(*(SORT_BY_NAME(.uk_boot_earlytab[0-9])))		\
		uk_boot_earlytab_end = .;				\
	}
```

include/uk/plat/bootstrap.h
```diff
@@ -50,19 +50,7 @@ extern "C" {
  * @param argc Number of arguments
```
mogasergiu: Hmm, maybe also delete the `@param` comments here?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+#include <uk/argparse.h>
+#include <uk/assert.h>
+#include <uk/boot/earlytab.h>
+#include <uk/essentials.h>
+#include <uk/plat/common/bootinfo.h>
+#include <uk/prio.h>
+
+#ifdef CONFIG_LIBUKLIBPARAM
```
mogasergiu: `#if`?

lib/ukboot/boot.c
```diff
@@ -111,6 +107,9 @@
 #include <uk/intctlr.h>
 #endif /* CONFIG_LIBUKINTCTLR */
 
+extern char *argv[CONFIG_LIBUKBOOT_MAXNBARGS];
```
mogasergiu: I feel like these names are a bit "established"? Maybe we should just name them as `boot_argv`/`boot_argc` or something? I am fine with this as well but it's just that `argv` and `argc` sound like something other components may want to unexpectedly define for themselves.

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+#include <uk/libparam.h>
+#endif /* CONFIG_LIBUKLIBPARAM */
+
+char *argv[CONFIG_LIBUKBOOT_MAXNBARGS];
+int argc;
+
+static char cmdline[CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN];
+static __sz cmdline_len;
+
+#ifdef CONFIG_LIBUKLIBPARAM
```
mogasergiu: `#if`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+static int init_libparam(int argc, char *argv[])
+{
+	int rc;
```
mogasergiu: Nitpick: a newline afterwards?

Perhaps some assertions for `argv` as well? Especially since `argv[1]` is there.

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+#endif /* CONFIG_LIBUKLIBPARAM */
+
+char *argv[CONFIG_LIBUKBOOT_MAXNBARGS];
+int argc;
+
+static char cmdline[CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN];
+static __sz cmdline_len;
+
+#ifdef CONFIG_LIBUKLIBPARAM
+static int init_libparam(int argc, char *argv[])
```
mogasergiu: See, you are using these names as arguments already. I understand the compiler is smart and stuff but I think it could create confusion.

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+	 *       hide from the parser.
+	 */
+	rc = uk_libparam_parse(argc - 1, &argv[1], UK_LIBPARAM_F_SCAN);
+	if (rc > 0 && rc < (argc - 1)) {
+		/* In this case, we did successfully scan for uklibparam
+		 * arguments and stop sequence is at rc < (argc - 1).
+		 */
+		/* Run a second pass for parsing */
+		rc = uk_libparam_parse(argc - 1, &argv[1], UK_LIBPARAM_F_USAGE);
+		if (rc < 0) /* go down on errors (including USAGE) */
```
mogasergiu: `unlikely`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+{
+	int rc;
+	/*
+	 * First, we scan if we can find the stop sequence in the kernel
+	 * cmdline. If not, we assume that there are no uklibparam arguments in
+	 * the command line.
+	 * NOTE: argv[0] contains the kernel/program name that we need to
+	 *       hide from the parser.
+	 */
+	rc = uk_libparam_parse(argc - 1, &argv[1], UK_LIBPARAM_F_SCAN);
```
mogasergiu: Nitpick: perhaps `rc` is not really the proper name for the variable. Maybe `arg_idx`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+		argv = &argv[rc];
+	}
+	return 0;
+}
+#endif /* CONFIG_LIBUKLIBPARAM */
+
+static int uk_boot_early_init_parse_cmdl(struct ukplat_bootinfo *bi)
+{
+	const char *cmdl;
+	int rc __maybe_unused;
```
mogasergiu, line 36: AM I SEEING THIS RIGHT? IS THIS.... NOT REVERSE XMAS TREE? :monocle_face: :laughing:

## UNACCEPTABLE

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+
+	UK_ASSERT(bi);
+
+	bi = ukplat_bootinfo_get();
```
mogasergiu: Why do we need to fetch it again? Misplaced getter maybe?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+	 */
+	if (!bi->cmdline_len) {
+		cmdl = CONFIG_UK_NAME;
+		cmdline_len = sizeof(CONFIG_UK_NAME);
+	} else {
+		cmdl = (const char *)bi->cmdline;
+		cmdline_len = bi->cmdline_len;
+	}
+
+	if (cmdline_len > ARRAY_SIZE(cmdline))
```
mogasergiu: `unlikely`? And why `ARRAY_SIZE` instead of `sizeof`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+		return -EINVAL;
+
+	strncpy(cmdline, cmdl, ARRAY_SIZE(cmdline));
```
mogasergiu: Ditto

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+	cmdline[cmdline_len] = 0;
+
+	/* Break down cmdline to argv */
+	argc = uk_argnparse(cmdline, cmdline_len, &argv[0],
+			    CONFIG_LIBUKBOOT_MAXNBARGS);
+
+#ifdef CONFIG_LIBUKLIBPARAM
```
mogasergiu: `#if`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,128 @@
...
+		UK_ASSERT(entry);
+		rc = (*entry->init)(bi);
+		if (unlikely(rc))
+			UK_CRASH("Early init call @ %p failed with %d",
+				 entry->init, rc);
+	}
+}
+
+/* Parse cmdline parameters at the earlies priority
+ * to make them available the rest of early init
```
mogasergiu:
```diff
- Parse cmdline parameters at the earlies priority
+ Parse cmdline parameters at the earliest priority
- to make them available the rest of early init
+ to make them available to the rest of early init
```

lib/ukboot/include/uk/boot.h
```diff
@@ -9,12 +9,18 @@
...
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+/**
+ * Executes the early_init boot stage
+ */
+void uk_boot_early_init(struct ukplat_bootinfo *bi);
```
mogasergiu, line 27: Maybe a `@param` arg comment as well?

plat/common/bootinfo_fdt.c
```diff
@@ -85,38 +85,23 @@ static void fdt_bootinfo_mem_mrd(struct ukplat_bootinfo *bi, void *fdtp)
 		ukplat_bootinfo_crash("Could not add free memory descriptor");
 }
 
-static void fdt_bootinfo_cmdl_mrd(struct ukplat_bootinfo *bi)
+static void fdt_bootinfo_cmdl_init(struct ukplat_bootinfo *bi, void *fdtp)
```
mogasergiu, line 88: In the commit msg: `s/rmd/mrd/`.
Also maybe `const void *fdtp`?

mogasergiu (changes requested):
Greetings, mister sir! :wave: :monocle_face: 
Very happy about this early tab addition. One step closer to welcoming ACPI into a beautiful home! Also cool how you took care of Xen as well :pray:

lib/ukboot/include/uk/boot/earlytab.h
michpappas: Very much so 👍🏼

plat/common/include/uk/plat/common/common.lds.h
michpappas: I'm moving this to an library-internal linker script. I'll keep the `uk_` prefix, as it seems to be used elsewhere too:
```
 Section to Segment mapping:
  Segment Sections...
   00     .text 
   01     .rodata .uk_thread_inittab .uk_libinfo .uk_posix_clonetab .eh_frame .eh_frame_hdr .uk_ctortab .uk_inittab .uk_eventtab 
   02     .data .uk_libparam_params .posix_socket_driver_list .uk_bootinfo .bss 
   03     .tdata .tbss 
   04     .tdata 
   05     
```

lib/ukboot/early_init.c
michpappas: Damn

lib/ukboot/early_init.c
michpappas: I just migrated that from `boot.c` verbatim. Will add.

lib/ukboot/early_init.c
michpappas: Again, that's migrated from `boot.c`.

lib/ukboot/early_init.c
michpappas: It's okay. We don't have an established rule for that anyway.

lib/ukboot/early_init.c
michpappas: Yes, it's been forgotten. I'll remove 👍🏼

lib/ukboot/early_init.c
michpappas: `unlikely` ok. Why *not* `ARRAY_SIZE`?

include/uk/plat/bootstrap.h
michpappas: Ah yes, but this goes away altogether in a later commit in this PR.

lib/ukboot/boot.c
michpappas: Yes,  I thought about that too. But then again these vars really hold `argc` and `argv`. Either way, I don't have a strong opinion.

lib/ukboot/early_init.c
michpappas: I'll add an assertion for `argv`.

lib/ukboot/early_init.c
michpappas: It's actually the scope of identifiers as defined by the C standard. Anyway, I can rename them.

plat/common/bootinfo_fdt.c
michpappas: > In the commit msg: s/rmd/mrd/.

Ok

>  Also maybe const void *fdtp?

I'm following the original author's convention 😉  Saying that, `fdtp` is redundant in most of these functions, as it can be obtained from the `bi` parameter already, so I can add a commit to clean that up.

lib/ukboot/early_init.c
mogasergiu: I like `sizeof` better because it is built-in and makes it clear we are dealing with a string. Maybe it is just personal preference though?

plat/common/bootinfo_fdt.c
mogasergiu: Makes sense.

lib/ukboot/earlytab.lds.S
```diff
@@ -0,0 +1,9 @@
+SECTIONS {
```
mogasergiu, line 7: License?

lib/ukboot/include/uk/boot/earlytab.h
```diff
@@ -0,0 +1,48 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2023, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: 2024

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,137 @@
...
+	if (!bi->cmdline_len) {
+		cmdl = CONFIG_UK_NAME;
+		cmdline_len = sizeof(CONFIG_UK_NAME);
+	} else {
+		cmdl = (const char *)bi->cmdline;
+		cmdline_len = bi->cmdline_len;
+	}
+
+	if (unlikely(cmdline_len > ARRAY_SIZE(cmdline)))
+		return -EINVAL;
```
mogasergiu: Maybe an error message. Why `EINVAL`? Would `E2BIG` fit better here?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,137 @@
...
+
+	/* Break down cmdline to argv */
+	strncpy(cmdline_argv, cmdl, ARRAY_SIZE(cmdline));
+	cmdline_argv[cmdline_len] = 0;
+
+	boot_argc = uk_argnparse(cmdline, cmdline_len, &boot_argv[0],
+				 CONFIG_LIBUKBOOT_MAXNBARGS);
+
+#if CONFIG_LIBUKLIBPARAM
+	rc = init_libparam(boot_argc, boot_argv);
```
mogasergiu: Why not just `return init_libparam`. Either that or I would go as you have now and add an additional error message. The more the better (I know strings increase the size of the binary, but I think it is not something that big...)

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+/* Parse cmdline parameters at the earliest priority
+ * to make them available to the rest of early init
+ */
+uk_boot_earlytab_register(uk_boot_early_init_parse_cmdl, UK_PRIO_EARLIEST);
+
+/* Reserve UK_PRIO_LATEST for post-coalesce operations */
+uk_boot_earlytab_register(uk_boot_early_init_mrd_coalesce,
```
mogasergiu: I wonder if there is an easy way to ensure the fact that no one allocates mrd's before coalesce.

plat/common/bootinfo_fdt.c
```diff
@@ -199,21 +184,7 @@ void ukplat_bootinfo_fdt_setup(void *fdtp)
...
-{
-	struct ukplat_bootinfo *bi;
+	fdt_bootinfo_cmdl_init(bi, fdtp);
 
-	bi = ukplat_bootinfo_get();
-	if (unlikely(!bi))
-		ukplat_bootinfo_crash("Invalid bootinfo");
-
-	fdt_bootinfo_cmdl_mrd(bi);
+	bi->dtb = (__u64)fdtp;
```
mogasergiu, line 189: >> I'm following the original author's convention 😉 Saying that, fdtp is redundant in most of these functions, as it can be obtained from the bi parameter already, so I can add a commit to clean that up.

> Makes sense.

Oh but wait it seems to be assigned later

mogasergiu (changes requested):

lib/ukboot/early_init.c
michpappas: Yeah, I think so as others may argue that it's safer to always use `ARRAY_SIZE` regardless of the underlying type.

lib/ukboot/include/uk/boot/earlytab.h
michpappas: Some people never learn...

lib/ukboot/early_init.c
michpappas: Nice, I forgot about `E2BIG`.

lib/ukboot/early_init.c
michpappas: I guess we could add some logic for that, as well as to enforce a single call of coalesce. If bootinfo becomes part of `libukboot` that makes sense, otherwise if it becomes a standalone lib it seems to add too much policy so I would probably make that behavior configurable. We can keep it in mind as a feature.

plat/common/bootinfo_fdt.c
michpappas: Yes, ut that doesn't stop us from assigning it as soon as we the sanity check succeeds.
```diff
diff --git a/plat/common/bootinfo_fdt.c b/plat/common/bootinfo_fdt.c
index ddf88e4e9..e9953438d 100644
--- a/plat/common/bootinfo_fdt.c
+++ b/plat/common/bootinfo_fdt.c
@@ -180,11 +180,11 @@ void ukplat_bootinfo_fdt_setup(void *fdtp)
        if (unlikely(fdt_check_header(fdtp)))
                ukplat_bootinfo_crash("Invalid DTB");
 
+       bi->dtb = (__u64)fdtp;
+
        fdt_bootinfo_fdt_mrd(bi, fdtp);
        fdt_bootinfo_mem_mrd(bi, fdtp);
        fdt_bootinfo_initrd_mrd(bi, fdtp);
 
        fdt_bootinfo_cmdl_init(bi, fdtp);
-
-       bi->dtb = (__u64)fdtp;
 }
```

What do you think?

michpappas:
Rebased to the latest state of #1471

lib/ukboot/early_init.c
mogasergiu: Ok fine. I just find it unusual being used for strings.

plat/common/bootinfo_fdt.c
mogasergiu: Ok makes sense. It still feels awkward but I dislike having unnecessary secondary arg. :cold_sweat:

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,143 @@
...
+	UK_ASSERT(argv);
+
+	/*
+	 * First, we scan if we can find the stop sequence in the kernel
+	 * cmdline. If not, we assume that there are no uklibparam arguments in
+	 * the command line.
+	 * NOTE: argv[0] contains the kernel/program name that we need to
+	 *       hide from the parser.
+	 */
+	rc = uk_libparam_parse(*argc - 1, &argv[1], UK_LIBPARAM_F_SCAN);
```
mogasergiu: Maybe we should assert `argc` as well?

plat/kvm/x86/multiboot.c
```diff
@@ -71,18 +71,6 @@ void multiboot_entry(struct lcpu *lcpu, struct multiboot_info *mi)
 	if (mi->flags & MULTIBOOT_INFO_CMDLINE) {
 		if (mi->cmdline) {
 			cmdline_len = strlen((const char *)(__uptr)mi->cmdline);
```
mogasergiu: Maybe we can get rid of `cmdline_len` variable entirely and directly assign `bi->cmdline_len` to this `strlen`.

plat/kvm/x86/setup.c
```diff
@@ -24,57 +25,12 @@
...
-}
-
 void _ukplat_entry(struct lcpu *lcpu, struct ukplat_bootinfo *bi)
 {
-	int rc;
 	void *bstack;
+	int rc;
 
-	_libkvmplat_init_console();
+	uk_boot_early_init(bi);
```
mogasergiu: Since this has assertions, maybe we should move it below `traps_table_init()`?

plat/kvm/x86/setup.c
```diff
@@ -134,5 +85,13 @@ void _ukplat_entry(struct lcpu *lcpu, struct ukplat_bootinfo *bi)
...
 	/* Switch away from the bootstrap stack */
 	uk_pr_info("Switch from bootstrap stack to stack @%p\n", bstack);
-	lcpu_arch_jump_to(bstack, _ukplat_entry2);
+
+	/* It's not possible to unwind past this function, because the stack
+	 * pointer was overwritten in lcpu_arch_jump_to. Therefore, mark the
+	 * previous instruction pointer as undefined, so that debuggers or
+	 * profilers stop unwinding here.
+	 */
+	ukarch_cfi_unwind_end();
```
mogasergiu: Are you sure this is ok to use before the `jmp`? I would still play it safe and use an intermediate function.

plat/kvm/x86/setup.c
```diff
@@ -134,5 +85,13 @@ void _ukplat_entry(struct lcpu *lcpu, struct ukplat_bootinfo *bi)
...
+	lcpu_arch_jump_to(bstack, uk_boot_entry);
+	ukplat_lcpu_halt();
```
mogasergiu: I know this is from before. Maybe we should have `UK_BUG()` instead?

plat/xen/arm/setup64.c
```diff
@@ -185,16 +185,7 @@ static inline void _get_cmdline(struct ukplat_bootinfo *bi)
...
-	if (!cmdline)
-		UK_CRASH("Could not allocate command-line memory");
-
-	memcpy(cmdline, cmdl, cmdline_len);
-	/* ensure null termination */
-	cmdline[cmdline_len] = '\0';
-
-	bi->cmdline = (__u64)cmdline;
+	bi->cmdline = (__u64)cmdl;
 	bi->cmdline_len = cmdline_len;
```
mogasergiu: Ditto about `cmdline_len` here. Maybe we can use `len` directly.

plat/xen/x86/setup.c
```diff
@@ -183,49 +184,10 @@ static int _init_mem(struct ukplat_bootinfo *const bi)
...
-		cmdline_len);
-	cmdline[cmdline_len] = '\0';
-}
-
-static void _libxenplat_x86bootinfo_setup(void)
+static void _libxenplat_x86bootinfo_setup(struct ukplat_bootinfo *bi)
 {
 	const char bl[] = "Xen";
 	const char bp[] = "PVH";
-	struct ukplat_bootinfo *bi;
```
mogasergiu, line 224: Should we assert for `bi`?

plat/xen/arm/setup64.c
```diff
@@ -169,7 +170,6 @@ static inline void _get_cmdline(struct ukplat_bootinfo *bi)
 {
 	int fdtchosen, len;
 	const char *fdtcmdline, *cmdl;
-	char *cmdline;
```
mogasergiu, line 172: Ditto for the `bi` here.

mogasergiu (changes requested):

plat/kvm/x86/multiboot.c
michpappas: Actually IMO these are different things: `bi->cmdline_len` reflects what was provided by the boot protocol and it also refers to the length of `bi->cmdline`. If we update `bi->cmdline_len` it would make sense to also update `bi->cmdline`, but I'm not sure there is value on that.

plat/kvm/x86/setup.c
michpappas: For sure. I didn't like that `ukplat_lcpu_halt()` either.

plat/kvm/x86/setup.c
michpappas: Sure, I'm not knowledgeable enough on x86's cfi so I'll revert just in case.

plat/kvm/x86/multiboot.c
michpappas: Wait sorry somehow I thought you referred to `uk_boot_early_init_parse_cmdline()`. The proposed change makes sense in multiboot, I'll push an update.

mogasergiu (approved):
Appreciate these changes :pray:. They will serve the reincarnation of our lord and savior ACPI well in the future :godmode:.

michpappas:
Rebased from #1471, no functional changes.

lib/ukboot/include/uk/boot/earlytab.h
```diff
@@ -0,0 +1,48 @@
...
+#include <uk/plat/common/common.lds.h>
+#include <uk/essentials.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef int (*uk_boot_earlytab_func_t)(struct ukplat_bootinfo *bi);
+
+struct uk_boot_earlytab_entry {
```
skuenzer, line 21: Would we need the struct for future extension? Do you have something already in mind? Otherwise, I think we can handle this table similar as standard constructors.

michpappas: Since we align the struct to 8 bytes I don't think it makes any difference. Below you can see two entries registered from two different libraries:
```
Hex dump of section '.uk_boot_earlytab':
  0x001621b8 10861200 00000000 c0851200 00000000 ................
``` 
It's more of a matter of habit / good practice in case we need to extend later. Saying that, I don't have an extension in mind atm, so if you have a strong preference we can easily switch to functions.

skuenzer: This is actually true, since it is a single-field `struct` with a pointer, you should be able to cast it to a function pointer even: no real difference in memory. I am also undecided which version is better, normally I tend to minimalism but here we have zero runtime overhead due to the `struct`. Could be worth keeping it. 🤔 Let's keep it, as is.

lib/ukboot/boot.c
```diff
@@ -238,8 +238,7 @@ static struct uk_alloc *heap_init()
 extern char **environ;
 #endif /* CONFIG_LIBPOSIX_ENVIRON */
 
-/* defined in <uk/plat.h> */
-void ukplat_entry(void)
+void uk_boot_entry(void)
```
skuenzer, line 241: There is a `_` too much in the commit title in `uk_plat_entry`, should be `ukplat_entry`.

michpappas:
Rebased from `staging` to pick up #1471

lib/ukboot/boot.c
michpappas: Actually `_ukplat_entry` is correct, I am referring to the platform entry, not the boot entry: https://github.com/unikraft/unikraft/blob/staging/plat/kvm/x86/setup.c#L72 🙂

lib/ukboot/boot.c
```diff
@@ -443,7 +396,8 @@ void ukplat_entry(int argc, char *argv[])
 
 #else /* CONFIG_LIBUKBOOT_MAINTHREAD */
 	m = uk_sched_thread_create_fn2(s, main_thread,
-				       (void *)((long)argc), (void *)argv,
+				       (void *)((long)boot_argc),
```
skuenzer: Hum, shouldn't it be `ictx.cmdline.argc` and `ictx.cmdline.argv` here?

skuenzer: The idea was that init functions would be able to replace/modify argc and argv.

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,150 @@
...
+/* Copy of bi->cmdline, as the boot protocol buffer may
+ * be unmapped by paged memory init.
+ */
+static char cmdline[CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN];
+static __sz cmdline_len;
+
+/* Mutable version of the cmdline. This is tokenized by uk_argnparse().
+ * Keep this as a global as `argv` points to tokens in this string.
+ */
+static char cmdline_argv[CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN];
```
skuenzer: Could we find a different name than `cmdline_argv`. With `_argv` I am biased to read `char *argv[]` 🤔

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	/* Update bi->cmdline */
+	strncpy(cmdline, cmdl, ARRAY_SIZE(cmdline));
+	cmdline[cmdline_len] = 0;
+	bi->cmdline = (__u64)cmdline;
+
+	/* Break down cmdline to argv */
+	strncpy(cmdline_argv, cmdl, ARRAY_SIZE(cmdline));
+	cmdline_argv[cmdline_len] = 0;
```
skuenzer: I'd prefer to stay with the target data type: Can we place `'\0'` here, instead of `0`?

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,150 @@
...
+		return rc;
+	}
+
+	/* Update bi->cmdline */
+	strncpy(cmdline, cmdl, ARRAY_SIZE(cmdline));
+	cmdline[cmdline_len] = 0;
+	bi->cmdline = (__u64)cmdline;
+
+	/* Break down cmdline to argv */
+	strncpy(cmdline_argv, cmdl, ARRAY_SIZE(cmdline));
```
skuenzer: Shouldn't the size be derived from `cmdline_argv`?: `ARRAY_SIZE(cmdline_argv)`

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if (unlikely(cmdline_len > ARRAY_SIZE(cmdline))) {
+		rc = -E2BIG;
+		uk_pr_err("Could not parse cmdline (%d)\n", rc);
+		return rc;
+	}
+
+	/* Update bi->cmdline */
+	strncpy(cmdline, cmdl, ARRAY_SIZE(cmdline));
```
skuenzer: Why do we do two copies of `cmdl`? Wouldn't `boot_argv` and `boot_argc` be good enough to keep only?

lib/ukboot/early_init.c
skuenzer: Plus, see my questions below.

lib/ukboot/boot.c
skuenzer: What I meant is the title of this commit that does this change:
```
lib/ukboot: Rename uk_plat_entry() to uk_boot_entry() 
```
Shouldn't it be?:
```
lib/ukboot: Rename ukplat_entry() to uk_boot_entry() 
```

lib/ukboot/include/uk/boot.h
```diff
@@ -19,14 +19,25 @@ extern "C" {
...
 void uk_boot_early_init(struct ukplat_bootinfo *bi);
 
+/**
+ * Unikraft boot entry
+ *
+ * This function is called directly by platform libraries for
+ * bootstrapping. Before calling this function a platform must
+ * have called uk_boot_early_init().
+ *
+ * Notice: Once all platform code registers to earlyrtab this
```
skuenzer: - Typo: `earlyrtab`
- I would change `Notice:` to `TODO:` or `NOTE:`

plat/kvm/arm/setup.c
```diff
@@ -231,5 +194,6 @@ void __no_pauth _ukplat_entry(void)
 	 */
 	uk_pr_info("Switch from bootstrap stack to stack @%p\n", bstack);
 
-	lcpu_arch_jump_to(bstack, _ukplat_entry2);
+	lcpu_arch_jump_to(bstack, uk_boot_entry);
```
skuenzer, line 197: I guess all calls to `uk_boot_entry` imply adding `lib/ukboot` as dependency (`select`) to the corresponding platform's `Config.uk`.

plat/xen/arm/setup64.c
```diff
@@ -31,6 +31,7 @@
 #include <common/events.h>
 #include <common/console.h>
 #include <xen/xen.h>
+#include <uk/boot.h>
```
skuenzer, line 34: Same here: A `lib/ukboot` dependency

plat/xen/x86/setup.c
```diff
@@ -71,6 +71,7 @@
 #include <uk/config.h>
 #include <uk/print.h>
 #include <uk/assert.h>
+#include <uk/boot.h>
```
skuenzer, line 74: Another `lib/ukboot` dependeny for `Config.uk`

lib/ukboot/boot.c
michpappas: That one yes 👍🏼

lib/ukboot/early_init.c
michpappas: `cmdline` is required for preserving `bi->cmdline` as the boot protocol originally points `bi->cmdline` to memory that, depending on the boot protocol, it may get unmapped later (and we print it with `ukplat_bootinfo_print()`). This is not new, we were doing it already in platform code for all platforms.

`cmdline_argv` is used for parsing, as it is updated in-place with `\0` by `uk_argnparse()`. This is not new either, it is just that now the two copies stand out as they're centralized in `early_init()`.

See also: https://github.com/unikraft/unikraft/pull/1472/commits/57aaa2df8d9335b04f17839968b7aefdcefce9b8#diff-02014d06ce1464d989f9bc78a64dc2159ff644d8cc71f1882a75730c55076895R21

lib/ukboot/early_init.c
michpappas: Will rename to `cmdline_argparse`

plat/xen/arm/setup64.c
```diff
@@ -167,35 +168,23 @@ static void *map_fdt(paddr_t device_tree)
...
+	bi->cmdline_len = sizeof(CONFIG_UK_NAME) - 1;
 
 	fdtchosen = fdt_path_offset(HYPERVISOR_dtb, "/chosen");
 	if (fdtchosen) {
 		fdtcmdline = fdt_getprop(HYPERVISOR_dtb, fdtchosen, "bootargs",
 					 &len);
 		if (fdtcmdline) {
-			cmdl = fdtcmdline;
-			cmdline_len = len;
+			bi->cmdline = fdtcmdline;
```
oleksiimoisieiev, line 184: I think this a bit unsafe. fdt_getprop call is used to get fdtcmdline property. Which returns address of the fdtproperty->data member. fdt_getprop is the part of fdt_ro, which means that this data is ro. So if we put this pointer to __u64 cmdline in the bi structure - there is a potential risk that this memory may be changed which may lead to unexpected consequences, such as wrong cmdline in the device tree blob or even fdt corruption.
That's why cmdline was copied on the previous version.
I think in this case either copy of cmdline should be done or __u64 cmdline in bi structuure should be const. What do you think?

plat/xen/x86/setup.c
```diff
@@ -183,63 +184,32 @@ static int _init_mem(struct ukplat_bootinfo *const bi)
...
 
 	memcpy(bi->bootloader, bl, sizeof(bl));
-
 	memcpy(bi->bootprotocol, bp, sizeof(bp));
 
 	if (_init_mem(bi) < 0)
 		UK_CRASH("Failed to initialize memory\n");
 
-	_libxenplat_x86bootinfo_setup_cmdl(bi);
+	bi->cmdline_len = strlen((char *)HYPERVISOR_start_info->cmd_line);
```
oleksiimoisieiev, line 200: I would use strnlen((char *)HYPERVISOR_start_info->cmd_line, CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN);

plat/xen/x86/setup.c
```diff
@@ -183,63 +184,32 @@ static int _init_mem(struct ukplat_bootinfo *const bi)
...
+	bi->cmdline = (__u64)HYPERVISOR_start_info->cmd_line;
 }
 
 void _libxenplat_x86entry(void *start_info) __noreturn;
 void _libxenplat_x86entry(void *start_info)
 {
+	struct ukplat_bootinfo *bi;
+
+	bi = ukplat_bootinfo_get();
```
oleksiimoisieiev, line 209: `struct ukplat_bootinfo *bi = ukplat_bootinfo_get();` ??

lib/ukboot/boot.c
```diff
@@ -403,8 +356,8 @@ void ukplat_entry(int argc, char *argv[])
 	uk_sched_start(s);
 #endif /* CONFIG_LIBUKBOOT_INITSCHED */
 
-	ictx.cmdline.argc = argc;
-	ictx.cmdline.argv = argv;
+	ictx.cmdline.argc = boot_argc;
```
oleksiimoisieiev, line 364: Probably some check needs to be here to ensure that boot_argv and boot_argc are already processed at this stage.

lib/ukboot/early_init.c
```diff
@@ -0,0 +1,150 @@
...
+	cmdline[cmdline_len] = 0;
+	bi->cmdline = (__u64)cmdline;
+
+	/* Break down cmdline to argv */
+	strncpy(cmdline_argparse, cmdl, ARRAY_SIZE(cmdline_argparse));
+	cmdline_argparse[cmdline_len] = '\0';
```
oleksiimoisieiev: As I can see from the code - cmdline_argparse is not used. Here are search results:
```
lib/ukboot/early_init.c:static char cmdline_argparse[CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN];
lib/ukboot/early_init.c:        strncpy(cmdline_argparse, cmdl, ARRAY_SIZE(cmdline_argparse));
lib/ukboot/early_init.c:        cmdline_argparse[cmdline_len] = '\0';
````

oleksiimoisieiev (changes requested):

plat/xen/arm/setup64.c
michpappas: You're right, and as a matter of fact we do copy the cmdline, but now we have formalized that behavior for all platforms (as every platform was copying its cmdline anyway) at the implementation of `uk_boot_early_init_parse_cmdl()`: https://github.com/unikraft/unikraft/pull/1472/commits/9930d23655e811a430ba49b8d3f67a58df8eba19#diff-02014d06ce1464d989f9bc78a64dc2159ff644d8cc71f1882a75730c55076895R98.
So essentially nothing has changed.

plat/xen/x86/setup.c
michpappas: That's a good point, however we don't export `CONFIG_LIBUKBOOT_CMDLINE_MAX_LEN` outside `libukboot`. What we do instead is expect that the boot protocol sets `bi->cmdline` and `bi->cmdline_len` to memory it controls, and manage the copy - including the check to the size of the buffers controlled by libukboot - internally in `libukboot`. That provides some separation between the responsibilities of each component involved with boot, and avoids having the boot protocol make any assumptions about `libukboot` internals.

plat/xen/x86/setup.c
michpappas: That's because of the coding convention discouraging this style: https://unikraft.org/docs/contributing/coding-conventions#variable-definitions

lib/ukboot/boot.c
michpappas: Good point, I'll add an assertion 👍🏼

lib/ukboot/early_init.c
michpappas: Thanks for catching that, it should be passed to `uk_argnparse()`. This is a regression introduced in the last update of the PR. Will fix 👍🏼

andreistan26 (approved):
After all the heavy lifting was done, nothing new from my side, additions look good!

Reviewed-by: Andrei Stan <andreistan2003@gmail.com>

michpappas:
@oleksiimoisieiev I addressed your comments (see replies inline as well as https://github.com/unikraft/unikraft/compare/8babcb3e87b5df4ff540f0bdb020ea0ab70fb2e3..b11bdd5176da6dc2d58418ac30360a1fd5595208) and additionally fixed a bug during parsing that caused incorrect arguments passed to the application.

oleksiimoisieiev:
LGTM 

Reviewed-by: Oleksii Moisieiev <oleksii_moisieiev@epam.com>

oleksiimoisieiev (approved):

lib/ukboot/early_init.c
skuenzer: Okay, the question is then rather if there is a use case for keeping `bi->cmdline`? However, I think we can resolve that in a different PR.

To explain a bit better my motivation: if we would provide an ukargparse variant that would work cpy and parse a buffer, we could use that function directly to read from the original cmdline and copy the parsed arguments directly over.

michpappas: @skuenzer, you right.  What I propose is:

1. Formalize the call to `ukplat_bootinfo_print()` as Unikraft's "hello" by moving it to early_init before we iterate the earlytab. With the addition of the logbuf it will be printed once the console is up.
2. Set `bi->cmdline` to NULL as soon as we have parsed the cmdline. That will prevent manual calls to `ukplat_bootinfo_print()` - often used for debugging mrds - from causing an abort.
3. Remove the immutable copy of the cmdline.

I pushed an update with these changes.

skuenzer (approved):
Great work! Thanks a lot!

PR #1464, opened 2024-06-25
lib/ukconsole: Add generic console device interface

thass0:
This PR introduces `ukconsole`, a library that exposes a generic interface for console devices. This is groundwork for adding the gdb stub to Unikraft.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

Changes are tested on KVM using QEMU (both x86_64 and ARM64) and on Xen (x86_64).

### Base target

 - Architecture(s): N/A
 - Platform(s): `kvm`, `xen`
 - Application(s): N/A

### Additional configuration

The `app-helloworld` and `catalog/library/nginx/1.25` projects should both build without changing the configuration

### Description of changes

From the perspective of user libraries, all "print" and "read input" operations for consoles now go through the `ukconsole` interface (`console.h`). Before, the `ukplat_[cout|cin][k|d]` functions were used for that purpose. These `ukplat` functions differentiated between kernel and debug printing. This difference is gone with `ukconsole`. kernel/debug printing wasn't really used so the difference will be phased out over time.

Drivers can register themselves with `ukconsole` in one of two ways: statically or dynamically. All statically registered device drivers are initialized at once with the `uk_console_early_init` function. Dynamic registration can happen at any point in time. `ukconsole` expects a dynamically registered device to be initialized before it's registered. The distinction between dynamically and statically registered devices is useful because some device drivers (e.g., PL011, NS16550) need information that becomes available later during boot before they can be initialized. Other drivers (e.g., COM) have no such prerequisites and can be initialized very early on.

The PL011 and NS16550 drivers are adapted to use `ukconsole`. The COM driver and the VGA driver from `plat/kvm/` were moved to `drivers/uktty` and also now use `ukconsole`. The Xen console drivers were moved to `drivers/xen` and updated to use `ukconsole`.

Note that some changes introduced early on are removed in later commits. Namely, the dynamic registration mechanism is simplified a lot later on. The string console device names used at the start are also removed entirely later. 

GitHub-Depends-On: #1473

PR #1473, opened 2024-07-18
drivers/uktty: Improve early initialization

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`arm64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series various improvement on the early initialization of `ns16550` and `pl011`:
- Register with `uk_boot_earlytab`.
- Set up the console port at `early_init()`.
- Improve port setup on `ns16550`.
- Add runtime configuration: Systems can now configure the base address via a cmdline options. If one is not set, the drivers will attempt to discover the early console via dtb's `/chosen/stdout-path`.
- Various cleanups.

GitHub-Depends-On: #1471 
GitHub-Depends-On: #1472


lib/ukofw/fdt.c
```diff
@@ -391,3 +392,37 @@ bool fdt_prop_read_bool(const void *fdt, int start_offset,
 
 	return prop ? true : false;
 }
+
+int fdt_chosen_stdout_path(const void *fdt, int *offs, char **opt)
+{
+	const char *prop;
+	int prop_len, len;
```
mogasergiu, line 399: :christmas_tree: I know you like it :pleading_face:

lib/ukofw/fdt.c
```diff
@@ -391,3 +392,37 @@ bool fdt_prop_read_bool(const void *fdt, int start_offset,
...
+
+	nchosen = fdt_path_offset((void *)fdt, "/chosen");
+	if (unlikely(nchosen < 0)) {
+		uk_pr_err("Could not find 'chosen' node (%d)\n", nchosen);
+		return -FDT_ERR_NOTFOUND;
+	}
+
+	prop = fdt_getprop((void *)fdt, nchosen, "stdout-path",
+			   &prop_len);
+	if (unlikely(!prop || prop_len <= 0))
```
mogasergiu: Maybe a `Could not find 'stdout-path' property` error here as well just like above?

drivers/uktty/ns16550/ns16550.c
```diff
@@ -72,6 +73,10 @@
 #define NS16550_LSR_RX_EMPTY	0x01U
 #define NS16550_LSR_TX_EMPTY	0x40U
 
+static const char * const fdt_compatible[] = {
```
mogasergiu, line 76: `*const` no whitespace ? :thinking:

drivers/uktty/ns16550/ns16550.c
```diff
@@ -169,14 +178,60 @@ static int init_ns16550(void)
 }
 
 #if CONFIG_LIBUKTTY_NS16550_EARLY_CONSOLE
+static inline int config_fdt_chosen_stdout(void *dtb)
```
mogasergiu: `const void *dtb`?

drivers/uktty/pl011/pl011.c
```diff
@@ -110,21 +123,67 @@ static int init_pl011(void)
 }
 
 #if CONFIG_LIBUKTTY_PL011_EARLY_CONSOLE
+static inline int config_fdt_chosen_stdout(void *dtb)
```
mogasergiu: `const void *dtb`

mogasergiu (changes requested):
LGTM. As a side note, I noticed that in the prereq PRs as well as this one you are printing the error code. I am wondering, should we start printing the corresponding error string as well? For example: `printf("Failed to do X: %s %d\n", strerror(-rc), rc)`

michpappas:
> As a side note, I noticed that in the prereq PRs as well as this one you are printing the error code. I am wondering, should we start printing the corresponding error string as well? For example: printf("Failed to do X: %s %d\n", strerror(-rc), rc)

That's right, I think it's a good pattern to print the error code along with the error. Being able to use `strerror()` would be even better, but the one from stdlib sets `errno` on error, and therefore we can't use it before we run under a thread. It would be nice however to have a errno-free version for the kernel, similarly to the `strntoull()` implementation of `libukparam`.

lib/ukofw/fdt.c
michpappas: Okay, I'll add it for consistency although in practice it's unlikely to see anything given the nature of the function (looking for a way to print). It has to be a warning though as according to the spec the `stdout-path` is optional.

drivers/uktty/ns16550/ns16550.c
michpappas: Yes, if it was a variable name it wouldn't require space.

lib/ukofw/fdt.c
```diff
@@ -391,3 +392,39 @@ bool fdt_prop_read_bool(const void *fdt, int start_offset,
...
+	prop = fdt_getprop((void *)fdt, nchosen, "stdout-path",
+			   &prop_len);
+	if (unlikely(!prop || prop_len <= 0)) {
+		uk_pr_warn("Could not find 'stdout-path' property\n");
+		return -FDT_ERR_NOTFOUND;
+	}
+
+	p = strchr(prop, ':');
+
+	*opt = p ? p + 1 : NULL;
```
mogasergiu, line 422: This is just a nitpick and most likely just my personal taste. But I feel like
```C
if (p) {
   *opt = p + 1;
   len = p - prop;
} else {
   *opt = NULL;
   len = prop_len - 1;
}
```
would be easier to follow and maybe more pleasant during a GDB debugging session if someone happens to have to go through here. This also makes it easier to add individual prints for the two branches so that one may be able to visually debug logs, I believe.

drivers/uktty/ns16550/ns16550.c
```diff
@@ -29,6 +29,7 @@
  */
 #include <libfdt.h>
 #include <uk/assert.h>
+#include <uk/boot/earlytab.h>
```
mogasergiu: Does this add ukboot dep?

drivers/uktty/ns16550/ns16550.c
```diff
@@ -120,6 +133,13 @@ static void ns16550_reg_write(__u32 reg, __u32 value)
 
 static int init_ns16550(void)
 {
+	__u32 lcr_mask;
+
+	/* Clear DLAB to access IER, FCR, LCR */
+	ns16550_reg_write(NS16550_LCR_OFFSET,
+			  ns16550_reg_read(NS16550_LCR_OFFSET) &
+			  ~(NS16550_LCR_DLAB));
```
mogasergiu, line 148: What is this paranthesis for? :thinking: Oh wait... those below have it too :confused:

drivers/uktty/ns16550/ns16550.c
```diff
@@ -129,6 +149,22 @@ static int init_ns16550(void)
...
+
+	/* Set line control parameters (8n1) */
+	lcr_mask = NS16550_LCR_WL | NS16550_LCR_STOP | NS16550_LCR_PARITY;
+	ns16550_reg_write(NS16550_LCR_OFFSET,
+			  ns16550_reg_read(NS16550_LCR_OFFSET) & ~(lcr_mask));
+	ns16550_reg_write(NS16550_LCR_OFFSET, NS16550_LCR_8N1);
+
+	/* Set DLAB to access DLL / DLM */
+	ns16550_reg_write(NS16550_LCR_OFFSET,
+			  ns16550_reg_read(NS16550_LCR_OFFSET) &
```
mogasergiu, line 168: I may be missing something but why do we read this register twice and write three times? Are you sure this couldn't have been shortened to a read and a write? This also reminds me of the compile-time barriers missing.

drivers/uktty/ns16550/Config.uk
```diff
@@ -1,16 +1,16 @@
 menuconfig LIBUKTTY_NS16550
 	bool "NS16550 / 8250"
-	default n
 	depends on ARCH_ARM_64
+	select LIBUKOFW
```
mogasergiu, line 4: In the commit title: s/15550/16550

drivers/uktty/ns16550/ns16550.c
```diff
@@ -169,14 +178,60 @@ static int init_ns16550(void)
...
 
 	UK_ASSERT(bi);
 
-	ns16550_uart_base = CONFIG_LIBUKTTY_NS16550_EARLY_CONSOLE_BASE;
+	/* If the base is not set by the cmdline, try
+	 * the dtb chosen/stdout-path.
+	 */
+	if (!ns16550_uart_base) {
+		rc = config_fdt_chosen_stdout((void *)bi->dtb);
+		if (unlikely(rc < 0))  { /* libfdt error */
```
mogasergiu: Ditto

drivers/uktty/ns16550/ns16550.c
```diff
@@ -83,6 +88,10 @@
 static __u8 ns16550_uart_initialized;
 static __u64 ns16550_uart_base;
 
+#if CONFIG_LIBUKLIBPARAM
+UK_LIBPARAM_PARAM_ALIAS(base, &ns16550_uart_base, __u64, "ns15550 base");
```
mogasergiu: How does someone know about this? Should there maybe be a kconfig hint?

drivers/uktty/pl011/pl011.c
```diff
@@ -21,6 +21,7 @@
 #include <libfdt.h>
 #include <uk/assert.h>
 #include <uk/bitops.h>
+#include <uk/boot/earlytab.h>
```
mogasergiu: Ditto ukboot dep?

drivers/uktty/ns16550/ns16550.c
```diff
@@ -142,6 +142,13 @@ static int early_init(struct ukplat_bootinfo *bi)
 
 	ns16550_uart_base = CONFIG_LIBUKTTY_NS16550_EARLY_CONSOLE_BASE;
 
+	/* Configure the port */
+	rc = init_ns16550();
+	if (unlikely(rc)) {
+		uk_pr_err("Could not initialize ns16550 (%d)\n", rc);
```
mogasergiu, line 233: Wait so we print on failing to initialize console?

drivers/uktty/pl011/pl011.c
```diff
@@ -119,6 +119,13 @@ static int early_init(struct ukplat_bootinfo *bi)
 
 	pl011_base = CONFIG_LIBUKTTY_PL011_EARLY_CONSOLE_BASE;
 
+	/* Configure the port */
+	rc = init_pl011();
+	if (unlikely(rc)) {
+		uk_pr_err("Could not initialize pl011 (%d)\n", rc);
```
mogasergiu: Ditto

mogasergiu (changes requested):

drivers/uktty/ns16550/ns16550.c
michpappas: Good question. In theory every library should export a `README.md` and these should be automatically collected along with doxygen to autogenerated documentation. Also before https://github.com/unikraft/unikraft/pull/1471 `libuklibparam` printed usage info upon parse error or coming across the parameter `help`. Of course with early cmdline this creates a circular dependency with console, but perhaps we could keep a flag in ukboot and print these after the console is initialized, at least until we implement a console buffer. Assuming successful console initialization, that is. I'll push an update in https://github.com/unikraft/unikraft/pull/1471.

Update: I changed my mind, as such a change would need to be reverted once we add console buffering. Let's leave it as is.

drivers/uktty/ns16550/ns16550.c
michpappas: In https://github.com/unikraft/unikraft/pull/1472#discussion_r1694224280 you ask for an error message before the console is initialized, so which one is it? 😉 Either way, we could also state that we make no assumptions of the overall system's state at the library level (e.g. imagine VGA was already on) and print messages on all errors.

lib/ukofw/fdt.c
michpappas: I think it's a matter of personal taste, and as a matter of fact I find the terse notation cleaner.

drivers/uktty/ns16550/ns16550.c
michpappas: That's a big question as it applies to all libraries initialized via `libukboot`, and similarly to `libukinit` in the future. One option is to wrap this header as well as the registration with `#if CONFIG_LIBUKBOOT` without making the dependency to `libukboot` mandatory in `Config.uk`. On the other hand `libukboot` is so integral that it would be safe to assume that it is always enabled (or implemented as an external library that replaces the native one). @skuenzer thoughts?

lib/ukofw/fdt.c
mogasergiu: Ok. Just thought that someone may easily lose sight of the fact that `len` depends on the state of `p` after the `*opt` statement, instead of its state before that (I personally did).

drivers/uktty/ns16550/ns16550.c
mogasergiu: Yes this makes sense. This could be a nice track after we have the ISR plugin work in. We could extend that to check additional things such as this. Especially once we will have KASLR in and we need a clearly defined section for relocations before/after KASLR.

drivers/uktty/ns16550/ns16550.c
mogasergiu: > you ask

u right i lost sight of that

>  make no assumptions of the overall system's state at the library level

ok sure

drivers/uktty/ns16550/ns16550.c
michpappas: The value of the `LCR.DLAB` affects how registers are interpreted [1], so I would keep the operations that update DLAB separate from those that update the rest of the LCR bits. For the other two writes that set the line control you're right, I'll merge them into one.

[1] https://media.digikey.com/pdf/Data%20Sheets/Texas%20Instruments%20PDFs/PC16550D.pdf

michpappas:
Added a commit to enable earlycon on kvm/arm64.

michpappas:
Rebased to the latest state of #1471, #1472

mogasergiu (approved):
Other than the ukboot dep [comment](https://github.com/unikraft/unikraft/pull/1473#discussion_r1694648748) waiting on @skuenzer, all fine by me.

michpappas:
Rebased from #1472. No functional changes.

michpappas:
@mogasergiu I pushed two more patches to get rid of compiled-in parameters.

drivers/uktty/ns16550/ns16550.c
skuenzer: Agree with Michalis' comment. I imagine that with the `ukprint` rework we will have a kernel message buffer. This would in principle enable that we can print out the buffer buffer later on a different device and would see this message.

drivers/uktty/ns16550/ns16550.c
skuenzer: I think we should add `lib/ukboot` as a dependency as soon as we need headers from it. This is the case when the driver configuration enables early console mode. Otherwise such a `select` should be avoided.

drivers/uktty/pl011/pl011.c
skuenzer: Agree, same as mentioned earlier. Basically:
```
menuconfig LIBUKTTY_PL011_EARLY_CONSOLE
	bool "Early console"
	select CONFIG_LIBUKBOOT
```

lib/ukofw/include/uk/ofw/fdt.h
```diff
@@ -187,6 +187,22 @@ int fdt_node_offset_by_compatible_list(const void *fdt, int startoffset,
 int fdt_node_offset_idx_by_compatible_list(const void *fdt, int startoffset,
 				const char * const compatibles[], int *index);
 
+/**
+ * fdt_node_check_compatible_list - checks whether a node's 'compatible'
+ *                                  property matches a given list of
```
thass0, line 192: Seems like this sentence is unfinished

mogasergiu (approved):
Re-approving as it seems the ukboot dep comment has been addressed.

lib/ukofw/include/uk/ofw/fdt.h
michpappas: Ooops, fixed 👍🏼

michpappas:
Rebased from `staging`

thass0:
Reviewed-by: Thassilo Schulze <thassilo@unikraft.io>

drivers/uktty/ns16550/ns16550.c
```diff
@@ -29,35 +29,54 @@
  */
 #include <libfdt.h>
 #include <uk/assert.h>
+#include <uk/boot/earlytab.h>
 #include <uk/config.h>
 #include <uk/init.h>
+#include <uk/ofw/fdt.h>
 #include <uk/plat/common/bootinfo.h>
 
+#if CONFIG_LIBUKLIBPARAM
```
skuenzer: Those guards aren't needed because `lib/uklibparam` is providing the header always and switches configuration of the registration macros. I am not mentioning the other places where this is redundant as well, but you could remove the guards everywhere for improving readability.

drivers/uktty/ns16550/Config.uk
```diff
@@ -2,6 +2,7 @@ menuconfig LIBUKTTY_NS16550
 	bool "NS16550 / 8250"
 	default n
 	depends on ARCH_ARM_64
+	select LIBUKBOOT
```
skuenzer: I am not fully getting why the dependency is declared here and not in the previous commit where the header of libukboot is included that causes that dependency. By mistake? 🤔

drivers/uktty/ns16550/Config.uk
```diff
@@ -1,29 +1,12 @@
 menuconfig LIBUKTTY_NS16550
 	bool "NS16550 / 8250"
-	default n
 	depends on ARCH_ARM_64
+	select LIBUKBOOT
+	select LIBUKOFW
 
 if LIBUKTTY_NS16550
+
 menuconfig LIBUKTTY_NS16550_EARLY_CONSOLE
```
skuenzer: Since there aren't any submenu configs, we could turn this into `config`. But please ignore this comment if we plan to introduce sub options again with PR #1464 or the PR merging x86 and arm console drivers with NS16550.

drivers/uktty/ns16550/ns16550.c
```diff
@@ -69,11 +88,15 @@
 static __u8 ns16550_uart_initialized;
 static __u64 ns16550_uart_base;
 
+#if CONFIG_LIBUKLIBPARAM
+UK_LIBPARAM_PARAM_ALIAS(base, &ns16550_uart_base, __u64, "ns15550 base");
```
skuenzer: Is here a typo in the help? `ns16550 base`?
Btw, this PR removes the capability to use the driver without fdt and without kernel command line parser. Wouldn't it be useful to also have the option to configure defaults for that scenario?

drivers/uktty/ns16550/Config.uk
```diff
@@ -1,29 +1,12 @@
...
 	bool "Early console"
```
skuenzer, line 9: Two questions:
- I think we should provide a help description to this option (describing how it works and how the device is detected via fdt/kernel cmdline)?
- Can we introduce the `lib/ukboot` dependency only when this option is enabled. Otherwise we don't need `lib/ukboot` for the driver, correct?

drivers/uktty/pl011/Config.uk
```diff
@@ -1,14 +1,12 @@
 menuconfig LIBUKTTY_PL011
 	bool "PrimeCell UART (PL011)"
 	depends on ARCH_ARM_64
+	select LIBUKBOOT
+	select LIBUKOFW
 
 if LIBUKTTY_PL011
+
 menuconfig LIBUKTTY_PL011_EARLY_CONSOLE
```
skuenzer: Same here, could become `config` when the base address configuration is removed.

drivers/uktty/pl011/Config.uk
```diff
@@ -1,14 +1,12 @@
...
 	bool "Early console"
```
skuenzer, line 9: Here also the two questions similar to the ones to ns16550:
- I think we should provide a help description to this option (describing how it works and how the device is detected via fdt/kernel cmdline)?
- Can we introduce the `lib/ukboot` dependency only when this option is enabled. Otherwise we don't need `lib/ukboot` for the driver, correct?

drivers/uktty/pl011/Config.uk
```diff
@@ -5,13 +5,8 @@ menuconfig LIBUKTTY_PL011
 	select LIBUKOFW
 
 if LIBUKTTY_PL011
+
 menuconfig LIBUKTTY_PL011_EARLY_CONSOLE
 	bool "Early console"
 
-if LIBUKTTY_PL011_EARLY_CONSOLE
-config LIBUKTTY_PL011_EARLY_CONSOLE_BASE
```
skuenzer, line 10: Similar to my comment to ns16550: Do we really want to remove option to compile in a default for ultra minimal builds without fdt and without kernel command line parameters?

drivers/uktty/ns16550/ns16550.c
michpappas: I hadn't noticed it's always on, thanks for that 👍🏼

drivers/uktty/ns16550/ns16550.c
michpappas: > Is here a typo in the help? ns16550 base?

Of course 👍🏼 

> Btw, this PR removes the capability to use the driver without fdt and without kernel command line parser. Wouldn't it be useful to also have the option to configure defaults for that scenario?

That's right. Supporting boot without fdt would require that we provide Kconfig options for all drivers and for all their config options, the amount of config options introduced and their impact in the code would be pretty high. IMO this is not worth to support as the device-tree is the standard (and only) way to boot on arm-based embedded systems, that this requirement shouldn't come as a surprise. So unless we have a real use-case in mind, I would vouch for keeping things simple. What do you think?

drivers/uktty/ns16550/Config.uk
michpappas: Good point. I think we should do it correctly here, regardless of how things may evolve in the (near) future.

drivers/uktty/pl011/Config.uk
michpappas: Answered in https://github.com/unikraft/unikraft/pull/1473#discussion_r1726144448

drivers/uktty/ns16550/Config.uk
michpappas: Most likely accidentally squashed to the wrong commit. Will fix 👍🏼

drivers/uktty/ns16550/Config.uk
michpappas: > I think we should provide a help description to this option (describing how it works and how the device is detected via fdt/kernel cmdline)?

Same as above, I would treat the fdt requirement in the same scope as the rest of the drivers (i.e. no need to state in the `help` of every driver that it is configured via the fdt, but instead document somewhere globally that arm64 systems are configured via the fdt).

Ditto for the cmdline: Currently the only way to figure out the available parameters is to pass `help`. As we move forward, these could also be documented in each library's `README.md`.

> Can we introduce the lib/ukboot dependency only when this option is enabled. Otherwise we don't need lib/ukboot for the driver, correct?

That's right 👍🏼

drivers/uktty/pl011/Config.uk
michpappas: > I think we should provide a help description to this option (describing how it works and how the device is detected via fdt/kernel cmdline)?

Answered in https://github.com/unikraft/unikraft/pull/1473#discussion_r1726161076

> Can we introduce the lib/ukboot dependency only when this option is enabled. Otherwise we don't need lib/ukboot for the driver, correct?

Correct 👍🏼

drivers/uktty/ns16550/ns16550.c
skuenzer: I am more thinking towards virtualized environments, especially less feature-rich VMMs and in particular I am thinking of firecracker. Does it use fdts for booting? If so, then it is fine to take that as assumption for Arm.

drivers/uktty/ns16550/Config.uk
skuenzer: > Same as above, I would treat the fdt requirement in the same scope as the rest of the drivers (i.e. no need to state in the help of every driver that it is configured via the fdt, but instead document somewhere globally that arm64 systems are configured via the fdt).

What I meant is that the help for the `Early console` option explains a bit which fdt node is checked for enabling early console support because it looked to me a bit special for the early console device with `/chosen/stdout-path`.

drivers/uktty/ns16550/ns16550.c
michpappas: Yes, FC on arm64 also passes a device-tree blob it generates, and similarly for `kvm-tool` I am not aware of a VMM that doesn't provide an fdt for arm64.

drivers/uktty/ns16550/Config.uk
michpappas: Done

skuenzer (approved):
Thanks a lot for this work. This is really improving Unikraft's boot code and brings us closer to the state that we want to reach with platform-rearch.

drivers/uktty/ns16550/ns16550.c
skuenzer: Okay, fair enough. Makes sense. :+1:

skuenzer (approved):

PR #1474, opened 2024-07-31
lib/nolibc: Fix memory bug in vfscanf

thass0:
It wasn't guaranteed that the `buf` buffer would be null-terminated in all cases. The fix makes sure that there is always as null byte at the end.

### Prerequisite checklist

### Base target

### DIFF
--- a/lib/nolibc/scanf.c
+++ b/lib/nolibc/scanf.c
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 #include <stdarg.h>
+#include <errno.h>
 
 #include <uk/plat/console.h>
 
@@ -16,7 +17,14 @@ uk_scanf(void *buffer,  size_t *cnt)
 	size_t bytes_total = 0, count;
 	char *buf = buffer;
 
-	count = *cnt;
+	/* Need at least two bytes: one for user input and one for
+	 * the NULL termination. The byte for user input is required,
+	 * otherwise we run into an endless loop on `ukplat_cink`.
+	 */
+	if (*cnt <= 1)
+		return -EINVAL;
+
+	count = *cnt - 1;
 
 	do {
 		while ((bytes_read = ukplat_cink(buf,
@@ -46,6 +54,7 @@ uk_scanf(void *buffer,  size_t *cnt)
 	} while (bytes_total < count &&
 	((bytes_total == 0) || (*(buf - 1) != '\n' && *(buf - 1) != '\0')));
 
+	((char *)buffer)[bytes_total] = '\0';
 	*cnt = bytes_total;
 
 	return 0;


michpappas (changes requested):
Good catch 💯 Perhaps it would be reasonable for the termination to be performed by the function that reads the input.

thass0:
> Good catch 💯 Perhaps it would be reasonable for the termination to be performed by the function that reads the input.

I agree. I'm moving to to `uk_scanf`. I added an assertion in `uk_scanf`. The condition that's asserted should always hold given that `count` is initially assigned `*cnt` and not touched after that and the `while`-loop makes sure `bytes_total < count`. But better to assert this before writing out of bounds of the buffer.

thass0:
You're right, I moved the NULL-termination into `uk_scanf` itself. The assertion that I added should always hold given how `uk_scanf` is coded right now. I just added it to make sure that if something changes in the implementation of `uk_scanf`, we don't accidentally write outside the buffer.

michpappas:
> You're right, I moved the NULL-termination into `uk_scanf` itself. The assertion that I added should always hold given how `uk_scanf` is coded right now. I just added it to make sure that if something changes in the implementation of `uk_scanf`, we don't accidentally write outside the buffer.

Regarding the assertion, I would feel better if we had confidence that the implementation is correct. Is that the case? Moreover, with the proposed change we overwrite the last character passed by the user, if the user writes the maximum number of characters. Perhaps it would be better if we reserved the last byte in the first place, and allow the user to send `cnt - 1` characters?

thass0:
I pushed another update. Now we reserve a byte at the end. I also removed the assertion. The logic is this: the `cnt` parameter contains the number of bytes available in the buffer. When the `do { ... } while` loop finishes, `bytes_total` is at most equal to `cnt - 1`. So it's safe to use `bytes_total` to index into `buffer` (which is `cnt` bytes in size). `bytes_total` can't be bigger than `cnt - 1` when the loops finishes because each iteration of the loop increases `bytes_total` by at most `(cnt - 1) - bytes_total`.

lib/nolibc/scanf.c
```diff
@@ -8,6 +8,7 @@
 #include <stdarg.h>
 
 #include <uk/plat/console.h>
+#include <uk/assert.h>
```
michpappas: Now we removed the assertion we probably don't need this. Other than that it LGTM 👍🏼

thass0: Right, the include is gone now

michpappas:
Actually that function still doesn't look really robust. Consider for example what happens when the caller passes `cnt = 1`. The same thing happens if the caller input the same amount of characters as `sizeof(buffer) - 1`.

thass0:
If `cnt` is passed as one, then we run into an endless loop, right? Because `count` is `0` then so `ukplat_cink` will also return `0`. But the loop around `ukplat_cink` only ends if `ukplat_cink` returns a value greater than zero.

> The same thing happens if the caller input the same amount of characters as `sizeof(buffer) - 1`.

What happens exactly?

michpappas:
> If cnt is passed as one, then we run into an endless loop, right? 

Exactly, it runs into an endless loop. So your fix needs a bit more work to address that problem. With the opportunity you can also update that loop to handle `EOF` (Ctrl-D).

michpappas (changes requested):
Thanks. Overall this function could take a bit of rework, and we would probably need a tty layer to handle things like EOF, but until then let's integrate this fix to make sure that at least the buffer is terminated.

razvand:
@andreittr , could you also review this PR?

lib/nolibc/scanf.c
```diff
@@ -16,7 +17,14 @@ uk_scanf(void *buffer,  size_t *cnt)
 	size_t bytes_total = 0, count;
 	char *buf = buffer;
 
-	count = *cnt;
+	/* Need at least two bytes: one for user input and one for
```
skuenzer, line 20: Does this comment fit with the check below (`at least two bytes` != `<= 1`)?
In the case of `size = 0`, couldn't we return without reading a char from ukplat_cink?

skuenzer:
> > If cnt is passed as one, then we run into an endless loop, right?
> 
> Exactly, it runs into an endless loop. So your fix needs a bit more work to address that problem. With the opportunity you can also update that loop to handle `EOF` (Ctrl-D).

Saw this, too. This is fine, we can address later.  Probably we could in the case of `size = 1`, just return without reading a char from ukplat_cink (since there is no space).

skuenzer:
> @andreittr , could you also review this PR?

I think it this small bugfix PR is fine without a second reviewer.

skuenzer (approved):
Thanks!

skuenzer:
> Thanks. Overall this function could take a bit of rework, and we would probably need a tty layer to handle things like EOF, but until then let's integrate this fix to make sure that at least the buffer is terminated.
> 
> Reviewed-by: Michalis Pappas [michalis@unikraft.io](mailto:michalis@unikraft.io)

@michpappas Can you send this tag with a GitHub approval? ...otherwise our tooling will block merging the PR.

michpappas (approved):

PR #1477, opened 2024-08-13
lib/uksched: Export stats via ukstore

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

* `CONFIG_LIUKSCHED_STATS`: Invisible option selected by the scheduler implementation library.
* `CONFIG_LIBUKSCHEDCOOP_STATS`: Enables stats in `libukschedcoop` and selects `CONFIG_LIBUKSCHED_STATS`.

### Description of changes

This series updates `libuksched` to provide per-scheduler statistics via `libukstore`. The exported metrics are:
* `UK_SCHED_STATS_CPU_TOTAL_TIME`
* `UK_SCHED_STATS_CPU_IDLE_TIME`
* `UK_SCHED_STATS_NUM_SCHEDULES`
* `UK_SCHED_STATS_NUM_SWITCHES`
* `UK_SCHED_STATS_NUM_YIEDS`

As some metrics are exported from the scheduler implementation (e.g. `libukschedcoop`), enabling stats in `libuksched` is done via an invisible option selected by the scheduler implementation library.


lib/uksched/stats.c
```diff
@@ -0,0 +1,166 @@
...
+static int get_cpu_total_time(void *cookie __unused, __u64 *out)
+{
+	__nsec now = (__u64)ukplat_monotonic_clock();
+
+	*out = now - total_time_last;
+
+	total_time_last = now;
+
+	return 0;
+}
```
craciunoiuc: If the tiem is actually since the last query, isn't the function name a bit misleading?

I read the code first and I was a bit confused and then I saw the comment

lib/uksched/stats.c
```diff
@@ -0,0 +1,166 @@
...
+	UK_ASSERT(idle_thread);
+
+	now = idle_thread->exec_time;
+
+	*out = now - idle_time_last;
+
+	idle_time_last = now;
+
+	return 0;
+}
```
craciunoiuc: Same here

Maybe would make sense to rename them to `*_since`

craciunoiuc (commented):
Two comments from my side, I would not call them 'Request changes' 😅 

Looks good on my side

lib/uksched/include/uk/sched.h
```diff
@@ -98,6 +98,11 @@ struct uk_sched {
 	struct uk_alloc *a_auxstack; /**< default allocator for aux stacks */
 	struct uk_alloc *a_uktls; /**< default allocator for TLS+ectx */
 	struct uk_sched *next;
+#if CONFIG_LIBUKSCHED_STATS
+	unsigned long num_yields;
+	unsigned long num_schedules;
+	unsigned long num_switches;
```
skuenzer: I think those would make more sense to have per thread or at least additionally per thread?
How are they also different in practice?
I remember Linux had stats like `number of voluntary context switches` (sched_yield) and `number of involuntary context switches` (could be an expired time slice).

lib/uksched/stats.c
skuenzer: Hum... Can we do really the total time and not a diff. A diff could be computed by a caller, right?

lib/uksched/stats.c
skuenzer: I would also take the total and not compute the time between calls here. Callers can do compute the diff by doing two calls with a pause in-between.

lib/uksched/stats.c
craciunoiuc: at least for prometheus it's a time series, each query has the time appended and you can do the diff between calls, so yeah, makes sense

skuenzer: In case we need a diff for prometheus, I would implement that computation logic within prometheus exporter that would be a user of the ukstore entry.

skuenzer: I think this should be here the cpu time since scheduler started.

michpappas:
Test code below:
```c
#include <stdio.h>

#include <uk/event.h>
#include <uk/libid.h>
#include <uk/store.h>
#include <uk/sched/store.h>

const struct uk_store_entry *e_cpu_total;
const struct uk_store_entry *e_cpu_idle;
const struct uk_store_entry *e_num_sched;
const struct uk_store_entry *e_num_switch;
const struct uk_store_entry *e_num_yield;

void ukstore_sched_test(void)
{
	__u64 cpu_total, cpu_idle;
	__u64 num_sched;
	__u64 num_switch;
	__u64 num_yield;

	uk_store_get_value(e_cpu_total, u64, &cpu_total);
	uk_store_get_value(e_cpu_idle,  u64, &cpu_idle);
	uk_store_get_value(e_num_sched,  u64, &num_sched);
	uk_store_get_value(e_num_switch,  u64, &num_switch);
	uk_store_get_value(e_num_yield,  u64, &num_yield);

	printf("cpu total: %012ld, cpu idle: %012ld (%02ld%%), schedules: %03ld, context-switches: %03ld, yields: %03ld\n", 
	       cpu_total, cpu_idle, (100 * cpu_idle / cpu_total), num_sched, num_switch, num_yield);
}

static int handle_create_object(void *arg)
{
	struct uk_store_event_data *data = (struct uk_store_event_data *)arg;
	struct uk_store_object *obj;

	/* Is this coming from our library? */
	if (data->library_id != __UKLIBID_libuksched)
		return UK_EVENT_NOT_HANDLED; /* continue to next handler */

	/* Increments the refcount */
	obj = uk_store_obj_acquire(data->library_id, data->object_id);
	if (unlikely(!obj))
		return UK_EVENT_HANDLED; /* The object was already released */

	/* At this point we hold a valid pointer of the object. It won't be
	* freed until we decrement the refcount.
	*/
	e_cpu_total = uk_store_obj_entry_get(obj, UK_SCHED_STATS_CPU_TOTAL_TIME);
	e_cpu_idle  = uk_store_obj_entry_get(obj, UK_SCHED_STATS_CPU_IDLE_TIME);
	e_num_sched = uk_store_obj_entry_get(obj, UK_SCHED_STATS_NUM_SCHEDULES);
	e_num_switch = uk_store_obj_entry_get(obj, UK_SCHED_STATS_NUM_SWITCHES);
	e_num_yield = uk_store_obj_entry_get(obj, UK_SCHED_STATS_NUM_YIELDS);

	return UK_EVENT_HANDLED;
}

UK_EVENT_HANDLER(UKSTORE_EVENT_CREATE_OBJECT, handle_create_object);
```
This can be called from `app-helloworld` as:
```diff
 #if CONFIG_APPHELLOWORLD_SPINNER
        i = 0;
        printf("\n\n\n");
        for (;;) {
+#if CONFIG_LIBUKSCHEDCOOP_STATS
+               ukstore_sched_test();
+#else
                i %= (monkey3_frame_count * 3);
                printf("\r\033[2A %s \n", monkey3[i++]);
                printf(" %s \n",          monkey3[i++]);
                printf(" %s ",            monkey3[i++]);
                fflush(stdout);
+#endif
                millisleep(250);
        }
 #endif /* CONFIG_APPHELLOWORLD_SPINNER */
```
Sample output:
```
cpu total: 000010811567, cpu idle: 000000000000 (00%), schedules: 000, context-switches: 000, yields: 000
cpu total: 000263478569, cpu idle: 000250243766 (94%), schedules: 002, context-switches: 002, yields: 001
cpu total: 000516756275, cpu idle: 000500754869 (96%), schedules: 004, context-switches: 004, yields: 002
cpu total: 000770150043, cpu idle: 000750948963 (97%), schedules: 006, context-switches: 006, yields: 003
cpu total: 001024344583, cpu idle: 001001495271 (97%), schedules: 008, context-switches: 008, yields: 004
cpu total: 001278158148, cpu idle: 001252027924 (97%), schedules: 010, context-switches: 010, yields: 005
cpu total: 001532346024, cpu idle: 001502579991 (98%), schedules: 012, context-switches: 012, yields: 006
cpu total: 001785978298, cpu idle: 001752684393 (98%), schedules: 014, context-switches: 014, yields: 007
cpu total: 002038770609, cpu idle: 002002871525 (98%), schedules: 016, context-switches: 016, yields: 008
cpu total: 002292129264, cpu idle: 002253155334 (98%), schedules: 018, context-switches: 018, yields: 009
cpu total: 002545655314, cpu idle: 002503780225 (98%), schedules: 020, context-switches: 020, yields: 010
cpu total: 002798848115, cpu idle: 002753918116 (98%), schedules: 022, context-switches: 022, yields: 011
cpu total: 003052117065, cpu idle: 003004624855 (98%), schedules: 024, context-switches: 024, yields: 012
cpu total: 003305280364, cpu idle: 003254839799 (98%), schedules: 026, context-switches: 026, yields: 013
cpu total: 003558420289, cpu idle: 003505053369 (98%), schedules: 028, context-switches: 028, yields: 014
cpu total: 003812341518, cpu idle: 003755555605 (98%), schedules: 030, context-switches: 030, yields: 015
cpu total: 004066198963, cpu idle: 004005959173 (98%), schedules: 032, context-switches: 032, yields: 016
cpu total: 004319783311, cpu idle: 004256464407 (98%), schedules: 034, context-switches: 034, yields: 017
cpu total: 004573396509, cpu idle: 004506971142 (98%), schedules: 036, context-switches: 036, yields: 018
```

lib/uksched/stats.c
michpappas: Done

lib/uksched/stats.c
michpappas: Done

lib/uksched/include/uk/sched.h
michpappas: We can definitely do that. What we implement in this PR is CPU aggregate statistics, i.e. the equivalent of `/proc/schedstat`. Per-thread stats involve a few more challenges:

* Application threads need to be provided by `posix-process`, when that is enabled. We can store ukstore objects in the thread struct that `posix-process` can access when constructing the process hierarchy. FTR, linux provides aggregate process stats `/proc/<pid>/schedstat` as well as per-thread stats `/proc/<pid>/task/<tid>/schedstat`.
* Kernel threads need some sort of identifier in lieu of TID, which the caller must provide.

Given the above, I see this as a separate task, so I propose that we handle this on a separate PR.

michpappas:
Rebased from `staging`

craciunoiuc:
Ping me when I should do a final pass, I remember this working

michpappas:
> Ping me when I should do a final pass, I remember this working

Unless @skuenzer has additional feedback, there are pending changes from my side :)

lib/ukschedcoop/Config.uk
```diff
@@ -1,5 +1,14 @@
-config LIBUKSCHEDCOOP
+menuconfig LIBUKSCHEDCOOP
 	bool "ukschedcoop: Cooperative Round-Robin scheduler"
 	default y
 	select LIBNOLIBC if !HAVE_LIBC
 	select LIBUKSCHED
+
+if LIBUKSCHEDCOOP
+
+config LIBUKSCHEDCOOP_STATS
```
skuenzer: Is this used somewhere?

lib/ukschedcoop/schedcoop.c
```diff
@@ -117,6 +117,10 @@ static void schedcoop_schedule(struct uk_sched *s)
 		uk_thread_clear_queueable(next);
 	}
 
+#ifdef CONFIG_LIBUKSCHED_STATS
```
skuenzer: Isn't this counter almost counting the same as what is counted in `uk_sched_yield()`?

lib/uksched/include/uk/sched_impl.h
```diff
@@ -110,6 +110,11 @@ void uk_sched_thread_switch(struct uk_thread *next)
 
 	ukplat_lcpu_set_auxsp(next->auxsp);
 
+#ifdef CONFIG_LIBUKSCHED_STATS
+	UK_ASSERT(next->sched);
+	next->sched->num_switches++;
```
skuenzer: This stat makes sense, I am just wondering if it actually makes more sense to take the scheduler reference from `prev` because that is the one that caused the switch (although this condition should be met `prev->sched == next->sched`). I think it is also safer to do this before we load any context of the new thread.

lib/uksched/include/uk/sched/store.h
```diff
@@ -0,0 +1,22 @@
...
+/* uksched stats store object
+ *
+ * By convention, for the stats object ID we use the index
+ * of the cpu this scheduler is running on. Yet we cannot
+ * export that as a macro that expands to ukplat_lcpu_idx()
+ * as the consumer may be executing on a different core.
+ */
+#define UK_SCHED_STATS_CPU_TOTAL_TIME		0x01
+#define UK_SCHED_STATS_CPU_IDLE_TIME		0x02
+#define UK_SCHED_STATS_NUM_SCHEDULES		0x03
```
skuenzer: Can you write a comment next to each `num` item what the value means?

lib/uksched/include/uk/sched/store.h
```diff
@@ -0,0 +1,22 @@
...
+#define UK_SCHED_STATS_NUM_SWITCHES		0x04
+#define UK_SCHED_STATS_NUM_YIELDS		0x05
```
skuenzer: I am wondering if we should keep both: `NUM_SCHEDULES` and `NUM_YIELDS`. For a preemptive scheduler it makes sense, considering `yield` are voluntary switches and `switches` is the sum of involuntary and voluntary. However, I think I would try to count involuntary switches independently. Maybe it makes sense to introduce something like this when we do preempt scheduling. What do you think?

skuenzer (changes requested):

lib/uksched/include/uk/sched.h
skuenzer: Sounds good, I agree.

craciunoiuc (approved):
I think Simon already covered everything :pray: 

lib/ukschedcoop/Config.uk
michpappas: Yes, it's purpose is to select `LIBUKSCHED_STATS`. That is because `libuksched` can't know at Kconfig level which library provides the implementation, so the path needs to be `LIBUKSCHEDCOOP_STATS` -> `LIBUKSCHED_STATS`. You will also notice that there is no option to enable stats in `libuksched`.

lib/ukschedcoop/schedcoop.c
michpappas: That's right. Let's continue the discussion in https://github.com/unikraft/unikraft/pull/1477#discussion_r1778845747.

lib/uksched/include/uk/sched/store.h
michpappas: As the metrics are defined by `libuksched` which has no knowledge of the scheduler's implementation, what I had in mind was to provide a more generic set of metrics that would also suit a pre-emptive scheduler.

I am posting here the outcome of our offline discussion.

```
/* uksched stats store object
 *
 * Exported metrics
 * ================
 *                                         ____________ 
 * uk_sched_yield()--[YIELD_COUNT++]-+    |            |
 *                                   |--->| schedule() | [SCHED_COUNT++]
 * preempt --------------------------+    |____________|
 *                                        /   |    |
 * resume current thread <---------------+    |    |
 *                                            |    |
 *                             [IDLE_COUNT++] |    | [SWITCH_COUNT++]
 *                                            v    v
 *                                         idle    next
 *                                       thread    thread
 *
 * YIELD_COUNT:  Number of calls to yield().
 * SCHED_COUNT:  Number of calls to schedule().
 * SWITCH_COUNT: Number of context switches.
 * IDLE_COUNT:   Number of switches to the idle.
 *
 * Given the above it is possible to derive other metrics:
 * - preempt_count = sched_count - yield_count
 * - resume_count = sched_count - idle_count - switch_count
 */
```

I updated the PR with this information and added a counter for context switches to the idle thread,. I also did some renaming, and the diff is quite large, so please have another look.

Notice that the `idle_thread_fn()` of libukschedcoop calls `schedcoop_schedule()` instead of `uk_sched_yield()`, and that causes the yield counter not to be incremented, which is then incorrectly calculated as  preemptions:
```
[cpu_time]  total: 000253551447, cpu idle: 000250102784 (98%)
[schedules] total: 002, yield: 001, preempt: 001
[switches]  next: 001, idle: 001, current: 000

[cpu_time]  total: 000504332479, cpu idle: 000500248698 (99%)
[schedules] total: 004, yield: 002, preempt: 002
[switches]  next: 002, idle: 002, current: 000

[cpu_time]  total: 000755436715, cpu idle: 000750446436 (99%)
[schedules] total: 006, yield: 003, preempt: 003
[switches]  next: 003, idle: 003, current: 000
```
Replacing with `uk_sched_yield()`  results into:
```
[cpu_time]  total: 000253603696, cpu idle: 000250266799 (98%)
[schedules] total: 002, yield: 002, preempt: 000
[switches]  next: 001, idle: 001, current: 000

[cpu_time]  total: 000504544314, cpu idle: 000500397108 (99%)
[schedules] total: 004, yield: 004, preempt: 000
[switches]  next: 002, idle: 002, current: 000

[cpu_time]  total: 000755111327, cpu idle: 000750595826 (99%)
[schedules] total: 006, yield: 006, preempt: 000
[switches]  next: 003, idle: 003, current: 000
```
Do you see any problem with this change?

lib/uksched/include/uk/sched/store.h
michpappas: Will do, see https://github.com/unikraft/unikraft/pull/1477#discussion_r1778845747

lib/uksched/include/uk/sched_impl.h
michpappas: Ok

lib/uksched/Config.uk
```diff
@@ -7,12 +7,17 @@ menuconfig LIBUKSCHED
...
+# Invisible symbol to enable TCB initialization
+config LIBUKSCHED_TCB_INIT
+	bool
+
+config LIBUKSCHED_DEBUG
+	bool "Enable debug messages"
+
+config LIBUKSCHED_STATS
+	bool
+	select LIBUKLIBID
```
flpostolache: Why is `UKLIBID` selected here? Won't it be active when [`UKSTORE`](https://github.com/unikraft/unikraft/blob/staging/lib/ukstore/Config.uk#L5) is selected?

flpostolache (commented):
Just one question from my side. Otherwise it looks good to me! Thank you! 💪🏿

lib/uksched/Config.uk
michpappas: Generally speaking, dependencies must be defined explicitly. This applies to both Kconfig and headers. Saying that, in this particular case I don't see why `uklibid` is there, most probably it's something left as things changed during development. Thanks for spotting it out 👍🏼

flpostolache (approved):
Thank you! 

lib/uksched/include/uk/sched/store.h
```diff
@@ -0,0 +1,47 @@
...
+ *                                       thread    thread
+ *
+ * YIELD_COUNT: Number of calls to yield().
+ * SCHED_COUNT: Number of calls to schedule().
+ * NEXT_COUNT:  Number of context switches.
+ * IDLE_COUNT:  Number of switches to the idle.
+ *
+ * Given the above it is possible to derive other metrics:
+ * - preempt_count = sched_count - yield_count
+ * - resume_count = sched_count - switch_count - idle_count
```
skuenzer: Maybe it is easier for somebody trying to understand this to replace `switch_count` to `next_count`. Or we rename `NEXT_COUNT` to `SWITCH_COUNT`. Otherwise, this looks perfect. 💯

lib/ukschedcoop/Config.uk
skuenzer: Okay, understand. I would prefer if we design this similar to `ukalloc`, for simplicity. I would not provide an option to enable stats on each schedulers individually, I would enable them given what is configured in `lib/uksched`. In order to simplify instrumentation of the actual scheduler code we should provide helpers for counting events: https://github.com/unikraft/unikraft/blob/staging/lib/ukalloc/include/uk/alloc_impl.h#L161 Those helpers are either compiling in the code for counting or compiling it out, depending on how `uksched` is configured.

What I like is that any scheduler that doesn't support stats would keep the counters at `0` automatically, because the instrumentation would never be called. Although we have this effect already with the current code base.

lib/uksched/include/uk/sched.h
```diff
@@ -98,6 +98,13 @@ struct uk_sched {
 	struct uk_alloc *a_auxstack; /**< default allocator for aux stacks */
 	struct uk_alloc *a_uktls; /**< default allocator for TLS+ectx */
 	struct uk_sched *next;
+#if CONFIG_LIBUKSCHED_STATS
+	__nsec start_time;
+	unsigned long sched_count;
+	unsigned long yield_count;
+	unsigned long next_count;
+	unsigned long idle_count;
```
skuenzer: I am just seeing that we should also update the `uk_sched_init` macro so that the counters are reset. I am thinking to use the partial struct initialization because that makes sure that any new field in the future is automatically initialized with `0`:
```c
*(s) = (struct uk_sched){ \
    .thread_add = (thread_add_func);
    /* .. */
}
```

lib/uksched/include/uk/sched/store.h
michpappas: Yes, `switch_count` was accidentally left during development while I was changing things around. Thanks for spotting it.

skuenzer (approved):
The updates look very good. Thanks for this feature addition!

PR #1489, opened 2024-09-16
plat/kvm: Enable VGA by default on QEMU x86_64

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [`x86_64`]
 - Platform(s): [`kvm`]
 - Application(s): [N/A]

### Description of changes

Enable VGA by default on QEMU x86_64. This was removed during the migration of VGA to drivers.


StefanJum (approved):
All good, thanks.

mogasergiu:
Why do we enable VGA by default? :thinking: 
It does not work properly with color codes and it is also super slow and crashes with SMP. Would it be better to just keep the COMs only?

mogasergiu (approved):

michpappas:
> Why do we enable VGA by default? 🤔 It does not work properly with color codes and it is also super slow and crashes with SMP. Would it be better to just keep the COMs only?

This is ultimately on @simon, but just FTR, it's not that we enable it now. This PR restores the default which is VGA on, which was unintentionally disabled in #1464.

skuenzer:
> > Why do we enable VGA by default? 🤔 It does not work properly with color codes and it is also super slow and crashes with SMP. Would it be better to just keep the COMs only?
> 
> This is ultimately on @simon, but just FTR, it's not that we enable it now. This PR restores the default which is VGA on, which was unintentionally disabled in #1464.

This is a good point. To keep the previous behavior (before #1464), I would keep it on for now as well. This is for ease of getting started, especially for people running the QEMU image on environments like `libvirt`.
However, I agree, adding VGA is against our "as minimal as possible" principle. I am thinking we can fix this with completing re-arch of platforms and maybe introduce 2 different QEMU/x86 targets, one without having a VGA.

skuenzer (approved):

PR #1492, opened 2024-09-17
lib/ukdebug: Add `README.md` to `ukdebug`

thass0:
The primary function of this README is to document how the GDB stub is used. In future commits, this README could potentially be extended with extra documentation describing the other features of `ukdebug`.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### DIFF
--- a/lib/ukdebug/README.md
+++ b/lib/ukdebug/README.md
@@ -1,0 +1,154 @@
+# `ukdebug`: Unikraft's debugging features
+
+This library contains features to debug Unikraft.
+
+## GDB stub
+
+The GDB stub allows remotely debugging Unikraft with GDB using the [GDB Remote Serial Protocol (RSP)](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Remote-Protocol.html).
+The debugging process involves two machines.
+The development machine (the GDB host) runs GDB and implements the client side of the RSP.
+The other machine runs Unikraft (the target), which implements the GDB stub and serves requests issued by the host.
+The two sides communicate over a serial connection.
+With this setup, the machine running Unikraft can be a local VM or a physically remote machine.
+
+```plain
+   Development machine                     Unikraft
+  +-------------+                         +-------------+
+  |             |    Serial connection    |             |
+  |  GDB Host <-+-------------------------+-> GDB Stub  |
+  |             |                         |             |
+  +-------------+                         +-------------+
+```
+
+The stub runs after every debug trap, meaning the stub executes in an exception context.
+A loop inside the stub waits to receive packets from the host and to act upon them.
+Once the stub has finished handling a packet, it will send the response to the host and wait for the next packet.
+Some packets instruct the stub to leave the loop and resume execution of the kernel.
+
+The GDB stub uses a serial console device to communicate with the host.
+This can be either the same console that is used for standard output, or a separate console dedicated for debugging.
+If the GDB stub and standard output use the same console, Unikraft's standard output is printed inline with GDB's output in the GDB CLI.
+
+### Kernel parameters
+
+Two kernel parameters control the GDB stub.
+
+- `debug.gdbcon=<id>`: ID of the console device that is used to communicate with the GDB host.
+  If there's only a single console on the system, its ID is `0`, so `debug.gdbcon=0` should be used to select it.
+  The GDB stub is only enabled at runtime if `debug.gdbcon` is set.
+- `debug.gdbmode=[0, 1, -1]`: A tristate value controlling the mode of the GDB stub.
+  The three states are "shared" (`1`), "not shared" (`0`) and "auto" (`-1`).
+  The default it "auto".
+  In "shared" mode, Unikraft's standard output is printed inline with GDB's output in the GDB CLI on the host.
+  In "not shared" mode, Unikraft's standard output is never printed inline with GDB's output.
+  If the console chosen to communicate with the GDB host is used for standard output, "auto" will enable the "shared" mode.
+
+### QEMU & GDB example
+
+What follows is a concrete example of using the GDB stub with QEMU and GDB.
+For starters, configure your project to enable the GDB stub and a single console device.
+Note that the GDB stub depends on `uklibparam`, so you need to select that before the GDB stub shows up in the `ukdebug` menu in kconfig.
+Let's assume we are using x86.
+A possible QEMU command to start a VM that the GDB host can connect to is:
+
+```bash
+qemu-system-x86_64 -kernel build/app-helloworld_qemu-x86_64 \
+                   -cpu max \
+                   -display none \
+                   -serial tcp::1235,server,nowait \
+                   -append "build/app-helloworld_qemu-x86_64 debug.gdbcon=0 --"
+```
+
+The only two interesting options here are `-serial` and `-append`.
+`-serial tcp::1235,server,nowait` makes QEMU create a TCP socket on `localhost:1235` that's connected to COM1 inside the VM.
+This TCP socket will be used to connect the GDB host to the serial console in the VM.
+`-append` specifies the kernel command line that's passed to the kernel at runtime.
+Passing the name of the image as the first parameter is an Unikraft convention.
+If the first parameter is a kernel parameter, this kernel parameter will be dropped, so be careful.
+With `debug.gdbcon=0`, we tell the GBD stub to use `con0` to communicate with the GDB host.
+On x86 this is COM1, and it's connected to `localhost:1235`.
+
+You can use `-serial stdio` instead of `-serial tcp::1235,server,nowait` to see if the GDB stub is enabled and set up correctly.
+If you can run the above command with `-serial stdio`, you should see some output that ends with something like the below (assuming verbosity is set to >= `info`):
+
+```plain
+[    0.104746] Info: [libukdebug] <gdbstub.c @ 1237> Waiting for debugger connection on con0...
+[    0.104944] Info: [libukconsole] <console.c @  176> Registered con1: GDB virtual console, flags: --
+$S05#b8
+```
+
+The last line is the first packet that the GDB stub sends to the GDB host.
+Switch back to `-serial tcp::1235,server,nowait` and you'll stop seeing any output.
+Using `-serial stdio` can be useful when you're unsure if your configured kernel even boots.
+
+Note that the `-serial` flag is position-sensitive on x86.
+COM1 is connected to whatever the first occurrence (from left to right) of the `-serial` flag specifies.
+COM2 is configured based on the second occurrence of `-serial`, and so forth.
+E.g., as an alternative to the above, you could configure COM1 and COM2 in kconfig and then connect the GDB stub to COM2 and keep kernel output on COM1 (`debug.gdbcon` is `1` this time):
+
+```bash
+qemu-system-x86_64 -kernel build/app-helloworld_qemu-x86_64 \
+                   -cpu max \
+                   -display none \
+                   -serial stdio \
+                   -serial tcp::1235,server,nowait \
+                   -append "build/app-helloworld_qemu-x86_64 debug.gdbcon=1 --"
+```
+
+But let's stick to the initial example and assume that only COM1 is configured and that `debug.gdbcon=0`.
+Start the VM and in another terminal run GDB like:
+
+```bash
+gdb build/app-helloworld_qemu-x86_64.dbg
+```
+
+There are two things to note here.
+First, make sure to use the `.dbg` image, as it contains symbol and debug information.
+GDB won't find any symbols without it.
+Second, make sure you're using a  GDB binary compiled for the target architecture.
+If you're cross-compiling, you must use the GDB binary provided by the cross-compiling toolchain.
+
+Run the command `target remote localhost:1235` in the GDB CLI to connect to the GDB stub on TCP socket `localhost:1235`.
+GDB should manage to connect to the VM and you should see where execution was stopped before entering the stub.
+E.g., it might look like this:
+
+```plain
+(gdb) tar remote :1235
+Remote debugging using :1235
+0x0000000000120e02 in gdb_entry (ctx=<optimized out>) at /home/thasso/Work/oss-unikraft/lib/ukdebug/gdbstub.c:1247
+1247    }
+```
+
+You are now connected to the GDB stub that runs inside Unikraft.
+All the usual GDB commands like `break`, `backtrace`, or `continue` should work.
+Since we only configured a single console and didn't set `debug.gdbmode`, standard output should be printed inline in the GDB CLI.
+Run `break main` and `continue` to see standard output show up in the GDB CLI.
+
+### Debugging the GDB stub
+
+This section provides information for developers who need to debug Unikraft's GDB stub.
+First, note that GDB stub is not reentrant and should not be used to debug itself.
+The GDB command `set debug remote 1` comes in handy when debugging any problem with the GDB stub.
+This command will print all packets that the GDB host sends the stub along with all responses that the host receives.
+
+QEMU has a built-in GDB stub as part of the VMM, which can be used to debug the Unikraft-native GDB stub.
+The `-s` flag enables the QEMU-internal GDB stub, and the `-S` makes QEMU wait for a GDB host to connect to this stub before properly starting the VM.
+`-s` creates a TCP socket on `localhost:1234` that exposes the internal GDB stub.
+If we wanted to use the QEMU-internal stub in addition to the Unikraft-native stub in the example above, the QEMU command would look like this:
+
+```bash
+qemu-system-x86_64 -s -S \
+                   -kernel build/app-helloworld_qemu-x86_64 \
+                   -cpu max \
+                   -display none \
+                   -serial tcp::1235,server,nowait \
+                   -append "build/app-helloworld_qemu-x86_64 debug.gdbcon=0 --"
+```
+
+First, connect one instance of GDB to the QEMU-internal stub, then set some breakpoints and resume execution.
+Now the VM will start and enter the Unikraft-native GDB stub.
+You can connect a second instance of GDB to the Unikraft-native stub at this point.
+With the QEMU-internal GDB stub, you can safely set breakpoints on functions in the Unikraft-native stub.
+
+The connection between the Unikraft-native GDB stub and its GDB host can time out if you set a breakpoint at an unfortunate point in the GDB stub and take too long to do your debugging.
+For this reason, it's recommended that you use the command `set remotetimeout 1000000` to make the GDB host that's connected to the Unikraft-native stub wait for a very long time before timing out (here `1000000` is the number of seconds that GDB waits before it times out, any high value works).


lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
+# `ukdebug`: Unikraft's debugging features
+
+This library contains features to debug Unikraft.
+
+## GDB stub
+
+The GDB stub allows remotely debugging Unikraft with GDB using a feature of
```
michpappas: > with GDB using a feature of GDB called the remote protocol

A way more suitable for technical writing would be  "with GDB, using the [GDB Remote Serial Protocol](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Remote-Protocol.html)".

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+```plain
+  Your machine (any OS)                Other machine (Unikraft)
+  ┌─────────────┐                         ┌─────────────┐
+  │             │    Serial connection    │             │
+  │  GDB Host ◄─┼─────────────────────────┼─► GDB Stub  │
+  │             │                         │             │
+  └─────────────┘                         └─────────────┘
+```
+
+Unikraft's GDB stub lives in `gdbstub.c` with architecture-specific code in
```
michpappas: This sentence can be dropped. In principle it's best to avoid documenting implementation as much as possible as such documentation needs to be maintained along with the code. In practice this means it will eventually rot as people tend not to update documentation.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+This is done by re-routing all kernel output and sending it to the GDB host
+wrapped in a special packet. The GDB host then prints the output that it
+receives to `stdout`, inline with the normal output from the GDB CLI.
+
+### Kernel parameters
+
+Two kernel parameters control the backing console and the mode of the GDB stub.
+
+- `debug.gdbcon`: This is the console ID of the console device from `ukconsole`
+  that should be used as the backing console; it's a 16-bit unsigned integer.
```
michpappas: >  it's a 16-bit unsigned integer.

For reasons described earlier, I would avoid documenting the implementation in such detail.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+  If there's only a single console on the system, its ID is `0`, so
+  `debug.gdbcon=0` should be used to select it. The GDB stub is only enabled at
+  runtime if `debug.gdbcon` is set.
+- `debug.gdbmode`: This is a tristate value controlling the mode of the GDB
+  stub. Its type is a signed integer. The three states for the tristate are
```
michpappas: > Its type is a signed integer

Again this is unnecessary.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+### Kernel parameters
+
+Two kernel parameters control the backing console and the mode of the GDB stub.
+
+- `debug.gdbcon`: This is the console ID of the console device from `ukconsole`
+  that should be used as the backing console; it's a 16-bit unsigned integer.
+  If there's only a single console on the system, its ID is `0`, so
+  `debug.gdbcon=0` should be used to select it. The GDB stub is only enabled at
+  runtime if `debug.gdbcon` is set.
+- `debug.gdbmode`: This is a tristate value controlling the mode of the GDB
```
michpappas: "A tristate value controlling..."

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
+# `ukdebug`: Unikraft's debugging features
+
+This library contains features to debug Unikraft.
+
+## GDB stub
+
+The GDB stub allows remotely debugging Unikraft with GDB using a feature of
+GDB called the remote protocol. The way this works is you run an instance of
```
michpappas: > The way this works is you run an instance of GDB on your local computer that you connect to Unikraft.

Ditto, this can be improved to sound a bit less informal:

"The debugging process involves two machines. A development machine (the GDB host) that runs GDB, which implements the client side of the RSP. The other machine runs Unikraft (the target) implements the GDB stub, which serves the requests issued by the host. The two sides communicate over a serial connection. With this setup, the machine running Unikraft can be a local VM or a physically remote machine."

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+  │  GDB Host ◄─┼─────────────────────────┼─► GDB Stub  │
+  │             │                         │             │
+  └─────────────┘                         └─────────────┘
+```
+
+Unikraft's GDB stub lives in `gdbstub.c` with architecture-specific code in
+`arch/$(UK_ARCH)/gdbsup.c`.
+
+The stub runs after every debug trap (`UKARCH_TRAP_DEBUG`) meaning the stub
+executes in an exception context. There is a loop inside the stub that waits to
```
michpappas: >  There is a loop inside the stub that waits

"A loop inside the stub waits..."

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+receive packets from the host and act upon them. Once the stub is finished
+handling a packet, it will send the response to the host and wait for the next
+packet. Some packets instruct the stub to leave the loop and resume execution
+of the kernel.
+
+The GDB stub uses a serial console device (`struct uk_console`) to communicate
```
michpappas: Please remove (`struct uk_console`).

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+The GDB stub allows remotely debugging Unikraft with GDB using a feature of
+GDB called the remote protocol. The way this works is you run an instance of
+GDB on your local computer that you connect to Unikraft. This instance of
+GDB is called the GDB host, and it implements one side of the GDB remote
+protocol. The other side is implemented by the GDB stub in Unikraft and the
+two sides talk to each other over a serial connection. With this setup, the
+machine running Unikraft can be a local VM or a physically remote machine.
+
+```plain
+  Your machine (any OS)                Other machine (Unikraft)
```
michpappas: "Your machine (any OS)" -> "Development machine"

"Other machine (Unikraft)" -> "Unikraft".

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+to share the backing console between the GDB communications and kernel output.
+This is done by re-routing all kernel output and sending it to the GDB host
+wrapped in a special packet. The GDB host then prints the output that it
+receives to `stdout`, inline with the normal output from the GDB CLI.
+
+### Kernel parameters
+
+Two kernel parameters control the backing console and the mode of the GDB stub.
+
+- `debug.gdbcon`: This is the console ID of the console device from `ukconsole`
```
michpappas: `debug.gdbcon=<id>`

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+receive packets from the host and act upon them. Once the stub is finished
+handling a packet, it will send the response to the host and wait for the next
+packet. Some packets instruct the stub to leave the loop and resume execution
+of the kernel.
+
+The GDB stub uses a serial console device (`struct uk_console`) to communicate
+with the host. This device is referred to as the "backing console" because it
+backs the connection to the host. When the stub is initially set up, it
+retrieves a reference to the console device for the backing console from
+`ukconsole` and clears the `UK_CONSOLE_FLAG_STDIN` and `UK_CONSOLE_FLAG_STDOUT`
```
michpappas: The whole description of flags (`UK_CONSOLE_FLAG_STDIN` and `UK_CONSOLE_FLAG_STDOUT`) should not be here for maintainability reasons described above. Same for the implementation of backing / virtual console and how the stub takes over.

Additionally to the above these details only adds confusion to a user that is not aware of the implementation (consider the case that the reader's intention is to debug their application).

As the next paragraph is also based on this information, I would replace the two with a more high-level description:

"The GDB stub uses a serial console device to communicate with the host. This can be either the same console that is used for standard output, or a separate console dedicated for debugging. If a single console is used, Unikraft's standard output is printed inline with GDB's output in the GDB CLI.
"

lib/ukdebug/README.md
michpappas: > `debug.gdbmode:`

`debug.gdbmode=[0, 1, -1]:`.

> The three states for the tristate are "shared", "not shared", and "auto". Any negative value is interpreted to mean  "auto", a value of zero is interpreted to mean "not shared" and a positive value is interpreted as "shared".

In retrospective we should had used keywords for these instead of numbers. 

Either way, by defining the syntax as above you can now be more terse: "The three states for the tristate are "shared" (`0`), "not shared" (`1`), and "auto" (`-1`).". The default is "auto".

Although technically correct, I would avoid documenting the any positive / negative value behavior as it makes no sense to use values like `3` or `-6` anyway.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+- `debug.gdbcon`: This is the console ID of the console device from `ukconsole`
+  that should be used as the backing console; it's a 16-bit unsigned integer.
+  If there's only a single console on the system, its ID is `0`, so
+  `debug.gdbcon=0` should be used to select it. The GDB stub is only enabled at
+  runtime if `debug.gdbcon` is set.
+- `debug.gdbmode`: This is a tristate value controlling the mode of the GDB
+  stub. Its type is a signed integer. The three states for the tristate are
+  "shared", "not shared", and "auto". Any negative value is interpreted to mean
+  "auto", a value of zero is interpreted to mean "not shared" and a positive
+  value is interpreted as "shared". "Shared" mode is useful when the system
```
michpappas: > "Shared" mode is useful when the system only has a single console, its operation was described above. 

"The "shared" mode...".

> If you don't include this kernel parameter, "auto" will be used.

You can skip that by adopting the sentece in comment above that states "The default is "auto"."

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+  only has a single console, its operation was described above. If you don't
+  include this kernel parameter, "auto" will be used. With "auto" the selection
```
michpappas: > With "auto" the selection between "shared" and "not shared" mode is made based on a simple policy. If the device used as the backing console had any flags set before being taking over, then "auto" will select "shared" mode. Otherwise, if the backing console didn't have flags set, "auto" will select "not shared" mode.

This should be rephrased to more high-level terms (no implementation-specific description like flags).

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+  between "shared" and "not shared" mode is made based on a simple policy. If
+  the device used as the backing console had any flags set before being taking
+  over, then "auto" will select "shared" mode. Otherwise, if the backing
+  console didn't have flags set, "auto" will select "not shared" mode.
+
+### QEMU & GDB example
+
+What follows is a concrete example of using the GDB stub with QEMU and GDB. For
+starters, configure your project to enable the GDB stub and a single console
+device (COM1 on x86, PL011 on Arm). Note that the GDB stub depends on
```
michpappas: > (COM1 on x86, PL011 on Arm)

Arm can use other drivers, so maybe that's unnecessary.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+Passing the name of the image as the first parameter is an Unikraft convention.
+If the first parameter is a kernel parameter, this kernel parameter will be
+dropped, so be careful. With `debug.gdbcon=0`, we tell the GBD stub to use
+`con0` as the backing console. On x86 this is COM1 and it's connected to
+`localhost:1235`.
+
+You can use `-serial stdio` instead of `-serial tcp::1235,server,nowait` to
+see if the GDB stub is enabled and set up correctly. If you can run above
+command with `-serial stdio`, you should see some output that ends with
+something like the below (assuming you have debug messages enabled):
```
michpappas: > assuming you have debug messages enabled)

assuming verbosity set to >= `info`

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+qemu-system-x86_64 -kernel build/app-helloworld_qemu-x86_64 \
+                   -cpu max \
+                   -display none \
+                   -serial stdio \
+                   -serial tcp::1235,server,nowait \
+                   -append "build/app-helloworld_qemu-x86_64 debug.gdbcon=1 --"
+```
+
+But let's stick to the initial example and assume that only COM1 is configured
+and that `debug.gdbcon=0`. Start the VM and in another terminal run GDB like
```
michpappas: > run GDB like this:

"run GDB like:"

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+this:
+
+```bash
+gdb build/app-helloworld_qemu-x86_64.dbg
+```
+
+There are two things to note here. First, don't forget the `.dbg` extension.
```
michpappas: > First, don't forget the `.dbg` extension. GDB won't find any symbols without it.

"Make sure you use the `.dgb` image, as it contains object symbols."

> Second, make sure you're using the right GDB binary. E.g., if you're developing on an x86 system and the kernel is compiled for Arm, you need to use GDB for Arm, not the stock GDB of your system.

I would avoid making specific assumptions about the host here, as with Mac it's also common to be on Arm and cross-compile for `x86_64`, so: "Second, make sure you're using a GDB binary compiled for the target architecture. If you're cross-compiling you must use the GDB binary provided by the cross-compiling toolchain".

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+GDB won't find any symbols without it. Second, make sure you're using the right
+GDB binary. E.g., if you're developing on an x86 system and the kernel is
+compiled for Arm, you need to use GDB for Arm, not the stock GDB of your system.
+
+Run the command `target remote localhost:1235` in the GDB REPL to connect to
```
michpappas: I dont't think most people know what a REPL is, please use "CLI" or "interface".

lib/ukdebug/README.md
michpappas: >  This is the console ID of the console device from `ukconsole`  that should be used as the backing console.

"The ID assigned to the console GDB should connect to."

lib/ukdebug/README.md
```diff
@@ -0,0 +1,201 @@
...
+You are now connected to the GDB stub that runs inside Unikraft. All the
+usual GDB commands like `break`, `backtrace`, or `continue` should work. Since
+we only configured a single console and didn't set `debug.gdbmode` on the
+command line, the GDB should automatically share the backing console for kernel
+output. If you `break main` and `continue`, you should see all kernel output
+printed inline with the other output from GDB.
+
+### Debugging the GDB stub
+
+This is guidance for anyone working on the GDB stub who needs to debug it.
```
michpappas: > This is guidance for anyone working on the GDB stub who needs to debug it.

"This section provides information for developers who need to debug Unikraft's GDB stub."

michpappas (changes requested):
@thass0 thanks for that, review comments below. Also please notice that the convention in markdown documents is one sentence per line, as that makes looking at diffs easier.

michpappas (commented):
@thass0 thanks for that, I think it looks great. One last question from me: On line 77 is `$S05#b8` part of the output or did it accidentally end up there?

thass0:
The `$S05#b8` is part of the output. It's the first packet that the GDB stub sends to the host.

michpappas (approved):
Looks good from my side.

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
...
+The two sides communicate over a serial connection.
+With this setup, the machine running Unikraft can be a local VM or a physically remote machine.
+
+```plain
+   Development machine                     Unikraft
+  ┌─────────────┐                         ┌─────────────┐
+  │             │    Serial connection    │             │
+  │  GDB Host ◄─┼─────────────────────────┼─► GDB Stub  │
+  │             │                         │             │
+  └─────────────┘                         └─────────────┘
```
StefanJum: Use ASCII characters for this pls

StefanJum (changes requested):
Great work @thass0, thanks a lot. Only one small comment.

lib/ukdebug/README.md
michpappas: @StefanJum can you please elaborate on the rationale behind this requirement?

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
+# `ukdebug`: Unikraft's debugging features
+
+This library contains features to debug Unikraft.
+
+## GDB stub
+
+The GDB stub allows remotely debugging Unikraft with GDB using the [GDB Remote Serial Protocol (RSP)](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Remote-Protocol.html).
+The debugging process involves two machines.
+A development machine (the GDB host) that runs GDB and implements the client side of the RSP.
```
mariasfiraiala:
```suggestion
A development machine (the GDB host) runs GDB and implements the client side of the RSP.
```

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
...
+   Development machine                     Unikraft
+  ┌─────────────┐                         ┌─────────────┐
+  │             │    Serial connection    │             │
+  │  GDB Host ◄─┼─────────────────────────┼─► GDB Stub  │
+  │             │                         │             │
+  └─────────────┘                         └─────────────┘
+```
+
+The stub runs after every debug trap, meaning the stub executes in an exception context.
+A loop inside the stub waits to receive packets from the host and act upon them.
```
mariasfiraiala:
```suggestion
A loop inside the stub waits to receive packets from the host and acts upon them.
```

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
...
+Once the stub is finished handling a packet, it will send the response to the host and wait for the next packet.
```
mariasfiraiala:
```suggestion
Once the stub finishes handling a packet, it will send the response to the host and wait for the next packet.
```

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Kernel parameters
+
+Two kernel parameters control the GDB stub.
+
+- `debug.gdbcon=<id>`: ID of the console device that is used to communicate with the GDB host.
+  If there's only a single console on the system, its ID is `0`, so `debug.gdbcon=0` should be used to select it.
+  The GDB stub is only enabled at runtime if `debug.gdbcon` is set.
+- `debug.gdbmode=[0, 1, -1]`: A tristate value controlling the mode of the GDB stub.
+  The three states for the tristate are "shared" (`1`), "not shared" (`0`) and "auto" (`-1`).
```
mariasfiraiala:
```suggestion
  The three states are "shared" (`1`), "not shared" (`0`) and "auto" (`-1`).
```

lib/ukdebug/README.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+There are two things to note here.
+First, make sure to use the `.dbg` image, as it contains symbol and debug information.
+GDB won't find any symbols without it.
+Second, make sure you're using a  GDB binary compiled for the target architecture.
+If you're cross-compiling, you must use the GDB binary provided by the cross-compiling toolchain.
+
+Run the command `target remote localhost:1235` in the GDB CLI to connect to the GDB stub on TCP socket `localhost:1235`.
+GDB should manage to connect to the VM and you should see where in the kernel execution was stopped.
```
mariasfiraiala:
```suggestion
GDB should manage to connect to the VM and you should see where in the kernel execution it was stopped.
```

mariasfiraiala (changes requested):
Thanks @thass0, great work. I left some minor comments, you'll find them below.

lib/ukdebug/README.md
thass0: I prefer `is finished` here. Saying `Once the stub finishes ...` sounds like the stub sends the response at the same time as it finishes handling the response while, in fact, it sends the response after.

lib/ukdebug/README.md
thass0: I meant to refer to the `execution` as a single noun here, not `kernel execution` as your change suggests. This makes me think I need to reword this sentence to make it more clear. See if you like the update.

lib/ukdebug/README.md
thass0: This sentence was also phrased ambiguously. I meant to say that inside of the loop packets are received and acted upon. I switched from `act` to `to act` to hopefully make this more clear.

lib/ukdebug/README.md
thass0: I disagree with this. IMO the `that` makes sense in this sentence.

thass0:
@mariasfiraiala Thanks for your feedback and for taking the time to go over this. Please see the comments I left and the changes I made.

lib/ukdebug/README.md
mariasfiraiala: It's not a full sentence tho. I don't have a hard preference here however.

lib/ukdebug/README.md
mariasfiraiala: The original meaning stays intact due to the usage of the _once_ conjunction, which here means _as soon as_. I don't have a hard preference for either one of the wordings, however.

mariasfiraiala:
@thass0 I am quite happy with the wording now, thanks! As a general rule of thumb, in the future, try avoiding passive voice; it can lead to quite ambiguous sentences as we've come across here, for instance.

I left some comments for you to decide the final outcome of. When you're ready I'll approve this.

lib/ukdebug/README.md
thass0: OK you have convinced me :)  I removed the `that`.

lib/ukdebug/README.md
thass0: OK I'm keeping it as it is now.

thass0:
OK I'm fine with what's it's like now. Again, thanks for the feedback.

mariasfiraiala (approved):
Very nice work, @thass0!

lib/ukdebug/README.md
StefanJum: (As discussed in the maintainers meeting)
1. We can be sure ASCII characters are rendered properly everywhere, no matter the layout/language etc.
2. The README files will eventually be part of the documentation website, were we already use the ASCII-only convention, so it makes sense for consistency reasons.

thass0: Fair points. I switched the graphic to ASCII :+1:

StefanJum (approved):
All good now, thanks @thass0 

michpappas:

PR #1496, opened 2024-09-20
libukrandom: Introduce driver for CPU-generated entropy + misc. fixes

### LINKED ISSUE #1502: plat/xen/arm64: Xen does not boot with default cpu type
### Describe the bug

When testing Xen on IMX (imx8qmmek) the system does not print any output when configured with the default CPU type (`CONFIG_MCPU_ARM64_NONE`). Changing to `CONFIG_MCPU_ARM64_GENERIC` fixes the issue.

### Steps to reproduce

unikraft:  `6550307b3b97072d9659bd6c7e80dbe65a713e70` 

From a clean config select:
* Arch: arm64
* Plat: Xen

Compile and run on `imx8qmmek` with config:
```
name          = 'hello'
vcpus         = '1'
memory        = '64'
kernel        = '/path/to/app-helloworld_xen-arm64'
```

### Expected behavior

Unikraft should boot and print output as normal:
```
$ xl create -c app-helloworld_xen-arm64.cfg 
Parsing config from app-helloworld_xen-arm64.cfg
Powered by
               Telesto 0.16.3~e453b74a2
Hello world!
Arguments:  "app-helloworld" "app-helloworld"
```

arm64

xen

### Relevant log output

_No response_

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): `arm64`, `x86_64`]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series introduces the driver API of `libukrandom` and introduces a driver for CPU-generated randomness based on the implementation of `ukarch_random`, which is now deprecated.

It also adds a new Kconfing option, namely `LIBUKRANDOM_CMDLINE_INIT` to seed the CSPRNG via the cmdline. In ChaCha this is used to pass 8x 32-bit words for the key in `random.seed` cmdline paramter. When this option is enabled AND initialziation paramters are passed to the cmdline, the underlying RNG driver is not used. This replaces `CONFIG_LIBUKRANDOM_SEED_INSECURE` as it allows users of legacy hardware or older QEMU versions to seed the CSPRNG from the host's entropy sources.

Finally, it additionally relaxes the requirement of `libukrandom` in the config of KVM to avoid compatibility issues with legacy hardware and versions of QEMU < 8 on applications that don't require an RNG.


nderjung:
One usability request to simplify seeding the driver would be to allow for using a simple numeric, such that the following is possible:

```
random.seed=12345678
```

vs.

```
random.seed=[ 0xb0 0xb0 0xf0 0x0d 0xb0 0xb0 0xf0 0x0d 0xd0 0xd0 ]
```

lib/ukrandom/chacha.c
```diff
@@ -35,7 +35,7 @@
 #include <uk/assert.h>
 #include <uk/ctors.h>
 #include <uk/random.h>
-#include <uk/arch/random.h>
```
Alex-deVis, line 38: This file was deleted in the previous commit, are we OK with intermediary commits that do not compile?

mariapana (approved):
Looks good to me.

drivers/ukrandom/lcpu/arch/arm64/random.c
```diff
@@ -0,0 +1,128 @@
...
+static int seed_bytes_fb(__u8 *buf, __sz size)
+{
+	int rc;
+
+	/* It is unclear whether this provides any benefit as Arm
+	 * specifies different possibilities for the implementation
+	 * of RNDRRS (DDI 0487K.a Sect. DD23.2.148)
+	 */
+	rc = seed_bytes(buf, size);
+	if (unlikely(rc))
```
Alex-deVis, line 102: I think adding a debug message here would be helpful, as the failure of `seed_bytes` is a hint that the CPU does not provide enough true randomness for the app.

drivers/ukrandom/lcpu/arch/x86_64/random.c
```diff
@@ -0,0 +1,149 @@
...
+	.seed_bytes_fb = seed_bytes_fb,
+};
+
+struct uk_random_driver_ops *device_init(void)
+{
+	__u32 eax, ebx, ecx, edx;
+
+	ukarch_x86_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	if (unlikely(!(ecx & X86_CPUID1_ECX_RDRAND))) {
+		uk_pr_err("CPU does not implement RDRAND\n");
```
Alex-deVis: I think this should be "The CPU does not ..." to match the RDSEED warn message.

drivers/ukrandom/lcpu/arch/x86_64/random.c
```diff
@@ -0,0 +1,149 @@
...
+
+	return 0;
+}
+
+static inline int rdseed(__u64 *val)
+{
+	__u8 res;
+
+	if (!have_rdseed)
+		return -ENOTSUP;
```
Alex-deVis, line 43: The commit message states that **-ENODEV** is returned if RDSEED is not supported. I think the `-ENOTSUP` is more appropriate and the commit message should be updated.

Alex-deVis (commented):
This looks very nice, I left a few inline comments.

Besides, there is a typo in the **drivers/ukrandom: Introduce driver for CPU generated randomness** commit's description: "On CPUs that don't support RDSEED init will no **longer** fail ...".

lib/ukrandom/chacha.c
michpappas: Yes that's fine, as long as the PR compiles, as the amount of work it often takes to make every commit to compile is disproportional to the benefit.

drivers/ukrandom/lcpu/arch/x86_64/random.c
michpappas: Good catch, thanks 👍🏼

drivers/ukrandom/lcpu/arch/arm64/random.c
michpappas: I'd like to keep the very low level code minimal. As the driver ops are protected with the `__check_result`, the caller (`lib/ukrandom/chacha.c`) should handle this and print an appropriate message (it does).

drivers/ukrandom/lcpu/arch/x86_64/random.c
michpappas: Ok

michpappas:
Added unit tests for ChaCha20

Alex-deVis (approved):
Reviewed-by: Alex Apostolescu <alexx.apostolescu@gmail.com>

lib/ukrandom/dev.c
```diff
@@ -79,13 +70,17 @@ static int dev_random_read(struct device *dev __unused, struct uio *uio,
...
 	char *buf;
+	int ret;
 
 	buf = uio->uio_iov->iov_base;
 	count = uio->uio_iov->iov_len;
 
-	uk_random_fill_buffer(buf, count);
+	ret = uk_random_fill_buffer(buf, count);
+	if (unlikely(ret))
+		return rc;
```
razvand:
```suggestion
		return ret;
```

Otherwise, this gives a build error. If you enable `LIBUKRANDOM_DEVFS`.

razvand (changes requested):
@michpappas , see the build error after enabling `CONFIG_LIBUKRANDOM_DEVFS`.

For a Python application build (such as the one [here](https://github.com/unikraft/catalog-core/tree/main/python3-hello)), I need to enable `CONFIG_LIBUKRANDOM_GETRANDOM` for it to work (as part of the `Config.uk` file of the application). Otherwise, the application cannot generate random numbers:

```text
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized
```

Shouldn't this configuration (`CONFIG_LIBUKRANDOM` and `CONFIG_LIBUKRANDOM_GETRANDOOM`) be part of the `Config.uk` file of the [`lib-python3` library](https://github.com/unikraft/lib-python3)? Because it should be enabled irrespective of the application. Any time you use Python3, this is going to be required.

lib/ukrandom/dev.c
michpappas: Thanks, fixed

michpappas:
> @michpappas , see the build error after enabling `CONFIG_LIBUKRANDOM_DEVFS`.
> 
> For a Python application build (such as the one [here](https://github.com/unikraft/catalog-core/tree/main/python3-hello)), I need to enable `CONFIG_LIBUKRANDOM_GETRANDOM` for it to work (as part of the `Config.uk` file of the application). Otherwise, the application cannot generate random numbers:
> 
> ```
> Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
> Python runtime state: preinitialized
> ```
> 
> Shouldn't this configuration (`CONFIG_LIBUKRANDOM` and `CONFIG_LIBUKRANDOM_GETRANDOOM`) be part of the `Config.uk` file of the [`lib-python3` library](https://github.com/unikraft/lib-python3)? Because it should be enabled irrespective of the application. Any time you use Python3, this is going to be required.

@razvand that's right, we need to identify which libraries / applications require secure randomness and enable `CONFIG_LIBUKRANDOM` there. Moreover, libraries that require the `getrandom()` syscall need to enable `CONFIG_LIBSYSCALL_SHIM` and `CONFIG_LIBUKRANDOM_GETRANDOM`. Libraries that require device nodes need to enable `CONFIG_LIBVFSCORE` and `LIBUKRANDOM_DEVFS`.

razvand:
> @razvand that's right, we need to identify which libraries / applications require secure randomness and enable `CONFIG_LIBUKRANDOM` there. Moreover, libraries that require the `getrandom()` syscall need to enable `CONFIG_LIBSYSCALL_SHIM` and `CONFIG_LIBUKRANDOM_GETRANDOM`. Libraries that require device nodes need to enable `CONFIG_LIBVFSCORE` and `LIBUKRANDOM_DEVFS`.

@michpappas , roger. Taking a look.

drivers/ukrandom/lcpu/arch/x86_64/random.c
```diff
@@ -0,0 +1,149 @@
...
+	.seed_bytes_fb = seed_bytes_fb,
+};
+
+struct uk_random_driver_ops *device_init(void)
+{
+	__u32 eax, ebx, ecx, edx;
+
+	ukarch_x86_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	if (unlikely(!(ecx & X86_CPUID1_ECX_RDRAND))) {
+		uk_pr_err("The CPU does not implement RDRAND\n");
```
skuenzer: Hum, should we really say `"implement"`? On x86 you are able to mask out and switch off CPU features for guests (via cpuid), although the CPU provides the feature in hardware. This is often done to allow live migration between different x86 CPU generations by allowing only a common feature set.

What would you think of:
`Hardware random number generation (RDRAND) not available on this CPU`

drivers/ukrandom/lcpu/arch/x86_64/random.c
```diff
@@ -0,0 +1,149 @@
...
+		return ERR2PTR(-ENOTSUP);
+	}
+
+	ukarch_x86_cpuid(7, 0, &eax, &ebx, &ecx, &edx);
+	if (ebx & X86_CPUID7_EBX_RDSEED)
+		have_rdseed = __true;
+	else
+		uk_pr_warn("The CPU does not implement RDSEED");
```
skuenzer: Suggestion:
`Hardware random seed generation (RDSEED) not available on this CPU, using RDRAND`

lib/ukrandom/chacha.c
skuenzer: However, this is not excluding that it is in general a good practice to keep compile-able individual commits. It helps to structure PRs and thinking step-wise. However, we relaxed the requirement because for bigger re-arch PRs, it can be in fact difficult to achieve that.

lib/ukrandom/Config.uk
```diff
@@ -1,25 +1,23 @@
...
-		do not feature a TRNG.
-
-		Select this if you know what you're doing.
-
-		DO NOT USE IN PRODUCTION SYSTEMS
+		Selecting this option will cause libukrandom to obtain its
+		seed from the cmdline, if the `random.seed` parameter is
+		passed with a 256-bit key formatted as an array of
+		32-bit words. This option has higher precedence than the
+		device-tree and drivers to hardware RNG devices.
```
skuenzer: Should we maybe write one sentence that the cmdline parameter user is responsible for securing access the the command line origin (e.g., boot loader)?

drivers/ukrandom/lcpu/Config.uk
```diff
@@ -0,0 +1,11 @@
+menuconfig LIBUKRANDOM_LCPU
+	bool "CPU generated randomness"
+	depends on (ARCH_X86_64 || (ARCH_ARM_64))
+	select HAVE_ARM64_FEAT_RNG if ARCH_ARM_64
```
skuenzer, line 11: I am not fully understanding that we select `HAVE_ARM64_FEAT_RNG` (setting to true) here and at the same time in the Arm CPU configuration. Shouldn't the random driver depend on `HAVE_ARM64_FEAT_RNG`?

drivers/ukrandom/lcpu/init.c
```diff
@@ -0,0 +1,44 @@
...
+	rc = uk_random_init(&driver);
+	if (unlikely(rc)) {
+		uk_pr_err("Could not register with libukrandom (%d)\n", rc);
+		return rc;
+	}
+
+	return 0;
+}
+
+#if CONFIG_LIBUKBOOT
```
skuenzer: Hum, is it actually possible that we do not need to compile this file at all if `CONFIG_LIBUKBOOT` is absent (via `Makefile.uk`)? I expect that we will see an unused warning for `uk_random_lcpu_init` if `libukboot` is not part of the build (this is theoretically for now of course).

arch/arm/arm64/memtag.c
```diff
@@ -72,11 +72,11 @@ int ukarch_memtag_init(void)
...
 
 	/* Set seed for RGSR_EL1 */
-	if (ukarch_random_init())
-		UK_CRASH("Arch random not available\n");
-
-	if (ukarch_random_seed_u64(&seed))
-		UK_CRASH("Could not generate MTE seed\n");
+	rc = uk_random_fill_buffer(&seed, sizeof(seed));
+	if (unlikely(rc)) {
+		uk_pr_err("Could not generate MTE key (%d)\n", rc);
```
skuenzer, line 77: Somehow I am tending to use `"Failed to generate..."` instead of `"Could not generate..."` since it is underlining that there is a serious problem. However, we are talking here probably about preference.
If we fail, what is the actual effect? Are we able to boot or do we fail if we do not have MTE?

plat/common/arm/pauth.c
```diff
@@ -72,27 +28,40 @@ static inline bool pauth_supported(void)
...
 
 int __no_pauth ukplat_pauth_init(void)
 {
 	__u64 key_hi, key_lo;
 	__u64 reg;
+	int rc;
 
-	if (unlikely(!pauth_supported()))
+	if (unlikely(!pauth_supported())) {
+		uk_pr_err("The CPU does not implement FEAT_PAUTH\n");
```
skuenzer, line 48: I have a similar comment here regarding the message as I mentioned to the x86 `RDRAND` feature regarding "implement". Is it possible that features can be masked out on Arm for virtualized guests?

plat/common/arm/pauth.c
```diff
@@ -72,27 +28,40 @@ static inline bool pauth_supported(void)
...
 		return -ENOTSUP;
+	}
 
 	/* Program instruction Key A */
-	ukplat_pauth_gen_key(&key_hi, &key_lo);
+	rc = uk_random_fill_buffer(&key_hi, sizeof(key_hi));
+	if (unlikely(rc)) {
+		uk_pr_err("Could not generate APIAKey (%d)\n", rc);
```
skuenzer, line 55: Also here and the message at line 62, should we say `"Failed to ..."`? Again probably preference and I am not strongly object `"Could not..."`.

drivers/ukrandom/lcpu/init.c
michpappas: The motivation was to leave `init()` available for platforms that may not wish to use `libukboot`, so that they can still initialize the library by calling `init()` directly. In that case, we can address the warning by adding `__maybe_unused`. Overall I don't have a strong opinion, as we will need to formalize the behavior of library initialization in the scope of `libukboot` rearch anyway.

arch/arm/arm64/memtag.c
michpappas: > Somehow I am tending to use "Failed to generate..." instead of "Could not generate..." since it is underlining that there is a serious problem. However, we are talking here probably about preference.

Indeed, I think this is subjective :) Saying that, the idea is that the library does not enforce policy, so the severity of a failure is decided by the caller. In our case we treat failures of initcalls as critical.

> If we fail, what is the actual effect? Are we able to boot or do we fail if we do not have MTE?

Right now we call `UK_CRASH()`. The same will happen when we move MTE initialization to `earlytab`. IMO, if I had to choose between continuing boot with a warning or crashing, I would favor security and choose the latter, given that the fact that someone enabled MTE explicitly in Kconfig implies that they care about that feature. Nevertheless, we can also consider adding a feature in `libukboot` or individual libraries like `ukmemtag` that allows the user to select the system behavior upon failure.

plat/common/arm/pauth.c
michpappas: See https://github.com/unikraft/unikraft/pull/1496#discussion_r1858175667

drivers/ukrandom/lcpu/arch/x86_64/random.c
michpappas: That's true, but IMO it also depends on the perspective. That is, whether virtualized or not, from the operating system's point of view the CPU still appears not to implement the feature. Besides, normally the assumption is that the system runs natively, and virtualization is a special case. In fact, even there, (ideally) the OS is not aware that it's executing over virtualization. So in conclusion I think that this change is unnecessary. If you have a strong opinion though, I can change it.

drivers/ukrandom/lcpu/arch/x86_64/random.c
michpappas: Adding "using RDRAND" for now, for the rest see https://github.com/unikraft/unikraft/pull/1496#discussion_r1858301885.

plat/common/arm/pauth.c
michpappas: Yes, on `arm64` the hypervisor can trap the `mrs` to the feature identification registers and return a custom value to the guest. For the rest let's continue the discussion in https://github.com/unikraft/unikraft/pull/1496#discussion_r1858301885.

drivers/ukrandom/lcpu/Config.uk
michpappas: So in the Arm CPU configuration  `HAVE_ARM64_FEAT_RNG` is an invisible option only, that is also selected by features that depend on randomness. It needs to be there because we use it to generate GCC flags: https://github.com/unikraft/unikraft/blob/staging/arch/arm/arm64/Makefile.uk#L68. Ideally I would had liked that these flags are generated per-library, but unfortunately there are some inter-dependencies on the `march` flag that we need to resolve first at the build system. Example: 

* `FEAT_RNG` resolves to ARCH_REV := armv8.5-a and ARCH_OPT := $(ARCH_OPT)+rng
* `FEAT_PAUTH` resolves to ARCH_REV := armv8.3-a and BRANCH_PROTECTION := pac-ret+leaf
 
At the final compiler flags we need to select the highest architectural revision, otherwise we'll get a compiler error that `FEAT_RNG` is not supported in `armv8.3-a`. For now I choose to have libraries select the architectural features they support (e.g. `FEAT_RNG`) and delegate the compiler flags generation to a single Makefile where we easily can do the dependency resolution by specifying features of later architectural revisions last, so that they override earlier ones.

> Shouldn't the random driver depend on HAVE_ARM64_FEAT_RNG?

The reason we can't have `libukrandom` depend on `FEAT_RNG` is that we don't know if the underlying CPU implements `FEAT_RNG`. On `arm64` GCC defines some generic CPUs , i.e. `MCPU_ARM64_NONE`, `MCPU_ARM64_NATIVE` and `MCPU_ARM64_GENERIC`, and the default we use in Kconfig is`MCPU_ARM64_NONE`. So when the user selects the lcpu rng driver of `arm64`, it is expected that the user implies that the CPU implements `FEAT_RNG`.

drivers/ukrandom/lcpu/arch/x86_64/random.c
skuenzer: Hum, you could even argue that on a firmware level (e.g., BIOS) such features could be switched off although present on the CPU. Because of this, I'd prefer to name it `"not available"`. The thought behind this is that it does not lead to any conclusion if the CPU is in principle able to do it or not.

drivers/ukrandom/lcpu/arch/x86_64/random.c
skuenzer: See my answer of that comment, too :+1:

drivers/ukrandom/lcpu/init.c
skuenzer: Okay, in such a case `uk_random_lcpu_init()` should not be declared static and we should provide it via `exportsyms.uk`. Secondly, can't we ignore the input parameter and just declare it with unspecified parameters?: `int uk_random_lcpu_init()`. This would look more compatible to non-`libukboot` builds. Additionally we would need to provide the function prototype with a driver-specific header, like `uk/ukrandom/lcpu/lcpu.h`. We can then decide if we make it only available for non-`libukboot` builds which I guess would make sense.

plat/common/arm/pauth.c
skuenzer: Okay, then the same answer in that comment applies here 😄

drivers/ukrandom/lcpu/Config.uk
skuenzer: Okay, understand. As we discussed, let's re-visit this integration when we touch the compiler settings as part of the final re-arch of the architecture/lcpu libraries. I am thinking that we may want to rename `HAVE_ARM64_FEAT_RNG` to `NEED_ARM64_FEAT_RNG` or `WANT_ARM64_FEAT_RNG` - but getting to a conclusion is not trivial at this point and needs further discussion. We shouldn't block this PR because of this. Can you add a comment to the source code that describes why we are doing this here (for now) with `HAVE_ARM64_FEAT_RNG`?

drivers/ukrandom/lcpu/init.c
michpappas: Okay, since we don't have a final decision on the behavior of drivers when it comes to that aspect I'll follow your initial suggestion and make compilation of that file conditional to libukboot.

drivers/ukrandom/lcpu/Config.uk
michpappas: Done

drivers/ukrandom/lcpu/arch/x86_64/random.c
michpappas: Done

skuenzer (approved):
Thank you so much for this addition!

skuenzer (approved):
Updates are looking good! Thanks!

PR #1501, opened 2024-09-23
drivers/*: Adopting serial driver configuration and making Xen x86_64 (PV) work again

skuenzer:
### Description of changes

This PR reworks the driver configuration with the main focus on Xen and serial drivers. Without it, builds for Xen/x86_64 (PV) where broken or failed.

The PR is also suggesting to name driver libraries directly with a representing device name. The original convention tended to cause very long driver library names and did not fit to multi-purpose devices. A concern I have with shorter naming is that it does not protect well against name clashing: For example, `libpl011` is more likely to clash than `libukconsole_pl011`. I am open for other suggestions in this PR. I had the idea to prefix driver libraries with `drv` instead of `lib`. Alternatively we could include our `uk` prefix in the driver names to state that a driver is developed for Unikraft's driver APIs.

Additionally, this PR switches off the option that multiple platforms images can be compiled with one configuration. This mode is not supported anymore since we started re-architecting the platforms by moving re-usable and (potentially) shared code under `drivers/`. Because this work is ongoing, this PR is providing a temporary fix.


plat/Config.uk
```diff
@@ -115,3 +115,12 @@ config FPSIMD
 	depends on ARCH_ARM_64
 	help
 		Enable support FPU usage in application
+
+# FIXME: Temporary workaround to display a warning when more than one platform
```
oleksiimoisieiev, line 119: I think this is reasonable to make menu where only one platform can be selected. This is because (at least as I saw) KVM and XEN has different entry functions with separate init sequence. In the Linux Kernel, where init sequence is generic - you can enable support for both platforms. So my standpoint is that (at least for now) only one platform should be selected. If your intention is to use both drivers, then you may consider renaming PLAT_{XEN,KVM} to something else. You can take a look in ARM-TF as the reference.

oleksiimoisieiev: Maybe it's better to set depend PLAT_XEN from !PLAT_KVM and vice versa?

drivers/xen/xenapi/Config.uk
```diff
@@ -0,0 +1,5 @@
+config LIBXENAPI
```
oleksiimoisieiev: I would rename it to LIBXEN similar to another projects, such as https://github.com/xen-troops/zephyr-xenlib.

drivers/xen/consfront/Config.uk
```diff
@@ -1,8 +1,8 @@
-menuconfig LIBXEN_CONSOLE
-	bool "Hypervisor console"
+config LIBCONSFRONT
+	bool "consfront"
```
oleksiimoisieiev: I would make it LIBXENCONSOLE. That's because there is no such entity as xenconsback. Blk, net, 9p are calling fronts because there are backends for them. Such backends may be called as process which is running in some domain. Typically DomD and you can configure domain with running backend. For 9p it's qemu process.
Same for sndbe and displbe.

drivers/xen/Makefile.uk
```diff
@@ -12,4 +12,4 @@ $(eval $(call import_lib,$(UK_DRIV_XEN_BASE)/blkfront))
 $(eval $(call import_lib,$(UK_DRIV_XEN_BASE)/netfront))
```
oleksiimoisieiev, line 13: drivert -> driver in the commit description.

plat/xen/Config.uk
```diff
@@ -11,7 +11,6 @@ menuconfig PLAT_XEN
        select LIBNOLIBC if !HAVE_LIBC
        select HAVE_FDT if ARCH_ARM_64
        imply LIBFDT if (ARCH_ARM_32 || ARCH_ARM_64)
-       select XEN_DBGEMERGENCY if ARCH_ARM_32
```
oleksiimoisieiev, line 13: no mention about this change in the commit description.

oleksiimoisieiev (changes requested):

oleksiimoisieiev:
Review is still in progress

oleksiimoisieiev:
Review done

andreistan26 (approved):
Thanks a lot for the changes, apart from the Oleksii's comments I don't have anything to add.

plat/Config.uk
skuenzer: Yes, the idea is to change the menu in order to let only one platform selectable. Because we also support platforms from external repos, we need to sit down first and come up with a solution on how we can properly auto-generate the `choice`. Because we originally supported multiple platform images at the same time, we could simply include each platform's `Config.uk` and compile the menu that way. Problem is that with a `choice` this is a bit more difficult: we get basically get two separated blocks in KConfig: 1) the `choice` listing all platforms, 2) the configuration items of each platform and they must be encapsulated with `if` condition so that they are only visible depending on the chosen platform. For building this mechanics, we need to be able to somehow extract the name of a platform form a platform repo.
Because all of this, the patch here is really for bridging the gap until we make it clean. Since we began already re-architecting the driver organization (wip, ongoing), we broke already the property that allowed us to compile multiple images at the same time. So this must be avoided that this can be configured and we won't go back to support multiple builds.

skuenzer: > Maybe it's better to set depend PLAT_XEN from !PLAT_KVM and vice versa?

Hum, this is also an option how we could do this temporarily. However, transition a config from KVM to XEN and vise versa is less obvious with that. I guess this is mostly preference. What do you think? What do you think @michpappas ?

drivers/xen/xenapi/Config.uk
skuenzer: Hum, Zephyr's Xenlib looks to me like a complete platform library and thus richer than just a place to maintain the unmodified copy of the official Xen headers retrieved from: [xen.git/xen/include/public](https://xenbits.xen.org/gitweb/?p=xen.git;a=tree;f=xen/include/public;h=316fe2841855015f6da322d37513569777b41b24;hb=bd51e573a730efc569646379cd59ccba967cde97). Background is that I expect that we want to move code all code that is under `plat/xen` (and actually `plat/`) into individual driver libs, for example `libxenhypercall`. A platform is then nothing else than the selection (via imply) of according drivers. However, the shared headers need to be accessible for these micro-driver libraries: this is what I intended `LIBXENAPI` for.
I was also original thinking to call it `LIBXENHEADERS` or `LIBXENHDRS` to make that a bit clearer. Should we rename it to that?

plat/xen/Config.uk
skuenzer: Good point. Actually, I should keep the line and only adopt it to the updated driver name and update it to use an `imply` instead of `select` (coming from re-arch considerations). We decided some time back to retire Xen/Arm32 support because it is unmaintained and outdated. I also do not have a machine anymore where I could test it. However, that should go to a different PR anyways. Do you have use and interest for Xen/Arm32?

drivers/xen/consfront/Config.uk
skuenzer: Funnily, in Xen's Mini-OS, they named the driver [`consfront`](https://xenbits.xen.org/gitweb/?p=mini-os.git;a=tree), they also have `fbfront` for a FrameBuffer/Keyboard driver - they got completely convinced with the `*front` naming, independently if there is a backend. However, their [Windows PV drivers](https://xenbits.xen.org/gitweb/?a=project_list;pf=pvdrivers/win) are named with `xencons`, `xennet`, `xenblk`... So, I must say I like your reasoning and would adopt it to `LIBXENCONS`.

drivers/xen/xenapi/Config.uk
andreistan26: IMO `LIBXENHEADERS` is clearer.

michpappas: In the current state of this PR `Xen -> Xen API headers` appears in the `Device Drivers` menu also when KVM is enabled. Given that you can't do much by selecting this library alone, i.e. without any drivers that implement the Xen API, would it make more sense if `CONFIG_LIBXENAPI` was a hidden option? If the reasoning is for some external platforms that may potentially implement the Xen API, IMO it's still okay to have this as an invisible option that these drivers would select.

drivers/xen/xenapi/Config.uk
```diff
@@ -0,0 +1,5 @@
...
+	bool "Xen API headers"
+	select LIBNOLIBC if !HAVE_LIBC
```
michpappas: Not sure if the dependency of libc is necessary here. From a quick look (well, actually grep) it seems that the only `libxenvchan.h` includes headers from libc, which can be easily resolved by switching to kernel types.

drivers/xen/9pfront/Makefile.uk
```diff
@@ -0,0 +1,8 @@
+$(eval $(call addlib_s,lib9pfront,$(CONFIG_LIB9PFRONT)))
+
+# FIXME: Extra includes should be removed
```
michpappas: Here and it the rest or rename commits, what exactly does this comment refer to?

plat/xen/arm/setup64.c
```diff
@@ -29,7 +29,7 @@
 #include <xen-arm/os.h>
 #include <xen-arm/mm.h>
 #include <common/events.h>
-#include <xen/console.h>
+#include <uk/xen/console.h>
```
michpappas: This breaks the build if `libukconsole` is not selected. The same applies to `plat/xen/x86/setup.c` and `plat/xen/shutdown.c`. (well, chances are it was broken already, but maybe we can fix it with the opportunity of this PR).

drivers/xen/consfront/include/uk/xen/console.h
michpappas, line 1: Regarding the rename from `include/xen/console.h` and `include/uk/xen/console.h`, I notice some inconsistency among Xen driver libraries: While `xenbus` and `consfront` use `include/uk/xen`, the `xenapi` uses `include/xen`.

plat/Config.uk
michpappas: If the user declares an external platform via the `P` parameter in `Makefile.uk`, perhaps it's reasonable to assume that the user intends to build that platform. In that case we can include the `Config.uk` of the external platform only, and not show the internal ones. Of course it comes with a tradeoff, i.e during development one may with to switch between two platform builds, but then again during development once needs to change things around anyway.

drivers/ukconsole/ns16550/Config.uk
```diff
@@ -1,150 +1,149 @@
-menuconfig LIBUKCONSOLE_NS16550
+menuconfig LIBNS16550
 	bool "NS16550 / 8250"
 	depends on (ARCH_ARM_64 || ARCH_X86_64)
+	depends on (!PLAT_XEN && ARCH_X86_64)
```
michpappas: So since this is replaced in a following commit with `HAVE_IBMPC_NS16550` this intermediate step should probably doesn't need to be in the history.

drivers/ukconsole/ns16550/ns16550.c
```diff
@@ -310,7 +310,7 @@ static int early_init(struct ukplat_bootinfo *bi)
 		return rc;
 	}
 
-	uk_console_init(&earlycon.dev, "NS16550", &ns16550_ops,
+	uk_console_init(&earlycon.dev, "ns16550", &ns16550_ops,
```
michpappas: Actually the device name is `NS16550` (with uppercase): https://archive.org/details/bitsavers_nationaldamunicationsElementsDataBook_19316911

drivers/ukconsole/ns16550/ns16550.c
```diff
@@ -356,7 +356,7 @@ static int fdt_get_device(struct ns16550_device *dev, const void *dtb,
 
 	uk_pr_debug("ns16550 @ 0x%lx - 0x%lx\n", reg_base, reg_base + reg_size);
 
-	uk_console_init(&dev->dev, "NS16550", &ns16550_ops, 0);
+	uk_console_init(&dev->dev, "ns16550", &ns16550_ops, 0);
```
michpappas: Same here, the uppercase was correct: https://archive.org/details/bitsavers_nationaldamunicationsElementsDataBook_19316911

drivers/ukconsole/ns16550/Config.uk
```diff
@@ -1,23 +1,28 @@
+# Port-IO mode for IBM PC/AT compatible machines
+config HAVE_IBMPC_NS16550
+	bool
+	depends on ARCH_X86_64
```
michpappas, line 4: I wonder if `depends` is necessary here, given that it's an invisible option, and given that in certain platforms (Firecracker) we select that option conditionally to the architecture anyway (i.e. `select HAVE_IBMPC_NS16550 if ARCH_X86_64`). Perhaps we can make all selections of `IBMPC` conditional to `ARCH_X86_64` at the platform configuration?

plat/kvm/Config.uk
```diff
@@ -97,9 +97,9 @@ config KVM_VMM_QEMU
 	bool "QEMU"
 	imply LIBUKCONSOLE_PL011 if ARCH_ARM_64
 	imply LIBUKCONSOLE_PL011_EARLY_CONSOLE if ARCH_ARM_64
-	imply LIBUKCONSOLE_NS16550 if ARCH_X86_64
-	imply LIBUKCONSOLE_NS16550_EARLY_CONSOLE if ARCH_X86_64
-	imply LIBUKCONSOLE_NS16550_COM1 if ARCH_X86_64
+	imply LIBUKCONSOLE_NS16550 if LIBUKCONSOLE && ARCH_X86_64
```
michpappas: Good catch, we missed making this change when introducing `libukconsole` 🙏🏼

michpappas (changes requested):
Hi @skuenzer just a few comments from my side. +1 from me on the proposed change in the driver's naming convention (btw first time we have libraries without the uk prefix!).

drivers/xen/xenapi/Config.uk
skuenzer: I am renaming 👍🏻

plat/Config.uk
skuenzer: Good arguments and I like your suggestion @michpappas. I will do the following: I am implementing for this PR @oleksiimoisieiev's suggestion with `PLAT_XEN depends on !PLAT_KVM` (and vice versa) to cover the configuration problem we have with the current tree.
The build system does stop compiling if we have too many platforms configured (e.g., when external platforms are added and selected): so this is a okay~ish behavior: at least nothing invalid is compiled. The last missing part with the menu, I would do in a separate PR because of the amount of changes that this involves is not covered by the intention of this PR anymore. I think there is also some discussion needed to get to a final conclusion.

skuenzer: Update: This change causes a circular dependency with KConfig 😭 :
```diff
diff --git a/plat/kvm/Config.uk b/plat/kvm/Config.uk
index c6640cef3..8ec06d16f 100644
--- a/plat/kvm/Config.uk
+++ b/plat/kvm/Config.uk
@@ -1,6 +1,6 @@
 menuconfig PLAT_KVM
        bool "KVM guest"
-       default n
+       depends on !PLAT_XEN
        depends on (ARCH_X86_64 || ARCH_ARM_64)
        select LIBUKDEBUG
        select LIBUKBOOT
diff --git a/plat/xen/Config.uk b/plat/xen/Config.uk
index 81dbe2f2f..cfdcd34a6 100644
--- a/plat/xen/Config.uk
+++ b/plat/xen/Config.uk
@@ -1,6 +1,6 @@
 menuconfig PLAT_XEN
        bool "Xen guest image"
-       default n
+       depends on !PLAT_KVM
        depends on ((ARCH_X86_32 && !HAVE_SMP) || (ARCH_X86_64 && !HAVE_SMP) || (ARCH_ARM_32 && !HAVE_SMP) || (ARCH_ARM_64 && !HAVE_SMP))
        depends on ((ARCH_X86_32 && !HAVE_SYSCALL) || (ARCH_X86_64 && !HAVE_SYSCALL) || (ARCH_ARM_32 && !HAVE_SYSCALL) || (ARCH_ARM_64 && !HAVE_SYSCALL))
        depends on !HAVE_PAGING
```

Hum, I would say that we keep then this warning message for now and solve it with a separate PR where we clean-up the platform configuration menu. It looks like a `choice` is needed.

drivers/xen/xenapi/Config.uk
skuenzer: Which is actually correct behavior and intention was the example that you just gave. But it is also true that it might be confusing, let's hide it 👍🏻

drivers/xen/xenapi/Config.uk
skuenzer: The headers are unmodified copies from the official Xen sources. I would like to keep it this way so that future updates of the Xen API headers (sync from Xen project) does not need to take modifications into account. Because of this, the dependency is actually correct and needed.

plat/xen/arm/setup64.c
skuenzer: Good point 👍🏻

drivers/xen/consfront/include/uk/xen/console.h
skuenzer: This is on purpose so that the official Xen API headers are under `xen/`. This would also allow compiling some user space Xen tools (in principle). Everything Xen-related that we provide goes under `uk/xen/`.

drivers/ukconsole/ns16550/Config.uk
skuenzer: An artifact of building a history that is commit-wise "bisectable". A preference of style IMO.

plat/xen/arm/setup64.c
skuenzer: Hum these are in fact artifacts that should be removed. Only `shutdown.c` requires access to a console function. I will guard it there.

drivers/ukconsole/ns16550/ns16550.c
skuenzer: I'll change it back.

drivers/xen/9pfront/Makefile.uk
skuenzer: That libraries should not just include some foreign headers by themselves if there is no good reason. This is a leftover that needs fixing with rearch.

drivers/ukconsole/ns16550/Config.uk
skuenzer: Hum.. I think it is a question of how complete we are here. The thing what I wanted to express is that `HAVE_IBMPC_NS16550` is a pure x86 standard because it implies `portio` instructions at specific addresses but you have a point.

skuenzer:
@michpappas, @oleksiimoisieiev I just pushed a new version addressing (most) of your comments. A few items are open, I kept them unresolved in the previous discussion threads.

drivers/xen/9pfront/Makefile.uk
michpappas: Okay, I guess you referring to the direct inclusions of xenplat / and xenbus. Then maybe we can replace "extra headers" with a more clear explanation based on what you posted above? Or separate these two `CINCLUDES` from others so that it's more clear that the comment refers to these? On some drivers that depend on `plat ` we have marked such dependencies similarly to:
```
...
LIBUKBUS_PCI_CINCLUDES-y += -I$(LIBUKBUS_PCI_BASE)/include

# TODO Remove as soon as plat dependencies go away
LIBUKBUS_PCI_CINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include

LIBUKBUS_PCI_SRCS-y += $(LIBUKBUS_PCI_BASE)/pci_bus.c
...
```

drivers/xen/consfront/include/uk/xen/console.h
michpappas: Okay thanks, I wasn't aware 👍🏼

drivers/ukconsole/ns16550/Config.uk
michpappas: Hmmm okay, let it be then.

drivers/xen/xenapi/Config.uk
michpappas: Maintainability is definitely a valid point. My only concern is whether such a dependency may make sense in the environment where these headers originate from, but on in our case. Anyway, I see Xen already depends on `HAVE_LIBC` in other configs, and most notably in `plat/xen/Config.uk`, so I guess for now it's okay. It would be down to the maintainers of Xen (in Unikraft) to take a decision on whether to depend on a stdlib. The other problem is that we don't have very fundamental features like `memcpy()`, without at least nolibc, which is another big topic to discuss in the scope of rearch. So in conclusion let's leave it as is for now then, until this topic is addressed altogether.

drivers/xen/9pfront/Makefile.uk
skuenzer: Yes, correct: Will do that :+1:

drivers/ukconsole/ns16550/Config.uk
michpappas: > The thing what I wanted to express is that HAVE_IBMPC_NS16550 is a pure x86 standard because it implies portio instructions at specific addresses

That also makes sense, as this is the authoritative source of information for this config option. My only concern is that it will look very strange on the platform's `Config.uk` without the condition of `ARCH_X86_64`. Saying that,  I think it won't hurt if  people define it conditional to `ARCH_X86_64` on the platform's `Config.uk` too. If you agree, let's leave it as is here, and add a (redundant) condition at the platform's side too.

drivers/xen/9pfront/Makefile.uk
skuenzer: Yes, exactly. I tried a bit and found finally a solution that removes these lines already from the Xen drivers. There are two more commits on this PR that I'd like you to check. But basically I achieved it by the following:
- Properly define and export the public API of `drivers/xen/xenbus`
- Similar to other libraries, export headers and compile flags of `libxenplat` globally: I basically define an API.
Because of this and the fact that I try to keep my commits "bisect"-able, I did not change the message because the latest commits will remove it again.

drivers/xen/xenapi/Config.uk
skuenzer: I am not fully understand what you mean, because the line says to include `libnolibc` if there is no other `libc` (`!HAVE_LIBC`). I think this is totally reasonable this way. This provides us the required data types for the headers. In general (also for other libraries) at least `nolibc` is almost always needed, just because of `memcmp` or `memcpy`.

michpappas: That's right, but the problem is that nolibc keeps growing and I feel we haven't defined a clear scope for it, or at least I am not aware of one. For many cases we indlude `nolibc` just for errno-compatible exit codes, and a few primitives like `memcpy()`. Anyway, as said earlier this is out of the scope of this PR, so I think we can resolve this comment and resume discussion in the scope of rearch.

drivers/ukconsole/ns16550/ns16550.c
michpappas: Looks like this one hasn't been fixed, unresolving.

drivers/ukconsole/ns16550/ns16550.c
michpappas: This one hasn't been fixed either, unresolving here too.

drivers/ukconsole/ns16550/Config.uk
michpappas: Actually it looks like you're doing that already, so it's all good. Please feel free to resolve.
```
plat/kvm/Config.uk:	select HAVE_IBMPC_NS16550 if ARCH_X86_64
```

plat/Config.uk
michpappas: > The last missing part with the menu, I would do in a separate PR because of the amount of changes that this involves is not covered by the intention of this PR anymore

Ok from my side 👍🏼 Perhaps open an issue for our backlog with discussion so far? (okay if it merely links here)

drivers/xen/9pfront/Makefile.uk
michpappas: Nice 👌🏼 I like this intermediate step approach, as it allows at least drivers to be clean.

>    With this change, the compile flag `CONFIG_PARAVIRT` is name spaced to  `XEN_PARAVIRT`, as well.

Could you elaborate on the commit message with some reasoning for that change?

>  Because of this and the fact that I try to keep my commits "bisect"-able, I did not change the message because the latest commits will remove it again.

I'm all on for bisectability, but IMO that is not a reason to introduce transient changes. The commits in a PR should reflect the final state of the work, and not the development process. So to go by the book, the API should had been defined in an earlier commit. Anyway, I know restructuring commits is a mundane task, so if you are satisfied with the current state, I won't stand on the way.

michpappas:
@skuenzer I'm done with this round. Please notice that there are a few issues reported by checkpatch.

drivers/ukconsole/ns16550/ns16550.c
skuenzer: There was probably a mistake happening during push. It is back in now.

drivers/ukconsole/ns16550/ns16550.c
skuenzer: There was probably a mistake happening during push. It is back in now.

oleksiimoisieiev:
LGTM. Please fix conflitcs.
`Reviewed-by: Oleksii Moisieiev <oleksii_moisieiev@epam.com>`

drivers/xen/xenapi/Config.uk
skuenzer: `nolibc's` definition is to support pure core repository builds and it should be also dead-code-elimination friendly.

skuenzer:
> LGTM. Please fix conflitcs. `Reviewed-by: Oleksii Moisieiev <oleksii_moisieiev@epam.com>`

@oleksiimoisieiev @michpappas Thanks a lot for the reviews. I updated the PR and solved the conflict caused by #1517. I noticed that in that PR, a `select` instead of a `depends on` should have been used and that `9pfront` and `blkfront` require grant tables as well. With `depends on` it would been slightly harder to achieve automatic platform driver configuration if an application states network dependency by just selecting `uknetdev`.
I tried fixing these in this PR by also decomposing `libxenplat` further: I moved the `gnttab` code into an own library and stated the dependencies accordingly.
In principle this decomposition can be continued as clean-up process of the Xen platform code. Next step could be to move the event channel implementation into an own library. However, I wanted to stop such efforts because I think we are already adding too much additional work into this PR which had its actual focus on serial consoles drivers.

drivers/xen/xengnttab/Makefile.uk
```diff
@@ -0,0 +1,7 @@
+$(eval $(call addlib_s,libxengnttab,$(CONFIG_LIBXENGNTTAB)))
+
+CINCLUDES-$(CONFIG_LIBXENGNTTAB)        += -I$(LIBXENGNTTAB_BASE)/include
+
+LIBXENGNTTAB_SRCS-y                     += $(LIBXENGNTTAB_BASE)/gnttab.c
+LIBXENGNTTAB_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBXENGNTTAB_BASE)/x86/gnttab.c|x86
```
michpappas: Unless we have a different convention on Xen, under the current conventions these should be under `$(LIBXENGNTTAB_BASE)/arch/arm64/` and `$(LIBXENGNTTAB_BASE)/arch/x86_64/`

drivers/xen/9pfront/Config.uk
```diff
@@ -2,6 +2,7 @@ menuconfig LIB9PFRONT
 	bool "9pfront: Xen 9pfs volumes"
 	select LIBXENBUS
 	select LIBXENHEADERS
+	select LIBXENGNTTAB
```
michpappas, line 5: In #1517 `XEN_GNNTAB` was selected conditional to `XEN_PV || ARCH_ARM_64`. Should we be preserving that condition in drivers?

skuenzer: Actually, the default `y` wasn't making much sense and breaking the minimalism of the image. Grants are the mechism for Xen to share pages from a guest with a driver domain, this is especially needed for virtual devices. I was going over the code and saw that just 9pfront, blkfront, and netfront are making use of this. For kernels that do not use any of these drivers, we can build a more minimal kernel without page grants.

drivers/xen/xengnttab/Makefile.uk
skuenzer: Sure, I will adopt it. I kept the original structure during the moving and currently think that we might be too strict with our convention. But you are right, we should stick to it and update the convention first.

drivers/xen/9pfront/Makefile.uk
skuenzer: > Could you elaborate on the commit message with some reasoning for that change?

I updated the commit message to point out that with the platform rearch we are basically removing the platform-internal scope where drivers were originally in. This means that API definitions are public usable, even by application code. Because of this, proper name spacing is required.

skuenzer: > I'm all on for bisectability, but IMO that is not a reason to introduce transient changes. The commits in a PR should reflect the final state of the work, and not the development process. So to go by the book, the API should had been defined in an earlier commit. Anyway, I know restructuring commits is a mundane task, so if you are satisfied with the current state, I won't stand on the way.

If you mean the exporting of the XenBus API, the commit is actually doing only that. The API is not redefined or changed, just "properly" exported by moving files. But I agree with your general statement that irrespective of bisect-able commits or not: transition commits do not make much sense and should be avoided. Every commit should directly go towards the final target. With a proper commit history, this is typically feasible.

michpappas (approved):
@skuenzer I think it looks good now. Thank you for going through the lengthy comments on this one.

skuenzer:
> @skuenzer I think it looks good now. Thank you for going through the lengthy comments on this one.
> 
> Reviewed-by: Michalis Pappas [michalis@unikraft.io](mailto:michalis@unikraft.io)

Thanks @michpappas. I just pushed a last update because #1518 got merged. I only removed this commit from the PR: `plat/xen/arm64: Fix invalid assignments to boot info struct`

michpappas:

PR #1527, opened 2024-10-21
lib/ukprint: Introduce libukprint

### LINKED ISSUE #665: `lib/ukprint`: Configurable log format and printing

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series introduces `libukprint` that is entasked to handle all printing and logging facilities. The changes are summarized below:
* `libukdebug/asmdump` is deprecated, as it relies on `zydis`, and effectively was never used.
* The implementation of the `uk_printk()` API and the `uk_pr_` helpers are migrated from `libukdebug` into the newly introduced `libukprint`.
* Debug printing is now done via a new `uk_printk()` severity level, `UK_PRINT_KLVL_DEBUG`.  `uk_printd()` and console redirection are deprecated. The `uk_pr_debug()` helper is implemented by a call to `uk_printk()` with `UK_PRINT_KLVL_DEBUG`. The functionality of enabling debug prints on individual files by defining `UK_DEBUG` remains unchanged.
* The `uk_printk()` API is further enhanced with features that allow raw and unconditional printing. Raw printing does not prepend output with metadata and is implemented by a new flag, `UK_PRINT_RAW`. Unconditional printing implements messages that appear regardless of the configured severity level and is implemented by a new pseudo-severity flag `UK_PRINT_KLVL_NONE`. These features are useful for use-cases of the `syscall_shim` and `uktest`. 
* `ukstore` entries are added to `libukprint` to allow derive and update the severity level at runtime. These can later be used to implement control via `/proc/sys/kernel/printk`.
*  Introduce logbuffer support. This provides `dmesg` functionality to Unikraft via the newly introduced `uk_print_dmesg()`. The logbuffer can be enabled independently of the kernel console.
* Headers of `libukprint` are always included in the build. This formalizes convention that calls to kernel printing facilities from various libraries do not need to be conditionally compiled. Moreover, libraries do not need to select `LIBUKPRINT` in their `Config.uk`.
* Various updates on depending libraries.


michpappas:
TBD:
* Add cmdline control.
* Verbosity controlled via cmdline / ukstore is bound to `UK_PRINT_KLVL_MAX`. Normally it is useful to increase verbosity for investigation.
* Runtime controlled verbosity works as expected if `UK_DEBUG` is not set.
* Finalize `uktest` / `syscall_shim` flags.
* Document required Kconfig changes in `Kraftfile` (if any)

lib/ukprint/include/uk/print/store.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef __UK_PRINT_STORE_H__
+#define __UK_PRINT_STORE_H__
+
+#include <uk/store.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* ukstore entry tha controls the verbosity at runtime.
```
mogasergiu: s/tha/that

lib/syscall_shim/uk_syscall_binary.c
```diff
@@ -176,18 +173,7 @@ static void binary_syscall_strace(struct uk_syscall_exit_ctx *exit_ctx)
...
-	 * library once that exists. We also don't want to print all the meta
-	 * data that `uk_pr_info` includes. Note also that right now, debug
-	 * print calls also turn into a no-op if `ukconsole` is not available.
-	 */
-#if CONFIG_LIBUKCONSOLE
-	uk_console_out(prsyscallbuf, (__sz)prsyscalllen);
-#else /* !CONFIG_LIBUKCONSOLE */
-	uk_pr_info(prsyscallbuf);
-#endif /* !CONFIG_LIBUKCONSOLE */
+	uk_printk(UK_PRINT_KLVL_INFO | UK_PRINT_RAW, "%s", prsyscallbuf);
```
mogasergiu: Oops, I think you meant to use just the `UK_PRINT_RAW` flag. Otherwise you cannot get `strace` without INFO level messages enabled. (I do see there was a `uk_pr_info` there at some point?? I don't know why)

lib/ukprint/console.c
```diff
@@ -0,0 +1,306 @@
...
+
+	return len;
+}
+#endif /* CONFIG_LIBUKCONSOLE */
+
+/* Console state for kernel output */
+static struct vprint_console kconsole  = {
+#if CONFIG_LIBUKCONSOLE
+	.cout = console_out,
+#else
```
mogasergiu: I think you forgot the `!CONFIG_LIBUKCONSOLE` here. There also other `#if`/`#else`/`#endif`s here missing it. Won't mention each one as you can easily see them now that I brought it up.

lib/ukprint/console.c
```diff
@@ -0,0 +1,306 @@
...
+#endif /* CONFIG_LIBUKCONSOLE */
+	.newline = 1,
+	.prevlvl = INT_MIN
+};
+
+static inline void vprint_cout(struct vprint_console *cons,
+			       const char *buf, __sz len)
+{
+	if (cons->cout)
+		cons->cout(buf, len);
```
mogasergiu, line 134: I notice this returns a `__ssz`. Should that matter? Though it's not like we could print a warning/error message here in this scenario :rofl:

lib/ukprint/console.c
```diff
@@ -0,0 +1,306 @@
...
+	int prevlvl;
+};
+
+#if CONFIG_LIBUKCONSOLE
+/* TODO: Some consoles require both a newline and a carriage return to
+ * go to the start of the next line. This kind of behavior should be in
+ * a single place in posix-tty. We keep this workaround until we have feature
+ * in posix-tty that handles newline characters correctly.
+ */
+static inline __ssz console_out(const char *buf, __sz len)
```
mogasergiu: Probably a remnant from development but `inline` I think is redundant here.

lib/ukprint/console.c
```diff
@@ -0,0 +1,306 @@
...
+			  "{r:%p,f:%p} ", (void *)ra, (void *)fa);
+	vprint_cout(cons, (char *)buf, len);
+}
+#endif /* CONFIG_LIBUKPRINT_PRINT_CALLER */
+
+void uk_print_console_write(struct uk_print_msg *msg)
+{
+	char lbuf[BUFLEN];
+	int len, llen;
+	const char *msghdr = NULL;
```
mogasergiu: Should `__NULL` be used instead since I see internal types everywhere else? (Ditto for the other `NULL`s in this file)

lib/ukprint/console.c
```diff
@@ -0,0 +1,306 @@
...
+				    strlen(msghdr));
+#if CONFIG_LIBUKPRINT_PRINT_THREAD
+			print_thread(cons);
+#endif
+#if CONFIG_LIBUKPRINT_PRINT_CALLER
+			print_caller(cons, msg->retaddr, msg->frameaddr);
+#endif
+			if (libname) {
+				vprint_cout(cons, LVLC_RESET LVLC_LIBNAME "[",
+					    strlen(LVLC_RESET LVLC_LIBNAME) + 1);
```
mogasergiu: I believe `LVLC_*` are all string literals. Maybe we could replace `strlen`s with a plain `sizeof`

lib/ukprint/uk_print_msg.h
```diff
@@ -0,0 +1,48 @@
...
+	__u16 libid;
+#if CONFIG_LIBUKPRINT_PRINT_SRCNAME
+	const char *srcname;
+	unsigned int srcline;
+#endif /* CONFIG_LIBUKPRINT_PRINT_SRCNAME */
+#if CONFIG_LIBUKPRINT_PRINT_TIME
+	__nsec timestamp;
+#endif /* CONFIG_LIBUKPRINT_PRINT_TIME */
+#if CONFIG_LIBUKPRINT_PRINT_THREAD
+	const char *threadname; /* may be NULL */
```
mogasergiu: Interesting, this seems to be the only useless field and is never used :thinking: . Its `UK_PRINT_MSG_ARGS_CALLER` does nothing

EDIT: It's been like this forever, I guess no point in pondering on it now.

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,144 @@
...
+
+int uk_print_logbuf_write(struct uk_print_msg *msg)
+{
+	struct uk_print_logmsg *logmsg;
+	va_list cp;
+
+	uk_spin_lock(&logbuf_lock);
+
+	va_copy(cp, msg->ap);
+	msg->len = vsnprintf(msg->msg, 0, msg->fmt, cp);
```
mogasergiu: I see that `msg->msg` is a flexible array but all the places that pass the `uk_print_msg ` argument to this function declare it as a plain stack local variable. Is it really safe to write to `msg->msg` here? Or maybe I am missing something.

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,144 @@
...
+
+	if (LOGMSG_SIZE(msg->len) > sizeof(logbuf)) {
+		uk_spin_unlock(&logbuf_lock);
+		return -EMSGSIZE;
+	}
+
+	logmsg = (struct uk_print_logmsg *)logbuf_next();
+
+	/* If we exceed the buffer, wrap around */
+	if ((__uptr)logmsg + LOGMSG_SIZE(msg->len) > (__uptr)logbuf + sizeof(logbuf))
```
mogasergiu: Doesn't the previously called `logbuf_next` already take care of this?

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,144 @@
...
+}
+
+static inline void update_head(struct uk_print_logmsg *logmsg, __sz msg_len)
+{
+	struct uk_print_logmsg *lm;
+	int found = 0; /* bool */
+
+	if (RANGE_OVERLAP((__uptr)logbuf_head, LOGMSG_SIZE(logbuf_head->msg.len),
+			  (__uptr)logmsg, LOGMSG_SIZE(msg_len))) {
+		for (lm = logbuf_head; lm != lm->next; lm = lm->next) {
```
mogasergiu: Initially, the tail points to itself. Is it possible this might not iterate correctly and skip the tail? Also, but maybe I am just too paranoid, but I would add a `UK_ASSERT(lm)` for each iteration.

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,144 @@
...
+}
+
+int uk_print_logbuf_write(struct uk_print_msg *msg)
+{
+	struct uk_print_logmsg *logmsg;
+	va_list cp;
+
+	uk_spin_lock(&logbuf_lock);
+
+	va_copy(cp, msg->ap);
```
mogasergiu, line 83: I believe you might be leaking `va_list` objects - each `va_copy` should be paired with `va_end`, right?

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,144 @@
...
+	 * fmt and va_list will no longer be available.
+	 */
+	va_copy(cp, msg->ap);
+	vsnprintf(logmsg->msg.msg, msg->len, msg->fmt, cp);
+
+	if (!logbuf_head) {
+		logbuf_head = logmsg;
+		logbuf_tail = logmsg;
+	}
+	logmsg->next = logmsg;
```
mogasergiu, line 136: ~~I might be misunderstanding but, for the first message, you set `logmsg->next = logmsg`, then immediately overwrite it with `logbuf_tail->next = logmsg` (which is the same). This works, but for subsequent messages you're setting `logmsg->next = logmsg` before linking it, which temporarily breaks the list.~~

EDIT: Actually nevermind, I don't see anything wrong, I must have misunderstood something.

lib/syscall_shim/uk_syscall_binary.c
michpappas: Yup, thanks for catching that. I think previously it was set to `uk_pr_info` in lieu of something better :)

lib/ukprint/console.c
michpappas: Yes exactly... I think it's okay to ignore

lib/ukprint/console.c
michpappas: Yes and I'll also replace `uintptr_t` with `__uptr` and get rid of `stdint.h` and `stdarg.h`.

lib/ukprint/console.c
michpappas: Good idea. Unfortunately we still don't get rid of `strlen()`, because of some dynamic strings left around.

lib/ukprint/logbuf.c
michpappas: Actually no write in `msg->msg` takes place here, we just obtain the formatted string's size and later use it when we write to the log buffer. I'll leave a comment as indeed it's not obvious.

lib/ukprint/logbuf.c
michpappas: `logbuf_next` does not take into account the message length, it only receives the next free position in the buffer (incl. wraparound if the next position is past the buffer's end). Based on that position, we then take the message length into account with again the possibility to wrap around.

lib/ukprint/logbuf.c
michpappas: That refers to the case where the buffer contains a single message, and a second message that overlaps the first one. Notice that since there is an overlap, then the 2nd message wraps around the buffer, and the caller has already update its position to the beginning of the buffer:
```c
        /* Now take the message length into account.
         * If we exceed the buffer, wrap around.
         */
        if ((__uptr)logmsg + LOGMSG_SIZE(msg->len) > (__uptr)logbuf + sizeof(logbuf))
                logmsg = (struct uk_print_logmsg *)logbuf;
```
So now in the `for` loop is not taken as `lm->next` equals to `lm` (still one message so far in buf), and we fall through to correctly update the head to the new message:
```c
                if (!found) {
                        /* This logmsg is the last element in the buffer.
                         * Set head to the first element.
                         */
                        logbuf_head = (struct uk_print_logmsg *)logbuf;
                }
```

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,164 @@
...
+
+	/* Update previous tail before we do the write.
+	 * As we may be overwriting that member, performing
+	 * the update now ensures that we don't corrupt our
+	 * own memory.
+	 */
+	if (lb->tail)
+		lb->tail->next = logmsg;
+
+	memcpy(&logmsg->msg, msg, sizeof(*msg));
```
mogasergiu, line 120: I believe you could do direct assignment here.

lib/ukprint/logbuf.c
```diff
@@ -0,0 +1,164 @@
...
+
+	UK_ASSERT(lb);
+	UK_ASSERT(msg);
+
+	uk_spin_lock(&lb->lock);
+
+	va_copy(cp, msg->ap);
+
+	/* pass zero size to obtain formatted string's length */
+	msg->len = vsnprintf(msg->msg, 0, msg->fmt, cp);
```
mogasergiu: So I noticed that we don't actually check if `vsnprintf` returned an error. But besides this... I just noticed that in some places `uk_vsnprintf` is called while in others it is `vsnprintf`. They also seem to have the same implementation? :thinking:

mogasergiu (commented):
Just one `memcpy` nit (feel free to ignore if you want) and one, I think, important observation about `vsnprintf`.

lib/ukprint/logbuf.c
michpappas: I don't think it makes a difference: The `snprintf.h` header (private) defines the implementation of `uk_vsnprintf()` conditionally to nolibc (in case you want to build unikraft without [no]libc), yet the rest of the implementation depends on a libc, and that is also expressed in `Config.uk`. Also if you use musl you can't configure it without stdio facilities, incl. `vsnprintf()`. I'll push an update for consistency, but overall I think we have reached a point where further iterations are a waste of time.

mogasergiu (approved):

PR #1530, opened 2024-10-30
drivers/ukconsole/vga: Add initial sanity check

florommel:
### Prerequisite checklist

### Base target

 - Architecture(s): `x86_64`
 - Platform(s): `kvm`, `xen`
 - Application(s): N/A

### Additional configuration

 - `CONFIG_LIBUKCONSOLE_VGA=y`

### Description of changes

When the VGA driver is enabled in firecracker builds, the kernel fails the following assertion during boot:
```
<memory.c @  374> Assertion failure: ml->flags == mr->flags
```
caused by the illicitly registered memory region for the (in this case, non-existing) VGA controller.

This PR adds an initial sanity check for the VGA controller that should detect a non-existing VGA controller and also an unexpectedly configured VGA controller.

I have tested this with firecracker (no VGA) and QEMU (with VGA), but not yet with XEN.

An alternative to this PR would be to statically disallow VGA for firecracker builds.

### DIFF
--- a/drivers/ukconsole/vgacons/vga.c
+++ b/drivers/ukconsole/vgacons/vga.c
@@ -213,6 +213,35 @@ static struct uk_console_ops vga_ops = { .out = vga_console_out };
 
 static struct uk_console vga_dev;
 
+#define VGA_MISC_OUTPUT_REG 0x3cc
+#define VGA_IO_ADDRESS_SELECT_MASK 0b01
+#define VGA_ERAM_MASK 0b10
+#define VGA_GRAPHICS_ADDRESS_REG 0x3ce
+#define VGA_GRAPHICS_DATA_REG 0x3cf
+#define VGA_MISC_GRAPHICS_REG 0x06
+#define VGA_MEMORY_MAP_SELECT_MASK 0b1100
+
+static int vga_check_settings(void)
+{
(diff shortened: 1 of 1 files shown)


skuenzer:
Thanks for this PR! This is a great addition to the VGA driver. We are undergoing a re-architecture of the architecture, platform and driver code. Some things aren't finished yet, but the goal is that VGA should not be available for firecracker, Xen, and any arch, platform that doesn't support it at all. However, your patch makes still a lot of sense to add, considering the flexibility you have with QEMU.

florommel:
Okay, that seems reasonable.

I checked the relevant QEMU (v.9.1.1) options (`-vga none`, `-nodefaults`). It seems that QEMU provides the basic VGA controller functionality (including the memory region) even when VGA is disabled.
Nevertheless, I think the proposed sanity check is valuable. There may be QEMU configurations that I haven't tested (and it may also prove useful in other virtualization environments supported in the future).

skuenzer (approved):
Thanks a lot @florommel, the PR looks great, we will take it in.

michpappas:
Hi @florommel, merging this failed with conflicts, could you rebase?

razvand:
@michpappas , this is now rebased. Merge at your earliest convenience.

PR #1114, opened 2023-09-26
lib/ukdebug: Crash screen improvements

jschlumpp:
### Prerequisite checklist

 - [ ] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [ ] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): `x86_64`, `aarch64`
 - Platform(s): `kvm`, `xen`, `linuxu`
 - Application(s): N/A
 
### Description of changes

This adds a more informative crash screen and changes the assertion macro to use this crash screen. 

To allow probing memory in exception and interrupt handlers, it introduces support to temporarily disable IST on x86. Otherwise, the page fault handler of uknofault would overwrite the stack of the active exception/interrupt handler.

The PR also removes the x86 specific register dumps in unhandled exceptions and instead relies on the crash screen to print those.

PR #1531, opened 2024-11-05
lib/ukdebug: Crash screen improvements

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This is a rework of #1114 with re-arch conventions in mind. It effectively decouples the crash screen logic from the rest of the kernel by implementing it as an exception handler provided by `libukdebug`.

Depends-on: #1527
Deprecates: #1114


lib/ukdebug/Config.uk
```diff
@@ -12,6 +12,61 @@ config LIBUKDEBUG_ENABLE_ASSERT
 	help
 	  Build code with assertions.
 
+menuconfig LIBUKDEBUG_CRASH_SCREEN
+	bool "Enable crash report"
+	default y
+	depends on LIBUKPRINT
+	help
+		Output information on the kernel console if a crash occurred.
```
mogasergiu, line 20: ~~I know this commit is old and ported. I think the other configs here use correct indentation for the help message. We should update these as well.~~ (The new empty line added at the end might also be unnecessary) 

EDIT: Actually there may have been a confusion. We might actually indeed use tabs in this case, unlike LK KConfig styling. Resolving this and let's leave as is for now.

lib/ukdebug/crash.c
```diff
@@ -0,0 +1,138 @@
...
+#endif /* !CONFIG_LIBUKDEBUG_CRASH_ACTION_REBOOT */
+}
+
+UK_EVENT(UK_CRASH_EVENT);
+
+int uk_debug_crash_handler(void *data)
+{
+	struct uk_event_crash_parameter param;
+	struct ukarch_trap_ctx *ctx;
+#ifdef UKPLAT_LCPU_MULTICORE
```
mogasergiu: I don't think this macro exists (or ever existed before). Maybe we should just delete related code for now?

lib/ukdebug/crash.c
```diff
@@ -0,0 +1,138 @@
...
+	if (ukarch_compare_exchange_sync(&crash_cpu, -1, cpu_id) != cpu_id) {
+		/* TODO: Finish SMP Support
+		 * Freeze CPU or wait until the crash_cpu initiates a freeze
+		 * (e.g., through IPI). For now, just busy wait.
+		 */
+		ukplat_lcpu_halt();
+	}
+#endif /* UKPLAT_LCPU_MULTICORE */
+
+#ifdef CONFIG_LIBUKDEBUG_CRASH_SCREEN
```
mogasergiu: `#if`?

lib/ukdebug/crashdump.c
```diff
@@ -0,0 +1,72 @@
...
+
+#define STACK_WORDS CONFIG_LIBUKDEBUG_CRASH_PRINT_STACK_WORDS
+#define CRASH_PRINT_STACK_FMT				    \
+	(UK_HXDF_ADDR | UK_HXDF_GRPQWORD | UK_HXDF_ASCIISEC)
+
+#if CONFIG_LIBUKDEBUG_CRASH_PRINT_STACK
+void cdmp_gen_print_stack(unsigned long sp)
+{
+	char buf[64];
+	size_t ul_len = sizeof(unsigned long);
```
mogasergiu: `__sz`

lib/ukdebug/exportsyms.uk
```diff
@@ -1,2 +1,4 @@
+_uk_crashdumpk
```
mogasergiu: Does this symbol exist anymore ?

lib/ukdebug/include/uk/crash.h
```diff
@@ -0,0 +1,90 @@
...
+#include <uk/plat/bootstrap.h>
+#include <uk/arch/lcpu.h>
+#include <uk/essentials.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum uk_crash_reason {
+	UK_CRASH_REASON_INVALID,
```
mogasergiu, line 37: Since this is in `uk/debug/`, wondering if all of these should have been namespaced as `UK_DEBUG_CRASH` or maybe the header should have instead been `uk/crash.h` like is the case for `uk/trace.h` and `uk/assert.h`

lib/ukdebug/include/uk/debug/crash.h
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * Contains the crash reason.
+	 * The meaning of the following fields change depending on the reason.
+	 */
+	enum uk_crash_reason reason;
+
+	/**
+	 * Contains the errno code the systems exits with.
+	 */
+	int errno;
```
mogasergiu: Let's not forget that dumb musl name clash :sneezing_face:

lib/ukdebug/include/uk/debug/crash.h
```diff
@@ -0,0 +1,90 @@
...
+	__u64 arg2;
+
+	/**
+	 * For UK_CRASH_REASON_UNHANDLED_TRAP:
+	 *   error code of the trap
+	 */
+	__u64 arg3;
+};
+
+struct uk_event_crash_parameter {
```
mogasergiu: I would name this as `uk_crash_...`

lib/ukdebug/Makefile.uk
```diff
@@ -4,15 +4,26 @@ $(eval $(call addlib_paramprefix,libukdebug,debug))
...
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/gdbtargetxml.ld
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/gdbtargetxml.S
 LIBUKDEBUG_GDBTARGETXML_CDEPS += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/target.xml
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/gdbsup.c|isr
 
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/crashsup.c
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/dumpsup.c
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/crashdump.c
+
+LIBUKDEBUG_SRCS-y += $(LIBUKDEBUG_BASE)/crash.c # FIXME conditional?
```
mogasergiu: I believe we could revise this FIXME? Also wondering if we should do `|isr` for everything crash related... or maybe it doesn't matter as we are supposedly crashing anyway...

arch/x86/x86_64/include/uk/asm/traps.h
```diff
@@ -1,33 +1,8 @@
...
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
- * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
- * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
- * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
- * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
- * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
- * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- * POSSIBILITY OF SUCH DAMAGE.
+/* Copyright (c) 2021, Karlsruhe Institute of Technology. All rights reserved.
```
mogasergiu, line 2: Should this commit and its ARM equivalent actually come before the crash dump commit in the history? Saying this as the latter already uses `uk_errno` but it's only introduced afterwards.

lib/ukdebug/include/uk/debug/crash.h
```diff
@@ -0,0 +1,90 @@
...
+	/** Optional register state at the crash location */
+	struct __regs *regs;
+	/** Optional explicit crash description */
+	struct uk_crash_descr descr;
+};
+
+/**
+ * Event for crashes. Will receive a uk_crash_event_param as data argument.
+ */
+#define UK_CRASH_EVENT uk_crash_event
```
mogasergiu: I am not sure I understand the need for such a macro (I know it's from an old version of this work).

lib/ukdebug/arch/arm64/include/uk/arch/crash.h
```diff
@@ -0,0 +1,32 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
mogasergiu, line 1: I know this patch set is old. But IIRC we are trying to get rid of asm directories for header inclusion paths

lib/ukdebug/crash.c
michpappas: Wow I'm impressed how this survived that long. That directive was introduced speculatively very early along with `plat/`  (828efb76ebb235966a52c9c7f7b76ef3bec36ec6) and parasitically lived in parallel to `HAVE_SMP` when multicore was introduced, before anyone noticed, at which point it started being removed. I'll replace with `HAVE_SMP`.

lib/ukdebug/Makefile.uk
michpappas: Actually that file must be unconditionally compiled as it manages `UK_EVENT_UNHANDLED_EXCEPTION`, which involves actions taken regardless `CONFIG_LIBUKDEBUG_CRASH_SCREEN`, as well as `UK_CRASH()` / `UK_CRASH_EX()`. I'll just remove the comment.

lib/ukdebug/arch/arm64/include/uk/arch/crash.h
michpappas: Good catch, will replace with `arch/`

arch/x86/x86_64/include/uk/asm/traps.h
michpappas: Sure I can reorder the commits

lib/ukdebug/include/uk/debug/crash.h
michpappas: Adding some context here as that caught me by surprise. Apparently musl defines `errno` as a macro: https://git.musl-libc.org/cgit/musl/tree/include/errno.h#n16. Will switch to a different name.

lib/ukdebug/include/uk/crash.h
michpappas: So this generally against our conventions, but I agree it makes sense to make an exception here give the ubiquitous nature of `libukdebug`. Will do 👍🏼

lib/ukdebug/include/uk/debug/crash.h
michpappas: I think it's there to document the event in the public API, as you can't use `UK_EVENT()` in a header you include in more than one files.

plat/kvm/arm/exceptions.S
```diff
@@ -81,6 +83,12 @@
 	 */
 	msr     tpidrro_el0, x0
 
+	ldr	x0, =ukarch_except_switch_stack
```
mogasergiu: Should've this been `ur_ldr` instead?
Also nit: I believe you could replace `cmp` + `beq` with a `cbz`:
```
ldr x0, =ukarch_except_switch_stack
cbz x0, 0f                    // Compare and Branch if Zero
```
It could help with branch prediction as this check is, although important, useless unless we are crashing.

plat/kvm/x86/traps.c
```diff
@@ -151,19 +151,77 @@ static void tss_init(__lcpuidx idx)
...
 
+static __align(8)
+UKPLAT_PER_LCPU_ARRAY_DEFINE(struct seg_gate_desc64, cpu_idt, IDT_NUM_ENTRIES);
+
+static
+UKPLAT_PER_LCPU_DEFINE(__u8, idt_ist_disable_nesting);
+
+#define IDT_IST_SAVE_LEN 32
+/* Space for the IST values of all exception vectors */
+static __u8 idt_ist_saved[IDT_IST_SAVE_LEN];
```
mogasergiu: Should this also be per CPU?

plat/kvm/x86/traps.c
```diff
@@ -151,19 +151,77 @@ static void tss_init(__lcpuidx idx)
...
+	__u8 *disable_nesting;
+	unsigned int i;
+
+	disable_nesting = &ukplat_per_lcpu_current(idt_ist_disable_nesting);
+	UK_ASSERT(*disable_nesting < __U8_MAX);
+
+	if (*disable_nesting++)
+		return;
+
+	idt = ukplat_per_lcpu_current(cpu_idt);
```
mogasergiu: This might prove to be unsafe as you cannot know for sure if you can safely access the lcpu idx field, given that it is an arbitrary section that does not guarantee the state of gs_base. Perhaps we could use `lcpu_get_current_in_except` to get the index and then reference the variable through `ukplat_per_lcpu`.

plat/common/include/x86/traps.h
```diff
@@ -108,23 +86,23 @@ static inline int _raise_event_##event(int trapnr, struct __regs *regs,	\
 void do_##name(struct __regs *regs)					\
 {									\
 	int rc;								\
-	rc = _raise_event_##event(TRAP_##name, regs, 0);		\
+	rc = _raise_event_##event(TRAP_##name, str, regs, 0);		\
 	if (unlikely(rc < 0))						\
 		uk_pr_crit("trap handler returned error: %d\n", rc);	\
 									\
-	if (!rc)							\
+	if (rc == UK_EVENT_NOT_HANDLED)					\
```
mogasergiu, line 93: Should this be `unlikely(rc == UK_EVENT_NOT_HANDLED || rc < 0)` (ditto below)?

plat/common/arm/traps_arm64.c
```diff
@@ -164,6 +165,21 @@ static enum aarch64_trap esr_to_trap(__u64 esr)
...
+__u8 ukarch_except_switch_stack;
+
+void ukarch_push_nested_exceptions(void)
+{
+	ukarch_except_switch_stack++;
+}
+
+void ukarch_pop_nested_exceptions(void)
+{
+	--ukplat_per_lcpu_current;
```
mogasergiu: uhh, I think you are fixing this by mistake in the next commit :sweat_smile:

lib/ukdebug/crash.c
```diff
@@ -0,0 +1,153 @@
...
+	param.descr = descr;
+
+	/* Ignore return value, we can't handle it anyway */
+	uk_raise_event(UK_CRASH_EVENT, &param);
+
+	/* Halt or reboot the system */
+	crash_shutdown();
+}
+
+int uk_debug_crash_handler(void *data)
```
mogasergiu: Should this be `uk_crash_handler` instead and `static`?

lib/ukdebug/crashdump.h
```diff
@@ -0,0 +1,37 @@
...
+
+void crash_print_crashdump(struct __regs *regs);
+
+void cdmp_gen_print_stack(unsigned long sp);
+void cdmp_gen_print_call_trace_entry(unsigned long addr);
+
+/* Must be implemented in the architecture-specific ukdebug support */
+void cdmp_arch_print_registers(struct __regs *regs);
+void cdmp_arch_print_stack(struct __regs *regs);
+#if !__OMIT_FRAMEPOINTER__
```
mogasergiu: Should this also be guarded by `CONFIG_LIBUKDEBUG_CRASH_PRINT_CALL_TRACE`?

lib/ukdebug/include/uk/crash.h
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * For UK_CRASH_REASON_UNHANDLED_TRAP:
+	 *   error code of the trap
+	 */
+	__u64 arg3;
+};
+
+/**
+ * Event raised after a crashdump to give libraries a chance
```
mogasergiu, line 92: Ok so I guess we will keep this.
This might mean that when we touch already existing (or add new) events, we should also add a similar macro with docs for them, as there are several that don't do this.

lib/ukdebug/Makefile.uk
```diff
@@ -4,15 +4,26 @@ $(eval $(call addlib_paramprefix,libukdebug,debug))
...
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/gdbtargetxml.ld
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/gdbtargetxml.S
 LIBUKDEBUG_GDBTARGETXML_CDEPS += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/target.xml
 LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_GDBSTUB) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/gdbsup.c|isr
 
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/crashsup.c
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/arch/$(CONFIG_UK_ARCH)/dumpsup.c
+LIBUKDEBUG_SRCS-$(CONFIG_LIBUKDEBUG_CRASH_SCREEN) += $(LIBUKDEBUG_BASE)/crashdump.c
+
+LIBUKDEBUG_SRCS-y += $(LIBUKDEBUG_BASE)/crash.c
```
mogasergiu, line 20: since this uses inclusions that are guarded by `LIBUKDEBUG_CRASH_SCREEN`, maybe it should also be guarded by it? In which case, its `#if CONFIG_LIBUKDEBUG_CRASH_SCREEN` could also be removed.
(Btw I just noticed but what is the `LIBUKDEBUG_CINCLUDES-y  += -I$(LIBUKDEBUG_BASE)` above for?)

lib/ukdebug/arch/arm64/dumpsup.c
```diff
@@ -0,0 +1,67 @@
...
+		}
+#endif /* CONFIG_LIBUKNOFAULT */
+
+		cdmp_gen_print_call_trace_entry(frame[1]);
+
+		/* Goto next frame */
+		fp = frame[0];
+	}
+}
+#endif /* !__OMIT_FRAMEPOINTER__ && CONFIG_LIBUKNOFAULT */
```
mogasergiu: I think you meant `CONFIG_LIBUKDEBUG_CRASH_PRINT_CALL_TRACE && !__OMIT_FRAMEPOINTER__`
(Ditto x86)

lib/ukdebug/arch/x86_64/include/uk/arch/crash.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <uk/essentials.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern __bool _uk_crash_explicit;
+
+#define uk_crash_trigger()						\
```
mogasergiu, line 18: Shouldn't this be `ukarch_crash_trigger`? Same for the header guards, `__UK_ARCH_CRASH_H__`?
(Ditto x86)

michpappas: No, we don't have such a convention

lib/ukdebug/Makefile.uk
michpappas: `crash.c` contains behavior that is not limited to the crash screen and most importantly the implementation of `UK_CRASH()` so it should not be compiled conditionally

plat/kvm/arm/exceptions.S
michpappas: ~~This file contains several `ldr`, so I would rather not change one of them arbitrarily. If it needs to be converted to use `ldr_ur` that is a separate task.~~ **Edit:** Sorry, I didn't understand that the suggestion on `ur_ldr` was about literal pools, this is now fixed.

plat/common/include/x86/traps.h
michpappas: There is a separate check for `rc` right above

lib/ukdebug/Makefile.uk
michpappas: >  (Btw I just noticed but what is the LIBUKDEBUG_CINCLUDES-y  += -I$(LIBUKDEBUG_BASE) above for?)

Yes, that is a leftover from the original patches, thanks for spotting.

mogasergiu (approved):

PR #1540, opened 2024-11-21
lib/vfscore: Support embedded initrd with custom compiled-in fstabs

skuenzer:
### Description of changes

This PR enables compiling images with embedded initrds with "custom" compiled-in mount tables (fstabs). As soon as one of the activated compiled-in mount table references to the device `"embedded"`, the configuration requires a path for an initrd for embedding at compile time. 

Prio this change, an initrd could only be embedded via using pre-configured (and non-"custom") table configuration that depends on an embedded initrd, for example: "Embedded InitRD (CPIO)".


razvand:
@skuenzer , could you share a sample `vfs.fstab` configuration that would work for this update? What didn't work before and what would work now?

razvand:
> @skuenzer , could you share a sample `vfs.fstab` configuration that would work for this update? What didn't work before and what would work now?

My understanding is that you can now embedded multiple initrds? Is that the case? Up to 4 custom mount images, if `embedded` is the device used? In which case, they would have to be mounted in different mount points, right?

skuenzer:
> @skuenzer , could you share a sample `vfs.fstab` configuration that would work for this update? What didn't work before and what would work now?

The thing that the PR fixes is if you use `CONFIG_LIBVFSCORE_AUTOMOUNT_CI_CUSTOM` or `CONFIG_LIBVFSCORE_AUTOMOUNT_FB_CUSTOM` to manually configure the compiled-in fstab entries, you could not make use of an embedded initrd at all. This PR fixes that by enabling the embedded initrd config in the main vfscore menu as soon an embedded initrd was specified in the manual table. The change does not have anything to do with the ‘vfs.fstab‘ kernel parameter.

skuenzer:
> > @skuenzer , could you share a sample `vfs.fstab` configuration that would work for this update? What didn't work before and what would work now?
> 
> 
> 
> My understanding is that you can now embedded multiple initrds? Is that the case? Up to 4 custom mount images, if `embedded` is the device used? In which case, they would have to be mounted in different mount points, right?

No, you can still embed only one initrd to the compiled kernel image.  But principle you could extract it multiple times.

razvand (approved):

PR #1564, opened 2025-01-13
lib/syscall-shim: Do not dereference optional clone() parameters

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The `parent_tid` parameter of `clone()` is used to instruct the kernel where to store the child TID in parent's memory. Similarly, the `child_tid` is used to instruct the kernel where to store the child TID in the child's memory. Both parameters are optional, and are interpreted conditionally to whether the CLONE_CHILD_SETTID and CLONE_PARENT_SETTID flags are set, respectively.
    
Do not interpret these options as PT_REF, as the pointer will not be valid if the caller does not set corresponding flags.

### DIFF
--- a/lib/syscall_shim/uk_prsyscall.c
+++ b/lib/syscall_shim/uk_prsyscall.c
@@ -1220,21 +1220,39 @@ static void pr_syscall(struct uk_streambuf *sb, int fmtf,
 
 #ifdef HAVE_uk_syscall_clone
 	case SYS_clone:
+		do {
+			unsigned long pt_tid_parent_ref;
+			unsigned long pt_tid_child_ref;
+			unsigned long flags;
+
+			flags = (unsigned long)va_arg(args, long);
+
+			if (flags & CLONE_PARENT_SETTID)
+				pt_tid_parent_ref = PT_VADDR | PT_REF;
+			else
+				pt_tid_parent_ref = PT_VADDR;
+
+			if (flags & CLONE_CHILD_SETTID)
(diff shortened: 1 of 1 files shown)


lib/syscall_shim/uk_prsyscall.c
```diff
@@ -1224,16 +1224,16 @@ static void pr_syscall(struct uk_streambuf *sb, int fmtf,
 		VPR_SYSCALL(sb, fmtf, syscall_num, args, rc >= 0,
 			    PT_CLONEFLAGS,
 			    PT_VADDR, /* sp */
-			    PT_TID | PT_REF, /* ref to parent tid */
```
mogasergiu, line 1227: Looks good. But, `pr_param`'s `PT_REF` seems to **supposedly** handle NULL though? Could the bug be somewhere else instead?

michpappas: Sorry, the description is incorrect as I opened the PR long time after I wrote the patch. The actual problem is that if the caller does not pass `CLONE_CHILD_SETTID` / `CLONE_PARENT_SETTID` the pointers will be invalid, so a check for `NULL` does not suffice. While I was writing that though I came up with a better fix that retains the previous behavior if these flags are set. Will push and update the commit message & cover letter.

mogasergiu (approved):

PR #1570, opened 2025-01-27
lib/uksched: Do not save/restore ECTX on coop thread switching

mogasergiu:
### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Since we only have cooperative scheduling at the moment, saving/restoring ECTX when thread switching is not necessary, since the ABI treats these registers as scratch registers and thread switching essentially happens through actual explicit function calls.

On the other hand, this is obviously not the case with preemptive scheduling since thread switching can happen out of the blue, typically on a timer IRQ, so then ECTX saving/restoring is definetely necessary.

But, again, we are only doing cooperative scheduling at the moment and this operation adds unnecessary overhead so throw it out.

### DIFF
--- a/lib/uksched/include/uk/sched_impl.h
+++ b/lib/uksched/include/uk/sched_impl.h
@@ -109,19 +109,26 @@ void uk_sched_thread_switch(struct uk_thread *next)
 	ukplat_per_lcpu_current(__uk_sched_thread_current) = next;
 
 	prev->tlsp = ukplat_tlsp_get();
-	if (prev->ectx)
-		ukarch_ectx_store(prev->ectx);
 
 	/* Load next TLS and extended registers before context switch.
 	 * This avoids requiring special initialization code for newly
 	 * created threads to do the loading.
 	 */
 	ukplat_tlsp_set(next->tlsp);
-	if (next->ectx)
-		ukarch_ectx_load(next->ectx);
 
 	ukplat_lcpu_set_auxsp(next->auxsp);
 
+	/**
+	 * NOTE: There is also prev->ectx/next->ectx! Since we only have
+	 * cooperative scheduling at the moment, saving/restoring ECTX
+	 * when thread switching is not necessary, since the ABI treats these
+	 * registers as scratch registers and thread switching essentially
+	 * happens through actual explicit function calls.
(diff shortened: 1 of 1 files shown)


andreittr (commented):
The optimization evidently makes sense for a co-op scheduler, although I don't like the fact that `uksched` takes on responsibility for policy decisions (something that should be exclusively handled by scheduler _implementations_ not APIs). If/when a scheduler impl comes around that needs ectx save/restore, we rely on a `NOTE` comment to change behavior in the API code -- something pretty brittle IMO.

How about instead we have `uksched` explicitly expose two methods for switching threads (+ clear docstrings for both):
- `uk_sched_thread_switch_intr` -- the current implementation that saves/restores ectx (and generally assumes thread execution can be arbitrarily interrupted)
- `uk_sched_thread_switch_yield` -- your new optimized impl that skips ectx (and generally assumes threads yield execution on function call ABI semantics)

and have our sched implementations (`ukschedcoop` only atm) call one or the other depending on their guarantees (e.g., coop scheds call `switch_yield`, pre-emptive scheds call `switch_intr`, neither needs to care what an `ectx` is -- neat).

This way we keep `uksched` cleanly free of caring about what _exact_ scheduler we use, better in line with its design.

mogasergiu:
This sounds like a good idea, however this is not enough for an `_intr` variant. As you can see, it also uses per-lcpu semantics, which is wrong to do in interrupt context as you don't know if that is accessible. I have a PR #1348 that will help us do that, but as of now, the ectx part is not enough to differentiate between the two. I would like to have this differentiation as well but I am afraid that if we were to do so now so early before we understand what preemption on our side entails, we will just end up adding a function no-one will use and that will likely need to be rewritten anyway. I also think a comment should do it for now. It is true it is brittle, but I doubt there will be any significant changes in this direction without bumping into a comment that is the size of the function itself.

andreittr (approved):
Ok, that's a fair point, we can revisit this issue when expanding our schedulers.

michpappas (approved):

michpappas:

andreittr:

PR #1575, opened 2025-02-04
lib/posix-fdio: Add internal implementations of futime(n)s

andreittr:
### Description of changes

This changeset adds internal API (`uk_sys_*`) implementations of the `futimes` and `futimens` operations on file references, as well as exposing the statx-to-stat field conversions used by `posix-fdio` for use in related code (specifically, `*at`-type operations that can target either open files or VFS paths).

(Binary) syscalls fall outside the scope of this PR, due to how Linux ties both of these operations to VFS-specific syscalls: `futimesat` and `utimensat`, respectively. Since these target the VFS they must remain in `vfscore` for now.

Permissions checking is also out of scope, it is to be handled File/VFS-wide by a single later PR.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/posix-fdio/fdstat.c
+++ b/lib/posix-fdio/fdstat.c
@@ -9,8 +9,15 @@
 #include <string.h>
 
 #include <uk/posix-fdio.h>
+#include <uk/posix-time.h>
+#include <uk/timeutil.h>
 
 #include "fdio-impl.h"
+
+#define statx_time(ts) ((struct uk_statx_timestamp){ \
+	.tv_sec = (ts).tv_sec, \
+	.tv_nsec = (ts).tv_nsec \
+})
 
 static inline
 dev_t nums2dev(__u32 major, __u32 minor)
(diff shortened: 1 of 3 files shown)


lib/posix-fdio/fdstat.c
```diff
@@ -117,3 +124,52 @@ int uk_sys_fchown(struct uk_ofile *of, uid_t owner, gid_t group)
...
+
+	if (iolock)
+		uk_file_wlock(of->file);
+	r = uk_file_setstat(of->file, mask, &sx);
+	if (iolock)
+		uk_file_wunlock(of->file);
+	return r;
+}
+
+int uk_sys_futimes(struct uk_ofile *of, const struct timeval *tv)
```
mogasergiu, line 183: Random question, not necessarily related to this. But I remember we had a couple of nested syscalls in the CPIO/RAMFS code, caused by utimes (the name reminded me of this :D). Wouldn't it be useful for these to be exportable for use in such scenarios? Unless you have a better plan to address that stuff.

lib/posix-fdio/fdstat.c
```diff
@@ -117,3 +124,52 @@ int uk_sys_fchown(struct uk_ofile *of, uid_t owner, gid_t group)
...
+	unsigned int mask;
+	struct uk_statx sx;
+	struct timespec t;
+	const int iolock = _SHOULD_LOCK(of->mode);
+
+	/* TODO: check if uid is root or file owner for custom times */
+
+	if (!times ||
+	    times[0].tv_nsec == UTIME_NOW || times[1].tv_nsec == UTIME_NOW)
+		uk_sys_clock_gettime(CLOCK_REALTIME, &t);
```
mogasergiu: Btw, shouldn't we also check the return value of this call?

andreittr: :thinking: You're right, even though futime(n)s manpages don't say anything about erroring out on an unavailable realtime clock (probably due to how Linux handles it internally), it is an error case we must handle as per `uk_sys_clock_gettime` API (regardless if it can ever fail in practice). Great catch! Will fix.

lib/posix-fdio/fdstat.c
andreittr: Using these functions instead of syscalls is exactly what they are meant for, and indeed they are exported.
`posix-fd` has no `exportsyms.uk` file on purpose (it is compartmentalized so as to not do private calls between its compilation units), and so all non-static symbols are exported by default (just like regular C linkage under *nix). I like writing libs like this when I can, it makes a function's public/private status obvious from the source file (I also understand the need for `exportsyms.uk` in more complex libs too).

A `ukfile`-based ramfs will most surely use these functions instead of syscalls, yes.

Unfortunately, these cannot be used by `ukcpio` as long as `vfscore` is around, as these only work on `ukfile`s; vfscore has its own way of handling `(f)utime(n)s`, and shimming between these two can only be done at the syscall level.

mogasergiu: I see so I guess this will be addressed once vfscore is retired then?

lib/posix-fdio/fdstat.c
```diff
@@ -117,3 +124,63 @@ int uk_sys_fchown(struct uk_ofile *of, uid_t owner, gid_t group)
...
+	struct uk_statx sx;
+	struct timespec t;
+	const int iolock = _SHOULD_LOCK(of->mode);
+
+	/* TODO: check if uid is root or file owner for custom times */
+
+	if (!times ||
+	    times[0].tv_nsec == UTIME_NOW || times[1].tv_nsec == UTIME_NOW) {
+		r = uk_sys_clock_gettime(CLOCK_REALTIME, &t);
+		UK_ASSERT(r <= 0);
```
mogasergiu: Why do we also assert here? :thinking:

andreittr: In short, it's to ensure we follow our API and only return 0 or negative values, regardless of whether the `clock_gettime` implementation follows its documented API (the _documentation_ says it returns `<= 0`, but locally the compiler can only reason it's an `int` and doesn't protect us from a misbehaving impl -- we have to explicitly assert).

That said, we can shave a branch off the success path by moving the assert into the error case. Will push update.

mogasergiu (approved):
Thanks. Merging.

Issue #1505, opened 2024-09-28
`lib/ukallocbuddy`: Boot error message for Nginx (QEMU, KVM, x86_64)


razvand:
### Describe the bug

Booting an Nginx Unikraft image on QEMU KVM fox x86_64 gives a boot error message (booting works, it's just a message).

### Steps to reproduce

Build Nginx in the [`catalog-core` repository](https://github.com/unikraft/catalog-core). Follow the steps:

1. Clone the repository:

   ```console
   git clone https://github.com/unikraft/catalog-core
   ```

1. Enter the `nginx/` directory:

   ```console
   cd catalog-core/nginx/
   ```
   
1. Switch to the `scripts` branch:

   ```console
   git checkout scripts
   ```

1. Build Nginx for QEMU/x86_64:

   ```console
   ./build.qemu.x86_64
   ```

1. Run Nginx:

   ```console
   ./run.qemu.x86_64
   ```

   The error message appears. It's just a message. The unikernel still boots.

To close the QEMU virtual machine, use the `Ctrl+a x` keyboard shortcut;
that is, press the `Ctrl` and `a` keys at the same time and then, separately, press the `x` key.

### Expected behavior

No error messages appears at boot.

### Which architectures were you using or does this bug affect?

x86_64

### Which platforms were you using or does this bug affect?

kvm

### Relevant log output

```shell
ERR:  [libukallocbbuddy] <bbuddy.c @  489> 11000: Failed to add memory region 322000-323000: Not enough space after applying page alignments
```

ThestralWarrior:
hi, can i work on this issue?

ryuusama09:
@razvand 
SeaBIOS (version 1.16.3-debian-1.16.3-2)


iPXE (http://ipxe.org) 00:03.0 CA00 PCI2.10 PnP PMM+0078EE00+006CEE00 CA00
                                                                               


Booting from ROM..Powered by
o.   .o       _ _               __ _
Oo   Oo  ___ (_) | __ __  __ _ ' _) :_
oO   oO ' _ `| | |/ /  _)' _` | |_|  _)
oOo oOO| | | | |   (| | | (_) |  _) :_
 OoOoO ._, ._:_:_,\_._,  .__,_:_, \___)
                Calypso 0.17.0~ba5e036c
Hello from Unikraft!

this is the output (and the kernel closes ) i am getting if i am replicating the steps above . am I missing anything ?

razvand:
> hi, can i work on this issue?

[...]

PR #1576, opened 2025-02-05
lib/ukalloc: Make palloc/pfree page-granular (+ document main public `ukalloc` API)

andreittr:
### Description of changes

Previously bbuddy's palloc/pfree implementations behaved like a page-order allocator, with the assumption of dealing only in power-of-2 numbers of pages instead of the page-granular API _implied_ by `ukalloc`. This mismatch between expected/assumed and actual behavior would lead to memory corruption and unpredictable crashes if bbuddy was called without following its undocumented and unasserted assumptions.

Fortunately, I have found no code in core Unikraft breaks these assumptions in a way to cause visible bugs ( :crossed_fingers: ), but this is (1) a ticking timebomb, and (2) severely limiting developent of code that uses page allocation more seriously.

The issue of the `ukalloc` API being undocumented was a leading cause of this situation, and so this PR starts off with documenting the core "public-facing" functions of `ukalloc`:
- object allocation (`malloc()` & friends, `free()`)
- page allocation (`palloc()` & `pfree()`)

Next, we need to consider that true page allocation semantics cannot be implemented on top of `malloc`&co (most notably due to page allocs being partially freeable), and thus the compat functions `uk_palloc_compat` and `uk_pfree_compat` are not correct and would break non-trivial page-allocating code if used.
Thus, these two functions are removed, and any malloc-only allocator will now _expliciitly_ signal that it cannot provide `palloc/pfree`, with the caller deciding whether `memalign` provides a satisfactory fallback or if an error is warranted.

The only users of palloc/pfree in the core is xen platform/driver code. I have checked these callsites along with @mogasergiu and they all would seem to work fine with memalign too (on xen at least). I've added explicit fallback logic to all of them.

Finally, this PR fixes bbuddy's palloc/pfree implementations to be truly page-granular, fully conforming to the now-documented ukalloc API.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukallocbbuddy/bbuddy.c
+++ b/lib/ukallocbbuddy/bbuddy.c
@@ -5,6 +5,7 @@
  * (C) 2003 - Rolf Neugebauer - Intel Research Cambridge
  * (C) 2005 - Grzegorz Milos - Intel Research Cambridge
  * (C) 2017 - Simon Kuenzer - NEC Europe Ltd.
+ * (C) 2025 - Unikraft GmbH and The Unikraft Authors
  ****************************************************************************
  *
  *        File: mm.c
@@ -12,8 +13,10 @@
  *     Changes: Grzegorz Milos
  *     Changes: Simon Kuenzer <simon.kuenzer@neclab.eu>
  *     Changes: Nour-eddine Taleb <contact@noureddine.xyz>
+ *     Changes: Andrei Tatar <ttr@unikraft.io>
  *
- *        Date: Aug 2003, changes Aug 2005, changes Oct 2017, changes Dec 2022
+ *        Date: Aug 2003, changes Aug 2005, changes Oct 2017, changes Dec 2022,
+ *              changes Feb/Jul 2025
  *
  * Environment: Unikraft
  * Description: buddy page allocator from Xen.
@@ -49,24 +52,20 @@
 #include <uk/bitops/bitscan.h>
 #include <uk/print.h>
 #include <uk/assert.h>
+#include <uk/list.h>
 #include <uk/page.h>
 
-typedef struct chunk_head_st chunk_head_t;
-typedef struct chunk_tail_st chunk_tail_t;
-
-struct chunk_head_st {
-	chunk_head_t *next;
-	chunk_head_t **pprev;
-	unsigned int level;
+struct chunk_head {
+	struct uk_hlist_node link;
+	unsigned int page_order;
 };
 
-struct chunk_tail_st {
-	unsigned int level;
-};
+#define CHUNK_END(buf, ord) \
+	((void *)(((char *)(buf)) + (1ULL << ((ord) + __PAGE_SHIFT))))
 
 /* Linked lists of free chunks of different powers-of-two in size. */
 #define FREELIST_SIZE ((sizeof(void *) << 3) - __PAGE_SHIFT)
-#define FREELIST_EMPTY(_l) ((_l)->next == NULL)
+#define FREELIST_EMPTY(_l) uk_hlist_empty((_l))
 #define FREELIST_ALIGNED(ptr, lvl) \
 	!((uintptr_t)(ptr) & ((1ULL << ((lvl) + __PAGE_SHIFT)) - 1))
 
@@ -81,8 +80,7 @@ struct uk_bbpalloc_memr {
 
 struct uk_bbpalloc {
 	unsigned long nr_free_pages;
-	chunk_head_t *free_head[FREELIST_SIZE];
-	chunk_head_t free_tail[FREELIST_SIZE];
+	struct uk_hlist_head free_head[FREELIST_SIZE];
 	struct uk_bbpalloc_memr *memr_head;
 };
 
@@ -99,42 +97,36 @@ struct uk_bbpalloc {
 
 #define _FREESAN_LOCFMT "\t@ %p (free_head[%zu](%p) + %zu): "
 
-#define _FREESAN_HEAD(head) \
-do { \
-	size_t off = 0; \
-	for (chunk_head_t *c = head; c; c = c->next, off++) { \
-		if (c->next != NULL && c->level != i) { \
-			uk_pr_err("Bad page level" _FREESAN_LOCFMT \
-			          "got %u, expected %zu\n", \
-			          c, i, head, off, c->level, i); \
-		} \
-		if (_FREESAN_BAD_CHUNKPTR(c->pprev)) { \
-			uk_pr_err("Bad pprev pointer" _FREESAN_LOCFMT "%p\n", \
-			          c, i, head, off, c->pprev); \
-		} else if (*c->pprev != c) { \
-			uk_pr_err("Bad backward link" _FREESAN_LOCFMT \
-			          "got %p, expected %p\n", \
-			          c, i, head, off, *c->pprev, c); \
-		} \
-		if (_FREESAN_BAD_CHUNKPTR(c->next)) { \
-			uk_pr_err("Bad next pointer" _FREESAN_LOCFMT "%p\n", \
-			          c, i, head, off, c->next); \
-			break; \
-		} else if (!FREELIST_ALIGNED(c->next, i) && \
-		           c->next->next != NULL) { \
-			uk_pr_err("Unaligned next page" _FREESAN_LOCFMT \
-			          "%p not aligned to %zx boundary\n", \
-			          c, i, head, off, c->next, \
-			          (size_t)1 << (__PAGE_SHIFT + i)); \
-		} \
-	} \
+#define _FREESAN_HEAD(head, ord)					\
+do {									\
+	size_t off = 0;							\
+	struct chunk_head *c;						\
+									\
+	uk_hlist_for_each_entry(c, (head), link) {			\
+		if (_FREESAN_BAD_CHUNKPTR(c)) {				\
+			uk_pr_err("Invalid chunk pointer" _FREESAN_LOCFMT "\n",\
+				  c, (ord), (head), off);		\
+			break;						\
+		}							\
+		if (!FREELIST_ALIGNED(c, (ord)))			\
+			uk_pr_err("Unaligned chunk" _FREESAN_LOCFMT	\
+				  "%p not aligned to %llx boundary\n",	\
+				  c, (ord), (head), off, c, BBUDDY_LEN((ord)));\
+		if (c->page_order != (ord))				\
+			uk_pr_err("Bad page level" _FREESAN_LOCFMT	\
+				  "got %u, expected %zu\n",		\
+				  c, (ord), (head), off, c->page_order, (ord));\
+		if (c->link.pprev && (*c->link.pprev)->next != &c->link) \
+			uk_pr_err("Bad backlink" _FREESAN_LOCFMT	\
+				  "got %p, expected %p\n",		\
+				  c, (ord), (head), off,		\
+				  (*c->link.pprev)->next, &c->link);	\
+	}								\
 } while (0)
 
-#define freelist_sanitycheck(free_head) \
-for (size_t i = 0; i < FREELIST_SIZE; i++) { \
-	UK_ASSERT((free_head)[i] != NULL); \
-	_FREESAN_HEAD((free_head)[i]); \
-}
+#define freelist_sanitycheck(free_head)					\
+for (size_t i = 0; i < FREELIST_SIZE; i++)				\
+	_FREESAN_HEAD(&(free_head)[i], i)
 
 #else /* !CONFIG_LIBUKALLOCBBUDDY_FREELIST_SANITY */
 
@@ -274,12 +266,25 @@ static void map_free(struct uk_bbpalloc *b, uintptr_t first_page,
 	b->nr_free_pages += nr_pages;
 }
 
+/* Number of pages covered by page order `ord` */
+#define BBUDDY_PAGES(ord) (1ULL << (ord))
+
+/* Bit corresponding to page order `ord`; identical to size of chunk at `ord` */
+#define BBUDDY_LEN(ord) (BBUDDY_PAGES((ord)) << __PAGE_SHIFT)
+
+/* Address of buddy page of `ptr` at page order `ord` */
+#define BBUDDY_BUDDY_ADDR(ptr, ord) \
+	((void *)(((uintptr_t)(ptr)) ^ BBUDDY_LEN(ord)))
+
+/* Non-zero if `ptr` would be the head buddy in page order `ord` */
+#define BBUDDY_ISHEAD(ptr, ord) (!(((uintptr_t)(ptr)) & BBUDDY_LEN(ord)))
+
 /* return log of the next power of two of passed number */
 static inline unsigned long num_pages_to_order(unsigned long num_pages)
 {
 	UK_ASSERT(num_pages != 0);
 
-	/* uk_flsl has undefined behavior when called with zero */
+	/* uk_mssbl has undefined behavior when called with zero */
 	if (num_pages == 1)
 		return 0;
 
@@ -291,135 +296,210 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
 	return uk_mssbl(num_pages - 1) + 1;
 }
 
+/* return the highest page order that `ptr` is aligned to, up to maxord */
+static inline unsigned long ptr_order(void *ptr, unsigned long maxord)
+{
+	unsigned long v = (uintptr_t)ptr;
+
+	UK_ASSERT(IS_ALIGNED(v, __PAGE_SIZE));
+	v >>= __PAGE_SHIFT;
+	if (!v)
+		return maxord; /* page 0 is aligned to any order */
+	return MIN(uk_lssbl(v), maxord);
+}
+
+/* return the highest power of two less than or equal to num_pages */
+static inline unsigned long npages_order(unsigned long num_pages)
+{
+	UK_ASSERT(num_pages);
+	return uk_mssbl(num_pages);
+}
+
 /*********************
  * BINARY BUDDY PAGE ALLOCATOR
  */
+
+/**
+ * INTERNAL. Trim chunk `ch` of page order `ord` down to `num_pages` pages.
+ *
+ * Page-order-sized chunks are trimmed off the end of `ch` and placed back into
+ * their respective freelists as needed.
+ */
+static inline void bbuddy_trim(struct uk_bbpalloc *b, struct chunk_head *ch,
+			       size_t ord, unsigned long num_pages)
+{
+	size_t spare_pages;
+	size_t spare_ord;
+	size_t spare_len;
+	char *spare_end;
+	struct chunk_head *spare_ch;
+
+	UK_ASSERT(ch->page_order == ord);
+
+	spare_pages = BBUDDY_PAGES(ord) - num_pages;
+	spare_end = CHUNK_END(ch, ord);
+	while (spare_pages) {
+		spare_ord = npages_order(spare_pages);
+		spare_len = BBUDDY_LEN(spare_ord);
+		spare_ch = (struct chunk_head *)(spare_end - spare_len);
+
+		UK_ASSERT(spare_ord < FREELIST_SIZE);
+		UK_ASSERT(BBUDDY_PAGES(spare_ord) <= spare_pages);
+
+		/* Populate chunk & link in */
+		spare_ch->page_order = spare_ord;
+		uk_hlist_add_head(&spare_ch->link, &b->free_head[spare_ord]);
+
+		spare_pages -= BBUDDY_PAGES(spare_ord);
+		spare_end -= BBUDDY_LEN(spare_ord);
+	}
+}
+
 static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
 {
-	struct uk_bbpalloc *b;
-	size_t i;
-	chunk_head_t *alloc_ch, *spare_ch;
-	chunk_tail_t *spare_ct;
+	struct uk_bbpalloc *const b = (struct uk_bbpalloc *)&a->priv;
+	struct chunk_head *alloc_ch;
+	size_t ord;
 
-	UK_ASSERT(a != NULL);
-	b = (struct uk_bbpalloc *)&a->priv;
-
+	UK_ASSERT(a);
+	UK_ASSERT(num_pages);
 	freelist_sanitycheck(b->free_head);
 
-	size_t order = (size_t)num_pages_to_order(num_pages);
+	ord = num_pages_to_order(num_pages);
+	/* Find the smallest order of free memory that satisfies the request */
+	while (ord < FREELIST_SIZE && FREELIST_EMPTY(&b->free_head[ord]))
+		ord++;
+	/* We use >= as ord may have been set arbitrarily high by num_pages */
+	if (ord >= FREELIST_SIZE)
+		goto err_nomem;
 
-	/* Find smallest order which can satisfy the request. */
-	for (i = order; i < FREELIST_SIZE; i++) {
-		if (!FREELIST_EMPTY(b->free_head[i]))
-			break;
-	}
-	if (i >= FREELIST_SIZE)
-		goto no_memory;
+	/* Grab & unlink a chunk */
+	alloc_ch = uk_hlist_entry(b->free_head[ord].first, struct chunk_head,
+				  link);
+	uk_hlist_del(&alloc_ch->link);
+	UK_ASSERT(FREELIST_ALIGNED(alloc_ch, ord));
 
-	/* Unlink a chunk. */
-	alloc_ch = b->free_head[i];
-	b->free_head[i] = alloc_ch->next;
-	alloc_ch->next->pprev = alloc_ch->pprev;
+	/* Trim off any extra pages off the end */
+	bbuddy_trim(b, alloc_ch, ord, num_pages);
 
-	/* We may have to break the chunk a number of times. */
-	while (i != order) {
-		/* Split into two equal parts. */
-		i--;
-		spare_ch = (chunk_head_t *)((char *)alloc_ch
-					    + (1UL << (i + __PAGE_SHIFT)));
-		spare_ct = (chunk_tail_t *)((char *)spare_ch
-					    + (1UL << (i + __PAGE_SHIFT))) - 1;
-
-		/* Create new header for spare chunk. */
-		spare_ch->level = i;
-		spare_ch->next = b->free_head[i];
-		spare_ch->pprev = &b->free_head[i];
-		spare_ct->level = i;
-
-		/* Link in the spare chunk. */
-		spare_ch->next->pprev = &spare_ch->next;
-		b->free_head[i] = spare_ch;
-	}
-	UK_ASSERT(FREELIST_ALIGNED(alloc_ch, order));
-	map_alloc(b, (uintptr_t)alloc_ch, 1UL << order);
+	/* Mark as in use and return */
+	map_alloc(b, (uintptr_t)alloc_ch, num_pages);
 
 	uk_alloc_stats_count_palloc(a, (void *) alloc_ch, num_pages);
 	freelist_sanitycheck(b->free_head);
-	return ((void *)alloc_ch);
 
-no_memory:
-	uk_pr_warn("%"__PRIuptr": Cannot handle palloc request of order %"__PRIsz": Out of memory\n",
-		   (uintptr_t)a, order);
+	return (void *)alloc_ch;
+
+err_nomem:
+	uk_pr_warn("%p: Cannot handle palloc request of %lu: Out of memory\n",
+		   a, num_pages);
 
 	uk_alloc_stats_count_penomem(a, num_pages);
 	errno = ENOMEM;
 	return NULL;
 }
 
+/**
+ * INTERNAL. Maximally merge chunk `*chp` of order `ord` with any free buddies,
+ * updating its value along the way.
+ *
+ * `*chp` must be marked as "allocated" in bbuddy's bookkeeping.
+ *
+ * The merge process for a given to-be-freed chunk is:
+ * 1. calculate ch's buddy's starting page address using BBUDDY_BUDDY_ADDR
+ * 2. check in the bitmap whether that page is allocated or free
+ *   - if allocated, we know buddy cannot be free, so we stop
+ *   - if free, we know it has to be a chunk head (see below), and thus has
+ *     valid metadata
+ * 3. read chunk_head from starting page to determine buddy's actual size/order
+ *   - if it matches our own, we can merge and move onto a higher order
+ *   - if it differs, we stop
+ *
+ * Generally speaking, a page marked free in the bitmap is a chunk head of order
+ * ord if-and-only-if:
+ * (a) the page is aligned to that particular order (all bits < ord are 0), AND
+ * (b) the page is not part of a chunk of larger order
+ *
+ * In the particular care of bbuddy_merge(ch):
+ * (a) is guaranteed because ch is aligned to ord, and thus its candidate buddy
+ * must necessarily be aligned too.
+ * (b) is guaranteed to never happen, as buddy would need ch to already be free
+ * in order to form a larger order chunk
+ *
+ * @return Order of `*ch` after merging
+ */
+static inline size_t bbuddy_merge(struct uk_bbpalloc *b,
+				  struct chunk_head **chp, size_t ord)
+{
+	struct chunk_head *ch = *chp;
+	struct chunk_head *buddy;
+
+	/* After successfully merging with our first buddy, we may be able to
+	 * merge again with a higher-order buddy, and so on, up to max order.
+	 */
+	while (ord < FREELIST_SIZE - 1) {
+		/* It is safe to deref this only if marked free in bitmap */
+		buddy = BBUDDY_BUDDY_ADDR(ch, ord);
+
+		/* Stop if buddy is not free or of wrong order */
+		if (allocated_in_map(b, (uintptr_t)buddy) ||
+		    buddy->page_order != ord)
+			break;
+
+		if (BBUDDY_ISHEAD(buddy, ord))
+			/* buddy is predecessor; merge downwards */
+			ch = buddy;
+		/* else: ch is already correct base, increase ord to merge up */
+
+		/* Unlink buddy from freelist & continue to next order */
+		uk_hlist_del(&buddy->link);
+		ord++;
+	}
+	*chp = ch;
+	return ord;
+}
+
 static void bbuddy_pfree(struct uk_alloc *a, void *obj, unsigned long num_pages)
 {
-	struct uk_bbpalloc *b;
-	chunk_head_t *freed_ch, *to_merge_ch;
-	chunk_tail_t *freed_ct;
-	unsigned long mask;
+	struct uk_bbpalloc *const b = (struct uk_bbpalloc *)&a->priv;
+	char *base = obj;
+	size_t freed_pages;
+	size_t ord;
+	struct chunk_head *ch;
 
-	UK_ASSERT(a != NULL);
+	UK_ASSERT(a);
+	UK_ASSERT(obj);
+	UK_ASSERT(num_pages);
+	UK_ASSERT(IS_ALIGNED((uintptr_t)obj, __PAGE_SIZE));
 
 	uk_alloc_stats_count_pfree(a, obj, num_pages);
-	b = (struct uk_bbpalloc *)&a->priv;
-
 	freelist_sanitycheck(b->free_head);
 
-	size_t order = (size_t)num_pages_to_order(num_pages);
+	/* Since obj can span an arbitrary page range, we may need to return it
+	 * in multiple page-order-sized chunks. At each iteration we pick the
+	 * largest chunk size permitted by alignment, attempt to merge it with
+	 * any free buddies, then link it into the appropriate freelist.
+	 */
+	do {
+		ord = ptr_order(base, npages_order(num_pages));
+		freed_pages = BBUDDY_PAGES(ord);
+		ch = (struct chunk_head *)base;
 
-	/* if the object is not page aligned it was clearly not from us */
-	UK_ASSERT((((uintptr_t)obj) & (__PAGE_SIZE - 1)) == 0);
+		UK_ASSERT(ord < FREELIST_SIZE);
+		UK_ASSERT(freed_pages <= num_pages);
 
-	/* First free the chunk */
-	map_free(b, (uintptr_t)obj, 1UL << order);
+		ord = bbuddy_merge(b, &ch, ord);
 
-	/* Create free chunk */
-	freed_ch = (chunk_head_t *)obj;
-	freed_ct = (chunk_tail_t *)((char *)obj
-				    + (1UL << (order + __PAGE_SHIFT))) - 1;
+		/* Populate chunk (must be done before marking free) */
+		ch->page_order = ord;
+		/* Mark pages as free (must be done before linking in) */
+		map_free(b, (uintptr_t)base, freed_pages);
+		/* Link into freelist */
+		uk_hlist_add_head(&ch->link, &b->free_head[ord]);
 
-	/* Now, possibly we can conseal chunks together */
-	while (order < FREELIST_SIZE) {
-		mask = 1UL << (order + __PAGE_SHIFT);
-		if ((unsigned long)freed_ch & mask) {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch - mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
-				break;
-
-			/* Merge with predecessor */
-			freed_ch = to_merge_ch;
-		} else {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch + mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
-				break;
-
-			/* Merge with successor */
-			freed_ct =
-			    (chunk_tail_t *)((char *)to_merge_ch + mask) - 1;
-		}
-
-		/* We are commited to merging, unlink the chunk */
-		*(to_merge_ch->pprev) = to_merge_ch->next;
-		to_merge_ch->next->pprev = to_merge_ch->pprev;
-
-		order++;
-	}
-
-	/* Link the new chunk */
-	freed_ch->level = order;
-	freed_ch->next = b->free_head[order];
-	freed_ch->pprev = &b->free_head[order];
-	freed_ct->level = order;
-
-	freed_ch->next->pprev = &freed_ch->next;
-	b->free_head[order] = freed_ch;
+		num_pages -= freed_pages;
+		base += freed_pages * __PAGE_SIZE;
+	} while (num_pages);
 
 	freelist_sanitycheck(b->free_head);
 }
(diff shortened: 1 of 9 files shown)


lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
 	return uk_flsl(num_pages - 1) + 1;
 }
 
+/* return the highest page order that `ptr` is aligned to, up to maxord */
+static inline unsigned long ptr_order(void *ptr, unsigned long maxord)
+{
+	unsigned long v = (uintptr_t)ptr;
+
+	UK_ASSERT(IS_ALIGNED(v, __PAGE_SIZE));
```
mogasergiu, line 304: Maybe `PAGE_ALIGNED` would have worked better here?
But then again, in the future it will probably be moved to a paging library. Even though you don't need to enable a library to use its headers, it would like an added dependency. Though I'd mention this. Maybe it's best left as is here with `IS_ALIGNED` indeed.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
 	freelist_sanitycheck(b->free_head);
-	return ((void *)alloc_ch);
 
-no_memory:
-	uk_pr_warn("%"__PRIuptr": Cannot handle palloc request of order %"__PRIsz": Out of memory\n",
-		   (uintptr_t)a, order);
+	return (void *)alloc_ch;
+
+err_nomem:
+	uk_pr_warn("%p: Cannot handle palloc request of %lu: Out of memory\n",
```
mogasergiu, line 394: Interesting, printed as a warning even though on error path :thinking:. Oh well this was like this even before the PR anyway...

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-		spare_ch->pprev = &b->free_head[i];
-		spare_ct->level = i;
-
-		/* Link in the spare chunk. */
+	UK_ASSERT(FREELIST_ALIGNED(alloc_ch, ord));
+
+	/* Trim off any extra pages */
+	slack = (1ULL << ord) - num_pages;
+	buf = CHUNK_END(alloc_ch, ord);
+	while (slack) {
```
mogasergiu: Since you are even declaring subscope variables and it's quite lengthy, this looks like it could go into its own separate (inline) function. Something that would take `buf` and `slack` as arguments maybe?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -358,68 +401,76 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
...
-	/* Now, possibly we can conseal chunks together */
-	while (order < FREELIST_SIZE) {
-		mask = 1UL << (order + __PAGE_SHIFT);
-		if ((unsigned long)freed_ch & mask) {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch - mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
+	/* Might need to return the buffer in multiple chunks */
+	do {
+		/* release `np` pages starting at `buf` back into freelists */
```
mogasergiu: Ditto about moving to a function the inner `while` + I am not a fan of subscope variable declaration unless we are talking macros (IMO, generally speaking, if you needed subscope declarations, there is a chance you could have split that part in another function) but is there any reason you are declaring them inside the `do { } while` instead of at the beginning? They are going to end up existing anyway. I won't insist on this though, I guess it could also be personal taste :upside_down_face: 

So then maybe we could split out that inner while loop into a static inline separate function that declares and manipulates freed_ch/ct. Something that receives as arguments `bp`, `buf` and `num_pages` and assumes `num_pages` is indeed a power of 2?
That could then be used by other functions as well maybe?

Maybe I am understanding the code wrong though? I see the inner loop handling power of two num_pages and the while loop splitting the maybe not power of two num_pages to accomodate for the inner loop. It looks like it could be two functions if I am not mistaken.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	slack = (1ULL << ord) - num_pages;
+	buf = CHUNK_END(alloc_ch, ord);
+	while (slack) {
+		const size_t sord = npages_order(slack);
+		const size_t spare_len = BBUDDY_MASK(sord);
+		chunk_head_t *spare_ch;
+		chunk_tail_t *spare_ct;
+
+		UK_ASSERT(spare_len <= slack * __PAGE_SIZE);
+		UK_ASSERT(1ULL << sord <= slack);
```
mogasergiu: Maybe I am exaggerating, but I am getting anxious whenever I don't see paranthesis around bitwise ops in situations like these :sweat_smile:. Can we add them: `UK_ASSERT((1ULL << sord) <= slack);`

mogasergiu (commented):
Had a first look. Looks ok to me. Though I left some comments because I think you already know from past experience that I tend to not like subscope declarations as much :sweat_smile:. Alas, feel free to consider them as suggestions, I will not insist on them.

I will have another look tomorrow with a fresh mind and likely give my RB. Cheers :beers:!

lib/ukallocbbuddy/bbuddy.c
andreittr: We're in the context of just one allocator here (not even necessarily the system alloc); we can't know if an OOM condition here is trivial or catastrophic, and it's best to leave it to the caller to signal an error if warranted. IMHO even a warning is overkill, I'd've made it info at most, but I want to keep previous observable behavior similar.

mogasergiu: Yes, true. I am not requesting a change. It's just funny that it seems to be something [recurring](https://github.com/unikraft/unikraft/issues/1505#issuecomment-2553487095) with bbuddy :sweat_smile:

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	UK_ASSERT(num_pages);
 	freelist_sanitycheck(b->free_head);
 
-	size_t order = (size_t)num_pages_to_order(num_pages);
+	ord = num_pages_to_order(num_pages);
+	while (ord < FREELIST_SIZE && FREELIST_EMPTY(b->free_head[ord]))
+		ord++;
+	/* We use >= as ord may have been set arbitrarily high by num_pages */
+	if (ord >= FREELIST_SIZE)
+		goto err_nomem;
```
michpappas, line 374: Preserving the comment wouldn't hurt here.

```diff
+	/* Find the smallest order of free memory that satisfies the request */
	while (ord < FREELIST_SIZE && FREELIST_EMPTY(b->free_head[ord]))
```

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-		spare_ct->level = i;
-
-		/* Link in the spare chunk. */
+	UK_ASSERT(FREELIST_ALIGNED(alloc_ch, ord));
+
+	/* Trim off any extra pages */
+	slack = (1ULL << ord) - num_pages;
+	buf = CHUNK_END(alloc_ch, ord);
+	while (slack) {
+		const size_t sord = npages_order(slack);
```
michpappas: Nit: For consistency with the other vars you might as well rename `sord` to `spare_ord`. (also what is a `sord`  ⚔️)

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,65 +308,91 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-			break;
-	}
-	if (i >= FREELIST_SIZE)
-		goto no_memory;
-
-	/* Unlink a chunk. */
-	alloc_ch = b->free_head[i];
-	b->free_head[i] = alloc_ch->next;
+	/* Grab & unlink a chunk */
+	alloc_ch = b->free_head[ord];
```
michpappas: For the sake of correctness and ease of review, both link and unlink would benefit from becoming inline functions or macros. We can also leave for a future iteration though, as it's not the only thing that could use a revamp.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -358,68 +401,76 @@ static void *bbuddy_palloc(struct uk_alloc *a, unsigned long num_pages)
...
+		const size_t np = 1UL << ord;
+		chunk_head_t *freed_ch = (chunk_head_t *)buf;
+		chunk_tail_t *freed_ct = CHUNK_TAIL_PTR(buf, ord);
+
+		UK_ASSERT(ord < FREELIST_SIZE);
+		UK_ASSERT(np <= num_pages);
+
+		/* Try merging with existing free chunks */
+		while (ord < FREELIST_SIZE - 1) {
+			chunk_head_t *const targ = BBUDDY_BUDDY(freed_ch, ord);
```
michpappas: Another nit: Maybe it's just me, but `targ` didn't make any sense (`t<something> arg`?) until I you said it stands for `target` (and now it seems obvious of course). As this implementation is complex enough already, let's use clear variable names. Consider whether `buddy` would be more a more suitable name, as you use that in the documentation above, or at least whether it's worth renaming to `target`.

lib/ukallocbbuddy/bbuddy.c
andreittr: I think there's an argument to be made for Unikraft to provide (via a paging lib or otherwise) 2 kinds of paging API for different consumers (at least conceptually):
- "kernel" paging API: all the arch goodness, knowledge about page table levels, hugepages, PTE flags, the works; basically what `<uk/arch/paging.h>` does and is used by the little code that genuinely cares about paging at a deep level (e.g., `ukvmem`)
- "user" paging API: `PAGE_SIZE` just as a "unit of finest granularity at which we can map memory" -- this is used by _a lot_ of code that doesn't really care about the details of paging, rather just wants a "default chunk size" that matches neatly with the underlying hardware. `bbuddy` is a perfect example of this.

Like I said, this difference should only be in concept; in practice these can be in the same (or dependent) headers. But it's useful to make it clear _how_ a piece of code cares about pages.

lib/ukallocbbuddy/bbuddy.c
andreittr: Such a function would have a sig like `func(struct uk_bbpalloc *b, unsigned long *slackp, char **bufp)` since it needs `b` for context and has to iterate buf & slack in tandem (having caller iterate them separately is asking for bugs). Leaving aside my personal distaste for mixed in/out arguments, such a function would have a hard time arguing its separate existence.
Consider the very unusual and specific pre- and post-conditions under which this code operates (which would need to be explicitly documented & human-enforced before/after calls for a static inline):
- pre: `buf` points to the first byte _after_ the target memory (an otherwise "invalid" pointer, but very useful in this loop)
- pre: `b` must be in an otherwise invalid bbuddy state: target memory is marked free in bitmap but not present in freelists
- post: bbuddy state is valid if-and-only-if `func` returns with a slack of `0`; otherwise you _must keep calling it_ before you do anything else, otherwise everything breaks

IMHO these make it a case of when splitting off a function actually hurts understandability and resilience to human error.

lib/ukallocbbuddy/bbuddy.c
andreittr: The main loop does roughly 3 things every iteration until it exhausts the buffer:
1. find a candidate first chunk to free -- `ptr_order(buf, npages_order(num_pages))` does all the heavy lifting, everything else is simple arithmetic
2. "merge with free buddies" -- this is what the inner loop does, but in practice it just kicks free buddies out of the freelist and updates `freed_ch`, `freed_ct`, and `ord` accordingly. No actual "merging" is done until step 3
3. "create new power-of-2 chunk & link into freelist + advance buffer" -- this is when the actual merge happens and pages are visibly freed

This tight interdependence between steps poses a few issues if one were to attempt splitting off into separate functions (see pre- & post-conditions in `palloc` as well):
- between the start of step 2 and the end of step 3, bbuddy is in an invalid state: pages are marked free but missing from freelists (much like the invalid state in `palloc`'s inner loop) => step 2 must _always_ be followed by step 3
- the loop in step 2 must always be executed until exhaustion, otherwise bbuddy will be in invalid state: "at rest" all pages must be maximally merged (otherwise the quick order-based size check in `palloc` is wrong)
- step 2 requires the initial value of `ord` to have been calculated _exactly_ as it is in step 1, otherwise the entire argument for why the target buddy, if free, has a valid head struct does not hold

Considering these, I make the same case as for `palloc` that splitting off would do more harm than good here.

andreittr:
Pushed changes addressing some comments & improving documentation.
Cosmetic & comment changes ONLY; no functionality diff.

lib/ukallocbbuddy/bbuddy.c
michpappas: There's no doubt it's useful, and as a matter of fact I've meant to add it for a while now, but chose to wait until we move paging into its own library as well as some other re-arch work that will help us determine how to make `__PAGE_SIZE` (which in `arm64` can be configured to values other than 4K for the smallest granule) available even when the paging library is not enabled.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,207 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+
+		UK_ASSERT(ord < FREELIST_SIZE);
+		UK_ASSERT(np <= num_pages);
+
+		/* Try merging with existing free chunks */
+		while (ord < FREELIST_SIZE - 1) {
+			/* It is safe to deref this iff targ is marked free */
+			chunk_head_t *const targ = BBUDDY_BUDDY(freed_ch, ord);
+
+			/* Stop if buddy is not free or of wrong order */
```
michpappas: I was about to say that a few words about why it's possible to get a buddy at a different (not wrong per se) order would help, but now I'm thinking it's paradoxical to request a buddy *of a given order* a few lines above with `BBUDDY_BUDDY(freed_ch, ord)` and in the very next line check if the return buddy is indeed of that order. I understand why it happens, but perhaps we should handle that case at the macro instead and return NULL?

mogasergiu (approved):

michpappas (changes requested):

lib/ukallocbbuddy/bbuddy.c
andreittr: Yeah, might even be worth switching over to [`<uk/list.h>`](https://github.com/unikraft/unikraft/blob/staging/include/uk/list.h) as it's clear we need doubly-linked lists here, and one fewer bespoke implementation is always a good thing. I'll consider it in the pass.

lib/ukallocbbuddy/bbuddy.c
andreittr: Yeah, renaming to `buddy` makes sense, will do. The declaration will be very satisfyingly buddified :smile:  
```c
chunk_head_t *const buddy = BBUDDY_BUDDY(freed_ch, org);
```

lib/ukallocbbuddy/bbuddy.c
andreittr: The two lines (`BBUDDY_BUDDY` & `buddy->level` check) do conceptually different things, and the reasoning here goes as follows:
1. `BBUDDY_BUDDY` is pure arithmetic: it answers "I have a chunk of a particular order; what would the first page of my buddy chunk (i.e., the one to merge with to get a chunk of higher order) be?"
2. At this point we have a pointer to a "target" page, we know nothing about whether our "buddy chunk" actually exists or is free, just that if it were, it would start at this particular page. And that page could be either occupied or free
2a. If target page is occupied, there is no way our buddy could exist and be free, so we stop
3. If the target page is free, we know that is must be a chunk head (see reasoning in comment higher up); we still know nothing about _how big_ this target chunk is.
3a. If the target chunk is indeed of the same size as our chunk (i.e., same ord, read explicitly from the chunk header) then we know we found our buddy, it is free, and we can merge
3b. If the target chunk is of a different size, then it is _not_ our buddy yet -- it still needs to merge with someone else before merging with us. Thus we stop.

Considering the above flow, I don't think it would make sense to burden the `BBUDDY_BUDDY` macro with all 3 steps (which are required to correctly answer the question: "does my buddy at this order exist in whole and is free?"). _Maybe_ there's a case to be made for these 3 steps being put into a `static inline` :shrug: 

I think I should maybe make this logic flow clearer though, with a comment or smth, because indeed it is not obvious why we need to explicitly check the chunk level in addition to buddy arithmetic + checking if free.

andreittr:
Rebased to current staging

andreittr:
Pushed update addressing comments + the previously unconsidered issue that `uk_palloc/free_compat` can no longer be correctly implemented under the newly-clarified page alloc API. Details in the PR description.

andreittr:
Pushed update adding commit to fix type mismatch of `uk_ffsl` between x86 and ARM, leading to compile errors on the latter.

andreittr:

andreittr:
Pushed rebase to current staging + changes & drop of 1 redundant commit after merging of #1652 .

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,218 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-		spare_ch->level = i;
-		spare_ch->next = b->free_head[i];
-		spare_ch->pprev = &b->free_head[i];
-		spare_ct->level = i;
-
-		/* Link in the spare chunk. */
+	UK_ASSERT(FREELIST_ALIGNED(alloc_ch, ord));
+
+	/* Trim off any extra pages */
+	slack = (1ULL << ord) - num_pages;
```
michpappas: For readability, consider renaming `slack` and `buf` to `spare_base` and `spare_page_count`, or similar, to be consistent with `spare_ord`, `spare_len`, and the rest of the variables used for the same context below.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,218 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	buf = CHUNK_END(alloc_ch, ord);
+	while (slack) {
+		const size_t spare_ord = npages_order(slack);
```
michpappas: Please delcare at the function scope.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,218 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+		const size_t spare_len = BBUDDY_MASK(spare_ord);
```
michpappas: Setting length to mask is quite unintuitive, perhaps it would be worth to alias this macro to a more meaningful name? Ideally we should rename `BBUDDY_MASK` too given that it's impossible to tell what it does.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,218 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-		mask = 1UL << (order + __PAGE_SHIFT);
-		if ((unsigned long)freed_ch & mask) {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch - mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
-				break;
+	/* Might need to return the buffer in multiple chunks */
+	do {
+		/* release `np` pages starting at `buf` back into freelists */
+		size_t ord = ptr_order(buf, npages_order(num_pages));
```
michpappas: Please declare at the function scope.

lib/ukallocbbuddy/bbuddy.c
michpappas: Coming back to review this after a while I came up with exactly the same question (i.e. how is it possible for the buddy to be at the wrong order), so maybe that's a sign that there's something to improve here. 
I propose (1) renaming to  `BBUDDY_BUDDY_ADDR()` and (2) to elaborate on the process in the comment on line 474. I post a proposal, feel free to adapt:
```
               /* Try to merge with buddy at incremental orders.
                 * 
                 * At a given level, if the buddy is available to merge (see below)
                 * we create a merged chunk, and continue iterating to 
                 * the next level to see if we can merge into a larger chunk.
                 * Once we reach the max order, or we can't merge with buddy
                 * at the current order (see below), we break from the loop and 
                 * link the resulting chunk to the current order's freelist.
                 *
                 * Looking up the merge buddy involves 3 steps (order is important):
                 * 1. Compute ptr of freed_ch's buddy page at order `ord` (BBUDDY_BUDDY_ADDR)
                 *  - buddy may or may not be there (may have been broken down into smaller chunks).
                 * 3. Check whether said buddy page is free
                 *  - if allocated we can neither merge or trust its chunk_head_t to read level, so we stop
                 * 3. Read buddy chunk head to check size
                 *   - we know it's a valid chunk head because of comment above
                 */
```

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +308,218 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
 
-			/* Merge with successor */
-			freed_ct =
-			    (chunk_tail_t *)((char *)to_merge_ch + mask) - 1;
+			if (BBUDDY_ISHEAD(buddy, ord)) {
+				/* buddy is predecessor; merge downwards */
+				freed_ch = buddy;
+			} else {
+				/* buddy is successor; merge upwards */
+				freed_ct = CHUNK_TAIL_PTR(buddy, ord);
```
michpappas: While looking across the implementation I noticed that we populate the tail in different places but never actually make any use of it. So perhaps the whole `struct chunk_tail_st` / `struct chunk_tail_t` is redundant?

michpappas (changes requested):
Did another round, mostly cosmetic comments. Also before I forget, I believe the Commit Format Check in the CI fails because of the backticks in the PR's title. If you edit it won't be fixed, but just FYI for future PRs.

lib/ukallocbbuddy/bbuddy.c
andreittr: Why?! That would entirely prevent it from being `const`, uselessly expand its scope beyond where it makes any sense, and thus make code _less_ correct. [Coding conventions](https://unikraft.org/docs/contributing/coding-conventions#variable-definitions) say "Variables are defined at the beginning of the block. The block is separated from the code with a blank line." and this is indeed a block. No mention about "Local variables must be declared in the outermost scope possible", nor do I think that such a rule would ever make sense.

lib/ukallocbbuddy/bbuddy.c
andreittr: No, `ord` makes no sense outside a loop iteration, nor should its value mean anything across iterations; a declaration at function scope would make code less correct & readable not more.
I am very strict about minimizing variable scope, and if strongarmed, the only way I see to keep correctness is to make the loop body a `static inline` with ptr in/out args for `buf` and `num_pages`.

lib/ukallocbbuddy/bbuddy.c
andreittr: Sure, with the caveat that `buf` is not the _base_ of the spare area, rather its _end_; will reflect this in name.

lib/ukallocbbuddy/bbuddy.c
andreittr: I think so; I hadn't removed it in the first pass, expecting it to be used _eventually_, but it really does seem that chunk tails are never referenced; will remove.

andreittr:
Pushed major update, addressing comments, and refactoring more of bbuddy.

andreittr:
Pushed cosmetic update to conform to checkpatch.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -49,24 +52,20 @@
...
 
-typedef struct chunk_head_st chunk_head_t;
-typedef struct chunk_tail_st chunk_tail_t;
-
-struct chunk_head_st {
-	chunk_head_t *next;
-	chunk_head_t **pprev;
-	unsigned int level;
+struct chunk_head {
+	struct uk_hlist_node link;
```
michpappas, line 59: Nice change to use `uk_hlist` here, but was there something wrong with `next`? Let's keep conventional naming, no need to unnecessarily obscure things.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+
+static inline size_t bbuddy_trimchunk(struct uk_bbpalloc *b,
+				      unsigned long spare_pages,
+				      char *spare_end)
 {
-	struct uk_bbpalloc *b;
-	size_t i;
-	chunk_head_t *alloc_ch, *spare_ch;
-	chunk_tail_t *spare_ct;
+	const size_t spare_ord = npages_order(spare_pages);
```
michpappas: This block is too noisy. Mixing code declarations with function calls etc is generally discouraged (https://unikraft.org/docs/contributing/coding-conventions#variable-definitions), we usually limit that to simple initializations (e.g. integer or `NULL`) if at all. Please consider a cleaner alternative similar to:

```diff
diff --git a/lib/ukallocbbuddy/bbuddy.c b/lib/ukallocbbuddy/bbuddy.c
index 53f478d7b..b942247a9 100644
--- a/lib/ukallocbbuddy/bbuddy.c
+++ b/lib/ukallocbbuddy/bbuddy.c
@@ -323,16 +323,20 @@ static inline size_t bbuddy_trimchunk(struct uk_bbpalloc *b,
                                      unsigned long spare_pages,
                                      char *spare_end)
 {
-       const size_t spare_ord = npages_order(spare_pages);
-       const size_t spare_len = BBUDDY_LEN(spare_ord);
-       char *spare_base = spare_end - spare_len;
-       struct chunk_head *spare_ch = (struct chunk_head *)spare_base;
+       struct chunk_head *spare_ch;
+       const size_t spare_ord;
+       const size_t spare_len;
 
+       spare_ord = npages_order(spare_pages);
+       UK_ASSERT(spare_ord < FREELIST_SIZE);
+
+       spare_len = BBUDDY_LEN(spare_ord);
        UK_ASSERT(spare_len <= spare_pages * __PAGE_SIZE);
+
        UK_ASSERT(BBUDDY_PAGES(spare_ord) <= spare_pages);
-       UK_ASSERT(spare_ord < FREELIST_SIZE);
 
        /* Populate chunk & link in */
+       spare_ch = (struct chunk_head *)(spare_end - spare_len);
        spare_ch->page_order = spare_ord;
        uk_hlist_add_head(&spare_ch->link, &b->free_head[spare_ord]);
 ```

File: lib/ukallocbbuddy/bbuddy.c [hunk 18]
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	uk_hlist_del(&buddy->link);
+	return 1;
+}
 
-	/* First free the chunk */
-	map_free(b, (uintptr_t)obj, 1UL << order);
+static inline size_t bbuddy_freechunk(struct uk_bbpalloc *b, char *buf,
+				      unsigned long num_pages)
+{
+	size_t ord = ptr_order(buf, npages_order(num_pages));
```
michpappas: Same here, please consider a layout that's easier to read, e.g.:
```diff
@@ -435,13 +439,19 @@ static inline int bbuddy_trymerge(struct uk_bbpalloc *b,
 static inline size_t bbuddy_freechunk(struct uk_bbpalloc *b, char *buf,
                                      unsigned long num_pages)
 {
-       size_t ord = ptr_order(buf, npages_order(num_pages));
-       const size_t np = BBUDDY_PAGES(ord);
-       struct chunk_head *freed_ch = (struct chunk_head *)buf;
+       struct chunk_head *freed_ch;
+       const size_t np;
+       size_t ord;
 
+       ord = ptr_order(buf, npages_order(num_pages));
        UK_ASSERT(ord < FREELIST_SIZE);
+
+       np = BBUDDY_PAGES(ord);
        UK_ASSERT(np <= num_pages);
 
+       freed_ch = (struct chunk_head *)buf;
+       UK_ASSERT(freed_ch);
+
        while (ord < FREELIST_SIZE - 1)
                if (bbuddy_trymerge(b, &freed_ch, ord))
                        ord++;
```

PS: Added an assertion for `freed_ch`, not sure if  a check against `buf` was intentionally omitted before.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	return np;
+}
 
-		/* We are commited to merging, unlink the chunk */
-		*(to_merge_ch->pprev) = to_merge_ch->next;
-		to_merge_ch->next->pprev = to_merge_ch->pprev;
+static void bbuddy_pfree(struct uk_alloc *a, void *obj, unsigned long num_pages)
+{
+	struct uk_bbpalloc *const b = (struct uk_bbpalloc *)&a->priv;
+	char *buf = obj;
```
michpappas: Any reason for the aliasing of `obj` to `buf`? Unless I'm missing something both can be used interchangeably in the function. Btw, I was saving this for a later revision with readability in mind, but since we're there, I always found `buf` an unfortunate name in the various places used in this file, perhaps `base` would be more intuitive? I also had a look at `malloc(3)` and friends, and they use `ptr`. If you think this is not for the present we can leave it for later, but at least let's remove the aliasing if it's indeed not needed.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	size_t freed;
 
-		order++;
-	}
+	UK_ASSERT(a);
```
michpappas, line 470: Would  `UK_ASSERT(b)` also be useful here?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
 
-	/* Now, possibly we can conseal chunks together */
-	while (order < FREELIST_SIZE) {
-		mask = 1UL << (order + __PAGE_SHIFT);
-		if ((unsigned long)freed_ch & mask) {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch - mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
-				break;
+	while (ord < FREELIST_SIZE - 1)
```
michpappas: Maybe a brief comment for this while loop? To me it looks self explanatory, but then again I don't have a clear view because by now I am familiar with how it works. We also document the rest of the steps in this function, so I guess it wouldn't hurt adding a comment here too.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+ * ord if-and-only-if:
+ * (a) the page is aligned to that particular order (all bits < ord are 0), AND
+ * (b) the page is not part of a chunk of larger order
+ *
+ * In the particular care of pfree(X):
+ * (a) is guaranteed because X is aligned to ord, and thus its candidate buddy
+ * must necessarily be aligned too.
+ * (b) is guaranteed to never happen on free, as buddy would need X to already
+ * be free in order to form a larger order
+ */
```
michpappas: This newline is redundant.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+		return 0;
 
-	size_t order = (size_t)num_pages_to_order(num_pages);
+	if (BBUDDY_ISHEAD(buddy, ord))
+		/* buddy is predecessor; merge downwards */
+		*freed_ch = buddy;
 
-	/* if the object is not page aligned it was clearly not from us */
-	UK_ASSERT((((uintptr_t)obj) & (__PAGE_SIZE - 1)) == 0);
+	/* Unlink buddy & continue to next order*/
```
michpappas: "unlink and continue to next order" does not belong to this function, but to the caller. Which is of course not possible, as the caller doesn't know anything about the buddy. Now, I know this was made into static inline because of malicious compliance, but that's not the way forward. Maybe pass the candidate buddy's address to `bbuddy_trymerge()` instead, or merge this whole functionality into `bbuddy_freechunk()`?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	uk_hlist_del(&buddy->link);
+	return 1;
```
michpappas: The return value for success is a bit unconventional, convert into bool or reverse the logic?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
-	/* Now, possibly we can conseal chunks together */
-	while (order < FREELIST_SIZE) {
-		mask = 1UL << (order + __PAGE_SHIFT);
-		if ((unsigned long)freed_ch & mask) {
-			to_merge_ch = (chunk_head_t *)((char *)freed_ch - mask);
-			if (allocated_in_map(b, (uintptr_t)to_merge_ch)
-			    || to_merge_ch->level != order)
-				break;
+	while (ord < FREELIST_SIZE - 1)
+		if (bbuddy_trymerge(b, &freed_ch, ord))
```
michpappas: Perhaps should be called `buddy_canmerge()` instead? While reading this I expected that the function will also go ahead do the merging, if it can, which is not the case really.

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -291,135 +296,187 @@ static inline unsigned long num_pages_to_order(unsigned long num_pages)
...
+	struct chunk_head *const buddy = BBUDDY_BUDDY_ADDR(*freed_ch, ord);
 
-	freelist_sanitycheck(b->free_head);
+	/* Stop if buddy is not free or of wrong order */
+	if (allocated_in_map(b, (uintptr_t)buddy) || buddy->page_order != ord)
+		return 0;
 
-	size_t order = (size_t)num_pages_to_order(num_pages);
+	if (BBUDDY_ISHEAD(buddy, ord))
+		/* buddy is predecessor; merge downwards */
```
michpappas: Perhaps one more sentence to clarify that if the buddy is not the head no action required as `free_ch` already points to head?

lib/ukallocbbuddy/bbuddy.c
```diff
@@ -49,24 +52,20 @@
...
-typedef struct chunk_head_st chunk_head_t;
-typedef struct chunk_tail_st chunk_tail_t;
-
-struct chunk_head_st {
-	chunk_head_t *next;
-	chunk_head_t **pprev;
-	unsigned int level;
+struct chunk_head {
+	struct uk_hlist_node link;
+	unsigned int page_order;
```
michpappas, line 60: Also nice change to remove redundancy, it was one of the first things I noticed when I first read this 👍🏼

lib/ukallocbbuddy/bbuddy.c
michpappas: > [Coding conventions](https://unikraft.org/docs/contributing/coding-conventions#variable-definitions) say "Variables are defined at the beginning of the block. The block is separated from the code with a blank line." and this is indeed a block. No mention about "Local variables must be declared in the outermost scope possible", nor do I think that such a rule would ever make sense.

Not quite. It is in fact documented, although quite poorly (see the `num` variable in the example).

image

Anyway, let's take this offline.

michpappas (changes requested):
@andreittr with the cleanup of the `chunk_head` and the use of `uk_hlist` it looks much better! 🚀  Went through it once again, didn't find any issues. Please bare with the last comments on readability, ready to RB after this round.

lib/ukallocbbuddy/bbuddy.c
andreittr: `next` makes no sense as a field name here as a hlist link contains both a `.next` and a `.pprev` field (hlist is a DLL, and we do need that for quick removal out of freelists on buddy merging).
Previously yes, these names made sense as `chunk_head` contained the linked list pointers as raw fields within itself, wheras now with `hlist`, we need to use a single field of type `struct uk_hlist_node`, which itself has `.next` and `.pprev` fields.
Rather than naming things confusingly (but in keeping with tradition) I'd rather be clear. As example to contrast which of these look more sane:
```c
struct chunk_head *h;

h->next.next = next_elem;
h->next.pprev = &prev_elem;
```
or
```c
struct chunk_head *h;

h->link.next = next_elem;
h->link.pprev = &prev_elem;
```
(I know these ops are hidden behind macros/inlines, but that's no reason to introduce extra madness into one's code; plus, this is what people would see under a debugger).

Or, to answer the question strictly, yes, there was a lot wrong with `next`, as it was an explicit linked list pointer instead of a link struct field, as `uk_hlist` uses. Since we changed the meaning of the field, it makes sense to change its name to match.

lib/ukallocbbuddy/bbuddy.c
andreittr: ~~Uhm, assigning to `const` variables after their definition is a no-no; the above code will not compile. I'm assuming that's an honest oversight.~~

~~That said, are you arguing that cosmetics are worth more than the extra safety against human screwups that immutable vars gives you? It's fine if you are (even if I disagree completely), but I want it explicit.~~

edit: Similar to subscope declarations, this was a case of me completely misreading the coding conventions and assuming degrees of freedom that weren't there. In light of this, the previous comments were out of line; standards aren't to be argued ad-hoc and I agree this code wasn't following them. Fixing in refactor/pass.

lib/ukallocbbuddy/bbuddy.c
andreittr: Aw crap, I completely missed the details in the example and have been having the wrong mental picture all along :facepalm: ; you're right, my bad.

lib/ukallocbbuddy/bbuddy.c
andreittr: The aliasing is only for doing pointer arithmetic on `buf` later on. The `pfree` signature requires us to take a `void *` (which we can't do arithmetic on directly), and manually casting
```c
obj = (char *)obj + freed * __PAGE_SIZE;
```
looks a bit noisy to me compared to a straight addition to `buf` :shrug: .

Re: naming, sure, `base` sounds good; I somewhat prefer it over `ptr` in this case as it more clearly expresses its role ("the start (base) of the buffer we're trying to divvy up into bbuddy chunks when freeing").

lib/ukallocbbuddy/bbuddy.c
andreittr: :thinking: considering that `b` is always a constant away from `a` (a quick read of `ukalloc` says 14-16 pointers + the `uk_alloc_stats` (80 bytes) beyond `a`, depending on config). (note that `*b`'s contents are entirely embedded within `*a`, it's not a pointer field).

The first `UK_ASSERT(a)` guards us against the oopsie of someone passing a NULL allocator instance, an expected mistake and good to catch. A `UK_ASSERT(b)` would then only guard against passing `a` as `(void *)-208` (or -192/-128/-112 depending on config), which I feel is a bit too specific to be triggered by mistake, and thus I'm unsure whether it's worth guarding against at runtime.

lib/ukallocbbuddy/bbuddy.c
andreittr: Unlinking the buddy is actually completely done by `uk_hlist_del` below; the caller merely decides whether to move onto the next order or stop according to the return value of this function (thus `return 1` _technically_ tells the caller to continue to next order).
But I see how logic jumping from one func to another is a bit convoluted, I'll see if I can refactor these 3 functions better in another pass.

lib/ukallocbbuddy/bbuddy.c
andreittr: Yeah, without context from the caller (which should never be required to understand a function) it's ambiguous what this means; will `bool`ify / refactor in pass.

lib/ukallocbbuddy/bbuddy.c
andreittr: > Added an assertion for freed_ch, not sure if a check against buf was intentionally omitted before.

Good catch, thanks; a NULL `obj`/`buf` should definitely be handled _somehow_. Whether gracefully or not I'm not sure; NULL is indeed an err value for `palloc`, so either way it should not be handled normally by `pfree`.
- On one hand, decreeing that `obj` must be non-NULL and crashing on non-compliance makes sense (and the [proposed wording](https://github.com/unikraft/unikraft/pull/1576/files#diff-c4f328020b85f182463f93103962fa30d9bd65d7e882c5784d75ee4f1322c94cR344) of pfree docs kinda says that already)
- OTOH, related APIs (e.g., `free()`) are explicitly NOPs when fed NULL, and people might expect symmetry / the principle of least surprise (although `palloc`/`pfree` is different enough to warrant being special :shrug: )

I'll go ahead and assert the argument, but let me know if you think NOPing out on NULL is the better call.

lib/ukallocbbuddy/bbuddy.c
andreittr: `bbuddy_trymerge` does indeed do the merging; it's just that "merging" in this context means strictly only:
- [unlinking](https://github.com/unikraft/unikraft/pull/1576/files#diff-7a854c8df048714d0efcab86bedab15bf77f36afe24270e946fd1bdb578c14afR431) the corresponding buddy from its freelist
- (if buddy was a lower address than `freed_ch`) [updating `freed_ch`](https://github.com/unikraft/unikraft/pull/1576/files#diff-7a854c8df048714d0efcab86bedab15bf77f36afe24270e946fd1bdb578c14afR428) to point to the lower end

Any additional bookkeeping isn't really required until the last step when we insert the final, maximal chunk back into freelists.

But yeah, having merge logic split up like this makes it harder to grok; I'll try refactoring in the pass.

andreittr:
Pushed minor refactor & addressing comments.

lib/ukallocbbuddy/bbuddy.c
michpappas: @andreittr many thanks for that. The part about `const` was of course an oversight of mine, so apologies. Review comments are always in good faith 🙃 Regarding the conventions: They are in many places quite unclear and could take a lot of improvement, so let's consider this an opportunity give them a revamp to avoid similar cases in the future.

lib/ukallocbbuddy/bbuddy.c
michpappas: Oh right, I see your point. Somehow my mind at that moment was hardwired to the previous state where `chunk_head_st` had `next` and `pprev` embedded.

lib/ukallocbbuddy/bbuddy.c
michpappas: Okay fair enough, it's more of a matter of personal preference then.

lib/ukallocbbuddy/bbuddy.c
michpappas: Hmmmm, I don't have a strong opinion. The important thing is to handle it somehow, so let's go for the suggested approach 👍🏼 We can always change it later if there is good reason.

lib/ukallocbbuddy/bbuddy.c
michpappas: > note that *b's contents are entirely embedded within *a, it's not a pointer field

Yup, that was it. I took it to be a pointer. It's all good now.

michpappas (approved):
Good to go 🚀  @andreittr thanks for putting up with the elaborate review on this one.

PR #1577, opened 2025-02-06
lib/syscall_shim: Autogenerate `uk_syscall_*` symbols at build time

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Introduce a new AWK script that autogenerates an additional, custom 
defined, exportsyms that contains all `uk_syscall_*` symbols as per 
each library's `UK_SYSCALLS_PROVIDED-*`.                            
                                                                    
This will help in reducing exportsyms.uk noise from now on as we no 
longer have to manually add these supposedly internal symbols.      

Thus, also remove such no longer necessary symbols.

This was done with the Introduction of `EACHOLIB_EXPORTS`,
the equivalent of `EACHOLIB_LOCALS`, but for exportsyms.uk.  
Now, one is able to add custom defined/named exportsyms.uk, in addition
to an already existing exportsyms.uk. If no exportsyms.uk exists, then,
unlike `EACHOLIB_LOCALS` which is processed regardless of whether a    
localsyms.uk exists or not, `EACHOLIB_EXPORTS` will be ignored.        
However, its result files may still be built as per Makefile recipe    
dependencies.


lib/posix-futex/exportsyms.uk
```diff
@@ -1,4 +1 @@
-uk_syscall_e_futex
-uk_syscall_r_futex
-uk_syscall_e_set_tid_address
-uk_syscall_r_set_tid_address
+none
```
michpappas: This exports a symbol called `none`, so what was done in `posix-futex` is definitely a hack to work around some bug in the build system. Do you think it's something easy to fix instead of propagating this hack? What I had in mind was wrapping calls to objcopy around a function that first checks if `exportsyms.uk` is present.

michpappas (changes requested):

mogasergiu:
Rebase force push

lib/posix-futex/exportsyms.uk
mogasergiu: Opened an issue for this so that we can look at it later.

mogasergiu:
Rebase push

lib/posix-futex/exportsyms.uk
mogasergiu: Pushed an update that uses `# No public symbols` instead of `none` to be more explicit and consistent.

mogasergiu: So I tried allowing empty `exportsyms.uk` to exist
```diff
diff --git a/support/build/Makefile.rules b/support/build/Makefile.rules
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -1024,8 +1024,13 @@ $(call libname2olib,$(1)): $(call libname2preolib,$(1)) \
                $$($(call vprefix_lib,$(1),LOCALS)) $$($(call vprefix_lib,$(1),LOCALS-y)) \
                $$(EACHOLIB_LOCALS) $$(EACHOLIB_LOCALS-y)
     $(call build_cmd,OBJCOPY,,$(call libname2olib,$(1)),\
-        $(OBJCOPY) $$(addprefix --keep-global-symbols=,$$($(call vprefix_lib,$(1),EXPORTS)) $$($(call vprefix_lib,$(1),EXPORTS-y))) \
-               $$(addprefix --localize-symbols=,$$($(call vprefix_lib,$(1),LOCALS)) $$($(call vprefix_lib,$(1),LOCALS-y))) \
+        $(OBJCOPY) \
+               $$(for f in $$($(call vprefix_lib,$(1),EXPORTS)) $$($(call vprefix_lib,$(1),EXPORTS-y)); do \
+                [ -s "$$$$f" ] && echo --keep-global-symbols="$$$$f"; \
+               done) \
+               $$(for f in $$($(call vprefix_lib,$(1),LOCALS)) $$($(call vprefix_lib,$(1),LOCALS-y)) $$(EACHOLIB_LOCALS) $$(EACHOLIB_LOCALS-y); do \
+                [ -s "$$$$f" ] && echo --localize-symbols="$$$$f"; \
+               done) \
                $$(addprefix --localize-symbols=,$$(EACHOLIB_LOCALS) $$(EACHOLIB_LOCALS-y)) \
                    $$(OBJCFLAGS) $$(OBJCFLAGS-y) \
                    $$($(call vprefix_lib,$(1),OBJCFLAGS)) $$($(call vprefix_lib,$(1),OBJCFLAGS-y)) \
```
while this works for not making objcopy error out on empty files given an argument to `--keep-global-symbols`/`--localize-symbols`, it is not enough to ensure that no symbols get exported. What we have to do is to essentially create the equivalent of a `localsyms.uk` for all public symbols of that library.

Therefore, since this is seemingly not just a quick fix and may require some additional thought, I have opened an internal issue for this in the backlog.

michpappas:
Very cool to see so much noise symbols go away 🚀 

michpappas (approved):

PR #1583, opened 2025-02-07
 lib/posix-process: Define kernel internal `gettid`/`getppid`/`getpid`

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Add the kernel internal variant of `gettid`, `uk_sys_gettid`. This  
allows kernel internal code to call this system call's logic without
having the syscall shim wrapper logic intervene.                    

While at it, also use this variant where possible, only place at the moment being in posix-futex.

### DIFF
--- a/lib/posix-process/include/uk/process.h
+++ b/lib/posix-process/include/uk/process.h
@@ -149,6 +149,7 @@ static inline int uk_sys_setrlimit(int resource, const struct rlimit *rlim)
 
 pid_t uk_sys_gettid(void);
 pid_t uk_sys_getppid(void);
+pid_t uk_sys_getpid(void);
 
 #if CONFIG_LIBUKSCHED
 int uk_posix_process_create(struct uk_alloc *a,
--- a/lib/posix-process/include/uk/process.h
+++ b/lib/posix-process/include/uk/process.h
@@ -38,6 +38,7 @@
 #include <uk/config.h>
 #include <stdbool.h>
 #include <sys/resource.h>
+#include <sys/types.h> /* pid_t */
 #if CONFIG_LIBUKSCHED
 #include <uk/thread.h>
 #endif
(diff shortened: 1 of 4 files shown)


andreittr (commented):
I like the changes, but am going to cheekily ask to expand the scope of this PR to include the same treatment for `getpid` and `getppid`, as it would feel far cleaner than just this piecemeal approach. AFAICT these latter 2 aren't called anywhere internally, so the risk is minimal.

lib/posix-process/include/uk/process.h
```diff
@@ -131,6 +132,9 @@
 #endif
 #endif /* CONFIG_LIBPOSIX_PROCESS_CLONE */
 
+pid_t uk_sys_gettid(void);
+pid_t uk_sys_getppid(void);
```
andreittr, line 151: I believe you forgot `uk_sys_getpid(void)`

lib/posix-process/process.c
```diff
@@ -530,7 +530,7 @@ pid_t ukthread2pid(struct uk_thread *thread)
 	return pthread->process->pid;
 }
 
-UK_SYSCALL_R_DEFINE(pid_t, getpid)
+int uk_sys_getpid(void)
```
andreittr: I was about to ask why you changed the return type from `pid_t` to `int`, but I see it has to do with signature inconsistencies between syscall declarations on either side of the `#ifdef #else` :facepalm: this return type inconsistency will surely bite us in the ass eventually.

I think that previous code was a mistake and having everybody (syscalls & `uk_sys_` alike) return `pid_t` here is the right call here; AFAICT the definition of `pid_t` does not depend on `CONFIG_LIBPOSIX_PROCESS_PIDS` so it should be safe. What do you think?

andreittr (changes requested):
Thanks for the PR expansion/DLC :smile: :pray: 
Just a few minor nitpicks, and we should be good to go.

andreittr (approved):
Looks great, thanks for the work! I'm going to wait on @michpappas 's review before merge.

michpappas:
Nice, thanks! This will also allow to prepending logs with pid / tid in multiprocess.

andreittr:
Merging, thanks!

PR #1584, opened 2025-02-07
lib/posix-process: Define kernel internal resource limit syscalls

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Add the kernel internal variants of `prlimit64`, `getrlimit` and
`setrlimit`: `uk_sys_prlimit64`, `uk_sys_getrlimit` and
`uk_sys_setrlimit` respectively.
This allows kernel internal code to call these system calls' logic
without having the syscall shim wrapper logic intervene.

Note how only `uk_sys_prlimit64` has been added to `exportsyms.uk`
since the others are defined as inline.

### DIFF
--- a/lib/posix-process/include/uk/process.h
+++ b/lib/posix-process/include/uk/process.h
@@ -37,6 +37,7 @@
 #include <arch/clone.h>
 #include <uk/config.h>
 #include <stdbool.h>
+#include <sys/resource.h>
 #if CONFIG_LIBUKSCHED
 #include <uk/thread.h>
 #endif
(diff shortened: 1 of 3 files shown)


lib/posix-process/include/uk/process.h
```diff
@@ -131,6 +132,9 @@
 #endif
 #endif /* CONFIG_LIBPOSIX_PROCESS_CLONE */
 
+int uk_sys_prlimit64(int pid, unsigned int resource,
+		     struct rlimit *new_limit, struct rlimit *old_limit);
+
```
andreittr, line 137: More of a QoL enhancement: consider adding
```c
static inline
int uk_sys_getrlimit(int resource, struct rlimit *rlim)
{
        return uk_sys_prlimit64(0, resource, NULL, rlim);
}
```
and similar ones for `setrlimit` and `getrusage`, so you encapsulate the "conversion to prlimit64" logic in `uk_sys_*` rather than at the syscall level.
Then you can call these from their syscall implementations just like `uk_sys_prlimit64`.

andreittr (commented):
Looks good, I've only one very minor request.

andreittr (approved):
Thanks for the patch! I see you also fixed the previously-uncaught issue with the `prlimit` libc func.
I think this is good to go.

andreittr:

PR #1364, opened 2024-03-21
lib/posix-process: Graceful shutdown

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

New options:

* `LIBPOSIX_PROCESS_GRACEFUL_SHUTDOWN` enables graceful shutdown (see below).
* `LIBPOSIX_PROCESS_GRACEFUL_SHUTDOWN_SIGNAL` controls the signal sent on graceful shutdown (defaults to `SIGTERM`).
* * `LIBPOSIX_PROCESS_GRACEFUL_SHUTDOWN_SIGNAL` controls the time to wait for before forcefully terminating a process (defaults to 3 sec).

Deprecated options:
* The behavior or `LIBPOSIX_PROCESS_INIT_PIDS` is now enabled by `LIBPOSIX_PROCESS_PIDS`, and the former option is deprecated.

### Description of changes

This series implements support for clean system shutdown, and adds a few associated syscalls. Specifically, it adds a late initicall to terminate all processes on system shutdown whether gracefully or not. Graceful process termination is controlled by a newly introduced Kconfig option, where the implementation sends every child of `init` a configuration-controlled signal (`SIGTERM` by default) and waits for the process to terminate on its own until a predefined timeout. Upon timeout the process is terminated forcefully.

When graceful shutdown is not enabled the system falls back into forceful termination.

Syscall implementations:
* `wait4()`, `waitpid()`
* `exit()`, `exit_group()`

Github-Depends-On: #1248 
Github-Depends-On: #1244

PR #1585, opened 2025-02-07
 lib/posix-process: Define kernel internal `exit_group`

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Add the kernel internal variant of `exit_group`, `pprocess_exit`. This allows kernel internal code to call this system call's logic without having the syscall shim wrapper logic intervene.

### DIFF
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -580,23 +580,28 @@ UK_LLSYSCALL_R_DEFINE(int, exit, int, status)
 	return -EFAULT;
 }
 
-UK_LLSYSCALL_R_DEFINE(int, exit_group, int, status)
+static int pprocess_exit(int status __unused)
 {
 	uk_posix_process_kill(uk_thread_current()); /* won't return */
 	UK_CRASH("sys_exit_group() unexpectedly returned\n");
 	return -EFAULT;
 }
 
+UK_LLSYSCALL_R_DEFINE(int, exit_group, int, status)
+{
+	return pprocess_exit(status);
+}
+
 #if UK_LIBC_SYSCALLS
 __noreturn void exit(int status)
 {
-	uk_syscall_r_exit_group(status);
+	pprocess_exit(status);
 	UK_CRASH("sys_exit_group() unexpectedly returned\n");
 }
 
 __noreturn void exit_group(int status)
 {
-	uk_syscall_r_exit_group(status);
+	pprocess_exit(status);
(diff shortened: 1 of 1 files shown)


lib/posix-process/exportsyms.uk
```diff
@@ -73,6 +73,7 @@ uk_syscall_e_gettid
 exit
 uk_syscall_r_exit
 uk_syscall_e_exit
+uk_sys_exit_group
```
michpappas: I think we don't have a use for exporting `exit_group()` in the kernel API. I also had a look at the [patch](https://github.com/unikraft/unikraft/pull/1364/commits/a4d929510ab7035c3871c3c715420fcb32cfd9b2) in the graceful shutdown PR that implements `exit()` / `exit_group()` and confirmed we only use such a function internally in `libposixprocess`. So maybe we can make `uk_sys_exit_group()` a `static` instead to avoid making a syscall from the kernel, and since it's no longer part of the kernel API, to rename it to something else (`pprocess_exit()` is okay as I'm using that in #1364).

michpappas (changes requested):

michpappas (approved):

lib/posix-process/process.c
```diff
@@ -580,23 +580,28 @@ UK_LLSYSCALL_R_DEFINE(int, exit, int, status)
 	return -EFAULT;
 }
 
-UK_LLSYSCALL_R_DEFINE(int, exit_group, int, status)
+static int pprocess_exit(int status __maybe_unused)
```
andreittr: Is `status` only _maybe_ unused? I don't see a code path that might use it in the body, but maybe there's a good reason for not using plain `__unused`?
(I know they both map to the same compiler attribute eventually, hence this nitpick's irrelevance)

andreittr (approved):
Yeah, as @michpappas said, it makes more sense to keep this internal to `posix-process` for now. Changes look good modulo one irrelevant nitpick that I'll leave up to you to fix or not before merge.

lib/posix-process/process.c
mogasergiu: Huh, I used `__maybe_unused` without thinking too much since that is what the `UK_LLSYSCALL_R_DEFINE` does under the hood. Though it does do that for a good reason, in order to stay as generic as possible. Totally not the case here, good point. Pushing an update

andreittr:
Thanks! Merging.

PR #1586, opened 2025-02-07
lib/vfscore: Use kernel internal `clock_gettime` variant

mogasergiu:
### Prerequisite checklist

 - [ ] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [ ] Tested your changes against relevant architectures and platforms;
 - [ ] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [e.g. `x86_64` or N/A]
 - Platform(s): [e.g. `kvm`, `xen` or N/A]
 - Application(s): [e.g. `app-python3` or N/A]

### Description of changes

Use the kernel internal variant of `clock_gettime`, `uk_sys_clock_gettime`. This helps avoid unnecessary execution of the syscall wrappers' logic of syscall shim that would have otherwise been run through `uk_syscall_r_clock_gettime`. Additionally, make sure to also handle errors of said syscall.

PR #1587, opened 2025-02-07
lib/ramfs: Use kernel internal `clock_gettime` variant

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Use the kernel internal variant of `clock_gettime`, `uk_sys_clock_gettime`.
This helps avoid unnecessary execution of the syscall wrappers' logic of syscall shim that would have otherwise been run through `uk_syscall_r_clock_gettime`.

Additionally, make sure to also handle errors of said syscall.

Lastly, since now RAMFS uses a definition available only through posix-time, add a dependency to it in the Config.uk. Ideally, this should have been a depends on HAVE_TIME and an imply, but seeing that currently this is the only library offering us time services, do it like this for now.

### DIFF
--- a/lib/ramfs/ramfs_vnops.c
+++ b/lib/ramfs/ramfs_vnops.c
@@ -56,6 +56,8 @@
 #include <fcntl.h>
 #include <vfscore/fs.h>
 
+#include <uk/posix-time.h>
+
 /* 16 bits are enough for file mode as defined by POSIX, the rest we can use */
 #define RAMFS_MODEMASK 0xffff
 #define RAMFS_DELMODE 0x10000
@@ -70,8 +72,19 @@ set_times_to_now(struct timespec *time1, struct timespec *time2,
 		 struct timespec *time3)
 {
 	struct timespec now;
+	int rc;
 
-	clock_gettime(CLOCK_REALTIME, &now);
+	rc = uk_sys_clock_gettime(CLOCK_REALTIME, &now);
+	if (unlikely(rc)) {
+		/*
+		 * We do not consider this a fatal error, but rather
+		 * something that might indicate a misconfigured kernel.
+		 * Just print the error and ignore.
+		 */
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		UK_ASSERT(rc < 0);
+	}
+
 	if (time1)
 		memcpy(time1, &now, sizeof(struct timespec));
 	if (time2)
--- a/lib/ramfs/Config.uk
+++ b/lib/ramfs/Config.uk
@@ -2,3 +2,4 @@ config LIBRAMFS
 	bool "ramfs: simple RAM file system"
 	default n
 	depends on LIBVFSCORE
+	select LIBPOSIX_TIME


lib/ramfs/ramfs_vnops.c
```diff
@@ -65,25 +67,34 @@
...
 set_times_to_now(struct timespec *time1, struct timespec *time2,
 		 struct timespec *time3)
 {
 	struct timespec now;
+	int rc;
+
+	rc = uk_sys_clock_gettime(CLOCK_REALTIME, &now);
+	if (unlikely(rc)) {
+		UK_ASSERT(rc < 0);
+		return rc;
```
andreittr: Here you return negative error codes, but in later calls (e.g., `ramfs_remove_node`) you return this exit code verbatim, when it should be made positive instead. Either make this func follow the (unusual) vfscore practice of positive error codes, or make sure to handle it properly later (IMO former is easier and less error-prone).

lib/ramfs/ramfs_vnops.c
```diff
@@ -108,7 +119,12 @@ ramfs_allocate_node(const char *name, int type, mode_t mode)
 	else
 		np->rn_mode = S_IFREG|mode;
 
-	set_times_to_now(&(np->rn_ctime), &(np->rn_atime), &(np->rn_mtime));
+	rc = set_times_to_now(&(np->rn_ctime), &(np->rn_atime),
+			      &(np->rn_mtime));
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		return NULL;
```
andreittr: Ooof, this eats up any meaningful errno from `rc` & converts to NULL, which is signaled as ENOMEM later (yes, we do have the print, but calling code won't know).
This is not ideal, but converting the API to errptrs is very invasive and risky.

Perhaps failure to set the current time isn't fatal in this context?
As opposed to `utime(ns)` from https://github.com/unikraft/unikraft/pull/1586, who _explicitly_ depend on having a valid time to work with, here the main operation -- allocating a new ramfs node -- is successful.

It might be worth just printing an error/warning and continuing; I believe that would have far fewer changes of breaking things.

lib/ramfs/ramfs_vnops.c
```diff
@@ -145,7 +162,11 @@ ramfs_add_node(struct ramfs_node *dnp, const char *name, int type, mode_t mode)
 		prev->rn_next = np;
 	}
 
-	set_times_to_now(&(dnp->rn_mtime), &(dnp->rn_ctime), NULL);
+	rc = set_times_to_now(&(dnp->rn_mtime), &(dnp->rn_ctime), NULL);
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		return NULL;
```
andreittr: Similar argument to the one in `ramfs_allocate_node` above: should we consider making this a non-fatal condition?

lib/ramfs/ramfs_vnops.c
```diff
@@ -177,7 +199,11 @@ ramfs_remove_node(struct ramfs_node *dnp, struct ramfs_node *np)
 	/* Mark node as deleted */
 	RAMFS_MARK_DELETED(np);
 
-	set_times_to_now(&(dnp->rn_mtime), &(dnp->rn_ctime), NULL);
+	rc = set_times_to_now(&(dnp->rn_mtime), &(dnp->rn_ctime), NULL);
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		return rc;
```
andreittr: Here we should _definitely_ not error out, because the remove operation has successfully completed by this point. Callers will _never_ expect a failing `unlink()` to still remove the file.

You need to do one of the following:
- move the `set_times_to_now()` call & errorout before any persistent changes to the fs
- revert any changes to the fs before erroring out
- treat as non-fatal, similar to `ramfs_alloc_node`
(I favor the last option; first 2 are invasive and risky)

lib/ramfs/ramfs_vnops.c
```diff
@@ -206,7 +233,12 @@ ramfs_rename_node(struct ramfs_node *np, const char *name)
 		np->rn_name = tmp;
 	}
 	np->rn_namelen = len;
-	set_times_to_now(&(np->rn_ctime), NULL, NULL);
+	rc = set_times_to_now(&(np->rn_ctime), NULL, NULL);
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		return rc;
```
andreittr: Similar argument to the one in `ramfs_remove_node`: FS changes are persistent at this point, we shouldn't error out.

lib/ramfs/ramfs_vnops.c
```diff
@@ -322,7 +355,12 @@ ramfs_readlink(struct vnode *vp, struct uio *uio)
 	else
 		len = uio->uio_resid;
 
-	set_times_to_now(&(np->rn_atime), NULL, NULL);
+	rc = set_times_to_now(&(np->rn_atime), NULL, NULL);
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to set current time for vnode: %d\n", rc);
+		return rc;
```
andreittr: This is the one case where it might be fine as-is (modulo `rc` being of wrong sign):
- It is before any persistent fs changes
- The rc is not eaten up but returned

I leave it up to you whether you want to keep erroring here, or follow the reasoning as for `ramfs_alloc_node` and treat it as non-fatal.

andreittr (changes requested):
Great initiative, though there are some issues, see comments.

lib/ramfs/ramfs_vnops.c
mogasergiu: Oh yeah, definetely the first one (positive errors), I did this part before doing the vfscore changes so I was not aware of this weird pattern. I was taken by surprise when I first saw it :laughing:. While vfscore did its job while it lasted, I can't wait for you to retire it :pray:.

lib/ramfs/ramfs_vnops.c
mogasergiu: I guess we have to do it that way for now, yes. I don't really like the way this whole file feels, but oh well... it is legacy code after all.

andreittr (approved):
Looks good, thanks! Looking forward to less `strace` spam :smile: 

andreittr:
Let's get this merged!

PR #1591, opened 2025-02-14
lib/vfscore: Use `uk_syscall_do` for syscall delegation

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Since `uk_syscall_r_` symbols tend to also invoke the system call enter
and exit tables if the `syscall_shim` library is enabled, replace such 
calls with the `uk_syscall_do_` symbol variant which does not involve  
any tables.

### DIFF
--- a/lib/vfscore/main.c
+++ b/lib/vfscore/main.c
@@ -186,7 +186,7 @@ UK_LLSYSCALL_R_DEFINE(int, openat, int, dirfd, const char *, pathname,
 		      int, flags, int, mode)
 {
 	if (pathname[0] == '/' || dirfd == AT_FDCWD) {
-		return uk_syscall_r_open((long int)pathname, flags, mode);
+		return uk_syscall_do_open((long int)pathname, flags, mode);
 	}
 
 	struct vfscore_file *fp;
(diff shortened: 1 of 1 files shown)


mogasergiu:
NOTE: Hastily submitted so we can merge #1277. Thorough review required.

mogasergiu:
Rebase fpush

andreittr (changes requested):
GH won't let me comment on a line not in the diff :roll_eyes:.
I think you also need to patch `uk_syscall_e_open` in the `creat` syscall impl @ line 241.

mogasergiu:
> GH won't let me comment on a line not in the diff 🙄. I think you also need to patch `uk_syscall_e_open` in the `creat` syscall impl @ line 241.

Good catch, I only just `%s/uk_syscall_r_/uk_syscall_do_`. I forgot about the wrongly registered system calls (not using the raw variant).

However, actually, I am not so sure. This  is could not the case here. Note that `creat` is **sadly** registered with the `UK_SYSCALL_DEFINE` variant which handles `errno` differently. However, the `open` syscall is registered with `UK_LLSYSCALL_R_DEFINE`, which handles it the exact opposite way. Correct me if I am wrong though. I am afraid that we either have to convert `creat` to `UK_SYSCALL_R_DEFINE` or simply accept that it will end up calling into the tables twice.

andreittr:
> Note that creat is sadly registered with the `UK_SYSCALL_DEFINE` variant which handles errno differently

Oooof, I had not noticed that; how many syscall definition macros do we have? :smile: 
Considering every other similar syscall in vfscore is `_R`, I think converting to `UK_SYSCALL_R_DEFINE` + using `_do_open` is the best course here.

andreittr (approved):
Alright, looks good, thanks!

andreittr:

PR #1595, opened 2025-03-03
lib/ukfs: Unikraft Filesystem Interface

andreittr:
### Description of changes

This PR introduces `ukfs`, the core filesystem API for `ukfile`s, enabled on-demand by selecting the `ukfs` library for build.
This adds a "filesystem ops" table reference to every file, with "true filesystem" files having it populated by their driver, and pseudo-files setting it to `NULL`.
Existing file implementations that initialize their `struct uk_file` with designated initializers (all from what I can tell) are automatically backwards-compatible.

This PR provides no drivers or callers into the `ukfs` API; rather it serves as a point of discussion of the interface itself and its logical consistency.

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/ukfs/Makefile.uk
```diff
@@ -0,0 +1,9 @@
+$(eval $(call addlib_s,libukfs,$(CONFIG_LIBUKFS)))
+
+CINCLUDES-$(CONFIG_LIBUKFS) += -I$(LIBUKFS_BASE)/include
+CXXINCLUDES-$(CONFIG_LIBUKFS) += -I$(LIBUKFS_BASE)/include
+
+LIBUKFS_SRCS-y += $(LIBUKFS_BASE)/common-ops.c
+LIBUKFS_SRCS-y += $(LIBUKFS_BASE)/drivers.c
+
+LIBUKFS_SRCS-y += $(LIBUKFS_BASE)/driver-list.ld
```
mogasergiu, line 9: I don't see any other source file with `-` in the name in our codebase, feels inconsistent :thinking:
Also maybe this should have used the `.lds.S` suffix instead?

lib/ukfs/drivers.c
```diff
@@ -0,0 +1,38 @@
...
+extern const struct uk_fs_drv uk_fs_driver_list_start[];
+extern const struct uk_fs_drv uk_fs_driver_list_end[];
+
+#define UKFS_DRIVER_COUNT \
+	(uk_fs_driver_list_end - uk_fs_driver_list_start)
+
+uk_fs_vopen_func uk_fs_driver(const char *fstype)
+{
+	const struct uk_fs_drv *drv = uk_fs_driver_list_start;
+	size_t len = UKFS_DRIVER_COUNT;
```
mogasergiu, line 21: `__sz`? Ditto for other occurences of `size_t`, especially since it's supposed to be Unikraft internal.

lib/ukfs/include/uk/fs/driver.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Convenience macro to register a ukfs driver.
+ *
+ * @param fsname Filesystem name to register; do not enclose in quotes
+ * @param drv_vopen vopen function of filesystem driver
+ */
+#define UK_FS_DRIVER_REGISTER(fsname, drv_vopen)		\
+	__used __align(8)					\
+		__section("." STRINGIFY(uk_fs_driver_##fsname))	\
```
mogasergiu, line 58: Interesting, so each driver declares its own section that will be part of `uk_fs_driver_list`. Is there a reason for this? :thinking:

mogasergiu (commented):
Very nice addition, can't wait to see first registered driver. In the meantime I had a brief read and only a few simple questions, since there is not much runtime code to look at **yet**.

lib/ukfs/include/uk/fs/driver.h
andreittr: The reason is so that linker magic will sort these structs by name in the array, allowing us to do binary search + strcmp.
The magic itself happens by sorting the drivers' sections at link time.

lib/ukfs/Makefile.uk
andreittr: > I don't see any other source file with - in the name in our codebase, feels inconsistent

[Plenty](https://github.com/unikraft/unikraft/blob/staging/lib/ukfile/file-nops.c) of [examples](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fd/include/uk/posix-fd.h) across [much](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/fd-shim.c) of the [new](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/fdio-impl.h) file/vfs [stack](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/include/uk/posix-fdio.h). This is following the same pattern.

> Also maybe this should have used the .lds.S suffix instead?

I'm not sure; [other](https://github.com/unikraft/unikraft/blob/staging/lib/ukalloc/libstats.ld) libs [also](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/clonetab.ld) use `.ld` for linker scripts, though `.lds.S` is also used in about as many places :shrug:

lib/ukfs/drivers.c
andreittr: I disagree, `ukfs` builds on top of `ukfile`, which by design always pulls in a libc and uses libc types; and if we're in a libc-enabled env anyway, I don't like mixing libc and `__` types in the same source file.
This dependency on libc types goes beyond primitives and into specific structs (e.g., `struct statfs` used directly by `ukfs`, with others used by `ukfile`) so it's not something easy to remove.

lib/ukfs/Makefile.uk
mogasergiu: > [Plenty](https://github.com/unikraft/unikraft/blob/staging/lib/ukfile/file-nops.c) of [examples](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fd/include/uk/posix-fd.h) across [much](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/fd-shim.c) of the [new](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/fdio-impl.h) file/vfs [stack](https://github.com/unikraft/unikraft/blob/staging/lib/posix-fdio/include/uk/posix-fdio.h). This is following the same pattern.

Introduced by you.. :sweat_smile: , ok sure, fine.

Ok so this was inconsistent from the very start:
ld:
```
./lib/ukdebug/trace.ld           
./lib/ukdebug/gdbtargetxml.ld    
./lib/vfscore/extra.ld           
./lib/uktest/tests.ld            
./lib/posix-socket/driver_list.ld
./lib/ukalloc/libstats.ld        
./lib/uknofault/excpttab.ld      
./lib/posix-process/clonetab.ld  
./lib/uksched/extra.ld           
./lib/uklibid/infosec.ld         
```
lds.S:
```
./plat/kvm/x86/link64.lds.S               
./plat/kvm/arm/link64.lds.S               
./plat/xen/x86/link64.lds.S               
./plat/xen/arm/link32.lds.S               
./plat/xen/arm/link64.lds.S               
./plat/common/bootinfo.lds.S              
./lib/uklibparam/libparam.lds.S           
./lib/syscall_shim/syscall_exittab.lds.S  
./lib/syscall_shim/syscall_entertab.lds.S 
./lib/ukboot/earlytab.lds.S               
./lib/ukreloc/reloc.lds.S                 
./lib/ukgcov/libukgcov.lds.S              
```
Ok, fine... can't help but feel this is very inconsistent.

andreittr:
Pushed revision of special file API.

andreittr:

andreittr:
Pushed minor change to `create` API.

lib/ukfs/include/uk/fs/driver.h
```diff
@@ -0,0 +1,46 @@
...
+	static const struct uk_fs_drv uk_fs_driver_##fsname = {	\
```
mogasergiu, line 59: This looks like a constant registration that stays constant. Any reason the driver list section is inserted after `.data` instead of `.rodata`?

andreittr: :eyes: good catch, it does make more sense to be after `.rodata`.
I must've uncritically copied the linker script of [posix-socket](https://github.com/unikraft/unikraft/blob/staging/lib/posix-socket/driver_list.ld#L10) and that indeed has non-const driver entries (_why_ that is eludes me though).

andreittr:

lib/ukfs/include/uk/fs/driver.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Retrieve the vopen function of the filesystem driver registered for the
+ * filesystem named `fstype`.
+ *
+ * @return
+ *  != NULL: vopen function of desired filesystem
+ *  == NULL: Filesystem name `fstype` not recognized
+ */
+uk_fs_vopen_func uk_fs_driver(const char *fstype);
```
mogasergiu, line 40: This is technically not retrieving the driver itself but its `vopen` function :thinking: . Why not `uk_fs_vopen_by_name` or something like that.

lib/ukfs/include/uk/fs/driver.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param fsname Filesystem name to register; do not enclose in quotes
+ * @param drv_vopen vopen function of filesystem driver
+ */
+#define UK_FS_DRIVER_REGISTER(fsname, drv_vopen)		\
+	__used __align(8)					\
+		__section("." STRINGIFY(uk_fs_driver_##fsname))	\
+	static const struct uk_fs_drv uk_fs_driver_##fsname = {	\
+		.fstype = STRINGIFY(fsname),			\
+		.vopen = drv_vopen				\
```
mogasergiu: Would it make sense to surround `drv_vopen` with paranthesis?

lib/ukfs/include/uk/fs.h
```diff
@@ -0,0 +1,591 @@
...
+#define UKFS_MODE_NOSEEK   010
+/* Driver handles synchronization internally */
+#define UKFS_MODE_NOIOLOCK 020
+
+/* Only filesystem nodes have `fsops` set */
+#define UKFS_FILE_ISFS(f) (!!(f)->fsops)
+/* Non-directories must not implement `listdir` */
+#define UKFS_FILE_ISDIR(f) (UKFS_FILE_ISFS(f) && (f)->fsops->listdir)
+/* Only symlinks may implement `readlink` */
+#define UKFS_FILE_ISSYM(f) (UKFS_FILE_ISFS(f) && (f)->fsops->readlink)
```
mogasergiu: I see multi-eval macro again where someone could pass a function call as an argument. Would it be worth making it a static inline instead?

mogasergiu (changes requested):

lib/ukfs/include/uk/fs.h
andreittr: Yeah, I guess static inlines would work just as well here and be more typesafe :+1: .

lib/ukfs/include/uk/fs/driver.h
andreittr: :thinking: yeah, looks like parantheses won't break anything here, must've been an oversight :+1:

lib/ukfs/include/uk/fs/driver.h
andreittr: In ukfs, at a very fundamental level, a driver _is_ its vopen function, and there is no distinction between the two -- we have no centralized ops table or anything, all driver functionality is encapsulated in the files it produces. `vopen` is the seed from which you get your first file. This is something I should probably insist on when doing the high-level doc pass over ukfile/ukfs :smile: .

Furthermore, at this very high abstraction layer the naming should best reflect the intentions of the caller -- to find the ukfs driver for `fstype` -- and be less about the details of _how_ a driver is represented.

If we ever want ukfs drivers to expose more metadata about themselves (and thus break the very strict driver-vopen equivalence mentioned above), we'll then define a
```c
struct uk_fs_driver_info {
    uk_fs_vopen_func vopen;
    /* misc metadata */
};
```
have drivers register using that, and change to
```c
const struct uk_fs_driver_info *uk_fs_driver(const char *fstype);
```
thus keeping an appropriate abstraction.
(We don't do this struct thing now to prevent an unnecessary extra indirection)

andreittr:
Pushed addressing comments.

andreittr:
Pushed minor change reordering the fsops struct to have the most commonly used operations at the start. API remains unchanged.

mogasergiu (approved):

andreittr:
Pushed update adding header `uk/fs/prio.h` to synchronize init/term prio among fs drivers & consumers thereof. Re-requesting review for just the additional header.

andreittr:

mogasergiu (approved):

andreittr:
Pushed very minor correction to the public API:
`vopen` and its function type is now declared in `uk/fs/driver.h`, as it it relevant only when talking about fs drivers:
- A driver implementing its vopen function
- A ukfs caller that wishes to look up a driver by name with `uk_fs_driver`

A caller only interested in filesystem operations (including the main `uk/fs.h` header) need not know or care about `vopen`.

andreittr:
Pushed update with 2 changes:
- added `exportsyms.uk` file to explicitly express developer intentions
- squashed the previous 2 commits into one, as the intermediate state would lead to a broken build with ukfs enabled

michpappas:

PR #1596, opened 2025-03-05
kconfig: Increase shell command output buffer size

### LINKED ISSUE #1588: kconfig `do_shell` reports invalid value on paths longer than 256. Fix available in kernel upstream.
### Describe the bug

When compiling with some deep sources paths, for example in yocto, `kconfig`  `do_shell` implementation can truncate long paths.

It has an hardcoded limit of `256` which might not be enough.

https://github.com/unikraft/unikraft/blob/d4394a1c7e5124633668c55ca5267c9914798019/support/kconfig/preprocess.c#L143

The kernel did introduce a fix in their `kconfig` implementation which could be backported and fixes the issue https://github.com/torvalds/linux/commit/8a4c5b2a6d8ea079fa36034e8167de87ab6f8880

### Steps to reproduce

Use a long directory name for all your sources, something that would make these `shell` calls return more than `256` characters.

https://github.com/unikraft/unikraft/blob/d4394a1c7e5124633668c55ca5267c9914798019/Config.uk#L43-L57

The generated value will be truncated and `source` will ignore the file.

### Expected behavior

Full path is handled.

_No response_

_No response_

### Relevant log output

```shell

```

shauryarane05:
### Description of changes
Backport upstream Linux kernel fix (commit 8a4c5b2) to increase the buffer size in do_shell from 256 to 4096 bytes. This fixes an issue where paths longer than 256 characters would be truncated.

The shell built-in is used to return paths in various Kconfig operations, and the previous buffer size was insufficient for deep directory structures, particularly in environments like Yocto builds.

Fixes: #1588

### Prerequisite checklist

 - [ x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [ x] Tested your changes against relevant architectures and platforms;
 - [ x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;

### Base target

### DIFF
--- a/support/kconfig/preprocess.c
+++ b/support/kconfig/preprocess.c
@@ -140,7 +140,7 @@ static char *do_lineno(int argc, char *argv[])
 static char *do_shell(int argc, char *argv[])
 {
 	FILE *p;
-	char buf[256];
+	char buf[4096];
 	char *cmd;
 	size_t nread;
 	int i;


shauryarane05:
solved the issue with the commit name

andreistan26:
Makes sense, maybe not diverging from the mainline would be better, but I don't see a reason that we would need that.

cladmi:
> Makes sense, maybe not diverging from the mainline would be better, but I don't see a reason that we would need that.

The shell commands in https://github.com/unikraft/unikraft/blob/30f46524e73327d19a9f51d8fc656259ba28f82d/Config.uk#L43-L57 do return a filepath.
Depending on the compilation environment depth, the value returned would be truncated without raising an error.

256 is easy to reach.

shauryarane05:
Thank you for reviewing my PR. I agree that the simple approach of increasing the buffer size to 4096 bytes is sufficient for this issue.
This also aligns with the upstream Linux kernel fix (commit 8a4c5b2) which took a similar approach.
Let me know if you'd like me to make any additional changes to the PR."

shauryarane05:
@razvand 
can this PR be merged?

cladmi:
I am not a maintainer.

PR #1600, opened 2025-03-12
lib/posix-process: Add per-process `sbrk`/`brk` implementations

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Implement support for per-process `sbrk`/`brk`. This feature has two
variants: one that uses libukvmem and one that alternatively simply
allocates a big enough area from the global allocator.

In both cases, since we have to track per process the brk base and
its current position this makes the feature depend on there being an
associated `struct posix_process` with the thread upon initialization.

Additionally, one is able to choose the size of the backing memory area
by configuring `CONFIG_LIBPOSIX_PROCESS_BRK_PAGE_ORDER` (default 2MiB).

Another important note is that, by default, we adopt the behavior of
glibc with MORECORE_CLEARS == 2, meaning that whenever there is a bump
through (s)brk, we zero out that difference.
This increment zeroing behavior is configurable through
`LIBPOSIX_PROCESS_BRK_INC_MEMZERO`.

For the non-ukvmem variant we simply request a big enough area from the
global allocator and just move the brk cursor around based on requests,
while taking care to zero out on bumps.

For the ukvmem variant, we initialize an anonymous, demand mapped, VMA.
Unlike the previous variant, this one does not interfere with the global
allocator.

As far as maintaining `MORECORE_CLEARS == 2` behavior goes, every newly
faulted page will be guaranteed to be zeroed out by the page fault
handler due to the VMA not having the UK_VMA_MAP_UNINITIALIZED flag.
On one hand, this helps us in not having to memset big, page boundary
crossing, bumps all at once and, instead, have the page fault handler
zero them out as they are being used. On the other hand, we must still
manually take care of in-page bumps: if the bump does not cross a page
boundary, it will not fault and thus not be zeroed out, forcing us to do
it ourselves.

What is more, the ukvmem variant is able to free up pages upon brk
decrement requests, if the difference is a multiple of page size.
This behavior is configurable through
`LIBPOSIX_PROCESS_BRK_DEC_DONTNEED`.
This is POSIX-compliant as POSIX specifies that the pages upon
reincrement have undefined values but remains optional as it is not
Linux-compatible since there the pages do not get released and thus
reincremented pages would keep previous data (although some libcs can
do manual zeroing).

Lastly, add the declarations/signatures for `brk`/`sbrk`.

## WARNING++: This removes/adds the `(s)brk` configurations and will ruin all catalog builds.


mogasergiu:
Rebase force push

lib/posix-process/Config.uk
```diff
@@ -26,6 +26,23 @@ config LIBPOSIX_PROCESS_SIGNAL
...
+	bool "brk() system call and sbrk() libc function"
+	select LIBPOSIX_PROCESS_PIDS
+	select LIBPOSIX_PROCESS_INIT_PIDS
+	select LIBUKALLOC
+	imply LIBUKVMEM
+
+if LIBPOSIX_PROCESS_BRK
+
+config LIBPOSIX_PROCESS_BRK_NPAGES
+	int "Heap size for brk (number of pages)"
```
michpappas: Since we allocate pages, perhaps it would make more sense to use page order to be consistent with Kconfig options of stacks.

lib/posix-process/Config.uk
```diff
@@ -26,6 +26,23 @@ config LIBPOSIX_PROCESS_SIGNAL
 	select LIBSYSCALL_SHIM
 	select LIBUKLOCK
 
+menuconfig LIBPOSIX_PROCESS_BRK
+	bool "brk() system call and sbrk() libc function"
+	select LIBPOSIX_PROCESS_PIDS
```
michpappas: These need to be updated because of the changes we made in `0.20.0`: 
* `CONFIG_LIBPOSIX_PROCESS_PIDS`: Replace with `CONFIG_LIBPOSIX_PROCESS_MULTITHREADING`. 
* `CONFIG_LIBPOSIX_PROCESS_INIT_PIDS`: Remove.

lib/posix-process/Config.uk
```diff
@@ -26,6 +26,23 @@ config LIBPOSIX_PROCESS_SIGNAL
...
+	select LIBPOSIX_PROCESS_INIT_PIDS
+	select LIBUKALLOC
+	imply LIBUKVMEM
```
michpappas: Not sure whether we should have this `imply LIBUKVMEM` here because of the problem that the only way to unselect these options is manually. Generally we should probably reserve `imply` for platform configs. In practice we have `imply` in a few more places (ukboot, uknetdev, posix-fdio) but most likely these shouldn't be there either.

lib/posix-process/process.c
```diff
@@ -272,6 +272,14 @@ int uk_posix_process_create(struct uk_alloc *a,
 	}
 #endif /* CONFIG_LIBPOSIX_PROCESS_SIGNAL */
 
+#if CONFIG_LIBPOSIX_PROCESS_BRK
```
michpappas: Now that we raise the clone event for PID1 (b4eeed0a9443efea76b5974fc62aec515c2a03dd), the initialization & termination of brk can be moved into a `POSIX_PROCESS_CLONE_HANDLER()` in `brk.c`. That will also work well if we eventually move brk into its own library. I am also planning to remove signal initialization / termination from `process.c` too, so that we limit `process.c` to very core implementation of pprocess / pthread.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+	rc = pprocess_brk_free_zone(pprocess);
+	if (unlikely(rc))
+		UK_CRASH("Failed to free brk space (%lu KiB).\n",
+			 (__u64)HEAP_LEN / 1024);
+
+	pprocess->brk_ctx.pos = 0;
+	pprocess->brk_ctx.base = __NULL;
+
+	uk_pr_debug("Freed brk heap region: %p-%p\n",
+		    pprocess->brk_ctx.base,
```
michpappas: UAF: `pprocess->brk_ctx.base` was set to `__NULL` right already. This debug message be moved up, before `free()`.

While we're there, some parts of this file use `NULL`, perhaps the implementation should be consistent and use one or the other.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+	if (!IN_RANGE(addr_arg, base, HEAP_LEN)) {
+		uk_pr_debug("0x%lx outside PID %d's brk range 0x%lx-0x%lx, keep brk @ 0x%lx\n",
+			    addr_arg, pprocess->pid,
+			    base, base + HEAP_LEN, cur_brk);
+		return (void *)cur_brk;
+	}
+
+	rc = do_brk_inrange(cur_brk, addr_arg);
+	if (unlikely(rc))
+		return NULL;
```
michpappas: I think `NULL` is not a legit return value for the `brk()`, at least the man only mentions the current break:

> C library/kernel differences
> 
> The return value described above for brk() is the behavior provided by the glibc wrapper function for the Linux brk() system call.  (On most  other  implementations,  the  return value  from  brk()  is the same; this return value was also specified in SUSv2.)  However, the actual Linux system call returns the new 
program break on success.  On failure, the system call returns the current break.  The glibc wrapper 
function does some work (i.e., checks whether the new break is less than addr) to provide the 0 and -1 return values described above.

If that is the case indeed, the check for the return value in the libc wrapper of brk and sbrk need to be updated too, as both currently checks `!rc`.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+UK_LLSYSCALL_R_DEFINE(void *, brk, void *, addr)
+{
+	struct posix_process *pprocess;
+	__uptr base, cur_brk, addr_arg;
+	int rc;
+
+	pprocess = uk_pprocess_current();
+	UK_ASSERT(pprocess);
+
+	base = (__uptr)pprocess->brk_ctx.base;
```
michpappas, line 265: As `brk_ctx.pos` can be udpated, I think the implementation needs to protect against the case where > 1 threads in the process call `brk()`.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+
+	return addr;
+}
+
+#if UK_LIBC_SYSCALLS
+int brk(void *addr)
+{
+	long rc;
+
+	rc = uk_syscall_r_brk((long)addr);
```
michpappas: The syscall takes `void *`, so I wonder why the cast is needed.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return 0;
+}
+
+#if CONFIG_LIBNOLIBC
+void *sbrk(intptr_t inc)
+{
+	struct posix_process *pprocess;
+	long req_base, rc;
```
michpappas: `brk()` takes `void *`, so I wonder why `req_base` is declared as `long`.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+	void *prev_base;
+
+	pprocess = uk_pprocess_current();
+	UK_ASSERT(pprocess);
+
+	/* We are increasing or reducing our range */
+	prev_base = (__u8 *)pprocess->brk_ctx.base + pprocess->brk_ctx.pos;
```
michpappas: `prev_base` should probably be `prev_brk`, as it carries the previous break value, not the base. Similarly `req_base` should be `req_brk`.

michpappas: Nit: Although casting to `__u8 *` is valid, the implementation of the `brk()` syscall casts to `uptr_t` for arithmetic. It would be nice if there was some consistency across the file.

lib/posix-process/brk.c
```diff
@@ -0,0 +1,268 @@
...
+	 * the backing VMA does not have the UK_VMA_MAP_UNINITIALIZED flag which
+	 * means that every new page we fault on will be zeroed out on demand
+	 * instead of us having to explicitly memset the difference every
+	 * time. This wouldn't be achievable if we didn't free the decrements
+	 * on page boundaries.
+	 *
+	 * See glibc's MORECORE_CLEARS. We go with MORECORE_CLEARS == 2 to
+	 * be on the safe side.
+	 */
+	if (addr >= cur_brk) {
```
michpappas: Should that be `addr > cur_brk`? Perhaps this function should not be entered in the first place if `addr` == `cur_brk`.

michpappas (commented):
Hi @mogasergiu I think part of the implementation is just migrated from app-elfloader, but I took the opportunity to make a full review. Comments below.

mogasergiu:
Rebase force push.

lib/posix-process/brk.c
mogasergiu: The signatures created by the syscall define macros have all of their arguments as `long`.

lib/posix-process/brk.c
mogasergiu: See https://github.com/unikraft/unikraft/pull/1600#discussion_r2702526694

lib/posix-process/brk.c
mogasergiu: This is generally more of an SMP worry-point. However, it does no harm so I added it. There may be a problem in the future w.r.t. ukvmem so I added a comment/TODO about it.

michpappas (approved):
Thanks! 🚀 

PR #1606, opened 2025-03-21
lib/uksparsebuf: Introduce sparse buffer library

andreittr:
### Description of changes

This change introduces uksparsebuf, a core library providing a data structure to represent a buffer of indeterminate size consisting of fragmented allocated and sparse areas.

The core sparse buffer API allows swapping out implementations at config time while remaining source-compatible to its callers.
Along with the core API, this change introduces a basic implementation using rank-balanced trees.

#### FAQ:
**Q**: Why is the embedded head pointer interface so baroque?
**A**: Writing an interface that is (1) source-compatible across any reasonable implementation and (2) does not introduce overhead compared to doing it "natively" isn't trivial. This is the best I could come up with.

**Q**: Why multiple implementations though? Aren't RBTrees good enough?
**A**: RBTrees serve the purpose decently and require little bespoke code, but do suffer from poor locality and high memory overhead and churn. It works well for a basic correct implementation that more performant future ones can be compared and validated against.

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/uksparsebuf/include/uk/sparsebuf.h
```diff
@@ -0,0 +1,530 @@
...
+static
+int uk_sparsebuf_lookup(struct uk_sparsebuf_blk **headp, __u64 pgoff,
+			struct uk_sparsebuf_cur *cur);
+
+/**
+ * For-like macro to iterate over all slices in a sparse buffer.
+ *
+ * The sparse buffer must not be modified over the course of the iteration.
+ */
+#define UK_SPARSEBUF_FOREACH(headp, cur) for (				\
```
mogasergiu: Can we align this foreach with every other foreach we have and place this `for (` on the line below the `#define` line?

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,320 @@
...
+
+/* Common utils & ops for sparsebuf implementations.
+ *
+ * Do not include in more than one source file per build.
+ */
+
+#ifndef __UK_SPARSEBUF_IMPL_COMMON_H__
+#define __UK_SPARSEBUF_IMPL_COMMON_H__
+
+/* We define these errnos here so as not to pull in a full libc */
```
mogasergiu, line 15: What is wrong about including `errno.h`?

lib/uksparsebuf/include/uk/sparsebuf/util.h
```diff
@@ -0,0 +1,123 @@
...
+ * @return
+ *   == 0: Success
+ *    < 0: Negative error code
+ */
+static inline
+int uk_sparsebuf_truncate(const struct uk_sparsebuf_ctx *ctx,
+			  struct uk_sparsebuf_blk **headp,
+			  __sz oldsz, __sz newsz)
+{
+	if (newsz < oldsz) {
```
mogasergiu: Maybe you can save some indentation by returning early if `newsz >= oldsz`?

lib/uksparsebuf/include/uk/sparsebuf/util.h
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Zero out `len` bytes starting at sparse buffer offset `off`, using `cur`.
+ */
+static inline
+void uk_sparsebuf_memclear(struct uk_sparsebuf_cur *cur, __sz off, __sz len)
+{
+	char *buf = __NULL;
+	__sz blen = uk_sparsebuf_memat(off, cur, &buf);
+
+	if (blen && buf) {
```
mogasergiu, line 29: Return early and save indentation?

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+		if (unlikely(r))
+			return r;
+	}
+	/* Walk through slices ensuring filled/refd */
+	while (blk && npages) {
+		struct uk_sparsebuf_blk *next = blk;
+		struct uk_sparsebuf_slice *sl = &blk->sl;
+		__u32 np;
+
+		if (uk_sparsebuf_pg_before(pgoff, sl)) {
```
mogasergiu, line 278: Just a question, but would it be possible/easy to move this big chunk from the while loop in its own separate inline function? :thinking: I believe it could help with readability.

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * Free the buffer of slice `sl` if applicable, using context `ctx`.
+ */
+static inline
+void sparsebuf_slice_buffree(const struct uk_sparsebuf_ctx *ctx,
+			     const struct uk_sparsebuf_slice *sl)
+{
+	if (UKSB_SLICE_SHOULD_FREE(sl))
+		ctx->funcs->alloc_free(sl->buf, sl->pgoff, sl->npages,
```
mogasergiu, line 73: Another opportunity to save up on indentation? :D

lib/uksparsebuf/include/uk/sparsebuf.h
```diff
@@ -0,0 +1,530 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Sparse buffers
```
mogasergiu, line 7: Any reason this multi-line comment is inconsistent with the others as in it does not begin with `/**`?

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,320 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Common utils & ops for sparsebuf implementations.
```
mogasergiu, line 7: Ditto (actually there are more). Genuinely just a question, is there a convention we have for this?

lib/uksparsebuf/Config.uk
```diff
@@ -0,0 +1,16 @@
...
+	bool "uksparsebuf: Sparse page buffers"
+	select LIBUKALLOC
+
+if LIBUKSPARSEBUF
+
+choice LIBUKSPARSEBUF_IMPL
+	prompt "Implementation"
+	default LIBUKSPARSEBUF_IMPL_RBTREE
+
+config LIBUKSPARSEBUF_IMPL_RBTREE
```
mogasergiu, line 11: So it won't be possible to have, for example, two different instances of sparsebuffers with two different implementations backing implementations?

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+	/* Shorten original slice */
+	sl->npages = lnp;
+	/* Populate new slice */
+	blk->sl = (struct uk_sparsebuf_slice){
+		.pgoff = pgoff,
+		.external = sl->external,
+		.npages = rnp,
+		.refcnt =
+			UK_REFCOUNT_INITIALIZER(uk_refcount_read(&sl->refcnt)),
+		.buf = sl->buf + lnp * PAGE_SIZE
```
mogasergiu, line 123: Nit, but it is generally a good convention to end static initializations of structs with `,` :thinking:

mogasergiu (commented):
LGTM. I only have a few nitpicks. Switching now to ukpod :)

lib/uksparsebuf/include/uk/sparsebuf.h
andreittr: Sure thing; I had written it from scratch and had no idea we had a house style for foreaches.

lib/uksparsebuf/impl-common.h
andreittr: Because I don't want a kernel data structure lib to pull in libc.

I believe this highlights the need for us to either have a "types-only libc" that's even more minimalist than nolibc, or to actually enforce some types/constants ourselves that ported libcs then make use of, rather than the other way around.

lib/uksparsebuf/include/uk/sparsebuf.h
andreittr: `/**` open docstrings that are in the familiar doxygen format. Normal, run-of-the-mill comments (whether single- or multi-line) should always start with `/*`. Most editors also distinguish the two when highlighting as they serve fundamentally different purposes (function documentation vs. comments).

lib/uksparsebuf/Config.uk
andreittr: Not at the moment no; I don't foresee that being very useful, and I'd rather not pay the cost of indirection upfront. This can change if we decide we need it.

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,320 @@
...
+#ifndef ENOMEM
+#define ENOMEM 12
+#endif /* ENOMEM */
+
+#ifndef EBUSY
+#define EBUSY 16
+#endif /* EBUSY */
+
+/* We should free all memory that is not externally assigned */
+#define UKSB_SLICE_SHOULD_FREE(sl) (!((sl)->external))
```
mogasergiu: Isn't it a bit inconsistent how most other macro's have `UK_` with an underscore and this one doesn't? :thinking: 
This is also the case for like `UKCPIO` macro's or `ukarch_` API but oh well :man_facepalming:

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,320 @@
...
+	struct uk_sparsebuf_cur tmp = *cur;
+
+	while (!uk_sparsebuf_finished(&tmp)) {
+		struct uk_sparsebuf_slice *sl = uk_sparsebuf_slice_at(&tmp);
+
+		if (UKSB_START_OR_BEFORE(end, sl))
+			break;
+		if (!UKSB_SLICE_CAN_RELEASE(sl))
+			return -1;
+		uk_sparsebuf_next(&tmp);
```
mogasergiu: To me it feels awkward how `uk_sparsebuf_next` doesn't return anything it just doesn't do anything if it can't. At least if it were named something like `uk_sparsebuf_try_next` or something like that? To me it would make more sense if it returned `__NULL` or the next entry.

Additionally, I think maybe such `while` loops could look better as a `uk_sparsebuf_foreach_slice` macro?

lib/uksparsebuf/include/uk/sparsebuf.h
```diff
@@ -0,0 +1,530 @@
...
+	return pg > sl->pgoff && pg < uk_sparsebuf_slice_pgend(sl);
+}
+
+/**
+ * Return non-zero if the page frame at `pg` lies entirely within `sl`.
+ */
+static inline
+int uk_sparsebuf_pg_within(__u64 pg, const struct uk_sparsebuf_slice *sl)
+{
+	return uk_sparsebuf_pg_atstart(pg, sl) || uk_sparsebuf_pg_mid(pg, sl);
```
mogasergiu: Note that we also have `IN_RANGE` macro as well as others that might help

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+	};
+	/* Insert after original */
+	UK_RB_INSERT_NEXT(sparsebuf_map, map, cur->p, blk);
+	return 0;
+}
+
+static
+int sparsebuf_separate(struct uk_sparsebuf_blk **headp,
+		       __u64 pgoff, __u32 npages,
+		       struct uk_sparsebuf_cur *cur, struct uk_alloc *alloc)
```
mogasergiu, line 133: This feels a bit weird how both `cut` and `separate` receive dedicated `alloc` arguments but when freed they use the `ctx->alloc`. Why not pass the `ctx` as argument? It's a bit as if saying that somewhere there is a guarantee that this allocator is that of `ctx`. :thinking:

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+			 struct uk_sparsebuf_blk **headp,
+			 struct uk_sparsebuf_cur *cur, __u64 pgend)
+{
+	/* The compile-time assert above ensures we can cast this safely. */
+	struct sparsebuf_map *map = (struct sparsebuf_map *)headp;
+	struct uk_sparsebuf_blk *blk;
+	struct uk_sparsebuf_blk *tmp = cur->p;
+
+	/* Free slices in collapse region */
+	UK_RB_FOREACH_FROM(blk, sparsebuf_map, tmp) {
```
mogasergiu, line 167: `UK_RB_FOREACH_FROM` I was just about to comment why not use the `SAFE` variant only to then see that this does the same thing as the safe variant behind the scenes. A bit inconsistent with other API's, for example `uk_list_for_each_entry_from` and `uk_list_for_each_entry_safe_from`.

lib/uksparsebuf/include/uk/sparsebuf.h
```diff
@@ -0,0 +1,530 @@
...
+ *
+ * @param ctx Caller context
+ * @param headp Head pointer of sparse buffer to operate on
+ */
+void uk_sparsebuf_clear(const struct uk_sparsebuf_ctx *ctx,
+			struct uk_sparsebuf_blk **headp);
+
+/**
+ * Assign the memory region of `npages` starting at `buf` to be the backing
+ * for the sparse file page offset `pgoff`, replacing whatever had been there.
```
mogasergiu: s/file/buffer ?

lib/uksparsebuf/include/uk/sparsebuf/util.h
```diff
@@ -0,0 +1,123 @@
...
+ * @return
+ *   == 0: Success
+ *    < 0: Negative error code
+ */
+static inline
+int uk_sparsebuf_punch_hole(const struct uk_sparsebuf_ctx *ctx,
+			    struct uk_sparsebuf_blk **headp,
+			    __sz off, __sz len)
+{
+	const __sz pglo = ALIGN_UP(off, PAGE_SIZE);
```
mogasergiu, line 102: Should this be `__PAGE_SIZE` instead?

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: The loop could use a small refactor, yeah; it's seen its fair share of changes during dev while keeping the original structure.

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+	/* No failures beyond this point */
+	/* Shorten original slice */
+	sl->npages = lnp;
+	/* Populate new slice */
+	blk->sl = (struct uk_sparsebuf_slice){
+		.pgoff = pgoff,
+		.external = sl->external,
+		.npages = rnp,
+		.refcnt =
+			UK_REFCOUNT_INITIALIZER(uk_refcount_read(&sl->refcnt)),
```
mogasergiu: What if the actual concurrent ref holder is looking at the first split half but you initialize the second half with the same refcount value, although they can't have the same refcount at the same time right?

What I am mainly saying, is it really correct to initialize the refs of both split chunks to be the same? When other holders would release their hold they will end up  releasing the refcount of the original slice although in fact they may have acquired it just to use the part of it that would have been the other half.

lib/uksparsebuf/include/uk/sparsebuf/util.h
andreittr: The computations below are only well-defined if `newsz < oldsz`, I very much like const correctness, and style demands declarations before statements; my hands are tied. I'll just move the `truncate_down` op into its own inline and remove this one (it's kinda undefined what "growing" a sparse buffer even means -- they're _always_ the size of the a `__sz` addr space -- so a general-purpose `truncate` makes little sense)

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,384 @@
...
+			struct uk_sparsebuf_blk **headp,
+			__u64 pgoff, __u32 npages, void *buf)
+{
+	/* The compile-time assert above ensures we can cast this safely. */
+	struct sparsebuf_map *map = (struct sparsebuf_map *)headp;
+	const __u64 pgend = pgoff + npages;
+	struct uk_sparsebuf_cur cur;
+	struct uk_sparsebuf_blk *blk;
+	int r;
+
```
mogasergiu, line 396: Perhaps an assertion `UK_ASSERT(PAGE_ALIGNED(buf))`

lib/uksparsebuf/impl-common.h
andreittr: Yeah, I didn't invest too much time in naming here, as these aren't suppose to be "exported" per-se; yes, they are in a header, but no public code should use them. And since prefixing macros with underscore(s) is problematic, the `UK*` namespace is where I ended up.

lib/uksparsebuf/include/uk/sparsebuf.h
andreittr: Might be worth considering; still not 100% happy with the naming here, so I might do a pass.

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: I chose to restrict their arguments to better convey what bits&pieces of `ctx` they're allowed to use; e.g., neither `cut` nor `separate` are allowed to call any buffer management funcs, only do metadata allocs. Never passing them the mgmt funcs to begin makes this very explicit. It's part of how sparsebuf implementations work together with `impl-common.h` and is thus a very tightly coupled interface.
If it were a public API I would naturally use `ctx` everywhere (as is the case for the actual public API), but here I value clean separation of responsibilities a lot more.

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: I don't think it's needed, nor strictly speaking correct -- all parts of the sparsebuf API and impl don't _really_ care about the backing buffer's aligment, only that it's (at least) a multiple of `PAGE_SIZE` long.

This is on purpose, you could very well have the buf alloc functions grab memory from `malloc` and everything works perfectly fine (provided they do the needed song&dance to support partial frees/drops (or never bother with free at all)). If a caller wants page-alignment, they ensure it themselves through appropriate alloc functions.

Edit: I just noticed the `assign` docstring mentioned page-alignment as requirement for buffers; will update to remove.

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: Slice refcounts are not meant to be manipulated independently from using `fill` and `ref_release`, which the caller themselves is responsible for synchronizing.
This, along with the fact that `cut` is allowed to change the data structure, but not what the data structure _represents_, means we have to assign the same refcount since the two slices after the split represent the same region of the sparsebuffer, with the same reference count.

But now that you mention it, I realise there's no need for actual atomic refcounts in slices and I can replace this with just an `int` or something to save on atomic ops. Will change in pass.

lib/uksparsebuf/include/uk/sparsebuf/util.h
andreittr: `uksparsebuf` uses `<uk/arch/paging.h>` and `PAGE_SIZE` seems to be what it defines. See related comment on https://github.com/unikraft/unikraft/pull/1607#discussion_r2033093601

lib/uksparsebuf/include/uk/sparsebuf.h
andreittr: Oh yeah, thanks

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: Blame it on the [freebsd code](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/tree.h#L1037) we imported our RBtrees from.
Keeping diff to a minimum helps us pull in future changes from upstream, I don't think we should harmonize.

lib/uksparsebuf/impl-common.h
andreittr: Name aside, you're not really supposed to call `next` on a cursor at the end of the buffer and expect it to do anything. Originally I had it assert & crash, but that meant a lot of messy extra checks in loops. Ultimately I decided that a no-op next makes calling code cleaner. As to why it modifies `cur` in place, it's a safer solution than passing an arbitrarily-sized struct around by value. I could make it return an `int` if it made progress, but that would make writing for-loops with a single loop variable trickier. Thus the current signature.

As for the name, yeah, I agree `next` is more suggestive of it returning something and a bit misleading; I'll change it to something like `advance` or `step`.

> Additionally, I think maybe such while loops could look better as a uk_sparsebuf_foreach_slice macro?

I need to start at a particular `cur`, else I'd've used `UK_SPARSEBUF_FOREACH`. But a good argument for a `UK_SPARSEBUF_FOREACH_FROM` macro.

lib/uksparsebuf/impl-common.h
andreittr: Not at the expense of clarity, mental control-flow and linecount (in that order) IMO.
I see early `return` in a `void` function akin to a `goto`: only to be used sparingly and if every other option is much uglier.

lib/uksparsebuf/include/uk/sparsebuf/util.h
andreittr: See other comment about returning early from `void`. Also, here it helps make our compliance with the `memat` API very obvious: only if `blen` AND `buf` (in that order) are we meant to touch `buf[0:blen]`.

Which reminds me, I need to see if I can remove that `= __NULL` initialization of `buf`. It's supposed to remain untouched/uninited for `blen == 0` on purpose, and we want the compiler to warn you about it if you do use it. Though I'm not sure how much insight it has into `memat`.

andreittr:
Pushed addressing comments.

andreittr:

lib/uksparsebuf/sparsebuf-rbtree.c
mogasergiu: (Yeah, that was why I was asking)

andreittr:
Pushed minor code cleanup.

lib/uksparsebuf/sparsebuf-rbtree.c
```diff
@@ -0,0 +1,434 @@
...
+		}
+
+		/* No failures beyond this point */
+		/* Free range */
+		sparsebuf_freerange(ctx, headp, &cur, pgend);
+	}
+	/* Insert new slice w/ gifted mem */
+	blk->sl = (struct uk_sparsebuf_slice){
+		.pgoff = pgoff,
+		.external = -1,
```
mogasergiu: Why `-1` instead of `1`, wouldn't it be the same result? This is a 1-bit bitfield, it looks a bit odd setting to `-1`, the docs just say it is either 0 or != 0.

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,337 @@
...
+#ifndef EBUSY
+#define EBUSY 16
+#endif /* EBUSY */
+
+/* We should free all memory that is not externally assigned */
+#define UKSB_SLICE_SHOULD_FREE(sl) (!((sl)->external))
+/* We can release only if there are no held refcounts */
+#define UKSB_SLICE_CAN_RELEASE(sl) (!((sl)->refcnt))
+
+#define UKSB_START_OR_BEFORE(pg, sl) \
```
mogasergiu: Would it be better to have this as a static inline instead? For example at line 240/280 the `UKSB_START_OR_BEFORE` assertions would call the `slice_at` twice (yes i know it's just an inline getter, but still).

lib/uksparsebuf/include/uk/sparsebuf/util.h
```diff
@@ -0,0 +1,124 @@
...
+#include <uk/sparsebuf.h>
+
+/**
+ * Zero out `len` bytes starting at sparse buffer offset `off`, using `cur`.
+ */
+static inline
+void uk_sparsebuf_memclear(struct uk_sparsebuf_cur *cur, __sz off, __sz len)
+{
+	char *buf = __NULL;
+	__sz blen = uk_sparsebuf_memat(off, cur, &buf);
```
mogasergiu, line 27: Just a heads up, this `uk_sparsebuf_memat` is also inlined and called in several places (2 more in ukfs-ramfs), just FYI, cus `memclear` is also inlined and called in several places as well (e.g. twice in `uk_sparsebuf_punch_hole`).

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: I had of course at first done that, but found out that since it's a single-bit _signed_ integer, its only two valid values are `0` and `-1` (it makes a lot of sense once you think about it). If using `1` the compiler will annoy you with a warning that "hey, that 1 you set there actually overflows into -1, just a heads-up"; setting to -1 is simpler than selectively silencing an otherwise useful warning.

lib/uksparsebuf/impl-common.h
andreittr: Yeah, good point, this is left over from when all slice checking was macros.

lib/uksparsebuf/include/uk/sparsebuf/util.h
andreittr: Yes, I am aware, do you suggest that this inlining is overoptimization and would be better served as a linked function call? Because, unless we're aiming for code size first, I'm not sure. Either of these functions are intended to be part of larger operations, and any multiple calls from the same codebase (e.g., in ramfs) should happen in unrelated code paths (e.g., different FS ops in ramfs). In that case I believe we do want the benefit of local optimizations vs. code compactness.

I agree both of these aren't trivial inlines, containing a loop and possible func calls, but they're not intended to be called "in the loop", rather they're meant to "be the loop", if that makes sense :smile: . In a hackier, less typesafe world I'd've made them macros.

mogasergiu: Nope I am not suggesting anything. Just making an observation in case you lost track of this as you were writing the rest of the code. If you think it's fine then ok, it's not used in that many places anyway.

lib/uksparsebuf/include/uk/sparsebuf.h
```diff
@@ -0,0 +1,613 @@
...
+ * by the sparse buffer, or provided by the caller using `assign`. This latter
+ * case is signaled by the `external` flag.
+ */
+struct uk_sparsebuf_slice {
+	/* Page offset in sparse buffer that this region maps */
+	__u64 pgoff;
+	struct {
+		/* != 0 if caller-assigned, 0 if automatically allocated */
+		char external:1;
+		/* Number of contiguoous pages that this region spans */
```
mogasergiu: `s/contiguoous/contiguous` :ghost:

lib/uksparsebuf/sparsebuf-rbtree.c
andreittr: That said, and taking a closer look, maybe it's smarter to silence this warning by making `.external` explicitly an `unsigned char` :thinking: . Not sure if I've tried that before, will try in the pass.

edit: Yeah `unsigned char external:1` shuts the compiler up. I'll update to that.

andreittr:
Pushed addressing comments.

andreittr:
Pushed minor docstring change to reflect reality.

lib/uksparsebuf/impl-common.h
```diff
@@ -0,0 +1,342 @@
...
+#endif /* ENOMEM */
+
+#ifndef EBUSY
+#define EBUSY 16
+#endif /* EBUSY */
+
+/* We should free all memory that is not externally assigned */
+#define UKSB_SLICE_SHOULD_FREE(sl) (!((sl)->external))
+/* We can release only if there are no held refcounts */
+#define UKSB_SLICE_CAN_RELEASE(sl) (!((sl)->refcnt))
```
mogasergiu: At this point I think we should static inline these as well :man_shrugging:

mogasergiu (approved):
LGTM, just a minor comment.

andreittr:
Pushed update addressing comment & adding `exportsyms.uk` file to explicitly express developer intent.

michpappas:

PR #1607, opened 2025-03-21
lib/ukpod: Introduce Pages-on-Demand library

andreittr:
### Description of changes

This change introduces ukpod, a core library providing an internal API and backends for managing virtual memory areas with file-like mmap semantics, ensured through custom callbacks.
This PR contains the core ukpod API along with a fallback backend that eagerly allocates pages.

~~Depends on #1576 and has it included as squashed commit; please ignore during review.~~
~~Squashed commit will be removed on merge of #1576 and subsequent rebase.~~

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/ukpod/include/uk/pod/types.h
```diff
@@ -0,0 +1,103 @@
...
+/* Common PoD types & callback API */
+
+#ifndef __UK_POD_TYPES_H__
+#define __UK_POD_TYPES_H__
+
+#include <uk/arch/paging.h>
+#include <uk/arch/types.h>
+#include <uk/essentials.h>
+
+#define UK_POD_BAD_PHYSADDR ((__paddr_t)-1)
```
mogasergiu: Note, we also have `ukarch_paddr_isvalid` and `__PADDR_INV`.

lib/ukpod/Config.uk
```diff
@@ -0,0 +1,45 @@
...
+	depends on LIBUKPOD_DEFAULT
+	default LIBUKPOD_DEFAULT_EAGER
+
+config LIBUKPOD_DEFAULT_EAGER
+	bool "Eager page allocator"
+	select LIBUKPOD_EAGER
+	help
+		Fallback implementation that eagerly allocates RAM.
+
+config LIBUKPOD_DEFAULT_EXTERNAL
```
mogasergiu, line 20: So there can be only one external implementation at a time?

lib/ukpod/anon.c
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Page-in function for anonymous memory; fills pages with zeros.
+ */
+__ssz uk_pod_anon_pagein(void *addr, __sz npages, __paddr_t pa __unused,
+			 void *arg __unused, __sz pgoff __unused)
+{
+	char *__align(PAGE_SIZE) p;
+
+	UK_ASSERT(PAGE_ALIGNED((__uptr)addr));
```
mogasergiu, line 23: Can you please leave a more detailed comment as to why you chose to do it this way? I.e. how you noticed that doing this case with our particular way of building yields in better results?

lib/ukpod/include/uk/pod/types.h
andreittr: Good call, will use those and rm duplication. (+ I don't want ukpod to take on the responsibility of ensuring `((__paddr_t)-1)` is indeed not a valid paddr)

lib/ukpod/Config.uk
andreittr: There can be only one system PoD per config, internal or external; that's by design, it's the one backend guaranteed to be available by `select LIBUKPOD_DEFAULT`, and the only one that does not take an explicit "state" argument in its API (the system PoD state is implicitly kept by the lib implementing it).

If a caller wishes to use a specific PoD backend it can always `#include` its specific headers and create an instance (== backend-specific state) which it can freely (and privately) use. Whether the backend is one provided by `libukpod`, or any number of external libs doesn't matter.

andreittr:
Pushed addressing comments

lib/ukpod/eager.c
```diff
@@ -0,0 +1,111 @@
...
+static
+__ssz loop_pgin(uk_pod_pgin_func func,
+		char *addr, __sz npages, void *arg, __sz base)
+{
+	__sz done = 0;
+	__ssz r;
+
+	while (done < npages) {
+		r = func(addr, npages - done, __PADDR_INV, arg, base + done);
+		if (unlikely(r < 0))
```
mogasergiu, line 28: What if you did this as `if (unlikely(r <= 0))` and then assert for `UK_ASSERT(r < 0)` inside the if error's scope? Ditto `loop_pgwb`.

lib/ukpod/eager.c
```diff
@@ -0,0 +1,111 @@
...
+			return r;
+		UK_ASSERT(r > 0); /* func must make progress or error out */
+		done += r;
+		addr += r * PAGE_SIZE;
```
mogasergiu, line 32: Should we use `__PAGE_SIZE` instead? Ditto for the rest of this commit of course.

lib/ukpod/eager.c
```diff
@@ -0,0 +1,111 @@
...
+			 void *arg, __sz base)
+{
+	__ssz r;
+	void *buf = uk_palloc(al, npages);
+
+	if (unlikely(!buf))
+		return ERR2PTR(-ENOMEM);
+	UK_ASSERT(PAGE_ALIGNED((__uptr)buf));
+	r = loop_pgin(ops->pagein, buf, npages, arg, base);
+	if (unlikely(r < 0)) {
```
mogasergiu, line 68: What if you do `if (unlikely((__sz)r != npages))` and then you `UK_ASSERT(r < 0)` within the error case's scope? Ditto for the other similar ones below.

lib/ukpod/eager.c
andreittr: I believe `PAGE_SIZE` is the macro [exported](https://github.com/unikraft/unikraft/blob/staging/include/uk/arch/paging.h#L306) by `<uk/arch/paging.h>`, which ukpod globally uses (through `<uk/pod/types.h>`); I don't think it's a case of libc-vs-kernel types.

lib/ukpod/eager.c
andreittr: Actually, since `loop_pgin`/`_pgwb` are local static functions, it makes little sense to waste an assert validating their contract. I'll remove it entirely.

lib/ukpod/eager.c
andreittr: I guess moving the contract validation of `func` inside the error path makes sense and doesn't hurt readability too much.

andreittr: On 2nd thought, moving the assert would change the proscribed behavior and have the loop functions return 0 instead of infinite looping with a misbehaving `func`. The latter is unfortunate, but easy to debug, while the former is a pandora's box of undefined behavior.
Thus, no, I believe it is not correct to move the assert, and I'll keep an eye out for recent code that may have suffered this "optimization", to make sure that it doesn't have similarly broken logic.

andreittr:
Pushed addressing comment.

andreittr:
Pushed rebase to `RELEASE-0.19.0` and latest version of #1576 .

lib/ukpod/include/uk/pod/default.h
```diff
@@ -0,0 +1,233 @@
...
+#define UK_POD_DEFAULT_CTX (uk_pod_eager_init_ctx)
+
+#elif CONFIG_LIBUKPOD_DEFAULT_EXTERNAL
+
+#include <uk/pod/external.h>
+
+#define UK_POD_DEFAULT_TYPE external
+#define UK_POD_DEFAULT_CTX (&uk_pod_external_init_ctx)
+
+#else
```
mogasergiu: A `/* !CONFIG_LIBUKPOD_DEFAULT_* */` here as well?

lib/ukpod/anon.c
```diff
@@ -0,0 +1,56 @@
...
+	 * - remove dependency on memset (and thus a libc)
+	 * - faster code because we know the buf & size are page-aligned
+	 */
+	char *__align(PAGE_SIZE) p;
+
+	UK_ASSERT(PAGE_ALIGNED((__uptr)addr));
+	p = addr;
+	/* The compiler will unroll this byte-wise loop with appropriate SIMD */
+	for (__sz i = 0; i < npages * PAGE_SIZE; i++)
+		p[i] = 0;
```
mogasergiu, line 27: Hmm, on ARM64 builds I get a plain byte by byte memset?
```
000000004013be60 <uk_pod_anon_pagein>:
    4013be60:	aa0003e2 	mov	x2, x0
    4013be64:	f2402c1f 	tst	x0, #0xfff
    4013be68:	54000141 	b.ne	4013be90 <uk_pod_anon_pagein+0x30>  // b.any
    4013be6c:	aa0103e0 	mov	x0, x1
    4013be70:	d374cc21 	lsl	x1, x1, #12
    4013be74:	8b010043 	add	x3, x2, x1
    4013be78:	b40002a1 	cbz	x1, 4013becc <uk_pod_anon_pagein+0x6c>
    4013be7c:	d503201f 	nop
    4013be80:	3800145f 	strb	wzr, [x2], #1
    4013be84:	eb03005f 	cmp	x2, x3
    4013be88:	54ffffc1 	b.ne	4013be80 <uk_pod_anon_pagein+0x20>  // b.any
    4013be8c:	d65f03c0 	ret
    4013be90:	a9bf7bfd 	stp	x29, x30, [sp, #-16]!
    4013be94:	f0000d60 	adrp	x0, 402ea000 <__func__.0.lto_priv.0+0xa30>
    4013be98:	f0000b65 	adrp	x5, 402aa000 <_etext+0x7000>
    4013be9c:	910003fd 	mov	x29, sp
    4013bea0:	79534c01 	ldrh	w1, [x0, #2470]
    4013bea4:	910920a5 	add	x5, x5, #0x248
    4013bea8:	b0000b44 	adrp	x4, 402a4000 <_etext+0x1000>
    4013beac:	f0000b62 	adrp	x2, 402aa000 <_etext+0x7000>
    4013beb0:	911e4084 	add	x4, x4, #0x790
    4013beb4:	910a0042 	add	x2, x2, #0x280
    4013beb8:	528002e3 	mov	w3, #0x17                  	// #23
    4013bebc:	52800000 	mov	w0, #0x0                   	// #0
    4013bec0:	97ffe0d0 	bl	40134200 <_uk_printk>
    4013bec4:	52800060 	mov	w0, #0x3                   	// #3
    4013bec8:	97ff2b5a 	bl	40106c30 <ukplat_terminate>
    4013becc:	d65f03c0 	ret
```
The
```
    4013be80:	3800145f 	strb	wzr, [x2], #1
    4013be84:	eb03005f 	cmp	x2, x3
```
is pretty much a byte-by-byte zeroing as far as I can tell :thinking:

andreittr:
Pushed addressing comment.

lib/ukpod/include/uk/pod.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+/* PoD functions work with lengths and offsets in units of "pages".
+ *
+ * In the context of PoD a "page" is defined to be exactly PAGE_SIZE bytes,
+ * regardless of the actual hardware page sizes in use (e.g., hugepages).
+ */
+
+/* PoD functions all take as arguments `ops` -- page I/O operations -- as well
+ * as `arg` and `base`, which are used when calling the readin/writeback ops.
```
mogasergiu: I have just noticed this but you use `readin` instead of `pagein` throughout the docstrings.

lib/ukpod/include/uk/pod.h
```diff
@@ -0,0 +1,92 @@
...
+ * It is undefined behavior if any of the previous do not hold true.
+ */
+
+/* PoD functions that operate on existing allocations all take the arguments
+ * `addr` and `npages` to describe the buffer being operated on.
+ *
+ * It is undefined behavior if:
+ * - `addr` is not page-aligned
+ * - `addr` + `npages` describe a buffer that is not entirely backed by
+ *   allocations returned from previous calls to `uk_pod_*_acquire`
```
mogasergiu: Did you mean `uk_pod_*_alloc`?

lib/ukpod/include/uk/pod.h
```diff
@@ -0,0 +1,92 @@
...
+ * - free -- free (destroy) a PoD area
+ * - writeback -- explicitly request writeback of pages in a PoD area
+ * - drop -- explicitly request discarding pages within a PoD area
+ * - lock -- ensure pages in a PoD area remain resident in memory
+ * - unlock -- reverse the effects of lock
+ * - populate -- ensure pages in a PoD area are memory resident and populated
+ * The operations are presented in more detail in <uk/pod/default.h>, consult
+ * that file for more information.
+ *
+ * In addition to the system-wide PoD backend, and in order to satifsy bespoke
```
mogasergiu: s/satifsy/satisfy

mogasergiu (changes requested):

andreittr:
Pushed rebase to latest staging & #1576 .

andreittr:
Pushed addressing comments.

mogasergiu (approved):
LGTM just don't forget about [this](https://github.com/unikraft/unikraft/pull/1607#discussion_r2157250227)

andreittr:
Pushed rebase to current staging, removing the prerequisite PR commit.

andreittr:
Pushed update adding `exportsyms.uk` file to explicitly express developer intent.

michpappas:

PR #1411, opened 2024-05-06
lib/ukfile: Add backing buffer shared memory API

andreittr:
### Description of changes

This change expands the ukfile API to provide direct shared memory access to file backing buffers, useful for implementing memory mappings (mmap) or single-copy file-to-file transfers (sendfile).
In broad strokes, consumers would use it as such:
1. single-copy file-to-file (`sendfile`):
  - lock source file
  - `mem(BORROW)`
  - write into dest file the iovec received from above
  - unlock source file
2. shared memory mapping (file `mmap`s):
  - `mmap()` does `mem(ACQUIRE)`; this ensures
    (1) backing buffer is mapped by the file for the respective area and
    (2) addresses returned by `mem(RETRIEVE)` are stable until RELEASE
  - `munmap()` does `mem(RELEASE)`
  - sometime between `mmap` and paging in the memory, do `mem(RETRIEVE)` to get specific memory ranges
3. memory gifting (i.e., make the file take some memory and use it as its own)
  - call `mem(GIFT)` with memory you own and want the file to take over
  - like a reverse mmap: the file takes _your_ memory as its backing buffer, rather than the other way around
  - does not have a POSIX/Linux equivalent; can be used for nifty things

A boilerplate no-op implementation is provided, and existing ukfiles are patched to use it.
No actual implementations or consumers of this API are included. Rather, the intent of this PR is to review the API and ensure it is properly documented in isolation.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): [e.g. `x86_64` or N/A]
 - Platform(s): [e.g. `kvm`, `xen` or N/A]
 - Application(s): [e.g. `app-python3` or N/A]

### Additional configuration

 N/A

PR #1608, opened 2025-03-21
lib/ukfs: Introduce live reference driver template

andreittr:
### Description of changes

This change introduces the ukfs live ref filesystem driver template, a header with common boilerplate code for implementing the ukfs driver API on top of driver code that naturally deals in live filesystem node references -- objects whose lifetime and runtime state it manages internally.

As prerequisite for the live template, common utilities for dealing with pathnames and dirents have been added to `ukfs`.

~~Depends on the following PRs and has them included as squashed commits; please ignore during review:~~
~~- #1595~~
~~- #1411~~

~~Squashed commits will be removed on merging of respective PRs and and subsequent rebase.~~

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukfs/include/uk/fs/dirent.h
+++ b/lib/ukfs/include/uk/fs/dirent.h
@@ -1,0 +1,84 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Utilities for handling directory entries to be used by filesystem drivers */
+
+#ifndef __UKFS_FS_DIRENT_H__
+#define __UKFS_FS_DIRENT_H__
+
+#include <dirent.h>
+
+#include <uk/fs.h>
+
+/* Identical to libc struct dirent64, but with dynamically-sized name field */
+struct uk_fs_dirent {
+	ino_t d_ino;
+	off_t d_off;
+	unsigned short d_reclen;
+	unsigned char d_type;
+	char d_name[];
+};
+
+/* Asserts to ensure dirent64 & uk_fs_dirent have identical layouts in memory */
+UK_CTASSERT(__offsetof(struct uk_fs_dirent, d_ino) ==
+	    __offsetof(struct dirent64, d_ino));
+UK_CTASSERT(__offsetof(struct uk_fs_dirent, d_off) ==
+	    __offsetof(struct dirent64, d_off));
+UK_CTASSERT(__offsetof(struct uk_fs_dirent, d_reclen) ==
+	    __offsetof(struct dirent64, d_reclen));
+UK_CTASSERT(__offsetof(struct uk_fs_dirent, d_type) ==
+	    __offsetof(struct dirent64, d_type));
+UK_CTASSERT(__offsetof(struct uk_fs_dirent, d_name) ==
+	    __offsetof(struct dirent64, d_name));
+
+/**
+ * Compute the record length for a dirent64 with name of length `namelen`.
+ *
+ * This takes into account the terminating NUL in the file name as well as
+ * alignment requirements for a following dirent64.
+ */
+#define UKFS_DIRENT_RECLEN(namelen) \
+	ALIGN_UP(sizeof(struct uk_fs_dirent) + (namelen) + 1, \
+		 __alignof__(struct dirent64))
+
+/**
+ * Get the next dirent after entry `dp` of size `sz`.
+ */
+#define UKFS_DIRENT_NEXT(dp, sz) \
+	((struct uk_fs_dirent *)((char *)(dp) + (sz)))
+
+/**
+ * Output a record for entry "." into `out` with inode number `inode`.
+ *
+ * This macro fills in only the essential fields; drivers are free to write out
+ * any additional dirent fields they wish.
+ */
+#define UKFS_DIRENT_OUT_DOT(out, inode) \
+do { \
+	(out)->d_ino = (inode); \
+	(out)->d_reclen = UKFS_DIRENT_RECLEN(1); \
+	(out)->d_type = DT_DIR; \
+	(out)->d_name[0] = '.'; \
+	(out)->d_name[1] = '\0'; \
+} while (0)
+
+/**
+ * Output a record for entry ".." into `out` with inode number `inode`.
+ *
+ * This macro fills in only the essential fields; drivers are free to write out
+ * any additional dirent fields they wish.
+ */
+#define UKFS_DIRENT_OUT_DOTDOT(out, inode) \
+do { \
+	(out)->d_ino = (inode); \
+	(out)->d_reclen = UKFS_DIRENT_RECLEN(2); \
+	(out)->d_type = DT_DIR; \
+	(out)->d_name[0] = '.'; \
+	(out)->d_name[1] = '.'; \
+	(out)->d_name[2] = '\0'; \
+} while (0)
+
+#endif /* __UKFS_FS_DIRENT_H__ */
--- a/lib/ukfs/include/uk/fs/pathutil.h
+++ b/lib/ukfs/include/uk/fs/pathutil.h
@@ -1,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/* Path utils */
+
+#ifndef __UKFS_FS_PATHUTIL_H__
+#define __UKFS_FS_PATHUTIL_H__
+
+#include <stddef.h>
+#include <string.h>
+
+/**
+ * Check if the first path component in string `path` of length `len` is ".".
+ */
+static inline
+int uk_fs_path_isdot(const char *path, size_t len)
+{
+	return len >= 1 && path[0] == '.' && (len == 1 || path[1] == '/');
+}
+
+/**
+ * Check if the first path component in string `path` of length `len` is "..".
+ */
+static inline
+int uk_fs_path_isdotdot(const char *path, size_t len)
+{
+	return len >= 2 && path[0] == '.' && path[1] == '.' &&
+	       (len == 2 || path[2] == '/');
+}
+
+/**
+ * Return the offset of the next path separator ('/') in `path` of length `len`,
+ * or `len` if no separator is found.
+ *
+ * @param path String containing path (does not need NUL terminator)
+ * @param len Length of path
+ *
+ * @return
+ *   < `len`: Offset of first path separator in `path`
+ *  == `len`: Path separator not found
+ */
+static inline
+size_t uk_fs_path_sep(const char *path, size_t len)
+{
+	const char *const p = memchr(path, '/', len);
+
+	if (!p)
+		return len;
+
+	UK_ASSERT(p >= path);
+	return (size_t)(p - path);
+}
+
+#endif /* __UKFS_FS_PATHUTIL_H__ */
--- a/lib/ukfs/include/uk/fs/template/live.h
+++ b/lib/ukfs/include/uk/fs/template/live.h
@@ -1,0 +1,2007 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/*
+ * Unikraft filesystem template for live reference drivers.
+ *
+ * This header file contains boilerplate code for implementing a Unikraft
+ * filesystem driver based on code that natively handles live references.
+ * This functionality is provided in the form of code generation macros that
+ * define functions conforming to the ukfile and ukfs file operation APIs,
+ * functions which in turn call into driver-implemented live reference ops.
+ *
+ * A "live reference" to a file is understood to be a reference that handles
+ * both file state (struct uk_file_state) as well as node refcounting and
+ * lifetime. A live reference does not need to concern itself with other runtime
+ * properties such as mount points or filesystem instances.
+ */
+
+#ifndef __UKFS_FS_TEMPLATE_LIVE_H__
+#define __UKFS_FS_TEMPLATE_LIVE_H__
+
+#include <sys/mount.h>
+#include <sys/stat.h>
+
+#include <uk/alloc.h>
+#include <uk/atomic.h>
+#include <uk/essentials.h>
+#include <uk/errptr.h>
+#include <uk/fs.h>
+#include <uk/fs/common-ops.h>
+#include <uk/fs/pathutil.h>
+#include <uk/refcount.h>
+#include <uk/rwlock.h>
+#include <uk/sched.h>
+#include <uk/spinlock.h>
+#include <uk/tree.h>
+
+/* Types of special filesystem nodes recognized by the live driver template */
+enum uk_fs_tmpl_node_kind {
+	UK_FS_TMPL_DIR, /* Directory */
+	UK_FS_TMPL_SYM /* Symbolic link */
+};
+
+/* Driver-specific types */
+
+#define UK_FS_TMPL_LIVE_CREATE_TARGET(drv) drv##_LIVE_CREATE_TARGET
+
+/**
+ * Define the specific types of live driver `drv` with nodes of `nodetype`.
+ *
+ * @param drv Driver name
+ * @param nodetype Type used for live node references
+ */
+#define UK_FS_TMPL_LIVE_TYPES(drv, nodetype)				\
+/* Target argument type for live create operation */			\
+union UK_FS_TMPL_LIVE_CREATE_TARGET(drv) {				\
+	nodetype livenode;						\
+	union uk_fs_create_target ukfs;					\
+}
+
+/* Live ops declarations
+ *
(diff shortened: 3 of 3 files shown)


lib/ukfs/include/uk/fs/pathutil.h
```diff
@@ -0,0 +1,52 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
mogasergiu, line 1: In commit message of ` lib/ukfs: Add utilities for processing paths `, `s/filessystem/filesystem`

lib/ukfs/include/uk/fs/pathutil.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+static inline
+size_t uk_fs_path_sep(const char *path, size_t len)
+{
+	const char *p = memchr(path, '/', len);
+
+	if (p) {
+		UK_ASSERT(p >= path);
+		return (size_t)(p - path);
+	} else {
```
mogasergiu: I think you can get rid of `else` and just return. Better yet,
```C
if (!p)
    return len;
    
UK_ASSERT(p >= path);
return (size_t)(p - path);
```

lib/ukfs/include/uk/fs/dirent.h
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Get the next dirent after entry `dp` of size `sz`.
+ */
+#define UKFS_DIRENT_NEXT(dp, sz) \
+	((struct uk_fs_dirent *)((char *)(dp) + (sz)))
+
+/**
+ * Output a record for entry "." into `out` with inode number `inode`.
+ */
+#define UKFS_DIRENT_OUT_DOT(out, inode) do { \
```
mogasergiu: Could we align this one with the other `do { } while` macro's in the codebase and place the `do` on the next line (ditto the macro below)? :)

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+}									\
+\
+UK_RB_KEY_GENERATE_STATIC(drv##_FILEMAP, drv##_FILE_ALLOC, rb_entry,	\
+			  nodecmp, drv##_FILEMAP_KEY);			\
+\
+attr struct drv##_ISTATE *drv##_ISTATE_NEW(struct uk_alloc *al)		\
+{									\
+	struct drv##_ISTATE *istate = uk_malloc(al, sizeof(*istate));	\
+									\
+	if (likely(istate)) {						\
```
mogasergiu: `unlikely` and early return?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+	ret = UK_RB_FIND(drv##_FILEMAP, &s->map, n);			\
+	if (ret) {							\
+		if (unlikely(!uk_file_try_acquire(&ret->f)))		\
+			/* We found file mid-destruction; report not found */\
+			ret = NULL;					\
+	}								\
+	uk_rwlock_runlock(&s->lock);					\
+	if (ret)							\
+		return &ret->f;						\
+	else								\
```
mogasergiu: you can remove `else` and just return

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+\
+attr const struct uk_file *drv##_FILE_GET(struct drv##_ISTATE *s,	\
+					  UK_FS_TMPL_LIVE_NODETYPE(drv) n)\
+{									\
+	struct drv##_FILE_ALLOC *ret;					\
+									\
+	uk_rwlock_rlock(&s->lock);					\
+	ret = UK_RB_FIND(drv##_FILEMAP, &s->map, n);			\
+	if (ret) {							\
+		if (unlikely(!uk_file_try_acquire(&ret->f)))		\
```
mogasergiu: `if (ret && unlikely(!...)` ?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+}									\
+\
+static void drv##_ISTATE_ACQUIRE(struct drv##_ISTATE *istate)		\
+{									\
+	uk_refcount_acquire(&istate->refcnt);				\
+}									\
+\
+attr void drv##_ISTATE_RELEASE(struct drv##_ISTATE *istate)		\
+{									\
+	if (uk_refcount_release(&istate->refcnt)) {			\
```
mogasergiu, line 508: `if (!uk_refcount_release)` and early return for less indentation? :sweat_smile:

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+						    &s->map, ret);	\
+				UK_ASSERT(!prev);			\
+			}						\
+		}							\
+		uk_rwlock_wunlock(&s->lock);				\
+		break;							\
+	}								\
+	if (ret)							\
+		return &ret->f;						\
+	else								\
```
mogasergiu: Unnecessary else?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+	}								\
+}									\
+\
+attr									\
+struct drv##_FILE_ALLOC *drv##_FILE_FROMNODE(UK_FS_TMPL_LIVE_NODETYPE(drv) n,\
+					     struct drv##_ISTATE *istate)\
+{									\
+	struct drv##_FILE_ALLOC *al = uk_malloc(istate->al, sizeof(*al));\
+									\
+	if (likely(al)) {						\
```
mogasergiu: `unlikely(!al)` and early return + less indentation?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1907 @@
...
+									\
+		if (likely(uk_compare_exchange_n(&idata->mnt, &prev, t)))\
+			uk_file_acquire(t);				\
+		else							\
+			return -EEXIST;					\
+	} else {							\
+		uk_spin_lock(&idata->mntlock);				\
+		t = uk_exchange_n(&idata->mnt, NULL);			\
+		uk_spin_unlock(&idata->mntlock);			\
+		if (likely(t))						\
```
mogasergiu: `if (unlikely(!t)) return -ENOENT` and no else/indentation?

mogasergiu (commented):
Not much to say other than some suggestions and nitpicks. Next up is iomem and then ramfs. It is then I will see these new additions put in motion.

Another thing worth noting is that, while this template stuff makes writing fs drivers much easier, I think it is going to make source debugging a lot harder. No other way I guess :man_shrugging:.

lib/ukfs/include/uk/fs/dirent.h
andreittr: Sure, no idea we had a macro house style

lib/ukfs/include/uk/fs/template/live.h
andreittr: That would make us trying to take a refcount (& its side-effects) infinitely more obscured. No.
In fact, this should probably be even more explicit and drop the `unlikely` altogether, as it's not an error condition.
```c
if (ret) {
    const int r = uk_file_try_acquire(&ret->f);

    if (!r)
        ret = NULL;
}
```

andreittr:
Pushed rebase to `RELEASE-0.19.0` & latest version of #1595 ; minor changes were required with newest #1595.

andreittr:
Pushed addressing comments & minor fixes.

andreittr:
Pushed rebase to latest #1595 .

lib/ukfs/include/uk/fs/pathutil.h
```diff
@@ -0,0 +1,51 @@
...
+#define __UKFS_FS_PATHUTIL_H__
+
+#include <stddef.h>
+#include <string.h>
+
+/**
+ * Check if the first path component in string `path` of length `len` is ".".
+ */
+#define UKFS_PATH_ISDOT(path, len) \
+	((len) >= 1 && (path)[0] == '.' && ((len) == 1 || (path)[1] == '/'))
```
mogasergiu: Wondering if these macros would go better as static inlines instead. For example, since they are usable by other libs AFAICT, the fact that they evaluate the parameters multiple times might lead to undesired behavior if someone uses these with function calls as arguments :thinking: . But at the same time, the compiler may be free to not inline them if it does not want to (although I don't see why it wouldn't).
There is no validation for correct parameters anyway so I guess it is assumed the users should know what they are doing (maybe static inline + assert?).

lib/ukfs/include/uk/fs/dirent.h
```diff
@@ -0,0 +1,78 @@
...
+	((struct uk_fs_dirent *)((char *)(dp) + (sz)))
+
+/**
+ * Output a record for entry "." into `out` with inode number `inode`.
+ */
+#define UKFS_DIRENT_OUT_DOT(out, inode) \
+do { \
+	(out)->d_ino = (inode); \
+	(out)->d_reclen = UKFS_DIRENT_RECLEN(1); \
+	(out)->d_type = DT_DIR; \
```
mogasergiu, line 63: What about `d_off`, ditto for DOTDOT

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1923 @@
...
+	struct drv##_ISTATE *istate = drv##_FILE_ISTATE(f);		\
+	struct drv##_ISTATE *newstate = drv##_ISTATE_NEW(istate->al);	\
+	struct drv##_FILE_ALLOC *newf;					\
+	struct drv##_FILE_ALLOC *prev __maybe_unused;			\
+									\
+	if (unlikely(!newstate))					\
+		return ERR2PTR(-ENOMEM);				\
+	newstate->mntflags = flags;					\
+	newf = drv##_FILE_FROMNODE(node, newstate);			\
+	drv##_ISTATE_RELEASE(newstate);					\
```
mogasergiu, line 1827: Shouldn't this happen after we are sire the allocation of `newf` succeeded?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1923 @@
...
+ * (refcounted) reference that the caller is responsible for releasing.
+ * Such operations should signal error conditions by special values of
+ * `nodetype` that are mapped to negative error codes by `live_errnode`.
+ *
+ * For best optimizaton, these functions should be implemented statically in
+ * the same compilation unit as the UK_FS_TMPL_LIVE_GEN_* macro that uses them.
+ */
+#define UK_FS_TMPL_LIVE_OPS(drv, sname, nodetype)			\
+/* Target argument type for live create operation */			\
+union UK_FS_TMPL_LIVE_CREATE_TARGET(drv) {				\
```
mogasergiu: I am likely being overly cautious here, but should this be made more unique? Imagine this macro is used more than once in the same file, I think that could lead to a redefinition error for the union right? maybe `drv_##sname`?
(Maybe worth considering in other places where the sole param is used?)

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1923 @@
...
+					   live_errnode, live_release,	\
+					   attr)			\
+attr UK_FS_TMPL_LIVE_DECL_FS_CREATE(drv)				\
+{									\
+	const unsigned int ftype = mode & S_IFMT;			\
+	int err;							\
+	UK_FS_TMPL_LIVE_NODETYPE(drv) rnode;				\
+	union UK_FS_TMPL_LIVE_CREATE_TARGET(drv) live_target;		\
+	const struct uk_file *ret;					\
+	struct drv##_FILE_ALLOC *prev __maybe_unused;			\
```
mogasergiu: What do you need this prev for? :thinking:

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1923 @@
...
+	if (istate) {							\
+		istate->mntflags = flags;				\
+		ret = drv##_FILE_IOPEN(istate, rootnode);		\
+		if (!ret)						\
+			ret = ERR2PTR(-ENOMEM);				\
+	} else {							\
+		ret = ERR2PTR(-ENOMEM);					\
+	}								\
+	(live_release)(rootnode);					\
+	drv##_ISTATE_RELEASE(istate);					\
```
mogasergiu: Is it safe to call this even if `istate` is `NULL`?

lib/ukfs/include/uk/fs/template/live.h
```diff
@@ -0,0 +1,1923 @@
...
+				uk_rwlock_wunlock(&s->lock);		\
+				uk_sched_yield();			\
+				continue;				\
+			}						\
+		/* Else try to create new file */			\
+		} else if ((ret = drv##_FILE_FROMNODE(n, s))) {		\
+			struct drv##_FILE_ALLOC *prev __maybe_unused;	\
+									\
+			prev = UK_RB_INSERT(drv##_FILEMAP, &s->map, ret);\
+			UK_ASSERT(!prev);				\
```
mogasergiu, line 571: I see other places where you do this. Is it ok to have it catch insertion errors only on builds with assertions enabled?

lib/ukfs/include/uk/fs/dirent.h
andreittr: These two macros are intended to writeout the minimum valid dirent for `.` & `..`, leaving the driver to do the rest.

According to [getdents64(2)](https://www.man7.org/linux/man-pages/man2/getdents64.2.html#DESCRIPTION): "d_off is a filesystem-specific value with no specific meaning to user space", so the fs driver is best posed to write this value out how it wishes, or indeed leave it unitialized [as ramfs does](https://github.com/unikraft/unikraft/pull/1609/files#diff-35e93af7eedc209bef6cb0bdb0648bf0c75ae5ebf376679624553d6f970c359dR956) :smile: .

lib/ukfs/include/uk/fs/template/live.h
andreittr: No, this is on purpose:
- if `newf` succeeded, it took a refcount on `newstate` so this simply ticks it down by 1
- if `newf` failed, this handles cleanup of the (no longer needed) `newstate`

This neatly encapsulates cleanup on error without dedicated code in the `if (unlikely(!newf))` block.

But now that you mention it, I do agree that the later use of `newstate` does look problematic -- we released the reference so _we shouldn't be using it anymore_. This happens to be fine because `newf` has a reference to `newstate`, but it does look very bad.
Fortunately `drv##_FILE_SET` AFAICT can be implemented with just `newf` as arg, making the API much cleaner (and explicitly using the _file's_ counted reference in its impl). Thanks for that :+1: I'll change it in the pass.

lib/ukfs/include/uk/fs/template/live.h
andreittr: Oof, good catch. I think this is left over from before `drv##_FILE_IOPEN` was a thing and the compiler (rightfully) never complained :sweat_smile: .

lib/ukfs/include/uk/fs/template/live.h
andreittr: This macro is only meant to be used exactly once in single source file drivers that need to define an `ops_table` struct for later use in `UK_FS_TMPL_LIVE_GENERATE_STATIC` (another macro meant to be used exactly once). More modular drivers never need to see or worry about it.

That said, thanks for catching the fact that `union UK_FS_TMPL_LIVE_CREATE_TARGET(drv)` should probably be declared somewhere else than here.

lib/ukfs/include/uk/fs/template/live.h
andreittr: I think so; on this code path (`UK_RB_FIND` returns NULL) for `UK_RB_INSERT` to return a prev node can only mean a bug in either the RBtree implementation or here in the live driver boilerplate; something OK to catch with an assert IMO. If the code is bug-free this should be impossible to trigger.

lib/ukfs/include/uk/fs/template/live.h
andreittr: Good catch, yeah, this should be above in the `if` branch after `drv##_FILE_IOPEN` :facepalm: .

lib/ukfs/include/uk/fs/pathutil.h
andreittr: Yeah, I think you're right; these expressions are complex enough that a static inline makes more sense, and them being a macro doesn't bring any concrete benefit. Removing multiple eval & typesafety can't hurt either.

Re: validation: no, these aren't meant to take unsanitized input any more than `memcmp` is meant to.

andreittr:
Pushed rebase to updated #1595 .

andreittr:

lib/ukfs/include/uk/fs/dirent.h
mogasergiu: Can you add a comment about that here as well?

andreittr:
Pushed rebase to latest #1595 + required changes.

andreittr:
Pushed update including #1411 as dependency, along with support for the `ukfile.mem` operation for liveref driver files.

andreittr:
Pushed addressing comment.

andreittr:
Pushed typo fix in type name.

andreittr:
Pushed rebase to latest #1411 .

mogasergiu (approved):

andreittr:
Pushed rebase to current staging &  latest #1595 .

andreittr:
Pushed rebase to latest staging after merging of all prerequisites.

michpappas:

PR #1609, opened 2025-03-21
lib/ukfs-ramfs: Introduce RAMfs filesystem for ukfs

andreittr:
### Description of changes

This change introduces a memory resident volatile filesystem registered as "ramfs" under `ukfs`.
It makes use of `uksparsebuf` and `ukpod` for file contents, and the ukfs live driver template.

~~Depends on the following PRs (included as squash commits, ignore during review):~~
~~- #1608~~ 
~~- #1607~~ 
~~- #1606~~ 
~~- #1629~~ 

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/ukfile/include/uk/file/iovutil.h
```diff
@@ -14,10 +14,59 @@
...
+		 size_t *iovip, size_t *curp)
+{
+	size_t ret = 0;
+	size_t i = *iovip;
+	size_t cur = *curp;
+
+	while (i < iovcnt && len) {
+		const size_t sz = MIN(len, iov[i].iov_len - cur);
+
+		UK_ASSERT(cur <= iov[i].iov_len);
```
mogasergiu: Is there any point to having this assert in the while loop? :thinking:

lib/ukfile/include/uk/file/iovutil.h
```diff
@@ -31,34 +80,23 @@ size_t uk_iov_zero(const struct iovec *iov, size_t iovcnt, size_t len,
...
+	while (i < iovcnt && len) {
+		const size_t sz = MIN(len, iov[i].iov_len - cur);
 
-		memset((char *)iov[i].iov_base + cur, 0, l);
-		ret += l;
-		len -= l;
-		cur += l;
+		UK_ASSERT(cur <= iov[i].iov_len); /* cur must be in bounds */
+		UK_ASSERT(sz || cur == iov[i].iov_len); /* must make progress */
+		if (sz) {
```
mogasergiu: I guess this is here for the case where we handle `cur == iov[i].iov_len`. Wouldn't that be considered that `cur` is not in bounds since it is at the very edge? In the comment you've specified "< `iovcnt`" as upper bound for `iovi` so why not the same for `cur` in relation to `iov[iovi].iov_len`? And in case you do want to keep this check, is it possible to reach this condition within the while loop to justify checking for it on every iteration or would it be better done once outside the while loop?

(Ditto for the other inlines below)

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+};
+
+#define RAMFS_MAX_NAMELEN USHRT_MAX
+
+static
+struct ramfs_dentry *ramfs_dentry_new(const char *name, unsigned short namelen)
+{
+	struct ramfs_dentry *ret = ramfs_obj_alloc(sizeof(*ret) + namelen);
+
+	if (ret) {
```
mogasergiu, line 93: `if (unlikely(!ret))` and early return? Same goes for other places in this PR where I believe there are wasted opportunities to avoid indentation and other minor stuff like that, I won't mention on each one in order to not be spammy, I believe you can spot and adjust them yourself if you agree with my suggestion.

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+		return ERR2PTR(-ENOMEM);
+	ramfs_init_dir(rootnode, 0777);
+	return rootnode;
+}
+
+/* File ops */
+
+static inline
+int ramfs_check_regfile(const struct ramfs_node *n)
+{
```
mogasergiu, line 545: `unlikely` x 2 here?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+	if (ramfs_live_nodekind(targ, UK_FS_TMPL_SYM)) {
+		ramfs_live_acquire(n);
+		out_nod[1] = n;
+		out_nod[0] = targ;
+		*nout = cur;
+		return UKFS_STOP_SYM;
+	} else if (cur == len) {
+		out_nod[0] = targ;
+		return UKFS_SUCCESS;
+	} else {
```
mogasergiu: Unnecessary `else`?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+		return ERR2PTR(-ENAMETOOLONG);
+
+	newnode = ramfs_node_create(mode, target);
+	/* Link in if success & not tmpfile */
+	if (!PTRISERR(newnode) && !(flags & O_TMPFILE)) {
+		/* Lookup name dentry */
+		struct ramfs_dentry *dp = ramfs_dir_find(n, name, namelen);
+
+		if (unlikely(dp && (flags & O_EXCL))) {
+			ramfs_live_release(newnode);
```
mogasergiu: Hmm I noticed in the cration function that the live acquire is only done for `S_IFMT`

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+	struct ramfs_node *newnode;
+
+	/* refuse to operate on '.' and '..' */
+	if (unlikely(!(flags & O_TMPFILE) &&
+		     (UKFS_PATH_ISDOT(name, namelen) ||
+		      UKFS_PATH_ISDOTDOT(name, namelen))))
+		return ERR2PTR(-EEXIST);
+	if (unlikely(!(flags & O_TMPFILE) && namelen > RAMFS_MAX_NAMELEN))
+		return ERR2PTR(-ENAMETOOLONG);
+
```
mogasergiu, line 1202: Should we also check and return `EINVAL` for `RENAME_WHITEOUT`?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+			ramfs_live_release(dd->node);
+			dd->node = sd->node;
+			ramfs_node_relink(dd->node, dest);
+			ramfs_dir_remove(n, sd);
+			ramfs_obj_free(sd);
+			ret = 0;
+		}
+	} else {
+		/* Destination does not exist */
+		if (flags & RENAME_EXCHANGE) {
```
mogasergiu: `unlikely`?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1200 @@
...
+ */
+
+struct ramfs_node;
+
+/* RAMfs directory entry; associates a filename with a filesystem node */
+struct ramfs_dentry {
+	/* Bindings for lookup by name */
+	UK_RB_ENTRY(ramfs_dentry) rb_name;
+	/* Node referenced by this dentry */
+	struct ramfs_node *node;
```
mogasergiu: Nice, the node is finally decoupled from the dentry!

mogasergiu (commented):
Had a first look. Moving on to posix-vfs and will likely come back to either of the prereq PR's the more context I build as I am reading more.

lib/ukfile/include/uk/file/iovutil.h
andreittr: We need `cur == iov[i].iov_len` to support the case where `cur == iov_len == 0`, as userspace likes to send you sparse iovecs sometimes :upside_down_face:

lib/ukfs-ramfs/ramfs.c
andreittr: Yep, on purpose.
- `S_IFMT` lends you a node that you're meant to take a ref on yourself before linking in, we do that with `live_acquire`; this release then undos the acquire and leaves the borrowed ref as before
- any other `S_` constants have you actually create a new node; any newly-created node starts off with a refcount of 1; this release then ticks off the last refcount, triggering destruction

lib/ukfs-ramfs/ramfs.c
andreittr: Why would we do that in the create function :thinking: ?
But, assuming you meant to put it in rename, I have to check if it's safe to ignore this flag altogether. The intended use seems to be for overlay/union fs _implementations_ to do renames using syscalls (?!) and it all smells like it should never have leaked into userspace to begin with.
If not, hard-erroring on this flag is also valid.

andreittr:
Pushed rebase to staging + dependency on #1629 (removing the iovutil commit)

andreittr:
Pushed rebase to latest version of prerequisite PRs + changes reflecting updates to #1608 and #1606 .

andreittr:
Pushed rebase to latest #1608 & #1629 .

lib/ukfs-ramfs/ramfs.c
andreittr: Since it's a static inline that computes a value based off some memory reads and its result is always used in an `unlikely` branch anyway, I don't think adding branch hints here would help any.

andreittr:
Pushed addressing comments + minor cosmetic tweaks.

lib/ukfs-ramfs/ramfs-file.h
```diff
@@ -0,0 +1,220 @@
...
+	return ret;
+}
+
+static
+size_t ramfs_file_data_write(struct ramfs_file_data *file_data,
+			     const struct iovec *iov, size_t iovcnt,
+			     size_t off, size_t rem)
+{
+	const size_t pgoff = off / PAGE_SIZE;
+	const size_t pgend = ALIGN_UP(off + rem, PAGE_SIZE) / PAGE_SIZE;
```
mogasergiu: I think you can use `PAGE_COUNT`. Ditto for `pgoff`

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1393 @@
...
+		size_t sz = UKFS_DIRENT_RECLEN(p ? 2 : 1);
+
+		if (unlikely(sz > len)) {
+			full = 1;
+			break;
+		}
+		if (p == 0) { /* 0: "." */
+			UKFS_DIRENT_OUT_DOT(out, (ino_t)(uintptr_t)n);
+		} else { /* 1: ".." */
+			struct ramfs_node *parent = n->dir_data.parent;
```
mogasergiu, line 1042: I see this does not acquire the plock like in other places. Is this okay?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1393 @@
...
+	if (unlikely(ISROFS(mntflags)))
+		return -EROFS;
+	if (unlikely(r))
+		return r;
+	if (unlikely(!rem))
+		return 0;
+
+	written = ramfs_file_data_write(&n->file_data, iov, iovcnt, off, rem);
+	if (unlikely(!written))
+		return -ENOSPC;
```
mogasergiu, line 595: Why not ENOMEM?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1393 @@
...
+
+static
+int ramfs_live_fs_sync(struct ramfs_node *n __unused)
+{
+	/* Nothing to do, ramfs is always synced */
+	return 0;
+}
+
+static
+int ramfs_dir_isempty(struct ramfs_node *n)
```
mogasergiu, line 891: What is the reason that only `isempty` locks `children` but insertion/removals don't?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1393 @@
...
+	UK_ASSERT(ramfs_live_nodekind(n, UK_FS_TMPL_DIR));
+	return UK_RB_FIND(ramfs_name_map, &n->dir_data.children,
+		((struct ramfs_entry_name){ .len = len, .name = name }));
+}
+
+static
+void ramfs_dir_insert(struct ramfs_node *n, struct ramfs_dentry *dp)
+{
+	UK_ASSERT(ramfs_live_nodekind(n, UK_FS_TMPL_DIR));
+	UK_RB_INSERT(ramfs_name_map, &n->dir_data.children, dp);
```
mogasergiu: I think this returns something. Why do we not check for it? Ditto for `UK_RB_REMOVE` below.

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1393 @@
...
+		ramfs_live_acquire(dp->target.node);
+		out_nod[0] = dp->target.node;
+		if (ramfs_live_nodekind(dp->target.node, UK_FS_TMPL_SYM)) {
+			ramfs_live_acquire(n);
+			out_nod[1] = n;
+			return UKFS_STOP_SYM;
+		}
+		return UKFS_STOP_NOD;
+	case RAMFS_DENT_FILE:
+		uk_file_acquire(dp->target.file);
```
mogasergiu, line 984: Looking at the LOOKUP up from live template, are we sure this will always have `deep = 1`? Because otherwise it may call `live_release` instead of `uk_file_release` and I guess that would not be good.

mogasergiu (commented):
Had another round and have a few questions. I will definitely have to do another round though looking into the expansion of the live ops. I only looked at the static generate one.

lib/ukfs-ramfs/ramfs-file.h
andreittr: For `pgoff` absolutely not, as we explicitly want to round _down_ a page.
For `pgend` the maths works out, but `PAGE_COUNT` is misleading as a name, as there's no "counting" going on, we're looking for an end offset instead. Using `DIV_ROUND_UP` directly would be the better choice.

lib/ukfs-ramfs/ramfs.c
andreittr: Yes, as we are only ever using the _value_ of the `parent` pointer not any of the memory contents therein, so we don't need to take a refcount.

The `plock` spinlock is meant to allow one to atomically (1) read the `parent` pointer and (2) take a ref on it. This is in order to prevent a UAF if `n` somehow gets reparented between steps (1) and (2) and the last active ref on parent is dropped in the process. Here we do not have this case.

lib/ukfs-ramfs/ramfs.c
andreittr: Because `ENOSPC` is the error code documented by [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html#ERRORS) for when
```
       ENOSPC The device containing the file referred to by fd has no room for the data.
```
and the error the vast majority of code dealing with the FS is likely to expect and have handlers for. This is also what any other filesystem would return when running out of space for _data_.

ENOMEM is best kept for when OOMing when allocating kernel bookkeeping data structures -- a separation which ramfs keeps, using memory allocators for metadata & ukpod for data, each of which can fail independently.

lib/ukfs-ramfs/ramfs.c
andreittr: Because I wrote `isempty` strictly for the convenience in [unlink](https://github.com/unikraft/unikraft/pull/1609/files/dcbe4bbcad81112009d157878d36cc525bba3de8..39d1086bb4a15f190cf7093038fbf7c11ca676d7#diff-35e93af7eedc209bef6cb0bdb0648bf0c75ae5ebf376679624553d6f970c359dR1183) where it's called with a _child_ of the target node which is not locked.

insert/remove are meant to be called only on target nodes which, by the convention set by the liveref template and ukfs, should already be synchronized on their state locks.

lib/ukfs-ramfs/ramfs.c
andreittr: Yeah, good point, we should probably assert their success here :+1: .

Edit: on closer reading of the [RB tree implementation](https://github.com/unikraft/unikraft/blob/staging/include/uk/tree.h#L808) I see no code path where it would return anything other than its 2nd arg (`dp` in this case). I also see no case of the return value being captured in the codebase; I'm not sure why it's written this way, possibly to enable chaining like
```c
UK_RB_INSERT(mymap, map1, UK_RB_REMOVE(mymap, map2, item));
```
:shrug: . I'll leave `UK_RB_REMOVE` as-is and only add retval validation for INSERT.

lib/ukfs-ramfs/ramfs.c
andreittr: I don't see how these would interact, could you please elaborate? :thinking: 

In LOOKUP the value of `deep` only matters `if (!d)`, which on lookup of the first path element will not be true (`d` is [initialized to `f`](https://github.com/unikraft/unikraft/pull/1608/files/99cd5751ceb3d081b4681b2e6b060dc8accb8a94..117b8403570670129d919e02d61c05b58cee2873#diff-a967b350d3c0827907a6429f0cf874bed3cfaccae36a5d1694b41cc0b71d3faaR1263) before the lookup loop and only gets replaced on `UKFS_SUCCESS`/`UKFS_STOP_NOD`).
`UKFS_STOP_FILE` which we return here [shortcuts the loop and exits](https://github.com/unikraft/unikraft/pull/1608/files/99cd5751ceb3d081b4681b2e6b060dc8accb8a94..117b8403570670129d919e02d61c05b58cee2873#diff-a967b350d3c0827907a6429f0cf874bed3cfaccae36a5d1694b41cc0b71d3faaR1438).

andreittr:
Pushed rebase to latest #1608 , #1607 , #1606 + required changes for compat.

andreittr:
Pushed addressing comments.

lib/nolibc/include/fcntl.h
```diff
@@ -162,6 +162,9 @@ struct f_owner_ex {
 
 #define FALLOC_FL_KEEP_SIZE 1
 #define FALLOC_FL_PUNCH_HOLE 2
+#define FALLOC_FL_COLLAPSE_RANGE 8
+#define FALLOC_FL_ZERO_RANGE 0x10
```
mogasergiu: Adding comment here so we don't forget: This is defined by kernel headers not libc's. So native builds with musl for example would fail at build time since it does not define these.

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1402 @@
...
+
+static
+void ramfs_dir_remove(struct ramfs_node *n, struct ramfs_dentry *dp)
+{
+	UK_ASSERT(ramfs_live_nodekind(n, UK_FS_TMPL_DIR));
+	UK_RB_REMOVE(ramfs_name_map, &n->dir_data.children, dp);
+}
+
+static
+int ramfs_dir_lookup(struct ramfs_node *n,const char *name, size_t len,
```
mogasergiu: missing space after ` *n,`

lib/ukfs-ramfs/ramfs.c
```diff
@@ -0,0 +1,1402 @@
...
+}
+
+static
+void ramfs_init_sym(struct ramfs_node *n, unsigned int mode,
+		    const char *path, size_t len)
+{
+	ramfs_init_statx(n, RAMFS_FMODE(mode, S_IFLNK));
+	/* Init sym_data */
+	n->sym_data.len = len;
+	memcpy(n->sym_data.path, path, len);
```
mogasergiu, line 244: so `ramfs_node_create` already passes `xlen` as argument which also counts the `\0` for proper node allocation. Why do we copy the `\0` twice? Unless you intended to pass `len - 1` to `sym_data.len`, which I think is the case as per `uk_fs_path` documentation (length of `strlen`) and have this be `	memcpy(n->sym_data.path, path, len - 1);` instead

mogasergiu (changes requested):

lib/ukfs-ramfs/ramfs-file.h
```diff
@@ -0,0 +1,220 @@
...
+/* TODO: Make configurable */
+/* TODO: Make compile-time when default alloc is constexpr */
+static
+int init_ramfs_sbctx(struct uk_init_ctx *ictx __unused)
+{
+	ramfs_sb_ctx.alloc = uk_alloc_get_default();
+	return 0;
+}
+
+uk_rootfs_initcall_prio(init_ramfs_sbctx, 0x0, 0);
```
mogasergiu: I don't really like our `UK_CTOR`.. but would it fit better here? :thinking:

andreittr: Nope; unfortunately the default system allocator is initialized by `ukboot` [during heap init](https://github.com/unikraft/unikraft/blob/staging/lib/ukboot/boot.c#L291) which runs after the constructors. Either way (constructor or inittab) it's still suboptimal to have runtime set this.
But as the comment says, ideally at some point in the future we'll have `uk_alloc_default` be a `constexpr` (and drop the "get" from the name) and then we'd have all this resolvable at compile-time.

andreittr:
Pushed rebase to staging & latest #1608 , including support for `ukfile.mem`, as introduced by #1411 .

lib/nolibc/include/fcntl.h
andreittr: Oof, yeah that's right :/ , indeed apparently all `FALLOC_*` constants are usually defined in the `<linux/falloc.h>` header, which on GNU systems gets included verbatim by glibc's `<fcntl.h>` via `<bits/fcntl.h>`.

Musl, for whatever reason, foregoes this and defines its own (insufficient) subset in `fcntl.h` instead of using kernel headers, and that served as inspiration for our own `FALLOC_*` constants in `nolibc`.
Regardless of libc, file drivers need to have all these definitions on hand, so I guess we need to bring in `falloc.h` under `ukfile` :disappointed: .

lib/ukfs-ramfs/ramfs.c
andreittr: Nice catch, thanks, I believe the intent was to pass the `strlen` to `ramfs_init_sym`, instead of `xlen` as that produces the correct semantics for following `readlink` calls. Will change to `xlen - 1`.
As it is now we may be writing 1 byte out of bounds for long symlinks :eyes: whoops.

andreittr:
Pushed rebase to latest #1608 and #1606 .

andreittr:
Pushed larger pass addressing comments, relocating the `FALLOC_*` constants out of nolibc, and several other issues.

mogasergiu (approved):

andreittr:
Pushed rebase to current staging & latest #1608 + #1607 .

andreittr:
Pushed update to use the ukfs-provided priorities for init instead of hardcoded.

andreittr:
Pushed update adding `exportsyms.uk` file to explicitly express developer intent.

andreittr:

michpappas:

PR #1602, opened 2025-03-19
lib/posix-fd: Move heap-allocated open file descriptions out of fdtab

andreittr:
### Description of changes

This PR moves the responsibility of allocating, managing the lifetime of, and freeing open file descriptions out of `posix-fdtab` and into inlines in the main `posix-fd` API.

This decouples dynamic open file descriptions from the fdtab instance that would have previously managed them, allowing for any library to create kernel-wide dynamic open file descriptions usable by any other library. This is of particular importance after fdtabs themselves have become dynamic instances and not always guaranteed to exist in the current thread's context.
On first iteration, dynamic open file descriptions are allocated with the default system allocator on the heap; this can easily change in the future.

Specifically, this changes the following in the `posix-fd` API, with heap-allocated open file descriptions gated behind a Kconfig option:
- `struct uk_ofile` contains a refcount only if heap-allocated ofiles are enabled
- `uk_ofile_new()` & `uk_ofile_free()` -- allocation and freeing
- `uk_ofile_acquire()` and `uk_ofile_release()` have been gated behind heap-allocated ofiles
- `uk_ofile_release()` now returns `void` and performs cleanup automatically (what the old `uk_fdtab_ret` did)

Relieved of its responsibility to manage dynamic ofiles, `posix-fdtab` has had the following changes:
- `uk_fdtab_new_desc()` has been removed; callers should switch to `uk_ofile_new()`
- `uk_fdtab_ret()` has been removed; callers should release the ofile directly with `uk_ofile_release()`
- the restriction on `uk_fdtab_open_desc` of using only ofiles opened by that fdtab instance has been lifted
- implementation in `fdtab.c` has been cleaned up to use public ofile heap allocation

All callers of both APIs have been patched to reflect the changes.

As a trivial addition to the above, 3 minor fixes are included that touch the same code base; see individual commits for details.

Conflicts with #1619 , whichever merges first mandates a patch to the other.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [x] Updated relevant documentation.

### Base target

 - Architecture(s): [e.g. `x86_64` or N/A]
 - Platform(s): [e.g. `kvm`, `xen` or N/A]
[...]

PR #1610, opened 2025-03-21
lib/posix-vfs: Introduce VFS interface library

andreittr:
### Description of changes

This change introduces posix-vfs, a library providing a posix-like virtual filesystem interface on top of ukfs filesystems.
Its functionality is exposed through a Unikraft-internal API (uk_sys_*), as well as Linux ABI-compatible syscalls.

Various minor prerequisites or related fixes are included as commits.

~~Depends on the following PRs (included as squash commits, ignore during review):~~
~~- #1609~~

This PR's branch serves as the merged, full-stack VFS implementation that can be meaningfully tested.

~~Interacts/conflicts with #1411 ; whichever merges first requires patches to the other.~~

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/nolibc/musl-imported/src/dirent/opendir.c
```diff
@@ -13,7 +14,7 @@ DIR *opendir(const char *name)
 	if ((fd = open(name, O_RDONLY|O_DIRECTORY|O_CLOEXEC)) < 0)
 		return 0;
 	if (!(dir = calloc(1, sizeof *dir))) {
-		__syscall(SYS_close, fd);
+		uk_syscall_r_close(fd);
```
mogasergiu, line 17: I think you should switch to `uk_syscall_do_close` now, unless you do intend for the syscall enter/exit tables to be run here.

lib/nolibc/musl-imported/src/unistd/getcwd.c
```diff
@@ -14,7 +15,7 @@ char *getcwd(char *buf, size_t size)
 		errno = EINVAL;
 		return 0;
 	}
-	long ret = syscall(SYS_getcwd, buf, size);
+	long ret = uk_syscall_r_getcwd(buf, size);
```
mogasergiu, line 18: Ditto

lib/nolibc/musl-imported/src/dirent/readdir.c
```diff
@@ -1,18 +1,19 @@
...
 	offsetof(struct __dirstream, buf) % sizeof(off_t))];
 
 struct dirent *readdir(DIR *dir)
 {
 	struct dirent *de;
-	
+
 	if (dir->buf_pos >= dir->buf_end) {
-		int len = __syscall(SYS_getdents, dir->fd, dir->buf, sizeof dir->buf);
+		int len = getdents64(dir->fd, (void *)dir->buf,
```
mogasergiu: Hmm, why the call to the libc wrapper?

EDIT: Actually, I think this should be the way, for the previous 2 comments above as well: it's a libc (nolibc), so it should call libc wrappers, right?

EDIT 2: Nevermind, let's keep in sync with musl

lib/posix-vfs/Config.uk
```diff
@@ -0,0 +1,83 @@
...
+
+if LIBPOSIX_VFS
+
+config LIBPOSIX_VFS_SYSCALLS
+	bool "Provide filesystem syscalls"
+	depends on !LIBVFSCORE
+	select LIBPOSIX_FDTAB
+	default y
+	help
+		This enables VFS syscalls to be provided by this library in
```
mogasergiu: Nit, but as a convention I believe that kconfig help messages have an indentation of 2 spaces after the indentation of the menu entry they appertain to.

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,231 @@
...
+
+static
+int root_getstat(const struct uk_file *f __maybe_unused,
+		 unsigned int mask __unused, struct uk_statx *arg)
+{
+	UK_ASSERT(f->vol == ROOT_VOLID);
+	*arg = (struct uk_statx){
+		.stx_mask = UK_STATX_NLINK | UK_STATX_TYPE | UK_STATX_MODE |
+			    UK_STATX_UID | UK_STATX_GID | UK_STATX_INO,
+		.stx_blksize = 4096,
```
mogasergiu, line 35: PAGE_SIZE?

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,231 @@
...
+					if (cur + 2 == len) {
+						cur += 2;
+						break;
+					}
+					if (cur + 2 < len &&
+					    path[cur + 2] == '/') {
+						cur += 3;
+						continue;
+					}
+				}
```
mogasergiu, line 100: I may be misreading this, but asking just to be sure. So I guess you did not add a `break` for this case because it is intended to just continue into the `break` out of the while loop? Wouldn't it still exit the switch the same way if you were to conventionally add a break as well as a dummy `default: break`?

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,231 @@
...
+int root_mount(const struct uk_file *f __maybe_unused,
+	       const struct uk_file **target)
+{
+	const struct uk_file *t = *target;
+
+	UK_ASSERT(f == &uk_posix_vfsroot_file);
+	if (t) {
+		const struct uk_file *prev = NULL;
+
+		if (likely(uk_compare_exchange_n(&rootmnt, &prev, t)))
```
mogasergiu: Why did you choose the `likely` instead of the `unlikely` and early return to save some indentation... ? :sweat_smile:

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,231 @@
...
+								 3);
+
+const struct uk_file uk_posix_vfsroot_file = {
+	.vol = ROOT_VOLID,
+	.node = NULL,
+	.ops = &rootfops,
+	.fsops = &rootfsops,
+	.refcnt = &root_refcnt,
+	.state = &root_fstate,
+	._release = uk_file_static_release,
```
mogasergiu: I can't help but notice that in some places you do add `,` for the last statically initialized struct member but in other places you don't... :sweat_smile:

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2025 @@
...
+		int r = uk_fs_mountat(point, &prev);
+
+		if (unlikely(r))
+			uk_pr_warn("Unable to unmount %p: %d\n", point, r);
+		else
+			uk_file_release(prev);
+	}
+}
+
+uk_rootfs_initcall_prio(0x0, vfs_mtab_cleanup, 4);
```
mogasergiu: Would it make sense to have this as `UK_PRIO_EARLIEST` instead? Is there anything that would make sense for rootfs initcalls to come before this? The only one I see is `init_ramfs_sbctx` but all it does is fetch the allocator, should it be moved later?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2025 @@
...
+
+#if CONFIG_LIBPOSIX_PROCESS_CLONE
+
+static
+struct vfs_proc_state *vfs_ctx_duplicate(struct vfs_proc_state *ctx)
+{
+	struct vfs_proc_state *newctx = uk_malloc(uk_alloc_get_default(),
+						  sizeof(*newctx));
+
+	if (newctx) {
```
mogasergiu, line 259: `unlikely` and early return?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2025 @@
...
+	if ((cl_args->flags & CLONE_FS)) {
+		/* Inherit parent's vfs ctx */
+		/* As a compat stop-gap, the raw thread already inherited the
+		 * parent's vfs ctx; we don't need to do anything.
+		 *
+		 * TODO: move inheritance here once stopgap is removed.
+		 */
+		UK_ASSERT(uk_thread_uktls_var(child, vfs_ctx) == ctx);
+		return 0;
+	} else {
```
mogasergiu, line 300: Unecessary `else`?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2025 @@
...
+			uk_file_release(sym);
+		if (UKFS_FILE_ISSYM(target)) {
+			struct uk_fs_path p = uk_fs_readlink(target);
+
+			uk_file_release(f);
+			f = dir;
+			path = p.s;
+			pathlen = p.len;
+			sym = target;
+			symlvl++;
```
mogasergiu, line 1807: Should this maybe also be guarded by `CONFIG_LIBPOSIX_VFS_SYMLIMIT_DEPTH` and be declared with a `__maybe_unused`?

mogasergiu (commented):
Really appreciate the attention to ensuring no references/allocations are lost/leaked on error paths :muscle: 

(Though I must say, our coding styles are polar opposites, I can't help but squint at all the lost opportunities of saving up on indentation, but ok :laughing: )

Anyway, as of this comment, I've finished my first walk of the PR series. Will begin a new walk in the following days now that I have more context. This is light years ahead of vfscore :100:

lib/nolibc/musl-imported/src/dirent/opendir.c
andreittr: Yeah I guess that's the way now.
Just to be clear: the `do` versions are intended for any and all native userspace code, and `r` versions for bincompat only?

lib/posix-vfs/posix-vfsroot/vfsroot.c
andreittr: No, I don't want it linked to anything, I just want an arbitrary, non-zero number that's typically found as a blocksize in irl devices. Could've picked 512. I'll just make it 1 and see what breaks.

lib/posix-vfs/vfs.c
andreittr: But it is used, regardless of `_DEPTH` or `_NUM`, either passed by value or reference to `_vfs_lookupat`, respectively.

lib/nolibc/musl-imported/src/dirent/opendir.c
mogasergiu: The difference between `r` and `do` is that the former also calls the routines from the syscall enter/exit tables before/after calling the actual implementation of the syscall.

For example, on bincompat, the big syscall switch case calls `do` because the entry into the binary syscall already runs the tables. On native, musl calls the `r` variants. IMO we should try to only call the `do` variants internally, though this call is technically in a libc (our `nolibc`) so you could argue it doesn;t hurt having a `r` called here. I have no preference, but if it's a libc then it may as well call `close` directly, as it alraedy calls `open` above anyway.

lib/posix-vfs/posix-vfsroot/vfsroot.c
andreittr: I think adding those 3 lines would have the same behavior, but would be very redundant and akin to littering. Think "unnecessary else" but on all the steroids.

Although tbh, I don't know if this code is special enough to deserve its own bespoke loop, or if I should just use the (slightly less efficient but much clearer) macros from `pathutil.h` and be done with it.

lib/nolibc/musl-imported/src/dirent/opendir.c
andreittr: Ok, so userspace code (libc) does `r`, kernel code does `do` (when it can't make do with `uk_sys_`); got it.

I can't know why musl chose to mix-and-match `open()`, the libcall, and `close()`, the syscall, in one function implementation, but I'm not about to change _anything_  about semantics unless there are very good reasons for it. We want to keep the diff to upstream minimal, not roll our own `DIR *` functions.

mogasergiu: Ok, makes sense, good point. Then I think you should instead maybe call `uk_syscall(SYS_close, fd)`?

lib/posix-vfs/Config.uk
andreittr: Sounds like we need a codebase-wide style fix commit; tab indends far outnumber double-space help strings :disappointed:

lib/posix-vfs/vfs.c
andreittr: I honestly don't know; the hard-coded prio `4` I grabbed from vfscore early in development _just in case_ there was some unspoken assumption about this somewhere, but I haven't really checked if anything else works.

Off the top of my head, this _should_ be fine to run at any time after the last filesystem _path lookup_ has been completed by userspace; that's the earliest point. Open files, etc., should work perfectly fine even after the mtab is cleaned up.
IIRC there isn't any other cleanup code that _depends_ on there being no active mounts, so I don't have a _latest_ point by which this must be done.

lib/nolibc/musl-imported/src/dirent/opendir.c
andreittr: We have `uk_syscall(NUM, ...)` too? :scream: 
_How many ways are there to do a syscall in Unikraft?!_  :joy: 

Jokes aside, the latter might be a better choice whatwith the `SYS_` constants already being in scope.

lib/posix-vfs/Config.uk
mogasergiu: Yep, in fact pretty much all the recent PR's that have touched KConfigs's have also updated the syling.

lib/posix-vfs/posix-vfsroot/vfsroot.c
mogasergiu: > slightly less efficient

nah, this good bro :rofl:

lib/posix-vfs/vfs.c
mogasergiu: I think we should try then to set the prio to 0 or at least find out why it was 4? Indeed I do remember working on this part of the codebase at some point and asking the same question but did not insist at the time as I did not bother. But now since we are about to bring major changes, I think it's a question worth resolving. As for `init_ramfs_sbctx` since it just assigns the allocator, it pretty much looks like it would fit just fine with the other `UK_CTOR`'s instead of the inittab... although I personally don't like the `UK_CTOR`s and don't see their point anymore.

andreittr: Sure thing, I'll just switch to plain `uk_rootfs_initcall` and we'll use something finer-grained when needed.
(insert rant about init/termtab being generally unsuitable, as system startup & teardown are more naturally DAGs rather than ordered queues)

> since it just assigns the allocator

Tbh I'm waiting until the default allocator is a compile-time constant to completely remove the need for `init_ramfs_sbctx`; or _at least_ have the allocator API do a sensible thing like
```
if (alloc == NULL)
    alloc = uk_alloc_get_default();
/* call func with alloc */
```
Then you create a runtime init func only if you want a _runtime selectable_ (via e.g., kernel args) allocator for a specific lib.

lib/nolibc/musl-imported/src/dirent/readdir.c
andreittr: Darn, I no longer remember if I had good reasons to use the libc wrapper instead of `_r` syscalls here. Will check closely and change if warranted.

andreittr:
Pushed rebase to staging + rebase of #1609 + removal of #1602 after its merge + add dependency to #1617 (replacing the fdtab fix commit)

andreittr:
Pushed update to latest #1609 , dropped #1617 as it was merged & minor changes reflecting updates to #1609 .

andreittr:
Pushed rebase to latest #1609 .

lib/nolibc/musl-imported/src/dirent/opendir.c
andreittr: On further thought, no, I think we should keep the `uk_syscall_r_name()` form here, as `uk_syscall_r(SYS_num` not only involves additional runtime matching of `SYS_num` to the actual syscall, but also well-defined runtime behavior on wrong syscall numbers.
Considering nolibc is our kernel-internal minimal libc, a missing `close` syscall here indicates a misconfigured kernel and it's by far more convenient to get a link error here. Same argument applies for `getcwd`.

lib/nolibc/musl-imported/src/dirent/readdir.c
andreittr: Ok, it's pretty evident from the return value check later that it expects syscall-style negative return codes; will change to raw syscall.

lib/posix-vfs/vfs.c
andreittr: This form is here on purpose, as once the compat stop-gap is removed (which btw, I need to check if it's possible now) the `return` on the if-branch will disappear and exec will continue after the `else` block. This serves as a reminder of the intended control flow.

andreittr:
Pushed addressing comments + minor updates.

andreittr:
Pushed fixes to a few issues + public path length/separator find inlines + make ukcpio VFS agnostic.

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2013 @@
...
+ */
+static
+int vfs_check_perms(const struct uk_file *f, mode_t desired)
+{
+	struct uk_statx sx;
+	int r = uk_file_getstat(f, VFS_MODEMASK, &sx);
+
+	if (!r) {
+		UK_ASSERT(sx->stx_mask & VFS_MODEMASK == VFS_MODEMASK);
+		r = uk_fs_checkperm(desired, sx->stx_mode,
```
mogasergiu: I think `sx` is not a pointer. How did this build for you?

lib/posix-vfs/Config.uk
```diff
@@ -0,0 +1,84 @@
...
+config LIBPOSIX_VFS_RELAX_UNLINK
+	bool "unlink(): Treat directories like files"
+
+config LIBPOSIX_VFS_RELAX_RMDIR
+	bool "Allow deletion of non-empty directories"
+
+# Hidden, will autoselect when required
+config LIBPOSIX_VFS_MULTICTX
+	bool
+	default y if LIBPOSIX_PROCESS_CLONE
```
mogasergiu: This will need to be updated to MULTITHREADING I think

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,229 @@
...
+	UK_ASSERT(f == &uk_posix_vfsroot_file);
+	if (t) {
+		const struct uk_file *prev = NULL;
+
+		if (unlikely(!uk_compare_exchange_n(&rootmnt, &prev, t)))
+			return -EEXIST;
+		uk_file_acquire(t);
+	} else {
+		uk_spin_lock(&umount_lock);
+		t = uk_exchange_n(&rootmnt, NULL);
```
mogasergiu, line 151: Why does this have to be atomic as well if already surrounded by locks? (Same for other places like this)

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,229 @@
...
+		uk_spin_unlock(&umount_lock);
+		if (unlikely(!t))
+			return -ENOENT;
+		*target = t;
```
mogasergiu, line 155: Shouldn't this `t` also be acquired?

lib/posix-vfs/posix-vfsroot/vfsroot.c
```diff
@@ -0,0 +1,229 @@
...
+		const char *path, size_t len, unsigned int flags,
+		union uk_fs_lookup_out *out, size_t *nout)
+{
+	const struct uk_file *mnt = NULL;
+	size_t cur = 0;
+
+	UK_ASSERT(f == &uk_posix_vfsroot_file);
+	/* Skip any number of '', '.', and '..' leading components */
+	while (cur < len) {
+		switch (path[cur]) {
```
mogasergiu, line 77: Does this handle `/.` as well? Or could it lead to `-ENOENT`?

lib/posix-vfs/syscalls.c
```diff
@@ -0,0 +1,730 @@
...
+
+/* FS ops */
+
+UK_SYSCALL_R_DEFINE(int, sync)
+{
+	uk_sys_sync();
+	return 0;
+}
+
+UK_SYSCALL_R_DEFINE(int, syncfs, int, fd)
```
mogasergiu, line 83: I see we do not also adopt the `UK_TRACEPOINT`s from vfscore. That's ok for now I guess, I think we should add an issue somewhere to track adding this later somewhere. I think it will prove valuable in the future.

lib/posix-vfs/syscalls.c
```diff
@@ -0,0 +1,730 @@
...
+UK_SYSCALL_R_DEFINE(int, linkat, int, olddfd, const char *, oldname,
+		    int, newdfd, const char *, newname, int, flags)
+{
+	int ret;
+	struct uk_ofile *oldof;
+	struct uk_ofile *newof;
+
+	if (unlikely((ret = _vfs_atfd(olddfd, oldname, &oldof))))
+		return ret;
+	if (unlikely((ret = _vfs_atfd(newdfd, newname, &newof))))
```
mogasergiu, line 481: So if the `_vfs_atfd` above succeeded but this one failed, won't `oldof` remain unreleased?
(Ditto for renameat)

lib/posix-vfs/syscalls.c
```diff
@@ -0,0 +1,730 @@
...
+	if (unlikely(!of))
+		return -EBADF;
+	if (unlikely(!UKFS_FILE_ISDIR(of->file)))
+		return -ENOTDIR;
+	/* Directories may not have the NOSEEK flag set */
+	UK_ASSERT(!(of->mode & UKFD_O_NOSEEK));
+
+	uk_mutex_lock(&of->lock);
+	dcur = of->pos;
+	ret = uk_sys_getdents64(of->file, &dcur, dirp, count);
```
mogasergiu: If `uk_sys_getdents64` failed, is it ok to restore afterwards to the `dcur` which `uk_sys_getdents64` may have modified during its execution? Maybe we should use another variable to pass as argument to `uk_sys_getdents64` and in case of success assign pos to that and otherwise to the old dcur?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2013 @@
...
+	const struct uk_file *ret;
+	size_t pos;
+
+	/* Skip leading '/' */
+	while (len && path[0] == '/') {
+		path++;
+		len--;
+	}
+	/* We only require fsroot for pointer comparison; can simply read val */
+	fsroot = vfs_ctx->fsroot;
```
mogasergiu, line 575: I am not sure I understand how the above comment clarifies the need to not protect this with lock/unlock. Do we only need the locks when acquire/release-ing fsroot?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2013 @@
...
+
+static struct vfs_mtab mtab = {
+	.lock = UK_MUTEX_INITIALIZER(mtab.lock),
+	.top = 0
+};
+
+#if CONFIG_LIBPOSIX_VFS_MOUNT_CLEANUP
+
+static
+void vfs_mtab_cleanup(struct uk_term_ctx *tctx __unused)
```
mogasergiu, line 71: I guess we don't need to protect this with mtab lock cause it is during init? If so, can you add a comment?

lib/posix-vfs/vfs.c
```diff
@@ -0,0 +1,2013 @@
...
+static
+const struct vfs_mtab_entry *vfs_parent_mount(const struct uk_file *f)
+{
+	const struct uk_file *d = f;
+	const struct vfs_mtab_entry *ret = NULL;
+	const struct uk_file *parent __maybe_unused;
+	unsigned int lflags = UKFS_LOOKUP_IGNMNT | UKFS_LOOKUP_NO_MNTAUX;
+
+	/* Clearing flag for later UK_ASSERT */
+	UK_ASSERT((lflags &= ~UKFS_LOOKUP_NO_MNTAUX));
```
mogasergiu, line 106: This looks... dangerous? I am not I understand why this needs to happen for assertions only.

lib/posix-vfs/vfs.c
andreittr: Whoops, guess I only tested with `CONFIG_LIBPOSIX_VFS_PERMISSIONS=n` since `uk_fs_checkperm` is still a no-op :sweat_smile: .

lib/posix-vfs/posix-vfsroot/vfsroot.c
andreittr: The spinlock is meant to make reading & taking ref atomic in lookup, similar to [here](https://github.com/unikraft/unikraft/pull/1609#discussion_r2166632749). Here it's used to sync with lookup, so that it either grabs and refs a valid `rootmnt` or gets a NULL.

The `uk_exchange_n` here is meant to synchronize with the `uk_compare_exchange_n` in the `if` branch above, which doesn't use the `umount_lock`.

More succintly, we have 3 ops: lookup, mount, unmount, which must be independently synchronized w.r.t. each other:
- mount vs unmount: this is a pointer write/cmpxchg which can be synced with atomic ops
- lookup vs unmount: this is a ptr read + conditional refcount acq vs pointer write; needs a (spin)lock

Note that lookup vs mount need no sync, as any possible ordering is consistent:
- if `rootmnt != NULL` the mount op will always fail; lookup can happen at any time and return the same
- if `rootmnt == NULL`:
  - lookup read before mount: lookup happy with NULL, mount takes ref on new root
  - mount before lookup read: lookup takes ref on new root, mount takes ref on new root

In any case there's no risk of inconsistency or UAF.

Now you _could_ argue that this may be overly fine-grained and just mutexing all 3 ops would be faster, but without performance data to back up that argument, I'd rather keep the actual dependencies clear in code.

lib/posix-vfs/posix-vfsroot/vfsroot.c
andreittr: No, because it is the previous (non-null) value of `rootmnt`, and the only way it could have gotten there is through a previous mount operation, which takes a reference on success. The return path here simply passes on the ownership of that reference to the caller.

That said, you pointing this out made me see an issue with the mount code and a possible UAF in some (very unlucky pre-emptive) cases; the `uk_file_acquire(t)` should run before the cmpxchg to be correct. Will update :+1: .

lib/posix-vfs/posix-vfsroot/vfsroot.c
andreittr: Whoops, I think I forgot to handle that case below; thanks :+1: .

lib/posix-vfs/syscalls.c
andreittr: I agree, that's totally on me and not quite understanding the `UK_TRACEPOINT` system, nor how to meaningfully use/test it. So instead of a half-baked copy-paste job I skipped them altogether. (We always have the vfscore code around to serve as inspiration for when we do port it)

I would be happy to finally find out what they're about and re-add them, possibly to `posix-fdio` and others as well.

lib/posix-vfs/syscalls.c
andreittr: Ooof, yeah, rookie mistake on my end :sweat_smile: . Thanks.

lib/posix-vfs/syscalls.c
andreittr: The `dcur` in/out arg here lets the driver (via `uk_fs_listdir`) directly update the file position (think `lseek`) of an open directory, something pretty unique in the ukfile/ukfs stack; usually file position can be entirely handled by the posix layer, including any guarantees on it changing or not on error.

While not always documented in manpages, AFAIK all file-related syscalls guarantee to not touch the file position on failure. This is what userspace expects of us and what IMO the ukfs API should explicitly ask of drivers. It's not too much to ask and means better chances for optimization.
Indeed I could make this more explicit by passing `&of->pos` directly to `uk_sys_getdents64`.

Going through ukfs, I see that "not touching curp" [not explicitly required](https://github.com/unikraft/unikraft/pull/1595/files#diff-893f022b7a0f07a7709bdc598d4b6bf9643e22da7063e0d668153c0c38472901R268); looking at ramfs, it's [not clear it's respected either](https://github.com/unikraft/unikraft/pull/1609/files/dcbe4bbcad81112009d157878d36cc525bba3de8..39d1086bb4a15f190cf7093038fbf7c11ca676d7#diff-35e93af7eedc209bef6cb0bdb0648bf0c75ae5ebf376679624553d6f970c359dR1003); will push a change to both to enforce this.

lib/posix-vfs/vfs.c
andreittr: Yes, the fsroot/cwd spinlocks on vfs state are there to make atomic (1) reading the pointer, and (2) if ptr acquire refcount (then after unlock use ptr as you see fit). This is to prevent a UAF, similar to [here](https://github.com/unikraft/unikraft/pull/1609#discussion_r2166632749) and in the vfsroot code.
If we only use the numerical value of the pointer, we don't need a ref, so no locking.

lib/posix-vfs/vfs.c
andreittr: Ok, yeah, I agree this looks sketchy af :smile: , so I think a bit of context is needed:

The `UKFS_LOOKUP_NO_MNTAUX` flag exists as an optimization, so that drivers need not return the _mount point_ node in addition to the _mounted_ node on lookup; in 99% of cases you want only the latter; returning the former may involve a lot of work and memory allocation, which is a darn shame if the very first thing you do with it is `uk_file_release()` and cause it to be freed.
Memory churn is bad; completely predictable and avoidable memory churn is worse. Thus this flag and its use here.

_However_ we sometimes do want to get a ref to the mount point, for e.g., sanity-checking purposes like asserts. In that case (and only that case) do I want `lflags` to have `NO_MNTAUX` cleared -- else non-assert builds would waste time producing a ref that immediately gets killed.

That said, I completely agree this looks scary and pretty brittle if we change the initval of `lflags`. If you know of a nicer way of evaluating a statement only if `UK_ASSERT` works (without `#ifdefs`) I'm happy to change over.

andreittr:
Pushed rebase to latest #1609 + required changes for compat.

andreittr:
Pushed addressing comments.

lib/posix-vfs/vfs.c
mogasergiu: Nothing comes to mind but maybe you can add a more detailed comment repeating what this flag does and why you do this in the assertion.

andreittr:
Pushed rebase to staging and latest #1609 + required compat changes.

mogasergiu (approved):

andreittr:
Pushed rebase to current staging & latest #1609 .

andreittr:
Pushed fix of 2 issues:
- using the old posix-process `_CLONE` config instead of `_MULTITHREADING`
- reference miscounting in `vfs.c`

andreittr:
Pushed update adding `exportsyms.uk` files to explicitly express developer intent.

andreittr:
Pushed rebase to latest staging after merging of all prerequisites.

michpappas:

PR #1618, opened 2025-03-25
lib/posix-process: Define `libc` wrappers for `clone`/`vfork`

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

NOTE: This PR begins with several minor fixes/improvements as prerequisites

`clone`:
In order to successfully call the `clone` system call through a libc
wrapper we need to be able to have an assembly sequence that both
translates the arguments passed by the libc wrapper to those of the
underlying syscall, since their signatures differ, as well as redirect
the parent and child accordingly upon exit from said syscall: child
must run its requested function and argument and parent must return
to the clone caller with an unscathed register context.

Said assembly sequence must be aware of the signature differences
between architectures, e.g. ARM64 vs x86_64 signatures of the
clone system call.

`vfork`:
As other libc's also state, it is impossible to write the libc
wrapper of the `vfork` system call in C since the child might
end up reusing and overwriting the parent's return address when it
pops it and then calls `execve`, leaving the parent to return to an
invalid state - typically would overwrite it with the return address
from execve, making it look as if the parent returned from execve.

Other libc's solve this by writing the libc wrapper as something
among the lines of:
popq    %<register that the syscall guarantees to preserve>
movq    $SYS_vfork, %rax
syscall
pushq   %<register we popped return address into 3 lines above>
ret

On ARM64 as well as others, clone() with the flags
`SIGCHLD | CLONE_VM | CLONE_VFORK` is called instead but let's just
still use our `vfork` implementation since it essentially also does
the same exact thing.

However, as a Unikernel with function calls for system calls we do
not need to care about any of that, as our execenv prologue can store
and restore everything without touching the stack. Thus, we can write
our native libc wrappers as a basic jump/branch to said prologue.


mogasergiu:
Depends-on: [#86](https://github.com/unikraft/lib-musl/pull/86)

lib/syscall_shim/arch/x86_64/include/arch/syscall_prologue.h
```diff
@@ -47,8 +47,16 @@
...
 		"pushq	$(0x10)\n\t"					\
-		"/* Push saving original rsp stored in r11 */\n\t"	\
+		"/* Push saving original rsp - 8 stored in r11 */\n\t"	\
 		"pushq	%r11\n\t"					\
+		"/* Above pushed rsp is actually the caller's\n\t"	\
+		" * rsp minus 8, because the call instruction\n\t"	\
+		" * pushes the address the ret instruction is\n\t"	\
+		" * supposed to return to. This means that to truly\n\t"\
+		" * mimic a trap/syscall we must store/restore\n\t"	\
+		" * the rsp we were given, plus 8.\n\t"			\
```
andreittr, line 63: The code makes sense, but I had to read the comments through a couple of times before getting what's going on. It might parse easier if this large comment were moved earlier, where we save the current sp (`movq %rsp, %r11`) and reworded to clearly explain what's going on with the stack. Something like

"Our stack top now contains a return address pushed by `call`; this must be ignored when saving the stack pointer to the interrupt return structure, but taken into account when we actually return execution", with smaller, single-single sentence reminders here when we `addq`, and later when we `subq`.

(If my rewording completely missed the point, then I believe it's evidence enough that the comment could do better :smile: )

lib/syscall_shim/arch/x86_64/include/arch/syscall_prologue.h
```diff
@@ -111,6 +119,11 @@
 		"popq	%rbx\n\t"					\
 		"/* Restore rsp from where it was stored */\n\t"	\
 		"movq   104(%rsp), %rsp\n\t"				\
+		"/* Now subtract those 8 bytes we previously added\n\t"	\
+		" * because the caller/ret do not expect the rsp to\n\t"\
+		" * not point to the actual return address.\n\t"	\
+		" */\n\t"						\
```
andreittr, line 148: Tangential to above, could use simplification & removal of double-negative: "[ ... ] do not expect [...] to not point":

"Adjust saved stack to original value; `ret` expects return address on top of stack"

arch/x86/x86_64/include/uk/asm/lcpu.h
```diff
@@ -66,6 +66,22 @@
...
+ */
+
+#define __fn_rarg0		rdi
+#define __fn_rarg1		rsi
+#define __fn_rarg2		rdx
+#define __fn_rarg3		rcx
+#define __fn_rarg4		r8
+#define __fn_rarg5		r9
+
+#define __fn_rret		rax
```
andreittr: Can I assume we are explicitly discounting the case of return value in `%rax:%rdx`?
AFAICT these macros are used for syscalls, and all syscalls return at most a machine word, so _should be fine_.

lib/posix-process/clone.c
```diff
@@ -593,12 +593,94 @@ UK_LLSYSCALL_R_E_DEFINE(int, clone,
...
+	"call	uk_syscall_e_clone\n\t"
+	"/* Let the parent just return */\n\t"
+	"test	%eax, %eax\n\t"
+	"jnz	1f\n\t"
+	"/* The child should now just call the saved fn(arg) */\n\t"
+	"movq	%rbp, %rdi\n\t"
+	"/* Mark the outermost frame (avoid unwinding beyond this)\n\t"
+	"xorq	%rbp, %rbp\n\t"
+	"call	*%rbx\n\t"
+	"/* If managed to return from fn(arg) we must call exit(x0) */\n\t"
```
andreittr: What is `x0` in `exit(x0)`?
- It is a typo and intended is `exit(0)`? In that case you need a `xor %edi, %edi` before the call
- Is it intended as `exit(rax)` (i.e., return code of `fn`)? Then you need a `mov %rax, %rdi` before the call.

lib/posix-process/clone.c
```diff
@@ -593,12 +593,94 @@ UK_LLSYSCALL_R_E_DEFINE(int, clone,
...
+	"/* tls */\n\t"
+	"movq	%r9, %r8\n\t"
+	"call	uk_syscall_e_clone\n\t"
+	"/* Let the parent just return */\n\t"
+	"test	%eax, %eax\n\t"
+	"jnz	1f\n\t"
+	"/* The child should now just call the saved fn(arg) */\n\t"
+	"movq	%rbp, %rdi\n\t"
+	"/* Mark the outermost frame (avoid unwinding beyond this)\n\t"
+	"xorq	%rbp, %rbp\n\t"
```
andreittr: You can save a byte by doing `xor %ebp, %ebp`, and in general making use of the defined behavior of zeroing out the upper half of 64bit regs when their lower halves are written to.
```
   0:	48 31 ed             	xor    %rbp,%rbp
   3:	31 ed                	xor    %ebp,%ebp
```

andreittr (commented):
On 1st pass looks ok modulo some comments.

arch/x86/x86_64/include/uk/asm/lcpu.h
mogasergiu: Doesn't hurt to add it. Will do.

lib/posix-process/clone.c
mogasergiu: Yup, the latter, good catch.

lib/posix-process/clone.c
mogasergiu: I've been told this so many times and I still lose track of it when I am writing assembly :rofl: .
Will do, thanks! :+1:

mogasergiu:
Rebase force push

lib/posix-process/Config.uk
```diff
@@ -26,18 +26,19 @@ config LIBPOSIX_PROCESS_SIGNAL
 	select LIBSYSCALL_SHIM
 	select LIBUKLOCK
 
-config LIBPOSIX_PROCESS_CLONE
+menuconfig LIBPOSIX_PROCESS_CLONE
```
michpappas: This is a bit out of scope of this PR and I'm about to open a PR that reworks the config options of `libposix-process`, which - among others - deprecates `LIBPOSIX_PROCESS_CLONE` in favor of `LIBPOSIX_PROCESS_MULTITHREADING`, and also moves `PREFER_CHILD` under `LIBPOSIX_PROCESS_MULTITHREADING` too. So maybe we can save that change for that PR?

lib/posix-process/clone.c
```diff
@@ -593,12 +593,95 @@ UK_LLSYSCALL_R_E_DEFINE(int, clone,
...
 	  int flags __unused, void *arg __unused,
-	  ... /* pid_t *parent_tid, void *tls, pid_t *child_tid */)
-{
-	/* TODO */
-	errno = EINVAL;
-	return -1;
-}
+	  ... /* pid_t *parent_tid, void *tls, pid_t *child_tid */);
+
+#if CONFIG_ARCH_X86_64
```
michpappas: Is there any reason these are not under `lib/posix-process/arch/<arch>/clone.S`?

lib/posix-process/vfork.c
```diff
@@ -16,6 +16,25 @@
 
 #include "process.h"
 
+#if UK_LIBC_SYSCALLS
```
michpappas: Same as in `clone()`, these should be under `lib/posix-process/arch/<arch>/vfork.S`.

lib/posix-process/clone.c
```diff
@@ -593,12 +593,95 @@ UK_LLSYSCALL_R_E_DEFINE(int, clone,
...
+	"ret\n\t"
+);
+#elif CONFIG_ARCH_ARM_64
+__asm__(
+	".global clone\n\t"
+	"clone:\n\t"
+	"/* Save fn and arg in callee-saved registers, but first make sure\n\t"
+	" * we can also restore them before returning in the parent.\n\t"
+	" */\n\t"
+	"mov	x20, x0\n\t"
```
michpappas: First save and then corrupt?

michpappas (changes requested):
Nice set of changes! In addition to the inline comments, I would suggest that you added a noticable banner with cross-referencing comments in the duplicated code here and in musl to make sure that anyone updating one keeps the other one in sync too.

On the asm wrappers I focused on arm64 that I'm more familiar with, as @andreittr covered x86_64 already.

Some nits below:

As general remark, as I have also noticed in other PRs, please refrain from using markdown in the title of PRs, as that causes the PR formatting check to fail as you can see below. To make things even worse, the script ignores edits, so even if it's corrected, the error will persist 🤦🏼‍♂️ 

Finally some small things I noticed in the commit messages:

In `lib/syscall_shim/arch/x86_64: Save proper rsp in execenv prologue`:

> to actual users of thie execenv
"to actual users of this execenv"

In `lib/posix-process: Define libc wrappers for vfork`:

> As other libcs also state

The code this patch applies to is not a libc though. It would be nice to provide links to the relevant code as references. A possible rephrase could be:

"As stated in the implementation of libc [1] and musl [2]..."

lib/posix-process/Config.uk
mogasergiu: OK I can drop this commit.

lib/posix-process/clone.c
mogasergiu: Just laziness. In fact, Andrei also suggested I do that. Makes sense to me. It would also make source debugging easier.

lib/posix-process/clone.c
mogasergiu: Yes I like my crash dumps confused.

... No idea how I ended up misplacing that `stp`. I made the same mistake in the musl PR. Will change there as well. :+1:

michpappas (approved):

andreittr (approved):
I much rather like separate asm files for the wrappers, nice change.

Shame `UK_LIBC_SYSCALLS` is not a true config option but computed by `syscall_shim` :disappointed: ; we could've saved ourselves an `#ifdef` and put all conditions for the asm files entirely in the Makefile. Oh well :smile: 

michpappas (approved):

lib/posix-process/arch/x86_64/clone.S
```diff
@@ -0,0 +1,74 @@
...
+
+	/* stack is already in rsi */
+
+	/* parent_tid */
+	movq	%r8, %rdx
+
+	/* child_tid: this is passed on the stack as it does not fit
+	 * in the registers used for the arguments
+	 */
+	movq	24(%rsp), %r10
```
andreittr, line 50: Oof, having to place the 4th argument of a function call in the register used by syscalls will 100% be something that trips future readers up, some of whom might decide to "fix" it to `rcx`.
I (vaguely) understand it's to do with execenv syscalls being special and requiring binary args, but I believe future people deserve a very explicit warning saying that `r10` is here on purpose and why.

andreittr (commented):
Code looks fine, shame we didn't catch the wrong stack address in previous review :disappointed: .
I've got only one documentation/comment request.

andreittr (approved):
Thanks for the changes, looks good!

michpappas:

PR #1619, opened 2025-03-25
lib/posix-process: Add `signalfd`/`signalfd4` support

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Add support for the `signalfd`/`signalfd4` syscalls which will     
allow callers to add monitored signal files to the current process,
allowing them to be notified of pending signals.


lib/posix-process/signal/signal.h
```diff
@@ -9,6 +9,7 @@
 
 #include <limits.h>
 #include <signal.h>
+#include <stdbool.h>
```
andreittr, line 12: Do the signal file additions require this header? I don't see any `bool`s being used there :thinking: 
Or is this a fix for an unrelated bool use somewhere else in the file?

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,107 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PROCESS_SIGNAL_H__
+#error Do not include this header directly
+#endif
+
+#include <uk/file.h>
+#include <uk/posix-fd.h>
```
andreittr, line 13: I believe you can remove the `posix-fd` include; this file doesn't look like it uses any types/funcs defined there.

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,107 @@
...
+ *   Pointer to base file structure
+ *
+ * @return
+ *      Containing uk_signal_file structure
+ */
+static inline
+struct uk_signal_file *uk_file_to_signal_file(const struct uk_file *f)
+{
+	return __containerof(f, struct uk_signal_file, f);
+}
```
andreittr: As a file driver you have complete freedom over the values in the `.vol` and `.node`, and I noticed later in the implementation that you store the mask in node.
That's perfectly fine, but you could get even cleaner code by storing the `struct uk_signal_file *` in `f->node` instead; that way `f->node` provides all the internal runtime context of a sigfile for `read()`, `poll()`, and signal delivery ops.

Of course it's a design freedom, and it's up to you how you structure things. I just want to make sure you're aware of all possibilities.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+{
+	struct signalfd_siginfo *si_arr;
+	struct posix_thread *pthread;
+	struct posix_process *pproc;
+	__sz sn, ninfo = 0, si_len;
+	struct uk_signal *sig;
+	uk_sigset_t mask;
+
+	if (unlikely(!FILE_IS_SIGNAL(f)))
+		return -EINVAL;
```
andreittr: You can safely turn this into an assert.

The only way this function is called with `f` is if `f->ops->read == sigfile_read`. Because both files as well as their ops tables are immutable once created, this can only happen if a non-signal file has signal file ops -- indicative of either memory corruption or an unambiguous kernel bug that is not recoverable.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+	/* Handle signals sent with sigqueue regardless of signal number */
+	if (si->si_code == SI_QUEUE || si->si_code == SI_USER) {
+		ssi->ssi_int = si->si_value.sival_int;
+		ssi->ssi_ptr = (uint64_t)si->si_value.sival_ptr;
+	}
+}
+
+static ssize_t sigfile_read(const struct uk_file *f,
+			    const struct iovec *iov, size_t iovcnt __unused,
+			    size_t off __unused, long flags __unused)
```
andreittr: `iovcnt` and `off` are being used in args validation, you can remove the `__unused` attrs.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+
+	mask = *(uk_sigset_t *)f->node;
+
+	pthread = uk_pthread_current();
+	pproc = uk_pprocess_current();
+
+	/*
+	 * signalfd is only meant for read() base syscall variant so there
+	 * is only one buffer.
+	 */
```
andreittr: While I (think I) understand what you meant here -- signalfd's `read` can only output into one contiguous buffer by design -- the wording suggests that it wouldn't work with other read variants even if given the correct args (which I _hope_ is not the case).

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+		uk_signal_free(pthread->_a, sig);
+
+		ninfo++;
+
+		/* Do we have enough room left in caller's buffer? */
+		if (si_len - ninfo * sizeof(*si_arr) < sizeof(*si_arr))
+			break;
+	}
+
+	/* No pending signals were queues - try again */
```
andreittr: Typo: "queued" I assume

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+		/* Free */
+		priv = uk_file_to_signalfd_priv(f);
+		pprocess_signal_file_del(pproc, &priv->sigf);
+		uk_free(priv->a, priv);
+	}
+}
+
+static struct uk_file *uk_sigfile_create(const uk_sigset_t *mask,
+					 __sz masksz __unused,
+					 int flags __unused)
```
andreittr: (1) If `masksz` and `flags` are unused, and this is a static function, why keep them?
(2) If this is a static func, why is it prefixed with `uk_` unlike other sigfile funcs?
(3) You're a file driver that's including libc types by design, why are you using `__sz`? Is that mandated by some other code this interacts with?

Did you intend for this func to be public (at some future point)?

Also: If you intend to export the sigfile API, consider adding a
```c
void uk_sigfile_set_mask(const struct uk_file *f, const uk_sigset_t *mask);
```

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+	/*
+	 * It is not possible to receive SIGKILL or SIGSTOP signals via a
+	 * signal file descriptor; these signals are silently ignored if
+	 * specified in mask.
+	 */
+	uk_sigcopyset(&priv->sigf.mask, mask);
+	uk_sigdelset(&priv->sigf.mask, SIGKILL);
+	uk_sigdelset(&priv->sigf.mask, SIGSTOP);
+
+	pproc = uk_pprocess_current();
```
andreittr: (This is assuming you intend to make this function public; disregard if it stays private)

Might be worth moving this higher up in the function and erroring/asserting that this func is called only in a pprocess context, and not from a raw kernel thread.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+		      int, flags)
+{
+	return uk_sys_signalfd(fd, mask, masksz, flags);
+}
+
+#if UK_LIBC_SYSCALLS
+int signalfd(int fd, const sigset_t *mask, int masksz)
+{
+	return (int)uk_syscall_e_signalfd((long)fd, (long)mask, (long)masksz);
+}
```
andreittr: Hmm, my manpages declare libc `signalfd` as
```c
#include <sys/signalfd.h>

       int signalfd(int fd, const sigset_t *mask, int flags);
```
not `int masksz`, I think you're misusing the args and it should be
```c
return (int)uk_syscall_e_signalfd4((long)fd, (long)mask, (long)(sizeof(*mask)), flags);
```

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+	pproc = uk_pprocess_current();
+	pprocess_signal_file_add(pproc, &priv->sigf);
+
+	return &priv->sigf.f;
+}
+
+#if CONFIG_LIBPOSIX_FDTAB
+static int uk_sys_signalfd_create(const uk_sigset_t *mask,
+				  __sz masksz,
+				  int flags)
```
andreittr: A few words regarding the `uk_sys_*` namespace (I am assuming you're preparing to have these public one day, otherwise pls dump the `uk_sys_` prefix entirely):
It's intended to contain Unikraft-internal versions of actual syscalls that have the exact same semantics, modulo using the appropriate kernel refs instead of user refs (e.g., `const struct uk_file *` or `uk_ofile *` instead of `int fd`) where possible. 
The naming should also match existing syscalls (with some exceptions for silly things and exceptions on top of those), but generally: if there's no syscall, you don't get a `uk_sys_*` name; you have to use your own namespace.

So in this specific instance, there's only the `signalfd` syscall that does everything (technically `signalfd4`, but that falls under the aforementioned "silly things"), so what one would expect here is just
```c
int uk_sys_signalfd(int fd, const uk_sigset_t *mask, size_t masksz, int flags);
```
The two helpers IMO should not be exposed under `uk_sys_`, as they have no corresponding syscall.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,345 @@
...
+	/*
+	 * It is not possible to receive SIGKILL or SIGSTOP signals via a
+	 * signal file descriptor; these signals are silently ignored if
+	 * specified in mask.
+	 */
+	uk_sigcopyset(mask, new_mask);
+	uk_sigdelset(mask, SIGKILL);
+	uk_sigdelset(mask, SIGSTOP);
+
+	return 0;
```
andreittr: I believe you forgot to release the open file after you're done (`uk_fdtab_ret` before #1602 , `uk_ofile_release` after).

andreittr (commented):
A rough first pass over only the file parts of the PR. There are some broader issues with synchronization and lifetime, but I'll leave those for a more in-depth review after a rebase.

mogasergiu:
Force pushed rebase

lib/posix-process/signal/signal.h
mogasergiu: Nope, this should've gone in signal_file.h
The changes had a very different placement initially, everything was defined in signalfd.c while testing/developing.

lib/posix-process/signal/signal_file.h
mogasergiu: Sounds ok to me. I even had it like this initially but then I changed it to follow the pattern of timerfd and only have in node the private data that matters. Will change :+1:.

lib/posix-process/signal/signalfd.c
mogasergiu: Ah, right, I copy-pasted the checks from eventfd.c after writing the function :sweat_smile:

lib/posix-process/signal/signalfd.c
mogasergiu: Will update comment to make it more clear. What I was referring to was that we are assuming we are not just called in a readv with split buffer.

andreittr: Ok, but is this assumption documented (or at least observed) in Linux, or is it a limitation we introduce? Either way is fine, but the comment must explicitly say which.

lib/posix-process/signal/signalfd.c
mogasergiu: for 1) and 2), they are artifacts from the time i was initially writing this

3) I'd rather use our own internal types and use libc types in the syscall entry wrappers and convert if needed. Maybe I should add a cast just to make it clear, even if it does nothing. I want to use our internal types internally because it is inevitable that we will call kernel internal functions.

lib/posix-process/signal/signalfd.c
mogasergiu: Will stay private.

lib/posix-process/signal/signal.h
mogasergiu: Oh nevermind, it is indeed used in a later commit. I'll add it in both files in the separate commits.

lib/posix-process/signal/signalfd.c
andreittr: That's fair; if this func stays private and you don't intend to export the sigfile API, feel free to use whatever mix of types suits you best.
I'm only gatekeeping the public `uk_file`-related APIs to use userspace types; everyone else carry on :smile:

lib/posix-process/signal/signalfd.c
mogasergiu: Heh, this is funny, I was about to say "observed in linux"... but the version I used for reference did not have this [commit](https://lore.kernel.org/linux-kernel/8de27ff2-d53f-4900-b755-6a05bff98f78@kernel.dk/T/)... will change to use the iov API.

lib/posix-process/signal/signalfd.c
mogasergiu: Perfectly justified, agreed.

lib/posix-process/signal/signalfd.c
mogasergiu: No idea what happened, I guess miscopied from the signature of signalfd4.

lib/posix-process/signal/signal_file.h
mogasergiu: Hmm, seems I included it because of UKFD_POLLIN - so it should go into signal.h instead

andreittr:
Conflicts/interacts with #1602 , whichever merges first requires patches to the other.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,342 @@
...
+int signalfd(int fd, const sigset_t *mask, int flags)
+{
+	return (int)uk_syscall_e_signalfd((long)fd, (long)mask, (long)flags);
+}
+#endif /* UK_LIBC_SYSCALLS */
+
+UK_LLSYSCALL_R_DEFINE(int, signalfd,
+		      int, fd,
+		      const uk_sigset_t *, mask,
+		      int, flags)
```
andreittr: The args for the signalfd syscall are:
```c
int ufd, sigset_t *user_mask, size_t sizemask
```
Only `signalfd4` takes flags; both of them take `sizemask` as 3rd arg. [Docs for reference](https://www.man7.org/linux/man-pages/man2/signalfd.2.html#VERSIONS)

lib/posix-process/signal/signalfd.c
andreittr: The wrapper is still wrong after the change, because you had the wrong signature for the `signalfd` syscall; see previous comment / code immediately below. My previously suggested version should work correctly though.

lib/posix-process/signal/signalfd.c
```diff
@@ -0,0 +1,342 @@
...
+	return signalfd_set_mask(fd, mask);
+}
+
+UK_LLSYSCALL_R_DEFINE(int, signalfd4,
+		      int, fd,
+		      const uk_sigset_t *, mask,
+		      size_t, masksz,
+		      int, flags)
+{
+	return uk_sys_signalfd(fd, mask, flags);
```
andreittr: It's OK if our internal signal implementation has a fixed size for a sigset and we do not pass `masksz` on; however, because this value is coming from userspace, we should at the very least validate it (`masksz >= sizeof(uk_sigset_t)`) to prevent over-reads.

(As a tangent, `if (unlikely(!mask)) return -EFAULT;` wouldn't hurt either. Might make sense to have `uk_sys_signalfd` take the masksz and perform these validations itself)

andreittr (changes requested):

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,334 @@
...
+}
+
+UK_LLSYSCALL_R_DEFINE(int, signalfd4,
+		      int, fd,
+		      const uk_sigset_t *, mask,
+		      size_t, masksz,
+		      int, flags)
+{
+	if (unlikely(masksz != sizeof(uk_sigset_t)))
+		return -EINVAL;
```
andreittr: This condition will break in bincompat mode; both glibc and musl define a sigset that's 128 bytes / 1024 bits long (funnily enough both of them as a `struct` containing a `long` array) and will pass that exact 128 byte masksz to your syscall.
That's why I said to validate that `masksz >= sizeof(uk_sigset_t)` not exact equality.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,334 @@
...
+
+	return uk_sys_signalfd(fd, mask, flags);
+}
+
+#if UK_LIBC_SYSCALLS
+int signalfd(int fd, const sigset_t *mask, int flags)
+{
+	return (int)uk_syscall_e_signalfd((long)fd, (long)mask, (long)flags);
+}
```
andreittr, line 348: The libc wrapper is still wrong yet again, please fix as advised.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,334 @@
...
+#endif /* UK_LIBC_SYSCALLS */
+
+UK_LLSYSCALL_R_DEFINE(int, signalfd,
+		      int, fd,
+		      const uk_sigset_t *, mask,
+		      size_t, masksz)
+{
+	if (unlikely(masksz != sizeof(uk_sigset_t)))
+		return -EINVAL;
```
andreittr: Same issue with `masksz` validation as above. Also, unnecessary code duplication.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,334 @@
...
+
+#if CONFIG_LIBPOSIX_FDTAB
+static int signalfd_create(const uk_sigset_t *mask, int flags)
+{
+	unsigned int mode = O_RDONLY | UKFD_O_NOSEEK;
+	struct uk_file *sigf;
+	int fd;
+
+	if (unlikely(!mask))
+		return -EFAULT;
```
andreittr: Why do you only check `mask` here and not in `signalfd_set_mask`? Or, as previously advised, in the common wrapper?

andreittr (changes requested):

lib/posix-process/signal/signal_file.c
mogasergiu: Oops, forgot to update this one.

lib/posix-process/signal/signal_file.c
mogasergiu: The thing with masksz, I am not sure how ok this is. There is some incompatibility somewhere between how we do it and what the mask should look like maybe.

andreittr: Oof, disregard, it seems that libcs define larger sigsets than the kernel supports for various reasons.
That would be fine, if they passed the size of that struct onto the kernel; they don't, they always pass an arch-specific constant that the kernel checks _by exact value_.
That would be fine if just an observed behavioral quirk, but unfortunately it's [documented behavior](https://www.man7.org/linux/man-pages/man2/sigwaitinfo.2.html#VERSIONS) for signal syscalls to receive the _exact_ value the kernel expects; even if the documentation is wrong about what exact value that is :roll_eyes: .

mogasergiu:
Force-push rebase following merge of #1602 
+ `s/uk_fdtab_ret/uk_ofile_release` in `signal_file.c`

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,354 @@
...
+	return 0;
+}
+
+static inline int uk_sys_signalfd(int fd, const uk_sigset_t *mask,
+				  size_t masksz, int flags)
+{
+	if (fd == -1)
+		return signalfd_create(mask, masksz, flags);
+
+	return signalfd_set_mask(fd, mask, masksz);
```
andreittr: Is there a reason we don't validate `mask` and `masksz` here in `uk_sys_signalfd` instead of duplicated checks in `_create` and `_set_mask`? AFAICT neither of the two are called anywhere else and likely don't need independent arg validation.

lib/nolibc/musl-imported/include/sys/signalfd.h
```diff
@@ -0,0 +1,49 @@
+#ifndef _SYS_SIGNALFD_H
```
andreittr, line 1: Could you please add a few lines to the commit message importing this file detailing its provenance? It really helps us keep track of imported code. You can see examples in the commit history, but tl;dr something like:
```
Source of import:
Repository: https://git.musl-libc.org/cgit/musl
Tag: v1.2.4 (f5f55d65)
Path: include/sys/signalfd.h
```
(adapted to whatever origin & tag you used)

lib/posix-fdtab/include/uk/posix-fdtab.h
```diff
@@ -69,7 +69,7 @@ int uk_fdtab_open_desc(struct uk_ofile *of, unsigned int mode);
 /**
  * Gets the open file description associated with descriptor `fd`.
  *
- * Users should call uk_fdtab_ret when done with the open file reference.
+ * Users should call uk_ofile_release when done with the open file reference.
```
andreittr, line 72: Leave it to API users to find bugs in documentation :smile: cheers for this.

lib/posix-process/signal/signal.h
```diff
@@ -295,6 +303,62 @@ void pprocess_signal_arch_get_ucontext(ucontext_t *ucontext,
...
+#endif /* CONFIG_LIBPOSIX_PROCESS_SIGNALFD */
+
+/*
+ * Check if a signal is sendable: either is not ignored by the process or
+ * there are signal files that are monitoring for this signal.
+ */
+#if CONFIG_LIBPOSIX_PROCESS_SIGNAL
+#if CONFIG_LIBPOSIX_PROCESS_SIGNALFD
+static inline
+bool pprocess_signal_sendable(struct posix_process *pproc, int signum)
```
andreittr: I appreciate you factored this out from its callsites in `signal.c`, although I believe you could have gone even further. Considering that both callsites negate the result, consider pushing that negation into the function itself, turning it into `pprocess_signal_masked`/`_dropped`/`_muted`/whatever better name you come up with.
This will also remove the slightly head-scratching double-negation when signalfds are disabled, and the callsites read as `if (!(!IS_IGNORED))`.

lib/posix-process/signal/signal.h
```diff
@@ -295,6 +303,62 @@ void pprocess_signal_arch_get_ucontext(ucontext_t *ucontext,
...
+}
+
+/* Notify the signal files of a thread */
+static inline
+void pthread_signal_files_notify(struct posix_process *pproc __unused,
+				 struct posix_thread *pthread __unused,
+				 int signum __unused)
+{
+	/* TODO: Set file event for one specific thread only */
+}
```
andreittr, line 342: While we don't yet have thread-specific file event notification, I believe we should until then notifiy all signalfd-waiting threads instead of doing nothing.
If we do nothing, a thread sleeping on a signalfd may never wake up, even as signals are queued to it, possibly breaking apps. The cost of needlessly waking up all other signalfd threads to immediately read an `-EAGAIN` is IMO worth it.
(Technical nit: since signalfds are explicitly polled files, threads may not even enter `read` on spurious wakeups, rather immediately `poll` and block again -- an even shorter path)

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,354 @@
...
+
+	sigf = uk_file_to_signal_file(f);
+	pthread = uk_pthread_current();
+	pproc = uk_pprocess_current();
+
+	pprocess_signal_foreach(sn) {
+		if (!uk_sigismember(&sigf->mask, sn) ||
+		    (!IS_PENDING(pthread->signal->sigqueue, sn) &&
+		     !IS_PENDING(pproc->signal->sigqueue, sn)))
+			continue;
```
andreittr: Might I suggest
```c
pprocess_signal_foreach(sn)
    if (uk_sigismember(&sigf->mask, sn) &&
        (IS_PENDING(pthread->signal->sigqueue, sn) ||
         IS_PENDING(pproc->signal->sigqueue, sn)))
        return UKFD_POLLIN;

return 0;
```
it's the same logic, but IMO much easier to parse without all those negations and control flow jumps.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	return 0;
+}
+
+/* Convert internal siginfo_t to signalfd_siginfo structure */
+static void si_to_ssi(siginfo_t *si, struct signalfd_siginfo *ssi)
+{
+	/* Zero out the structure first */
+	memset(ssi, 0, sizeof(*ssi));
```
andreittr, line 69: Do we strictly need this? As in, are there fields of `ssi` that we might not initialize and there is documented (or at least observed) behavior that signalfd syscalls zero out the rest?

It's not a rhetorical question (I've been burned by this in `stat` code). If we _need_ this zeroing out we should document it, so some eager future dev doesn't optimize it away. If we don't need it, we shouldn't bother.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,354 @@
...
+			sig = pprocess_signal_dequeue(pproc, NULL, sn);
+		else
+			continue;
+
+		if (!sig)
+			continue;
+
+		si_to_ssi(&sig->siginfo, &si);
+		uk_signal_free(pthread->_a, sig);
+		iovlen_wr += uk_iov_scatter(iov, iovcnt,
```
andreittr, line 171: Oof, you're doing double the memwrites here (using `si` as tmp), and I don't think the compiler is smart enough to optimize it away.
I do get _why_ you're doing it like this, and don't see a clean way of doing both type conversion and iovec scatter any other way (yet).
I'll make a mental note for `iovutil.h` :thinking:

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * param a
+ *   Allocator to use for signal file creation
+ * param mask
+ *   Signal mask to be monitored by the signal file
+ *
+ * @return
+ *   Pointer to the created signal file or NULL on creation failure
+ */
+struct uk_file *signal_file_create(struct uk_alloc *a,
```
andreittr: I recommend you return a `const struct uk_file *` to enforce the rule that all ukfiles, once created, are immutable.
I know this is a lib-internal header/API, but I don't want any precedent out there for mutable ukfile refs being passed around after being made public.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,354 @@
...
+	sigf = uk_file_to_signal_file(of->file);
+
+	/*
+	 * It is not possible to receive SIGKILL or SIGSTOP signals via a
+	 * signal file descriptor; these signals are silently ignored if
+	 * specified in mask.
+	 */
+	uk_sigcopyset(&sigf->mask, new_mask);
+	uk_sigdelset(&sigf->mask, SIGKILL);
+	uk_sigdelset(&sigf->mask, SIGSTOP);
```
andreittr, line 294: I believe you forgot to update the pproc allmask when an existing signalfd's mask is changed (you need to at least OR in any newly set bits to prevent missed signal deliveries; cleared bits can be updated by the usual pruning runs)

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,164 @@
...
+ * param signum
+ *   Signal number to check against for monitoring signal files
+ */
+static inline
+void uk_signal_files_ctx_notify(struct uk_signal_files_ctx *ctx,
+				int signum)
+{
+	struct uk_signal_file *it, *itn;
+
+	uk_sigemptyset(&ctx->allmask);
```
andreittr: Let's hope the allmask is never read while this function is executing, else we'll miss signals; especially since `uk_file_event_set` can end up doing callbacks.
I understand synchronization is not in scope but this still feels really scary; are there huge performance drawbacks to using a local temp to gather the new allmask before writing it in before returning?

andreittr (changes requested):
Another pass, with a closer-toothed comb; it's shaping up nicely in general.

Also, I assume synchronization is not in scope at all and will be addressed in future work.

lib/posix-process/signal/signal_file.c
mogasergiu: No reason, just laziness/copy-paste.

lib/posix-process/signal/signal_file.c
mogasergiu: This is so not my style but ok :laughing:

lib/posix-process/signal/signal_file.c
mogasergiu: oh good catch!

lib/posix-process/signal/signal_file.h
mogasergiu: None, I even did it like that while updating but then I decided to leave it for when we actually do synchronization. But now that makes two of us that think it might be a good idea so I'll do it :small_airplane:

lib/posix-process/signal/signal_file.c
mogasergiu: Yep, I did some tests on linux userspace and it seems the unused members are always zeroed out.

lib/posix-process/signal/signal.h
mogasergiu: Let's go with should_drop

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,167 @@
...
+/*
+ * Create a signal file with a given mask
+ *
+ * param a
+ *   Allocator to use for signal file creation
+ * param mask
+ *   Signal mask to be monitored by the signal file
+ *
+ * @return
+ *   Pointer to the created signal file or NULL on creation failure
```
andreittr: This line contradicts implementation where you return errptrs.
Judging by what this function does though, failing to allocate memory is really the only way it _can_ fail, so returning NULL may well be valid :shrug: 
I don't have strong feelings either way for NULL or ERRPTR, but please have the docs and impl be in line.

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,167 @@
...
+	uk_list_for_each_entry_safe(it, itn, &ctx->sigfiles, sigf_head) {
+		uk_sigorset(&allmask, &it->mask);
+
+		if (!uk_sigismember(&it->mask, signum))
+			continue;
+
+		uk_file_event_set(&it->f, UKFD_POLLIN);
+	}
+
+	uk_sigorset(&ctx->allmask, &allmask);
```
andreittr: I think this should be `uk_sigcopyset`, as ORing `allmask` in will not clear any pruned signals.

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,167 @@
...
+/*
+ * Convert base file structure to specialized signal file structure
+ * param f
+ *   Pointer to base file structure
+ *
+ * @return
+ *   Containing uk_signal_file structure
+ */
+static inline
+struct uk_signal_file *uk_file_to_signal_file(const struct uk_file *f)
```
andreittr: I believe you can move this over into `signal_file.c`, as no other inline in this header is using it AFAICT.
And if you move it, you can add a `UK_ASSERT(FILE_IS_SIGNAL(f))` for the extra safety against using garbage data. As a bonus, you get consistent asserts for all ops (poll, read, set_mask, and destructor) that touch the `uk_signal_file` struct without explicit checks.

andreittr (commented):
Pass 3, getting pretty close to a checkmark from my side.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,358 @@
...
+		if (uk_sigismember(&sigf->mask, sn) &&
+		    (IS_PENDING(pthread->signal->sigqueue, sn) ||
+		     IS_PENDING(pproc->signal->sigqueue, sn)))
+			return UKFD_POLLIN;
+
+	return 0;
+}
+
+/* Convert internal siginfo_t to signalfd_siginfo structure */
+static void si_to_ssi(siginfo_t *si, struct signalfd_siginfo *ssi)
```
andreittr: `const siginfo_t *si` perhaps? Won't change code generation, but might save us debug cycles if someone ever mistypes `si` instead of `ssi` in a future patch :smile:

lib/posix-process/signal/signal_file.h
```diff
@@ -0,0 +1,151 @@
...
+static inline
+void uk_signal_files_ctx_notify(struct uk_signal_files_ctx *ctx,
+				int signum)
+{
+	struct uk_signal_file *it, *itn;
+	uk_sigset_t allmask;
+
+	uk_sigemptyset(&allmask);
+
+	uk_list_for_each_entry_safe(it, itn, &ctx->sigfiles, sigf_head) {
```
andreittr: I _think_ you can use the plain `uk_list_for_each_entry` here; IIRC the safe versions are for when you intend to modify the list mid-traversal. Saves the `itn` tempvar.

lib/posix-process/signal/signal_file.c
```diff
@@ -0,0 +1,358 @@
...
+
+static ssize_t signal_file_read(const struct uk_file *f,
+				const struct iovec *iov, size_t iovcnt,
+				size_t off, long flags __unused)
+{
+	size_t iovlen_total, iovlen_wr, iovi, cur;
+	struct posix_thread *pthread;
+	struct posix_process *pproc;
+	struct uk_signal_file *sigf;
+	struct signalfd_siginfo si;
```
andreittr: Maybe rename `si` to `ssi` for better name consistency with `si_to_ssi()`?

andreittr (commented):
Ok, (hopefully) final pass, only trivial things left.

andreittr (approved):
Cheers! Looks good; can't wait to see the synchronization PR :smile: 

michpappas:

PR #1620, opened 2025-03-26
libposix-process: Create PID1 from thread passed via init

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Add an new element to `struct uk_init_ctx` for the thread that calls `main()`, and update `libposix-process` to use that thread when instantiating PID1. When selecting `LIBUKBOOT_MAINTHREAD` this allows excluding Unikraft's init thread from the init process. Update `libukboot` to start the main thread before running through inittab. Update the parameters of `main_thread()` to pass a pointer to `struct uk_init_ctx` to allow `argc` and `argv` to be updated by inittab.

### DIFF
--- a/include/uk/init.h
+++ b/include/uk/init.h
@@ -39,6 +39,10 @@
 #include <uk/essentials.h>
 #include <uk/prio.h>
 
+#if CONFIG_LIBUKSCHED
+#include <uk/thread.h>
+#endif /* CONFIG_LIBUKSCHED */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -48,6 +52,12 @@ struct uk_init_ctx {
 		int    argc;
 		char **argv;
 	} cmdline;
+#if CONFIG_LIBUKSCHED
+	/* Set if main() executes on a separate thread,
+	 * otherwise NULL.
+	 */
+	struct uk_thread *tmain;
+#endif /* CONFIG_LIBUKSCHED */
 
 	/* reserved for future additions */
 };
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -41,6 +41,8 @@
 #include <uk/config.h>
 #include <uk/syscall.h>
 
+struct uk_thread *pprocess_thread_main;
+
 #if CONFIG_LIBPOSIX_PROCESS_PIDS
 #include <uk/bitmap.h>
 #include <uk/list.h>
@@ -406,11 +408,26 @@ void uk_posix_process_kill(struct uk_thread *thread)
 }
 
 #if CONFIG_LIBPOSIX_PROCESS_INIT_PIDS
-static int posix_process_init(struct uk_init_ctx *ictx __unused)
+static int posix_process_init(struct uk_init_ctx *ictx)
 {
+	struct uk_thread *t;
+
+	UK_ASSERT(ictx);
+
+	/* If ictx->tmain in set, main() executes on a
+	 * separate uk_thread. Instantiate PID_INIT from
+	 * that thread, and set pprocess_thread_main, as
+	 * we will need that information later.
+	 */
+	if (ictx->tmain) {
+		t = ictx->tmain;
+		pprocess_thread_main = ictx->tmain;
+	} else {
+		t = uk_thread_current();
+	}
+
 	/* Create a POSIX process without parent ("init" process) */
-	return uk_posix_process_create(uk_alloc_get_default(),
-				       uk_thread_current(), NULL);
+	return uk_posix_process_create(uk_alloc_get_default(), t, NULL);
 }
 
 uk_late_initcall(posix_process_init, 0x0);
--- a/lib/ukboot/boot.c
+++ b/lib/ukboot/boot.c
@@ -114,8 +114,9 @@ int main(int argc, char *argv[]) __weak;
 static inline int do_main(int argc, char *argv[]);
 
 #if CONFIG_LIBUKBOOT_MAINTHREAD
-static __noreturn void main_thread(void *, void *);
+static __noreturn void main_thread(void *);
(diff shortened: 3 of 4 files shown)


include/uk/init.h
```diff
@@ -38,6 +38,7 @@
 #include <uk/plat/bootstrap.h>
 #include <uk/essentials.h>
 #include <uk/prio.h>
+#include <uk/thread.h>
```
andreittr, line 43: Since `<uk/thread.h>` is provided by `lib/uksched` we should probably guard this behind `HAVE_SCHED`.

include/uk/init.h
```diff
@@ -49,6 +50,9 @@ struct uk_init_ctx {
 		char **argv;
 	} cmdline;
 
+	/* The tread that executes main() */
+	struct uk_thread *tmain;
```
andreittr, line 59: Similar argument, should probably guard behind `HAVE_SCHED`.

Ideally we'd want to guard this behind an even finer-grained condition, that applies IFF this field is actually used (so, probably involving posix process to some degree), but I'm not sure how cleanly this can be done in a core kernel header :thinking: . `HAVE_SCHED` is probably good enough.

Also: typo "tread" -> "thread"

lib/ukboot/boot.c
```diff
@@ -364,6 +367,23 @@ void uk_boot_entry(void)
...
+				       0x0 /* default auxiliary stack size */,
+				       false, false,
+				       "main", NULL,
+				       main_thread_dtor);
+	if (unlikely(!m || PTRISERR(m)))
+		UK_CRASH("Failed to launch application's main()\n");
+
+	ictx.tmain = m;
+#else /* !CONFIG_LIBUKBOOT_MAINTHREAD */
+	ictx.tmain = uk_thread_current();
```
andreittr: AFAICT `ukboot` can be configured without a scheduler, so related to guarding `ictx.tmain` behind `HAVE_SCHED`, this should probably be
```c
#elif HAVE_SCHED
    ictx.tmain = uk_thread_current();
#endif /* HAVE_SCHED */
```

andreittr (changes requested):
The change looks nice, a clean way to remove any assumptions about what thread should become PID1. I do have one issue about builds w/o a scheduler though.

edit: in light of @mogasergiu 's comments, replace `HAVE_SCHED` with the appropriate config macro in below comments.

include/uk/init.h
mogasergiu: I think this header and the API are behind `CONFIG_LIBUKSCHED`'s API. Another scheduler library might offer a differently named API. I think the guards should maybe be `CONFIG_LIBUKSCHED` instead. I see `HAVE_SCHED` is only used for things that don't directly interact with any API but rather use something that implies there is some sort of scheduling/threading enabled, namely I see it is used for TLS variables. I see that `posix-time` also uses `HAVE_SCHED`, maybe there should have been `CONFIG_LIBUKSCHED` instead as well?

lib/ukboot/boot.c
mogasergiu: Maybe we could move this #if/else below the `#if CONFIG_LIBUKBOOT_INITSCHED` a few lines above? That being said, maybe we could write a separate function that does all this scheduler/thread stuff. This function has too many #if's already :thinking:

lib/posix-process/process.c
```diff
@@ -406,11 +406,13 @@ void uk_posix_process_kill(struct uk_thread *thread)
 }
 
 #if CONFIG_LIBPOSIX_PROCESS_INIT_PIDS
-static int posix_process_init(struct uk_init_ctx *ictx __unused)
+static int posix_process_init(struct uk_init_ctx *ictx)
 {
+	UK_ASSERT(ictx->tmain);
```
mogasergiu: Do we need to assert for this here? This function does not directly access it, it simply just passes it to `uk_posix_process_create`. I think the check for `NULL` should be offloaded to `uk_posix_process_create` instead since, unlike `posix_process_init`, it actually derefs the thread pointer/

include/uk/init.h
michpappas: Good catch thanks. As @mogasergiu said though, the right one is `CONFIG_LIBUKSCHED`, which is also inline with the rest of the dependency checks in `libukboot`. Will update.

lib/posix-process/process.c
michpappas: That depends on how you see it. One may argue that from the function's perspective it makes sense to evaluate the validity of its parameters before passing them on. Anyway I find this very minor, especially given that it's an assertion.

include/uk/init.h
michpappas: > Similar argument, should probably guard behind HAVE_SCHED.

That's right thanks. Will guard around `CONFIG_LIBUKSCHED` (see previous comment for reasoning).

> Ideally we'd want to guard this behind an even finer-grained condition, that applies IFF this field is actually used (so, probably involving posix process to some degree), but I'm not sure how cleanly this can be done in a core kernel header 🤔 . HAVE_SCHED is probably good enough.

IMO that defeats the purpose, as it wouldn't make sense do guard every element in init / term ctx to its consumer library. Let's keep it as is.

lib/ukboot/boot.c
michpappas: Will additionally guard against `CONFIG_LIBUKSCHED`, thanks.

andreittr (approved):
Thanks for the changes; this looks good on my end :+1: 

mogasergiu (approved):

andreittr (approved):
Looks good!

andreittr:
Let's merge!

PR #1626, opened 2025-04-01
lib/posix-process/signals: Fix conditional for sigsetops

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The implementation of sigsetops was conditional to HAVE_LIBC instead of CONFIG_HAVE_LIBC, causing a build error when musl is used, due to the definiton of sigset_t to different types. Update the conditional to fix. Notice that this works as each flavor of libc controls both the type and the implementation of the functions that operate on it.

PR #1627, opened 2025-04-01
lib/posix-process: Supplementary changes for multiprocess

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

⚠️ **Deprecation notice:**

* CONFIG_LIBPOSIX_PROCESS_PIDS
* CONFIG_LIBPOSIX_PROCESS_INIT_PIDS
* CONFIG_LIBPOSIX_PROCESS_CLONE
* CONFIG_LIBPOSIX_PROCESS_VFORK 

The following are consolidated into a single option, CONFIG_LIBPOSIX_PROCESS_MULTITHREADING:
* CONFIG_LIBPOSIX_PROCESS_PIDS
* CONFIG_LIBPOSIX_PROCESS_CLONE

The following is replaced by CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS:
* CONFIG_LIBPOSIX_PROCESS_VFORK 

CONFIG_LIBPOSIX_PROCESS_INIT_PIDS is deprecated, and its functionality is enabled by default.

### Description of changes

This series consists of a set of supplementary changes emerging from the work on multiprocess. These consist of fixes as well as major refactoring of libposix-process. In brief:
* Refactor Kconfig options of libposix-process for multithreading / multiprocess: This consolidates exiting options into more meaningful ones and removes redundancy. Affects both internal and external libraries.
* Make process management functions internal: Removes redundant functions from the public API of libposix-process.
* Assign a fixed pid to init: Deprecates CONFIG_LIBPOSIX_PROCESS_INIT_PIDS and makes its functionality enabled by default.
* Restrict pthread creation to clone(): This fixes a long standing issue where new kernel threads will always be associated to new pthreads.
* Various fixes

For details see the commit messages of individual PRs.

Depends-on: https://github.com/unikraft/app-elfloader/pull/90
Depends-on: https://github.com/unikraft/lib-musl/pull/87


michpappas:
Regarding the failure of the libc-test: Should be related to the outdated options in libmusl.

lib/posix-process/Config.uk
```diff
@@ -4,50 +4,44 @@ menuconfig LIBPOSIX_PROCESS
...
-	select LIBUKLOCK
-
-config LIBPOSIX_PROCESS_CLONE
-	bool "clone() system call"
-	select LIBPOSIX_PROCESS_PIDS
-
 config LIBPOSIX_PROCESS_CLONE_PREFER_CHILD
 	bool "Prefer scheduling of child"
-	depends on LIBPOSIX_PROCESS_CLONE
+	depends on LIBPOSIX_PROCESS_MULTITHREADING
```
andreittr: This `depends on` line might be redundant, as the config option is already in an `if LIBPOSIX_PROCESS_MULTITHREADING` block.

lib/posix-process/Config.uk
```diff
@@ -4,50 +4,44 @@ menuconfig LIBPOSIX_PROCESS
...
+	bool "Multithreading support"
 	select LIBUKSCHED
+	select LIBSYSCALL_SHIM
 
-if LIBPOSIX_PROCESS_PIDS
+if LIBPOSIX_PROCESS_MULTITHREADING
 
 config LIBPOSIX_PROCESS_MAX_PID
-	int "Largest PID"
+	int "Largest TID"
```
andreittr, line 12: I understand why the label change to "TID" (and why keep the config var the same); would it perhaps be more descriptive to users to say "Largest TID/PID" instead? Or do you think there would be more confusion if we mention PIDs for an option that pops up without multiprocessing enabled?

lib/posix-process/include/uk/process.h
```diff
@@ -43,7 +43,7 @@
 #include <uk/thread.h>
 #endif
```
andreittr: AFAICT (pls correct me if I'm wrong), `<uk/thread.h>` is used by multithreading code only (and the multithread Kconfig option ensures uksched is selected).
I believe we could move this `#include` into the `#if CONFIG_LIBPOSIX_PROCESS_MULTITHREADING` block lower down.

lib/posix-process/process.c
```diff
@@ -749,7 +752,25 @@ static int pprocess_clone_thread(const struct clone_args *cl_args __unused,
 				 struct uk_thread *child __unused,
 				 struct uk_thread *parent __unused)
```
andreittr: `child`, `parent`, and `cl_args` are now used, you can remove the `__unused` attr.

lib/posix-fdtab/Config.uk
```diff
@@ -13,7 +13,7 @@ if LIBPOSIX_FDTAB
 	# Hidden, will autoselect when required
 	config LIBPOSIX_FDTAB_MULTITAB
 	bool
-	default y if LIBPOSIX_PROCESS_CLONE
+	default y if LIBPOSIX_PROCESS_MULTITHREADING
```
andreittr, line 16: The multitab logic could (in theory) now be constrained to only `if LIBPOSIX_PROCESS_MULTIPROCESS`, and the compat stop-gaps removed, but that's a change in behavior outside the scope of this work.
Nevertheless, something to consider soon.

andreittr (commented):
Change looks nice, I like the featureset approach to `posix-process` config.
I've just some minor remarks.

lib/posix-process/Config.uk
```diff
@@ -4,50 +4,44 @@ menuconfig LIBPOSIX_PROCESS
 
 if LIBPOSIX_PROCESS
 
-menuconfig LIBPOSIX_PROCESS_PIDS
```
mogasergiu, line 7: Commit meassage headline for 
 lib/posix-process: Refactor Kconfig options for multithreading/multip…

…rocess

is too long and spills into the next line

I know you added a Checkpatch-Ignore, but I'd rather not ahve it like this in our upstream history.

lib/posix-process/clone.c
```diff
@@ -515,15 +515,6 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
 		    (void *) child->tlsp,
 		    (child->tlsp != child->uktlsp) ? "custom" : "Unikraft");
 
-	if (!(cl_args->flags & CLONE_THREAD)) {
-		uk_pr_debug("Creating new process for child\n");
```
mogasergiu, line 512: In commit message of ` lib/posix-process: Restrict pthread creation to clone() `: `s/rename then/rename them/`

lib/posix-process/include/uk/process.h
```diff
@@ -218,6 +211,15 @@ struct uk_posix_clonetab_entry {
...
 				 UK_PRIO_LATEST)
 
+/* Creates a pthread and attaches it to the current process
+ *
+ * DO NOT USE. This is only necessary when loading an ELF via initrd,
+ * where we want to POSIXise the kernel thread, as initrd does not
+ * support paths and therefore we cannot execve(). For more details
+ * see the implementation of app-elfloader.
+ */
+int uk_posix_process_create_pthread(struct uk_thread *thread);
```
mogasergiu, line 225: Why don't we move it to elf_binfmt in elfloader then?

lib/posix-process/vfork.c
```diff
@@ -18,21 +18,17 @@
 
 UK_LLSYSCALL_R_E_DEFINE(pid_t, vfork)
 {
-	struct posix_process *child_proc;
 	struct clone_args cl_args = {0};
-	pid_t child_tid;
+	pid_t pid; /* child */
```
mogasergiu, line 22: Commit subject is a bit misleading, maybe rewrite as to imply it refers to fixing nomenclature rather than C-type

lib/posix-process/Config.uk
```diff
@@ -4,50 +4,44 @@ menuconfig LIBPOSIX_PROCESS
...
-	bool "Process and thread IDs"
+menuconfig LIBPOSIX_PROCESS_MULTITHREADING
+	bool "Multithreading support"
 	select LIBUKSCHED
+	select LIBSYSCALL_SHIM
```
mogasergiu: Why do we need syscall shim as mandatory now?

lib/posix-process/Config.uk
michpappas: > Or do you think there would be more confusion if we mention PIDs for an option that pops up without multiprocessing enabled?

Exactly. And in addition, with multiprocess enabled one would expect to find it under that menu instead, so I opted to move it under multithreading and rename it to something sensible, and hope it would be intuitive to the user that when multiprocess is enabled it also implies max pid. Saying that, I don't like it very much either, so I'm open to suggestions.

lib/posix-process/include/uk/process.h
michpappas: That's right, the only thing is that without it we get a compile error because of `NULL` not being defined. I think `thread.h` resolves that by some chain include that uses `stddef.h`. Anyway, I'll move `thread.h` inside the `CONFIG_LIBPOSIX_PROCESS_MULTITHREADING` block, and include `stddef.h` for now, and at some point push another commit to replace boolean types and NULL with the equivalent kernel types.

lib/posix-process/vfork.c
michpappas: Good point, will update 👍🏼

lib/posix-process/Config.uk
michpappas: Okay so these are macros that are always compiled in right. Will update, thanks 👍🏼

lib/posix-process/Config.uk
andreittr: > I'm open to suggestions.

Only thing I can think of are 2 separate Kconfig options: "Largest TID" and "Largest TID/PID", first selected if `multithread && !multiproc`, 2nd if `multiproc`; then some Kconfig magic to turn these 2 vars into a single var that code uses.

Or, y'know, just keep TID, user is configuring a kernel, they should know what's up :smile:

michpappas:
Rebased to the latest state of #1618

lib/posix-process/clone.c
michpappas: This commit is now gone.

michpappas:
Addressed review comments, added #1626 into the dependencies to avoid a conflict, pushed an additional patch to fix a bug in `clone()`.

michpappas:
@andreittr, @mogasergiu I pushed a bunch updates. The TLDR is:
* Rebased to the latest state of #1618
* What https://github.com/unikraft/unikraft/pull/1627#issuecomment-2783362514 says
* The cummulative diff of chagnes excl. the rebase: https://github.com/unikraft/unikraft/compare/88b066ae8cf42f9ce18197925770a1898191baaf..1ab140af44a880ed24c4d204e99ec5aced775d16
* libc test fails because the relevant change in musl pending merge

mogasergiu (approved):

lib/Config.uk
```diff
@@ -7,31 +7,27 @@
...
 config HAVE_NW_STACK
        bool
-       default n
 
 config HAVE_SYSCALL
        bool
-       default n
+
+config HAVE_VFS
+	bool
```
andreittr, line 27: Thanks for this, it'll make the transition to `posix-vfs` smoother :pray: .

lib/posix-process/Config.uk
```diff
@@ -4,35 +4,17 @@ menuconfig LIBPOSIX_PROCESS
 
 if LIBPOSIX_PROCESS
 
-menuconfig LIBPOSIX_PROCESS_PIDS
-	bool "Process and thread IDs"
-	select LIBUKSCHED
-
-if LIBPOSIX_PROCESS_PIDS
+menuconfig LIBPOSIX_PROCESS_MULTITHREADING
```
andreittr, line 7: Typo in the "Refactor Kconfig options" commit message, last paragraph: multirpocess -> multiprocess

lib/posix-process/Makefile.uk
```diff
@@ -38,24 +37,22 @@ LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signal/kill.c
 LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signal/tgkill.c
 LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signal/tkill.c
 
-ifeq ($(CONFIG_LIBPOSIX_PROCESS_CLONE), y)
+ifeq ($(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING), y)
 LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signal/clone.c|signal
 endif
```
andreittr: Is there a reason I'm not seeing for keeping this an `ifeq` block rather than `LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING)`?

lib/syscall_shim/Makefile.uk
```diff
@@ -100,7 +100,7 @@ LIBSYSCALL_SHIM_LIBC_STUBS_FLAGS-$(call have_gcc) += -Wno-builtin-declaration-mi
 LIBSYSCALL_SHIM_SRCS-$(CONFIG_LIBSYSCALL_SHIM_HANDLER) += $(LIBSYSCALL_SHIM_BASE)/uk_syscall_binary.c|isr
 
 LIBSYSCALL_SHIM_SRCS-y += $(LIBSYSCALL_SHIM_BASE)/uk_prsyscall.c
-LIBSYSCALL_SHIM_SRCS-$(CONFIG_LIBPOSIX_PROCESS_EXECVE) += $(LIBSYSCALL_SHIM_BASE)/execve.c
+LIBSYSCALL_SHIM_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS) += $(LIBSYSCALL_SHIM_BASE)/execve.c
```
andreittr: Since `LIBPOSIX_PROCESS_EXECVE` remains a separate option, I think this change should be dropped here, as we want the `execve` syscall even if `!multiprocess`.

lib/posix-fdtab/fdtab.c
```diff
@@ -28,14 +28,14 @@
...
 #include <uk/process.h>
-#endif /* CONFIG_LIBPOSIX_PROCESS_CLONE */
+#endif /* CONFIG_LIBPOSIX_PROCESS_MULTITHREADING */
 
-#if CONFIG_LIBPOSIX_PROCESS_EXECVE
+#if CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS
 #include <uk/event.h>
 #include <uk/prio.h>
-#endif /* CONFIG_LIBPOSIX_PROCESS_EXECVE */
+#endif /* CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS */
```
andreittr: fdtab in particular should have its cloexec handler run on any exec, regardless of multiproc or not; I think this should remain `CONFIG_LIBPOSIX_PROCESS_EXECVE` (and similarly around the handler).

andreittr (commented):
Looks very good; did a full pass and have only one issue with execve vs. multiproc `ifdefs` in libs.

lib/posix-process/Makefile.uk
michpappas: I think it's how it evolved once making it unconditional due to the stubs. Will push an update in #1628 and update here accordingly, thanks.

lib/syscall_shim/Makefile.uk
michpappas: Ooops, I totally forgot to update these, thanks for the catch. Will also update the commit message to remove the paragraph about execve.

lib/posix-fdtab/fdtab.c
michpappas: Yup, same as above, forgot to update.

andreittr (approved):
Looks great!

michpappas:
Rebased from `staging`

andreittr:

PR #1628, opened 2025-04-03
lib/posix-process/signal: Add stub for CLONE_SIGHAND

michpappas:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [e.g.N/A]
 - Application(s): [e.g.N/A]

### Description of changes

Provide stub for CLONE_SIGHAND when LIBPOSIX_PROCESS_SIGNAL is not selected. This prevents clone() from returning -ENOTSUP when signals are not selected.

PR #1598, opened 2025-03-11
lib/syscall_shim: Use global scope inline asm for execenv prologue

mogasergiu:
### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s):  `x86_64` or `arm64`
 - Platform(s): N/A
 - Application(s): N/A

### Description of changes

Avoid having to use weird and unconventional function attributes by writing the inline assembly in a global scope. This is especially useful for ARM64 as it does not support the naked attribute on GCC.

PR #1630, opened 2025-04-04
lib/posix-process: exit() and wait()

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

This series implements the exit and wait functionality for multiprocess support in `libposix-process`. Specifically it implements the `_exit()`, `exit_group()`, `waitid()`, and `wait4()` syscalls, along with their libc wrappers for nolibc.

The possible execution paths to terminate a pthread are:
* On pthread return, by `uk_thread` calling `posix_thread_fini()`.
* On termination by `_exit()`, via `pprocess_exit_pthread()`.
* On termination by signal, via `pprocess_exit_pthread()`.

In all the above cases the flow is the same: The tread termination workflow implemented in `pprocess_exit_pthread()` is followed by `pprocess_release_pthread()` to free the pthread.

For processes, the workflow is similar with the exception that processes are released by `wait()` either during `pprocess_exit()` if the parent is already waiting, or they remain in zombie state until the parent calls `wait()`. One exception is `UK_PID_INIT`, which cannot be waited, so it's immediately released upon return.

The above logic also covers `POSIX_PROCESS_MULTITHREADING`: The application can exit individual threads, or the init process that is operating on.

In the case where `POSIX_PROCESS_MULTITHREADING` is not enabled (`libposix_process` syscalls are stubbed) the behavior of exit depends on the selection of  (`LIBUKSCHED` and `LIBUKBOOT_MAINTHREAD`). If `main()` executes on its own thread, `exit()` is emulated. Otherwise, the system crashes, as there is no thread to exit, or the only thread is the one Unikraft is executing on.

Depends-on: #1598
Depends-on: #1618
Depends-on: #1620
Depends-on: #1627
Depends-on: #1628


lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+UK_EVENT(POSIX_PROCESS_EXIT_EVENT);
+
+int pprocess_exit_status;
+
+#if CONFIG_LIBPOSIX_PROCESS_MULTITHREADING
+
+/* pprocess state to pthread state */
+#define PP2PT_STATE(_pp_state)						\
+	((_pp_state) == POSIX_PROCESS_EXITED ? POSIX_THREAD_EXITED :	\
+					      POSIX_THREAD_KILLED)
```
andreittr, line 44: Seeing as this macro does not handle pproc/pthread states generally (AFAICT it assumes it'll only ever convert pproc exit states into pthread exit states), perhaps including "exit state" in the name/description would help make callsites clearer.

If converting exit states is not all it does, and indeed it's valid to be called with `POSIX_PROCESS_RUNNING` as well, then that should also be made clear, as it's not immediately obvious and future editors might make the same error as I.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+{
+	struct posix_process *pchild, *pchildn;
+	struct posix_process *pprocess_init;
+
+	pprocess_init = pid2pprocess(UK_PID_INIT);
+
+	uk_list_for_each_entry_safe(pchild, pchildn, &pprocess->children,
+				    child_list_entry) {
+		uk_list_del(&pchild->child_list_entry);
+		if (pprocess_init && pprocess != pprocess_init) {
```
andreittr, line 56: IIUC this function is called for a `pprocess` during its `exit()` to reparent its children to `PID_INIT` (which I'll call PID1 for brevity). I wonder when this condition would fail
- `pprocess_init == NULL` I assume happens when there's no multiprocessing (only multithreading) and thus no real PID1; that's fair, but then how could there be any child process entries to reparent? My intuition says this loop should be empty in that case.
- `pprocess == pprocess_init` would happen iff you call `exit` from PID1, in which case the system is going down _by definition_ and setting parents to NULL is done (I assume) as a convenience for future cleanup code -- I can't see under what semantics a parentless process (that's not PID1) would work. Either way, exiting from PID1 is a special enough case that I think a check and special handling outside of this loop is warranted

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+
+#if CONFIG_LIBPOSIX_PROCESS_SIGNAL
+static inline int signal_exit(struct posix_process *pprocess)
+{
+	UK_ASSERT(pprocess);
+	UK_ASSERT(pprocess->signal);
+
+	if (pprocess->signal->sigaction[SIGCHLD].ks->ks_handler == SIG_IGN ||
+	    pprocess->signal->sigaction[SIGCHLD].ks->ks_flags & SA_NOCLDWAIT)
+		return 1;
```
andreittr, line 81: Not strictly in scope of this particular PR, but convenience macros for signal disposition of a process would help readability.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	UK_ASSERT(pthread->state != POSIX_THREAD_EXITED ||
+		  pthread->state != POSIX_THREAD_KILLED);
+
+	pprocess = pthread->process;
+	UK_ASSERT(pprocess);
+	UK_ASSERT(pprocess->state == POSIX_PROCESS_RUNNING);
+
+	uk_pr_debug("pid %d: exit tid %d\n", pprocess->pid, pthread->tid);
+
+	/* May be NULL if init terminated early */
```
andreittr, line 113: I guess related to my previous comment: do we meaningfully support normal execution (e.g., into a call to `_exit()`) _after_ PID1 has terminated?
I would assume we stop scheduling at that point and only run kernel cleanup code -- shutdown of userspace (graceful or otherwise) is meant to happen _before_ PID1 goes down, right?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	    parent_pthread->state == POSIX_THREAD_BLOCKED_VFORK) {
+		uk_thread_wake(parent_pthread->thread);
+		parent_pthread->state = POSIX_THREAD_RUNNING;
+	}
+
+	/* Release pthread and terminate the undelying uk_thread
+	 * unless it's the current one.
+	 */
+	pprocess_release_pthread(pthread);
+	if (pthread->thread != uk_thread_current())
```
andreittr: Is the `pthread->thread` deref still valid after `pprocess_release_pthread(pthread)`? A cursory look at the impl suggests it's UAF, as it calls `uk_free(pthread)`.
You probably want to remember the ukthread in a tmpvar or reverse the order.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	UK_ASSERT(pprocess->state == POSIX_PROCESS_RUNNING);
+
+	uk_pr_debug("pid %d: exit process\n", pprocess->pid);
+
+	/* May be NULL if init terminated early */
+	parent_process = pprocess->parent;
+
+	/* As this will be signaling the parent on behalf of the current pthread,
+	 * we must do this before we terminate the pthreads, in case we are called
+	 * are called from _exit(), exit_group() (i.e. operate on current).
```
andreittr: Typo: duplicate "are called" on consecutive lines

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+			    PTHREAD_BLOCKING_ON_SIGNAL(pt)) {
+				uk_semaphore_up(&parent_process->wait_semaphore);
+				break;
+			}
+		}
+	} else if (nowait == true) { /* release now */
+		uk_pprocess_foreach_pthread(pprocess, pt, ptn) {
+			if (pt->thread == uk_thread_current())
+				continue;
+			uk_sched_thread_terminate(pt->thread);
```
andreittr: Doesn't this pthread termination already happen in the `pprocess_exit_pthread` loop earlier?

Also: I believe the pprocess is never actually released if `nowait`.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+
+	pprocess_exit(pprocess, POSIX_PROCESS_EXITED, status);
+	uk_sched_thread_exit();
+}
+#else /* !CONFIG_LIBPOSIX_PROCESS_MULTITHREADING */
+void pprocess_exit_stub(int status)
+{
+	if (pprocess_thread_main) {
+		pprocess_exit_status = status;
+		uk_sched_thread_exit(); /* noreturn */
```
andreittr, line 278: If I understand right, the `POSIX_*_EXIT_EVENT` is only raised in the "real" exit implementations, and thus libraries can only rely on it if `CONFIG_LIBPOSIX_PROCESS_MULTITHREADING`.
Would it make sense for consistency to provide this event in the stub as well (with mocked/zero values for pid & tid)? Or should that be special-cased to termtab functions?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	/* May be NULL if init terminated early */
+	parent_pthread = pthread->parent;
+
+	/* Update state */
+	pthread->state = state;
+
+	/* Raise event */
+	event_data.thread = pthread->thread;
+	event_data.tid = pthread->tid;
+	event_data.tid = pprocess->pid;
```
andreittr: Perhaps typo: `event_data.pid` intended?

lib/posix-process/process.c
```diff
@@ -1,39 +1,15 @@
...
- * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
- * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
- * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- * POSSIBILITY OF SUCH DAMAGE.
+ * Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
  */
 
+#define UK_DEBUG 1
```
andreittr: Remnant from development I assume.

lib/posix-process/process.c
```diff
@@ -359,33 +344,28 @@ int pprocess_create(struct uk_alloc *a,
...
-			pchild->parent = NULL;
-			uk_pr_debug("Process PID %d loses its parent\n",
-				    pchild->pid);
+	UK_ASSERT(uk_list_empty(&pprocess->children));
+
+	/* Unlink this process from its parent */
+	if (pprocess->parent) {
+		uk_list_for_each_entry_safe(pchild, pchildn,
+					    &pprocess->parent->children,
+					    child_list_entry) {
```
andreittr: Do we need to iterate over the parent's child list? AFAICT we use a doubly-linked list and we could delete `pprocess` in-place.

lib/posix-process/process.h
```diff
@@ -120,9 +107,62 @@ struct posix_process *tid2pprocess(pid_t tid);
 pid_t ukthread2tid(struct uk_thread *thread);
 pid_t ukthread2pid(struct uk_thread *thread);
 
-void pprocess_kill(struct posix_process *pprocess);
+/**
+ * INTERNAL. Exit and release a pthread
+ *
+ * Performs termination tasks and raises the POSIX_PROCESS_EXIT_EVENT.
+ * Unlike processes, pthreads are released upon exit. Besides the prhtread,
```
andreittr: Typo: `pthread`

lib/posix-process/process.h
```diff
@@ -120,9 +107,62 @@ struct posix_process *tid2pprocess(pid_t tid);
...
+ * @param state        The new process state. Must be either
+ *                     POSIX_PROCESS_STATE_EXITED if terminated volunatrily, or
+ *                     POSIX_PROCESS_STATE_KILLED if terminated by signal.
+ * @param exit_status  The exit status of the terminating process.
+ */
+void pprocess_exit(struct posix_process *process, enum posix_process_state state,
+		   int exit_status);
+
+/**
+ * INTERNAL. Release ptrhead
```
andreittr: Typo: `pthread`

andreittr (commented):
1st pass over `exit` only (pipelining the review); will continue over `wait`

lib/posix-process/process.h
```diff
@@ -14,6 +14,8 @@
 #if CONFIG_LIBPOSIX_PROCESS_MULTITHREADING
 #include <linux/sched.h>
 #include <uk/arch/ctx.h>
+#include <uk/config.h>
+#include <uk/semaphore.h>
```
andreittr, line 17: I belive `#include <uk/semaphore.h>` should be moved to the exit commit, as that's where the semaphore fields were introduced. The `select LIBUKLOCK` line in `Config.uk` goes with these I believe.

`#include <uk/config.h>` is redundant, as you're already in an `#if` block that depends on a config var being set.

lib/posix-process/process.h
```diff
@@ -94,6 +98,10 @@ extern __uk_tls struct posix_thread *pthread_self;
 	uk_list_for_each_entry_safe((_pthread), (_pthreadn),		\
 				    &(_proc)->threads, thread_list_entry)
 
+#define uk_pprocess_foreach_child(_pparent, _p)				\
+	uk_pprocess_foreach((_p))					\
+		if ((_p)->parent == (_pparent))				\
```
andreittr: If this is meant to iterate over `_pparent`'s children, why are we going through the entire proclist and filtering? Is there a reason we're not iterating over `_pparent->children` instead?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+		/* Wait for a state change on one of our children */
+		uk_semaphore_down(&this_thread->process->wait_semaphore);
+	}
+
+out:
+	this_thread->state = POSIX_THREAD_RUNNING;
+	return ret;
+}
+
+pid_t uk_posix_process_wait(void)
```
andreittr: Is this function used anywhere? Name & non-static attr suggests it's intended for export, but no header file declares it. If intended as a kernel-internal syscall alternative, then might I suggest the `uk_sys_` namespace (and expanded args)? though I'm not sure what use exposing _just_ wait has; unless there are reasons I'm not aware of, I would suggest removing it.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+		 */
+		if (unlikely(!child || child->parent != this_thread->process)) {
+			ret = -ECHILD;
+			goto out;
+		}
 
+		/* If the process exited already, reap and return the pid */
+		if (child->state != POSIX_PROCESS_RUNNING) {
+			pprocess_reap(child, si);
+			goto out;
```
andreittr, line 106: This will go to out with an uninitialized `ret`. I believe you meant `ret = pprocess_reap(child, si)`.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+	ret = pprocess_wait(pid, &si, options | WEXITED);
+
+	if (ret > 0 && wstatus) {
+		if (si.si_status == CLD_KILLED)
+			*wstatus = si.si_code;
+		else /* CLD_EXITED */
+			*wstatus = si.si_code << 8;
+	}
+
+	return ret > 0 ? si.si_pid : ret;
```
andreittr: Is there any case where `ret != si.si_pid` on succes? If not, we can simplify to `return ret;`.
AFAICT `pprocess_wait` always returns the pid or error.

andreittr (commented):
pass1 over wait & the rest

lib/posix-process/execve.c
```diff
@@ -69,10 +69,16 @@ static void __noreturn execve_ctx_switch(long arg0, long arg1)
...
-static int pprocess_cleanup(struct uk_thread *thread __maybe_unused)
+static int pprocess_cleanup(struct posix_process *pprocess)
 {
-	/* Kill this thread's siblings */
-	pprocess_kill_siblings(thread);
+	struct posix_thread *pt, *ptn;
+
+	/* Kill all threads of this process except current */
+	uk_pprocess_foreach_pthread(pprocess, pt, ptn) {
+		if (pt->thread == uk_thread_current())
```
mogasergiu, line 78: Perhaps we can avoid doing this if by removing current thread's pthread from the list and then adding it back after we are done?

Or, better yet, how about a `uk_list_splice_init` of the pthread list into a temporary locally defined list head and then a `uk_list_del` on curent thread's pthread followed by an insert into the now empty pthread list?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2024, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu, line 2: 2025?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	this_pthread = uk_thread_uktls_var(this_thread, pthread_self);
+
+	UK_ASSERT(this_pthread);
+	UK_ASSERT(this_pthread->process);
+
+	/* Last thread, exit the process */
+	if (uk_list_is_singular(&this_pthread->process->threads)) {
+		pprocess_exit(this_pthread->process, POSIX_PROCESS_EXITED,
+			      status);
+		uk_sched_thread_exit();
```
mogasergiu, line 256: Should we put a UK_BUG()/UK_CRASH() after all calls to `uk_sched_thread_exit`?
Also, nit, but I would restructure this as
```C
if (uk_list_is_singular(&this_pthread->process->threads))
       pprocess_exit(this_pthread->process, POSIX_PROCESS_EXITED, status);
else
       pprocess_exit_pthread(this_pthread, POSIX_THREAD_EXITED, status);
       
uk_sched_thread_exit();
```

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+			 * set the SIGCHLD handler to SIG_IGN, the status is discarded
+			 * and the child dies immediately."
+			 */
+			uk_pr_info("Parent ignores SIGHLD, terminating\n");
+			nowait = true;
+		} else if (unlikely(ret < 0)) { /* no choice here but crash */
+			UK_CRASH("Could not signal parent (%d)\n", ret);
+		}
+#else /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+	nowait = true;
```
mogasergiu: I guess this is still in the `if`'s scope, indentation? :sweat_smile:

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	/* May be NULL if init terminated early */
+	parent_process = pprocess->parent;
+
+	/* As this will be signaling the parent on behalf of the current pthread,
+	 * we must do this before we terminate the pthreads, in case we are called
+	 * are called from _exit(), exit_group() (i.e. operate on current).
+	 */
+	if (parent_process) {
+#if CONFIG_LIBPOSIX_PROCESS_SIGNAL
+		ret = signal_exit(parent_process);
```
mogasergiu, line 171: AFAICT this can also return `0` so why not `ret >= 0`? Better yet, why not just a `if (unlikely(ret < 0))` and then the rest unindented. Or should there have been an `else` case?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	nowait = true;
+#endif /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+	}
+
+	/* Terminate pthreads.
+	 *
+	 * This must happen before we unblock wait(), so that threads
+	 * have already been released when we reap the process.
+	 */
+	uk_pprocess_foreach_pthread(pprocess, pt, ptn)
```
mogasergiu, line 190: ugh, I almost got fooled by the name, wish it had a `_safe` at the end instead of an implicit `ptn` :sweat_smile:

lib/posix-process/exit.c
mogasergiu, line 56: Why do we do this check in the loop? Do we really need a loop? Maybe a `uk_list_splice_init` is enough?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+		uk_pprocess_foreach_pthread(parent_process, pt, ptn) {
+			if (PTHREAD_WAITING_FOR_PID(pt, uk_sys_getpid()) ||
+			    PTHREAD_BLOCKING_ON_SIGNAL(pt)) {
+				uk_semaphore_up(&parent_process->wait_semaphore);
+				break;
+			}
+		}
+	} else if (nowait == true) { /* release now */
+		uk_pprocess_foreach_pthread(pprocess, pt, ptn) {
+			if (pt->thread == uk_thread_current())
```
mogasergiu: Ditto, maybe we can use `uk_splice_init` instead of having this `if` on every iteration?

lib/posix-process/exit.c
mogasergiu: Assigned `event_data.tid` twice by mistake? :sweat_smile:

lib/posix-process/exit.c
```diff
@@ -0,0 +1,313 @@
...
+	parent_pthread = pthread->parent;
+
+	/* Update state */
+	pthread->state = state;
+
+	/* Raise event */
+	event_data.thread = pthread->thread;
+	event_data.tid = pthread->tid;
+	event_data.tid = pprocess->pid;
+	uk_raise_event(POSIX_PROCESS_EXIT_EVENT, &event_data);
```
mogasergiu: Returns `int`

lib/posix-process/signal/siginfo.h
```diff
@@ -41,4 +41,26 @@ static inline void set_siginfo_sigqueue(int signum, siginfo_t *si,
...
+	case POSIX_PROCESS_EXITED:
+		si->si_status = CLD_EXITED;
+		si->si_code = pprocess->exit_status;
+		break;
+	case POSIX_PROCESS_KILLED:
+		si->si_status = CLD_KILLED;
+		si->si_code = pprocess->exit_status;
+		break;
+	default:
+		uk_pr_warn("CLD_STOPPED, CLD_CONTINUED, CLD_TRAPPED, CLD_DUMPED stubbed");
```
mogasergiu, line 63: It does nothing, but I believe it is good practice to still add `break` in `default:`.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+}
+
+UK_LLSYSCALL_R_DEFINE(int, waitid, idtype_t, idtype, id_t, id,
+		      siginfo_t *, infop, int, options,
+		      struct rusage * __unused, rusage)
+{
+	int ret;
+
+	if (options)
+		if (unlikely(!(options & (WEXITED | WSTOPPED | WCONTINUED |
```
mogasergiu, line 178: Why not part of the same `if`? Ditto `wait4`

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+					  WNOHANG | WNOWAIT))))
+			return -EINVAL;
+
+	if (uk_list_empty(&uk_pprocess_current()->children))
```
mogasergiu: Is this common or would it be worth an `unlikely`? Ditto `wait4`/`uk_posix_process_wait`

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+		return -ECHILD;
+
+	switch (idtype) {
+	case P_PID:
+		ret = pprocess_wait(id, infop, options);
+		break;
+	case P_PIDFD:
+	case P_PGID:
+		ret = -ENOTSUP;
```
mogasergiu: Why not return directly.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,262 @@
...
+ *
+ * @param pid: child pid to wait for or -1 for any
+ * @param opt: same as waitid()
+ * @return child pid, zero if no waitable, negative value on error
+ */
+static int pprocess_wait(pid_t pid, siginfo_t *si, int opt)
+{
+	struct posix_process *this_process;
+	struct posix_thread *this_thread;
+	struct posix_process *waitable = NULL;
```
mogasergiu, line 73: You have sullied the reverse xmas tree style. I know you also like it. I may forgive you this time :santa: :christmas_tree:

lib/posix-process/signal/deliver.c
```diff
@@ -20,8 +21,16 @@ void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
...
 static void uk_sigact_term(int __unused sig)
 {
-	uk_pr_warn("tid %d terminated by signal\n", uk_sys_gettid());
-	pprocess_kill(uk_pprocess_current());
+	struct posix_process *pprocess;
+
+	pprocess = uk_pprocess_current();
+	UK_ASSERT(pprocess);
+
+	uk_pr_warn("pid: %d terminated by signal\n", pprocess->pid);
```
mogasergiu: Why should this be a warning?

lib/posix-process/exit.c
michpappas: It is indeed only a helper to convert the exit state, will apply your suggestion to make it more clear 👍🏼

lib/posix-process/exit.c
michpappas: Ooops, thanks for that 👍🏼

lib/posix-process/exit.c
michpappas: > Doesn't this pthread termination already happen in the pprocess_exit_pthread loop earlier?

Yes I shuffled termination a few times and it looks like I missed that one. Will remove.

> Also: I believe the pprocess is never actually released if nowait.

You're right, thanks for catching that.

lib/posix-process/process.c
michpappas: Yes, thanks

lib/posix-process/exit.c
michpappas: I expected that exit events are the counterpart of clone, and since there is no clone, there is nothing to tear down. But since you're the consumer, if you have a use-case in mind I can trigger the event on the stub too.

andreittr: Let's cross that bridge if/when we get to it; atm I don't have a concrete use-case in mind, and now that you made it explicit that exit events act as the counterpart of clone events (with the exception of pid1 :speak_no_evil: ), I see even less use for raising the exit event in the stub.

lib/posix-process/exit.c
michpappas: The code was originally written in 2024. Bumping the date effectively gives up copyright for the previous year(s): https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated.

lib/posix-process/exit.c
michpappas: You can have a look at `uk_sched_thread_exit()`, it uses the `noreturn` attribute and is additionally with a `UK_CRASH()`.

lib/posix-process/exit.c
michpappas: Yes, and additionally we need to do a bit more work on passing consistent info in the event data. Right now you get things like pid / tid on execve and exit events, yet clonetab does not pass this info. We can rework this part once the main logic is in, and we shift our focus to process resource management by the various libraries.

lib/posix-process/exit.c
michpappas: I wonder who reviewed that PR

lib/posix-process/exit.c
michpappas: Tbh, I didn't like it either but left it as is thinking that once we move signals out of libposix-process we will need to export this and likely other functionality too, so we might as well do it all at once. But since it did stick out to you as well, I'll add some macros now.

lib/posix-process/exit.c
michpappas: A return value > 1 signifies a different than a return value of 0. The former means that the parent ignores `SIGCHLD` or has set `SA_NOCLDWAIT`, while the latter signifies an error while sending the signal. So these cannot be merged into a single condition. Before you ask, I chose to treat this case separately and return an positive error code, partially because I didn't want to make any assumptions on what error code the signalling function returns, and partially because I could't find a suitable error code for  the case where I currently return > 0.

lib/posix-process/exit.c
michpappas: Normally yes, but it is possible that the application implements its own init logic that somewhat misbehaves. So, although not the advertised behavior, to make sure that we libraries are able to release their resources, I opted in for a more flexible implementation that allows pid 1 to terminate first.

lib/posix-process/exit.c
michpappas: Same reason as described in https://github.com/unikraft/unikraft/pull/1630/files#r2032776145

lib/posix-process/exit.c
michpappas: Since this needs to reparent every child, I think it's okay to iterate through all of them.

mogasergiu: As I missed the `pchild->parent = ` part.

lib/posix-process/exit.c
mogasergiu: Why is `0` an error? `pprocess_signal_send` seems to return `0` on success

lib/posix-process/exit.c
mogasergiu: But my other point is, I don't think we need a check in the loop, I think we can do the loop within the checks instead, as they don't seem to have to do anything with the particular element being iterated upon.

lib/posix-process/exit.c
michpappas: Yes, zero is not an error, but both > 0 and < 0 signify different cases we need to handle. Also in my message above it should be: "A return value > 0 signifies..."

lib/posix-process/exit.c
michpappas: So are you proposing something along the lines of checking whether init exits before we enter the loop, based on the result reparent to init or to `NULL` in the loop? If that is the case, I'm pretty sure the compiler will do it anyway.

lib/posix-process/exit.c
michpappas: This loop is now gone, see https://github.com/unikraft/unikraft/pull/1630#discussion_r2031065096

lib/posix-process/signal/deliver.c
michpappas: I don't know, initially I set it to info, later I changed my mind and changed to warning. Now I think it should be info again.

lib/posix-process/wait.c
michpappas: I just forgot to remove it :)

lib/posix-process/wait.c
michpappas: Because I was concerned placing an `unlikely` condition together with a non-unlikely one as that may result into the opposite of the desired effect.

lib/posix-process/wait.c
michpappas: I would say, neither likely or unlikely. It depends on the application, so I refrained from making assumptions.

lib/posix-process/process.h
michpappas: Honestly I have no idea. Perhaps when I first prototyped this a year ago I wasn't familiar with pprocess and it remained. Will fix.

lib/posix-process/execve.c
michpappas: But what will that save? We need to iterate through all children anyway. Or is it about micro-optimizations on the branch predictor?

andreittr: It's not obvious that doing several extra memory writes would be faster than a branch. On modern arches very predictable branches (like this one) are basically free. I don't think it's worth the complete break in abstraction layers to think about "temporary reparenting" when you just want to kill all threads (except perhaps one).

Also: you're assuming that the current thread is part of the process being killed, which needn't always be the case. Doing a branch on a pointer comparison is simpler to parse and more obviously correct.

lib/posix-process/exit.c
mogasergiu: I am proposing 
```C
if (pprocess_init && pprocess != pprocess_init)
     for loop for pchild->parent = pprocess_init; uk_list_add
else
     for loop for pchild->parent = NULL
```

> I'm pretty sure the compiler will do it anyway.

If you say so...

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+#if CONFIG_LIBPOSIX_PROCESS_SIGNAL
+#define PTHREAD_BLOCKING_ON_SIGNAL(_pt)					\
+	((_pt)->state == POSIX_THREAD_BLOCKED_SIGNAL)
+#else /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+#define PTHREAD_BLOCKING_ON_SIGNAL(_pt)		    0
+#endif /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+
+UK_EVENT(POSIX_PROCESS_EXIT_EVENT);
+
+int pprocess_exit_status;
```
mogasergiu, line 37: I see this is not `static` and is also exported in the header. Is there anyone else using it?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+			    PTHREAD_BLOCKING_ON_SIGNAL(pt)) {
+				uk_semaphore_up(&parent_process->wait_semaphore);
+				break;
+			}
+		}
+	} else if (nowait == true) { /* release now */
+		pprocess_release(pprocess);
+	}
+
+	uk_semaphore_up(&pprocess->exit_semaphore);
```
mogasergiu: Is this safe to do if the `pprocess_release` branch above was taken?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+	if (parent_process && nowait == false) {
+		uk_pprocess_foreach_pthread(parent_process, pt, ptn) {
+			if (PTHREAD_WAITING_FOR_PID(pt, uk_sys_getpid()) ||
+			    PTHREAD_BLOCKING_ON_SIGNAL(pt)) {
+				uk_semaphore_up(&parent_process->wait_semaphore);
+				break;
+			}
+		}
+	} else if (nowait == true) { /* release now */
+		pprocess_release(pprocess);
```
mogasergiu, line 227: Can you leave a comment here that says that init process needs to manually call the release function since it cannot reach `nowait == true`?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+
+#if CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS
+#include <uk/process.h>
+#include <uk/semaphore.h>
+
+#include "process.h"
+#include "signal/siginfo.h"
+
+#define PPROCESS_STATE_CHANGED(_pp, _state)				\
+	((_state & WEXITED) && ((_pp)->state == POSIX_PROCESS_EXITED ||	\
```
mogasergiu: Probably should put paranthesis around `_state`

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+ */
+static int pprocess_wait(pid_t pid, siginfo_t *si, int opt)
+{
+	struct posix_process *this_process;
+	struct posix_thread *this_thread;
+	struct posix_process *waitable = NULL;
+	struct posix_process *child = NULL;
+	int ret;
+
+	UK_ASSERT(pid > 0 || pid == -1);
```
mogasergiu, line 77: Maybe `UK_ASSERT(pid >= UK_PID_INIT || pid == UK_PID_WAIT_ANY)` would look better?

lib/posix-process/exit.c
andreittr: After the latest change, this line is a UAF if `nowait`, because `pprocess` has already been released.
Assuming we move the semaphore call before `release`, can we be sure that no other code touches `pprocess` or its fields after `uk_semaphore_up(&pprocess->exit_semaphore)` returns?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+	int ret;
+
+	if (options)
+		if (unlikely(!(options & (WNOHANG | WUNTRACED | WCONTINUED))))
+			return -EINVAL;
+
+	if (uk_list_empty(&uk_pprocess_current()->children))
+		return -ECHILD;
+
+	if (pid < -1 || pid == 0) /* process groups */
```
mogasergiu, line 156: Ditto `UK_PID_INIT`/`UK_PID_WAIT_ANY`?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
 {
 	return -ECHILD;
 }
+#endif /* !CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS */
 
 #if UK_LIBC_SYSCALLS
 pid_t wait3(int *wstatus, int options, struct rusage *rusage)
 {
-	return uk_syscall_e_wait4((long) -1, (long) wstatus,
-				  (long) options, (long) rusage);
```
mogasergiu, line 64: The `-1`'s from here maybe should be `UK_PID_WAIT_ANY`?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+	}
+
+out:
+	this_thread->state = POSIX_THREAD_RUNNING;
+	return ret;
+}
 
+UK_SYSCALL_R_DEFINE(pid_t, wait4, pid_t, pid,
+		    int *, wstatus, int, options,
+		    struct rusage * __unused, rusage)
```
mogasergiu: No need for `__unused` the macro already adds it

lib/posix-process/Config.uk
```diff
@@ -29,6 +29,7 @@ endif
 menuconfig LIBPOSIX_PROCESS_MULTIPROCESS
 	bool "Multiprocess support"
 	select LIBPOSIX_PROCESS_MULTITHREADING
+	select LIBUKLOCK
```
mogasergiu, line 35: Maybe `LIBUKLOCK_SEMAPHORE` as well?

lib/posix-process/signal/deliver.c
```diff
@@ -20,8 +21,16 @@ void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
...
-	pprocess_kill(uk_pprocess_current());
+	struct posix_process *pprocess;
+
+	pprocess = uk_pprocess_current();
+	UK_ASSERT(pprocess);
+
+	uk_pr_info("pid: %d terminated by signal\n", pprocess->pid);
+
+	pprocess_exit(pprocess, POSIX_PROCESS_KILLED, sig);
+	uk_sched_yield();
```
mogasergiu: Correct me if I am wrong: `pprocess_exit` calls `pprocess_exit_pthread` for each of its pthreads (including current one) which calls `uk_sched_thread_terminate` for all threads **BUT** the current one. Am I missing something or could this thread actually return from `uk_sched_yield`? What stops it from not doing so.

mogasergiu (changes requested):

lib/posix-process/exit.c
michpappas: Yes, it is used in the upcoming PR for shutdown.

lib/posix-process/exit.c
michpappas: init is a bit special as it cannot be waited for. So it can either be released via the return path (), otherwise if it hasn't returned until shutdown, it will be released (after being force-terminated) with the system going down.

lib/posix-process/signal/deliver.c
michpappas: That's right, it's a leftover from a previous iteration that I forgot to replace with `uk_sched_thread_exit()`. Will fix 👍🏼

lib/posix-process/wait.c
michpappas: If I used `UK_PID_INIT` here, that would imply that `UK_PID_INIT` is always 1, which defeats the purpose of the macro. In other words, if we ever changed `UK_PID_INIT` to a different value, this code would break. Of course it's highly unlikely that this will ever happen, but at least that was my rationale for using an integer value. For `UK_PID_WAIT_ANY`, I am also inclined even to replace with `-1` in the single place where it's used, as both 0 and -1 are values defined by the man (and likely POSIX) that libc does not even provide macros for.

lib/posix-process/wait.c
michpappas: See https://github.com/unikraft/unikraft/pull/1630#discussion_r2035688158

lib/posix-process/wait.c
michpappas: See https://github.com/unikraft/unikraft/pull/1630#discussion_r2035688158

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+	struct posix_thread *parent_pthread;
+	struct uk_thread *thread;
+	int ret;
+
+	UK_ASSERT(state == POSIX_THREAD_EXITED ||
+		  state == POSIX_THREAD_KILLED);
+
+	UK_ASSERT(pthread);
+	UK_ASSERT(pthread->state != POSIX_THREAD_EXITED ||
+		  pthread->state != POSIX_THREAD_KILLED);
```
andreittr, line 105: Is this assert correct? I assume you want to ensure that `pthread` is not yet exited or killed. In that case it should be `&&` between conditions.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+			 * discarded and the child dies immediately."
+			 */
+			uk_pr_info("Parent ignores SIGHLD, terminating\n");
+			nowait = true;
+		} else if (unlikely(ret < 0)) { /* no choice here but crash */
+			UK_CRASH("Could not signal parent (%d)\n", ret);
+		}
+#else /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+		nowait = true;
+#endif /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
```
andreittr: Just to make sure I got this straight, in a config with multiproc but w/o signals (which AFAICT is possible):
- if `!parent_process` we're dealing with PID1 => `nowait = false` => no semaphore up & no `pprocess_release`
- if `parent_process` `nowait = true` => no semaphore up & yes `pprocess_release`

The first seems reasonable (there's nobody to wait for PID1, nor should we release its struct yet), but the 2nd one strikes me as off; shouldn't `wait()` be independent from signals? As I read it, without signals, children die immediately and wait calls never return :thinking: .

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+	return pprocess_signal_send(pprocess, SIGCHLD, NULL);
+}
+#endif /* CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+
+/* Exit AND release pthread.
+ * Notice: Unlike pprocesses, pthreads are released upon exit.
+ */
+void pprocess_exit_pthread(struct posix_thread *pthread,
+			   enum posix_thread_state state,
+			   int exit_status __unused)
```
andreittr, line 92: Is `exit_status` intended to be used in the future? It can't be kept anywhere, since `pthread` is freed by the end. Is this meant for a future impl of exit events, with richer `event_data`?

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+	UK_ASSERT(pprocess);
+	UK_ASSERT(pprocess->state == POSIX_PROCESS_RUNNING);
+
+	uk_pr_debug("pid %d: exit tid %d\n", pprocess->pid, pthread->tid);
+
+	/* May be NULL if init terminated early */
+	parent_pthread = pthread->parent;
+
+	/* Update state */
+	pthread->state = state;
```
andreittr, line 117: Why do we bother setting the state if we end up freeing `pthread` before returning? We don't pass `pthread` along to anything else before freeing either.

lib/posix-process/exit.c
```diff
@@ -0,0 +1,318 @@
...
+	if (state == POSIX_PROCESS_EXITED)
+		pprocess->exit_status = exit_status & 0xff;
+	else /* POSIX_PROCESS_KILLED */
+		pprocess->exit_status = exit_status;
+	pprocess->state = state;
+
+	/* Reparent child processes to init. The pprocess_parent
+	 * variable still refers to the original parent.
+	 */
+	pprocess_reparent_children(pprocess);
```
andreittr, line 206: I think you meant "parent_process variable".
Though, on 2nd thought, why is this comment needed? `pprocess_reparent_children(pprocess);` does not change `pprocess->parent` AFAICT.

lib/posix-process/process.h
```diff
@@ -120,9 +108,63 @@ struct posix_process *tid2pprocess(pid_t tid);
...
+/**
+ * INTERNAL. Exit process
+ *
+ * Performs process termination tasks and raises POSIX_PROCESS_EXIT_EVENT.
+ * Does not release the process. Upon completion, if the process has not been
+ * reaped by an existing wait, the process becomes a zombie and its resources
+ * remain allocated until reaped by a subsequent call to wait().
+ *
+ * Notice: It's the caller's responsibility to terminate the current uk_thread
+ *         upon return, if calling this function from uk_thread_current().
```
andreittr: Maybe better phrasing would be "if the calling thread is a member of the terminating process".

As it stands it's a bit confusing; technically function calls always come from `uk_thread_current()`.

lib/posix-process/process.h
```diff
@@ -120,9 +108,63 @@ struct posix_process *tid2pprocess(pid_t tid);
...
 
-void pprocess_kill(struct posix_process *pprocess);
+/**
+ * INTERNAL. Exit and release a pthread
+ *
+ * Performs termination tasks and raises the POSIX_PROCESS_EXIT_EVENT.
+ * Unlike processes, pthreads are released upon exit. Besides the pthread,
+ * this function terminates the underlying uk_thread, unless called by
+ * uk_thread_curent(). In this last case, the caller should take care of
+ * terminating the underlying uk_thread, as needed.
```
andreittr: Similarly here, "unless called _on_ uk_thread_current()'s pthread" would be clearer.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
 
-/* FIXME: Provide with sys/wait.h */
-struct rusage;
-typedef unsigned int idtype_t;
-typedef unsigned int id_t;
-typedef unsigned int siginfo_t;
+	this_process = this_thread->process;
+	UK_ASSERT(this_process);
 
+	UK_ASSERT(!uk_list_empty(&this_process->children));
```
andreittr: Minor: I don't see the benefit of keeping this assert + 2 separate `-ECHILD` error checks in the syscalls, versus moving the childlist check & `-ECHILD` return here. It would even clean up the syscalls, not having them touch pprocess struct fields directly.

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+	this_thread->state = POSIX_THREAD_BLOCKED_WAIT;
+
+	if (pid > 0) {
+		child = pid2pprocess(pid);
+		/* Child does not exist or not our child. This also
+		 * covers the case where the process chose to ignore
+		 * SIGCHLD or set SA_NOCLDWAIT and therefore _exit()
+		 * terminated it without further ado.
+		 */
+		if (unlikely(!child || child->parent != this_thread->process)) {
```
andreittr: We have `this_process`, any reason we use `this_thread->process` here?

lib/posix-process/wait.c
```diff
@@ -1,78 +1,250 @@
...
+		if (pid > 0)
+			waitable = pprocess_state_changed(child, opt);
+		else /* any */
+			waitable = pprocess_child_state_changed(opt);
+
+		if (waitable) {
+			ret = waitable->pid;
+			if (!(opt & WNOWAIT))
+				pprocess_reap(waitable, si);
+			goto out;
```
andreittr: `break;` might be clearer and in-line with the `opt & WNOHANG` block lower down.

andreittr (commented):
Final full pass on my end, this is as reviewed as it's ever going to be.

lib/posix-process/exit.c
michpappas: Yeah okay, so if signals are not selected we are in uncharted waters wrt the behavior of other systems. I interpolated the state defined in the man from "if the parent ignores SIGCHLD" when signals are selected, to "if the parent does not support signals at all" when signals are not selected. On second thought I'm not sure if that's right, as in the case where the parent ignores SIGCHLD, it has done that explicitly by changing the default disposition. I think I will remove line 182, and let it fall back to the default behavior when signals are not selected.

lib/posix-process/exit.c
michpappas: I went through that thinking path too, and was close to deprecating it, but in the end decided against it. I'd say let's keep it until posix-process matures up a bit more, to avoid adding it back if we need it later.

lib/posix-process/exit.c
michpappas: Only in case some of the event handlers needs to check / assert against the thread's the state. That could also be a case for exit_status at some point. We can also remove it now, I don't have a strong opinion.

lib/posix-process/exit.c
michpappas: Leftover from a previous iteration where I was also unlinking from parent. Will remove.

lib/posix-process/wait.c
michpappas: No reason, perhaps `this_process` was added later and forgot to update that line.

lib/posix-process/wait.c
michpappas: Yeah okay I was between two minds on where to delegate the responsibility, then I thought it's easier to see the check at the syscall when going reading the syscall implementations, but you have a point. Also it's easy to miss it if adding another call later. Will update.

lib/posix-process/exit.c
andreittr: We can leave state updates in for now until we decide on event_data contents. A memory write on thread termination is trivial.

lib/posix-process/process.h
andreittr: The [latest change](https://github.com/unikraft/unikraft/compare/417c1a246cf785b6ef4a5f2091b6ffc695855d4f..fcc4f8c49d460021af5dcea4230b80b630728b57#diff-efec2683f632d2de2fdf79c43b8941cbd262040f18925c93d03ac8b59c0cf4e8R122) has a duplicate "unless called by unless called on" on lines 121-122.

michpappas: Ooops, sorry about that

andreittr (approved):
With these latest changes all looks good, thanks a lot!

michpappas:
Another rebase from `staging` + fix a conflict resolution in last reabse.

mogasergiu (approved):
Let's merge! :steam_locomotive: choo choo chuga chuga

andreittr:

PR #1632, opened 2025-04-15
lib/posix-fdio: Add generic sendfile syscall + fallback

andreittr:
### Description of changes

This PR adds a generic sendfile syscall implementation that accepts any combination of ukfiles and vfscore files.
Pure ukfile `sendfile` is delegated to a dedicated operation `uk_sys_sendfile` that for now always returns error -- in agreement with `sendfile(2)` that requires input files to be mmap-able, which no ukfiles currently are.
Mixed or pure vfscore sendfile uses a fallback implementation with two I/O ops over a temp buffer.

To achieve this, vfscore code has been adapted with the _minimum_ changes necessary; improving vfscore in any way is outside the scope.

A true single-copy implementation of `uk_sys_sendfile` requires #1411 and driver support.

### Prerequisite checklist

### Base target

### Additional configuration

N/A


andreittr:
Rebased to release 0.19.0.

lib/posix-fdio/sendfile-shim.c
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * Shim implementations involving vfscore are not synchronized.
+ */
+
+static inline
+int shim_get(int fd, union uk_shim_file *sf)
+{
+	int r = uk_fdtab_shim_get(fd, sf);
+
+	if (r < 0)
```
mogasergiu: why no unlikely? :thinking:

lib/posix-fdio/sendfile-shim.c
```diff
@@ -0,0 +1,277 @@
...
+			if (r <= 0)
+				break;
+			remain -= r;
+			written += r;
+			sent += r;
+			count -= r;
+		} while (remain);
+		/* Rewind infile offset to before last read & break on error */
+		if (unlikely(remain)) {
+			off_t rc __maybe_unused;
```
mogasergiu: Is the `__maybe_unused` needed? Does it warn without it?

lib/posix-fdio/sendfile-shim.c
```diff
@@ -0,0 +1,277 @@
...
+#endif /* CONFIG_LIBVFSCORE */
+
+UK_SYSCALL_R_DEFINE(ssize_t, sendfile, int, out_fd, int, in_fd, off_t *, offp,
+		    size_t, count)
+{
+	union uk_shim_file out_sf;
+	union uk_shim_file in_sf;
+	int out_type;
+	int in_type;
+	ssize_t ret;
```
mogasergiu, line 213: So I assume the switch has all possible cases included, since `ret` is uninitialized and there is no `default`

mogasergiu (commented):
LGTM just a couple of questions

lib/posix-fdio/sendfile-shim.c
andreittr: Because where `shim_get` is called, its result is already interpreted in a [hinted branch](https://github.com/unikraft/unikraft/pull/1632/files/aae3f05debbaf3d9b4fe49068294239406bb3357..dcc31a58d732a473d483cb2ee1a35c2743d256b7#diff-d8edc20028129c5c579b9cf8aa97622cbdd5b2da5dec8fc79f71e66d66b96131R221).

~~But on closer look, I don't see a good reason for keeping `shim_get` around any more; AFAICT it doesn't do anything that `uk_fdtab_shim_get` doesn't already do; must've been left over from a previous iteration. Will fold into its callsites.~~
edit: actually I see why it was used: to isolate in 1 place the validation that fdtab only returns the file types that we support in the current build (i.e., no UK_SHIM_LEGACY if no vfscore enabled). I'll keep the function, but remove the early return, as it's already handled gracefully by the switch below, and we don't need any error propagation from fdtab (we return -EBADF on fdtab lookup regardless of internal error).

lib/posix-fdio/sendfile-shim.c
andreittr: Huh, whaddya know, it doesn't warn w/o it. Good to know we need that only for _explicit_ deadcode elimination (like with `UK_ASSERT`, or args to `uk_print*`). I guess use in a branch (even an empty one that will get eliminated by DCE) counts as "used" from the compiler's PoV :smile: .

lib/posix-fdio/sendfile-shim.c
andreittr: Yep, the `default` and init value are left out on purpose, so that the compiler would warn loudly had I missed handling any of the values of `enum shim_combo`.

andreittr:
Pushed addressing comments.

mogasergiu (approved):

michpappas (approved):

PR #1633, opened 2025-04-15
lib/posix-process: Graceful shutdown 

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

New options:

* `LIBUKBOOT_INIT`: Enables `/sbin/init` logic.
* `LIBUKBOOT_GRACEFUL_SHUTDOWN`: Enables graceful shutdown.
* `LIBUKBOOT_GRACEFUL_SHUTDOWN_SIGNAL`: Controls the signal sent on graceful shutdown (defaults to `SIGTERM`).
 * `LIBUKBOOT_GRACEFUL_SHUTDOWN_TIMEOUT`: Controls the time to wait for before forcefully terminating a process (defaults to 5 sec).

### Description of changes

This series implements graceful shutdown at the process level. The objective is to have processes and their associated libraries release their resources both at the application and kernel level, before the system goes down. This is implemented with coordination of  `libposix-process` and `libukboot`. The most significant changes are descirbed below:

#### libposix-process

Add an initcall to handle system shutdown. This serves the purpose of cleaning up process resources at the kernel level, and return the application's exit status back to Unikraft. The implementation varies depending on system configuration. When multithreading is not enabled, the handler simply returns the exit status of the application. When multithreading or multiprocess are enabled, the handler sends a SIGTERM to the appliation and blocks until the application returns. It then proceeds to forcefully terminate any remaining process. When signals are not enabled the implementation  falls back into forceful termination.

#### libukboot

Add `/sbin/init` logic for multiprocess configurations. This is controlled by a new option, CONFIG_LIBUKBOOT_INIT. The init logic is implemented as part of PID 1, and spawns the application into a new process. Besides fostering reparented orphans, and reaping children, it also manages system shutdown and returning the application's exit status back to ukboot.
    
A second option, namely CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN, provides the ability to signal children with a configurable signal, and block until they exit until a configurable timeout. If the timeout is reached and children are still alive, init returns and libposix-process forcefully terminates all remaining processes.

Depends-on: #1598 
Depends-on: #1618
Depends-on: #1619
Depends-on: #1620
Depends-on: #1626
Depends-on: #1627
Depends-on: #1628
Depends-on: #1630

Supersedes: #1364


lib/posix-process/shutdown.c
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#if CONFIG_LIBPOSIX_PROCESS_MULTITHREADING
+/* We can enter this path if UK_PID_INIT returns, or
+ * while it's still running via the shutdown signal.
+ */
+#if CONFIG_LIBPOSIX_PROCESS_SIGNAL
+/* Since signals are enabled, if the application has not returned,
+ * send a signal and block until it terminates. Once the application
+ * is terminated, kill any remaining processes.
```
andreittr, line 40: At first I was confused as to why block indefinitely after `SIGTERM` and not force-kill. But I realised that _needing_ to force-kill PID1 is not a well-defined state and suggests a misconfigured kernel. At that point one might as well force-shutdown the entire machine on an external timeout.

I do see potential in harmonizing behavior with the nosignals impl, and setting `ctx->exit_code` to `SIGKILL`, but that would require another Kconfig timeout, with the practical result of timing out on the minimum of the two. In that light, requring PID1 to never hang is the lesser evil.

lib/posix-process/include/uk/process.h
```diff
@@ -226,8 +226,27 @@ int uk_posix_process_create_pthread(struct uk_thread *thread);
 
 #endif /* CONFIG_LIBPOSIX_PROCESS_MULTITHREADING */
 
-#if CONFIG_LIBPOSIX_PROCESS_EXECVE
+#if CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS
+
+typedef int (*uk_posix_process_mainlike_func_t)(int argc, char *argv[]);
```
andreittr: The `*_t` namespace is [reserved by POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_02_02) and we should avoid using it. Just `*_func` is fine for func typedefs, [other new code](https://github.com/unikraft/unikraft/blob/staging/lib/ukfile/include/uk/file.h#L38) uses that as well.
Also (and since this is a public API), I think `uk_posix_process_main_func` would be shorter and clearer as a type name.

lib/posix-process/include/uk/process.h
```diff
@@ -226,8 +226,27 @@ int uk_posix_process_create_pthread(struct uk_thread *thread);
...
+/**
+ * Spawn a process that jumps into function.
+ *
+ * DO NOT USE. This is only necessary when we create a new process
+ * for main() in multiprocess.
+ *
+ * @param fn      Function to jump to.
+ * @param a       Allocator to assign to the new process.
+ * @param thread  Container thread to create process from.
+ * @param parent  Parent thread.
```
andreittr: Docstring is mismatched with signature: `a`, `thread`, `parent` are missing from sig, `argc` & `argv` are missing from doc.

lib/posix-process/process.c
```diff
@@ -344,6 +345,73 @@ int pprocess_create(struct uk_alloc *a,
 	return ret;
 }
 
+pid_t uk_posix_process_run(uk_posix_process_mainlike_func_t fn,
```
andreittr: Return type mismatch vs. declaration in `<uk/process.h>`: `pid_t`, expected `int`.

lib/posix-process/process.c
```diff
@@ -344,6 +345,73 @@ int pprocess_create(struct uk_alloc *a,
...
+	struct uk_sched *s = uk_sched_current();
+	struct clone_args cl_args;
+	struct uk_thread *thread;
+	struct uk_thread *parent;
+	int ret;
+
+	UK_ASSERT(s);
+
+	parent = uk_thread_current();
+	UK_ASSERT(ukthread2pid(parent));
```
andreittr: I believe this might not behave as intended: `ukthread2pid` [does not return 0 on failure](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/process.c#L615), rather `-ENOTSUP` (or a straight assert).
I assume you wanted to assert that `uk_posix_process_run` is only run in the context of a parent with a real PID (I also assume there's a case where `pthread(parent)->process != NULL` _but also_ that process has pid 0, otherwise asserting the pprocess ptr directly is maybe better).

Also, it's not obvious why this function particularly needs to assert this condition, as I don't see `parent`'s pid or pproccess-ness used directly anywhere. I assume it's for pprocess_create, but shouldn't that function be in charge of validating/asserting its own input preconditions?

edit: whoops, wrong link earlier

lib/posix-process/process.c
```diff
@@ -344,6 +345,73 @@ int pprocess_create(struct uk_alloc *a,
...
+	if (unlikely(!thread)) {
+		uk_pr_err("Could not create thread\n");
+		return -ENOMEM;
+	}
+
+	/* Create new process */
+	ret = pprocess_create(uk_alloc_get_default(), thread, parent);
+	if (unlikely(ret)) {
+		uk_pr_err("Could not create process (%d)\n", ret);
+		return ret;
```
andreittr: `thread` is leaked unfreed by this error path.

lib/posix-process/process.c
```diff
@@ -344,6 +345,73 @@ int pprocess_create(struct uk_alloc *a,
...
+		.flags       = CLONE_VM | CLONE_VFORK,
+		.child_tid   = ukthread2tid(thread),
+		.parent_tid  = ukthread2tid(parent),
+		.exit_signal = SIGCHLD,
+	};
+	ret = pprocess_clonetab_init(&cl_args, sizeof(cl_args), 0,
+				     thread, parent);
+	if (unlikely(ret)) {
+		uk_pr_err("clonetab execution error (%d)\n", ret);
+		return ret;
```
andreittr: `thread` + its pprocess are leaked by this error path.

lib/posix-process/process.c
```diff
@@ -344,6 +345,73 @@ int pprocess_create(struct uk_alloc *a,
...
+	}
+
+	/* Raise the execve event */
+	event_data.thread = thread;
+	ret = pprocess_raise_execve_event(&event_data);
+	if (unlikely(ret < 0)) {
+		uk_pr_err("exeve event error (%d)\n", ret);
+		return ret;
```
andreittr: `thread` + its pprocess are leaked by this error path.
Also, the process is "live" from the point of view of external libs after the clonetab entries are run, so we should now run their term code too.

lib/posix-process/execve.c
```diff
@@ -83,6 +83,11 @@ static int pprocess_cleanup(struct posix_process *pprocess)
 	return 0;
 }
 
+int pprocess_raise_execve_event(struct posix_process_execve_event_data *data)
+{
+	return uk_raise_event(POSIX_PROCESS_EXECVE_EVENT, data);
+}
```
andreittr, line 89: Since the implementation is trivial, this should be a static inline or a macro in `process.h`.
Then again:
```c
pprocess_raise_execve_event(event_data)
uk_raise_event(POSIX_PROCESS_EXECVE_EVENT, event_data)
```
isn't a huge difference in charcount, and the latter uses only publicly exported names.
For my own taste, the latter reads clearer with fewer logical steps, but could go either way; your call.

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,43 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
 	  if it is requested by a driver on the shutdown_req event queue
 	  (e.g., ACPI power button event).
 
+config LIBUKBOOT_INIT
+	bool "Provide /sbin/init"
```
andreittr: The description would better read as "Provide /sbin/init logic". As it is, the pathname suggests it would create an actual `/sbin/init` file.

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,43 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
...
+		select this in a multiprocess configuration, you need
+		to provide your own init.
+
+menuconfig LIBUKBOOT_GRACEFUL_SHUTDOWN
+	bool "Graceful shutdown"
+	depends on LIBUKBOOT_INIT
+	help
+		Signal the application on system shutdown to give them
+		a chance to terminate gracefully. If timeout is
+		reached, children are killed forecefully.
```
andreittr: Ambiguous phrasing, suggest replacing "them" & "children" by "it" & "all processes", respectively.

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,43 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
...
+	default 15 if (ARCH_ARM_64 || ARCH_X86_64)
+	help
+		Signal to send the application to gracefully terminate
+		before shutting down. Defaults to SIGTERM.
+
+config LIBUKBOOT_GRACEFUL_SHUTDOWN_TIMEOUT
+	int "Termination timeout (sec)"
+	default 5
+	help
+		Time (sec) to wait for children to gracefully terminate
```
andreittr: Would this be worth expressing in finer units, e.g., `ms`? I can imagine low-latency configurations where even 1sec is a long time.

lib/ukboot/boot.c
```diff
@@ -111,7 +111,8 @@ extern char **boot_argv;
 extern int boot_argc;
 
 int main(int argc, char *argv[]) __weak;
-static inline int do_main(int argc, char *argv[]);
+int do_main(int argc, char *argv[]);
+int do_init(int argc, char *argv[]);
```
andreittr: I think these two can go into an internal header (`init.h` IMO), along with a short description of the semantics of each, as it's not immediately obvious what "do_init" does without the context of this PR.
And `do_init` is worth guarding behind `CONFIG_LIBUKBOOT_INIT`, as it really doesn't make sense to declare it otherwise.

lib/ukboot/Makefile.uk
```diff
@@ -5,6 +5,8 @@ CXXINCLUDES-$(CONFIG_LIBUKBOOT)		+= -I$(LIBUKBOOT_BASE)/include
 LIBUKBOOT_ASINCLUDES-y			+= -I$(UK_PLAT_COMMON_BASE)/include
 LIBUKBOOT_CINCLUDES-y			+= -I$(UK_PLAT_COMMON_BASE)/include
 
+LIBUKBOOT_SRCS-y += $(LIBUKBOOT_BASE)/init.c
```
andreittr: This should be `LIBUKBOOT_SRCS-$(CONFIG_LIBUKBOOT_INIT)`, as it pulls in some headers that aren't always available (and is only used if init).

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,43 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
 	  if it is requested by a driver on the shutdown_req event queue
 	  (e.g., ACPI power button event).
 
+config LIBUKBOOT_INIT
+	bool "Provide /sbin/init"
+	depends on LIBPOSIX_PROCESS_MULTIPROCESS
+	select LIBPOSIX_PROCESS_SIGNAL
+	select LIBPOSIX_PROCESS_SIGNALFD
+	select LIBUKBOOT_MAINTHREAD
```
andreittr, line 205: I normally wouldn't gripe about this, since posix-process (an implicit dependency) pulls in (no)libc, but seeing as `ukboot` is a lower-level component, it would be worth explicitly saying that it uses libc code directly (`select LIBNOLIBC if !HAVE_LIBC`).

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+#define GRACEFUL_SHUTDOWN_TIMEOUT			    \
+	CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN_TIMEOUT
+#endif /* CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN */
+
+/* We use the shell convention to set the high bit
+ * when the process is terminated by signal.
+ */
+#define TERM_BY_SIGNAL_BIT  0x80
+
+int do_main(int argc, char *argv[]);
```
andreittr: See comment on internal header for this; would prevent duplicate code here.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+
+pid_t application_pid;
+
+int application_returned;
+int application_status;
```
andreittr: These 3 should probably be `static`.

Actually, after reading the entire file, is there any reason to have them in global scope? They could very well be locals in `do_init` and passed to `wait_nonblocking` as args.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+static inline pid_t wait_nonblocking(void)
+{
+	int wstatus;
+	pid_t pid;
+
+	pid = waitpid(-1, &wstatus, WNOHANG);
+
+	/* Save the application's return code. This will
+	 * the value Unikraft returns upon exit. If the
+	 * application was killed we use convention to set
```
andreittr: Minor signal drop:
- "This will the" -> "This will be the"
- "we use convention" -> "we use the convention"
(or, rephrase, "[was killed] we set the high bit as by convention.")

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+	time_t now;
+#endif /* CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN */
+	int sigfd;
+	int ret;
+
+	/* Block all signals except SIGCHLD and SIGTERM */
+	sigemptyset(&mask);
+	sigaddset(&mask, SIGCHLD);
+	sigaddset(&mask, SIGTERM);
+	sigprocmask(SIG_BLOCK, &mask, NULL);
```
andreittr: If my reading of the manpages is correct, this does the opposite of what the comment says: it blocks only SIGCHLD and SIGTERM.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+
+	/* Create signalfd for SIGCHLD and SIGTERM */
+	sigfd = signalfd(-1, &mask, 0);
+	if (sigfd < 0) {
+		uk_pr_err("signalfd error (%d)\n", -errno);
+		return -errno;
```
andreittr: `errno` is positive, is there a reason we want to return (& print) negative error codes?

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+			break;
+		}
+
+		if (info.ssi_signo == SIGCHLD) {
+			/* Reap children. If the last child terminated,
+			 * initiate shutdown.
+			 */
+			ret = wait_nonblocking();
+			if (unlikely(ret < 0 && ret != -ECHILD))
+				UK_BUG();
```
andreittr: Could probably be replaced by `UK_ASSERT(ret >= 0 || ret == -ECHILD);`.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+
+	kill(0, GRACEFUL_SHUTDOWN_SIGNAL);
+
+	/* Wait for processes to terminate gracefully. We don't
+	 * differentiate between the application processes and
+	 * reparented children, as these are likely the result
+	 * of daemonization so we also need to treat as equal.
+	 */
+	now = time(NULL);
+	while (time(NULL) - now < GRACEFUL_SHUTDOWN_TIMEOUT) {
```
andreittr: Considering this loop condition, the naming of `now` is very misleading :smile: .

Also, the entire loop should probably be a `do { ... } while ()`, because you don't want to call `time` twice before doing anything useful.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	/* Close the signalfd */
+	close(sigfd);
+
+#if CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN
+	/* Initiate graceful shutdown */
+	uk_pr_debug("Signalling children\n");
+
+	kill(0, GRACEFUL_SHUTDOWN_SIGNAL);
```
andreittr: Does `kill` yield execution to give a chance for the recipient to respond to the signal? If not, the `wait_nonblocking` call lower down will never succeed on first try, and we always wait 1 second regardless.

lib/ukboot/init.c
```diff
@@ -0,0 +1,155 @@
...
+	 * differentiate between the application processes and
+	 * reparented children, as these are likely the result
+	 * of daemonization so we also need to treat as equal.
+	 */
+	now = time(NULL);
+	while (time(NULL) - now < GRACEFUL_SHUTDOWN_TIMEOUT) {
+		ret = wait_nonblocking();
+		if (ret == -ECHILD)
+			return application_status;
+		sleep(1);
```
andreittr: Ok, now I see why the 1s granularity in the timeout :facepalm: ; we do have `usleep` for shorter sleeps, but waking up every few ms is spammy.

What we need is a proper wait-with-deadline, which I realise is not practical in PID1 for a number of reasons. Best I can think of is keeping `sigfd` open just a tad longer and calling `poll()` on it: that supports millisecond precision and doesn't sleep more than it needs to.

andreittr (commented):
1st pass

lib/ukboot/init.c
andreittr: On a more caffeinated 2nd read: I get what you're trying to do here, but I think it's incomplete.
What you basically want is:
- Block all signals except `SIGCHLD` and `SIGTERM`, because we don't care about them
- Block `SIGCHLD` and `SIGTERM` too, because you're processing them via signalfd and not handlers
- => we want to block all signals, and have mask contain `SIGCHLD` and `SIGTERM` only for the signalfd call

Maybe something like
```c
sigfillset(&mask);
sigprocmask(SIG_BLOCK, &mask, NULL);
sigemptyset(&mask);
sigaddset(&mask, SIGCHLD);
sigaddset(&mask, SIGTERM);
```
then continue to signalfd.

lib/posix-process/include/uk/process.h
michpappas: > The *_t namespace is [reserved by POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_02_02) and we should avoid using it. Just *_func is fine for func typedefs, [other new code](https://github.com/unikraft/unikraft/blob/staging/lib/ukfile/include/uk/file.h#L38) uses that as well.

Indeed: https://unikraft.org/docs/contributing/coding-conventions#typedefs

> Also (and since this is a public API), I think uk_posix_process_main_func would be shorter and clearer as a type name.

I used `mainlike` instead of `main`, to indicate that it doesn't have to be a `main()` function, but the signature has to look like `main()`. Either way, this causes a warning because the type does not match with that of  `uk_thread_fn2` passed to `uk_thread_container_init_fn2()`. I still need to address that before we merge, so `uk_posix_process_mainlike_func` may be dropped altogether.

lib/posix-process/process.c
michpappas: Should be `pid_t` in declaration 👍🏼

lib/posix-process/process.c
michpappas: Maybe I didn't understand the proposed solution correctly, but the thing is that we don't know `pthread`, so we can't do `pthread(parent)->process`. Also, we can safely assume that there's no PID 0 (traditionally assigned to the scheduler). Nevertheless, I agree that comparing pointers is more safe. What we lack is the helpers to derive these for anything else but current. I can add a FIXME now and improve in the next release, as I have a few more things to address there.

> Also, it's not obvious why this function particularly needs to assert this condition, as I don't see parent's pid or pproccess-ness used directly anywhere. I assume it's for pprocess_create, but shouldn't that function be in charge of validating/asserting its own input preconditions?

~~Honestly I don't remember, and `pprocess_create()` does accept NULL as a parent. Could be a development artifact. I'll remove it.~~ We need it, it's for the parent's tid passed to clonetab, which implies that the parent thread must also be part of a process. Will rewrite to make it more clear.

lib/posix-process/execve.c
michpappas: The problem here is how `UK_EVENT()` is implemented. Raising the event, or even getting a pointer to the event,  needs to be in the same file that declares the event. So I cannot move it out, and consequently neither make it a static inline or a macro in `process.h`.

lib/ukboot/Config.uk
michpappas: Heheh so you found out why in https://github.com/unikraft/unikraft/pull/1633#discussion_r2047271495. Let's continue the discussion there :)

lib/ukboot/Config.uk
michpappas: The libc dependency is already explicitly stated for the libukboot:
```
menuconfig LIBUKBOOT
	bool "ukboot: Unikraft bootstrapping"
	select LIBNOLIBC if !HAVE_LIBC <------------------------------
	select LIBUKDEBUG
	select LIBUKARGPARSE
	select HAVE_BOOTENTRY
	imply LIBUKBOOT_MAINTHREAD if HAVE_SHUTDOWN_DISPATCHER
	imply LIBUKBOOT_SHUTDOWNREQ_HANDLER if HAVE_SHUTDOWN_DISPATCHER
	default y
```
Or maybe I didn't understand the comment correctly?

lib/ukboot/init.c
michpappas: > Actually, after reading the entire file, is there any reason to have them in global scope? They could very well be locals in do_init and passed to wait_nonblocking as args.

To avoid passing them as args 😄 I'll declare them `static`.

lib/ukboot/init.c
michpappas: Because the recipient is the kernel, where the convention is to use negative error codes.

lib/ukboot/init.c
michpappas: Good point. Generally speaking I would avoid making assumptions about the scheduler from the application. In practice though, since `kill()` results into a syscall, I expect that the scheduler will schedule the next thread.

lib/ukboot/init.c
michpappas: I'll rename to `start`. It also needs to be conditionally declared to `CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN`.

lib/ukboot/boot.c
michpappas: I was saving that for when we rearch libukboot, but we might as well do it now. The proposed name fits the future split of libukboot.

lib/posix-process/process.c
andreittr: > Also, we can safely assume that there's no PID 0

In that case, there is no way for the condition inside the assert to be false, as `ukthread2pid` will either return a non-zero "real" PID, or [`-ENOTSUP`](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/process.c#L615) (sorry, initial comment had wrongly pasted link), neither of which evaluate to `false` to trigger the assert.
Or are you implicitly relying on it triggering [this assert](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/process.c#L617) in the implementation?

> We need it, it's for the parent's tid passed to clonetab

In that case shouldn't we be asserting `ukthread2tid` (what we're actually using) instead / as well?

lib/posix-process/execve.c
andreittr: Oooh, ok, gotcha, makes sense in that case.

lib/ukboot/Config.uk
andreittr: Oh crap, I had missed that; sorry, disregard.

lib/ukboot/init.c
andreittr: > To avoid passing them as args 😄

So you'd rather have them be allocated .bss and be accessed as pc-relative-memory instead of regs? Tsk tsk.
It also makes `do_init` harder to understand, with it returning a global out of nowhere.

lib/ukboot/init.c
andreittr: So, just to make 100% sure, `tctx.exit_code` is meant to be negative on kernel-internal errors and positive only on user-supplied exit codes? `do_main` seems to call the workload-supplied `main` and return its exit code directly; I believe user `main` is supposed to return positive.

lib/ukboot/init.c
andreittr: The more I think about it, the more sense it makes to use [poll](https://www.man7.org/linux/man-pages/man2/poll.2.html) (or, in practice, epoll, as it's more efficient atm for [complicated reasons](https://github.com/unikraft/unikraft/blob/staging/lib/posix-poll/poll.c#L27)), and have `LIBUKBOOT_INIT` pull in `LIBPOSIX_POLL` (it's not an unreasonable ask for multi-process workloads). Sleep-waiting makes it both needlessly wake up for nothing _and_ oversleep when processes are done.

lib/ukboot/init.c
michpappas: Yes exactly, the only difference is that instead of blocking all signals, I only block the ones delivered via signalfd, and leave the rest to their default disposition. Since we're executing on a unikernel I didn't see a reason to block the rest, but let's do that anyway just in case.

lib/posix-process/process.c
michpappas: Yes and yes, when I said I will rewrite to make more clear I meant exactly that:
```diff
parent = uk_thread_current();
-UK_ASSERT(ukthread2pid(parent));
+parent_tid = ukthread2tid(parent);
+ UK_ASSERT(parent_tid > 0);
```

lib/ukboot/init.c
michpappas: Hmmm on second thought, so if we treat init as part of the application (I think we should) then I guess it should return positive values too. It also simplifies the handling logic on `uk_boot_entry()` as it maintains clear separation between application and kernel code, i.e. we treat `do_init()` the same way as do `do_main()`, assume that the main thread (if one) executes application code and so on. Will update to return a positive value 👍🏼

lib/ukboot/init.c
andreittr: I agree it's not stricly needed to block other signals, and might be more correct to block only SIGTERM and SIGCHLD, and leave the others at their default disposition. I had initially assumed the comment reflected the true intent (and the code was wrong) rather than the other way around (as it now seems is the case). Either way is OK by me as long as comment matches code.

lib/ukboot/init.c
michpappas: IIRC back in the first implementation I was printing a dot every second to show that the system has not hang, hence the sleep. I think it's better to use epoll to avoid the issues described, and should the requirement arises later, we can always add a config option that shows an animation instead.

lib/ukboot/init.c
michpappas: Alright, as init is not intended to be signalled by appliations I'll go ahead and block all signals and update the comment.

lib/ukboot/init.c
andreittr: Off the top of my head: If we want to print a "liveness" indicator, we can always have epoll wait with a deadline shorter than the total timeout, print a dot/spinner/whatever, then go back to wait for `MIN(liveness_period, remaining_timeout)`.

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,44 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
...
+config LIBUKBOOT_GRACEFUL_SHUTDOWN_SIGNAL
+	int "Termination signal"
+	default 15 if (ARCH_ARM_64 || ARCH_X86_64)
+	help
+		Signal to send the application to gracefully terminate
+		before shutting down. Defaults to SIGTERM.
+
+config LIBUKBOOT_GRACEFUL_SHUTDOWN_TIMEOUT
+	int "Termination timeout (msec)"
+	default 5000
```
andreittr, line 232: Negative values for `timeout` in (e)poll signify to "wait forever"; do we intend to support this behavior?
If yes, I believe it would be useful to mention this in the helptext: "Negative values will wait indefinitely".
If not, would it be worth limiting this to positive values only (via Kconfig or compile-time asserts)? Or is it safe to rely on the user to not shoot themselves in the foot here?

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+			application_status = TERM_BY_SIGNAL_BIT | WTERMSIG(wstatus);
+	}
+
+	return pid;
+}
+
+#if CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN
+/* Wait for processes to terminate gracefully. We don't differentiate between
+ * the application processes and reparented children, as these are likely the
+ * result of daemonization so we also need to treat as equal.
```
andreittr: Just to make sure, by not differentiating between app process and reparented children, you mean in the sense of sending them the shutdown signal and reaping them, _not_ in the sense of grabbing their exit code as the Unikraft exit code.
The latter is not true, because we treat the `application_pid` special when setting `application_returned` and `application_status`. I want to make sure this is intended and not an oversight.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+ */
+static int graceful_shutdown(int sigfd)
+{
+	struct epoll_event ev, events[1];
```
andreittr: It is common to reuse `ev` for both input to `epoll_ctl` and output of `epoll_wait` when monitoring a single file.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+	int epollfd;
+	int nfds;
+	int ret;
+
+	uk_pr_debug("Signalling children\n");
+
+	kill(0, GRACEFUL_SHUTDOWN_SIGNAL);
+
+	/* Make signalfd non-blocking */
+	ret = fcntl(sigfd, F_SETFL, fcntl(sigfd, F_GETFL) | O_NONBLOCK);
```
andreittr, line 104: This double `fcntl` call is unfortunate, but I see why you need it.
There is the alternative of
```c
preadv2(fd, (struct iovec){ .iov_base = &info, .iov_len = sizeof(info) }, 1, 0, RWF_NOWAIT);
```
that does a nonblocking read without mucking about with `fcntl`, but now I see that our nolibc doesn't provide `preadv2` nor `RWF_NOWAIT` :facepalm: .
Let's leave it like this for now and switch to preadv2 later. There are some more improvements to nolibc that this could go along with.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+	ret = epoll_ctl(epollfd, EPOLL_CTL_ADD, sigfd, &ev);
+	if (unlikely(ret == -1)) {
+		uk_pr_err("epoll_ctl failed (%d)\n", errno);
+		ret = errno;
+		goto err_close_epollfd;
+	}
+
+	timeout_ms = GRACEFUL_SHUTDOWN_TIMEOUT;
+
+	while (1) {
```
andreittr, line 131: I believe this loop as it is might oversleep until the deadline; AFAICT it breaks on (1) error, (2) timeout, and (3) trying to reap a child when none are present. (1) & (2) are obvious, but (3) I think can never happen: we reach the `ret == -ECHILD` check only after (a) a wakeup from `epoll_wait`, and (b) a successful `read` from `sigfd`. Thus, after reaping the last child successfully, the code goes back into `epoll_wait` that hangs until the deadline, as there are no more signals to wake up sigfd/epoll. We also process just 1 child per sigfd/epoll wakeup, but IIRC if multiple SIGCHLD signals are sent to pid1 but not processed, they get squashed together into one -- so we might miss children and oversleep.

Furthermore, `timeout_ms` is not updated between wakeups from `epoll_wait`, resulting in the countdown timer being reset on every reap; while this can only prolong shutdown for a finite amount of time, it's probably not intended behavior.

Also, I think the order of operations should be flipped; in nonblocking I/O it is more common to try-then-wait rather than the other way around. So something like
```c
deadline = now() + GRACEFUL_SHUTDOWN_TIMEOUT;
while (1) {
	/* Try to reap */
	ret = wait_nonblocking();
	if (ret == -ECHILD)
		break; /* Out of children, done */
	if (ret > 0)
		continue; /* Reaped, maybe more on the way */

loop_wait:
	/* ret == 0 => need to wait */
	timeout_ms = deadline - now();
	if (timeout_ms <= 0)
		break; /* We've passed deadline, no more waiting */
	nfds = epoll_wait(epollfd, &ev, 1, timeout_ms);
	if (unlikely(nfds == -1)) {
		ret = errno;
		break;
	}
	if (!nfds)
		break; /* Timeout */
	
	/* Read sigfd to clear events; use of output is optimization */
	bytes = read(sigfd, &info, sizeof(info));
	if (!(bytes == sizeof(info) && info.ssi_signo == SIGCHLD))
		goto loop_wait; /* Not a SIGCHLD, go back to wait */
}
```
with `now()` being a function that gives you a monotonic clock in ms (epoch irrelevant). The last line with `goto loop_wait` is a (over)optimization, we could probably remove that entire `if` and loop back to try a `wait_nonblocking()` regardless.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+		/* Wait for events or timeout */
+		nfds = epoll_wait(epollfd, events, 1, timeout_ms);
+		if (unlikely(nfds == -1)) {
+			uk_pr_err("epoll_wait failed (%d)\n", errno);
+			break;
```
andreittr, line 136: Are we purposely not returning `errno` here or is it an oversight?

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+				break;
+			}
+		} else if (info.ssi_signo == SIGTERM) {
+			uk_pr_info("Received SIGTERM. Initiating shutdown...\n");
+			break;
+		}
+	}
+
+#if CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN
+	graceful_shutdown(sigfd);
```
andreittr, line 231: This function returns an int, do we want to ignore it? If so, it'd probably be better as `void`.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+	return ret;
+}
+#endif /* CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN */
+
+int do_init(int argc, char *argv[])
+{
+	struct signalfd_siginfo info;
+	ssize_t bytes;
+	sigset_t mask;
+
```
andreittr: Is this blank line intentional?

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#define UK_DEBUG 1
```
andreittr: Likely leftover from dev

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+	sigfillset(&mask);
+	sigprocmask(SIG_BLOCK, &mask, NULL);
+
+	/* Prepare mask for signalfd */
+	sigemptyset(&mask);
+	sigaddset(&mask, SIGCHLD);
+	sigaddset(&mask, SIGTERM);
+
+	/* Create signalfd for SIGCHLD and SIGTERM */
+	sigfd = signalfd(-1, &mask, 0);
```
andreittr: We probably want this opened with `SFD_CLOEXEC` so we don't pollute the application process' fdtab with it.

lib/ukboot/init.c
```diff
@@ -0,0 +1,218 @@
...
+		if (bytes != sizeof(info)) {
+			uk_pr_err("Read from signalfd failed\n");
+			goto err_close_signalfd;
+		}
+
+		if (info.ssi_signo == SIGCHLD) {
+			/* Reap children. If the last child terminated,
+			 * initiate shutdown.
+			 */
+			ret = wait_nonblocking();
```
andreittr: Will this ever return `-ECHILD` if we only call it after receiving SIGCHLD? There's nobody else who could reap the child between the two calls :thinking: I believe this loop would only break on SIGTERM.

Also, IIRC multiple SIGCHLD signals are squashed on delivery and only handled once, thus doing a single wait might miss exited children.

andreittr (commented):
A 2nd pass, focusing on the /sbin/init logic

lib/ukboot/init.c
michpappas: Damn

lib/ukboot/init.c
michpappas: Very good point

lib/ukboot/init.c
michpappas: The problem here is that we can't return these errors as they would shadow the application's return value, if the application has returned by this point. Ideally we would need a way to differentiate between the application's return code and collateral failures such as this. I can see the following options:
1. Print errors (will end up in the syslog once #1527 is in)
2. Crash the kernel (the opposite extreme of the above)
3. Somehow return both values back to ukboot and have ukboot handle any error (goto 1)

I silently chose option 1 seeing such failures not much different than non-main threads did not returning gracefully, as in both cases they will be force-killed. Saying that, if calls like epoll fail there's something clearly wrong with the system, but as it's a shutdown operation, instead of hard crashing we might as well continue execution in the hope that forceful termination will successfully call the exit handlers.

lib/ukboot/init.c
michpappas: For the same reasons as https://github.com/unikraft/unikraft/pull/1633#discussion_r2055889747. Let's continue the discussion there.

lib/ukboot/init.c
michpappas: It's true, but I will rephrase the comment to make sure there's no doubt.

lib/ukboot/init.c
andreittr: In principle I feel the "right thing to do" if any part of the graceful shutdown fails is to (loudly) complain with an error and continue with force-killing, same as if a timeout had happened.
In that sense, I think the current behavior is the right one, and we should simply make `graceful_shutdown` return `void` to make it explicit that it must handle/eat up all errors internally.

lib/ukboot/Config.uk
michpappas: Without trying to engage into philosophical discussions, I'd say there's no such thing as a good timeout value in the first place, yet it's still useful to support indefinite timeouts instead of having the user to work around it by setting an extremely high value. I'll take option 1 and document it.

lib/ukboot/init.c
michpappas: Good point, I'll wrap it around a do / wait loop so that it reaps all children. And with that, it is in fact possible that it returns `-ECHILD`.

lib/ukboot/init.c
michpappas: > I believe this loop as it is might oversleep until the deadline; AFAICT it breaks on (1) error, (2) timeout, and (3) trying to reap a child when none are present. We ignore (1), (2) is obvious, but (3) I think can never happen: we reach the ret == -ECHILD check only after (a) a wakeup from epoll_wait, and (b) a successful read from sigfd. Thus, after reaping the last child successfully, the code goes back into epoll_wait that hangs until the deadline, as there are no more signals to wake up sigfd/epoll. We also process just 1 child per sigfd/epoll wakeup, but IIRC if multiple SIGCHLD signals are sent to pid1 but not processed, they get squashed together into one -- so we might miss children and oversleep.

Will apply same resolution as in https://github.com/unikraft/unikraft/pull/1633#discussion_r2056049236.

> Furthermore, timeout_ms is not updated between wakeups from epoll_wait, resulting in the countdown timer being reset on every reap; while this can only prolong shutdown for a finite amount of time, it's probably not intended behavior.

Yup, will fix 👍🏼 

> in nonblocking I/O it is more common to try-then-wait rather than the other way around

Actually what is the difference?

lib/ukboot/init.c
michpappas: Okay so in principle we agree, will update to return `void`. Saying that, this matter can take a lot more discussion depending on the payload, as upon failure you don't know if it's for the better or the worse letting execution continue -- one may argue that exit handlers on an incorrectly executing system may do more harm than good. But as that discussion falls into a different realm, let's keep it as is for now.

lib/ukboot/init.c
andreittr: > Actually what is the difference?

Besides saving a no-op roundtrip through the polling code in case there is work available, in this very particular case (single-thread level-polling (see [epoll(7)](https://www.man7.org/linux/man-pages/man7/epoll.7.html#DESCRIPTION) ) & signalfd is guaranteed created before any SIGCHLD comes in) I believe the two should be functionally identical. I think I'm just too used to see "try-before-wait" loops in most epoll code that uses `EPOLLET`, where they're ~~mandatory~~ strongly encouraged :smile: ; feel free to do it in any order.

lib/ukboot/init.c
andreittr: Exit handlers _should_ ( :crossed_fingers: ) work regardless if `CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN`; so in that sense if `graceful_shutdown` (1) cleans up on error paths, and (2) doesn't break kernel state, it doesn't matter to exit handlers if or how it did its job. (1) we ensure with this PR, and (2) should be guaranteed by all syscalls leaving kernel state sane on error (a reasonable ask IMO) -- we use only syscalls in `init.c` after all.

TL;DR: exit handlers shouldn't care; we can capture the return of `graceful_shutdown` only if/when we want to pass richer return statuses back to `ukboot`; `void` is good for now.

lib/ukboot/init.c
```diff
@@ -0,0 +1,241 @@
...
+	ev.data.fd = sigfd;
+	ret = epoll_ctl(epollfd, EPOLL_CTL_ADD, sigfd, &ev);
+	if (unlikely(ret == -1)) {
+		uk_pr_err("epoll_ctl failed (%d)\n", errno);
+		goto err_close_epollfd;
+	}
+
+	/* Reap terminated children */
+	deadline = now() + GRACEFUL_SHUTDOWN_TIMEOUT;
+	timeout  = deadline;
```
andreittr: This initial value for `timeout` is too high, should be just `GRACEFUL_SHUTDOWN_TIMEOUT`.

Also, if the user has purposely set a negative timeout, we shouldn't even bother computing `deadline` or calling `now()`.

lib/ukboot/init.c
```diff
@@ -0,0 +1,241 @@
...
+ * Notice: We don't return on error as we don't have a way to handle these
+ *         other than printing an error message (do_init() is expected to
+ *         return the application's code, that we don't want to shadow if
+ *         the application has already returned by this point).
+ */
+static void graceful_shutdown(int sigfd)
+{
+	struct epoll_event ev;
+	struct signalfd_siginfo info;
+	int deadline; /* msec */
```
andreittr: Should likely be `long` to match with `now()`'s return size.

lib/ukboot/init.c
```diff
@@ -0,0 +1,241 @@
...
+			application_status = WEXITSTATUS(wstatus);
+		else if (WIFSIGNALED(wstatus))
+			application_status = TERM_BY_SIGNAL_BIT | WTERMSIG(wstatus);
+	}
+
+	return pid;
+}
+
+/* Monotonic time in msec */
+static inline unsigned long now(void)
```
andreittr: AFAICT this is not used unless `CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN`, so it can be moved within the `#if` block below. Also, I think we can make it signed `long` to propagate the signedness of `time_t`.

lib/ukboot/init.c
```diff
@@ -0,0 +1,241 @@
...
+{
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+
+	return (unsigned long)ts.tv_sec * 1000 + (unsigned long)ts.tv_nsec / 1000000;
```
andreittr: If you don't mind pulling in [`uktimeconv`](https://github.com/unikraft/unikraft/blob/staging/lib/uktimeconv/include/uk/timeutil.h) you can write this as
```c
return ukarch_time_nsec_to_msec(uk_time_spec_to_nsec(ts));
```
and save on manual maths. Also, I don't think it's necessary to cast either field to unsigned (see related comment on function's return type).

lib/ukboot/init.c
```diff
@@ -0,0 +1,241 @@
...
+
+			UK_ASSERT(ret == 0 || ret == -ECHILD);
+
+			if (ret == -ECHILD) {
+				uk_pr_debug("All children exited gracefully\n");
+				break;
+			}
+		}
+
+		timeout = deadline - now();
```
andreittr: We should explicitly check for negative values here and break. If a `timeout < 0` somehow reaches `epoll_wait` it will wait forever instead of figuring out that the deadline has passed.

Then again, if a user has specified a negative timeout to begin with, we shouldn't even bother recalculating `timeout`. So I propose:
```c
if (timeout >= 0) {
    timeout = deadline - now();
    if (timeout < 0)
        break;
}
```

lib/ukboot/init.c
andreittr: Also, I said a monotonic clock is the right choice for the timeout, but on 2nd thought I'm not so sure whether monotonic or realtime deadline is more appropriate:
- monotonic is better if we want to give apps X ms of _actual running time_ to gracefully shutdown, irrespective of system clock changes or stop-the-world events like suspend (which monotonic clocks don't count)
- realtime is better if we want the system to force-shutdown ASAP at a specific wall-clock time from init getting SIGTERM

The answer depends on policy, and whatever policy we pick we should do so explicitly.

michpappas:
Rebased from `staging`, no functional changes.

lib/ukboot/init.c
michpappas: I'd rather minimize the number of calls to the uk API, so far we only have one we can't avoid, `uk_posix_process_run()`.  I'll annotate numbers with `UL` and get rid of the casts.

michpappas:
@andreittr thanks a lot for the elaborate review. Latest update addresses the remaining of your comments.

lib/ukboot/init.c
andreittr: After reading the latest changes in their entirety, I realise my proposal is wrong for `timeout == 0` (i.e., `epoll_wait` never blocks), whether specified by the user or if by sheer luck `deadline - now()` works out exactly 0 in a previous iteration.
In both cases we shouldn't bother calling `now()`, and we shouldn't reference `deadline` in the former as it is uninitialized. What we should do is immediately break the loop as we've clearly run out of time.
In that light, what I can think of is:
```c
if (timeout >= 0) {
    if (!timeout) /* Out of time, don't bother continuing */
        break;
    timeout = deadline - now();
    if (timeout < 0) /* We still want 1 try of wait if timeout is exactly 0 */
        break;
}
```

michpappas: Good catch. I think the last iteration is not necessary though as we already reached deadline. Will push a slightly modified version that handles the case where the user passed zero:
```diff
diff --git a/lib/ukboot/init.c b/lib/ukboot/init.c
index 5eb79e8c0..c85c331cf 100644
--- a/lib/ukboot/init.c
+++ b/lib/ukboot/init.c
@@ -158,8 +158,10 @@ static void graceful_shutdown(int sigfd)
                }
 
                if (timeout >= 0) {
+                       if (!timeout) /* User set to zero */
+                               break;
                        timeout = deadline - now();
-                       if (timeout < 0)
+                       if (timeout <= 0)
                                break;
                }
        }
```

michpappas:
In addition to addressing the last review comment, the latest update contains:
* A fix on a conflict resolution in the last rebase of #1630 
* A fix on `shutdown.c` that compiles `force_kill()` only if `CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS` is selected.

lib/ukboot/Config.uk
```diff
@@ -196,6 +196,45 @@ config LIBUKBOOT_SHUTDOWNREQ_HANDLER
 	  if it is requested by a driver on the shutdown_req event queue
 	  (e.g., ACPI power button event).
 
+config LIBUKBOOT_INIT
+	bool "Provide /sbin/init logic"
+	depends on LIBPOSIX_PROCESS_MULTIPROCESS
+	select LIBPOSIX_PROCESS_SIGNAL
+	select LIBPOSIX_PROCESS_SIGNALFD
```
andreittr, line 204: We should probably `select LIBPOSIX_FDTAB` too, else we are not guaranteed fd-related syscalls/libc funcs.
(It is a conscious design decision of the unikraft file stack to separate "file" and "fd" functionality; thus signalfd by itself will only provide unikraft-internal signal file, the userspace-facing signalfd sys/libcalls are exposed only if fdtab is enabled)

lib/ukboot/init.c
```diff
@@ -0,0 +1,248 @@
...
+	ret = epoll_ctl(epollfd, EPOLL_CTL_ADD, sigfd, &ev);
+	if (unlikely(ret == -1)) {
+		uk_pr_err("epoll_ctl failed (%d)\n", errno);
+		goto err_close_epollfd;
+	}
+
+	/* Reap terminated children */
+	if (GRACEFUL_SHUTDOWN_TIMEOUT > 0)
+		deadline = now() + GRACEFUL_SHUTDOWN_TIMEOUT;
+	timeout  = GRACEFUL_SHUTDOWN_TIMEOUT;
```
andreittr: Nit: the extra space after `timeout` is no longer warranted for aligning with previous line, you can remove.

andreittr (commented):
A hopefully final pass, it looks really good, thanks for the changes! Only minor nitpicks remaining and it's ready for merge IMO.

andreittr (approved):
Looks great, thanks!

andreittr:

PR #1634, opened 2025-04-15
plat/xen: Add initrd support for arm64

andreistan26:
Add support for initrd on xen/arm64

### Base target

### DIFF
--- a/plat/xen/arm/setup64.c
+++ b/plat/xen/arm/setup64.c
@@ -208,6 +208,54 @@ int uk_intctlr_plat_probe(void *arg)
 	return 0;
 }
 
+/* Taken from plat/common/bootinfo_fdt.c and adapted for Xen
+ * because we need to translate the initrd address from the
+ * device tree to the virtual address.
+ */
+#define initrd_addr(val, len)                          \
+	(len == 4 ? fdt32_to_cpu(val) : fdt64_to_cpu(val))
+
+static int _get_ramdisk(struct ukplat_bootinfo *bi, void *fdtp)
+{
+	struct ukplat_memregion_desc mrd = {0};
+	const __u64 *fdt_initrd_start;
+	const __u64 *fdt_initrd_end;
+	int start_len, end_len;
+	__u64 initrd_base, initrd_end;
+	int nchosen;
+
+	nchosen = fdt_path_offset(fdtp, "/chosen");
+	if (unlikely(nchosen < 0))
+		return 0;
+
+	fdt_initrd_start = fdt_getprop(fdtp, nchosen, "linux,initrd-start",
+				       &start_len);
+	if (unlikely(!fdt_initrd_start || start_len <= 0))
+		return 0;
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return 0;
+
+	initrd_base = initrd_addr(fdt_initrd_start[0], start_len);
+	initrd_end = initrd_addr(fdt_initrd_end[0], end_len);
+
+	mrd.vbase = (__vaddr_t)to_virt(PAGE_ALIGN_DOWN(initrd_base));
+	mrd.pbase = (__paddr_t)mrd.vbase;
+	mrd.pg_off = initrd_base - mrd.pbase;
+	mrd.len = initrd_end - initrd_base;
+	mrd.pg_count = PAGE_COUNT(mrd.pg_off + mrd.len);
+	mrd.type = UKPLAT_MEMRT_INITRD;
+	mrd.flags = UKPLAT_MEMRF_READ;
+
+#if CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "initrd", sizeof(mrd.name) - 1);
+#endif
+
+	return ukplat_memregion_list_insert(&bi->mrds, &mrd);
+}
+
 static int _init_mem(struct ukplat_bootinfo *const bi, paddr_t physical_offset)
 {
 	int rc;
@@ -279,6 +327,10 @@ static int _init_mem(struct ukplat_bootinfo *const bi, paddr_t physical_offset)
 #endif
 	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
 	if (rc < 0)
+		return rc;
+
+	rc = _get_ramdisk(bi, HYPERVISOR_dtb);
+	if (unlikely(rc < 0))
 		return rc;
 
 	ukplat_memregion_list_coalesce(&bi->mrds);


plat/xen/arm/setup64.c
```diff
@@ -281,6 +327,11 @@ static int _init_mem(struct ukplat_bootinfo *const bi, paddr_t physical_offset)
 	if (rc < 0)
 		return rc;
 
+	rc = _get_ramdisk(bi);
+	if (rc < 0)
+		return rc;
+
+coalesce:
```
oleksiimoisieiev: I see you have added tag but no goto. Did I miss something?

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,52 @@ int uk_intctlr_plat_probe(void *arg)
...
+		return 1;
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return 1;
+
+	initrd_base = fdt64_to_cpu(fdt_initrd_start[0]);
+	initrd_end = fdt64_to_cpu(fdt_initrd_end[0]);
+	mrd = (struct ukplat_memregion_desc){0};
```
oleksiimoisieiev: Maybe it can be done on the definition stage?
Smth like: struct ukplat_memregion_desc mrd = {0};

plat/xen/arm/setup64.c
andreistan26: Thanks, moved things around and forgot that there

oleksiimoisieiev:
LGTM
`Reviewed-by: Oleksii Moisieiev <oleksii_moisieiev@epam.com>`

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,51 @@ int uk_intctlr_plat_probe(void *arg)
...
+				       &start_len);
+	if (unlikely(!fdt_initrd_start || start_len <= 0))
+		return 1;
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return 1;
+
+	initrd_base = fdt64_to_cpu(fdt_initrd_start[0]);
```
mogasergiu: Is there a reason you don't use the `initrd_addr` like in the [code](https://github.com/unikraft/unikraft/blob/staging/plat/common/bootinfo_fdt.c#L116) you duplicated this function from? The comment above seems to have a good reason :thinking: Without that I don't see what you're even initializing `start_len` for.

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,51 @@ int uk_intctlr_plat_probe(void *arg)
...
+
+	fdt_initrd_start = fdt_getprop(fdtp, nchosen, "linux,initrd-start",
+				       &start_len);
+	if (unlikely(!fdt_initrd_start || start_len <= 0))
+		return 1;
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return 1;
```
mogasergiu: Is there a reason for `return 1`? Why not `0`? Personally, I would say it feels more natural than a random non-zero positive value. :thinking: 
Maybe a comment also helps saying why we just return a non-negative value cause these are indicatives that the `initrd` properties are either malformed or not present.

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,51 @@ int uk_intctlr_plat_probe(void *arg)
...
+
+	initrd_base = fdt64_to_cpu(fdt_initrd_start[0]);
+	initrd_end = fdt64_to_cpu(fdt_initrd_end[0]);
+
+	mrd.vbase = to_virt(PAGE_ALIGN_DOWN(initrd_base)),
+	mrd.pbase = mrd.vbase,
+	mrd.len = initrd_end - initrd_base,
+	mrd.pg_off = initrd_base - PAGE_ALIGN_DOWN(initrd_base),
+	mrd.pg_count = PAGE_COUNT(mrd.len + mrd.pg_off),
```
mogasergiu: Is there a reason why you inverted the operands from the original code (same with removal of `initrd_end` above)? Maybe if we kept things as similar as possible to the original it will be easier to spot differences. I understand that you needed to duplicate it because of `to_virt` and needing to return a value to make it similar to the rest of the xen code.

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,51 @@ int uk_intctlr_plat_probe(void *arg)
...
+	mrd.type = UKPLAT_MEMRT_INITRD,
+	mrd.flags = UKPLAT_MEMRF_READ,
+
+#if CONFIG_UKPLAT_MEMRNAME
+	memcpy(mrd.name, "initrd", sizeof(mrd.name) - 1);
+#endif
+	rc = ukplat_memregion_list_insert(&bi->mrds, &mrd);
```
mogasergiu: Do you really need the `rc` variable?

plat/xen/arm/setup64.c
```diff
@@ -281,6 +326,10 @@ static int _init_mem(struct ukplat_bootinfo *const bi, paddr_t physical_offset)
 	if (rc < 0)
 		return rc;
 
+	rc = _get_ramdisk(bi);
+	if (rc < 0)
```
mogasergiu: `unlikely(rc < 0)`?

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,51 @@ int uk_intctlr_plat_probe(void *arg)
 	return 0;
 }
 
+static int _get_ramdisk(struct ukplat_bootinfo *bi)
+{
+	int rc;
+	struct ukplat_memregion_desc mrd = {0};
+	const __u64 *fdt_initrd_start;
+	const __u64 *fdt_initrd_end;
+	void *fdtp = HYPERVISOR_dtb;
```
mogasergiu: I would pass this as an argument like the original function [does](https://github.com/unikraft/unikraft/blob/staging/plat/common/bootinfo_fdt.c#L116). Just to keep it more similar.

mogasergiu (changes requested):
Looks good, just a couple of questions :muscle:, then will approve.

plat/xen/arm/setup64.c
andreistan26: Xen will always emit a 64 bit address but it makes sense to have it the same way.

plat/xen/arm/setup64.c
andreistan26: Agree, will keep it to `0`.

plat/xen/arm/setup64.c
andreistan26: Changed it back

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,50 @@ int uk_intctlr_plat_probe(void *arg)
...
+
+	fdt_initrd_end = fdt_getprop(fdtp, nchosen, "linux,initrd-end",
+				     &end_len);
+	if (unlikely(!fdt_initrd_end || end_len <= 0))
+		return 0;
+
+	initrd_base = initrd_addr(fdt_initrd_start[0], start_len);
+	initrd_end = initrd_addr(fdt_initrd_end[0], end_len);
+
+	mrd.pbase = (__paddr_t) to_virt(PAGE_ALIGN_DOWN(initrd_base));
```
mogasergiu: Are you sure it is `pbase` that needs the `to_virt`? I understand pbase is going to be equal to vbase anyway but I find it confusing.

Additionally, nitpick, while still on this line, I get this checkpatch `CHECK`
```
CHECK:SPACING: No space is necessary after a cast                     
#51: FILE: plat/xen/arm/setup64.c:240:                                
+       mrd.pbase = (__paddr_t) to_virt(PAGE_ALIGN_DOWN(initrd_base));
```
Up to you if you want to fix it, but since the rest of the code also respects this, I think it would be a good idea to fix it.

plat/xen/arm/setup64.c
```diff
@@ -208,6 +208,50 @@ int uk_intctlr_plat_probe(void *arg)
 	return 0;
 }
 
+#define initrd_addr(val, len)                          \
```
mogasergiu, line 215: Since you don't take the comment for `initrd_addr` macro as well, which makes sense, I would try to at least indicate through a comment to others that may bump into this code where this function was initially taken [from](https://github.com/unikraft/unikraft/blob/staging/plat/common/bootinfo_fdt.c#L116) and the reason for which it was a good idea at this time to duplicate, maybe something among the lines of `Taken from platform common bootinfo fdt initrd code and adapted for Xen because..` as other comments in `_init_mem` have also indicated why there is a need to deviate from how we typically do in other places.

I can see that you also indicated in the commit comment now after the last push, let's also give the reason there as well.

plat/xen/arm/setup64.c
mogasergiu: Yes this makes sense, my confusion stemmed from the unnecessary initialization of `start_len`/`end_len` that you were doing despite not using them. I guess it was just remnants from the original.

mogasergiu (changes requested):

plat/xen/arm/setup64.c
andreistan26: Makes sense, thanks!

mogasergiu (approved):
LGTM. One more potential nitpick, I am feeling unsure about direct mentions to `plat/common/bootinfo_fdt.c` in both comments and the commit message, since that is a location very likely to change in the near future. I am thinking maybe we should refer to it simply as "platform common bootinfo fdt code" (or "generic" instead of "platform common" ?) instead? But feel free to do as your intuition suggests as well.

razvand (approved):

sculamacaz (commented):
Reviewed-by: Luca Mazilescu luca.mazilescu@protonmail.com

PR #1636, opened 2025-04-29
lib/vfscore: Fix write const path in automount

andreittr:
### Description of changes

Previously vfscore_mount_volume would pass a const path to vfscore_ukopt_mkmp, which expects a mutable path, and indeed does modify it in-place during execution. This is wrong and rightfully triggers a compiler warning.
This change fixes the issue, by passing a temporary copy to mkmp.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/vfscore/automount.c
+++ b/lib/vfscore/automount.c
@@ -568,18 +568,19 @@ static int vfscore_extract_volume(const struct vfscore_volume *vv)
 #endif /* CONFIG_LIBUKCPIO */
 
 /* Handle `mkmp` Unikraft Mount Option */
-static int vfscore_ukopt_mkmp(char *path)
+static int vfscore_ukopt_mkmp(const char *path_arg)
 {
+	char path[strlen(path_arg) + 1];
 	char *pos, *prev_pos;
 	int rc;
 
-	UK_ASSERT(path);
-	UK_ASSERT(path[0] == '/');
+	UK_ASSERT(path_arg[0] == '/');
 
-	if (path[1] == '\0') {
+	if (path_arg[1] == '\0') {
 		uk_pr_debug(" mkmp: Called on '/', nothing to pre-create\n");
 		return 0;
 	}
+	strcpy(path, path_arg);
 
 	uk_pr_debug(" mkmp: Ensure mount path \"%s\" exists\n", path);
 	pos = path;


lib/vfscore/automount.c
```diff
@@ -687,7 +687,10 @@ static inline int vfscore_mount_volume(const struct vfscore_volume *vv)
 		    vv->ukopts);
 
 	if (vv->ukopts & LIBVFSCORE_UKOPT_MKMP) {
-		rc = vfscore_ukopt_mkmp(path);
+		char tmp[strlen(path) + 1];
```
michpappas: Not a blocker, but FYI: Similarly to Linux, our convention is not to use scope declarations, especially given that we don't yet have consensus on the standard the code should comply to.

lib/vfscore/automount.c
```diff
@@ -687,7 +687,10 @@ static inline int vfscore_mount_volume(const struct vfscore_volume *vv)
...
+
+		strcpy(tmp, path);
+		rc = vfscore_ukopt_mkmp(tmp);
```
michpappas: Looks like `vfscore_ukopt_mkmp()` modifies the temp path, while later we pass `path` to `mount()`. Is it safe to do so?

michpappas:
Also, as checkpatch suggests, the commit title should have a single scope component, i.e. `lib/vfscore: Fix write const path in automount`.

lib/vfscore/automount.c
andreittr: Yes; the modifications mkmp does to the path received as arg are undone by the time it exits. But it's still _writing_ to that memory, so a `const char *` is inappropriate.

lib/vfscore/automount.c
andreittr: > our convention is not to use scope declarations

That is news to me, as it's not mentioned in the [coding conventions](https://unikraft.org/docs/contributing/coding-conventions#variable-definitions). `tmp`  is indeed declared at the start of a block and separated by whitespace from code. But fine, I can move `tmp` and the `strcpy` to `mkmp`, doesn't make much of a difference for this case. For the record, I _absolutely_ disagree that "scope declarations are bad and we should avoid them", but that debate is out of scope.

> we don't yet have consensus on the standard the code should comply to

The core codebase uses C99 features extensively (whether by design or not, this is where we stand atm); Unikraft supporting C89/90 (which cannot handle intermingled declarations & code, and a major driver of most of Linux's very outdated codestyle) is IMHO impossible. And even C89 supports declarations at the beginning of any block, not just function scope. Thus I believe any concern about functionality/correctness is unfounded; this is only a question of style.

andreittr:
> Also, as checkpatch suggests, the commit title should have a single scope component, i.e. lib/vfscore: Fix write const path in automount.

That had been my initial title (w/ "write _of_"), but it overflows git's allowed/recommented title limit. Will fix to remove "of" & it fits.

lib/vfscore/automount.c
michpappas: > That is news to me, as it's not mentioned in the [coding conventions](https://unikraft.org/docs/contributing/coding-conventions#variable-definitions). tmp is indeed declared at the start of a block and separated by whitespace from code. But fine, I can move tmp and the strcpy to mkmp, doesn't make much of a difference for this case. For the record, I absolutely disagree that "scope declarations are bad and we should avoid them", but that debate is out of scope.

Indeed neither in our or in the linux's conventions it's not explicit. And for the record in principle I agree with you as I would also think they can be useful at least in some cases, but we cannot apply the conventions selectively depending to our preference. So as you suggest, let's have that discussion, but not in the scope of this PR 👍🏼 

Similarly for the topic of C99 (agreed, let's do it, not in this scope).

lib/vfscore/automount.c
michpappas: Thanks for confirming

michpappas:

michpappas:

PR #1639, opened 2025-04-30
lib/posix-process: Make rt_sigtimedwait stub block

michpappas:
### Prerequisite checklist

### Base target

### Description of changes

According to rt_sigtimedwait(2) the syscall should return a signal number on success (> 0) or -1 on failure, so the zero 
value currently returned by the stubbed implementation can be misinterpreted by applications that check against failure. Update the stub to return implementation to block and never return a signal.

### DIFF
--- a/lib/posix-process/signal/rt_sigtimedwait.c
+++ b/lib/posix-process/signal/rt_sigtimedwait.c
@@ -46,11 +46,15 @@ UK_LLSYSCALL_R_DEFINE(int, rt_sigtimedwait,
 	if ((sig = pprocess_signal_next_pending_t(pthread)))
 		goto out;
 
-	if (timeout)
+	if (timeout) {
+		if (unlikely(!(uk_time_spec_canonical(timeout) &&
+			       uk_time_spec_positive(timeout))))
+			return -EINVAL;
 		uk_semaphore_down_to(&pthread->signal->pending_semaphore,
 				     uk_time_spec_to_nsec(timeout));
-	else
+	} else {
 		uk_semaphore_down(&pthread->signal->pending_semaphore);
+	}
 
 	if ((sig = pprocess_signal_next_pending_t(pthread)))
 		goto out;
--- a/lib/posix-process/signal/rt_sigtimedwait.c
+++ b/lib/posix-process/signal/rt_sigtimedwait.c
@@ -12,6 +12,11 @@
 #include <uk/syscall.h>
 #include <uk/timeutil.h>
 
+#if !CONFIG_LIBPOSIX_PROCESS_SIGNAL
+#include <uk/plat/time.h>
+#include <uk/sched.h>
+#endif /* !CONFIG_LIBPOSIX_PROCESS_SIGNAL */
+
 #include "process.h"
 #include "signal.h"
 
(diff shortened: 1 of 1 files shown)


lib/posix-process/signal/rt_sigtimedwait.c
```diff
@@ -64,9 +69,28 @@ UK_LLSYSCALL_R_DEFINE(int, rt_sigtimedwait,
...
 {
-	UK_WARN_STUBBED();
+	struct uk_thread *current = uk_thread_current();
+	__nsec timeout_ns;
+	int rc;
+
+	if (timeout) {
+		timeout_ns = ukplat_monotonic_clock() +
+			ukarch_time_sec_to_nsec(timeout->tv_sec) +
+			timeout->tv_nsec;
```
andreittr, line 86: You might consider validating `timeout` to be both [positive and in canonical form](https://github.com/unikraft/unikraft/blob/staging/lib/uktimeconv/include/uk/timeutil.h#L46) before using it to compute the deadline. I can confirm testing the behavior of `sigtimedwait` and it errors out with `-EINVAL` on a bad timeout value.
Userspace can be arbitrarily silly, and it's good to be compatible.

Similar argument can apply to the implementation that does use signals, but that's not strictly in scope :shrug: .

andreittr (commented):
One minor comment; plus typo in commit title: "rt_sitgimedwait" -> "rt_sigtimedwait".
(I've already fixed the PR title)

lib/posix-process/signal/rt_sigtimedwait.c
michpappas: Good point. I will also add a commit to fix the actual implementation.

lib/posix-process/signal/rt_sigtimedwait.c
```diff
@@ -64,9 +69,30 @@ UK_LLSYSCALL_R_DEFINE(int, rt_sigtimedwait,
...
+	__nsec timeout_ns;
+	int rc;
+
+	if (timeout) {
+		if (unlikely(!uk_time_spec_canonical(timeout))
+			return -EINVAL;
+		timeout_ns = ukplat_monotonic_clock() +
+			ukarch_time_sec_to_nsec(timeout->tv_sec) +
+			timeout->tv_nsec;
+		uk_thread_block_until(current, (__snsec) timeout_ns);
```
mogasergiu: nitpick but no space after cast (i know that's just how it was initially when you picked it up)? it;s just a checkpatch `CHECK` but as you know (and as you also do) we tend to respect that :sweat_smile:

mogasergiu (approved):
just a tiny nit, feel free to disregard if you want

lib/posix-process/signal/rt_sigtimedwait.c
```diff
@@ -42,6 +47,8 @@ UK_LLSYSCALL_R_DEFINE(int, rt_sigtimedwait,
 		goto out;
 
 	if (timeout)
+		if (unlikely(!uk_time_spec_canonical(timeout))
```
mogasergiu: Wait I think you also meant to add { } to this if/else now

lib/posix-process/signal/rt_sigtimedwait.c
michpappas: Oh yes it's from the original author, thanks for catching. I wonder why checkpatch didn't catch....

lib/posix-process/signal/rt_sigtimedwait.c
michpappas: Oooops

lib/posix-process/signal/rt_sigtimedwait.c
```diff
@@ -64,9 +72,30 @@ UK_LLSYSCALL_R_DEFINE(int, rt_sigtimedwait,
 		      const struct timespec *, timeout,
 		      size_t, sigsetsize)
 {
-	UK_WARN_STUBBED();
+	struct uk_thread *current = uk_thread_current();
+	__nsec timeout_ns;
+	int rc;
+
+	if (timeout) {
+		if (unlikely(!uk_time_spec_canonical(timeout)))
```
andreittr: We should check both canonicity _and_ a positive value (the two are independent properties, as a canonical timespec can indeed represent negative time)
```c
if (unlikely(!(uk_time_spec_canonical(timeout) && uk_time_spec_positive(timeout))))
```

michpappas: Right, I got mislead by the check for `tv->tv_nsec` thinking that this would suffice. Would be useful if that was documented. Also a bit strange that negative time can only be expressed by the `sec` part but not the `nsec`, that would be also useful to be aware of.

andreittr: Because the two are independent properties. A `struct timespec`, as [documented](https://www.man7.org/linux/man-pages/man3/timespec.3type.html) can only have `tv_nsec` in the range `[0, 999'999'999]` regardless of type or signedness of `tv_nsec`, and regardless of sign of `tv_sec`. We call timespec structs that follow this rule "canonical". AFAICT all Linux/libc API outputs canonical timespecs and errors out if fed non-canonical values.

Orthogonal to canonicity is the actual value of the timespec (which depending on context is interpreted either as a relative "time offset from now" or absolute "offset from epoch") that can be either positive or negative. In the case of a timeout, only positive timespec values make sense, but for timestamps negative values are perfectly reasonable (e.g., `(struct timespec){ -1, 500000000 }` represents half a second before midnight Jan 1 1970).

andreittr (approved):
Looks good, thanks!

mogasergiu (approved):

PR #1641, opened 2025-05-02
lib/posix-process/arch/arm64: Save/restore lr/fp on clone asm stub

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Commit 87fa095ae8bf ("lib/syscall_shim/arch/arm64: Store caller's link register")
updated the way we save the link register to properly emulate a                  
system call trap, so that the actual link register substitued our                
supposed ELR_EL1 and the link register from before substitutes the               
actual link register of the function that trapped. In order to achieve           
the latter it does a level 1 stack unwinding to extract the link                 
register saved at offset 8 from the frame pointer.                               
                                                                                 
However, the commit forgot to also update the way we restore the link            
register as now it wrongly restores the link register saved from the             
unwinding instead of the one from where the syscall was called.                  
                                                                                 
Fix this by properly restoring the link register from where we                   
actually save it now after said commit.                                          

Additionally, since we are invoking `clone` as a regular function call instead
of a system call, there is nobody to remember the asm stub entry
link register for us once we branch into the syscall. Therefore 
save and restore it.

### DIFF
--- a/lib/posix-process/arch/arm64/clone.S
+++ b/lib/posix-process/arch/arm64/clone.S
@@ -20,10 +20,16 @@
 #if UK_LIBC_SYSCALLS
 .global clone
 clone:
+	/* We must also save the link/frame register because no one else will
+	 * do it for us since we are invoking a function call not a system call
+	 */
+	stp	x29, x30, [sp, #-32]!
+	mov	x29, sp
+
 	/* Save fn and arg in callee-saved registers, but first make sure
 	 * we can also restore them before returning in the parent.
 	 */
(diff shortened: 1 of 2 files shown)


mogasergiu:
Related musl PR: https://github.com/unikraft/lib-musl/pull/89

michpappas (approved):

sculamacaz (approved):
Reviewed-by: Luca Mazilescu luca.mazilescu@protonmail.com

michpappas:

PR #1644, opened 2025-05-06
lib/posix-process: Add README.md

michpappas:
Add README.md

### Prerequisite checklist

### Base target

### DIFF
--- a/lib/posix-process/README.md
+++ b/lib/posix-process/README.md
@@ -1,0 +1,118 @@
+# posix-process
+
+libposix-process provides a POSIX-compliant process implementation allowing Unikraft to run unmodified applications written for *NIX systems.
+It provides syscalls for task identification and lifetime management, as well as support for signals.
+
+To accommodate different application requirements while preserving Unikraft's minimal and modular design, libposix-process offers three feature levels:
+
+### Single thread
+Bare minimum process-like behavior for single process applications that execute on a single thread.
+Provides stubbed implementations of task identification facilities required by musl.
+Although appearing as single-threaded, the process does not necessarily need to be backed by an underlying `uk_thread`, as it does not rely on thread state.
+That is, this feature level is independent of [libuksched](https://github.com/unikraft/unikraft/tree/staging/lib/uksched).
+Configuration:
+  * Select `CONFIG_LIBPOSIX_PROCESS` without additional options.
+
+### Multithreading
+Single process with multithreading.
+Provides true task identification and clone() support for threads.
+Configuration:
+  * Select `CONFIG_LIBPOSIX_PROCESS` and `CONFIG_LIBPOSIOX_PROCESS_MULTITHREADING`.
+  * Optional: Select `CONFIG_LIBPOSIX_PROCESS_EXECVE` to enable `execve()` (requires VFS).
+  * Optional: Select `CONFIG_LIBPOSIX_PROCESS_SIGNAL` to enable POSIX signals.
+
+### Multiprocess:
+Full-featured multiprocess support.
+Configuration:
+  * Select `CONFIG_LIBPOSIX_PROCESS` and `CONFIG_LIBPOSIOX_PROCESS_MULTIPROCESS`. Requires VFS.
+  * Optional: Select `CONFIG_LIBPOSIX_PROCESS_SIGNAL` to enable POSIX signals.
+
+A summary of the syscalls provided by libposix-process and their behavior on each configuration is shown below:
+
+|                            | Single Thread     | Multithreading                   | Multiprocess |
+| -------------------------- | ----------------- | -------------------------------- | ------------ |
+| *gettid()*                 | Yes (stubbed)     | Yes                              | Yes          |
+| *getpid()*                 | Yes (stubbed)     | Yes                              | Yes          |
+| *getppid()*                | Yes (stubbed)     | Yes                              | Yes          |
+| *clone()*                  | No (`ENOSYS`)     | Yes (`ENOTSUP` on `CLONE_VFORK`) | Yes          |
+| *vfork()*                  | No (`ENOSYS`)     | No (`ENOSYS`)                    | Yes          |
+| *execve()*                 | No (`ENOSYS`)     | Conditional [^2]                 | Yes          |
+| *_exit()* / *exit_group()* | Conditional [^1]  | Yes                              | Yes          |
+| *wait4()* / *waitid()*     | No (`ECHILD`)     | No (`ECHILD`)                    | Yes          |
+
+[^1]: When `CONFIG_LIBUKBOOT_MAINTHREAD` is selected, the application is executed on a dedicated `uk_thread`.
+In that case, single process single thread configurations can emulate `_exit()` and `exit_group()`, as any of these syscalls will result in cleanly terminating the underlying `uk_thread`.
+If `CONFIG_LIBUKBOOT_MAINTHREAD` is not selected, the application executes on Unikraft's boot thread, which cannot be terminated.
+Similarly if `CONFIG_LIBUKSCHED` is not selected, the application executes without being backed by an underlying `uk_thread`.
+For these reasons `_exit()` / `exit_group()` cannot be meaningfully implemented under such configurations.
+Since these syscalls never return, and thus cannot signal failure to their caller, calling them is an unrecoverable runtime error, on which Unikraft will panic with an appropriate error message.
+
+[^2]: Calling `execve()` without multiprocessing support is possible, and results into the calling process' context being replaced with that of the executed program.
+Because `execve()` is passed a filesystem path, `CONFIG_LIBPOSIX_PROCESS_EXECVE` depends on VFS.
+To avoid unnecessary bloat on appliations that may not use `execve()`, `CONFIG_LIBPOSIX_PROCESS_EXECVE` is provided as an optional selection in multithreaded configurations.
+
+## Requirements of execve()
+
+As stated earlier `execve()` depends on VFS.
+Loading and execution is handled by [libukbinfmt](https://github.com/unikraft/unikraft/tree/staging/lib/ukbinfmt), which requires a binfmt handler for the corresponding binary format.
+The binfmt handler for ELF is provided by [app-elfloader](https://github.com/unikraft/app-elfloader).
+Additionally, since Unikraft executes on a single address space, it is required that executables passed to `execve()` must be compiled as PIE.
+
+## Notes on multiprocess
+
+Being a single-address-space operating system, instead of the traditional `fork()` that creates a new address space for the child, Unikraft implements multiprocess with `vfork()` only.
+That includes the `vfork(2)` syscall, the equivalent of calling `clone()` with flags set to `CLONE_VM | CLONE_VFORK | SIGCHLD`.
+When calling `vfork()`, the child shares the same address space as the parent, and the parent's execution is suspended until the child either calls `execve()`, or terminates by calling `_exit()` / receiving a terminating signal.
+This introduces limitations on what the child can do until it calls `execve()` (or terminates), as it must ensure that it doesn't leave the parent's memory in an incosistent state.
+Specifically, according to [vfork(2)](https://man7.org/linux/man-pages/man2/vfork.2.html), the only operations the child can do are: (1) update the memory location that stores return value of `vfork()`, (2) call `execve()`, (3) call `_exit()`.
+In any other case the resulting behavior is undefined.
+Fortunately it is common that appliations spawn new processes by calling `fork()` immediately followed by an `execve()`, so running a multiprocess application on Unikraft that does not natively support `vfork()` should be possible with minimal changes.
+
+> [!TIP]
+> Applications can use the `posix_spawn()` libc function that spawns a process using `vfork()` in a safe way.
+
+Another important aspect of multiprocess systems is the existence of an init process.
+Unikraft implements `/sbin/init`-like logic as part of [libukboot](https://github.com/unikraft/unikraft/tree/staging/lib/ukboot), and is enabled by selecting `CONFIG_LIBUKBOOT_INIT`.
+The implementation is minimal; essentially it spawns a new process for the application, fosters orphaned children, reaps zombies, and coordinates system shutdown.
+In most cases, an init process is necessary on multiprocess configurations, yet `CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS` does not enforce the selection of `CONFIG_LIBUKBOOT_INIT` by default, to give applications the possibility to override the default implementation, or operate without an init.
+The latter is possible if the application does not exit early leaving daemonized processes behind, and does not require graceful shutdown (described below).
+In conclusion, multiprocess applications should therefore either (1) select `CONFIG_LIBUKBOOT_INIT` (2) provide their own implementation of init, or (3) make sure that they don't need one.
+
+## System Shutdown
+
+From Unikraft 0.16.0, libukboot supports asynchronously triggered shutdown.
+At the most fundamental level, when the system receives a shutdown request, the application must be terminated before proceeding to terminate system-wide components like the network and filesystem.
+When libposix-process is selected, all processes must be terminated and their resources must be released.
+That requires additional handling at the level of libposix-process.
+
+The way libposix-process implements shutdown at the process level depends on the system configuration.
+If signals are enabled, the *application process*[^4] is sent `SIGTERM` and shutdown blocks until the application process terminates.
+Once the application process terminates, any remaining processes are terminated forcefully.
+If signals are not enabled, all processes are terminated forcefully.
+In both cases the result is the same: all process resources are directly released.
+
+In multiprocess configurations, we have the option to send a shutdown signal to the init process, who can in turn signal remaining processes and wait for a predetermined amount of time for them to terminate gracefully.
+This is referred to as *graceful shutdown*, and is implemented as part of the `/sbin/init` logic provided by `libukboot`.
+If the timeout is reached and there are processes still running, the init process returns and any surviving processes are terminated depending on the system configuration as described above.
+
+The table below summarizes process termination during shutdown for the possible configurations of libposix-process:
+
+|                           | Single Thread     | Multithreading | Multiprocess |
+| ------------------------- | ----------------- | -------------- | ------------ |
+| **Asynchronous Shutdown** | Yes[^3]           | Yes [^4]       | Yes [^4]     |
+| **Graceful Shutdown**     | No                | No [^5]        | Yes [^6]     |
+
+[^3]: The application thread is terminated and the system proceeds to shutdown.
+Notice that asynchronous shutdown depends on `CONFIG_LIBUKBOOT_MAINTHREAD`, which makes it possible to return the application's exit status.
+
+[^4]: If signals are enabled, libposix-process issues `SIGTERM` and blocks until the *application process* terminates.
+On multithreaded configurations that is the only process.
+On multiprocess with an implementation of init, that is the init process.
+On multiprocess without init, that is the process that calls `main()`.
+Once the application process terminates, the system proceeds to forcefully terminate any remaining processes by directly releasing their resources.
+
+[^5]: Graceful termination, however, can still be implemented by the application.
+From the system's perspective the termination path is the same one followed during asynchronous shutdown[^4].
+
+[^6]: Enabled by selecting `CONFIG_LIBUKBOOT_GRACEFUL_SHUTDOWN` in libukboot.
+Both the termination signal and the timeout are configurable.


lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
+# posix-process
+
+libposix-process provides a POSIX-compliant process implementation allowing Unikraft to run unmodified applications written for *NIX systems.
+To accommodate different application requirements while preserving Unikraft's minimal and modular design, libposix-process offers three configuration levels:
```
andreittr: A 2nd explanatory sentence about what functionality `posix-process` implements would go well here:
"It provides task identification (`getpid()`, `gettid()`, etc.), management (`clone()`, `execve()`), and signal support." You can then refer to these individual features by name when describing the configurations.

Perhaps "feature levels" would be more descriptive?

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Single thread
+Bare minimum process-like behavior for single process applications that execute on a single thread. Stubbed implementation required to build musl. Although acting as single-threaded, the process does not necessarily need to be backed by an underlying `uk_thread`, i.e. this configuration can be enabled even without libuksched. Configuration:
```
andreittr: 2nd sentence: "Provides stubbed implementations of task identification and management required by e.g., fully-featured libcs." -- I would drop the explicit naming of musl (an external lib) in a core component.

3rd sentence: "This feature level is independent of scheduler or (kernel) thread support, as these stubs never touch scheduler state." -- less detail on sched/`uk_thread` impl, but more explicit about _why_ single-thread is independent of sched support.

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+  * Select `CONFIG_LIBPOSIX_PROCESS` without additional options.
+
+### Multithreading
+Single process with multithreading. Configuration:
```
andreittr: Perhaps a 2nd sentence summarizing this feature level: "Provides true task identification and thread-like `clone()` support, with shared address space and resources."

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+  * Select `CONFIG_LIBPOSIX_PROCESS` and `CONFIG_LIBPOSIOX_PROCESS_MULTITHREADING`.
+  * Optionally select `CONFIG_LIBPOSIX_PROCESS_EXECVE` to enable `execve()` (requires VFS).
+  * Optional select `CONFIG_LIBPOSIX_PROCESS_SIGNAL` to enable POSIX signals.
```
andreittr: Go for one or the other for consistency:
- Optionally select ... (gramatically correct full sentence)
- (Optional) Select ... (qualifier in parantheses)

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Multiprocess:
+Full-featured multiprocess support. Configuration:
+  * Select `CONFIG_LIBPOSIX_PROCESS` and `CONFIG_LIBPOSIOX_PROCESS_MULTIPROCESS`. Requires VFS.
+  * Optional select `CONFIG_LIBPOSIX_PROCESS_SIGNAL` to enable POSIX signals.
```
andreittr: Perhaps move "Requires VFS" to the 2nd sentence of 1st paragraph? Makes it harder to miss.

Same argument for optionally/(optional) as above.

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+A summary of the syscalls provided by libposix-process and their behavior on each configuration is shown below:
+
+|  | Single Thread | Multithreading | Multiprocess |
+| ----- | ----- | ----- | ----- |
+| *gettid()* | Yes (emulated) | Yes | Yes |
+| *getpid()* | Yes (emulated) | Yes | Yes |
+| *getppid()* | Yes (emulated) | Yes | Yes |
```
andreittr: I believe "stubbed" is more in line with previous terminology than "emulated".

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+| *clone()* |	No (`ENOSYS`) | Yes (`ENOSUP` on `CLONE_VFORK`) | Yes |
```
andreittr: I believe the error is `ENOTSUP` (extra 'T').

Should we add a (foot)note to the "Yes" on the multiprocess column pointing to "Notes on multiprocess" below?

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+| *vfork()* |	No (`ENOSYS`) | No (`ENOSYS`) |	Yes |
+| *execve()* |	No (`ENOSYS`) | Tentative (see (2)) | Yes |
+| *_exit()* / *exit_group()* |	Tentative (see (1)) | Yes | Yes |
+| *wait4()* / *waitid()* | No (`ECHILD`) | No (`ECHILD`) | Yes |
+
+**Footnotes:**
+
+(1) When `CONFIG_LIBUKBOOT_MAINTHREAD` is selected, the application is executed on a dedicated `uk_thread`.
+In that case, single process single thread configurations can emulate `_exit()` and `exit_group()`, as any of these syscalls will results into terminating the underlying `uk_thread`.
```
andreittr: "will results into terminating" -> "will result in cleanly terminating"

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+If `CONFIG_LIBUKBOOT_MAINTHREAD` is not selected, the application executes on Unikraft's boot thread, which cannot be terminated.
+Similarly if `CONFIG_LIBUKSCHED` is not selected, the application executes without being backed by an underlying `uk_thread`.
+In both of these cases, it is not possible to handle a call to `_exit()` / `exit_group()` by propagating an error back to the application, as these syscalls do not return.
+In addition, it is not possible to catch these cases at compile time, as in bincompat mode (app-elfloader) the payload is provided in a separate binary image.
+Unikraft will therefore panic with an appropriate error message.
```
andreittr: I'd reword these last 3 sentences:
"For these reasons `_exit()`/`exit_group()` cannot be meaningfully implemented under such configurations."
"Since these syscalls never return, and thus cannot signal failure to their caller, calling them is an unrecoverable runtime error, on which Unikraft will panic with an appropriate error message."

It makes it more explicit that exit doesn't make sense under this config, and that calling it is a logical error. IMO we can drop the bit about bincompat and why it has to be a runtime error and not a compile-time one, as it's a fairly low-level implementaion detail.

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+(2) Calling `execve()` from a single-process application is possible, and results into the calling process' context being replaced with that of the executed program.
```
andreittr: Perhaps "from a single-process application" -> "without multiprocessing support" would be more in line with the config this footnote applies to, as "single-process application" focuses on the workload rather than config.

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+Because `execve()` is passed a filesystem path, `CONFIG_LIBPOSIX_PROCESS_EXECVE` depends on VFS.
+To avoid unnecessary bloat on appliations that may not use `execve()`, `CONFIG_LIBPOSIX_PROCESS_EXECVE` is provided as an optional selection in multithreaded configurations.
+
+## Requirements of execve()
+
+As stated earlier `execve()` depends on VFS. Additionally, since Unikraft executes on a single address space, it is required that executables passed to `execve()` must be compiled as PIE.
```
andreittr: Since you mention PIE, maybe add a sentence mentioning `ukbinfmt` and the fact that `execve` delegates binary support to it?

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+A summary of the syscalls provided by libposix-process and their behavior on each configuration is shown below:
+
+|  | Single Thread | Multithreading | Multiprocess |
+| ----- | ----- | ----- | ----- |
+| *gettid()* | Yes (emulated) | Yes | Yes |
+| *getpid()* | Yes (emulated) | Yes | Yes |
+| *getppid()* | Yes (emulated) | Yes | Yes |
+| *clone()* |	No (`ENOSYS`) | Yes (`ENOSUP` on `CLONE_VFORK`) | Yes |
+| *vfork()* |	No (`ENOSYS`) | No (`ENOSYS`) |	Yes |
+| *execve()* |	No (`ENOSYS`) | Tentative (see (2)) | Yes |
```
andreittr: Why is `execve()` only "Tentative" on multi-thread? AFAICT what we describe doing in footnote (2) ("replacing the only process with another binary image") is _exactly_ what `execve()` is expected to do in that case, and we're not missing any functionality :thinking: .

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+| *_exit()* / *exit_group()* |	Tentative (see (1)) | Yes | Yes |
```
andreittr: "Tentative" suggests support under this config might improve in the future. The footnote however explains that exit depends on an orthogonal core feature (ukboot mainthread), and I don't see how this limitation can be worked around. Perhaps a better word would be "Partial" or "Conditional", with the same footnote.

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+Because `execve()` is passed a filesystem path, `CONFIG_LIBPOSIX_PROCESS_EXECVE` depends on VFS.
+To avoid unnecessary bloat on appliations that may not use `execve()`, `CONFIG_LIBPOSIX_PROCESS_EXECVE` is provided as an optional selection in multithreaded configurations.
+
+## Requirements of execve()
+
+As stated earlier `execve()` depends on VFS. Additionally, since Unikraft executes on a single address space, it is required that executables passed to `execve()` must be compiled as PIE.
+
+## Notes on multiprocess
+
+Being a single-address-space operating system, instead of the traditional `fork()` that creates a new address space for the child, Unikraft implements multiprocess with `vfork()`.
```
andreittr: @ the end: "with `vfork()` only." would make it clearer we're contrasting against OSes that do _both_ fork and vfork (and that vfork isn't our invention, as some naiver readings of the current wording can imply), and underline that we're _not_ implementing fork (nor is it really on the table in a single addrspace OS).

lib/posix-process/README.md
```diff
@@ -0,0 +1,94 @@
...
+When calling `vfork()`, the child shares the same address space as the parent, and the parent's execution is suspended until the child either calls `execve()`, or terminates by calling `_exit()` / receiving a terminating signal.
+This introduces limitations on what the child can do until it calls `execve()` (or terminates), as it must ensure that it doesn't leave the parent's memory in an incosistent state.
+Specifically, according to [vfork(2)](https://man7.org/linux/man-pages/man2/vfork.2.html), the only operations the child can do are: (1) update the memory location that stores return value of `vfork()`, (2) call `execve()`, (3) call `_exit()`.
+In any other case the resulting behavior is undefined.
+Fortunately it is common that appliations spawn new processes by calling `fork()` immediately followed by an `execve()`, so running a multiprocess application on Unikraft that does not natively support `vfork()` should be possible with minimal changes.
+
+Another important aspect of multiprocess systems is the existence of an init process.
```
andreittr, line 74: This paragraph on init should be its own section, or at least a sub-section under multiprocess. It's a crucial part of how multiprocess works and we don't want users to miss it.

lib/posix-process/README.md
michpappas: Alright. The intention behind the sentence on bincompat was mostly so that we don't forget later, as well as future devs.

lib/posix-process/README.md
michpappas: You're right, I'll use "Conditional".

lib/posix-process/README.md
michpappas: Should be "Conditional", as it depends on the user selecting the feature. Will update.

lib/posix-process/README.md
michpappas: Yes, and in addition I think I will improve that altogether to clarify that when porting one can also call `posix_spawn()`.

lib/posix-process/README.md
michpappas: Will update to something along these lines.

lib/posix-process/README.md
michpappas: I find "thread-like clone() support" unclear - not to the reader by also to myself. Is that meant to mean that one can create threads with `clone()`?

lib/posix-process/README.md
michpappas: I just found out that markdown [supports footnotes for a while now](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/), so I will adapt.

> Should we add a (foot)note to the "Yes" on the multiprocess column pointing to "Notes on multiprocess" below?

That would require making every row into a link though, which is possible but maybe a bit ugly. I think the reader will get there anyway.

michpappas: > I just found out that markdown [supports footnotes for a while now](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/), so I will adapt.

Saying that, using proper footnotes would move these sentences to the bottom of the document. For me it's okay, but I could take a second opinion. Let me know what you think. In the unlikely case you don't know it already, you can see the rendered version by clicking on "View File" in "Files changed":

image

lib/posix-process/README.md
andreittr: You're right, that's a bad phrasing; how about: "[...] and `clone()`ing of posix threads"?

lib/posix-process/README.md
andreittr: Notes on the implementation are in my view best kept as source comments -- close to the implementation itself. Someone discovering a lib is likely to go for the README, while an experienced hacker would go straight for the headers & source.

lib/posix-process/README.md
michpappas: > 2nd sentence: "Provides stubbed implementations of task identification and management required by e.g., fully-featured libcs." -- I would drop the explicit naming of musl (and external lib) in a core component.

This is not entirely true either though as it doesn't really provide management facilities, except from exit() conditionally to very specific configuration, and also we don't know whether other libcs have the same requirement as musl. On the other hand, while musl is indeed an external library, it's still a bit special as it's the officially supported implementation of libc, and also it is actually a fact that the only reason this mode exists is for compiling musl (otherwise one could execute single process single thread applications that use libc without selecting libposix-process).

> 3rd sentence: "This feature level is independent of scheduler or (kernel) thread support, as these stubs never touch scheduler state." -- less detail on sched/uk_thread impl, but more explicit about why single-thread is independent of sched support.

We need some reference to the underlying uk_thread as we rely on it when we later describe the conditional functionality of exit() in single thread mode. I will update it to include the "why".

lib/posix-process/README.md
michpappas: This is a bit of a hairy topic. As this section is about requirements, the only reason `ukbinfmt` fits here is because the ELF part of binfmt lives in `app-elfloader`. This is however a violation of the principle that Unikraft must not depend on external libraries (appications are libraries too), so instead of documenting it perhaps we should discuss that instead.

lib/posix-process/README.md
michpappas: It's crucial indeed. On the other hand, all and all the multiprocess section is two paragraphs long, so I feel it's a bit strange to have a subsection, and it's not a wall of text so hard to miss once you're there anyway.

lib/posix-process/README.md
```diff
@@ -0,0 +1,111 @@
...
+That requires additional handling at the level of libposix-process.
+
+The way libposix-process implements shutdown at the process level depends on the system configuration.
+If signals are enabled, it is possible to signal the application and wait for it to terminate before shutting the rest of the system down.
+When that happens the application returns or calls `_exit()` / `exit_group()` directly and its resources are released.
+If signals are not enabled, all processes are terminated by forcefully and their resources are directly released.
+
+In multiprocess configurations, the init process can signal all processes and wait for a predetermined amount of time for them to terminate gracefully.
+This is referred to as *graceful shutdown*.
+If the timeout is reached and there are processes still running, the init process returns and any surviving processes are terminated depending on the system configuration as described above.
```
andreittr, line 96: If I understand right, this paragraph is meant to describe what _any_ `init` implementation might want to do to support graceful shutdown, as well as what `ukboot`'s init logic does.

In that case it would be useful to explain how the entire process is triggered and contrast it with normal async shutdown:
"In multiprocess configurations, we have the option to send a shutdown signal to the init process, who can in turn signal remaining processes and wait for a predetermined amount of time for them to terminate gracefully."

> the init process returns

I think "exits" or "terminates" might be more explicit than "returns".

Also, surviving processes after init exits are always forcefully terminated; the current wording suggests they would get a signal&wait instead.

Finally, AFAICT all graceful shutdown logic is actually outside of posix-process and in ukboot; now I completely understand why we'd want to document it here, but might be worth pointing to ukboot when introducing it:
"This feature is referred to as *graceful shutdown*, and is provided by `ukboot`."

lib/posix-process/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## System Shutdown
+
+From Unikraft 0.16.0, libukboot supports asynchronously triggered shutdown.
+At the most fundamental level, when the system receives a shutdown request, the application must be terminated before proceeding to terminate system-wide components like the network and filesystem.
+When libposix-process is selected, all processes must be terminated and their resources must be released.
+That requires additional handling at the level of libposix-process.
+
+The way libposix-process implements shutdown at the process level depends on the system configuration.
+If signals are enabled, it is possible to signal the application and wait for it to terminate before shutting the rest of the system down.
```
andreittr: "it is possible" is ambiguous phrasing that suggests this may or may not happen. AFAICT if signals are enabled we always signal the application process & wait, so we should positively say what happens (like we do lower down: "all processes are terminated")

lib/posix-process/README.md
```diff
@@ -0,0 +1,111 @@
...
+When that happens the application returns or calls `_exit()` / `exit_group()` directly and its resources are released.
+If signals are not enabled, all processes are terminated by forcefully and their resources are directly released.
```
andreittr: "are terminated by forcefully" -> "are terminated forcefully"

lib/posix-process/README.md
```diff
@@ -0,0 +1,111 @@
...
+  * Select `CONFIG_LIBPOSIX_PROCESS` and `CONFIG_LIBPOSIOX_PROCESS_MULTIPROCESS`. Requires VFS.
+  * Optional: Select `CONFIG_LIBPOSIX_PROCESS_SIGNAL` to enable POSIX signals.
+
+A summary of the syscalls provided by libposix-process and their behavior on each configuration is shown below:
+
+|  | Single Thread | Multithreading | Multiprocess |
+| ----- | ----- | ----- | ----- |
+| *gettid()* | Yes (stubbed) | Yes | Yes |
+| *getpid()* | Yes (stubbed) | Yes | Yes |
+| *getppid()* | Yes (stubbed) | Yes | Yes |
```
andreittr: Nit: could we please have this table nicely aligned in text form with some extra spaces? :pleading_face: 2nd table too

Markdown's coolest feature is equal human-readability in plaintext or rendered form, but this is about as easy to parse for humans as a LaTeX tabular :smile: .

lib/posix-process/README.md
andreittr: I think the footnotes look perfectly fine; in plaintext they're close by and easy to jump to when reading, and in the rendered version has hyperlinks :+1:

lib/posix-process/README.md
michpappas: > "In multiprocess configurations, we have the option to send a shutdown signal to the init process, who can in turn signal remaining processes and wait for a predetermined amount of time for them to terminate gracefully."

Ok

> I think "exits" or "terminates" might be more explicit than "returns".

It literally returns though

> Also, surviving processes after init exits are always forcefully terminated; the current wording suggests they would get a signal&wait instead.

It is correct that it depends on the configuration: If signals are enabled they get a `SIGKILL`  and we block indefinitely. See https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/shutdown.c.

> Finally, AFAICT all graceful shutdown logic is actually outside of posix-process and in ukboot; now I completely understand why we'd want to document it here, but might be worth pointing to ukboot when introducing it:
"This feature is referred to as graceful shutdown, and is provided by ukboot."

Ok, will change to "and is implemented as part of the init process provided by `libukboot`."

lib/posix-process/README.md
michpappas: Will rephrase to:
> If signals are enabled, the application process is sent `SIGKILL` and shutdown blocks until the application process terminates.
Once the application process terminates, any remaining processes are terminated forcefully.
If signals are not enabled, all processes are terminated forcefully.
In both cases the result is the same: all process resources are directly released.

lib/posix-process/README.md
michpappas: Actually nevermind, the fact is that tit's already there, and that doesn't change whether we document it or not. If we end up moving it out we can update the document.

lib/posix-process/README.md
andreittr: > If signals are enabled they get a SIGKILL and we block indefinitely.

Looking at the impl in shutdown.c; AFAICT [we only SIGTERM & wait PID_INIT](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/shutdown.c#L49) and [force-kill remaining processes](https://github.com/unikraft/unikraft/blob/staging/lib/posix-process/shutdown.c#L55). I don't see any way to signal & block _after_ the application process has exited, which is the situation you're describing here.

andreittr (approved):
Looks good, thanks!

lib/posix-process/README.md
michpappas: My bad, I meant `SIGTERM`. Also I see what you mean for the case of init, will update.

michpappas:
@andreittr I pushed one more update

andreittr (approved):
Re-reviewed latest changes, looks good!

andreittr:

PR #1647, opened 2025-05-16
lib/vfscore: Clean up relative path resolution in *at syscalls

andreittr:
### Description of changes

This changeset cleans up the implementation of `*at` syscalls, centralizing the construction of absolute paths from relative paths in `taskat_conv`, and removing duplicate code from syscalls.

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/vfscore/task.c
```diff
@@ -143,6 +143,57 @@ task_conv(struct task *t, const char *cpath, int acc __unused, char *full)
...
+		error = fget(dirfd, &dirfp);
+		if (error)
+			goto out_error;
+
+		dirvp = dirfp->f_dentry->d_vnode;
+		vn_lock(dirvp);
+
+		/* get working directory */
+		strlcpy(wdpath, dirfp->f_dentry->d_mount->m_path, PATH_MAX);
+		strlcat(wdpath, dirfp->f_dentry->d_path, PATH_MAX);
```
michpappas: Is it worth applying some error checking using `strclcat()`'s return value to detect truncation?

lib/vfscore/main.c
```diff
@@ -2147,34 +2147,41 @@ UK_SYSCALL_R_DEFINE(int, truncate, const char*, pathname, off_t, length)
...
 	size  = 0;
 	error = sys_readlink(path, buf, bufsize, &size);
 
 	if (error != 0)
-		goto out_errno;
+		goto out_err;
 
 	return size;
-	out_errno:
-	errno = error;
```
michpappas, line 2176: Finally this direct errno assignment is out, I remember being bitten by it at some point.

lib/vfscore/main.c
```diff
@@ -2147,34 +2147,41 @@ UK_SYSCALL_R_DEFINE(int, truncate, const char*, pathname, off_t, length)
...
+		    char *, buf, size_t, bufsize)
 {
 	struct task *t = main_task;
 	char path[PATH_MAX];
 	int error;
 	ssize_t size;
 
-	error = -EINVAL;
+	error = EINVAL;
 	if (bufsize <= 0)
```
michpappas, line 2018: Maybe annotate all checks for error conditions with `unlikely`? But then again from what I see vfscore doesn't fully follow the convention, and it's going the way of the Dodo anyway, so if you think it's unnecessary feel free to ignore.

lib/vfscore/main.c
```diff
@@ -1283,67 +1283,13 @@ UK_SYSCALL_R_DEFINE(int, rename, const char*, oldpath, const char*, newpath)
 
 	out_error:
 	trace_vfs_rename_err(error);
-	return -error;
+	return ((error > 0) ? -error : error);
```
michpappas, line 1225: This looks so ugly... I guess something called by `taskat_conv()` returns a positive value that `taskat_conv()` happily passes forward? The `error = ENOENT` could already be changed to a negative value. Btw, I also see the same pattern in `unlikat()`. Again if vfscore is a mess and we know these patches work as is, we can keep it as is given that we're deprecating it anyway.

michpappas (commented):
@andreittr I'm leaving some comments without requesting changes, as (1) vscore's codebase is messy (2) we're deprecating it (3) we know that these patches work. If you think we shouldn't address any of this please let me know and I will merge.

lib/vfscore/main.c
andreittr: vfscore's use of positive error codes is indeed a mess, and not something I'm too keen on fixing, due to the same points you raised. I say we look the other way until it's gone :see_no_evil: .

lib/vfscore/main.c
andreittr: Yeah, pushing branch hints across vfscore would greatly expand the scope here, not sure it's worth it.

lib/vfscore/task.c
andreittr: That's actually a very good point, and pretty trivial to add; will do in a quick patch :+1: .

andreittr:
Pushed small change adding error checking to path length.

michpappas (approved):

PR #1649, opened 2025-05-23
lib/posix-process: Fix clone() return value on thread create

michpappas:
### Prerequisite checklist

### Base target

### Description of changes

Commit 08c6ac9b0ffd ("lib/posix-process: Update semantics of vfork()'s return type") introduced a regression on the return value of clone() when new threads are created. clone() should return the tid on the creation of a new thread, and the pid when a new process is created. Although in the current implementation the pid of a new process matches the underlying tid, fix this by using the corresponding macro for each case, to keep semantics clear.

### DIFF
--- a/lib/posix-process/clone.c
+++ b/lib/posix-process/clone.c
@@ -559,8 +559,13 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
 
 	uk_thread_set_runnable(child);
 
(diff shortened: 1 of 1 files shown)


andreittr (approved):
Looks good; it's nice to keep PIDs & TIDs separate conceptually, even if they do share a namespace in the current implementation.

andreittr:

PR #1651, opened 2025-06-04
lib/uksched: Rework waitqueues to use thread data

andreittr:
### Description of changes

This changeset reworks the implementation of wait queues to use wakeup entries (tickets) embedded within the thread data structure instead of allocated on the execution stack of a thread.
This ensures a thread is waiting on at most one queue at a time, and enables detection when terminating a waiting thread.

The public wait queue API remains unchanged.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/uksched/include/uk/thread.h
+++ b/lib/uksched/include/uk/thread.h
@@ -46,6 +46,7 @@ extern "C" {
 #endif
 
 struct uk_sched;
+struct uk_thread;
 
 typedef void (*uk_thread_dtor_t)(struct uk_thread *);
 typedef void (*uk_thread_gc_t)(struct uk_thread *, void *);
@@ -65,6 +66,7 @@ struct uk_thread {
 	uint32_t flags;
 	__snsec wakeup_time;
 	struct uk_sched *sched;
+	struct uk_waitq_ticket wait_ticket;
 
 	struct {
 		struct uk_alloc *t_a;
@@ -100,6 +102,24 @@ static inline
 struct uk_thread *uk_thread_current(void)
 {
 	return ukplat_per_lcpu_current(__uk_sched_thread_current);
+}
+
+/**
+ * Return thread containing the waitqueue ticket `ticket`.
+ */
+static inline
+struct uk_thread *uk_thread_of_wait_ticket(struct uk_waitq_ticket *ticket)
+{
+	return __containerof(ticket, struct uk_thread, wait_ticket);
+}
+
+/**
+ * True if `thread` is linked in a wait queue.
+ */
+static inline
+bool uk_thread_in_waitq(struct uk_thread *thread)
+{
+	return !!thread->wait_ticket.wq;
 }
 
 /*
@@ -652,6 +672,19 @@ void uk_thread_block_until(struct uk_thread *thread, __snsec until);
 void uk_thread_block_timeout(struct uk_thread *thread, __nsec nsec);
 void uk_thread_block(struct uk_thread *thread);
 void uk_thread_wake(struct uk_thread *thread);
+
+/**
+ * Set the wakeup timer state for thread `t`.
+ *
+ * If `deadline`:
+ *  != 0: arm a wakeup timer for time `deadline`
+ *  == 0: disarm wakeup timer
+ */
+static inline
+void uk_thread_set_wakeup(struct uk_thread *t, __snsec deadline)
+{
+	t->wakeup_time = deadline;
+}
 
 /**
  * Macro to access a Unikraft thread-local (UKTLS) variable of a
--- a/lib/uksched/include/uk/wait.h
+++ b/lib/uksched/include/uk/wait.h
@@ -102,7 +102,7 @@ do { \
 	unsigned long flags; \
 	int timedout = 0; \
 	DEFINE_WAIT(__wait); \
-	if (!(condition)) { \
+	if (unlikely(!(condition))) { \
 		__current = uk_thread_current(); \
 		for (;;) { \
 			/* protect the list */ \
--- a/lib/uksched/include/uk/wait.h
+++ b/lib/uksched/include/uk/wait.h
@@ -1,27 +1,14 @@
-/* SPDX-License-Identifier: BSD-2-Clause */
-/*
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in the
- *    documentation and/or other materials provided with the distribution.
- *
- * THIS SOFTWARE IS PROVIDED BY AUTHOR AND CONTRIBUTORS ``AS IS'' AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
- * ARE DISCLAIMED.  IN NO EVENT SHALL AUTHOR OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
- * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
- * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
- * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
- * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
- * SUCH DAMAGE.
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
  */
-/* Ported from Mini-OS */
+
+/* Unikraft wait queues */
+
+/* This is a reimplementation of the API of waitqueues from Xen MiniOS, whose
+ * BSD-2 licensed implementation served as the first waitqueues in Unikraft.
+ */
 
 #ifndef __UK_SCHED_WAIT_H__
 #define __UK_SCHED_WAIT_H__
@@ -30,6 +17,7 @@
 #include <uk/plat/lcpu.h>
 #include <uk/plat/time.h>
 #include <uk/sched.h>
+#include <uk/thread.h>
 #include <uk/wait_types.h>
 
 #ifdef __cplusplus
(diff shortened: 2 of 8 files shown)


lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
-	if (lock) \
-		unlock_fn(lock); \
-	uk_sched_yield(); \
-	if (lock) \
-		lock_fn(lock); \
-	if (deadline_condition) \
-		timedout = 1; \
-	ukplat_spin_lock_irqsave(&((wq)->sl), flags); \
-	/* need to wake up */ \
-	uk_thread_wake(__current); \
```
mogasergiu, line 164: huh, so the `uk_thread_wake` from here and `__wq_wait_event_deadline were redundant?

mogasergiu (approved):
LGTM, just a question though, as I see that function was not kept.

lib/uksched/include/uk/wait.h
andreittr: Yep; the thread gets awoken when notified (otherwise _how could it even reach this execution point?!_) I had noticed as much earlier on when getting inspiration for the the ukfile/pollqueue code -- where this redundant self-wake was never implemented and it works perfectly -- but decided to not muck about with working waitq code without a good reason. That reason has come now :smile: .

lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
+				uk__waitq_timedout = 1;			\
+				break;					\
+			}						\
+			/* Still need to wait; back to sleep */		\
+		}							\
+		/* No more waiting, remove from queue & exit */		\
+		uk__waitq_remove(uk__waitq_current);			\
+		uk__waitq_unlock((wq), uk__waitq_flags);		\
+	}								\
+uk__waitq_wait_out:							\
```
mogasergiu: I had a random realisation while working on something else. Using this macro more than once in the same function will break the build as the label will be duplicated. I think we need a two-level macro to uniquefy this label. Something like this:

```diff
diff --git a/lib/uksched/include/uk/wait.h b/lib/uksched/include/uk/wait.h
--- a/lib/uksched/include/uk/wait.h
+++ b/lib/uksched/include/uk/wait.h
@@ -103,8 +103,8 @@ void uk__waitq_block_until(struct uk_thread *thread, __snsec deadline)
  * If `lock` evaluates true, it will be released before sleeping with
  * `unlock_fn` and re-acquired after wakeup with `lock_fn`.
  */
-#define uk__waitq_wait_until(wq, condition, deadline, clock_expr,	\
-			     lock_fn, unlock_fn, lock)			\
+#define uk__waitq_wait_until_uniq(wq, condition, deadline, clock_expr,	\
+			     lock_fn, unlock_fn, lock, uniq)			\
 ({									\
 	int uk__waitq_timedout = 0;					\
 									\
@@ -116,7 +116,7 @@ void uk__waitq_block_until(struct uk_thread *thread, __snsec deadline)
 		if ((condition)) {					\
 			/* Condition changed in the meantime */		\
 			uk__waitq_unlock((wq), uk__waitq_flags);	\
-			goto uk__waitq_wait_out;			\
+			goto UK_CONCAT(uk__waitq_wait_out, uniq);			\
 		}							\
 		/* Need to wait; add self to waitqueue & block */	\
 		uk__waitq_add((wq), uk__waitq_current);			\
@@ -142,10 +142,15 @@ void uk__waitq_block_until(struct uk_thread *thread, __snsec deadline)
 		uk__waitq_remove(uk__waitq_current);			\
 		uk__waitq_unlock((wq), uk__waitq_flags);		\
 	}								\
-uk__waitq_wait_out:							\
+UK_CONCAT(uk__waitq_wait_out, uniq):							\
 	uk__waitq_timedout;						\
 })
 
+#define uk__waitq_wait_until(wq, condition, deadline, clock_expr,	\
+			     lock_fn, unlock_fn, lock)			\
+	uk__waitq_wait_until_uniq(wq, condition, deadline, clock_expr, \
+				lock_fn, unlock_fn, lock, __COUNTER__)
+
 /**
  * INTERNAL. Wait exactly once using `wq` until awoken or a deadline.
  *
```

lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
 
+/**
+ * INTERNAL. Add the thread `thread` to waitqueue `wq`.
+ *
+ * `thread` must not be linked in any waitqueue and the lock of `wq` must be
+ * acquired.
+ */
 static inline
-int uk_waitq_empty(struct uk_waitq *wq)
+void uk__waitq_add(struct uk_waitq *wq, struct uk_thread *thread)
```
michpappas: The comment about `UK__` naming provided in the changes in `wait_types.h` applies here too, as well as  in the rest of the API calls prefixed with `uk__`.  These private inline functions should be placed in private headers instead.

lib/uksched/include/uk/wait.h
andreittr: Ugh, I don't like 2-level macros, it's cursed enough as it is; but you do make a good point about label redefinition.
Dijkstra was right all along, I'll convert the label-goto back into a loop as before.

andreittr:
Updated addressing last comment.

lib/uksched/include/uk/wait_types.h
```diff
@@ -33,34 +18,35 @@
...
 
-struct uk_waitq_entry {
-	int waiting;
-	struct uk_thread *thread;
-	UK_STAILQ_ENTRY(struct uk_waitq_entry) thread_list;
+struct uk_waitq;
+
+struct uk_waitq_ticket {
+	/* Entry to link into wait queue */
+	UK_TAILQ_ENTRY(struct uk_waitq_ticket) list;
```
michpappas: According to https://github.com/unikraft/unikraft/blob/9603a4ab18de82c9401581db0e7ce929cbfda7f2/include/uk/list.h#L435 the implementation in `compat_list.h` should be deprecated in favor of `uk/list.h`, so perhaps we should be using a data structure from there instead?

lib/uksched/include/uk/wait_types.h
```diff
@@ -33,34 +18,35 @@
...
-	UK_STAILQ_HEAD(wait_list, struct uk_waitq_entry) wait_list;
+	UK_TAILQ_HEAD(waiters, struct uk_waitq_ticket) waiters;
+	__spinlock lock;
 };
 
-#define __WAIT_QUEUE_INITIALIZER(name) \
-{ \
-	UKARCH_SPINLOCK_INITIALIZER(), \
-	UK_STAILQ_HEAD_INITIALIZER(name.wait_list) \
+#define UK__WAITQ_INITIALIZER(name)				\
```
michpappas: We don't have a `UK__` convention. The convention is for preprocessor directives that add another level of indirection (like here), and other private preprocessor definitions, to be prepended with double underscore. So the initial state was correct, although it should have also be prepended with `UK`, i.e. `__UK_WAIT_QUEUE_INITIALIZER`. But then again is the indirection necessary in the first place?

lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
 {
-	entry->thread = thread;
-	entry->waiting = 0;
+	return UK_TAILQ_EMPTY(&wq->waiters);
 }
 
+/**
+ * INTERNAL. Add the thread `thread` to waitqueue `wq`.
+ *
+ * `thread` must not be linked in any waitqueue and the lock of `wq` must be
```
michpappas: The wording initially confused me ( the "any" in particular), but then I realized you're referring to the `!t->wq` assertion. How about "`thread` must not be already in a waitqueue"? If you think it's just me feel free to ignore.

lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
+ * Expression returns non-zero if timed out.
+ *
+ * Depending on `condition` may wait 0 or multiple times in the queue.
+ * If `deadline` is non-zero, wait at most until that point in time;
+ * after wakeup compare `deadline` with the value supplied by `clock_expr` to
+ * determine whether a timeout has occurred.
+ * If `lock` evaluates true, it will be released before sleeping with
+ * `unlock_fn` and re-acquired after wakeup with `lock_fn`.
+ */
+#define uk__waitq_wait_until(wq, condition, deadline, clock_expr,	\
```
michpappas: Maybe this function deserves to be a `static inline` too?

lib/uksched/include/uk/wait.h
```diff
@@ -39,183 +25,227 @@ extern "C" {
...
+			uk__waitq_block_until(uk__waitq_current, (deadline));\
+			uk__waitq_unlock((wq), uk__waitq_flags);	\
+			if ((lock))					\
+				unlock_fn((lock));			\
+			uk_sched_yield();				\
+			/* Awoken */					\
+			if ((lock))					\
+				lock_fn((lock));			\
+			uk__waitq_lock((wq), uk__waitq_flags);		\
+			if ((condition))				\
```
michpappas, line 136: Although the order of the checks against condition and timeout is the same in the original implementation, this part caught my eye. It is possible that the condition changed AND a timeout occurred, yet in that case the caller won't know that both happened as we ignore the timeout. I guess it doesn't matter, as the caller primarily cares about the condition OR a timeout while the condition is still not met. On the other hand, if the reverse happens, i.e. the condition is met shortly after the timeout, the caller still doesn't care given that it requested that we inform it during that timeout window. So I think it's all good.

lib/uksched/include/uk/thread.h
```diff
@@ -102,6 +104,24 @@ struct uk_thread *uk_thread_current(void)
...
+struct uk_thread *uk_thread_of_wait_ticket(struct uk_waitq_ticket *ticket)
+{
+	return __containerof(ticket, struct uk_thread, wait_ticket);
+}
+
+/**
+ * Non-zero if `thread` is linked in a wait queue.
+ */
+static inline
+int uk_thread_in_waitq(struct uk_thread *thread)
```
michpappas: Because of the unconventional return value maybe better to use the internal `__Bool` type here? (see `include/uk/arch/types.h`).

lib/uksched/sched.c
```diff
@@ -302,6 +302,10 @@ void uk_sched_thread_terminate(struct uk_thread *thread)
 	uk_pr_debug("%p: thread %p (%s) terminated\n",
 		    sched, thread, thread->name ? thread->name : "<unnamed>");
 
+	if (uk_thread_in_waitq(thread))
+		uk_pr_warn("%p: thread %p (%s) terminated while waiting\n",
```
mogasergiu, line 306: Does it really have to be a warning level message? I would expect a debug/info level message maybe.

mogasergiu (approved):
LGTM

lib/uksched/sched.c
andreittr: It opens up a site for a UAF bug, I think it warrants at least a warning message. Either way, this warning will be superseded by a correct implementation that does _not_ leave waiting terminated threads (and thus has nothing to print).

michpappas (changes requested):
Besides fixing the underlying issue, this is a more elegant implementation. Only a few minor comments from me.

lib/uksched/include/uk/wait.h
andreittr: It cannot be, as `condition` and `clock_expr` are meant to be expressions evaluated inside the _caller's context_. Since C does not support true closures, the best we can do is callback functions, which for something as low-level and on the hotpath as waitqueues is an abstraction too many. Hence, macro magic. I don't find it the most elegant solution either, but it is what it has to be.

Another alternative would be to have static inlines for every `condition` and `clock_expr` used; for `clock_expr` that would be easy actually, but for `condition` we'd need at least `wait_until_eq`, `wait_until_neq`, `wait_until_gt`, `wait_until_lt` (+ variants to handle different types, because the loop inline has to re-evaluate `condition` on every check) to cover all the existing use-cases in the core codebase. And it gets much hairier if we try to do more clever things with a waitq ticket during the wait loop. I don't think this is feasible.

A final alternative would be for `wait.h` to publicly export all the primitives it uses for locking, waiting, etc. as static inlines (as it already defines most of them for internal use), and rely on callers to build their custom loops for anything but trivial "wait with a deadline" use-cases. This does not scale and opens up opportunities for bugs at every wait callsite, as wait loops are _really easy_ to get wrong, speaking from experience here :smile: .

lib/uksched/include/uk/wait.h
andreittr: I guess the order of checks ultimately is determined by the semantics you want from a "wait until condition or timeout" operation. If by the time this piece of code runs (which can be arbitrarily later than either the timeout alarm or the event that woke the thread from the queue) both `condition` and `timeout` are true, we cannot know which occured first (not without keeping a "last event wakeup" timestamp in `struct uk_thread` anyway), and we optimistically check for `condition` first: as that's ultimately what calling code is waiting on, and is expecting a timeout only if `condition` is not met at a time point `>= deadline`.

lib/uksched/include/uk/wait.h
michpappas: Okay sure, I think I wrote this comment on a separate pass than the one where I read the function, and at that time didn't notice the condition / expr. Let's keep it simple then, no need to add complexity just for that.

lib/uksched/include/uk/wait_types.h
andreittr: > But then again is the indirection necessary in the first place?

At closer inspection I don't think so; I'll remove.

> We don't have a UK__ convention.

That's fair, but we need a convention that's not underscore prefixes for stuff that ends up in public headers, as underscore prefixes are reserved. On a cursory reading, `_[_A-Z][0-9_a-zA-Z]*` is reserved by ISO C and then whatever posix throws on top. _Technically_ if you use any of these behavior is undefined. Practically we're probably fine, at worst this'll throw up a ton of compiler warning some years down the line when clang/gcc gets _really_ pedantic. Either way, I don't like consciously breaking standards without a good reason :smile: .
I guess underscore-lowercase is fine (TIL) but that doesn't look good for macros.

lib/uksched/include/uk/wait_types.h
andreittr: ~~I'm guessing that comment was meant in the sense of "TODO: actually implement all functionality of the old linked list implementation so we can get rid of it" because I tried to use the new lists, but they just did not cut it.~~

~~I explicitly needed a doubly linked tail queue (DLL w/ fast access to tail & remove in-place ability), and the closest thing is `uk_hlist`; that does not track the tail however, and not wishing to hack around it at callsites in the waitq implementation, I decided to just use the one in `compat_list.h`.~~

Edit: on much closer inspection, I _think_ I can port over to `uk_list`, as it might work out.

Edit2: not a huge fan of the `uk/list.h` API; needs a lot of `__containerof` use. But let's see after the pass how big of an issue that is.

Edit3: ok, it's not as bad as I first thought; there are some neat helpers in list.h; could use ~~better~~ some documentation though.

lib/uksched/include/uk/wait.h
andreittr: > These private inline functions should be placed in private headers instead.

You mean like `include/uk/wait_ops.h` (similar to `wait_types.h`) that only gets `#include`d by this header? Sure, it'd split the API and impl of waitq a bit better.

Re: naming, see response to `wait_types.h` comment.

lib/uksched/include/uk/wait_types.h
michpappas: Yes, fully aware of the potential conflicts with libc, and the current conventions were established with these in mind, while taking into account other tradeoffs too, given Unikraft's somewhat special nature. The conventions may or may not be perfect, and that's debatable (although doesn't scale well), but either way we should avoid applying new conventions unilaterally in an ad-hoc manner. What I propose in these cases is (1) stick to the convention in PRs (2) if *really* necessary, initiate discussion to revise a convention (but again we should avoid that where possible).

lib/uksched/include/uk/wait.h
michpappas: > You mean like include/uk/wait_ops.h (similar to wait_types.h) that only gets #included by this header?

Just to confirm, yes. And being private they shouldn't use the convention of the public API (`uk_xxx`).

lib/uksched/include/uk/thread.h
andreittr: On closer examination, `thread.h` already does `#include <stdbool.h>` so we have the stdC `bool` available. I'll use that.

andreittr:
Pushed addressing comments

michpappas (approved):

PR #1654, opened 2025-06-06
lib/posix-vfs-fstab: Introduce fstab library

andreittr:
### Description of changes

This change introduces posix-vfs-fstab, a library tasked with mounting filesystems at boot according to defined tables (fstabs).
Its general functionality mirrors that of legacy vfscore automount.

As a prerequisite the build system now supports `Config.uk.d` drop-in directories for Kconfig files in addition to the required `Config.uk`. This allows libraries (like this one) to better parcel out complex or verbose config options.

~~Depends on the following PR and has it included as a squashed commit; ignore during review:~~
~~- https://github.com/unikraft/unikraft/pull/1610~~

### Prerequisite checklist

### Base target

### Additional configuration

Config options for setting the builtin fstab(s) are similar to but not the same as the ones used by vfscore's `automount`. Users need to reconfigure their kernels, and adapt any scripts that emit `.config` files accordingly.

Legacy vfscore's automount functionality remains unchanged and orthogonally available.


support/build/config-submenu.sh
```diff
@@ -160,6 +160,15 @@ esac
 for L in "${LIBS[@]}"; do
 	CONFIG_UK="${L}/Config.uk"
 	printf "source \"%s\"\n" "$( "${CMD_READLINK}" -f "${CONFIG_UK}" )" >&7
+
+	CONFIG_UK_D="${CONFIG_UK}.d"
+	if [ -d "${CONFIG_UK_D}" ]; then
+		IFS=$IFS_NL
```
mogasergiu: Any reason you did not use `{ }` for consistency for the `IFS_NL` and `IFS_ORIG` vars?

lib/posix-vfs-fstab/fstab-builtin.h
```diff
@@ -0,0 +1,184 @@
...
+#error "Do not include this header directly"
+#endif /* __UK_POSIX_VFS_FSTAB_H__ */
+
+#if CONFIG_LIBPOSIX_VFS_FSTAB_BUILTIN
+
+#if CONFIG_LIBPOSIX_VFS_FSTAB_BI0
+const struct vfs_fstab_entry fstab_bi0 = {
+#ifdef CONFIG_LIBPOSIX_VFS_FSTAB_BI0_DEV_ARG
+	.dev = CONFIG_LIBPOSIX_VFS_FSTAB_BI0_DEV_ARG,
+#else
```
mogasergiu: `!CONFIG_LIBPOSIX_VFS_FSTAB_BI0_DEV_ARG` and ditto for `endif`? (Ditto for the others in this strucct as well)

lib/posix-vfs-fstab/einitrd.S
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/config.h>
+
+.globl vfs_einitrd_start
```
mogasergiu, line 9: Since there is no `exportsyms.uk` and I assume you don't want these syms exported, consider adding a `localsyms.uk` file.

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+#if CONFIG_LIBPOSIX_VFS_FSTAB_BUILTIN || CONFIG_LIBPOSIX_VFS_FSTAB_FALLBACK
+static
+int fstab_process_internal(const struct vfs_fstab_entry *const fstab[],
+			   size_t len,
+			   const struct ukplat_memregion_desc *initrd0)
+{
+	int mounted = 0;
+
+	for (size_t i = 0; i < len; i++)
+		if (likely(fstab[i])) {
```
mogasergiu: You'll hate me for mentioning it :sweat_smile: but I think an `unlikely` with `continue` and reduced indentation here would look better? But is possible for an entry to be NULL? Did you mean to compare against `""` instead?


(Ditto `dddfstab_process_user`)

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+		if (!opt) {
+			if (!pos)
+				break; /* End of option string */
+			continue; /* Empty option */
+		}
+		if (!strcmp(opt, "mkmp")) {
+			ret |= FSTAB_UKOPT_MKMP;
+			continue;
+		}
+		if (!strcmp(opt, "ifinitrd0")) {
```
mogasergiu: Maybe we could use `strncmp` with `sizeof("hardcoded string here")` instead? (Ditto for `fstab_mount`/`fstab_extract`'s already macronized strings)

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+
+	for (char *pos = ukopts_str;;) {
+		char *opt = uk_nextarg(&pos, FSTAB_UKOPTS_ARGS_SEP);
+
+		if (!opt) {
+			if (!pos)
+				break; /* End of option string */
+			continue; /* Empty option */
+		}
+		if (!strcmp(opt, "mkmp")) {
```
mogasergiu: I think these commands could be macronized instead of being hardcoded here?

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+	else if (!strcmp(e->dev, FSTAB_DEV_EINITRD)) {
+		base = (const void *)vfs_einitrd_start;
+		len = (uintptr_t)&vfs_einitrd_end - (uintptr_t)base;
+	}
+#endif /* CONFIG_LIBPOSIX_VFS_FSTAB_EINITRD */
+	else {
+		uk_pr_crit("Invalid initrd source: %s\n", e->dev);
+		return -EINVAL;
+	}
+
```
mogasergiu, line 71: Would it make sense to have two `UK_ASSERT`s here: one for `base` and one for `len`?

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+#define FSTAB_FSTYPE_EXTRACT "extract"
+
+#define FSTAB_DEV_INITRD0 "initrd0"
+#define FSTAB_DEV_EINITRD "embedded"
+
+/**
+ * INTERNAL. Get the memregion of initrd0.
+ *
+ * @return
+ *  Zero on success, non-zero on failure.
```
mogasergiu: This is documented as if it would return an integer :sweat_smile:

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+	if (!e->path)
+		e->path = "";
+	if (!e->fstype)
+		e->fstype = "";
+
+	if (flags_str && flags_str[0])
+		e->flags = strtol(flags_str, NULL, 0);
+	else
+		e->flags = 0;
+	e->ukopts = fstab_parse_ukopts(ukopts_str);
```
mogasergiu: I think we should have a check/assert somewhere for `ukopts_str`? Either here or in `fstab_parse_ukopts` :thinking:

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+
+	if (!e->dev)
+		e->dev = "";
+	if (!e->path)
+		e->path = "";
+	if (!e->fstype)
+		e->fstype = "";
+
+	if (flags_str && flags_str[0])
+		e->flags = strtol(flags_str, NULL, 0);
```
mogasergiu: `strtol`.. which also sets errno, should we check for underflow/overflow?

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+	/* Handle ukopts: skip if/no initrd; mkmp */
+	if (((e->ukopts & FSTAB_UKOPT_IFINITRD0) && !initrd0) ||
+	    ((e->ukopts & FSTAB_UKOPT_IFNOINITRD0) && initrd0))
+		return 0;
+	if (e->ukopts & FSTAB_UKOPT_MKMP) {
+		r = fstab_mkmp(e->path);
+		if (unlikely(r))
+			uk_pr_crit("Coult not create mount path '%s': %d\n",
+				   e->path, r);
+		/* Attempt to mount even on failure */
```
mogasergiu: Can you also mention the reason for this?

lib/posix-vfs-fstab/fstab.c
```diff
@@ -0,0 +1,353 @@
...
+	UK_ASSERT(e->fstype);
+
+	/* Handle ukopts: skip if/no initrd; mkmp */
+	if (((e->ukopts & FSTAB_UKOPT_IFINITRD0) && !initrd0) ||
+	    ((e->ukopts & FSTAB_UKOPT_IFNOINITRD0) && initrd0))
+		return 0;
+	if (e->ukopts & FSTAB_UKOPT_MKMP) {
+		r = fstab_mkmp(e->path);
+		if (unlikely(r))
+			uk_pr_crit("Coult not create mount path '%s': %d\n",
```
mogasergiu: s/Coult/Could

mogasergiu (changes requested):
A couple of comments/questions.

lib/posix-vfs-fstab/fstab.c
andreittr: I disagree; using `strncmp` with string literals would only harm things here: it requires either duplicating the literal for `sizeof` (which introduces the risk of divergence and hard-to-catch bugs) or macro noise to define the length of every literal we care to compare against.

`strcmp` is IMO fine here because bounds checking is guaranteed by the finite length of the string literal; `strncmp` would add nothing in terms of safety. And if you're worried about performance, any halfway decent compiler will LTO a `strcmp` where one of the strings is a literal to the unrolled equivalent
```c
if (str[0] && str[0] == lit[0] && str[1] && str[1] == lit[1] && ...)
```

lib/posix-vfs-fstab/fstab.c
andreittr: Not at all, I think it would make things more brittle. The code here in fstab is perfectly correct and consistent regardless of their values. How `ukcpio_extract` uses those values passed as args is none of our business here in fstab ([its API docs](https://github.com/unikraft/unikraft/blob/staging/lib/ukcpio/include/uk/cpio.h#L163) say nothing about zero values being forbidden, and [the implementation](https://github.com/unikraft/unikraft/blob/staging/lib/ukcpio/cpio.c#L333) doesn't seem to mind either).

I don't see what benefit we gain from changing previously OK behavior (extracting a 0-length CPIO is a NOP) to "illegal state we must crash the kernel".

lib/posix-vfs-fstab/fstab.c
andreittr: There's not much that fstab can meaningfully do on underflow/overflow except crash, and I'm not sure whether that's the right thing to do here :shrug: . But ignoring the error is probably not kosher either. I'll add some error checking and warnings.

Re: `strtol`, yes, it's the only function that does multi-base conversion without having to roll my own. And its touching errno is irrelevant, as everybody is allowed to set errno, [even functions that succeed](https://www.man7.org/linux/man-pages/man3/errno.3.html#DESCRIPTION).

lib/posix-vfs-fstab/fstab.c
andreittr: What for? A null value is handled perfectly gracefully by `fstab_parse_ukopts` and `uk_nextarg`. No need to crash the kernel out of paranoia.

That said, I see a logic error in how `ukopts_str` is assigned, since both vol and ukopts share the same separator char, the line
```c
ukopts_str = uk_nextarg(&p, FSTAB_VOL_ARGS_SEP);
```
would chop off all but the first ukopt before we got into `fstab_parse_ukopts`. Thus it should simply be
```c
ukopts_str = p;
```
and let `fstab_parse_ukopts` parse the rest until `\0`.

lib/posix-vfs-fstab/fstab.c
andreittr: > But is possible for an entry to be NULL?

[Yes](https://github.com/unikraft/unikraft/pull/1654/files/5f887d1c7641ba7eaeb765b700a2a950ac7de9e9..19e98078083e6a4b54bee75849505f082cf93ecc) although I'm not sure we need a `NULL` sentinel value if we know the array size :thinking: could probably rework this.

edit: and on closer look, user fstabs can have multiple NULL entries at the end, since it's a fixed-size array that may get populated by more or fewer entries at runtime. But there we should probably break the loop at the first NULL :thinking: .

Regardless, I'll remove branch hinting, as it doesn't denote an error condition or anything that would warrant a "cold path".

support/build/config-submenu.sh
andreittr: Force of habit most likely; but on 2nd look I don't like this invasive change to the buildsystem for something that is likely only to be used by libraries that do crazy `Config.uk` wizardry, which I'm hoping there won't be many of.
Will drop this commit in the pass and replace with bespoke `Config.uk` magic in `posix-vfs-fstab` itself.

edit: geez, I realised I had missed to include the `Config.uk.d` directory for fstab entirely :sweat_smile: .

andreittr:
Pushed rebase to staging & latest #1610 .

andreittr:
Pushed addressing comments + more:
- dropped changes to build system; usecase not common enough to warrant it
- added missing builtin & user fstab config dropins (oops)
- added missing library registration in `lib/Makefile.uk` (oops)
- builtin and fallback fstabs have had one layer of indirection removed; they are now flat struct arrays in the kernel image
- misc code improvements

andreittr:
Pushed cosmetic changes to fix checkpatch warnings.

lib/posix-vfs-fstab/user/Config.uk
```diff
@@ -0,0 +1,460 @@
...
+	config LIBPOSIX_VFS_FSTAB_FALLBACK_INITRD
+	bool "RAMfs + CPIO initrd"
+	select LIBUKFS_RAMFS
+	select LIBUKCPIO
+	help
+	  Extract a user-supplied initrd into an empty RAMfs.
+
+	config LIBPOSIX_VFS_FSTAB_FALLBACK_EINITRD
+	bool "RAMfs + compiled-in CPIO initrd"
+	select LIBPOSIX_VFS_FSTAB_EINITRD
```
mogasergiu, line 38: I am not sure how consistent the KConfig style is. Normally you don't indent these (yes i know we have that in some places but there's been changes in the right direction). Additionally, the help text in some places seems to be inconsistent and does not always begin 2 whitespaces indentation on the next line after `help` (e.g. `LIBPOSIX_VFS_FSTAB_FALLBACK_EINITRD_EXTRACT`). The style that I think is commonly used is
```
config NAME
<tab>bool/prompt/etc
<tab>select/depends on/etc
<tab>help
<tab>  <message here>
```
(You can look at `lib/ukboot/Config.uk` for example)

mogasergiu (changes requested):

lib/posix-vfs-fstab/user/Config.uk
andreittr: Good point, I had blindly copied over an older style, will update :+1:

andreittr:
Pushed rebase to current staging & latest #1610 .

andreittr:
Pushed update addressing comment & clarifying fstab's intentional positioning as latest in the rootfs init class.

lib/posix-vfs-fstab/user/Config.uk
```diff
@@ -0,0 +1,460 @@
...
+	select LIBUKCPIO
+	help
+	  Extract a user-supplied initrd into an existing directory.
+
+config LIBPOSIX_VFS_FSTAB_FALLBACK_EINITRD_EXTRACT
+	bool "CPIO compiled-in initrd, extract only"
+	select LIBPOSIX_VFS_FSTAB_EINITRD
+	select LIBUKCPIO
+	help
+		Extract a compiled-in initrd into an existing directory.
```
mogasergiu: Commenting so we don't forget: indent this line properly.

mogasergiu (approved):
LGTM just 1 more comment.

andreittr:
Pushed update adding `exportsyms.uk` file to explicitly express developer intentions, and remove a `localsyms.uk` file that was mistakenly put in the repo root :sweat_smile: .

andreittr:
Pushed update addressing final comment.

andreittr:

michpappas:

PR #1657, opened 2025-06-10
lib/posix-process: Fix stack corruption in signal delivery

michpappas:
### Prerequisite checklist

### Base target

### Description of changes

Rework signal delivery to fix a corruption bug on the auxstack triggered when the exception handling trampoline is entered while handling a signal.

### DIFF
--- a/lib/posix-process/arch/arm64/signal.S
+++ b/lib/posix-process/arch/arm64/signal.S
@@ -5,22 +5,91 @@
  */
 
 #include <uk/asm.h>
+#include <uk/arch/lcpu.h>
+#include <uk/arch/ctx.h>
+#include <uk/plat/common/lcpu.h>
 
-/* Switch to application stack and call signal handler.
- * Switch back to UK stack before return.
+#include "../../signal/signal.h"
+
+/* Switch to application's context and call signal handler.
  *
- * void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
- *                                              ucontext_t *ctx, void *handler,
- *                                              void *sp);
+ * Before executing the handler, align the auxsp and update
+ * auxspcb->current_frame so that any kernel entry that happens
+ * while the signal handler executes, starts off a clean state.
  *
- * Notice: Parameters to handler already saved to registers
- *         via this function's first four parameters.
+ * @param handler_ctx    ptr to struct pprocess_signal_handler_ctx (x0)
+ * @param execenv        ptr to application's execenv (x1)
  */
-ENTRY(pprocess_signal_call_handler_with_stack)
-	str	x19, [sp], #16	/* save x19 before we corrupt */
-	mov	x19, sp		/* save uk sp */
-	mov	sp, x4		/* switch to handler sp */
-	blr	x3		/* call handler */
-	mov	sp, x19		/* switch back to uk sp */
-	ldr	x19, [sp], #16	/* restore x19 */
+ENTRY(pprocess_signal_jmp_handler)
+	stp	x29, x30, [sp, #-16]!
+
+	stp	x19, x20, [sp, #-16]!
+	stp	x21, x22, [sp, #-16]!
+	stp	x23, x24, [sp, #-16]!
+	mov	x29, sp /* save pre-aligned sp */
+
+	mov	x19, x0 /* handler_ctx */
+	mov	x20, x1 /* execenv */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Save uk sysctx                                                    *
+	 *───────────────────────────────────────────────────────────────────*/
+	sub sp, sp, #UKARCH_SYSCTX_SIZE     /* reserve space for ukarch sysctx */
+	and sp, sp, #~0xf                         /* mare sure we're aligned */
+	mov x0, sp                                /* prepare args */
+	bl	ukarch_sysctx_store		/* save uk sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Align auxspcb->curr_fp.
+	 *───────────────────────────────────────────────────────────────────*/
+	mrs	x10, tpidr_el1			/* get auxspcb */
+	add	x10, x10, #(LCPU_AUXSP_OFFSET - (UKARCH_AUXSPCB_SIZE - UKARCH_AUXSPCB_OFFSETOF_CURR_FP))
+	ldr	x11, [x10]			/* save curr_fp and its addr */
+	stp	x11, x10, [sp, #-16]!
+
+	/* Save uk sp *before* we align auxsp for nested calls.
+	 * We will be restoring the saved value upon handler return.
+	 */
+	mov	x21, sp
+
+	/* Align auxsp */
+	and	x11, sp, #~(UKARCH_AUXSP_ALIGN - 1)
+	str	x11, [x10]
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Switch to the application's sysctx                                *
+	 *───────────────────────────────────────────────────────────────────*/
+	add	x0, x20, #UKARCH_EXECENV_OFFSETOF_SYSCTX
+	bl	ukarch_sysctx_load			/* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	ldr	x22, [x19, #PPROCESS_SIGHNDL_CTX_FN_OFFS]	/* handler_fn */
+	ldr	x23, [x19, #PPROCESS_SIGHNDL_CTX_SP_OFFS]	/* handler_sp */
+	ldr	w0, [x19, #PPROCESS_SIGHNDL_CTX_SN_OFFS]	/* signum */
+	ldr	x1, [x19, #PPROCESS_SIGHNDL_CTX_SI_OFFS]	/* siginfo_t */
+	ldr	x2, [x19, #PPROCESS_SIGHNDL_CTX_UC_OFFS]	/* ucontext_t */
+
+	mov	sp, x23			/* switch to handler stack */
+	blr	x22			/* call handler */
+	mov	sp, x21			/* switch back to uk sp */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore auxspcb->curr_fp                                          *
+	 *───────────────────────────────────────────────────────────────────*/
+	ldp	x11, x10, [sp], #16	/* pop auxpcb->curr_fp and &auxpcb->curr_fp */
+	str	x11, [x10]		/* restore old value */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore uk sysctx                                                 *
+	 *───────────────────────────────────────────────────────────────────*/
+	mov	x0, sp				/* prepare args */
+	bl	ukarch_sysctx_load		/* restore saved sysctx */
+
+	mov	sp, x29				/* restore pre-aligned sp */
+	ldp	x23, x24, [sp], #16
+	ldp	x21, x22, [sp], #16
+	ldp	x19, x20, [sp], #16
+	ldp	x29, x30, [sp], #16
 	ret
--- a/lib/posix-process/arch/x86_64/signal.S
+++ b/lib/posix-process/arch/x86_64/signal.S
@@ -5,22 +5,90 @@
  */
 
 #include <uk/asm.h>
+#include <uk/arch/lcpu.h>
+#include <uk/arch/ctx.h>
+#include <uk/plat/common/lcpu.h>
 
-/* Switch to application stack and call signal handler.
- * Switch back to UK stack before return.
+#include "../../signal/signal.h"
+
+/* Switch to application's context and call signal handler.
  *
- * void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
- *                                              ucontext_t *ctx, void *handler,
- *                                              void *sp);
+ * Before executing the handler, align the auxsp and update
+ * auxspcb->current_frame so that any kernel entry that happens
+ * while the signal handler executes, starts off a clean state.
  *
- * Notice: Parameters to handler already saved to registers
- *         via this function's first four parameters.
+ * @param handler_ctx    ptr to struct pprocess_signal_handler_ctx (rdi)
+ * @param execenv        ptr to application's execenv (rsi)
  */
-ENTRY(pprocess_signal_call_handler_with_stack)
-	push	%rbp		/* save rbp before we corrupt */
-	movq    %rsp, %rbp	/* save uk sp */
-	movq    %r8, %rsp	/* switch to handler sp */
-	call    *%rcx		/* call handler */
-	movq    %rbp, %rsp	/* switch back to uk sp */
-	pop	%rbp		/* restore rbp */
+ENTRY(pprocess_signal_jmp_handler)
+	pushq	%rbp
+	pushq	%rbx
+	pushq	%r12
+	pushq	%r13
+	movq	%rsp, %rbp /* save pre-aligned rsp */
+
+	movq	%rdi, %r12 /* handler_ctx */
+	movq	%rsi, %r13 /* execenv */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Save uk sysctx                                                    *
+	 *───────────────────────────────────────────────────────────────────*/
+	subq	$(UKARCH_SYSCTX_SIZE), %rsp /* allocate space for struct */
+	andq	$~0xf, %rsp		    /* align rsp */
+	movq	%rsp, %rdi		    /* prepare args */
+	call	ukarch_sysctx_store	    /* save uk sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Align auxspcb->curr_fp.
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	%gs:LCPU_AUXSP_OFFSET, %r10	/* get auxpcb */
+	subq	$(UKARCH_AUXSPCB_SIZE - UKARCH_AUXSPCB_OFFSETOF_CURR_FP), %r10
+	pushq	(%r10)				/* save curr_fp */
+	pushq	%r10				/* also save its addr */
+
+	/* Save uk rsp *before* we align auxsp for nested calls.
+	 * We will be restoring the saved value upon handler return.
+	 */
+	movq	%rsp, %rbx
+
+	movq	%rsp, %r11		/* now proceed to align auxsp */
+	andq	$~(UKARCH_AUXSP_ALIGN - 1), %r11
+	movq	%r11, (%r10)		/* write aligned address */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Switch to the application's sysctx                                *
+	 *───────────────────────────────────────────────────────────────────*/
+	leaq	UKARCH_EXECENV_OFFSETOF_SYSCTX(%r13), %rdi  /* &ee->sysctx in rdi */
+	call	ukarch_sysctx_load			    /* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12), %r10	/* handler_fn */
+	movq	PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %r11	/* handler_sp */
+	movl	PPROCESS_SIGHNDL_CTX_SN_OFFS(%r12), %edi	/* signum */
+	movq	PPROCESS_SIGHNDL_CTX_SI_OFFS(%r12), %rsi	/* siginfo_t */
+	movq	PPROCESS_SIGHNDL_CTX_UC_OFFS(%r12), %rdx	/* ucontext_t */
+
+	movq	%r11, %rsp	/* switch to handler stack */
+	call	*%r10		/* call handler */
+	movq	%rbx, %rsp	/* switch back to uk sp */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore auxspcb->curr_fp                                          *
+	 *───────────────────────────────────────────────────────────────────*/
+	popq	%r10		/* pop &auxpcb->curr_fp */
+	popq	(%r10)		/* pop &auxpcb->curr_fp */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore uk sysctx                                                 *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	%rsp, %rdi		    /* prepare args */
+	call	ukarch_sysctx_load	    /* restore saved sysctx */
+
+	movq	%rbp, %rsp  /* restore pre-aligned rsp */
+	popq	%r13
+	popq	%r12
+	popq	%rbx
+	popq	%rbp
 	ret
--- a/lib/posix-process/signal/deliver.c
+++ b/lib/posix-process/signal/deliver.c
@@ -4,7 +4,9 @@
  * You may not use this file except in compliance with the License.
  */
 
+#include <uk/arch/ctx.h>
 #include <uk/essentials.h>
+#include <uk/plat/config.h>
 #include <uk/prio.h>
 #include <uk/process.h>
 #include <uk/syscall.h>
@@ -14,10 +16,30 @@
 #include "signal.h"
 #include "siginfo.h"
 
-/* asm */
-void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
-					     ucontext_t *ctx, void *handler,
-					     void *sp);
+struct pprocess_signal_handler_ctx {
+	void *handler_fn;
+	void *handler_sp;
+	siginfo_t *siginfo;
+	ucontext_t *ucontext;
+	int signo;
+} __packed;
+
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, handler_fn) ==
+	    PPROCESS_SIGHNDL_CTX_FN_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, handler_sp) ==
+	    PPROCESS_SIGHNDL_CTX_SP_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, signo) ==
+	    PPROCESS_SIGHNDL_CTX_SN_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, siginfo) ==
+	    PPROCESS_SIGHNDL_CTX_SI_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, ucontext) ==
+	    PPROCESS_SIGHNDL_CTX_UC_OFFS);
+
+/* Notice: We implement this in pure asm, as there is a number of issues
+ *         with correctly preserving context in inline asm.
+ */
+void pprocess_signal_jmp_handler(struct pprocess_signal_handler_ctx *h,
+				 struct ukarch_execenv *execenv);
 
 static void uk_sigact_term(int __unused sig)
 {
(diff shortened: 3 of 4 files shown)


lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+
+	movq	%rsp, %r11		/* now proceed to align auxsp */
+	andq	$~(UKARCH_AUXSP_ALIGN - 1), %r11
+	movq	%r11, (%r10)		/* write aligned address */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Switch to the application's sysctx                                *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	%r13, %rdi
+	addq	$(UKARCH_EXECENV_OFFSETOF_SYSCTX), %rdi /* &ee->sysctx in rdi */
```
andreittr: The above 2 instructions could be more succintly (and idiomatically x86) expressed as
```asm
lea UKARCH_EXECENV_OFFSETOF_SYSCTX(%r13), %rdi
```
which encodes to as many bytes as just the `addq` instruction :smile: .

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+	popq	%r10		/* pop &auxpcb->curr_fp */
+	popq	%r11		/* pop auxpcb->curr_fp */
+	movq	%r11, (%r10)	/* restore old value */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore uk sysctx                                                 *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	%rsp, %rdi		    /* prepare args */
+	call	ukarch_sysctx_load	    /* restore saved sysctx */
+	addq	$(UKARCH_SYSCTX_SIZE), %rsp /* restore stack for saved syscx */
```
andreittr: This final stack add is redundant, as we restore the post-prologue `rsp` from `rbp` one instruction later.

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12), %r14	/* handler_fn */
+	movq	PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %r15	/* handler_sp */
+	movq	PPROCESS_SIGHNDL_CTX_SN_OFFS(%r12), %rdi	/* signum */
+	movq	PPROCESS_SIGHNDL_CTX_SI_OFFS(%r12), %rsi	/* siginfo_t */
+	movq	PPROCESS_SIGHNDL_CTX_UC_OFFS(%r12), %rdx	/* ucontext_t */
+
+	movq	%r15, %rsp	/* switch to handler stack */
+	call	*%r14		/* call handler */
```
andreittr: AFAICT you use the non-volatile r14 and r15 only here, which demands you save & restore them in the prologue/epilogue. You could remove these two clobbers and write
```asm
movq PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %rsp
call *PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12)
```
or, if that's too CISCey and unclear, you can use the available volatile r10 and r11
```asm
movq PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12), %r10	/* handler_fn */
movq PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %r11	/* handler_sp */
...
movq %r11, %rsp
call *%r10
```

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+	movq	%rbx, %rsp	/* switch back to uk sp */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore auxspcb->curr_fp                                          *
+	 *───────────────────────────────────────────────────────────────────*/
+	popq	%r10		/* pop &auxpcb->curr_fp */
+	popq	%r11		/* pop auxpcb->curr_fp */
+	movq	%r11, (%r10)	/* restore old value */
```
andreittr: You can write
```asm
popq (%r10)
```
to better align with the `pushq (%r10)` earlier (and save 2 bytes of encoding, plus potentially [lower latency](https://www.agner.org/optimize/instruction_tables.pdf))

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+	movq	%r13, %rdi
+	addq	$(UKARCH_EXECENV_OFFSETOF_SYSCTX), %rdi /* &ee->sysctx in rdi */
+	call	ukarch_sysctx_load			/* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12), %r14	/* handler_fn */
+	movq	PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %r15	/* handler_sp */
+	movq	PPROCESS_SIGHNDL_CTX_SN_OFFS(%r12), %rdi	/* signum */
```
andreittr: The `signo` field is defined as `int` in `struct pprocess_signal_handler_ctx`, and similarly the sighandler expects an `int` as its 1st arg, thus the read would correctly be
```asm
movl PPROCESS_SIGHNDL_CTX_SN_OFFS(%r12), %edi
```
This goes especially because the `pad` field (immediately following `signo`) never seems to be initialized by `handle_self`, and you don't want garbage data loaded into the upper half of `%rdi`.

IIRC the sysv ABI, if the called function expects an `int` then it only uses `%edi`, so we don't need to do any `movslq` shenaningans to sign-extend to all of `%rdi`.

lib/posix-process/signal/deliver.c
```diff
@@ -14,10 +16,34 @@
 #include "signal.h"
 #include "siginfo.h"
 
-/* asm */
-void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
-					     ucontext_t *ctx, void *handler,
-					     void *sp);
+struct pprocess_signal_handler_ctx {
+	void *handler_fn;
```
andreittr, line 20: I noticed this is `void *` while `struct kern_sigaction` (where this is assigned from) has a typed `void (*ks_handler)(int)` field.

AFAICT this field is used for both `void sighandler(int)` as well as `void sigaction(int, siginfo_t *, ucontext_t *)`. Leaving aside that the _interpretation_ of `handler_fn` as one or the other is done entirely in assembly, I see the need for `void *` as a generic fn pointer (although, we should really be using a fn ptr type union to be 100% correct).

My (somewhat out of scope) question then is: is there a reason I'm not seeing for `ks_handler` to be typed that way to begin with? Would it not make sense to have it as `void *` (or, better yet, a typesafe union)?

lib/posix-process/signal/deliver.c
```diff
@@ -14,10 +16,34 @@
...
+		       PPROCESS_SIGHNDL_CTX_SP_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, signo) ==
+		       PPROCESS_SIGHNDL_CTX_SN_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, siginfo) ==
+		       PPROCESS_SIGHNDL_CTX_SI_OFFS);
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, ucontext) ==
+		       PPROCESS_SIGHNDL_CTX_UC_OFFS);
+
+/* Execute system fault handlers after pf */
+int signal_syserr_handler_prio = CONFIG_LIBUKVMEM_PAGEFAULT_HANDLER_PRIO + 1;
```
andreittr: `static const` and ALL_CAPS, as it seems to be a module-level constant.

Then again, I don't see this used (or declared `extern`) anywhere, what is its purpose?

lib/posix-process/signal/signal.h
```diff
@@ -24,9 +25,18 @@
...
+#endif
 
 #define SIG_ARRAY_COUNT		_NSIG /* SIGRTMAX + 1 */
 
+/* struct pprocess_signal_handler_ctx */
+#define	PPROCESS_SIGHNDL_CTX_FN_OFFS	 0/* handler_fn */
+#define	PPROCESS_SIGHNDL_CTX_SP_OFFS	 8/* handler_sp */
+#define	PPROCESS_SIGHNDL_CTX_SN_OFFS	16/* signum */
+#define	PPROCESS_SIGHNDL_CTX_SI_OFFS	24/* siginfo_t */
+#define	PPROCESS_SIGHNDL_CTX_UC_OFFS	32/* ucontext_t */
```
andreittr: Since this `signal.h` is an internal header, would it not be better to move the definition of `struct pprocess_signal_handler_ctx` here? It would help give a lot more context to these defines. (you can keep the `CT_ASSERT()`s in just one source file for compilation performance if you wish)

Similarly, the declaration of
```c
void pprocess_signal_jmp_handler(struct pprocess_signal_handler_ctx *h,
				 struct ukarch_execenv *execenv);
```
could be moved here.

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	mov	x0, x20
+	add	x0, x0, #UKARCH_EXECENV_OFFSETOF_SYSCTX	/* &ee->sysctx in x0 */
+	bl	ukarch_sysctx_load			/* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	ldr	x22, [x19, #PPROCESS_SIGHNDL_CTX_FN_OFFS]	/* handler_fn */
+	ldr	x23, [x19, #PPROCESS_SIGHNDL_CTX_SP_OFFS]	/* handler_sp */
+	ldr	x0, [x19, #PPROCESS_SIGHNDL_CTX_SN_OFFS]	/* signum */
```
andreittr: (Analogous to x86 comment about `int signum`)
I _think_ I know enough arm asm to tell this should be
```asm
ldr w0, [x19, #PPROCESS_SIGHNDL_CTX_SN_OFFS]	/* signum */
```
(assuming the handler doesn't expect any sign-extension shenanigans on `x0`)

andreittr (commented):
Some minor-to-medium comments. I gave only a brief look at the arm64 asm, as that's not my forte.

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,97 @@
...
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore uk sysctx                                                 *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	%rsp, %rdi		    /* prepare args */
+	call	ukarch_sysctx_load	    /* restore saved sysctx */
+	addq	$(UKARCH_SYSCTX_SIZE), %rsp /* restore stack for saved syscx */
+
+	movq	%rbp, %rsp  /* restore pre-aligned rsp */
+	pop	%r15
```
mogasergiu: Not that it would impact runtime, just a nit, but `popq` here (and `pushq` at the beginning) to be consistent with the other assembly code we have? Which reminds me... we should maybe add CFI here as well, or remember to add it later if not now.

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	mov	x19, x0 /* handler_ctx */
+	mov	x20, x1 /* execenv */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Save uk sysctx                                                    *
+	 *───────────────────────────────────────────────────────────────────*/
+	sub	x23, sp, #UKARCH_SYSCTX_SIZE	/* reserve space for ukarch sysctx */
+	and	x23, x23, #~0xf			/* make sure we're aligned */
+	mov	sp, x23				/* update sp */
+	mov	x0, sp				/* prepare args */
```
mogasergiu: Do we need to touch `x23` here? :thinking: 
Would this be equivalent?
```
sub sp, sp, #UKARCH_SYSCTX_SIZE     /* reserve space for ukarch sysctx */
and sp, sp, #~0xf                         /* mare sure we're aligned */
mov x0, sp                                /* prepare args */
```


Also, feel free to disregard, but to use maybe even one less instructions we could do
```
sub sp, sp, #((UKARCH_SYSCTX_SIZE + 15) & ~0xf)
mov x0, sp
```
But this assume `sp` is 16-byte aligned (as it should be and is the case) and looks a bit ugly, idk. Same could go for x86 as well and we would save an instruction.

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	bl	ukarch_sysctx_store		/* save uk sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Align auxspcb->curr_fp.
+	 *───────────────────────────────────────────────────────────────────*/
+	mrs	x10, tpidr_el1			/* get auxspcb */
+	add	x10, x10, #LCPU_AUXSP_OFFSET
+	sub	x10, x10, #(UKARCH_AUXSPCB_SIZE - UKARCH_AUXSPCB_OFFSETOF_CURR_FP)
```
mogasergiu: I believe you could combine these 2 above into:
```
add	x10, x10, #(LCPU_AUXSP_OFFSET - (UKARCH_AUXSPCB_SIZE - UKARCH_AUXSPCB_OFFSETOF_CURR_FP))
```

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	ldr	x11, [x10]			/* save curr_fp and its addr */
+	stp	x11, x10, [sp, #-16]!
+
+	/* Save uk sp *before* we align auxsp for nested calls.
+	 * We will be restoring the saved value upon handler return.
+	 */
+	mov	x21, sp
+
+	mov	x11, sp				/* now proceed to align auxsp */
+	and	x11, x11, #~(UKARCH_AUXSP_ALIGN - 1)
```
mogasergiu: Ditto for these two:
```
and	x11, sp, #~(UKARCH_AUXSP_ALIGN - 1)	/* align auxsp directly */
```

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	str	x11, [x10]			/* write aligned address */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Switch to the application's sysctx                                *
+	 *───────────────────────────────────────────────────────────────────*/
+	mov	x0, x20
+	add	x0, x0, #UKARCH_EXECENV_OFFSETOF_SYSCTX	/* &ee->sysctx in x0 */
```
mogasergiu: Ditto
```
add	x0, x20, #UKARCH_EXECENV_OFFSETOF_SYSCTX
```

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	bl	ukarch_sysctx_load			/* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	ldr	x22, [x19, #PPROCESS_SIGHNDL_CTX_FN_OFFS]	/* handler_fn */
```
mogasergiu, line 68: I am wondering if we could do `ldp` for `x22`/`x23` and `x1`/`x2`... or should we leave as is in case the placements/values of these offsets change?

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,96 @@
...
+	 *───────────────────────────────────────────────────────────────────*/
+	ldp	x11, x10, [sp], #16	/* pop auxpcb->curr_fp and &auxpcb->curr_fp */
+	str	x11, [x10]		/* restore old value */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Restore uk sysctx                                                 *
+	 *───────────────────────────────────────────────────────────────────*/
+	mov	x0, sp				/* prepare args */
+	bl	ukarch_sysctx_load		/* restore saved sysctx */
+	add	sp, sp, #UKARCH_SYSCTX_SIZE	/* restore stack for saved syscx */
```
mogasergiu: Just as Andrei mentioned on the x86 side, I believe this add is redundant

lib/posix-process/signal/deliver.c
```diff
@@ -67,41 +93,22 @@ bool pprocess_signal_is_deliverable(struct posix_thread *pthread, int signum)
...
-	 */
-	pprocess_signal_call_handler_with_stack(signum, si, ctx, handler, sp);
-
-	ukarch_sysctx_load(&uk_sysctx);
-}
-
 static void handle_self(struct uk_signal *sig, const struct kern_sigaction *ks,
 			struct ukarch_execenv *execenv)
 {
+	struct pprocess_signal_handler_ctx handler_ctx;
```
mogasergiu, line 95: I think we may need to initialize the `pad` field somewhere (if we decide to keep it).
(Btw, not part of this PR, I know :rofl:, but do you remember why the `ucontext` below has a `__maybe_unused`?)

lib/posix-process/signal/deliver.c
```diff
@@ -133,43 +139,24 @@ static void handle_self(struct uk_signal *sig, const struct kern_sigaction *ks,
...
-			    ks->ks_handler, (void *)ulsp);
+		pprocess_signal_jmp_handler(&handler_ctx, execenv);
 		pprocess_signal_arch_get_ucontext(&ucontext, execenv);
 	} else {
-		jmp_handler(execenv, sig->siginfo.si_signo,
-			    NULL, NULL, ks->ks_handler,
-			    (void *)ulsp);
-	}
+		handler_ctx.siginfo = NULL;
+		handler_ctx.ucontext = NULL;
```
mogasergiu: Btw, should we use `__NULL` instead? I see we already use non-libc types here.
(Also nit but it looks like line 147 is empty, surprised checkpatch did not catch it. Seems that the ARM version of `pprocess_signal_jmp_handler` also has it above the first instruction)

lib/posix-process/signal/signal.h
```diff
@@ -24,9 +25,18 @@
 #if CONFIG_LIBPOSIX_PROCESS_SIGNALFD
 #include "signal_file.h"
 #endif /* CONFIG_LIBPOSIX_PROCESS_SIGNALFD */
+#endif
```
mogasergiu: I think you forgot the `/* !__ASSEMBLY__ */`.

lib/posix-process/signal/deliver.c
```diff
@@ -14,10 +16,34 @@
...
 
-/* asm */
-void pprocess_signal_call_handler_with_stack(int signum, siginfo_t *si,
-					     ucontext_t *ctx, void *handler,
-					     void *sp);
+struct pprocess_signal_handler_ctx {
+	void *handler_fn;
+	void *handler_sp;
+	int signo;
+	char pad[4];
```
mogasergiu: Would it be a bad idea to get rid of the `pad` by moving `signo` as the last member? This way I think we still keep necessary alignments.

mogasergiu (changes requested):
LGTM! Though I think we should maybe reword the commit message as it would make it look as if el trampo got merged before this. Probably should be reworded to claim fixing auxiliary stack consistency with **potential** :shushing_face: surprise handlers.

lib/posix-process/arch/x86_64/signal.S
michpappas: I'll change to `pushq` / `popq`

lib/posix-process/arch/arm64/signal.S
michpappas: I'll change to the 1st one

lib/posix-process/arch/arm64/signal.S
michpappas: You never know, so I'd rather stick to the failsafe choice and avoid making assumptions.

lib/posix-process/signal/deliver.c
michpappas: Good idea. For consistency I will also update the other two instances in the same file, although not in the context of this PR.

lib/posix-process/signal/deliver.c
michpappas: >  (Btw, not part of this PR, I know 🤣, but do you remember why the ucontext below has a __maybe_unused?)

Most likely from a previous state of the code, won't dig as it's no longer needed to be there, so I'll remove it now .

lib/posix-process/signal/deliver.c
michpappas: I don't remember why I didn't place it last in the first place, but I'll move it now as it seems to work fine.

lib/posix-process/arch/x86_64/signal.S
michpappas: Sure I'll go with the second mostly because it looks cleaner when loading the rest of the args into regs.

lib/posix-process/arch/x86_64/signal.S
michpappas: Moved `signo` to the end and got rid of padding. Also updated to use `%edi`.

lib/posix-process/signal/deliver.c
michpappas: Relic from a past revision, will remove.

lib/posix-process/signal/signal.h
michpappas: I consider that too. On the other hand the struct is only used in `deliver.c`, and if I moved its declaration to the private header I'd have to move it way below together with other struct declarations, so I don't see much difference. I think I'll keep it as is for now...

lib/posix-process/signal/deliver.c
michpappas: > I noticed this is void * while struct kern_sigaction (where this is assigned from) has a typed void (*ks_handler) (int) field.
> 
> AFAICT this field is used for both void sighandler(int) as well as void sigaction(int, siginfo_t *, ucontext_t *).  Leaving aside that the interpretation of handler_fn as one or the other is done entirely in assembly, I see the need for void * as a generic fn pointer (although, we should really be using a fn ptr type union to be 100% correct).

It actually shouldn't matter. What happens is that in `handle_self()` we either initialize the 2nd paramter or not, depending on whether `SA_SIGINFO` is set in flags. Then in asm we make no distinction whatsoever, just copy the value from the struct to `%rsi` / `x1. From there on, it's the handler that knows how many paramters to expect, depending on what the syscall it used to register itself 🙃 

> My (somewhat out of scope) question then is: is there a reason I'm not seeing for ks_handler to be typed that way to begin with? Would it not make sense to have it as void * (or, better yet, a typesafe union)?

Most likely when I started writing the implementation not knowing where this will end up I declared it properly, but since we can't enforce type safety there either, I'll change that to `void *` to prevent future confusion or doubt.

lib/posix-process/arch/x86_64/signal.S
andreittr: @mogasergiu fun fact: `push` is somewhat unique in that it does _not_ take a `REX.W` prefix to become 64-bit. There's no `pushl` or `pushq` _really_, just `push`.
Instead it takes the current operating mode to determine its size. So the same bytes mean different things in long mode vs protected mode:
```as
a.out:     file format elf64-x86-64

   0:	50                   	push   %rax
```
```as
a.out:     file format elf32-i386

   0:	50                   	push   %eax
```

What it _does_ accept is the `66` prefix to become 16-bit, because at this point _why not?!_. Which means that `pushw` is indeed a thing, and 16-bit and machine-word are the only two sizes you can ever push on the stack. Compare with a more normal instruction:
```as
   0:	48 ff c0             	inc    %rax
   3:	ff c0                	inc    %eax
   5:	66 ff c0             	inc    %ax
   8:	fe c0                	inc    %al
   a:	50                   	push   %rax
   b:	66 50                	push   %ax
```

_Isn't x86 just wonderful?!_

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,90 @@
...
+	 *───────────────────────────────────────────────────────────────────*/
+	lea UKARCH_EXECENV_OFFSETOF_SYSCTX(%r13), %rdi	/* &ee->sysctx in rdi */
+	call	ukarch_sysctx_load			/* load app sysctx */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Call signal handler                                               *
+	 *───────────────────────────────────────────────────────────────────*/
+	movq	PPROCESS_SIGHNDL_CTX_FN_OFFS(%r12), %r10	/* handler_fn */
+	movq	PPROCESS_SIGHNDL_CTX_SP_OFFS(%r12), %r11	/* handler_sp */
+	movl	PPROCESS_SIGHNDL_CTX_SN_OFFS(%r12), %edi	/* signum */
```
andreittr, line 69: Since `signum` is at the end of the struct now, should we make it the last load instruction? Not 100% sure what memory access patterns modern x86 CPUs are best at, but it's a safe assumption that "sequential" is one of them.

Similar argument to arm asm, though there I have even less of a clue re: access patterns.

andreittr (commented):
Looks good, thanks for the changes! One small nitpick though

lib/posix-process/arch/x86_64/signal.S
mogasergiu: Oh yeah it is the most intuitive architecture in the world! :earth_africa: :earth_americas: :earth_asia: :man_astronaut:

lib/posix-process/arch/arm64/signal.S
```diff
@@ -5,22 +5,91 @@
...
  */
-ENTRY(pprocess_signal_call_handler_with_stack)
-	str	x19, [sp], #16	/* save x19 before we corrupt */
-	mov	x19, sp		/* save uk sp */
-	mov	sp, x4		/* switch to handler sp */
-	blr	x3		/* call handler */
-	mov	sp, x19		/* switch back to uk sp */
-	ldr	x19, [sp], #16	/* restore x19 */
+ENTRY(pprocess_signal_jmp_handler)
+	stp	x29, x30, [sp, #-16]!
```
mogasergiu, line 24: Can we add a 1-line comment about store/restore-ing the callee-saved registers here and on x86?
(Also on ARM for the above for preserving frame/link register)

lib/posix-process/arch/x86_64/signal.S
```diff
@@ -5,22 +5,90 @@
...
+	movq	%rsp, %rbx
+
+	movq	%rsp, %r11		/* now proceed to align auxsp */
+	andq	$~(UKARCH_AUXSP_ALIGN - 1), %r11
+	movq	%r11, (%r10)		/* write aligned address */
+
+	/*───────────────────────────────────────────────────────────────────*
+	 * Switch to the application's sysctx                                *
+	 *───────────────────────────────────────────────────────────────────*/
+	lea UKARCH_EXECENV_OFFSETOF_SYSCTX(%r13), %rdi	/* &ee->sysctx in rdi */
```
mogasergiu: nit, `leaq`?

lib/posix-process/signal/deliver.c
```diff
@@ -14,10 +16,30 @@
...
+struct pprocess_signal_handler_ctx {
+	void *handler_fn;
+	void *handler_sp;
+	siginfo_t *siginfo;
+	ucontext_t *ucontext;
+	int signo;
+} __packed;
+
+UK_CTASSERT(__offsetof(struct pprocess_signal_handler_ctx, handler_fn) ==
+		       PPROCESS_SIGHNDL_CTX_FN_OFFS);
```
mogasergiu: I think you aligned these macros to the wrong paranthesis? :thinking:

mogasergiu (approved):
LGTM. Just some nits.

lib/posix-process/arch/x86_64/signal.S
michpappas: Hmmm I'm not a fan of changes motivated by micro-optimizations, as they create noise without any practical benefit, for many different reasons. But since you brought it up: I considered it too, but I think readability is more important (the order followed currently is that of parameters, as dictated by the calling convention). When it comes to caching, it depends on whether `signum` falls into the same cache line as the other parameters, but also on other microarchitectural properties that may vary from system to system anyway. Moreover, this code is not on some hot path anyway, so if we consider micro optimizations that should happen after profiling.

lib/posix-process/arch/arm64/signal.S
michpappas: A comment that we save and restore callee saved registers? I think it's not necessary.

mogasergiu (approved):
LGTM. Will approve/merge once @andreittr also gives his ack.

lib/posix-process/arch/x86_64/signal.S
andreittr: I hadn't noticed that the loads are in argument order; I had always thought they were written in `struct pprocess_signal_handler_ctx` order instead, and the ordering was mistakenly left over when `signum` got moved. Carry on then :+1: .

andreittr (approved):
Looks good, thanks!

mogasergiu (approved):
Merging... :robot: 

PR #1671, opened 2025-07-03
lib/ukfile: Rework pollq implementation to use waitq

andreittr:
### Description of changes

This changeset reworks the ukfile pollqueue implementation to use standard waitqueues instead of bespoke waiting code.
During this pass, unused parts of the pollqueue API were trimmed off and/or brought more in-line with other standard synchronization primitives.
The waitqueue interface has been expanded with a `cookie` field freely available to callers, in order to accomodate pollq's more complex use-case.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukfile/include/uk/file/pollqueue.h
+++ b/lib/ukfile/include/uk/file/pollqueue.h
@@ -569,14 +569,12 @@ out:
  *
  * @param q Target queue.
  * @param set Events to set.
- * @param n Maximum number of threads to wake up. If < 0 wake up all threads.
- *   Chained updates have their own defined notification semantics and may
- *   notify more threads than specified in `n`.
+ * @param one If zero, notify all waiting threads, if non-zero notify at most 1.
  *
  * @return
  *   The previous event set.
  */
-uk_pollevent uk_pollq_set_n(struct uk_pollq *q, uk_pollevent set, int n);
+uk_pollevent uk_pollq_set_n(struct uk_pollq *q, uk_pollevent set, int one);
 
 /**
  * Update events, clearing those in `clr`.
@@ -598,39 +596,15 @@ uk_pollevent uk_pollq_clear(struct uk_pollq *q, uk_pollevent clr);
  *
  * @param q Target queue.
  * @param val New event set.
- * @param n Maximum number of threads to wake up. If < 0 wake up all threads.
- *   Chained updates have their own defined notification semantics and may
- *   notify more threads than specified in `n`
+ * @param one If zero, notify all waiting threads, if non-zero notify at most 1.
  *
  * @return
  *   The previous event set.
  */
-uk_pollevent uk_pollq_assign_n(struct uk_pollq *q, uk_pollevent val, int n);
+uk_pollevent uk_pollq_assign_n(struct uk_pollq *q, uk_pollevent val, int one);
 
-#define UK_POLLQ_NOTIFY_ALL -1
+#define uk_pollq_set(q, s) uk_pollq_set_n(q, s, 0)
 
-/**
- * Update events, setting those in `set` and handling notifications.
- *
- * @param q Target queue.
- * @param set Events to set.
- *
- * @return
- *   The previous event set.
- */
-#define uk_pollq_set(q, s) uk_pollq_set_n(q, s, UK_POLLQ_NOTIFY_ALL)
-
-/**
- * Replace the events in `q` with `val` and handle notifications.
- *
- * Only available on level-triggered queues.
- *
- * @param q Target queue.
- * @param val New event set.
- *
- * @return
- *   The previous event set.
- */
-#define uk_pollq_assign(q, s) uk_pollq_assign_n(q, s, UK_POLLQ_NOTIFY_ALL)
+#define uk_pollq_assign(q, s) uk_pollq_assign_n(q, s, 0)
 
 #endif /* __UKFILE_POLLQUEUE_H__ */
--- a/lib/ukfile/include/uk/file/pollqueue.h
+++ b/lib/ukfile/include/uk/file/pollqueue.h
@@ -171,23 +171,23 @@ struct uk_pollq {
 };
 
 /*
- * Pollqueues come in two varieties: edge- and level-notified.
- * Edge-notified queues require drivers to only notify rising edges of events,
- * while providing a callback for fetching instantaneous levels.
- * Level-notified queues require drivers to notify both rising and falling edges
+ * Pollqueues come in two varieties: managed and polled.
+ * Polled queues require drivers to only notify rising edges of events, while
+ * providing a callback for fetching instantaneous levels (.poll).
+ * Managed queues require drivers to notify both rising and falling edges
  * of events, with the queue itself maintaining event levels.
  * See description of `uk_poll_func` for more details.
  *
- * Edge-notified queues require setting LIBUKFILE_POLLED during configuration.
+ * Polled queues require setting LIBUKFILE_POLLED during configuration.
  */
 /*
- * We define initializers separate from an initial values.
+ * We define initializers separate from initial values.
  * The former can only be used in (static) variable initializations, while the
  * latter is meant for assigning to variables or as anonymous data structures.
  */
 #if CONFIG_LIBUKFILE_CHAINUPDATE
 #if CONFIG_LIBUKFILE_POLLED
-#define _POLLQ_INIT(q, pollfunc, ev) { \
+#define __UK_POLLQ_INIT(q, pollfunc, ev) { \
 	.wait = NULL, \
 	.waitend = &(q).wait, \
 	.prop = NULL, \
@@ -200,7 +200,7 @@ struct uk_pollq {
 	.waitlock = UK_RWLOCK_INITIALIZER((q).waitlock, 0), \
 }
 #else /* !CONFIG_LIBUKFILE_POLLED */
-#define _POLLQ_INIT(q, pollfunc, ev) { \
+#define __UK_POLLQ_INIT(q, pollfunc, ev) { \
 	.wait = NULL, \
 	.waitend = &(q).wait, \
 	.prop = NULL, \
@@ -214,7 +214,7 @@ struct uk_pollq {
 #endif /* !CONFIG_LIBUKFILE_POLLED */
 #else /* !CONFIG_LIBUKFILE_CHAINUPDATE */
 #if CONFIG_LIBUKFILE_POLLED
-#define _POLLQ_INIT(q, pollfunc, ev) { \
+#define __UK_POLLQ_INIT(q, pollfunc, ev) { \
 	.wait = NULL, \
 	.waitend = &(q).wait, \
 	.poll = (pollfunc), \
@@ -223,7 +223,7 @@ struct uk_pollq {
 	.waitlock = UK_RWLOCK_INITIALIZER((q).waitlock, 0), \
 }
 #else /* !CONFIG_LIBUKFILE_POLLED */
-#define _POLLQ_INIT(q, pollfunc, ev) { \
+#define __UK_POLLQ_INIT(q, pollfunc, ev) { \
 	.wait = NULL, \
 	.waitend = &(q).wait, \
 	.events = (ev), \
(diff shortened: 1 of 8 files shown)


lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,26 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
 	_uk_waitq_unlock(wq, flags);
 }
 
+#define _uk_waitq_wake_up_if_then(wq, ticket_name, condition, aftermath) \
+do {									\
+	struct uk_waitq_ticket *ticket_name;				\
```
michpappas: How is `ticket_name` different than `ticket`?

lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,26 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
 	_uk_waitq_unlock(wq, flags);
 }
 
+#define _uk_waitq_wake_up_if_then(wq, ticket_name, condition, aftermath) \
```
michpappas: Perhaps it would be preferable to prepend macro parameters with underscores to prevent shadows existing symbols of that name that would cause confusing bugs. This is the pattern we normally use anyway.

lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,26 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
...
+do {									\
+	struct uk_waitq_ticket *ticket_name;				\
+	unsigned long _uk_waitq_flags;					\
+									\
+	_uk_waitq_lock((wq), _uk_waitq_flags);				\
+	uk_list_for_each_entry(ticket_name, &wq->waiters, link)		\
+		if ((condition)) {					\
+			_uk_waitq_wake((wq), ticket_name);		\
+			aftermath;					\
```
michpappas, line 298: I take it that the assumption is that if the user needs to pass > 1 statements they have to write something like:

```
uk_waitq_wake_up_if(wq, ticket, cond, ({ after(); math(); }));
```
or
```
uk_waitq_wake_up_if(wq, ticket, cond, do { after(); math(); } while (0));
```

lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,26 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
 	_uk_waitq_unlock(wq, flags);
 }
 
+#define _uk_waitq_wake_up_if_then(wq, ticket_name, condition, aftermath) \
+do {									\
+	struct uk_waitq_ticket *ticket_name;				\
+	unsigned long _uk_waitq_flags;					\
+									\
+	_uk_waitq_lock((wq), _uk_waitq_flags);				\
+	uk_list_for_each_entry(ticket_name, &wq->waiters, link)		\
```
michpappas: If my understanding that `ticket_name` is actually `ticket`, it should probably also be enclosed in parentheses.

lib/ukfile/pollqueue.c
```diff
@@ -4,60 +4,53 @@
  * You may not use this file except in compliance with the License.
  */
 
+#include <uk/assert.h>
 #include <uk/file/pollqueue.h>
 
-#include <uk/assert.h>
+#define pq_notify_one(q, mask, seen, broke)				\
+	uk_waitq_wake_up_one_if(&(q)->waitq, t,				\
```
michpappas: Uhhhhh so the caller is expected to have a varible `t` set, and in addition to what this macro expects? 😬 Also similarly to the comment in `waitq.h` perhaps it's best to prepend parameters with underscores to avoid unexpected shadowing? Maybe these macros are used by this file now but I feel that it's still safer to follow these defensive patterns.

michpappas (changes requested):
LGTM, just a few comments

lib/ukfile/pollqueue.c
andreittr: Tbh I no longer remember what reasoning I had to make these two macros. Seeing as how tightly they integrate with the loop in `pollq_notify_n` they might as well be folded into it. Will do in pass :+1: 

And no, `t` is/was not meant to be from the caller, it was the name of the ticket loop variable later used in the `condition` argument to the `wake_up_one_if` macro. But now that `ticket_name`->`ticket` and an external var, that changes. But the macros are going away anyway.

lib/uksched/include/uk/wait.h
andreittr: :thinking: yeah, `ticket` is definitely the better name; I think I wanted to make it clear that it's a _variable name_ containing the current ticket that you can access in `condition` as you wish. But maybe it'd be better to just take an already-existing var from the calling scope like the forloop macros do. Makes its type clear from the outer scope too.

lib/uksched/include/uk/wait.h
andreittr: I don't think macro argument names have any way of shadowing local vars in expressions supplied to them, due to the single-pass string replacement nature of macro arg handling.
Minimal example:
```c
#define macr(x, y, expr)                        \
do {                                            \
        (x) = "x in the macro";                 \
        expr;                                   \
        (y) = "y in the macro";                 \
} while (0)

void f(void)
{
        char *x, *y;

        macr(y, x, (x = "X in f", y = "Y in f"));
}
```
preprocesses to
```c
void f(void)
{
 char *x, *y;

 do { (y) = "X IN THE MACRO"; (x = "X in f", y = "Y in f"); (x) = "Y IN THE MACRO"; } while (0);
}
```
So as you can see, x and y are correctly handled (this happens regardless of macro arg order or if you eval the args multiple times).

Macro args can shadow e.g., struct fields or identifiers that the macro code itself uses, yes, but that's not the case here AFAICT. Unless needed for name conflicts, I think underscores would only add noise.

lib/uksched/include/uk/wait.h
andreittr: For multiple statements, yes; multiple expressions can be chained with the comma operator.

That said, `aftermath` is not intended for public use, and callers into the waitq API should not be using the `_uk_waitq_wake_up_if_then` macro directly ever (since it's clearly marked private), rather only use `wake_up_if` and `wake_up_one_if`. We might one day change our mind re: all this macro sorcery and bite the code duplication bullet and just write out two bespoke macros.

andreittr:
Pushed addressing comments.

lib/uksched/include/uk/wait.h
michpappas: > multiple expressions can be chained with the comma operator

But then are executed as a single expression IIRC so there may be differences in side-effects.

> That said, aftermath is not intended for public use, and callers into the waitq API should not be using the _uk_waitq_wake_up_if_then macro directly ever (since it's clearly marked private), rather only use wake_up_if and wake_up_one_if.

Yes for sure, mostly discussing, no action requested here 🙃

andreittr:
Pushed fix of potential propagation loop in pollqueues.

lib/uksched/include/uk/wait.h
michpappas: Yes okay, it's variables declared inside macros that can cause shadowing, not parameters, and I believe that's where the habit to prepend all macro vars (including parameters) with underscores emerges from, otherwise you end up using a mixed style that looks inconsistent. And because of that it's easier to establish a rule that says "use underscores in macros" universally rather than elaborating on the different cases. Saying that, I don't see anything in the conventions, and in the codebase I can see examples of both styles, so I won't insist 🙃

michpappas (approved):

lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,25 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
 	_uk_waitq_unlock(wq, flags);
 }
 
+#define _uk_waitq_wake_up_if_then(wq, ticket, condition, aftermath)	\
+do {									\
+	unsigned long _uk_waitq_flags;					\
+									\
+	_uk_waitq_lock((wq), _uk_waitq_flags);				\
+	uk_list_for_each_entry((ticket), &wq->waiters, link)		\
```
mogasergiu: Should this `wq` also have paranthesis?

lib/uksched/include/uk/wait.h
```diff
@@ -253,6 +265,25 @@ void uk_waitq_wake_up_one(struct uk_waitq *wq)
...
+		if ((condition)) {					\
+			_uk_waitq_wake((wq), (ticket));			\
+			aftermath;					\
```
mogasergiu, line 298: I think you can also surround `aftermath` with paranthesis.

lib/ukfile/pollqueue.c
```diff
@@ -133,18 +134,24 @@ uk_pollevent uk_pollq_assign_n(struct uk_pollq *q, uk_pollevent val, int one)
 {
 	uk_pollevent prev;
 	uk_pollevent set;
+#if CONFIG_LIBUKFILE_CHAINUPDATE
+	uk_pollevent clr;
```
mogasergiu, line 126: Feel free to disregard, but in my personal opinion, I find it cleaner to use `__maybe_unused` in cases like this.

lib/ukfile/include/uk/file/pollqueue.h
```diff
@@ -15,7 +15,12 @@
 #include <uk/atomic.h>
 #include <uk/rwlock.h>
 #include <uk/plat/time.h>
-#include <uk/thread.h>
+#include <uk/wait.h>
+
+#if CONFIG_LIBUKFILE_CHAINUPDATE
+#include <uk/list.h>
+#include <uk/mutex.h>
+#endif
```
mogasergiu: `/* CONFIG_LIBUKFILE_CHAINUPDATE */`?

lib/ukfile/include/uk/file/pollqueue.h
```diff
@@ -376,43 +308,22 @@ uk_pollevent _pollq_lock(struct uk_pollq *q, uk_pollevent req,
...
-	uk_thread_block_until(__current, deadline);
-	uk_rwlock_runlock(&q->waitlock);
-	uk_sched_yield();
-	/* Back, wake up, check if timed out & try again */
-	timeout = deadline && ukplat_monotonic_clock() >= deadline;
-	if (timeout)
-		uk_pollq_cancel_ticket(q, &tick);
-	return !timeout;
+	/* Set events as cookie & wait */
+	uk_waitq_set_cookie(req);
```
mogasergiu, line 316: Feels unfortunate how `_uk_pollq_wait` is called in a loop and it reaccesses a per-cpu variable every time (`uk_thread_current` so `%gs:offset` on x86 or `TPIDR_EL` on ARM) rather than caching its pointer and doing updates through that pointer. But I think it's not pondering on this at this time I guess...

lib/ukfile/include/uk/file/pollqueue.h
```diff
@@ -120,209 +110,150 @@ struct uk_poll_chain {
...
+	uk_pollevent events;
+	uk_pollevent waitmask;
+	struct uk_waitq waitq; /* Polling threads */
+	struct uk_rwlock waitlock;
 #if CONFIG_LIBUKFILE_CHAINUPDATE
 	void *_tag; /* Internal use */
-	struct uk_rwlock proplock; /* Chained updates list lock */
+	UK_STAILQ_HEAD(uk_pollq_chain_head, struct uk_poll_chain) prop;
+	struct uk_mutex proplock;
+	uk_pollevent propmask;
```
mogasergiu, line 148: I may be reading too much into it but I noticed that `propmask` is not within the same cacheline as the other prop-related fields, in favor of `_tag`.
```C
struct uk_pollq {
	uk_pollevent               events;               /*     0     4 */
	uk_pollevent               waitmask;             /*     4     4 */
	struct uk_waitq            waitq;                /*     8    16 */
	struct uk_rwlock           waitlock __attribute__((__aligned__(8))); /*    24    48 */
	/* --- cacheline 1 boundary (64 bytes) was 8 bytes ago --- */
	void *                     _tag;                 /*    72     8 */
	struct uk_pollq_chain_head prop;                 /*    80    16 */
	struct uk_mutex            proplock;             /*    96    32 */
	/* --- cacheline 2 boundary (128 bytes) --- */
	uk_pollevent               propmask;             /*   128     4 */

	/* size: 136, cachelines: 3, members: 8 */
	/* padding: 4 */
	/* forced alignments: 1 */
	/* last cacheline: 8 bytes */
} __attribute__((__aligned__(8)));
```
Is it more common for `_tag` to be more frequently accessed together with the other prop fields than `propmask`? Just asking. As far as I can tell that is not necessarily the case, but though I'd ask since, name-wise, it got my attention as it felt like a lost opportunity of having similarly named (and possibly frequently accessed together) struct fields within the same cacheline.

mogasergiu (commented):
LGTM. Just a couple of nits/questions.

lib/ukfile/include/uk/file/pollqueue.h
andreittr: `_uk_pollq_wait` is indeed called in a loop, but that's mostly for the sake of correctness, and the assumption for the vast majority of cases is that the loop terminates after the 1st iteration, either through timeout or successful polling. I would have added branch hints, but had no idea how without nesting a `likely` inside an `unlikely` :smile: .
The only cases where we'd wait more than once are:
- missed events (event got set, but cleared before we got to run): these should be rare, and it's normal for `uk_pollq_poll_until` to ignore them -- for precise tracking of events one would use update chaining
- spurious wakeups (got woken, but our events of interest were never set): these we can avoid altogether when writing code that generates events

lib/ukfile/include/uk/file/pollqueue.h
andreittr: When update chaining is enabled, `_tag` is checked in both `set` and `clr` before doing anything in order to break chaining loops; so it's read before `propmask` or any of the other `prop*`s, and it makes sense to have it "at hand".
That said, I do see 2 unfortunate things:
1. `waitlock` crossing cacheline boundaries is bad; will bite us with SMP
2. `propmask` being after `prop` is suboptimal, since the former is checked before touching the latter

~~I'll give field ordering a bit of thought in the pass.~~
Edit2: On 2nd thought, `struct uk_pollq` is never allocated by itself, but as part of `struct uk_file_state`, which in turn is embedded at various offsets in larger structs. Changing field order is premature optimization for now, but something worth considering for a future perf-guided optimization pass.

Edit: I also realised `_tag` should probably be checked in `assign` too.

lib/uksched/include/uk/wait.h
andreittr: Adding parantheses unfortunately does not work with the plain `break` used in `uk_waitq_wake_one_if` below. To make that work one would need to pad it up with braces and parantheses and semicolons to create a compound statement expression:
```c
#define uk_waitq_wake_up_one_if(wq, ticket, condition) \
	_uk_waitq_wake_up_if_then((wq), (ticket), (condition), ({ break; }))
```
and at that point I feel it hurts readability more than the plain `aftermath` does :shrug: .

andreittr:

andreittr:
Pushed addressing comments

mogasergiu (approved):

michpappas:

PR #1681, opened 2025-07-30
lib/ukfs-devfs: Introduce ukfs filesystem for special files under /dev

andreittr:
### Description of changes

This changeset introduces `ukfs-devfs`, a tiny pseudo-filesystem for Unikraft providing 2 things:
- a global anonymous ramfs root where libraries can register device-like special files, independent of VFS state
- the "devfs" ukfs driver, which returns an instance of the above root on `vopen()`, used for mounting on e.g. `/dev/`

As a first provider of dev special files, `posix-tty` has been patched to provide the `null`, `zero`, `stdin`, `stdout`, and `stderr` files.
Additional prerequisites to these features and tangential changes included:
- ukfs pseudofs driver template
- fix styling in `posix-tty`'s `Config.uk`
- fix a refleak bug in `posix-tty`

~~Depends on the following PR and has it included as squashed commit that will be removed on rebase, please ignore during review:~~
~~- https://github.com/unikraft/unikraft/pull/1609~~

### Prerequisite checklist

### Base target

### Additional configuration

`CONFIG_LIBPOSIX_TTY_DEV_PSEUDO` & `CONFIG_LIBPOSIX_TTY_DEV_STDIO`.


lib/posix-tty/pseudo.c
```diff
@@ -180,3 +187,31 @@ const struct uk_file *uk_zerofile_create(void)
...
+	r = uk_fs_createat(uk_fs_devfs_root, "null", 4, 0666, O_EXCL,
+			   (union uk_fs_create_target){ .file = &null_file });
+	if (unlikely(PTRISERR(r))) {
+		uk_pr_err("Failed to create /dev/null: %d\n", PTR2ERR(r));
+		return PTR2ERR(r);
+	}
+	r = uk_fs_createat(uk_fs_devfs_root, "zero", 4, 0666, O_EXCL,
+			   (union uk_fs_create_target){ .file = &zero_file });
+	if (unlikely(PTRISERR(r))) {
+		uk_pr_err("Failed to create /dev/null: %d\n", PTR2ERR(r));
```
mogasergiu: s/null/zero

Also, won't previously created `/dev/null` leak?

lib/posix-tty/tty.c
```diff
@@ -89,10 +129,14 @@ static int init_posix_tty(struct uk_init_ctx *ictx __unused)
 	}
 
 	r = 0;
+#if CONFIG_LIBPOSIX_TTY_DEV_STDIO
+	r = init_posix_tty_dev_stdio(in, out, out);
```
mogasergiu: Probably would make more sense to have #elif so that we don't have an unnecessary `r = 0` above? :sweat_smile:
(I know the compiler would optimize it out anyway but visually looks unpleasant)

lib/posix-tty/tty.c
```diff
@@ -68,23 +68,31 @@ static int init_posix_tty(struct uk_init_ctx *ictx __unused)
...
 		uk_pr_err("Failed to allocate fd for stderr: %d\n", r);
-		return (r < 0) ? r : -EBADF;
+		if (r > 0)
+			r = -EBADF;
+		goto out;
 	}
 
-	return 0;
+	r = 0;
+out:
```
mogasergiu: What about closing/releasing the fds created by `uk_fdtab_open_named` up until a certain point?

lib/posix-tty/pseudo.c
```diff
@@ -180,3 +187,31 @@ const struct uk_file *uk_zerofile_create(void)
...
+		return PTR2ERR(r);
+	}
+	return 0;
+}
+
+uk_rootfs_initcall_prio(init_posix_tty_dev_pseudo, 0x0, UK_FS_PRIO_FSAVAIL);
```
mogasergiu: Why not release the created files upon shutdown? Is it because it would not longer matter?
(Ditto for tty.c)

lib/posix-tty/tty.c
```diff
@@ -26,6 +32,40 @@
...
+	r = uk_fs_createat(uk_fs_devfs_root, "stdout", 4, 0222, O_EXCL,
+			   (union uk_fs_create_target){ .file = out });
+	if (unlikely(PTRISERR(r))) {
+		uk_pr_err("Failed to create /dev/stdout: %d\n", PTR2ERR(r));
+		return PTR2ERR(r);
+	}
+	r = uk_fs_createat(uk_fs_devfs_root, "stderr", 4, 0222, O_EXCL,
+			   (union uk_fs_create_target){ .file = err });
+	if (unlikely(PTRISERR(r))) {
+		uk_pr_err("Failed to create /dev/stderr: %d\n", PTR2ERR(r));
```
mogasergiu, line 49: Same question as for zero/null. Won't stdin/stdout leak?

lib/posix-tty/tty.c
andreittr: Sure, I've no strong feelings on which looks better, but an elif is definitely more explicit.

lib/posix-tty/tty.c
andreittr: :thinking: I was working under the assumption that any open fds will be cleaned up when the fdtab goes down on system shutdown, although there is an argument to be made about `init_posix_tty` not failing with partial side-effects. But then again, if a failure of an init function is always fatal we might not need to worry about it. ~~Then again again, I'm not sure failure of an init function will run the term functions or just panic :smile: ~~

~~So yeah, it's probably safest to clean up fds on failure too :+1: .~~

edit: I checked and we do indeed [run term functions if init func fails](https://github.com/unikraft/unikraft/blob/staging/lib/ukboot/boot.c#L435); so we can safely rely on fdtab's cleanup removing any partial side-effects without needing bespoke rollback code here.

lib/posix-tty/pseudo.c
andreittr: > s/null/zero

Oops.

> Also, won't previously created /dev/null leak?

I was relying on them being cleaned up by devfs' term function that releases the root. ~~But I have to check if term funcs run on init error or not.~~

edit: I checked and we do indeed [run term functions if init func fails](https://github.com/unikraft/unikraft/blob/staging/lib/ukboot/boot.c#L435); so we can safely rely on devfs' own cleanup here.

lib/posix-tty/pseudo.c
andreittr: On system shutdown devfs will release its root, which will in turn recursively drop references to everything it holds, releasing them in turn. That way libs don't need to cleanup the files they put in pseudofses (and thus needing to keep track of them in parallel with the filesystem), they just get cleaned up automatically by the fs on shutdown.

andreittr:
Pushed update addressing comments & adding `exportsyms.uk` file to explicitly express developer intentions.

mogasergiu (approved):

andreittr:

andreittr:
Pushed minor change adding comment.

andreittr:
Pushed major update with 2 additions:
- Moved the pseudofiles out of `posix-tty` and into their own library
- Added code to fstab for mounting devfs on `/dev` at boot

lib/ukfile-pseudo/pseudo.c
```diff
@@ -147,36 +143,11 @@ static const struct uk_file void_file = {
...
 	.vol = ZERO_VOLID,
 	.node = NULL,
 	.ops = &zero_ops,
 	.refcnt = &zero_ref,
 	.state = &zero_state,
 	._release = uk_file_static_release
 };
-
-
-const struct uk_file *uk_nullfile_create(void)
```
mogasergiu, line 160: I would also mention in the commit message why these have been removed in favor of static lifetime global vars.

mogasergiu (approved):

andreittr:

andreittr:
Pushed functional update & addressing comments:
- Clarified the reasoning for exporting the raw singleton files in `ukfile-pseudo` instead of refcounting getters
- Changed the serial singleton file in `posix-tty` to similarly export the raw file
- Fixed a minor refleak bug of the serial file during `init_posix_tty`
- Fixed a bug where `init_posix_tty` would silently fail to run if `ukfs-devfs` is not enabled

michpappas:

PR #1682, opened 2025-07-31
lib/ukallocregion: Misc improvements

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Remove unnecessary page size check and add branch hinting and adopt decremental bumping strategy instead of incremental (+10-15% faster). Align with `lib/ukallocpool` and embed `struct uk_alloc` into the region allocator structure and export some functions to help region allocator aware callers avoid `ukalloc` indirections.

### DIFF
--- a/lib/ukallocregion/region.c
+++ b/lib/ukallocregion/region.c
@@ -76,11 +76,11 @@ static void *uk_allocregion_malloc(struct uk_alloc *a, size_t size)
 	intptr = ALIGN_UP(((uintptr_t)b->heap_base), (uintptr_t)sizeof(void *));
 
 	newbase  = intptr + size;
-	if (newbase > (uintptr_t)b->heap_top)
+	if (unlikely(newbase > (uintptr_t)b->heap_top))
 		goto enomem; /* OOM */
 
 	/* Check for overflow, handle malloc(0) */
-	if (newbase <= (uintptr_t)b->heap_base)
+	if (unlikely(newbase <= (uintptr_t)b->heap_base))
 		goto enomem;
 
 	uk_alloc_stats_count_alloc(a, (void *)intptr,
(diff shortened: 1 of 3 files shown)


lib/ukallocregion/region.c
```diff
@@ -55,51 +55,40 @@
...
 
 	UK_ASSERT(a != NULL);
 
 	b = (struct uk_allocregion *)&a->priv;
 
-	UK_ASSERT(b != NULL);
+	if (unlikely(!size))
+		return NULL;
+
+	if (unlikely((uintptr_t)b->heap_pos - (uintptr_t)b->heap_base < size)) {
```
andreittr: Nit: you seem to use this same expression to check for space both here and in memalign. Since it's not exactly trivial, would it be worth a macro/static inline?

lib/ukallocregion/region.c
```diff
@@ -55,51 +55,40 @@
...
-	/* Check for overflow, handle malloc(0) */
-	if (unlikely(newbase <= (uintptr_t)b->heap_base))
-		goto enomem;
-
-	uk_alloc_stats_count_alloc(a, (void *)intptr,
-				   newbase - (uintptr_t)b->heap_base);
-
-	b->heap_base = (void *)newbase;
+	b->heap_pos = (void *)ALIGN_DOWN((uintptr_t)b->heap_pos - size,
+					 (uintptr_t)sizeof(void *));
```
andreittr: Similar nit argument to the space check expression: would the operation of "pop next region" be worth a macro/static inline?

andreittr (commented):
For the record, I'm on the fence about the proliferation of private allocator APIs vs. the common ukalloc API; not going to opine on the merits of either, as probably that's a bigger arch discussion. It's fine if we need it for now, we can always revisit this later.

Other than that, just two small nitpicks.

andreittr (approved):
Looks good, thanks!

andreittr:

PR #1683, opened 2025-07-31
 lib/ukcpio: Track and link inodes whose `nlink` is bigger than one 

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Use a simple singly-linked list structure to keep track of CPIO entries whose inode has multiple hardlinks (`nlink` > 1). A plain region/bump allocator is used to minimise allocator overhead, especially since we only need to do one single free (releasing the whole allocator) at the end.


lib/ukcpio/cpio.c
```diff
@@ -57,10 +59,71 @@ int uk_syscall_do_chmod(const char *, mode_t);
...
+
+		/*
+		 * If current element's inode is not the one we are looking for
+		 * and the next element's inode is bigger, then we haven't
+		 * encountered the inode we are looking for and this is the
+		 * first time we are seeing it so return current element to be
+		 * able to insert right after it and maintain the list
+		 * ascendingly sorted.
+		 */
+		next_elm = UK_SLIST_NEXT(elm, next);
```
andreittr: Consider using `UK_SLIST_FOREACH_SAFE` that sets `next_elm` for you.

lib/ukcpio/cpio.c
```diff
@@ -57,10 +59,71 @@ int uk_syscall_do_chmod(const char *, mode_t);
...
+		if (next_elm && next_elm->ino < ino)
```
andreittr: The logic on this line doesn't line up with the commit message and the comment above :thinking: If the list is indeed sorted by ascending `ino`, then the return should happen on `next_elm->ino > ino`.
If, instead you keep the list descending, you need to update the doc.

Assuming the condition is fixed & consistent for ascending order, I still think there are cornercases that this code is missing:
- Inserting a new maximum `ino` into a non-empty list: the loop will run to the end and return NULL => insert to the head (correct thing to do is add to the TAIL)
- Inserting a new minimum `ino` into a non-empty list: this `if` will trigger on 1st iteration and return the 1st element => insert after elm0 (correct thing: insert at HEAD)

We can't really control the ordering of files within a CPIO or the order inode numbers are picked in, so we should take extra care in making this robust.

lib/ukcpio/cpio.c
```diff
@@ -277,6 +345,54 @@ extract_section(const struct uk_cpio_header **headerp, char *fullpath,
...
+		}
+
+		/*
+		 * There is no element with current entry's inode and found
+		 * element's inode is smaller so create the new inode list
+		 * element to insert.
+		 */
+		new_elm = cpio_ilist_elm_create(ilist, ino,
+						fullpath + 2,
+						prefixlen + namesize - 2);
```
andreittr: What's the purpose of this +2/-2 over here? :thinking:

lib/ukcpio/cpio.c
```diff
@@ -348,9 +471,39 @@ ukcpio_extract(const char *dest, const void *buf, size_t buflen)
...
+	/*
+	 * Since we cannot know how many CPIO entries with nlink > 1 are in
+	 * a CPIO archive, just initialize the region allocator based on the
+	 * biggest possible contiguous allocation of the current default
+	 * allocator.
+	 *
+	 * NOTE: If in the future we do parallel booting, we may have to
+	 * reconsider this here so as not to temporarily steal too much memory.
+	 */
+	max_alloc = uk_alloc_maxalloc(a);
```
andreittr: While it's true we can't know how many ilist entries we'll need, we can set an upper bound on that as `CPIO_SIZE / sizeof(struct uk_cpio_header)`.

lib/ukcpio/cpio.c
```diff
@@ -277,6 +345,54 @@ extract_section(const struct uk_cpio_header **headerp, char *fullpath,
...
+		 * just link this current entry to it.
+		 */
+		if (elm && elm->ino == ino) {
+			uk_pr_err("Inode %u has already been linked by %s, creating new hardlink at %s\n",
+				  ino, elm->path, fullpath);
+			rc = uk_syscall_do_link(elm->path, fullpath);
+			if (unlikely(rc)) {
+				uk_pr_err("Failed to create new hard link %s (from %s).\n",
+					  fullpath, elm->path);
+				return -UKCPIO_LINK_FAILED;
```
andreittr, line 370: I don't think this is meant to be negative.

lib/ukcpio/cpio.c
```diff
@@ -57,10 +59,71 @@ int uk_syscall_do_chmod(const char *, mode_t);
...
+					     size_t pathlen)
+{
+	struct cpio_ilist_elm *elm;
+
+	elm = uk_allocregion_bump(ilist->ar, sizeof(*elm) + pathlen + 1);
+	if (unlikely(!elm))
+		return NULL;
+
+	elm->ino = ino;
+	strncpy(elm->path, path, pathlen);
```
andreittr: I don't think `strncpy` is the right function to use here, for a few reasons:
- we know the length of `path`, no point in probing every byte for a NUL during copy
- neither its truncation, nor NUL padding logic are desired or useful here
- the 3rd arg is supposed to be "number of bytes available in dest" not "length of source"
  - => the 3rd arg is off-by-one (it's supposed to be `pathlen + 1`)
  - this also means that you truncate the char array in `path`, not copying over the terminating NUL; you do get lucky by having most of system mem be zeroed at boot, but still not correct.

Considering that all paths you get in this function are validated and were [built right here in ukcpio](https://github.com/unikraft/unikraft/pull/1683/files#diff-16c7dcd3c3387d5b8d30d83b32014a1bd9f4354b64f2deb66bfde8e9eec83cccR346), and are guaranteed a terminating NUL, I think you can go for a plain `memcpy` :shrug: .

andreittr (commented):
On first pass, I have some concerns regarding correctness, see comments.

lib/ukcpio/cpio.c
mogasergiu: I think I got confused by this cpio.c. It seems that in case of errors, a negative enum is returned and otherwise a positive CPIO_SUCCESS enum is returned... so I guess this is correct to be negative but my UKCPIO_NOMEM must be negative as well.

andreittr: Ok, yeah, you're right, the _NOMEM returns need to be negative then.
I'm genuinely surprised the compiler lets you return arbitrary numbers as `enums` though :smile: .

lib/ukcpio/cpio.c
mogasergiu: Hmm, leftover from testing extraction as rootfs (`/.`). Will fix :+1:

lib/ukcpio/cpio.c
```diff
@@ -57,10 +59,75 @@ int uk_syscall_do_chmod(const char *, mode_t);
...
+		 * encountered the inode we are looking for and this is the
+		 * first time we are seeing it so return current element to be
+		 * able to insert right after it and maintain the list
+		 * ascendingly sorted.
+		 */
+		if (next_elm && next_elm->ino > ino)
+			break;
+	}
+
+	return elm;
```
andreittr, line 130: Correct me if I'm wrong, but the previous `FOREACH` will leave `elm == NULL` on exhaustion (e.g., when `ino` greater than all existing elements). In that case it will be added to the HEAD, which is not what we want.

lib/ukcpio/cpio.c
```diff
@@ -277,6 +349,53 @@ extract_section(const struct uk_cpio_header **headerp, char *fullpath,
...
+			*headerp = UKCPIO_NEXT(header, namesize, filesize);
+			return UKCPIO_SUCCESS;
+		}
+
+		/*
+		 * There is no element with current entry's inode and found
+		 * element's inode is smaller so create the new inode list
+		 * element to insert.
+		 */
+		new_elm = cpio_ilist_elm_create(ilist, ino, fullpath, namesize);
```
andreittr: The path length is too short there; AFAICT `namesize` is the length of the file path within the CPIO archive, including the terminator NUL, but excluding the extraction prefix. You probably need to add that back in to process all of `fullpath`.

lib/ukcpio/cpio.c
```diff
@@ -348,9 +473,40 @@ ukcpio_extract(const char *dest, const void *buf, size_t buflen)
 		pathbuf[destlen] = 0;
 	}
 
+	a = uk_alloc_get_default();
+
+	/*
+	 * Since we cannot know how many CPIO entries with nlink > 1 are in
+	 * a CPIO archive, just initialize the region allocator based on the
+	 * biggest possible contiguous allocation of the current default
+	 * allocator.
```
andreittr: Thanks for the upper bound on alloc size! Though I think this comment needs updating to reflect the new logic. (And perhaps remove the NOTE? :shrug: )

lib/ukcpio/cpio.c
```diff
@@ -57,10 +59,75 @@ int uk_syscall_do_chmod(const char *, mode_t);
...
+{
+	struct cpio_ilist_elm *elm;
+
+	elm = uk_allocregion_bump(ilist->ar, sizeof(*elm) + pathlen + 1);
+	if (unlikely(!elm))
+		return NULL;
+
+	elm->ino = ino;
+	memcpy(elm->path, path, pathlen);
+	elm->path[pathlen] = '\0';
```
andreittr: Very minor nit: since `path[pathlen]` already contains the terminating NUL (copied over from the CPIO), we can probably remove this explicit assignment, and cut 1 byte off the allocation size too.

andreittr (commented):
Thanks for the updates, just two more issues and minor nits.

lib/ukcpio/cpio.c
mogasergiu: Yeah got very confused locally when mounting on root (`/`) as it is also 1 byte in length and i forgot that `namesize` in CPIO includes the terminating null byte. So during testing when I saw that `namesize` is already the size of `fullpath` (when  mounting on `/`) I thought it was the correct parameters and that I would also need an explicit null terminating byte in the element creation :sweat_smile:

andreittr (approved):
Thanks, looks good!

andreittr:

PR #1684, opened 2025-07-31
arch: Check each `XSAVE` state component depending on the bitmap and add the check on ARM as well

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

According to the Intel SDM, XSTATE_BV is a bitmap that, depending on
which extended register context subcomponent is used, it may have its
corresponding bit marked as dirty through the CPU-internal state
tracking structure XINUSE. If a subcomponent is enabled (RFBM[i] = 1)
then it is stated that: If the state component is in its initial
configuration, XINUSE[i] may be either 0 or 1, and XSTATE_BV[i] may be
written with either 0 or 1. In other words, if the component happens to
be zeroed out entirely, its XSTATE_BV[i] can be either 0 or 1, both
being valid. Some microarchitectures asynchrounously track this state
and may unset the component's bit if its state is noticed to be clean.

This could lead to rare cases where even sequential XSAVEOPT
instructions, executed one after another, may save a different XSTATE_BV
value in the XSAVE header which is valid.

Therefore, we must split the memory comparing in case of XSAVE(OPT). We
have 3 individual subcomponents to check: x87, SSE and AVX states. If
a subcomponent's bit in the bitmap is 0 in both states to compare, it
means that subcomponent was not touched. Otherwise, if the bits differ
then we must check that the subcomponent of the XSAVE's image whose bit
is 1, is entirely zeroed out. If both bits are 1 then we must compare
the subcomponent of each XSAVE image against each other and they must
coincide.

Below are some interesting experiments and their results. Consider the
starting extended register state as clean and the following steps:
1. dirty an AVX register (e.g. VMOVDQU with non-zero values)
2. loop for a while
3. undirty AVX register
4. do xgetbv with RCX as 1 to get XINUSE bitmap until the XINUSE[2]
(2 being the bit corresponding to AVX) is undirtied

If you do VZEROALL/VZEROUPPER or a self-VPXOR of said register at
step 3, the loop at step 4 end instantly, as the bit is syncrhounously
cleared.
If you do VMOVDQU with zeroed out memory for step 3, the loop at step 4
will run for a while, indicating that the CPU does not syncrhounously set
the bit but rather after some arbitrary cycles it clears the bit after
noticing the state being clean again.

While at it, implement this check on ARM as well.

### DIFF
--- a/arch/arm/ectx.c
+++ b/arch/arm/ectx.c
@@ -41,6 +41,7 @@
 #include <uk/essentials.h>
 #include <uk/assert.h>
 #include <uk/print.h>
+#include <uk/hexdump.h>
 #include <uk/isr/string.h> /* memset_isr */
 #include <arm/cpu.h>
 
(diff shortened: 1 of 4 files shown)


michpappas (approved):
For the `arm64` patch:

andreittr (approved):
Looks good, thanks for the rigorous validation code on x86 :smile: .

andreittr:

PR #1687, opened 2025-08-08
include/uk/event: Make handler and raise return values type-safe

michpappas:
`uk_event` handlers often omit using one of the designated macro values, and return an integer instead. This is particularly problematic when returning zero to signify that the event was handled, as `event.h` assigns zero to `UK_EVENT_NOT_HANDLED`.

Convert `uk_event` handler return values to enum to catch such errors at compile-time. Introduce a `UK_EVENT_ERROR` state and an additional parameter in the handler prototype that handlers must populate before returning `UK_EVENT_ERROR`.

### Prerequisite checklist

### Base target

 - Architecture(s): N/A
 - Platform(s): N/A
 - Application(s): N/A

### DIFF
--- a/include/uk/event.h
+++ b/include/uk/event.h
@@ -74,13 +74,14 @@
  * void some_function(void)
  * {
  *       struct myevent_data = { .dummy = 42 };
+ *       int error;
  *       int rc;
  *
- *       rc = uk_raise_event(myevent, &myevent_data);
- *       if (rc < 0)
- *               uk_pr_crit("an event handler returned an error!\n");
- *       else if (!rc)
- *               uk_pr_err("myevent not handled!\n");
+ *       rc = uk_raise_event(myevent, &myevent_data, &error);
+ *       if (unlikely(rc < UK_EVENT_ERROR))
+ *               uk_pr_crit("an event handler returned an error (%d)\n", error);
+ *       else if (rc == UK_EVENT_NOT_HANDLED)
+ *               uk_pr_err("myevent not handled\n");
  * }
  * ```
  * **libB/myhandler.c:**
@@ -88,19 +89,19 @@
  * #include <uk/event.h>
  * #include <uk/myevent.h>
  *
- * int handler1(void *arg)
+ * enum uk_event_status handler1(void *arg, int *error)
  * {
  *       struct myevent_data *data = (struct myevent_data *)arg;
  *       ...
  *       return UK_EVENT_HANDLED_CONT;
  * }
  *
- * int handler2(void *arg)
+ * enum uk_event_status handler2(void *arg, int *error)
  * {
  *       return UK_EVENT_HANDLED_CONT;
  * }
  *
- * int handler3(void *arg)
+ * enum uk_event_status handler3(void *arg, int *error)
  * {
  *       return UK_EVENT_HANDLED;
  * }
@@ -136,14 +137,22 @@ extern "C" {
  *
  * @param data
  *   Optional data parameter. The data is supplied when raising the event
+ * @param error
+ *   Pointer to error code. Set by the handler when returning UK_EVENT_ERROR.
  * @return
  *   One of the UK_EVENT_* macros on success, errno on < 0
  */
-typedef int (*uk_event_handler_t)(void *data);
+typedef enum uk_event_status (*uk_event_handler_t)(void *data, int *error);
 
-#define UK_EVENT_NOT_HANDLED	0  /* Event not handled. Try next handler. */
-#define UK_EVENT_HANDLED	1  /* Event handled. Stop calling handlers. */
-#define UK_EVENT_HANDLED_CONT	2  /* Event handled. Call next handler. */
+enum uk_event_status {
+	UK_EVENT_NOT_HANDLED,	/* Event not handled. Try next handler. */
+	UK_EVENT_HANDLED,	/* Event handled. Stop calling handlers. */
+	UK_EVENT_HANDLED_CONT,	/* Event handled. Call next handler. */
+	UK_EVENT_ERROR		/* Error while processing the event. The
+				 * handler must return an error code via
+				 * *error.
+				 */
+};
 
 struct uk_event {
 	const uk_event_handler_t *hlist_end;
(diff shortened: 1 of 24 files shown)


include/uk/event.h
```diff
@@ -269,7 +271,8 @@ struct uk_event {
  *     other handler returned UK_EVENT_HANDLED.
  *   - UK_EVENT_NOT_HANDLED if no handler handled the event.
  */
-static inline int uk_raise_event_ptr(struct uk_event *e, void *data)
+static inline
+enum uk_event_status uk_raise_event_ptr(struct uk_event *e, void *data)
 {
 	const uk_event_handler_t *itr;
 	int rc;
```
mogasergiu: Shouldn't `rc` and `ret` also be enum given the comparisons below?

lib/ukdebug/arch/arm64/gdbsup.c
```diff
@@ -79,7 +79,7 @@ static int gdb_arch_check_brk(unsigned long pc)
 	return ((opcode & BRK_OPCODE_MASK) != BRK_OPCODE);
 }
 
-static int gdb_arch_debug_handler(void *data)
+static enum uk_event_status gdb_arch_debug_handler(void *data)
```
mogasergiu: `have_brk` and `r` are `int`

lib/ukdebug/arch/x86_64/gdbsup.c
```diff
@@ -37,7 +37,7 @@ static int gdb_arch_dbg_trap(int errnr, struct __regs *regs)
 	return 0;
 }
 
-static int gdb_arch_debug_handler(void *data)
+static enum uk_event_status gdb_arch_debug_handler(void *data)
```
mogasergiu: Ditto `r`.

lib/uknofault/excpttab.h
```diff
@@ -78,7 +78,8 @@ extern const struct nf_excpttab_entry uk_excpttab_end[];
 #define nf_excpttab_foreach(itr, excpttab_start, excpttab_end)		\
 	for ((itr) = (excpttab_start); (itr) < (excpttab_end); (itr)++)
 
-static inline int nf_handle_trap(__vaddr_t ip, struct ukarch_trap_ctx *ctx)
+static inline
+enum uk_event_status nf_handle_trap(__vaddr_t ip, struct ukarch_trap_ctx *ctx)
```
mogasergiu: There is a `return handler` statement and it seems that the definition is to return int.
```C
typedef int (*nf_excpt_handler)(const struct nf_excpttab_entry *e,
				struct ukarch_trap_ctx *ctx);
```

mogasergiu (changes requested):
LGTM, just a couple of missed type updated I think

include/uk/event.h
michpappas: Yes, and that makes me realize that I completely missed the case that the handlers can also return errors. Which means that `enum` won't do. I'll push an update to address the problem in a different way.

include/uk/event.h
```diff
@@ -74,33 +74,34 @@
...
+ *       int error;
  *       int rc;
  *
- *       rc = uk_raise_event(myevent, &myevent_data);
- *       if (rc < 0)
- *               uk_pr_crit("an event handler returned an error!\n");
- *       else if (!rc)
- *               uk_pr_err("myevent not handled!\n");
+ *       rc = uk_raise_event(myevent, &myevent_data, &error);
+ *       if (unlikely(rc < UK_EVENT_ERROR))
```
mogasergiu, line 81: I would drop the unlikely from the documentation and leave as it was. Also the type of `rc` might need to be updated.

include/uk/event.h
```diff
@@ -259,32 +268,35 @@ struct uk_event {
...
+ *     event. Event processing stops immediately in this case, and an errror
+ *     code is saved in *error.
  */
-static inline int uk_raise_event_ptr(struct uk_event *e, void *data)
+static inline enum uk_event_status
+uk_raise_event_ptr(struct uk_event *e, void *data, int *error)
 {
 	const uk_event_handler_t *itr;
-	int rc;
 	int ret = UK_EVENT_NOT_HANDLED;
```
mogasergiu, line 288: I think `ret` should have been enum?

include/uk/event.h
```diff
@@ -259,32 +268,35 @@ struct uk_event {
...
+		ret = ((*itr)(data, error));
+		if (unlikely(ret == UK_EVENT_ERROR))
+			return UK_EVENT_ERROR;
 
-		if (rc == UK_EVENT_HANDLED)
+		if (ret == UK_EVENT_HANDLED)
 			return UK_EVENT_HANDLED;
 
-		if (rc == UK_EVENT_HANDLED_CONT)
+		if (ret == UK_EVENT_HANDLED_CONT)
```
mogasergiu, line 299: I think this if statement is no longer necessary

include/uk/event.h
```diff
@@ -302,9 +314,9 @@ static inline int uk_raise_event_ptr(struct uk_event *e, void *data)
  * @return
  *   One of the UK_EVENT_* macros on success, errno on < 0
  */
-#define uk_raise_event(event, data)					\
+#define uk_raise_event(event, data, error)				\
 	({	_UK_EVT_IMPORT_EVENT(event);				\
-		uk_raise_event_ptr(UK_EVENT_PTR(event), data);	})
+		uk_raise_event_ptr(UK_EVENT_PTR(event), data, error);	})
```
mogasergiu, line 319: Maybe we should enclose `data` and `error` between paranthesis? Also don't forget to now add the `error` argument documentation to `uk_raise_event` as well.

lib/posix-process/execve.c
```diff
@@ -83,9 +83,12 @@ static int pprocess_cleanup(struct posix_process *pprocess)
 	return 0;
 }
 
-int pprocess_raise_execve_event(struct posix_process_execve_event_data *data)
+enum uk_event_status
+pprocess_raise_execve_event(struct posix_process_execve_event_data *event_data,
+			    int event_error)
```
mogasergiu, line 88: I think you meant `int *event_error`? (You do have it in the header definition :D)

lib/ukboot/shutdown_req.c
```diff
@@ -57,7 +58,8 @@ int uk_boot_shutdown_req_isr(enum ukplat_gstate target)
 }
 
 #if __INTERRUPTSAFE__ && CONFIG_LIBUKBOOT_SHUTDOWNREQ_HANDLER
-static int shutdown_req_handler(void *data)
+static
+enum uk_event_status shutdown_req_handler(void *data, int *error __unused)
```
mogasergiu, line 62: I think you can update `error` to return value of `rc`.

lib/ukvmem/arch/arm/pagefault64.c
```diff
@@ -6,14 +6,16 @@
...
 #include <uk/arch/types.h>
-#include <uk/print.h>
+#include <uk/assert.h>
+#include <uk/compiler.h>
 #include <uk/config.h>
-#include <string.h>
+#include <uk/print.h>
 
-static int vmem_arch_pagefault(void *data)
+static enum uk_event_status vmem_arch_pagefault(void *data, int *error __unused)
```
mogasergiu, line 18: Ditto (for x86 too)

mogasergiu (changes requested):
For me it feels viscerally wrong whenever I end up having to add functions with lateral effect and can't help but wonder if there is a better way. But, yeah, for now I can't see a better way than this that would fit all requirements either...
So I guess we can stick with this or... maybe leave it for later and brainstorm some more. I would approve either way.

PR #1688, opened 2025-08-14
plat: Add in-exception lcpu getters and `__isr` macro

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

In the case of some functions, we may want to be able to compile them
and be sure they do not use any extended registers (e.g., FPU, SIMD)
so that we can use them in an exception context or anywhere we do not
want to taint the said extended context.

To make it easy, add the `__isr` macro one can add to its function
to ensure it complies with the above.

Using said macro, introduce two lcpu getters:
- `lcpu_get_current_in_except`:
Assuming we are in an exception handler on an exception stack,
fetch  the current LCPU pointer in such a manner that we do not
assume any register state. This is done by dividng current stack
pointer relative to the LCPU stack base to find out the current
LCPU id.

- `ukplat_lcpu_get_auxsp_in_except`:
Implement `ukplat_lcpu_get_auxsp_in_except`, an Interrupt-safe variant
of `ukplat_lcpu_get_auxsp` that relies on the caller being in an CPU
exception handler routine. We can use this function in case we need
to get current LCPU's `auxsp` without having to do any costly
system register read/write.

### DIFF
--- a/arch/arm/arm64/include/uk/asm/compiler.h
+++ b/arch/arm/arm64/include/uk/asm/compiler.h
@@ -58,3 +58,7 @@
 #define __no_branch_protection
 #endif /* CONFIG_ARM64_FEAT_PAUTH || CONFIG_ARM64_FEAT_BTI */
 
+#define __isr						\
+	__attribute__((__target__(			\
+		"general-regs-only"			\
+	)))
--- a/arch/x86/x86_64/include/uk/asm/compiler.h
+++ b/arch/x86/x86_64/include/uk/asm/compiler.h
@@ -32,4 +32,10 @@
 #endif
 
 /* Architecture-specific compiler definitions */
-
+#define __isr						\
+	__attribute__((__target__(			\
+		"no-mmx,"				\
+		"no-sse,"				\
+		"no-avx,"				\
+		"general-regs-only"			\
+	)))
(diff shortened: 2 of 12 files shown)


michpappas (commented):
Just one question: Should we be using the (informally) conventional `_isr` suffix in the function name instead of `in_except`?

mogasergiu:
> Just one question: Should we be using the (informally) conventional `_isr` suffix in the function name instead of `in_except`?

I am not sure. I chose `in_except` because unlike the other `_isr` stuff, there is not only asking the compiler not to use extended registers for it, but it is also something that assumes that runs on one of the per-lcpu exception stacks. If that assumption is not met you will likely get a crash. I also don't like how `in_except` sounds, but it felt like I needed something stronger than `_isr` to reflect this assumption. Other naming suggestions are more than welcome.

michpappas:
> > Just one question: Should we be using the (informally) conventional `_isr` suffix in the function name instead of `in_except`?
> 
> I am not sure. I chose `in_except` because unlike the other `_isr` stuff, there is not only asking the compiler not to use extended registers for it, but it is also something that assumes that runs on one of the per-lcpu exception stacks. If that assumption is not met you will likely get a crash. I also don't like how `in_except` sounds, but it felt like I needed something stronger than `_isr` to reflect this assumption. Other naming suggestions are more than welcome.

No it's fine, the question was not about the name itself but more on whether we should keep consistency. I see your point now, let's keep it as is.

michpappas (approved):

michpappas:

PR #1690, opened 2025-08-18
lib/posix-process: Signal the application on system error

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Allow the appliation to handle signals generated on system error.
These are the usual SIGBUS, SIGFPE, SIGILL, SIGSEGV, and SIGTRAP.
    
We treat the case where the appliation cannot receive the signal as a non-recoverable error.
Right now this may cause cascading faults, and will be addressed by `UK_CRASH_EX()` introduced in #1531.

In addition to the above, demote the message printed by ukvmem on unhandled page faults to debug, 
As page faults can now signal the application, an unhandled page fault is not necessarily a critical condition.
This delegates the responsibility of unhandled faults to the trap vector, as it should.


lib/ukvmem/arch/x86_64/pagefault.c
```diff
@@ -40,11 +40,11 @@ static int vmem_arch_pagefault(void *data)
 	if (rc < 0) {
 		vas = uk_vas_get_active();
 		if (unlikely(vas && !(vas->flags & UK_VAS_FLAG_NO_PAGING)))
-			uk_pr_crit("Cannot handle %s page fault at 0x%"
-				   __PRIvaddr" (ec: 0x%x): %s (%d).\n",
-				   faultstr[faulttype &
+			uk_pr_debug("Cannot handle %s page fault at 0x%"
```
mogasergiu, line 43: What about the ARM64 side?

lib/posix-process/signal/deliver.c
```diff
@@ -369,3 +369,63 @@ static void uk_signal_deliver(struct uk_syscall_exit_ctx *exit_ctx)
...
+void sys_error_handler(struct ukarch_execenv *ee __unused, long arg)
+{
+	struct ukarch_auxspcb *auxspcb;
+	struct sys_error_desc *error;
+	struct posix_thread *pthread;
+	struct posix_process *pproc;
+	struct uk_signal sig = {0};
+
+	/* Now we can enable IRQs */
+	ukplat_lcpu_enable_irq();
```
mogasergiu, line 383: I think this needs more documentation as why we always enable IRQs, i.e., running an app/userspace sighandler will always have IRQs enabled.
But, I have a suggestion: what if we introduce a `ukarch_regs_get_irqf`? Then you could do `ukplat_lcpu_restore_irqf(ukarch_regs_get_irqf(&ee->regs));`. And this way you would be able to ensure consistent state with what was before.

lib/posix-process/signal/deliver.c
```diff
@@ -369,3 +369,63 @@ static void uk_signal_deliver(struct uk_syscall_exit_ctx *exit_ctx)
...
+		uk_sigact_term(SIGKILL);
+		UK_BUG(); /* noreturn */
+	}
+
+	/* If the application masks or ignores this signal, panic.
+	 * A general-purpose OS would terminate the applciation.
+	 * Being a unikernel, we treat this as a non-recoverable
+	 * error instead.
+	 */
+	if (!pprocess_signal_is_deliverable(pthread, error->signum))
```
mogasergiu, line 411: Should these 2 `if` statements have `unlikely`. Maybe they could also be part of the same statement so that you could get rid of the `goto`.

lib/posix-process/signal/deliver.c
```diff
@@ -369,3 +369,63 @@ static void uk_signal_deliver(struct uk_syscall_exit_ctx *exit_ctx)
...
+		goto err_panic;
+
+	if (KERN_SIGACTION(pproc, error->signum)->ks_handler == SIG_DFL)
+		goto err_panic;
+
+	/* Prepare siginfo */
+	set_siginfo_kill(error->signum, &sig.siginfo);
+
+	/* Execute standard delivery path */
+	do_deliver(pthread, &sig, ee);
```
mogasergiu, line 421: This returns an `int` :thinking:

lib/posix-process/signal/system_error.c
```diff
@@ -0,0 +1,103 @@
...
+	handler_ctx = (struct sys_error_desc *)handler_sp;
+	handler_ctx->signum = signum;
+	handler_ctx->auxsp = auxsp;
+	handler_ctx->vaddr = trap_ctx->fault_address;
+
+	/* Jump away from the exception context */
+	ukarch_ctx_init_ehtrampo(&ctx,
+				 trap_ctx->regs,
+				 handler_sp,
+				 sys_error_handler, handler_sp);
```
mogasergiu: Maybe pass `handler_ctx` here instead. I know they are the same at this point, but helps avoid confusion.

lib/posix-process/signal/system_error.c
```diff
@@ -0,0 +1,103 @@
...
+
+static int pprocess_signal_math_handler(void *arg)
+{
+	return sys_error_handler_except(SIGFPE, (struct ukarch_trap_ctx *)arg);
+}
+
+/* Execute system fault handlers after pf, and in fact last as we cannot
+ * return from the isr-safe context if we can't handle the signal.
+ */
+UK_EVENT_HANDLER_PRIO(UKARCH_TRAP_PAGE_FAULT, pprocess_signal_pf_handler,
```
mogasergiu, line 96: I think that `UKARCH_TRAP_PAGE_FAULT` is the only one that should be `UK_PRIO_LATEST`?

lib/posix-process/signal/deliver.c
```diff
@@ -369,3 +369,63 @@ static void uk_signal_deliver(struct uk_syscall_exit_ctx *exit_ctx)
...
+	UK_ASSERT(pthread);
+
+	/* If there's a SIGKILL pending, kill the process right away */
+	if (IS_PENDING(pproc->signal->sigqueue, SIGKILL)) {
+		uk_sigact_term(SIGKILL);
+		UK_BUG(); /* noreturn */
+	}
+
+	/* If the application masks or ignores this signal, panic.
+	 * A general-purpose OS would terminate the applciation.
```
mogasergiu: s/applciation/application

lib/posix-process/signal/signal.h
```diff
@@ -205,6 +205,13 @@ struct uk_signal_tdesc {
 	struct uk_semaphore pending_semaphore;
 };
 
+/* System error descriptor passed to handler */
+struct sys_error_desc {
+	int signum;
+	__uptr auxsp;
+	unsigned long vaddr;
```
mogasergiu: Myabe `__vaddr_t`?

lib/ukvmem/arch/x86_64/pagefault.c
michpappas: `arm64` will be transparently fixed in https://github.com/unikraft/unikraft/pull/1531 where we unify the trap descriptor structs.

lib/posix-process/signal/deliver.c
michpappas: > I think this needs more documentation as why we always enable IRQs, i.e., running an app/userspace sighandler will always have IRQs enabled.

Good point, I'll update the comment.

> I think this needs more documentation as why we always enable IRQs, i.e., running an app/userspace sighandler will always have IRQs enabled.
But, I have a suggestion: what if we introduce a ukarch_regs_get_irqf? Then you could do ukplat_lcpu_restore_irqf(ukarch_regs_get_irqf(&ee->regs));. And this way you would be able to ensure consistent state with what was before.

Possibly, but I think it's out of scope of this PR, as we can't anticipate a case where interrupts are disabled while the application is running.

lib/posix-process/signal/deliver.c
michpappas: Whether a signal is deliverable depends on whether the application has chosen to masks or ignores it. Because it's application-specific, we can't make any assumptions on whether that case is likely or unlikely, as opposed to something like an error that we can treat as unlikely with confidence.

lib/posix-process/signal/system_error.c
michpappas: Yes, and in fact I forgot to rename it to `handler_desc` as I renamed the struct to be a descriptor rather than a context.

lib/posix-process/signal/signal.h
michpappas: Sure. The reason I decided to stick to `unsigned long` is that `ukarch_trap_ctx` defines `fault_addr` as an `unsigned long`, which I don't like so I changed to `__vaddr_t` in #1531. I'll set it to `__vaddr_t` now in anticipation of #1531.

lib/posix-process/signal/system_error.c
michpappas: I think all system errors may be handled by some other component first (fpu emulation, gdb stub, etc), so signalling the application should be the last resort. I'll make the comment more generic.

lib/posix-process/signal/deliver.c
michpappas: That function always returns zero, so no calls check its result. I'll add a commit to change it  void.

mogasergiu (approved):

PR #1694, opened 2025-08-25
lib/posix-socket: Poll for connection error on blocking sockets

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

On a blocking socket, whenever the driver would return -EINPROGRESS, the core would attempt to poll for when the socket becomes available for writing. If the other end however is no longer available then the equivalent of an EPOLLHUP would be generated for which the aforementioned polling does not check. This leads to the calling thread blocking indefinetely.

Fix this by also polling for errors so that the event handler can set the socket event and thus wake the blocked thread.


andreittr (approved):
Nice catch! Indeed the manpage only specifies "polling for writability" for `EINPROGRESS`, but that's under the implicit understanding that HUP/ERR are _always_ polled for :smile: . Mental note for myself to check this behavior throughout `posix-fdio`.

andreittr:

PR #1695, opened 2025-08-26
lib/posix-process: Rework event API

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Rework the API for pprocess / pthread lifetime management events.

## Kernel API
The new API provides unified handler registration macros for task create, exec, and terminate. What follows is a brief summary, for more details see `lib/posix-process/inclue/uk/process/events.h`.

### Clone Event:

Raised on `clone()` and `uk_posix_process_run()` (spawn):
```c
struct posix_process_clone_event_data {
        const struct clone_args *cl_args;
        size_t cl_args_len;
        struct uk_thread *child;
        struct uk_thread *parent;
        pid_t ppid;
        pid_t pid;
        pid_t tid;
};

typedef int (*posix_process_clone_handler_fn)(void *event_data);

POSIX_PROCESS_CLONE_HANDLER(flags_mask, handler_fn);
POSIX_PROCESS_CLONE_HANDLER_PRIO(flags_mask, handler_fn, prio);
```
**Important** ⚠️  The semantics of the `flags_mask` have changed. With these changes `flags_mask` merely serves as a way to assist `libposix_process` to perform runtime checks on whether handlers are registered for all flags passed to `clone()`. The clone event invokes all handlers, regardless of the value of `flags_mask` each one of them passed during registration.

Handler parameters have been updated to pass `ppid`, `pid`, and `tid`, as these provide a reference to the equivalent members in the exec and exit events.

### Exec Event:
Raised on `execve()` and `uk_posix_process_run()` (spawn):

```c
struct posix_process_execve_event_data {
        struct uk_thread *thread;
        pid_t pid;
        pid_t tid;
};

typedef int (*posix_process_execve_handler_fn)(void *event_data);

POSIX_PROCESS_EXECVE_HANDLER(handler_fn);
POSIX_PROCESS_EXECVE_HANDLER_PRIO(handler_fn, prio);
```

### Terminate:
Raised on pprocess / pthread termination. This can happen due to various reasons, incl. calls to `exit_group()` & `_exit()`, pthread return, signals, and system shutdown. Upon this event handlers are expected to release all resources.
```c
struct posix_process_exit_event_data {
        struct uk_thread *thread;
        pid_t pid;
        pid_t tid;
};

typedef void (*posix_process_exit_handler_fn)(void *event_data);

POSIX_PROCESS_EXIT_HANDLER(handler_fn);
POSIX_PROCESS_EXIT_HANDLER_PRIO(handler_fn, prio);
```

## Internal API

This API is internally used in `libposix_process`:
```c
int pprocess_raise_clone_event(struct posix_process_clone_event_data *data);
int pprocess_raise_execve_event(struct posix_process_execve_event_data *data);
void pprocess_raise_exit_event(struct posix_process_exit_event_data *data);
```

## Additional Changes

All clone logic previously scattered around libposix_process in different handlers, is now moved into the implementation of `clone()`.

The functionality of `clonetab` has been reduced to flag registration and runtime validation:
* Handler invocation is managed by `uk_event`.
* Registering handlers for the clone event still requires passing clone flags, but these are only used at runtime, to check whether all flags passed to clone can be handled. All handlers are invoked regardless of the flags mask they pass on registration.
* `init` and `term` functions have been deprecated in favor a signle hander. Upon failure resource cleanup is performed by raising an exit event.
* Optional flags (`presence_only`) are now handled internally by `pprocess_raise_clone_event()`.
* Handlers no longer execute in child context. The original intention was to facilitate the initialization of the child's TLS, but in practice handlers explicitly specify TLS by `uk_thread_uktls_var()`.


lib/posix-process/Makefile.uk
```diff
@@ -48,6 +48,7 @@ LIBPOSIX_PROCESS_SRCS-y += $(LIBPOSIX_PROCESS_BASE)/signal/tkill.c
 LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING) += $(LIBPOSIX_PROCESS_BASE)/signal/clone.c|signal
 LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING) += $(LIBPOSIX_PROCESS_BASE)/clone.c
 LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING) += $(LIBPOSIX_PROCESS_BASE)/clonetab.ld
+LIBPOSIX_PROCESS_SRCS-$(CONFIG_LIBPOSIX_PROCESS_MULTITHREADING) += $(LIBPOSIX_PROCESS_BASE)/events.c
```
mogasergiu, line 51: In the commit message s/handlerd/handled

lib/posix-process/clone.c
```diff
@@ -544,13 +360,29 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
 		goto err_free_child;
 #endif /* CONFIG_LIBPOSIX_PROCESS_MULTIPROCESS */
 	}
+	child_tid = ukthread2tid(child);
+	child_pid = ukthread2pid(child);
+
+	if (flags & CLONE_CHILD_SETTID)
```
mogasergiu: Maybe we should set these in user arguments before raising the event instead and do it right before `yield()` when the child is confirmed to have been successfully scheduled? My point in general is mainly about modifying user arguments only at a point where it's needed and keeping them intact until then. I don't think it is necessary for them to be modified before running the handlers.

lib/posix-process/events.c
```diff
@@ -0,0 +1,137 @@
...
+	/* Make sure no handler returns UK_EVENT_HANDLER as that
+	 * would prevent propagating the event to the rest of
+	 * the handlers.
+	 */
+	UK_ASSERT(ret != UK_EVENT_HANDLED);
+
+	/* Set child TLS variables */
+	uk_thread_uktls_var(child, cl_status.is_cloned) = true;
+	uk_thread_uktls_var(child, cl_status.cl_flags)  = cl_args->flags;
+	ret = 0; /* success */
```
mogasergiu: Nit, I know this comes from the original code, but I think you could directly return 0 and `ret` would no longer need the initial assignment at declaration.

lib/posix-process/events.c
```diff
@@ -0,0 +1,137 @@
...
+/*
+ * Raise execve event
+ */
+int pprocess_raise_execve_event(struct posix_process_execve_event_data *data)
+{
+	int ret;
+
+	ret = uk_raise_event(POSIX_PROCESS_EXECVE_EVENT, data);
+
+	/* Make sure no handler returns UK_EVENT_HANDLER as that
```
mogasergiu: s/UK_EVENT_HANDLER/UK_EVENT_HANDLED (ditto `pprocess_raise_clone_event` and `pprocess_raise_exit_event`)

lib/posix-process/clone.c
```diff
@@ -544,13 +360,29 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
...
+		*((pid_t *)cl_args->child_tid) = child_tid;
+
+	if (flags & CLONE_PARENT_SETTID)
+		*((pid_t *)cl_args->parent_tid) = child_tid;
+
+	/* Raise clone event */
+	clone_event = (struct posix_process_clone_event_data) {
+		.cl_args = cl_args,
+		.cl_args_len = cl_args_len,
```
mogasergiu, line 368: ~~I believe you removed the assert for `cl_args_len`, originally in clonetab_init: `UK_ASSERT(cl_args_len >= CL_ARGS_REQUIRED_LEN);`
I would keep it just in case.~~
Edit: Nevermind, it's very unnecessary as there is already a check before that.

lib/posix-process/clone.c
```diff
@@ -586,7 +418,16 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
...
 out:
 	return ret;
-
+err_event:
+	/* Raise exit event to clean up resources */
+	if (cl_args->flags & CLONE_THREAD) {
+		exit_event.thread = child;
+		exit_event.tid = ukthread2tid(child);
+	} else {
+		exit_event.thread = NULL;
```
mogasergiu: Should we still initialize `tid` to something?

lib/posix-process/process.c
```diff
@@ -366,66 +368,84 @@ pid_t uk_posix_process_run(uk_posix_process_mainlike_func fn,
...
 	}
 
 	/* Schedule the process */
-	uk_thread_container_init_fn2(thread,
+	uk_thread_container_init_fn2(child,
 				     (uk_thread_fn2_t)fn,
 				     (void *)(unsigned long)argc,
 				     (void *)argv);
-	uk_sched_thread_add(s, thread);
+	uk_sched_thread_add(s, child);
```
mogasergiu: Huh, ig this is the only place in the codebase that doesn't check for `uk_sched_thread_add`'s return value. Should we add this check to make sure the thread is successfully added?

lib/posix-process/include/uk/process/events.h
```diff
@@ -0,0 +1,257 @@
...
+typedef int (*posix_process_clone_handler_fn)(void *event_data);
+
+struct posix_process_clonetab_entry {
+	__u64 flags_mask;
+};
+
+#define __POSIX_PROCESS_CLONETAB_ENTRY(arg_flags_mask, handler_fn)	    \
+	static const struct posix_process_clonetab_entry		    \
+	__used __section(".posix_process_clonetab") __align(8)		    \
+		__posix_process_clonetab ## clonetab ## _ ## handler_fn = { \
```
mogasergiu: I believe the resulting entries will have "clonetabclonetab" in their name :sweat_smile:, is that intended? Maybe we could use `arg_flags_mask` instead to uniquefy the entry

lib/posix-process/clone.c
michpappas: I see your point and in principle I agree, although I believe that there are no expectations of the values of these parameters if `clone()` fails. Regardless of that, the reason I placed these there is to preserve the previous behavior.  Speaking of handlers, notice that futex's clone handler saves the pointer `cl_arg->child_tid` to `child_tid_clear_ref`, but that's the *pointer* passed by the caller, not the value. In addition, we do pass the values of `pid`, `tid`, `ppid` to handlers, so yes I don't think there's a reason for handlers to expect that value being populated. I'll move these further down, but not right before yield, as that would miss the case of `CLONE_VFORK`.

lib/posix-process/events.c
michpappas: Of course 👍🏼

lib/posix-process/clone.c
michpappas: You will see I follow the same pattern every time I raise the exit event. The reason is that if the exit event refers to a process, the value `thread` is NULL, and the value of `tid` is undefined:
```
/**
 * Data delivered to the handlers of POSIX_PROCESS_EXIT_EVENT.
 * On pthread exit, all fields are populated. On pprocess exit,
 * thread is set to NULL and the value of tid is undefined.
 */
struct posix_process_exit_event_data {
	struct uk_thread *thread;
	pid_t pid;
	pid_t tid;
};
```
Overall I don't have a strong opinion whether we should update the API and set `tid` to zero in this case too.

lib/posix-process/process.c
michpappas: Why not. In fact there's is another unchecked call in `clone()`, I'll fix that too.

lib/posix-process/include/uk/process/events.h
michpappas: Hah, definitely not intended 😄 I'll remove the duplicate part.

mogasergiu (approved):

lib/posix-process/clone.c
mogasergiu: No strong opinion here either, especially since it's also documented. I will leave as unresolved for now as maybe @andreittr can give his 2c as well.

lib/posix-process/clone.c
```diff
@@ -586,7 +424,16 @@ int uk_clone(struct clone_args *cl_args, size_t cl_args_len,
 #endif /* CONFIG_LIBPOSIX_PROCESS_CLONE_PREFER_CHILD */
 out:
 	return ret;
-
+err_rollback:
```
andreittr: If I'm parsing this logic right, exit handlers can be called with as few as 0 clone handlers completing successfully, and should thus gracefully handle uninitialized / partially initialized task state.
To be clear, that's _perfectly fine_, and a clean design overall, though it might be useful to spell out this caveat explicitly in e.g., documentation of exit handlers. I'm positive I would make the silly assumption that the process is "fully-grown" by the time exit handlers run :smile: .

lib/posix-process/events.c
```diff
@@ -0,0 +1,136 @@
...
+	/* Test if we can handle all requested clone flags. Treat flags that we
+	 * handle internally in libposix_process as optional, and mask them out.
+	 * We should not fail if no handler registers them.
+	 */
+	flags = cl_args->flags;
+	flags &= ~(CLONE_SETTLS | CLONE_VFORK | CLONE_DETACHED | CLONE_VM |
+		   CLONE_THREAD | CLONE_CHILD_SETTID | CLONE_PARENT_SETTID);
+
+	posix_process_clonetab_foreach(itr)
+		flags &= ~itr->flags_mask;
```
andreittr: Fairly minor nitpick: AFAICT the set of clone handlers (along with their flags) is determined at compile/link time and does not change during runtime. As such, I think we don't need to walk the clonetab twice on every clone -- we could compute the flag set supported by handlers once (e.g. during init or on 1st call to clone), cache the results, and use it to validate the `flags` arg without a loop.

lib/posix-process/include/uk/process/events.h
```diff
@@ -0,0 +1,257 @@
...
+#ifndef CLONE_IO
+#define CLONE_IO		0x80000000
+#endif /* !CLONE_IO */
+#ifndef CLONE_CLEAR_SIGHAND
+#define CLONE_CLEAR_SIGHAND	0x100000000ULL
+#endif
+
+/**
+ * Data delivered to the handlers of POSIX_PROCESS_CLONE_EVENT
+ * When this event is raised for PID 1, ppid is set to zero.
```
andreittr: I like the documentation about the special case of PID1; could you please also add what happens to the `.parent` field in this case as well? (one could assume it is NULL, but looking at the code PID1 seems to always have a "real" kernel-thread parent; it's good to dispel these assumptions early)

lib/posix-process/clone.c
andreittr: @mogasergiu :eyes: In cases where there's no meaningful value for `tid`, I much prefer leaving (& documenting) it undefined, rather than specifying a dummy/duplicate value. The value of `.thread` already unambiguously tells us whether we're dealing with pprocess or pthread exit event.
Specifying `tid` as `== 0` (or `== pid` or whatever) at this point would only:
- create a difference between _valid_ (`thread == NULL && tid == 0`) and _invalid_ (`thread == NULL && tid != 0`) values for pprocess exit event data => these must either (1) be validated by handlers at runtime, or (2) shown to not matter for each handler in turn
- risk that misguided code will pick up `tid` as the ground source of truth rather than `thread` (they _are_ equivalent in theory, aren't they?), leading to brittle code that can break when fed dodgy input

TL;DR: IMO leaving `tid` undefined for `thread == NULL` is the tighter and preferable type model.

That said, this might be a case where two independent events (one for pprocess exit, one for pthread exit), each with their own args & handlers, makes for a cleaner design; perhaps worth considering.

andreittr (commented):
I like the new design unified around events, it makes following a task's lifetime much easier :+1: . Only a few minor comments.

lib/posix-process/clone.c
michpappas: I'll provide separate handlers for the two events, it's cleaner and leaves no room for ambiguity.

lib/posix-process/events.c
```diff
@@ -0,0 +1,161 @@
...
+static __uk_tls struct {
+	bool is_cloned;
+	__u64 cl_flags;
+} cl_status = { false, 0x0 };
+
+/* clone() flags we can handle. We initialize these to the flags
+ * we manage internally in libposix-process and during init append
+ * the flags handlers have registered for.
+ */
+__u64 cl_flags_handled = CLONE_SETTLS | CLONE_VFORK | CLONE_DETACHED | CLONE_VM |
```
mogasergiu: I think `cl_flags_handled` and `cl_flags_handled_set` could probably also be static.Unless of course you did intend for it to be callable within the lib (since you did not add it in exportsysms.uk), though I highly doubt it for the latter which is in inittab.

lib/posix-process/include/uk/process/events.h
```diff
@@ -0,0 +1,295 @@
...
+/**
+ * Data delivered to the handlers of POSIX_PROCESS_PTEXIT_EVENT.
+ */
+struct posix_process_ptexit_event_data {
+	struct uk_thread *thread;
+	pid_t tid;
+};
+
+/**
+ * POSIX_PROCESS_EXIT_EVENT handler function
```
mogasergiu: And I think you forgot to add "pt" to this typedef and its doc :D

lib/posix-process/events.c
```diff
@@ -0,0 +1,161 @@
...
+UK_EVENT(POSIX_PROCESS_EXECVE_EVENT);
+UK_EVENT(POSIX_PROCESS_PPEXIT_EVENT); /* pprocess */
+UK_EVENT(POSIX_PROCESS_PTEXIT_EVENT); /* pthread */
+UK_EVENT(POSIX_PROCESS_RELEASE_EVENT);
+
+extern const struct posix_process_clonetab_entry _posix_process_clonetab_start[];
+extern const struct posix_process_clonetab_entry _posix_process_clonetab_end;
+
+#define posix_process_clonetab_foreach(itr)				\
+	for ((itr) = DECONST(struct posix_process_clonetab_entry*,	\
```
mogasergiu: Very minor nit. I just realised, but do we actually need this to not be const? (the iterator could be const as well)

lib/posix-process/events.c
```diff
@@ -0,0 +1,161 @@
...
+ */
+static void pprocess_clear_clone_tls(struct uk_thread *child __maybe_unused)
+{
+	UK_ASSERT(child);
+	UK_ASSERT(ukplat_tlsp_get() == child->uktlsp);
+
+	if (!cl_status.is_cloned)
+		return;
+
+	cl_status.is_cloned = false;
```
mogasergiu, line 107: I missed this during first iteration but aren't these already initialised within the TLS ELF section? (I see default values at declaration)

lib/posix-process/events.c
michpappas: Honestly I took it as is from the existing implementation without thinking much about it, but you're right, I can declare `itr` as const and remove `DECONST()`.

lib/posix-process/events.c
michpappas: I don't understand this comment, this function is called on thread termination to cleanup the TLS.

mogasergiu: Oh, my bad, sorry, yes it is on term not init.

lib/posix-process/events.c
```diff
@@ -0,0 +1,161 @@
...
+	bool is_cloned;
+	__u64 cl_flags;
+} cl_status = { false, 0x0 };
+
+/* clone() flags we can handle. We initialize these to the flags
+ * we manage internally in libposix-process and during init append
+ * the flags handlers have registered for.
+ */
+static __u64 cl_flags_handled = CLONE_THREAD | CLONE_VFORK | CLONE_VM |
+				CLONE_CHILD_SETTID | CLONE_PARENT_SETTID|
```
andreittr: There is a missing space before the final `|`.

lib/posix-process/include/uk/process/events.h
```diff
@@ -0,0 +1,295 @@
...
+ * @param event_data Pointer to event data. Handler should cast to
+ *		     struct posix_process_exit_event_data.
+ * @return UK_EVENT_HANDLED_CONT if the event is handled,
+ *	   UK_EVENT_NOT_HANDLED if the event is not handled.
+ *	   Handlersof this event cannot return an error.
+ */
+typedef void (*posix_process_ptexit_handler_fn)(void *event_data);
+
+/**
+ * Registers a handler that is called when a pprocess / pthread is terminated.
```
andreittr: This docstring needs to be updated to refer to pthreads only; similarly the one below to refer to the `PTEXIT` macro.

andreittr (commented):
Looks great, just two _tiny_ cosmetic fixes and we're good to go!

lib/posix-process/include/uk/process/events.h
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * POSIX_PROCESS_PTEXIT_EVENT handler function
+ *
+ * Notice that the exit event may be issued as a means of rollback upon
+ * failure during process creation. Handlers shall not assume a complete
+ * process state.
+ *
+ * @param event_data Pointer to event data. Handler should cast to
+ *		     struct posix_process_exit_event_data.
```
mogasergiu: Ditto here "pt" and "Handlersof" typo below and for ppexit

mogasergiu (approved):
Just a couple of typos.

andreittr (approved):
Thanks!

mogasergiu (approved):

PR #1710, opened 2025-09-25
lib/ukvmem: Do not assert on uninitialized DMA VMA creation

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Some scenarios, where one may not want to pre-zero out an entire DMA area, could exist.

Generally, when doing DMA transfers with a device, some provided buffers may need to be zeroed-out, while some others don't (for example write buffers that are written fully).

In other scenarios, we may want to simply direct-map a region that already contains information that we do not want to be zeroed-out, or just want to direct-map a device memory region.

Thus, do not make it a hard requirement that DMA VMAs are zeroed-out for each faulted/populated page.


michpappas (approved):

PR #1714, opened 2025-10-01
arch/x86: AVX512 support

mogasergiu:
### Prerequisite checklist

### Base target

### Description of changes

Add necessary `ectx` code to handle `AVX512` and early assembly check to enable it.

### DIFF
--- a/plat/native/arch/x86_64/ectx.c
+++ b/plat/native/arch/x86_64/ectx.c
@@ -40,6 +40,8 @@ struct x86_xsave_hdr {
 #define X86_XSAVE_HDR_XSTATE_BV_X87F			(1UL <<  0)
 #define X86_XSAVE_HDR_XSTATE_BV_SSEF			(1UL <<  1)
 #define X86_XSAVE_HDR_XSTATE_BV_AVXF			(1UL <<  2)
+#define X86_XSAVE_HDR_XSTATE_BV_MPX_BNDREGSF		(1UL <<  3)
+#define X86_XSAVE_HDR_XSTATE_BV_MPX_BNDCSRF		(1UL <<  4)
 #define X86_XSAVE_HDR_XSTATE_BV_AVX512_OPMASKF		(1UL <<  5)
 #define X86_XSAVE_HDR_XSTATE_BV_AVX512_ZMM_HI256F	(1UL <<  6)
 #define X86_XSAVE_HDR_XSTATE_BV_AVX512_HI16_ZMMF	(1UL <<  7)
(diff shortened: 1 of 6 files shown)


mogasergiu:
fpush:
- rebase
- MPX-awareness

plat/native/arch/x86_64/ectx.c
```diff
@@ -121,11 +220,53 @@ static void _init_ectx_store(void)
...
-						  avx512_state));
+						  post_avx_state) ||
+			  ectx_size == __offsetof(struct x86_xsave_ctx,
+						  post_avx_state) +
+				       __offsetof(struct _mpx_and_avx512,
+						  avx512_state) ||
+			  ectx_size == __offsetof(struct x86_xsave_ctx,
+						  post_avx_state) +
+				       sizeof(struct x86_avx512_ctx) ||
+			  ectx_size == sizeof(struct x86_xsave_ctx));
```
andreittr, line 250: The role of this assert was to validate if our structs faithfully reflect the SDM by comparing to runtime cpuid info; using an assert is correct in this case, as the check can only fail iff either (1) our code is wrong, or (2) CPUs don't implement the SDM (astronomically unlikely). If neither applies, it _should never happen_, and if it does it gets caught in the dev process.

Orthogonally however, for validating our _assumption_ about a fixed MPX layout (something _not_ in the SDM) versus real-world hardware, we should make the check unconditional and fail boot if it doesn't hold.
Our code may be correct for current CPUs, but break on future ones (which implement the SDM just as faithfully); we _really_ want to know if that happens, even outside of development builds.

edit: I had misunderstood the purpose of the assert; most ectx fields are indeed at dynamic offsets, but we have made the conscious decision to special-case the common combinations and add as necessary. What I proposed would halfway amount to supporting true dynamic ectx as mandated by the SDM, and is thus better left for if/when we make that change. Resolving.

plat/native/arch/x86_64/ectx.c
```diff
@@ -435,14 +628,25 @@ void uk_plat_native_ectx_assert_equal(struct uk_plat_native_ectx *state)
...
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_X87F;
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_SSEF;
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_AVXF;
+		if (have_mpx) {
+			xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_MPX_BNDREGSF;
+			xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_MPX_BNDCSRF;
+		}
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_AVX512_OPMASKF;
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_AVX512_ZMM_HI256F;
+		xbvf[nflags++] = X86_XSAVE_HDR_XSTATE_BV_AVX512_HI16_ZMMF;
```
andreittr: Using an array to iterate over in a previous commit made sense as shorthand for the compiler to unpack, but actually building a cacheline-sized array at runtime IMO is overkill and just gives this function more to do.

Do we need to check the substate fields in order? Could we keep the const array for shorthand, and just add a branch on `have_mpx` to check those fields too?

andreittr (commented):
Since you added init code to enable AVX512 to kvm, I think we should have a commit that does the same to [xen x86 init code](https://github.com/unikraft/unikraft/blob/staging/plat/xen/x86/entry64.S#L121).

andreittr (approved):
Thanks!

andreittr:

PR #1724, opened 2025-10-15
lib/ukfs: Breaking API changes for 0.21

andreittr:
### Description of changes

This changeset updates the ukfs API with two major breaking changes:
- `vopen()` signature changed, it now accepts typed arguments
- `readlink()` is allowed to fail

These changes percolate across the fs stack, and as a consequence:
- the ukfs driver API is updated with drivers needing to declare what types their `vopen()` accepts
- drivers that accept paths as arguments no longer need to "call up" into the VFS layer to open files
- drivers that accept multiple types no longer need to use heuristics to disambiguate
- `posix-vfs` is responsible for interpreting the arguments of `mount()` if the fs driver so requests
- `posix-vfs-fstab` can unambiguously pass blobs in RAM (e.g., initrds) directly to a fs driver

Two minor issues were also fixed in the pass:
- minor docstring inconsistency in `<uk/fs.h>`
- fstab now ignores mkdir failures for special mounts (the logic failed overly eagerly; if the directory _really_ isn't there, the mount will fail instead)

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/posix-vfs-fstab/einitrd.h
+++ b/lib/posix-vfs-fstab/einitrd.h
@@ -17,6 +17,8 @@
 extern const char vfs_einitrd_start[];
 extern const char vfs_einitrd_end;
 
+#define VFS_EINITRD_LEN (&vfs_einitrd_end - vfs_einitrd_start)
+
 #endif /* CONFIG_LIBPOSIX_VFS_FSTAB_EINITRD */
 
 #endif /* __UK_POSIX_VFS_FSTAB_EINITRD_H__ */
--- a/lib/posix-vfs-fstab/fstab.c
+++ b/lib/posix-vfs-fstab/fstab.c
@@ -350,10 +350,9 @@ int fstab_mount_special(const char *path, const char *fstype,
 	int r;
 
 	r = fstab_mkdir(path);
-	if (unlikely(r)) {
-		uk_pr_err("Failed to create %s: %d\n", path, r);
-		return r;
-	}
+	if (unlikely(r))
+		uk_pr_warn("Failed to create %s: %d\n", path, r);
+		/* Try to mount anyway */
 	r = uk_sys_mount(source, path, fstype, flags, data);
 	if (unlikely(r)) {
 		uk_pr_err("Failed to mount special filesystem %s on %s: %d\n",
--- a/lib/posix-vfs-fstab/fstab.c
+++ b/lib/posix-vfs-fstab/fstab.c
@@ -11,6 +11,7 @@
 #include <uk/init.h>
 #include <uk/argparse.h>
 #include <uk/essentials.h>
+#include <uk/fs/driver.h>
 #include <uk/fs/pathutil.h>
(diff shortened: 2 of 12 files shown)


lib/posix-vfs-fstab/einitrd.h
```diff
@@ -17,6 +17,8 @@
 extern const char vfs_einitrd_start[];
 extern const char vfs_einitrd_end;
 
+static const __sz vfs_einitrd_len = &vfs_einitrd_end - vfs_einitrd_start;
```
mogasergiu: Why is it important for this variable to be declared in the header instead of the source file? (also surprised about this being `__sz` instead of `size_t`, since the rest of the code in such libs use libc types :thinking: )

lib/posix-vfs-fstab/fstab.c
```diff
@@ -178,12 +214,11 @@ int fstab_mount(const struct vfs_fstab_entry *e,
...
 		}
-		r = uk_sys_mount((const void *)initrd0->vbase, e->path,
-				 e->fstype, e->flags, e->opts);
+		r = fstab_mount_initrd(e, (const void *)initrd0->vbase,
+				       initrd0->len);
 #if CONFIG_LIBPOSIX_VFS_FSTAB_EINITRD
 	} else if (!strcmp(e->dev, FSTAB_DEV_EINITRD)) {
-		r = uk_sys_mount((const void *)vfs_einitrd_start, e->path,
-				 e->fstype, e->flags, e->opts);
+		r = fstab_mount_initrd(e, vfs_einitrd_start, vfs_einitrd_end);
```
mogasergiu: Did you mean `vfs_einitrd_len` instead of `vfs_einitrd_end`?

mogasergiu (commented):
LGTM. Just a couple of questions.

lib/posix-vfs-fstab/fstab.c
andreittr: Oof, good catch, indeed I did mean `_len` :+1:

lib/posix-vfs-fstab/einitrd.h
andreittr: Re: declaration inside header: mostly to avoid extra `#if`s in the main source file, and to keep its definition close to `_start` and `_end`; but I've no particularly strong preference.

Re: `__sz`, that's to do with not wanting this header to `#include <stdint.h>` just for smth that can hold a ptr diff. If declared in `fstab.c` I'd've use `size_t`.

andreittr: Oh further thought, a variable declaration isn't really warranted here; will replace with a macro, solving both issues.

andreittr:
Pushed addressing comments.

mogasergiu (approved):

michpappas (approved):
Nice, looks definitely cleaner this way.

PR #1729, opened 2025-10-22
lib/ukfs: Introduce node driver template

andreittr:
### Description of changes

This change introduces the ukfs node driver template, a header with common boilerplate code for implementing the live ref driver API on top of driver code that does little to no runtime bookkeeping internally.

In addition, included are two trivial fixes for the live reference template.

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/ukfs/include/uk/fs/template/node.h
+++ b/lib/ukfs/include/uk/fs/template/node.h
@@ -1,0 +1,2006 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+/*
+ * Unikraft filesystem template for node drivers.
+ *
+ * This header file contains boilerplate code for implementing a Unikraft
+ * filesystem driver based on code that natively handles raw references to
+ * filesystem nodes.
+ * This functionality is provided in the form of code generation macros that
+ * define functions conforming to the ukfs live driver template API defined in
+ * `uk/fs/template/live.h`. These functions are then fed through the live driver
+ * template to generate the final ukfile & ukfs operations.
+ *
+ * A raw reference to a filesystem node is understood to be a pair of
+ * identifiers (dev, node), of driver-specific types, that together uniquely
+ * identify a filesystem node in relationship to a "device". A node reference
+ * does not track any runtime state such as refcounting, file state, mount point
+ * status, or filesystem instances.
+ * A node driver is solely responsible for implementing operations focused on
+ * filesystem contents, and for signaling events (if so required).
+ *
+ * Limited support for tracking device and node lifetimes is provided.
+ * Devices are created, opened, and returned by vopen() ready to perform
+ * operations, with vclose() called when finished.
+ * The root node is returned by vopen(), with all other nodes returned by calls
+ * to lookup() or create(). Nodes that compare equal refer to the same
+ * filesystem object until explicitly closed or forgotten.
+ * A node may be "unopened" -- can only perform lookup() -- or "open" -- can
+ * perform all operations.
+ * A device's open nodes are guaranteed to be closed before `vclose()`; any
+ * unopened nodes are implicitly forgotten by `vclose()`.
+ *
+ * When returning the first occurrence of a node:
+ * - vopen() and lookup() return unopened nodes
+ * - create() returns open nodes
+ * When returning nodes previously output but yet to be closed/forgotten:
+ * - lookup() returns nodes as-is
+ * - create() ensures nodes are open before return
+ *
+ * The node driver provides callbacks for node lifetime:
+ * - nopen() -- called when opening a node
+ * - nclose() -- called on open nodes; the node shall not be referenced until
+ *               again returned by the node driver
+ * - nforget() -- called on unopened nodes; the node shall not be referenced
+ *                until again returned by the node driver
+ * Any of these is called at most once for a node value. Multiple "copies",
+ * obtained from lookup() or create(), reference a singular object that will be
+ * opened or cleaned up exactly once.
+ *
+ * Thus, a node may go through one of only 3 lifecycles:
+ * - create() -> [ops] -> nclose()
+ * - lookup()/vopen() -> [lookups] -> nforget()
+ * - lookup()/vopen() -> [lookups] -> nopen() -> [ops] -> nclose()
+ */
+
+#ifndef __UKFS_FS_TEMPLATE_NODE_H__
+#define __UKFS_FS_TEMPLATE_NODE_H__
+
+#include <uk/fs/template/live.h>
+#include <uk/mutex.h>
+
+/* Driver-specific types */
+
+#define UK_FS_TMPL_NODE_OPENDEV(drv) drv##_OPENDEV
+#define UK_FS_TMPL_NODE_CREATE_TARGET(drv) drv##_NODE_CREATE_TARGET
+
+/**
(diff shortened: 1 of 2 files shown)


lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+{									\
+	UK_CTASSERT(node_fs_rename);					\
+	return (node_fs_rename)(rs->vol->dev, rs->node, name, nlen,	\
+				dest->node, dname, dlen,		\
+				flags, mntflags);			\
+}
+
+/* fs readlink */
+
+#define UK_FS_TMPL_NODE_OP_LIVE_FS_READLINK(drv) dr##_LIVE_FS_READLINK
```
mogasergiu: `drv##_LIVE_FS_READLINK`*? (missing "v") :sweat_smile:

lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+ *
+ * Limited support for tracking device and node lifetimes is provided.
+ * Devices are created, opened, and returned by vopen() ready to perform
+ * operations, with vclose() called when finished.
+ * The root node is returned by vopen(), with all other nodes returned by calls
+ * to lookup() or create(). Nodes that compare equal must refer to the same
+ * filesystem object until explicitly closed or forgotten.
+ * A node may be "unopened" -- can only perform lookup() -- or "open" -- can
+ * perform all operations.
+ * - vopen() and lookup() return unopened nodes
```
mogasergiu, line 38: but `vopen` seems to call `_RSTATE_OPEN`?

lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+			if (unlikely(PTRISERR(out)))			\
+				return PTR2ERR(out);			\
+			out_nodes[0] = out;				\
+			return UKFS_SUCCESS;				\
+		}							\
+		ret = (node_fs_lookup)(rs->vol->dev, n, &path[cur], len - cur,\
+				       nod, out_ukfs, &prog);		\
+		if (below && (ret != UKFS_STOP_SYM || n != nod[1]))	\
+			/* n has no live ref & won't be used, forget */	\
+			(node_nforget)(rs->vol->dev, n);		\
```
mogasergiu, line 1317: What is the exact contract for `node_nforget`? Should it be called for every unopened node returned by lookup/vopen that won't be used?

lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+				uk_sched_yield();			\
+				continue;				\
+			}						\
+		} else {						\
+			struct drv##_RSTATE *prev __maybe_unused;	\
+									\
+			ret = drv##_RSTATE_FROMNODE(vs, n, !already_open);\
+			if (unlikely(PTRISERR(ret))) {			\
+				if (already_open)			\
+					(node_nclose)(vs->dev, n);	\
```
mogasergiu: Shouldn't it be forgotten (`node_nforget`) otherwise?

lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+ *
+ * @param drv Driver name
+ * @param devtype Type used for device identifiers
+ * @param nodetype Type used for filesystem node identifiers
+ */
+#define UK_FS_TMPL_NODE_TYPES(drv, devtype, nodetype)			\
+/* Result of opening a volume */					\
+struct UK_FS_TMPL_NODE_OPENDEV(drv) {					\
+	devtype dev; /* Device identifier */				\
+	nodetype rootnode; /* Identifier of root filesystem node */	\
```
mogasergiu, line 83: Could you clarify in the docs as well what "identifier" means?

If it's just an identifier like an inode number (e.g. `__u64`):
Multiple "copies" of the same nodetype value don't need individual cleanup
`nforget()` is just a hint that "we won't look this up again soon" (cache management)

If it's a counted reference (like a refcounted pointer, a `struct node *`):
Each time `node_fs_lookup` returns a `nodetype`, it increments some internal refcount
`nopen()` and `nforget()` decrement that refcount

lib/ukfs/include/uk/fs/template/node.h
```diff
@@ -0,0 +1,1977 @@
...
+			UK_ASSERT(prog);				\
+			/* If live ref exists, pass it up */		\
+			if ((out = drv##_RSTATE_GET(rs->vol, nod[0]))) {\
+				out_nodes[0] = out;			\
+				*nout = cur + prog;			\
+				return ret;				\
+			}						\
+			n = nod[0];					\
+			cur += prog;					\
+			below = 1;					\
```
mogasergiu: Correct me if I am wrong in this case (checking my understanding):
after `node_fs_lookup` returns `nod[0]` (an unopened node ID), and we find it already has an RSTATE (meaning someone already opened that ID) then the existing RSTATE is for the same entity that `node_fs_lookup` just returned. In this case, the ID `nod[0]` is still "in use" by that RSTATE.
We don't need to call `nforget` because the ID won't be freed until that RSTATE is closed.

lib/ukfs/include/uk/fs/template/node.h
andreittr: Whoops :eyes: thanks, that typo would've taken quite a bit to find otherwise :smile:

lib/ukfs/include/uk/fs/template/node.h
andreittr: ~~That's true, I think vopen is meant to return opened nodes; the docs are out of date w/ the code, will double-check and update :+1:~~

lib/ukfs/include/uk/fs/template/node.h
andreittr: Exactly, I mention the 3 possible lifecycles [here](https://github.com/unikraft/unikraft/pull/1729/files#diff-cea516ef7b098d2e0008f4514653b9b4034b9b129799b78e7f5caac083ecd5b8R45).

lib/ukfs/include/uk/fs/template/node.h
andreittr: Indeed, and moreover, since `_RSTATE_OPEN` is tasked with all error handling for the node passed in, I think we're missing one more forget still. Will fix in pass, thanks!

lib/ukfs/include/uk/fs/template/node.h
andreittr: Correct; maybe I should make it clearer in docs that lookup returns _new_ nodes unopened, in-use nodes are returned with unchanged status ("open" in this case).

lib/ukfs/include/uk/fs/template/node.h
andreittr: It indeed means the former; I'll do a pass on the docs to make that clearer.

lib/ukfs/include/uk/fs/template/node.h
andreittr: Wait, no, the docs are correct; the `node_vopen` function returns an _unopened_ rootnode, which the `live_vopen` function then passes to `_RSTATE_OPEN`, which in turn calls `nopen` on the rootnode.

andreittr:
Pushed addressing comments:
- Updated docs for more clarity
- Reworked error handling to have live lookup/create handle node cleanup rather than `_RSTATE_OPEN`

andreittr:
Pushed typo fix.

andreittr:
Pushed rework of lookup symlink error paths.

mogasergiu (approved):

PR #1731, opened 2025-10-27
lib/ukrandom: Fix incorrect size parameters + improvements

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

None

### Description of changes

This series fixes an issue where an incorrect size is passed to when seeding the CSPRNG, as well as when determining whether a seed is set via the cmdline.  It also comes with additional changes that:
1. Decouple the crypto implementation from the swrandom API, which maintains state and provides higher level operations. 
2. Re-implement `chacha_rand32()` to return all values of a given block before generating a new one. That improves efficiency by 16x.
3. Implement encryption tests from [RFC 8439](https://www.rfc-editor.org/rfc/rfc8439.html) to allow end-to-end testing of the high-level API.


lib/ukrandom/chacha.c
```diff
@@ -126,169 +59,42 @@ void uk_salsa20_wordtobyte(__u32 output[16], const __u32 input[16])
...
-		return 0x0;
-	return v[pos % c];
-}
-
-static void chacha_init(struct uk_swrand *r, const __u32 seedv[])
-{
-	UK_ASSERT(r);
-	/* Initialize chacha */
-	__u32 k[8], iv[2], i;
+		ctx->input[i] = ((__u32 *)sigma)[i];
```
andreittr: Technically out of scope, but worth mentioning: since we're casting `sigma` to `__u32 *`, we should be declaring it either with alignment
```c
static const char sigma[16] __nonstring __align(__alignof__(__u32))) = "expand 32-byte k";
```
or directly as a word array
```c
static const __u32 sigma[4] = { 0x61707865, 0x3320646e, 0x79622d32, 0x6b206574 };
```
(the latter would also be endian-agnostic if/when that would be an issue.)

lib/ukrandom/chacha.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+void chacha_init(struct chacha_ctx *ctx,
+		 const __u32 key[CHACHA20_SEED_WORDS],
+		 const __u32 iv[CHACHA20_IV_WORDS],
+		 __u64 counter);
+
+/**
+ * INTERNAL Generate a 32-bit random number
+ */
+__u32 chacha_rand32(struct chacha_ctx *ctx);
```
andreittr, line 31: Since we're declaring only `chacha_init` and `chacha_rand32` as "public", would it be worth converting the internal functions of `chacha.c` to `static` (and possibly drop the `uk_` prefix)?
(`uk_quarterround`, `uk_rotl32`, `uk_salsa20_wordtobyte`)

lib/ukrandom/swrand.h
```diff
@@ -8,12 +8,29 @@
...
+	.lock = UK_SPINLOCK_INITIALIZER(),  \
+}
+
+#define UK_SWRAND_CTX(name)		    \
+	struct uk_swrand_ctx name = UK_SWRAND_CTX_INIT()
+
+struct uk_swrand_ctx {
+	struct uk_random_driver *driver;
+	struct chacha_ctx chacha;
+	uk_spinlock lock;
```
andreittr: I know it was raw `uk_spinlock` in the file to begin with, but I believe `struct uk_spinlock` is [the preferred form](https://github.com/unikraft/unikraft/blob/staging/include/uk/arch/spinlock.h#L38). The former happens to work OK because `__spinlock` is typedef'd to the eponymous struct and `uk_spinlock` is [macroed to `__spinlock`](https://github.com/unikraft/unikraft/blob/staging/lib/uklock/include/uk/spinlock.h#L50), but it's quite brittle and AFAICT the only such instance in the core source tree. We can use this opportunity to fix it.

lib/ukrandom/chacha.c
```diff
@@ -83,18 +83,17 @@ __u32 chacha_rand32(struct chacha_ctx *ctx)
...
-			ctx->input[13]++;
-
-		if (ctx->k < 16) {
-			res = ctx->output[ctx->k];
-			ctx->k += 1;
-			return res;
-		}
-
-		ctx->k = 0;
+	if (ctx->k < 16) {
```
andreittr, line 87: Aside from the obvious improvement in not wasting 90+% of the PRNG stream, I like this form _so_ much better :+1: .
The previous version with the infinite loop was incredibly hard to read.

andreittr (commented):
Overall great changes, I've just a few minor comments.

lib/ukrandom/tests/test_chacha.c
```diff
@@ -7,27 +7,38 @@
...
  *
- * In the block tests we initialize the state using
- * the RFC8439 layout, i.e. 128-bit block and 128-bit
- * nonce, whichis different than the original version of
- * djb that we use. For these test the layout difference
- * doesn't make a difference, as the operations of the
- * block function are agnostic of semantics.
+ * In the quarter round and block function tests, we
+ * initialize the state using the RFC layout, i.e.
+ * 128-bit block counter and 128-bit nonce, which is
```
andreittr: I missed this initially: I think these should be 32-bit and 96-bit, respectively (the entire counter+nonce area is 128b).

lib/ukrandom/chacha.c
michpappas: Good point, I'll push a fix along on the next update. I think declaring as an array of `__u32` is the right way to go.

lib/ukrandom/chacha.h
michpappas: These have to be non-static as they need to be called directly from the tests. I'll add their definitions to the internal header and leave a comment there.

andreittr (approved):
Looks good, thanks!

andreittr:

PR #1732, opened 2025-10-27
lib/ukrandom: Add CSPRNG reseed

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Additional configuration

* `CONFIG_LIBUKRANDOM_RESEED`: Enable periodic reseed.
* `CONFIG_LIBUKRANDOM_RESEED_INTERVAL`: Reseed interval, defaults to 5min.

### Description of changes

Introduce CSPRNG reseed. This adds `uk_random_reseed()` into the API, to allow external components to trigger a reseed in events like system resume. Update `chacha_rand32()` to reseed the CSPRNG once the counter reaches its max value. Introduce  new config options for scheduled reseed for forward secrecy. Without periodic reseed an attacker obtaining the internal state of the CSPRNG can derive past and future outputs.

### DIFF
--- a/lib/ukrandom/random.c
+++ b/lib/ukrandom/random.c
@@ -130,5 +130,10 @@ int uk_random_early_init(struct ukplat_bootinfo __maybe_unused *bi)
 	return 0;
 }
 
+int __check_result uk_random_reseed(void)
+{
+	return uk_swrand_reseed();
+}
+
 UK_BOOT_EARLYTAB_ENTRY(uk_random_early_init, UK_PRIO_AFTER(2));
 #endif /* CONFIG_LIBUKRANDOM_DTB_SEED */
--- a/lib/ukrandom/random.c
+++ b/lib/ukrandom/random.c
@@ -43,9 +43,21 @@
 #include <uk/prio.h>
 #endif /* CONFIG_LIBUKRANDOM_DTB_SEED || CONFIG_LIBUKRANDOM_CMDLINE_SEED */
 
+#if CONFIG_LIBUKRANDOM_RESEED
+#include <uk/arch/time.h>
+#include <uk/errptr.h>
+#include <uk/prio.h>
+#include <uk/sched.h>
+#endif /* CONFIG_LIBUKRANDOM_RESEED */
+
 #include "swrand.h"
 
 struct uk_random_driver *driver;
+
+#if CONFIG_LIBUKRANDOM_RESEED
+static const
+__nsec reseed_interval = CONFIG_LIBUKRANDOM_RESEED_INTERVAL * UKARCH_NSEC_PER_SEC;
+#endif /* CONFIG_LIBUKRANDOM_RESEED */
 
 int __check_result uk_random_fill_buffer(void *buf, size_t buflen)
 {
@@ -98,13 +110,18 @@ int uk_random_init(struct uk_random_driver *drv)
 	return rc;
 }
 
+int __check_result uk_random_reseed(void)
+{
+	return uk_swrand_reseed();
+}
+
 #if CONFIG_LIBUKRANDOM_DTB_SEED || CONFIG_LIBUKRANDOM_CMDLINE_SEED
 /* Do not print a warning if the user did not pass a seed.
  * It may be the case that the application does not require
  * randomness, or that a hardware RNG will be initialized later
  * via init.
  *
- * For the same reasons, do not return an arror as that would cause
+ * For the same reasons, do not return an error as that would cause
  * init to trigger a fatal error. At the worst case the application
  * (or the kernel) will fail later, upon request for entropy.
  */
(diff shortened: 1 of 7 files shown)


lib/ukrandom/swrand.c
```diff
@@ -129,13 +129,50 @@ int uk_swrand_init(struct uk_random_driver **drv)
...
+
 	if (unlikely(!ctx.driver))
 		return -ENODEV;
 
 	uk_spin_lock(&ctx.lock);
 
+	if (chacha_counter(&ctx.chacha) == CHACHA_COUNTER_MAX) {
+		if (ctx.driver == (void *)UK_SWRAND_DRIVER_NONE) {
+			uk_pr_err("Could not reseed, no driver\n");
+			return -EIO;
```
andreittr: Is it intentional to have separate "no driver" error paths in `uk_swrand_reseed` and `uk_swrand_randr`? (i.e., do we want different error codes / messages?)
If not, we could leave error handling to `uk_swrand_reseed` alone.

lib/ukrandom/swrand.c
```diff
@@ -129,13 +129,50 @@ int uk_swrand_init(struct uk_random_driver **drv)
...
 int __check_result uk_swrand_randr(__u32 *val)
 {
+	int ret;
+
 	if (unlikely(!ctx.driver))
 		return -ENODEV;
 
 	uk_spin_lock(&ctx.lock);
 
+	if (chacha_counter(&ctx.chacha) == CHACHA_COUNTER_MAX) {
```
andreittr, line 170: AFAICT this check will trip once the second-to-last output block is generated, leaving the input block with counter == MAX, thus reseeding two blocks early.
To be clear, that's _perfectly fine_ (128 bytes out of 1ZiB is astronomically trivial), and I can see how working around this would just complicate the logic. I just want to make sure this is intentional and not an oversight.

lib/ukrandom/random.c
```diff
@@ -130,10 +145,36 @@ int uk_random_early_init(struct ukplat_bootinfo __maybe_unused *bi)
...
 {
-	return uk_swrand_reseed();
+	int ret;
+
+	while (1) {
+		uk_sched_thread_sleep(reseed_interval);
+		uk_pr_debug("CSPRNG scheduled reseed\n");
+		ret = uk_random_reseed();
+		if (unlikely(ret))
+			UK_CRASH("Could not reseed the CSPRNG\n");
```
andreittr: Since we're crashing, might be worth printing out the error value too.

lib/ukrandom/random.c
```diff
@@ -43,10 +43,20 @@
...
+#include <uk/prio.h>
+#include <uk/sched.h>
+#endif /* CONFIG_LIBUKRANDOM_RESEED */
+
 #include "swrand.h"
 
 struct uk_random_driver *driver;
 
+#if CONFIG_LIBUKRANDOM_RESEED
+static __nsec reseed_interval = CONFIG_LIBUKRANDOM_RESEED_INTERVAL * 1e9;
```
andreittr: Do we intend to adjust at runtime? Otherwise it can be `const`.

(Also: heart skipped a beat when I saw floating point :smile: but sufficient precision around big powers of 10 in doubles + constexpr evaluation + casting to integer type makes it fine)

andreittr (commented):
Looks good, only minor comments.

lib/ukrandom/random.c
michpappas: Good point, I'll add `const`.

> Also: heart skipped a beat when I saw floating point 😄 but sufficient precision around big powers of 10 in doubles + constexpr evaluation + casting to integer type makes it fine

Yes, I think it's fine indeed. But since it caught your eye, I'll do it the proper way and use `UKARCH_NSEC_PER_SEC`.

michpappas:
Rebased from `staging`

lib/ukrandom/swrand.c
michpappas: Yes, mostly for simplicity. The other option was to take into account the value of `k` too but that makes hard assumptions on the implementation of `chacha_rand32()` and IMO is too easy to break (i.e. generate a duplicate block) if something like the increment order changed there. To avoid that, we would normally need a unit test. I did consider that route, but it turns out it's a bit more involved as in theory we should also make sure `uk_swrand_r()` performs operations in the right order (re-seed before generate) and hence closely track the internal state and outputs of the RNG. Technically possible but not sure if it's worth, so I decided to keep it simple instead, at least for now. I 'll leave a comment to make sure it's clear.

andreittr (approved):
Looks great, thanks!

andreittr:

PR #1734, opened 2025-10-28
lib/posix-process: Raise clone event for PID1

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

Raise the clone event upon creation of PID1 to allow libraries to assign resources to the process.
Update clone handlers to manage the creation of parentless processes correctly.

### DIFF
--- a/lib/posix-process/process.c
+++ b/lib/posix-process/process.c
@@ -346,19 +346,67 @@ err_out:
 	return ERR2PTR(ret);
 }
 
+static inline int raise_process_clone_event(struct uk_thread *child,
+					    struct uk_thread *parent)
+{
+	struct posix_process_clone_event_data clone_data;
+	struct clone_args cl_args;
+	pid_t child_tid, child_pid;
+	pid_t parent_tid, parent_pid;
+	int ret;
+
+	UK_ASSERT(child);
+
+	child_tid = ukthread2tid(child);
+	child_pid = ukthread2pid(child);
+
+	if (parent) {
+		parent_tid = ukthread2tid(parent);
+		parent_pid = ukthread2pid(parent);
+	} else {
+		parent_tid = 0;
+		parent_pid = 0;
+	}
+
+	/* We pass the flags used when creating a new process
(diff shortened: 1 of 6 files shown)


lib/posix-process/process.c
```diff
@@ -502,7 +531,11 @@ static int posix_process_init(struct uk_init_ctx *ictx)
 		return PTR2ERR(p);
 	}
 
-	return 0;
+	ret = raise_clone_event(t, NULL);
+	if (unlikely(ret))
+		pprocess_release(tid2pprocess(ukthread2pid(t)));
```
andreittr: Since we have a reference to the new process in `p`, can we not call `pprocess_release(p)` directly?

lib/posix-process/process.c
```diff
@@ -346,6 +346,57 @@ struct posix_process *pprocess_create(struct uk_alloc *a,
 	return ERR2PTR(ret);
 }
 
+static inline int raise_clone_event(struct uk_thread *child,
```
andreittr: Very minor nitpick: the name `raise_clone_event` is slightly misleading, as it suggests it handles any clone event, as opposed to just "new process" clone events. Maybe something like `raise_process_create_event`, though it's an internal function so it doesn't matter that much.

lib/posix-process/process.c
michpappas: I did the rework just for that and forgot to update it 😄 Will update.

lib/posix-process/process.c
michpappas: Yeah it's hard to come up with a good name for this one. In practice it does raise the clone event so I feel that the name should provide that information, but I do see the ambiguity too. I'll rename to `raise_process_clone_event()`. Not very happy with it either but maybe a little bit better.

andreittr (approved):
Thanks!

andreittr:

PR #1735, opened 2025-10-28
lib/posix-process: Use exit_signal from clone()

michpappas:
### Prerequisite checklist

### Base target

 - Architecture(s): [N/A]
 - Platform(s): [N/A]
 - Application(s): [N/A]

### Description of changes

The exit_signal parameter of clone() instructs the kernel to notify the parent with a signal other than SIGCHLD, when
the child returns. Add a new field into struct posix_process and update clone and exit as needed.

Depends-on: #1734

### DIFF
--- a/lib/posix-process/exit.c
+++ b/lib/posix-process/exit.c
@@ -69,15 +69,21 @@ static void pprocess_reparent_children(struct posix_process *pprocess)
 static inline int signal_exit(struct posix_process *pprocess)
 {
 	struct kern_sigaction *ks;
+	__u64 exit_signal;
 
 	UK_ASSERT(pprocess);
 	UK_ASSERT(pprocess->signal);
 
-	ks = KERN_SIGACTION(pprocess, SIGCHLD);
+	if (pprocess->exit_signal)
+		exit_signal = pprocess->exit_signal;
+	else
+		exit_signal = SIGCHLD;
+
+	ks = KERN_SIGACTION(pprocess, exit_signal);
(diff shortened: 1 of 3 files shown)


lib/posix-process/exit.c
```diff
@@ -69,15 +69,21 @@ static void pprocess_reparent_children(struct posix_process *pprocess)
...
+	__u64 exit_signal;
 
 	UK_ASSERT(pprocess);
 	UK_ASSERT(pprocess->signal);
 
-	ks = KERN_SIGACTION(pprocess, SIGCHLD);
+	if (pprocess->exit_signal)
+		exit_signal = pprocess->exit_signal;
+	else
+		exit_signal = SIGCHLD;
```
andreittr, line 80: In what cases would the field be 0? If it's only a special case for PID1 (who does not go through `uk_clone`), then it might be more appropriate to assign `SIGCHLD` to the field by default in `pprocess_create`.

michpappas: Yes exactly, for `pprocess_create()` bu also when we create `PID1` on `posix_process_init()`. Was between two minds on whether to manually set `exit_signal` there and add an assertion in exit, or assigning a default if no exit signal is set before issuing the signal. In the end I concluded on the latter in favor of a single point of assignment. I don't have a very strong preference though as both approaches have pros and cons.

andreittr: Fair enough; might even be more robust as-is if we get even more complex pprocess initialization paths in the future.

andreittr (approved):

andreittr:

PR #1736, opened 2025-10-29
lib/posix-netlink: Introduce netlink library

andreittr:
### Description of changes

This introduces `posix-netlink` a socket library handling the `AF_NETLINK` protocol, and an intermediary for other libraries to register handlers for specific netlink families.

Based on earlier work by @skuenzer and @jschlumpp .

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/posix-netlink/include/uk/netlink/driver.h
```diff
@@ -0,0 +1,133 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
michpappas, line 1: According to our conventions this should be `include/uk/netlink.h`. Since it's only intended for drivers though, it could probably be in `include/uk/netlink/driver.h`.

lib/posix-netlink/include/linux/netlink.h
```diff
@@ -0,0 +1,354 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
```
michpappas, line 1: Since this is an imported header, ideally there should be some traceability info on the commit message.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+		err = ENOMEM;
+		goto err_out;
+	}
+	nl_ctx->a = drv->allocator;
+	nl_ctx->drv = nl_drv;
+#if CONFIG_LIBPOSIX_PROCESS
+	/* FIXME: PID is here like port number and must be unique. It just
+	 *        starts with PID if free.
+	 */
+	nl_ctx->nl_pid = getpid();
```
michpappas: Let's use `uk_sys_getpid()` to avoid making syscalls from the kernel's context

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+	uk_free(nl_ctx->a, nl_ctx);
+err_out:
+	return ERR2PTR(-err);
+}
+
+static int nl_close(posix_sock *s)
+{
+	struct nl_ctx *nl_ctx = sock2nlctx(s);
+	struct uk_streambuf *nlbuf;
+
```
michpappas, line 102: Perhaps it would be worth having a `UK_ASSERT(s)` before deriving the netlink context. Ditto for pointer parameters in the rest of ops.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+static int nl_sockname(posix_sock *s,
+		       struct sockaddr *restrict addr,
+		       socklen_t *restrict addr_len)
+{
+	struct sockaddr_nl *nl_addr;
+	struct nl_ctx *nl_ctx = sock2nlctx(s);
+
+	if (*addr_len < sizeof(*nl_addr)) {
+		/* addr_len is too small, only inform about needed size */
+		goto out;
```
michpappas: If I'm reading [getsockname(2)](https://man7.org/linux/man-pages/man2/getsockname.2.html) correctly, this case should also return `EINVAL`. Also this should probably be `unlikely()`. Same for a few more input checks in other ops.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+	if (addrlen < sizeof(*nl_addr))
+		return -EINVAL;
+
+	nl_addr = (struct sockaddr_nl *)dest_addr;
+	if (unlikely(nl_addr->nl_family != AF_NETLINK || nl_addr->nl_pid != 0))
+		return -EHOSTUNREACH;
+
+	ret = nl_handle(nl_ctx, buf, len);
+	if (unlikely(ret < 0))
+		return -EINVAL;
```
michpappas: Perhaps this should be returning the error value of `nl_handle()`.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+	struct sockaddr_nl *nl_addr;
+	int ret;
+
+	if (addrlen < sizeof(*nl_addr))
+		return -EINVAL;
+
+	nl_addr = (struct sockaddr_nl *)dest_addr;
+	if (unlikely(nl_addr->nl_family != AF_NETLINK || nl_addr->nl_pid != 0))
+		return -EHOSTUNREACH;
+
```
michpappas, line 274: Also a check that the socket is bound would be worth here:
```c
if (unlikely(!nl_ctx->bound))
    return -EDESTADDRREQ;
```

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+		}
+	}
+
+	/* forward message by message */
+	for (nlh = (const struct nlmsghdr *)buf;
+	     NLMSG_OK(nlh, len);
+	     nlh = NLMSG_NEXT(nlh, len)) {
+		uk_pr_debug("Call handler for msg %p (driver %s)\n",
+			    nlh, nl_ctx->drv->libname);
+		ret = nl_ctx->drv->ops->handle(nl_ctx, nlh);
```
michpappas, line 246: Without checking the value of `ret`, this continues execution and returns the error code of the last message sent.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,339 @@
...
+	cpylen = copy2iovec(msg->msg_iov, msg->msg_iovlen,
+			    nlbuf_data(nlbuf), nlbuf_len(nlbuf));
+	nlbuf_free(nlbuf);
+	uk_pr_debug("Message received\n");
+
+	/* Fill name structure */
+	if (msg->msg_name && msg->msg_namelen == sizeof(struct sockaddr_nl)) {
+		nl_addr = msg->msg_name;
+		nl_addr->nl_family = AF_NETLINK;
+		nl_addr->nl_pid = 0;
```
michpappas: I take it that for now we only support process to kernel communication hence pid is always set to 0?

michpappas (changes requested):
Hi @andreittr, minor comments only.

lib/posix-netlink/include/uk/netlink/driver.h
andreittr: The latter sounds good :+1:

lib/posix-netlink/Config.uk
```diff
@@ -0,0 +1,8 @@
+config LIBPOSIX_NETLINK
+	bool "posix-netlink: Abstraction for netlink protocols"
+	select LIBPOSIX_SOCKET
+	help
+	  This microlibrary allows other microlibraries to register a
```
mogasergiu: So, I've always thought that we use LK style for KConfig and I've even pointed it out to you in the past :sweat: . But recently it's been brought up to me that that is actually not the case and we actually do use tab instead of 2 whitespaces for the help message indentation.
CC: @michpappas

lib/posix-netlink/include/uk/netlink_protocol.h
```diff
@@ -0,0 +1,133 @@
...
+
+/*
+ * Creates a static struct posix_netlink_protocol for the AF protocol
+ */
+#define _POSIX_NETLINK_PROTOCOL_REGISTER(lib, prot, vops)		\
+	__used __align(8)						\
+		__section("." _POSIX_NETLINK_PROTOCOL_SECNAME(lib, prot))\
+	static struct posix_netlink_protocol				\
+	_POSIX_NETLINK_PROTOCOL_DRVRNAME(lib, prot) = {			\
+		.protocol = prot,					\
```
mogasergiu: Should `prot`/`vops` here be surrounded by `()`?

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+	struct nl_ctx *nl_ctx;
+	int err;
+
+	UK_ASSERT(drv);
+	UK_ASSERT(family == AF_NETLINK);
+
+	/* AF_NETLINK sockets must be created with SOCK_RAW or SOCK_DGRAM */
+	if (unlikely(((type & ~(SOCK_NONBLOCK | SOCK_CLOEXEC)) != SOCK_RAW) &&
+		     ((type & ~(SOCK_NONBLOCK | SOCK_CLOEXEC)) != SOCK_DGRAM))) {
+		err = EINVAL;
```
mogasergiu: Why do we use positive errno values in this function only to then return the negated version anyway?

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+static inline int nl_handle(struct nl_ctx *nl_ctx, const void *buf, size_t len)
+{
+	const struct nlmsghdr *nlh;
+	size_t tmp_len;
+	int ret = 0;
+
+	UK_ASSERT(nl_ctx);
+	UK_ASSERT(nl_ctx->drv);
+	UK_ASSERT(nl_ctx->drv->ops);
+	UK_ASSERT(nl_ctx->drv->ops->handle);
```
mogasergiu, line 223: so why is only `handle` asserted for but `close`/`create` are actively checked for? maybe this should also be documented that only `handle` is mandatory?

lib/posix-netlink/include/uk/netlink_protocol.h
```diff
@@ -0,0 +1,133 @@
...
+
+static inline struct uk_streambuf *nlbuf_alloc(struct nl_ctx *ctx, size_t len)
+{
+	return uk_streambuf_alloc2(ctx->a, len, sizeof(long),
+				   UK_STREAMBUF_C_WIPEZERO);
+}
+
+#define nlbuf_reserve(nlbuf, len)	uk_streambuf_reserve((nlbuf), (len))
+#define nlbuf_align(nlbuf, align)					       \
+	uk_streambuf_reserve((nlbuf), ALIGN_UP(uk_streambuf_seek((nlbuf)),     \
```
mogasergiu: Why does this "call" (I know it's a macro) `uk_streambuf_seek` twice in a row?

lib/posix-netlink/include/uk/netlink_protocol.h
```diff
@@ -0,0 +1,133 @@
...
+					       (align)) -		       \
+					       uk_streambuf_seek((nlbuf)))
+#define nlbuf_free(nlbuf)		uk_streambuf_free((nlbuf))
+
+static inline void nlbuf_send(struct nl_ctx *ctx, struct uk_streambuf *nlbuf)
+{
+	return uk_mbox_post(ctx->nl_recvqueue, nlbuf);
```
mogasergiu: This confused me and actually made me check if `uk_mbox_post` actually does return a value, but it doesn't. Should we remove the `return`?

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+static int nl_bind(posix_sock *s,
+		   const struct sockaddr *addr,
+		   socklen_t addr_len)
+{
+	const struct sockaddr_nl *nl_addr;
+	struct nl_ctx *nl_ctx = sock2nlctx(s);
+
+	if (unlikely(nl_ctx->bound))
```
mogasergiu: I see the manual says that `bind` returns `EINVAL` in this case?
>  EINVAL The socket is already bound to an address.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+		return -EINVAL;
+	nl_addr = (const struct sockaddr_nl *)addr;
+
+	if (unlikely((nl_addr->nl_family != AF_NETLINK) ||
+		     (nl_addr->nl_pad != 0))
+		return -EINVAL;
+
+	/* We only support the client to connect to us: the kernel */
+	if (nl_addr->nl_pid != 0)
+		return -EHOSTUNREACH;
```
mogasergiu: `bind` does not return `EHOSTUNREACH` (although I have seen syscalls that in practice do return undocumented error codes... looking at you `EFAULT` from `copy_from/to_user`).
Should it instead be `EADDRNOTAVAIL`? But it also sounds a bit like a more unikraft specific case.

lib/posix-netlink/netlink.c
mogasergiu: The comment I think is misleading. I see the man says:
>  The returned address is truncated if the buffer provided is too small; in this case, addrlen will return a value. greater than was supplied to the call.

My understanding is that it does write partial data.

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+{
+	struct nl_ctx *nl_ctx = sock2nlctx(s);
+	struct sockaddr_nl *nl_addr;
+	int ret;
+
+	if (addrlen < sizeof(*nl_addr))
+		return -EINVAL;
+
+	nl_addr = (struct sockaddr_nl *)dest_addr;
+	if (unlikely(nl_addr->nl_family != AF_NETLINK || nl_addr->nl_pid != 0))
```
mogasergiu: `sendto` does not return `EHOSTUNREACH` according to manual... but it does according to POSIX :exploding_head: 
But I think the error here should be:
```C
if (unlikely(nl_addr->nl_family != AF_NETLINK))
	return -EAFNOSUPPORT;
if (unlikely(nl_addr->nl_pid != 0))
	return -EINVAL;  // or ECONNREFUSED
```

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+	size_t left = len;
+	size_t cpylen;
+	size_t i = 0;
+
+	for (i = 0; i < iovec_len && left; ++i) {
+		cpylen = MIN(iovec[i].iov_len, left);
+		memcpy(iovec[i].iov_base, src, cpylen);
+
+		left -= cpylen;
+		src  += cpylen;
```
mogasergiu: void pointer arithmetic :thinking:

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+	struct uk_streambuf *nlbuf;
+	struct nl_ctx *nl_ctx = sock2nlctx(s);
+	size_t cpylen;
+
+	uk_pr_debug("Picking up packet from netlink mbox\n");
+	uk_mbox_recv(nl_ctx->nl_recvqueue, (void **)&nlbuf);
+	if (unlikely(!nlbuf))
+		return -EIO;
+
+	cpylen = copy2iovec(msg->msg_iov, msg->msg_iovlen,
```
mogasergiu: Should we validate the iovec?
```C
	if (unlikely(msg->msg_iovlen > 0 && !msg->msg_iov))
		return -EFAULT;
```

But also, something I am realising just now about both `uk_mbox_recv` and `uk_mbox_post` that are used across this driver.
They are blocking! Shouldn't the socket driver always be non-blocking? i.e. use the `try` variants.
(also `uk_mbox_recv` has its own return value and `EAGAIN` would then probably be more suitable)

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,309 @@
...
+			    nlbuf_data(nlbuf), nlbuf_len(nlbuf));
+	nlbuf_free(nlbuf);
+	uk_pr_debug("Message received\n");
+	return (ssize_t)cpylen;
```
mogasergiu, line 318: Are we sure the message does not get truncated? In that case, according to manual for `recvmsg`, we should set `MSG_TRUNC` flag

lib/posix-netlink/protocol_list.ld
```diff
@@ -0,0 +1,10 @@
+SECTIONS
+{
+	. = ALIGN(0x8);
+	.posix_netlink_protocol_list : {
+		PROVIDE(posix_netlink_protocol_list_start = .);
+		KEEP(*(SORT_BY_NAME(.posix_netlink_protocol_*)));
+		PROVIDE(posix_netlink_protocol_list_end = .);
+	}
+}
+INSERT BEFORE .data;
```
mogasergiu: Should these entries be writable? If not, then:
```
INSERT AFTER .rodata;
```

lib/posix-netlink/netlink.c
```diff
@@ -254,6 +255,16 @@ static ssize_t nl_recvmsg(posix_sock *s, struct msghdr *msg,
 			    nlbuf_data(nlbuf), nlbuf_len(nlbuf));
 	nlbuf_free(nlbuf);
 	uk_pr_debug("Message received\n");
+
+	/* Fill name structure */
+	if (msg->msg_name && msg->msg_namelen == sizeof(struct sockaddr_nl)) {
```
mogasergiu: If buffer is larger, is it not still valid?
If yes then maybe we should use
```C
if (msg->msg_name && msg->msg_namelen >= sizeof(struct sockaddr_nl))
```
and in that case, I think we should update `msg_namelen` to how much we actually wrote to the bigger buffer.

lib/posix-netlink/netlink.c
```diff
@@ -277,6 +277,9 @@ static ssize_t nl_recvfrom(posix_sock *s,
 	struct nl_ctx *nl_ctx = sock2nlctx(s);
 	size_t cpylen;
 
+	if (from && !fromlen)
```
mogasergiu: This commit made me check and actually should we also check for `!buf` and return `-EFAULT`?

lib/posix-netlink/netlink.c
```diff
@@ -295,6 +295,18 @@ static int nl_sockopt(const posix_sock *s __unused,
 	return 0;
 }
 
+static int nl_ioctl(const posix_sock *s __unused,
+		    int request __unused,
+		    void *argp __unused)
+{
+	return -ENOTSUP;
```
mogasergiu: Why do we need a separate commit for this and shouldn't it actually be `ENOTTY`?

andreittr: > Why do we need a separate commit for this

To preserve history and attribution.

> shouldn't it actually be ENOTTY?

I think that error only applies to ioctls specific to character device files; it probably should be `-EINVAL` though, as that's the generic "I don't support the args you passed".

lib/posix-netlink/netlink.c
andreittr: Might have to do with what error values the `create` netlink op is meant to return; the docs in the driver header are misleading and need updating. Will convert to negative and enforce netlink drivers return negative error codes.

lib/posix-netlink/netlink.c
andreittr: Both these asserts are done implicitly by [`posix_sock_get_node()`](https://github.com/unikraft/unikraft/blob/staging/lib/posix-socket/include/uk/socket_driver.h#L99) which `sock2nlctx` uses.
And in general (at least across the file/sock/fs stack), the 1st object argument to operations (akin to `this` or `self`) is always guaranteed non-NULL by caller inlines. The op function may at most wish to assert it's of the correct type.

lib/posix-netlink/Config.uk
```diff
@@ -0,0 +1,10 @@
+config LIBPOSIX_NETLINK
+	bool "posix-netlink: Abstraction for netlink protocols"
+	select LIBPOSIX_SOCKET
+	select LIBUKMPI
+	select LIBUKMPI_MBOX
+	help
+	  This microlibrary allows other microlibraries to register a
+	  netlink protocol family, such as NETLINK_ROUTE, for their netlink
+	  module implementation. This library regsiters the POSIX socket API
```
nderjung: registers*

andreittr:
Pushed update:
- rebasing existing commits on top of current staging -- no changes
- 1 trivial fix commit w/ a missing include
- 1 minor refactor commit addressing review comments and completing the socket API

lib/posix-netlink/netlink.c
```diff
@@ -0,0 +1,443 @@
...
+	struct sockaddr_nl *nl_addr;
+
+	/* We can only ever send messages to the kernel, thus connect only
+	 * validates its args and returns success.
+	 */
+
+	if (unlikely(!addr || addr_len < sizeof(*nl_addr)))
+		return -EINVAL;
+
+	nl_addr = (struct sockaddr_nl *)addr;
```
mogasergiu: Should we use `DECONST`?

mogasergiu (changes requested):

lib/posix-netlink/netlink.c
andreittr: It is indeed a unikraft-specific case for which we need to pick an error ourselves, as the upstream API leaves it undefined; since `EADDRNOTAVAIL` is at least _documented_ in `bind(2)` it's more likely to be gracefully handled by workloads and thus probably the better choice :+1: .

lib/posix-netlink/netlink.c
andreittr: I think this is also one case where unikraft is special; in my (possibly naive) understanding of netlink as a _best effort_ protocol, sending a message to a destination that's not listening should silently drop the packet.
Thus in theory we should _not_ error out here, rather drop the request immediately (since in the current impl there cannot be anyone but the kernel listening). However, I'd rather have it throw an error to notify us of this missing feature, than have it stay silent, however correct the latter may be.
Again as for `bind`, error is up to us: `EINVAL` is already too overloaded, `ECONNREFUSED` is not mentioned in the `sendto(2)` manpage (nor does it technically apply here, since netlink sockets are not connection-oriented), so `EHOSTUNREACH` seems reasonable to me (esp since POSIX _does_ document it for `sendto`).

lib/posix-netlink/netlink.c
andreittr: This is a true cast, not just getting rid of `const`: `const struct sockaddr *` -> `struct sockaddr_nl *`.
But now that you mention it, `nl_addr` should be const too, will fix.

lib/posix-netlink/netlink.c
andreittr: ukfile drivers (including sockets) are allowed to block if they can't help it, and the rest of the file stack works OK. Netlink sockets in particular, because their handlers atm run synchronously with the request, are even more amenable to this, since there's never any "waiting" for replies -- requests generate and write the replies before completion.
If the implementation evolves to support more features, and nonblocking operation will be genuinely required, we can revisit this. In the meantime I'll leave an explanatory comment.

Edit: on 2nd thought I should at least use `_try` versions on `recv`, to prevent silly workloads from deadlocking themselves if they decide to read an empty netlink socket. Polling/notification is still out of scope and only intended for full async support later if needed.

andreittr:
Pushed addressing comments

mogasergiu (approved):

michpappas (approved):
LGTM

PR #1741, opened 2025-12-03
drivers/virtio/fs: Introduce support for VirtIO FS devices

andreittr:
### Description of changes

This changeset adds initial support for VirtIO FS devices -- configuration & synchronous requests -- along with necessary prerequisites to the rest of the virtio driver subsystem.

### Prerequisite checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

### Base target

 - Architecture(s): [e.g. `x86_64` or N/A]
 - Platform(s): [e.g. `kvm`, `xen` or N/A]
 - Application(s): [e.g. `app-python3` or N/A]

### Additional configuration
N/A

PR #1742, opened 2025-12-05
lib/ukfs-virtiofs: Introduce virtiofs filesystem driver

andreittr:
### Description of changes

This changeset adds a ukfs driver for virtiofs, on top of virtio FS devices.

~~Depends on the following PR, and has it included as squashed commit:~~
~~- https://github.com/unikraft/unikraft/pull/1741~~

~~**PLEASE DO NOT MERGE UNTIL REBASE**~~

### Prerequisite checklist

### Base target

### Additional configuration

N/A


lib/ukfs-virtiofs/virtiofs-fuse.c
```diff
@@ -0,0 +1,1057 @@
...
+	if (unlikely(paylen < FIELDLEN(struct fuse_init_out, minor)))
+		return -ENODEV; /* Correct major but no minor; bad device */
+	if (unlikely(out.minor < VIRTIOFS_MIN_FUSE_MINOR)) {
+		/* Init valid, but device is too old; destroy & error */
+		err = -EPROTO;
+		goto err_destroy;
+	}
+	minor = MIN(out.minor, (uint32_t)FUSE_KERNEL_MINOR_VERSION);
+
+	uk_pr_debug("Negociated FUSE version %u.%u\n",
```
mogasergiu: s/Negociated/Negotiated

lib/ukfs-virtiofs/virtiofs-fuse.h
```diff
@@ -0,0 +1,230 @@
...
+	struct uk_virtiofs_dev *dev;
+	uint64_t unique;
+};
+
+/* Lifecycle messages */
+
+/**
+ * Perform FUSE session initialization on `fdev`.
+ *
+ * Sends the FUSE_INIT message and negociates protocol version and flags.
```
mogasergiu: Ditto :rofl:

lib/ukfs-virtiofs/virtiofs-fuse.c
```diff
@@ -0,0 +1,1057 @@
...
+		    FUSE_KERNEL_VERSION, minor);
+
+	/* Check flags */
+	if (unlikely(paylen < FIELDLEN(struct fuse_init_out, flags)))
+		return -ENODEV;
+	if (unlikely(out.flags != VIRTIOFS_FUSE_FLAGS)) {
```
mogasergiu, line 224: I think the device is allowed to return more flags than requested (for backward compatibility). Maybe just `(out.flags & VIRTIOFS_FUSE_FLAGS) != VIRTIOFS_FUSE_FLAGS`?

lib/ukfs-virtiofs/virtiofs-fuse.c
```diff
@@ -0,0 +1,1057 @@
...
+int virtiofs_fuse_syncfs(struct virtiofs_fuse_dev *fdev)
+{
+	struct fuse_in_header hdr = {
+		.nodeid = FUSE_ROOT_ID
+	};
+	struct fuse_syncfs_in in;
+	struct fuse_out_header ohdr;
+	const struct uk_virtiofs_iovec iov[] = {
+		{ &hdr, sizeof(hdr) },
+		{ &in, sizeof(in) },
```
mogasergiu: I think this `in` should be initialized somehow?

lib/ukfs-virtiofs/virtiofs-iomem.h
```diff
@@ -0,0 +1,253 @@
...
+		.iov_base = (void *)addr, /* OK to cast away const */
+		.iov_len = npages * PAGE_SIZE
+	};
+	ssize_t r;
+
+	r = virtiofs_file_write(n, pgoff * PAGE_SIZE, npages * PAGE_SIZE,
+				&iov, 1, 0);
+	if (unlikely(r < 0))
+		return r;
+	/* TODO: handle partial writes gracefully; this is brittle */
```
mogasergiu: shouldn't we have an actual check here not just an assert?

lib/ukfs-virtiofs/virtiofs-fuse.c
```diff
@@ -0,0 +1,1057 @@
...
+			 struct fuse_open_out *out)
+{
+	struct fuse_in_header hdr = {
+		.nodeid = node
+	};
+	struct fuse_out_header ohdr;
+	struct uk_virtiofs_iovec iov[] = {
+		{ &hdr, sizeof(hdr) },
+		{ (void *)in, sizeof(*in) },
+		{ (void *)name, len + 1 },
```
mogasergiu: Why does this one has `len + 1` (emphasis on `+1`) but the `virtiofs_fuse_lookup` doesn't?

mogasergiu (commented):
Had a first look. Just a couple of comments.

lib/ukfs-virtiofs/virtiofs-fuse.c
andreittr: I thought so too, but the entire struct is [just padding](https://github.com/unikraft/unikraft/pull/1742/files#diff-2ffbc0921207e6fee14d63306dc1b13f04157f5de3b6a534d77e55e6e4ec2533R1144) :joy: . Though I guess it's useful to clear the mem before sending the request.

lib/ukfs-virtiofs/virtiofs-iomem.h
andreittr: That's what the TODO is referring to; _fine_ I'll go ahead and make it a write loop.

lib/ukfs-virtiofs/virtiofs-fuse.c
andreittr: Good catch; it should indeed be only `len` :+1:

lib/ukfs-virtiofs/virtiofs-fuse.c
andreittr: The negotiation protocol of fuse INIT flags is unfortunately not as well documented as that for major/minor fuse versions, so we have to make some educated guesses.
Considering that a good number of these flags affect behavior in incompatible ways, I believe they work under a request-reply model: driver requests a set of features, device responds with a subset of those features it supports and _are now active_ on the session. This has been my experience during local testing too.

IMO this is the safer conservative assumption to make, as even if wrong, it's better to get an explicit error prompting investigation and a fix rather than silently running a driver on top of a device with subtly different/broken behavior.

andreittr:

lib/ukfs-virtiofs/virtiofs-iomem.h
```diff
@@ -0,0 +1,274 @@
...
+		if (unlikely(r < 0))
+			break;
+
+		wbytes += r;
+		off += r;
+		rem -= r;
+	} while (rem && !IS_ALIGNED(wbytes, PAGE_SIZE));
+
+	/* If we managed to writeback at least 1 page, report success */
+	wpages = wbytes / PAGE_SIZE;
```
mogasergiu, line 123: Is it possible for `wbytes` to be less than a page but still non-zero? In that case maybe `PAGE_COUNT` could be used instead. Or maybe we should use it either way.

andreittr: > Is it possible for wbytes to be less than a page but still non-zero?

It's definitely possible, and the division is round-down on purpose. The intended effect is to error out; because we have not completed a full page writeback, the caller cannot treat the page as "clean", needing to either retry the writeback or error out.

mogasergiu (approved):

andreittr:
Pushed rebase to current staging following merging of #1741 .

michpappas (approved):

PR #1746, opened 2025-12-12
lib/: Misc build and compiler warning fixes

andreittr:
### Description of changes

This changeset fixes a number of minor build errors and compiler warnings; more details in the commit descriptions, in brief:
- nolibc: switch away from syscalls for `getcwd` and `getdents64` to enable native builds
- posix-netlink: add missing select of `ukstreambuf`
- ukdebug, posix-process, ukfs-ramfs, uksparsebuf: silence compiler warnings

### Prerequisite checklist

### Base target

### Additional configuration

N/A

### DIFF
--- a/lib/posix-process/exit.c
+++ b/lib/posix-process/exit.c
@@ -271,7 +271,7 @@ UK_LLSYSCALL_R_DEFINE(int, exit_group, int, status)
 	uk_sched_thread_exit();
 }
 #else /* !CONFIG_LIBPOSIX_PROCESS_MULTITHREADING */
-void pprocess_exit_stub(int status)
+__noreturn void pprocess_exit_stub(int status)
 {
 	if (pprocess_thread_main) {
 		pprocess_exit_status = status;
--- a/lib/ukfs-ramfs/ramfs.c
+++ b/lib/ukfs-ramfs/ramfs.c
@@ -276,7 +276,9 @@ unsigned char ramfs_dentry_type(const struct ramfs_dentry *dp)
 	case RAMFS_DENT_FILE:
 		return DT_UNKNOWN;
 	default:
-		UK_BUG();
+		uk_pr_err("Invalid dentry type %d; likely bug or corruption\n",
+			  dp->type);
+		return DT_UNKNOWN;
 	}
 }
 
@@ -291,7 +293,9 @@ ino_t ramfs_dentry_inode(const struct ramfs_dentry *dp)
 	case RAMFS_DENT_SPEC:
 		return (ino_t)(uintptr_t)dp;
 	default:
-		UK_BUG();
+		uk_pr_err("Invalid dentry type %d; likely bug or corruption\n",
+			  dp->type);
+		return 0;
 	}
 }
 
@@ -649,8 +653,6 @@ ssize_t ramfs_live_mem(struct ramfs_node *n, enum uk_file_mem_op op,
 		if (unlikely(total != len))
 			return -EINVAL;
 		break;
-	default:
-		return -EINVAL;
 	}
 
 	if (unlikely(!len))
@@ -675,7 +677,7 @@ ssize_t ramfs_live_mem(struct ramfs_node *n, enum uk_file_mem_op op,
 			n->statx.stx_size = off + len;
 		return r;
 	default:
-		UK_BUG(); /* Validate op beforehand */
+		return -EINVAL;
 	}
 }
 
@@ -989,7 +991,9 @@ int ramfs_dir_lookup(struct ramfs_node *n, const char *name, size_t len,
 		out_ukfs->special = dp->target.special;
 		return UKFS_STOP_SPEC;
 	default:
-		UK_BUG();
+		uk_pr_err("Invalid dentry type %d; likely bug or corruption\n",
+			  dp->type);
+		return -EIO;
 	}
 }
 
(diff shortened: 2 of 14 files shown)


mogasergiu (approved):

lib/posix-process/exit.c
```diff
@@ -304,12 +304,12 @@ __noreturn void exit(int status)
 	 * process."
 	 */
 	uk_syscall_e_exit_group(status);
-	UK_BUG(); /* noreturn */
+	UK_CRASH("exit_group syscall returned"); /* noreturn */
```
michpappas: This looks more like a workaround. From what I see the problem originates from the fact that we can't declare syscalls (`uk_sys_exit_group()` in this case) as `__noreturn`. So perhaps we should provide a way to declare syscalls as noreturn? CC: @mogasergiu

Also, generally speaking, as these checks are a common pattern in our codebase: If we make a call *from* a noreturn function *to* a noreturn function, is  a check really necessary? Is it possible for a noreturn function one day to accidentally return?

lib/ukfs-ramfs/ramfs.c
```diff
@@ -276,7 +276,9 @@ unsigned char ramfs_dentry_type(const struct ramfs_dentry *dp)
 	case RAMFS_DENT_FILE:
 		return DT_UNKNOWN;
 	default:
-		UK_BUG();
+		uk_pr_err("Invalid dentry type %d; likely bug or corruption\n",
```
michpappas, line 279: This looks more like a workaround, as `UK_BUG()` should be the proper way to go in such cases. Similarly for the change below, as well as for that in `ukfs-ramfs`. IMO it should not be possible to disable `UK_BUG()`. Unlike `UK_ASSERT()` which is there to catch potential bugs, calling `UK_BUG()` implies certainty that a bug is there. To make things worse, disabling it alters the program flow, allowing execution to proceed if assertions are disabled. Shall we update libukdebug to make `UK_BUG()` always enabled similarly to `UK_CRASH()`?

michpappas (changes requested):

lib/ukfs-ramfs/ramfs.c
andreittr: > UK_BUG() should be the proper way to go in such cases. 

This update gives the ramfs code defined recoverable behavior on invalid directory contents (DT_UNKNOWN or EIO) instead of a hard system crash as before implied by `UK_BUG()`, which one of these is appropriate is not that obvious; at first I thought a hard-crash was OK, but further consideration pushed me more towards erroring out, as the latter would give at least the workload an opportunity to gracefully shutdown.

> Shall we update libukdebug to make UK_BUG() always enabled similarly to UK_CRASH()?

In that case, would there remain a difference between the two, or would they be essentially synonyms?

lib/posix-process/exit.c
andreittr: > If we make a call from a noreturn function to a noreturn function, is a check really necessary?

If both functions are declared noreturn no checks are needed (indeed, the compiler will just optimize them away as dead code). I believe most of these checks you mention lie in places where we _assume_ the function called to not return, without the compiler being able to know about or enforce this in any way (as we have here with the exit syscall).

lib/ukfs-ramfs/ramfs.c
michpappas: > This update gives the ramfs code defined recoverable behavior on invalid directory contents (DT_UNKNOWN or EIO) instead of a hard system crash as before implied by UK_BUG(), which one of these is appropriate is not that obvious; at first I thought a hard-crash was OK, but further consideration pushed me more towards erroring out, as the latter would give at least the workload an opportunity to gracefully shutdown.

Okay, but does this case justify different handling than any other case we would normally use `UK_BUG()` on?

> In that case, would there remain a difference between the two, or would they be essentially synonyms?

Functionally almost, as `UK_CRASH()` takes a paramter. In practice, I believe they have different semantics, as we would normally use `UK_CRASH()` in non-recoverable conditions of different type (panic). As is, `UK_BUG()` is pretty dangerous when assertions are disabled.

lib/posix-process/exit.c
michpappas: Exactly my thought (question was rhetoric, as execution should be deterministic). So in this particular case a check is only meaningful if we don't provide a way to declare syscalls as noreturn. And if we do indeed enable `UK_BUG()` unconditionally, even without declaring the syscall as noreturn, `UK_BUG()` would suffice.

andreittr:
Pushed 1 extra commit silencing a function cast mismatch warning.

lib/ukfs-ramfs/ramfs.c
andreittr: > Okay, but does this case justify different handling than any other case we would normally use UK_BUG() on?

I believe my initial use of `UK_BUG` was overly strict and wrong in this context; the right thing for a filesystem to do when encountering _data corruption_ is to error out, not crash the kernel. This fix goes beyond silencing a warning; it actually changes behavior to something I think makes more sense than before.

> use UK_CRASH() in non-recoverable conditions of different type (panic)

Do you mean that `UK_BUG()` should be used for recoverable conditions (e.g., an oops followed by a graceful shutdown), or that `UK_BUG()` is meant for non-recoverable conditions of a different type than panic? I am in favour of the former, but I think it needs more careful consideration than the scope of this PR allows.

Either way, I believe we agree that a `UK_BUG()` should never return to its calling context, as by using it you are basically saying "if we reach this line something has gone terribly wrong, do not continue". Any case where it made sense to return would be better served by a warning or such.

andreittr:
Pushed update addressing comments with the following changes:
- added 1 commit ([74e3ab9](https://github.com/unikraft/unikraft/pull/1746/commits/74e3ab99e72971fbdec2acb37fa2e1f3872c1084)) to make `UK_BUG()` an unconditional system halt, regardless of assertions being enabled
- updated the posix-process commit to remove UK_BUG -> UK_CRASH change
- updated the title & description of ramfs commit to highlight change in behavior (as that is its main purpose; warning silencing is a side-effect)

michpappas (approved):

mogasergiu (approved):

michpappas:

PR #1751, opened 2026-01-05
support: Generate compilation database on clang builds

michpappas:
### Description of Changes

The [JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) defines a standardized format to provide information on the generation of a build, independently of the underlying build system. The generated database can be used by various tools, a comprehensive list of which is available [here](https://sarcasm.github.io/notes/dev/compilation-database.html#clang-tools). The main motivation of this PR is to perform static analysis on Unikraft with [CodeChecker](https://github.com/Ericsson/CodeChecker/).

Generating a compile db is natively supported by some build systems like [Bazel](https://github.com/hedronvision/bazel-compile-commands-extractor), [CMake](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html), and [ninja](https://ninja-build.org/manual.html#_extra_tools), while tools like [Bear](https://github.com/rizsotto/Bear) and [compiledb](https://github.com/nickdiego/compiledb) can be used on systems that don't provide native support.

Support for Unikraft takes into account that CI workflows are based on [kraftkit](https://github.com/unikraft/kraftkit), which makes it inconvenient to use third-party tools that wrap around the build command.  Instead, Unikraft's build system is updated to natively generate a compile db utilizing clang's `-MJ` flag, when building with clang. The convention is that platforms generate `build/compile_commands.json`, which can be then used by various post-build tools.

### Related Work

N/A

### PR Checklist

### DIFF
--- a/plat/xen/Linker.uk
+++ b/plat/xen/Linker.uk
@@ -22,6 +22,7 @@ else
 XEN_IMAGE       := $(XEN_RAW_IMAGE)
 endif
 XEN_DEBUG_IMAGE := $(XEN_IMAGE).dbg
+XEN_COMPILE_DB_IMAGE := $(BUILD_DIR)/compile_commands.json
 
 XEN_LD_SCRIPT_FLAGS := $(addprefix -Wl$(comma)-dT$(comma),\
 			 $(UK_PLAT_XEN_DEF_LDS))
@@ -87,6 +88,9 @@ $(XEN_IMAGE).gz: $(XEN_IMAGE)
 
 endif
 
+$(XEN_COMPILE_DB_IMAGE): $(XEN_IMAGE)
+	$(call build_compile_db,$(BUILD_DIR),$@)
+
 # register images to the build
 ifeq ($(CONFIG_PLAT_XEN),y)
 UK_DEBUG_IMAGES-y                     += $(XEN_DEBUG_IMAGE)
@@ -102,6 +106,7 @@ UK_IMAGES-$(CONFIG_OPTIMIZE_COMPRESS) += $(XEN_RAW_IMAGE).gz
 else
 UK_IMAGES-$(CONFIG_OPTIMIZE_COMPRESS) += $(XEN_IMAGE).gz
 endif
+UK_IMAGES-y                           += $(XEN_COMPILE_DB_IMAGE)
 endif
 
 # register image to the build
--- a/support/build/Makefile.rules
+++ b/support/build/Makefile.rules
@@ -82,6 +82,12 @@ define measure_loc =
 $(shell $(OBJDUMP) -dl "$(1)" | grep '^/' | grep -v '^/[/*]' | $(SED) -e 's/ [(]discriminator .*[)]//g' | sort -u | wc -l)
 endef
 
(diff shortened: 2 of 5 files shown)


plat/xen/Linker.uk
```diff
@@ -87,6 +88,9 @@ $(XEN_IMAGE).gz: $(XEN_IMAGE)
 
 endif
 
+$(XEN_COMPILE_DB_IMAGE): $(XEN_IMAGE)
+	$(call collate_compile_db,$(BUILD_DIR),$@)
```
andreittr: I believe this was meant to be `build_compile_db` rather than `collate_compile_db` (unless the latter means something different, in which case it needs an implementation in `plat/common/Makefile.rules`).

andreittr (commented):
Looks good overall, thanks! Only one minor quibble

plat/xen/Linker.uk
michpappas: No, it should be `build`, I changed my mind about the name and probably forgot to update. Thanks for catching, will update.

support/build/Makefile.rules
```diff
@@ -396,6 +402,7 @@ endef
 #
 # build_clean $1:target
 build_clean = $(1) $(addsuffix .cmd,$(1))
+build_clean += $(1) $(addsuffix .ukcmpdb.json,$(1))
```
mogasergiu: I think you meant without the `$(1)`?

support/scripts/mkcompiledb.py
```diff
@@ -0,0 +1,37 @@
...
+
+    search_path = opt.path
+    if not os.path.isdir(search_path):
+        print("{}: Invalid path or not a directory".format(search_path))
+        exit(1)
+
+    files = glob.glob(search_path + "/**/*.ukcmpdb.json", recursive=True)
+
+    sys.stdout.write("[\n")
+    if files:
```
mogasergiu, line 30: Nit, I am not sure the files are ever closed. Maybe we should do:
```python
for f in files:
    with open(f, 'r') as fp:
        sys.stdout.write(fp.read().rstrip(",\n") + ",\n")
```
I am also wondering if it's a good idea to use `stdout` or provide a second argument instead for an output file so that we are aligned with the other scripts.

andreittr: All files are guaranteed to have their buffers flushed and closed on interpreter shutdown (whether normal or due to exception) -- and more generally all objects are guaranteed to be cleanly destructed.
If this were a function called in a longer running program, then yes, the files' uncertain destruction time might cause issues, and thus the `with` construct would be warranted; in one-shot scripts like this it'd be overkill IMO.

mogasergiu: ~~Fair, makes sense. Though given that there are hundreds of cmd files in a build (my elfloader build has 635 of them) and a server may want to run multiple builds at the same time, not sure how this would play out with the open files limit.~~ Nvm that is per process

mogasergiu (approved):

andreittr (approved):

andreittr:

PR #1752, opened 2026-01-12
lib: Introduce libukpal, libukpaging, and libukplat_native

michpappas:
This PR introduces Unikraft's Platform Abstraction Layer (PAL) and rearchitects the low-level paging code accordingly.

`libukpal` defines a standardized API between Unikraft libraries and platform implementations. The API abstracts platform-specific behavior, and especially privileged cpu / mmu operations. Other platform functionality like timers, RTC, and consoles should be implemented as platform drivers. In addition to libukpal, this series adds the paging PAL API. This derives from the current paging implementation.

`libukplat_native` implements the uk_pal API to provide support for native and non-paravirtualized environments. In addition, it provides the a platform API (`uk_plat_native`) to other platforms with similar implementation to facilitate code resuse.

Finally, with `libukpal` and the native platform in place, migrate paging implementation from plat/common into `libukpaging`. The library utilizes the PAL paging API to implement page-table management operations, and exports them to other libraries via the uk_paging API.

The rest of this series adapts the codebase to the above changes, and perform minor fixes along the way.

This PR has been tested with:
1. **bincompat:** `catalog/node/20` (x86_64)
2. **native:** `redis/7.0` (arm64, x86_64)

**Notice:** This series does not update Xen. This will be addressed along with the upcoming lcpu changes on a follow up PR.

### Related Work

N/A

### PR Checklist


lib/ukpal/include/uk/pal/paging.h
```diff
@@ -0,0 +1,567 @@
...
+#endif
+
+#ifndef UK_PAL_DIRECTMAP_AREA_END
+#error "UK_PAL_DIRECTMAP_AREA_END not defined"
+#endif
+
+/**
+ * Page table entry type
+ */
+typedef __pte_t	__pte_t;
```
mogasergiu: I think there is a misplaced tab between the two `__pte_t`s

lib/ukpaging/arch/arm64/include/uk/paging/arch.h
```diff
@@ -0,0 +1,102 @@
...
+#include <uk/arch/types.h>
+#include <uk/asm/lcpu.h>
+#include <uk/essentials.h>
+#include <uk/plat/pal/paging.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if CONFIG_LIBUKPAGING
```
mogasergiu, line 18: Why is this guard needed, I see it used in the other headers as well? Additionally, shouldn't we also have header guards here? Ditto x86

lib/ukpaging/arch/x86_64/arch.c
```diff
@@ -0,0 +1,70 @@
...
+		return -ENOMEM;
+
+	start = UK_PAL_PAGE_ALIGN_UP(start + fa_meta_size);
+	pages = (len - fa_meta_size) >> UK_PAL_PAGE_SHIFT;
+
+	dm_off = pgarch_directmap_paddr_to_vaddr(start);
+
+	UK_ASSERT(pt->fa->addmem);
+	rc = pt->fa->addmem(pt->fa, fa_meta, start, pages, dm_off);
+	if (unlikely(rc))
```
mogasergiu: Nit, I know it was like this originally, but maybe we can use this opportunity to just return directly? Ditto for the other pgarch funcs

lib/ukpaging/include/uk/paging.h
```diff
@@ -0,0 +1,832 @@
...
+struct uk_falloc;
+
+struct uk_pagetable {
+	__vaddr_t pt_vbase;
+	__paddr_t pt_pbase;
+
+	struct uk_falloc *fa;
+
+	/* Architecture-dependent part */
+	struct ukarch_pagetable arch;
```
mogasergiu: Is there any point in keeping this? From what I see nothing uses it.

lib/ukpaging/include/uk/paging.h
```diff
@@ -0,0 +1,832 @@
...
+ * Computes the number of page table pages in the given level to map the
+ * specified number of pages
+ *
+ * @param lvl page table level in the range [0..UK_PAL_PT_LEVELS - 1]
+ * @param pages number of pages to map
+ *
+ * @return the number of page table pages required
+ */
+#define UK_PAGING_PT_Lx_PT_PAGES(lvl, pages)			\
+	(DIV_ROUND_UP((pages), UK_PAL_PT_Lx_PTES(level)))
```
mogasergiu: uh, I think it should have been `lvl` instead of `level`. This bug has always existed it seems? 😅

lib/ukpaging/Config.uk
```diff
@@ -0,0 +1,25 @@
+config HAVE_PAGING
+	bool
+	default y if LIBUKPAGING
+
+config HAVE_PAGING_DIRECTMAP
+	bool
+	default y if LIBUKPAGING && ARCH_X86_64
```
mogasergiu: What about ARM64?
Maybe `ARCH_X86_64 || ARCH_X86_64` or should we just drop it?

lib/ukpaging/Makefile.uk
```diff
@@ -0,0 +1,20 @@
...
+
+CINCLUDES-y   += -I$(LIBUKPAGING_BASE)/include
+CINCLUDES-y   += -I$(LIBUKPAGING_BASE)/arch/$(CONFIG_UK_ARCH)/include
+
+CXXINCLUDES-y += -I$(LIBUKPAGING_BASE)/include
+CXXINCLUDES-y += -I$(LIBUKPAGING_BASE)/arch/$(CONFIG_UK_ARCH)/include
+
+LIBUKPAGING_SRCS-y += $(LIBUKPAGING_BASE)/paging.c|isr
+
+#arm64
```
mogasergiu: Nit, let's keep these aligned with the other makefile comments and have the whitespace follow the `#`

lib/ukpaging/include/uk/paging.h
```diff
@@ -0,0 +1,832 @@
...
+ *   - UKPLAT_PAGE_MAPX_ETOOBIG to indicate that the mapping should be retried
+ *     using a smaller page size
+ */
+typedef int (*uk_paging_page_mapx_func_t)(struct uk_pagetable *pt,
+					  __vaddr_t vaddr, __vaddr_t pt_vaddr,
+					  unsigned int level, __pte_t *pte,
+					  void *ctx);
+
+/* Page mapper function return codes */
+#define UKPLAT_PAGE_MAPX_ESKIP		1 /* Skip the current PTE */
```
mogasergiu: There still are some `UKPLAT_PAGE` references in `lib/ukpaging`

drivers/Config.uk
```diff
@@ -6,6 +6,10 @@ menu "Interrupt controller"
 source "$(shell,$(UK_BASE)/support/build/config-submenu.sh -q -o '$(KCONFIG_DIR)/drivers-intctlr.uk' -r '$(KCONFIG_DRIV_BASE)/ukintctlr' -l '$(KCONFIG_DRIV_BASE)/ukintctlr' -e '$(KCONFIG_EXCLUDEDIRS)')"
 endmenu
 
+menu "Platform"
```
mogasergiu, line 9: Couple of typos in the 3rd commit message:
the a platform API
to facilitate code resuse.

plat/native/Config.uk
```diff
@@ -0,0 +1,12 @@
...
+config LIBUKPLAT_NATIVE_PAL
+	bool
+
+# Platforms that call into the uk_plat_native API
+# must select this option
+config LIBUKPLAT_NATIVE
+	bool
+
+config LIBUKPLAT_NATIVE_5LEVEL_PAGING
+	depends on ARCH_X86_64
```
mogasergiu, line 43: I think this is missing the `bool`

While at it, should we maybe have different config menus, exportsyms and Makefile.uk for each architecture? Thinking it might be cleaner to allow each subcomponent to have its own. For example `plat/native/exportsyms.uk` has `uk_plat_native_x86_pg_maxphysaddr` which is unnecessarily shared with ARM64.

Of course this is harmless, so I am just mentioning it.

plat/native/arch/arm64/include/uk/plat/native/arch/paging.h
```diff
@@ -0,0 +1,312 @@
...
+#define UK_PLAT_NATIVE_PAGE_SHIFT	12
+#define UK_PLAT_NATIVE_PAGE_SIZE	0x1000UL
+#define UK_PLAT_NATIVE_PAGE_MASK	(~(UK_PLAT_NATIVE_PAGE_SIZE - 1))
+
+#define UK_PLAT_NATIVE_PAGE_ATTR_PROT_NONE		0x00
+#define UK_PLAT_NATIVE_PAGE_ATTR_PROT_READ		0x01
+#define UK_PLAT_NATIVE_PAGE_ATTR_PROT_WRITE		0x02
+#define UK_PLAT_NATIVE_PAGE_ATTR_PROT_EXEC		0x04
+
+#define PAGE_ATTR_TYPE_MASK		0x07
```
mogasergiu: Should these also have `UK_PLAT_NATIVE`?

plat/native/arch/arm64/paging.c
```diff
@@ -0,0 +1,207 @@
...
+	case PTE_ATTR_IDX(DEVICE_nGnRnE):
+		attr |= PAGE_ATTR_TYPE_DEVICE_nGnRnE;
+		break;
+	case PTE_ATTR_IDX(DEVICE_nGnRE):
+		attr |= PAGE_ATTR_TYPE_DEVICE_nGnRE;
+		break;
+	case PTE_ATTR_IDX(DEVICE_GRE):
+		attr |= PAGE_ATTR_TYPE_DEVICE_GRE;
+		break;
+	case PAGE_ATTR_TYPE_NORMAL_WB_TAGGED:
```
mogasergiu: Shouldn't this also be using `PTE_ATTR_IDX`?

plat/native/arch/arm64/paging.c
```diff
@@ -0,0 +1,207 @@
...
+		attr |= PAGE_ATTR_TYPE_NORMAL_WB_TAGGED;
+		break;
+	default:
+		UK_ASSERT(0 && "Invalid memory type\n");
```
mogasergiu: Should we replace this with a `UK_CRASH`?

plat/native/include/uk/plat/pal/paging.h
```diff
@@ -0,0 +1,217 @@
...
+
+#define UK_PAL_PT_Lx_PTE_SET_PADDR(pte, lvl, paddr)	    \
+	UK_PLAT_NATIVE_PT_Lx_PTE_SET_PADDR(pte, lvl, paddr)
+
+#endif /* !__ASSEMBLY__ */
+
+#define UK_PAL_PAGE_Lx_SHIFT(lvl)   UK_PLAT_NATIVE_PAGE_Lx_SHIFT(lvl)
+#define UK_PAL_PAGE_SHIFT_Lx(shift) UK_PLAT_NATIVE_PAGE_SHIFT_Lx(shift)
+#define UK_PAL_PAGE_Lx_SIZE(lvl)    UK_PLAT_NATIVE_PAGE_Lx_SIZE(lvl)
+#define UK_PAL_PAGE_Lx_MASK	    UK_PLAT_NATIVE_PAGE_Lx_MASK
```
mogasergiu: Forgot the argument here for some of these.

plat/native/include/uk/plat/pal/paging.h
```diff
@@ -0,0 +1,217 @@
...
+{
+	uk_plat_native_tlb_flush();
+}
+#endif /* !__ASSEMBLY__ */
+#endif /* CONFIG_HAVE_PAGING */
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* __UK_PLAT_PAL_H__ */
```
mogasergiu: `__UK_PLAT_PAL_PAGING_H__`

plat/native/pal/include/uk/plat/pal/paging.h
```diff
@@ -0,0 +1,217 @@
...
+
+#define UK_PAL_PAGE_ATTR_PROT_NONE	UK_PLAT_NATIVE_PAGE_ATTR_PROT_NONE
+#define UK_PAL_PAGE_ATTR_PROT_READ	UK_PLAT_NATIVE_PAGE_ATTR_PROT_READ
+#define UK_PAL_PAGE_ATTR_PROT_WRITE	UK_PLAT_NATIVE_PAGE_ATTR_PROT_WRITE
+#define UK_PAL_PAGE_ATTR_PROT_EXEC	UK_PLAT_NATIVE_PAGE_ATTR_PROT_EXEC
+
+#if CONFIG_HAVE_PAGING
+#if !__ASSEMBLY__
+
+#define UK_PAL_DIRECTMAP_AREA_START  UK_PLAT_NATIVE_DIRECTMAP_AREA_START
```
mogasergiu: I think we could have these accessible in asm as well (above the guard)

plat/native/pal/include/uk/plat/pal/paging.h
```diff
@@ -0,0 +1,217 @@
...
+{
+	uk_plat_native_tlb_flush_entry(vaddr);
+}
+
+static inline
+void uk_pal_tlb_flush(void)
+{
+	uk_plat_native_tlb_flush();
+}
+#endif /* !__ASSEMBLY__ */
```
mogasergiu, line 29: I think this has to open/close before/after the `__cplusplus`?

lib/ukfs-ramfs/ramfs-file.h
```diff
@@ -167,31 +167,39 @@ static
 ssize_t ramfs_file_range_acquire(struct ramfs_file_data *file_data,
 				 size_t off, size_t len)
 {
-	const size_t pgoff = off / PAGE_SIZE;
-	const size_t pgend = ALIGN_UP(off + len, PAGE_SIZE) / PAGE_SIZE;
-	const size_t npages = pgend - pgoff;
+	const size_t pgoff;
+	const size_t pgend;
```
mogasergiu: This is strange, uninitizlied const variables at declaration? Did this generate a checkpatch warning? Also the case for virtiofs

drivers/ukconsole/ns16550/ns16550.c
```diff
@@ -420,14 +420,14 @@ static int init(struct uk_init_ctx *ictx __unused)
...
-#if CONFIG_PAGING
+#if CONFIG_HAVE_PAGING
 		/* Map device region */
 		dev.base = uk_bus_pf_devmap(dev.base, dev.size);
 		if (unlikely(PTRISERR(dev.base))) {
 			uk_pr_err("Could not map ns16550\n");
 			return PTR2ERR(dev.base);
 		}
-#endif /* !CONFIG_PAGING */
+#endif /* !CONFIG_HAVE_PAGING */
```
mogasergiu: Without the `!`

mogasergiu: Without the `!`

drivers/ukconsole/pl011/pl011.c
```diff
@@ -351,14 +351,14 @@ static int init(struct uk_init_ctx *ictx __unused)
...
-#if CONFIG_PAGING
+#if CONFIG_HAVE_PAGING
 		/* Map device region */
 		dev.base = uk_bus_pf_devmap(dev.base, dev.size);
 		if (unlikely(PTRISERR(dev.base))) {
 			uk_pr_err("Could not map pl011\n");
 			return PTR2ERR(dev.base);
 		}
-#endif /* !CONFIG_PAGING */
+#endif /* !CONFIG_HAVE_PAGING */
```
mogasergiu: Ditto

mogasergiu (changes requested):

plat/native/include/uk/plat/native/paging.h
```diff
@@ -0,0 +1,184 @@
...
+
+#ifndef __UK_PLAT_NATIVE_PAGING_H__
+#define __UK_PLAT_NATIVE_PAGING_H__
+
+#include <uk/arch/types.h>
+#include <uk/plat/native/arch/paging.h>
+
+#if !__ASSEMBLY__
+
+#ifndef UK_PLAT_NATIVE_PT_Lx_IDX
```
mogasergiu: Guard removal, just `#include <uk/plat/native/arch/paging.h>` inclusion

plat/native/exportsyms.uk
```diff
@@ -0,0 +1,4 @@
+uk_plat_native_paging_init
```
mogasergiu, line 1: I think this should also include pal syms?

lib/ukpaging/arch/arm64/include/uk/paging/arch.h
michpappas: This is needed because some directives like `UK_PAGING_PAGE_SIZE` need to be available even when `libukpaging` is not selected. Previously these were in `uk/arch/paging.h`, while paging-related functionality could be included via `uk/plat/paging.h`. Now `libukpaging` provides both via `uk/paging.h`, which is available unconditionally to `CONFIG_LIBUKPAGING`.

lib/ukpaging/include/uk/paging.h
michpappas: I thought about it too and changed my mind a few times already. Overall I don't have a strong opinion, but if we keep it should probably be renamed to `uk_plat_pagetable`. Or drop it altogether...

Edit: I'll drop

lib/ukpaging/include/uk/paging.h
michpappas: Ooof, nice catch!

lib/ukpal/include/uk/pal/paging.h
```diff
@@ -0,0 +1,567 @@
...
+ *	entry with
+ * @param level the level of the page table [0..UK_PAL_PT_LEVELS - 1] to
+ *	create this entry for
+ * @param template a template page table entry to base the new entry on
+ * @param template_level the level [0..UK_PAL_PT_LEVELS - 1] of the template
+ *
+ * @return the new page table entry
+ */
+__pte_t uk_pal_pte_create(__paddr_t paddr, unsigned long attr,
+			  unsigned int level, __pte_t template,
```
andreittr: The macros in this header file suggest it is meant to be `#include`-able in C++ source files; however, `template` is a [reserved keyword in C++](https://en.cppreference.com/w/cpp/keyword/template.html), and would thus fail compilation (even within an `extern "C"` block).

If we desire C++ compatibility, we should avoid `template` as an identifier, both here in the PAL, as well as platform implementations (there are a couple of instances of the identifier across this PR).

plat/native/arch/arm64/paging.c
michpappas: Yes, looks like a bug in the original implementation, thanks for catching

lib/ukfs-ramfs/ramfs-file.h
michpappas: Ooops, didn't notice the const, lemme fix that.

plat/native/arch/arm64/include/uk/plat/native/arch/paging.h
michpappas: Yes. Notice that `uk_bus_pf_devmap()` previously set the region as `PAGE_ATTR_TYPE_DEVICE_nGnRnE`, and now it is updated to use the `UK_PLAT_NATIVE` type.

lib/ukpaging/Config.uk
michpappas: We can't drop it cause it's used by libukfallocbuddy. What I'm gonna do is:
1. Add a conditional for arm64. This is a general rule when we implement arch-specific config options.
2. Guard the mapping of the directmap area in paging.c with `CONFIG_HAVE_PAGING_DIRECTMAP`

plat/native/Config.uk
michpappas: Yes, you're right about the `bool`, I forgot to add.

> While at it, should we maybe have different config menus, exportsyms and Makefile.uk for each architecture? Thinking it might be cleaner to allow each subcomponent to have its own. For example plat/native/exportsyms.uk has uk_plat_native_x86_pg_maxphysaddr which is unnecessarily shared with ARM64.

If we have more arch-specific directives then possibly, but for the time being we have so few that I think it's not the effort. We can always reconsider later depending on how things evolve.

michpappas:
Note to self: Document the API in `plat/native/paging.h`

plat/native/pal/include/uk/plat/pal/paging.h
michpappas: If we don't have a use-case yet, let's leave it as is until we have one :)

plat/native/exportsyms.uk
michpappas: See the latest update. I introduced a pal/ directory with an exportsyms.uk for pal symbols as well as includes. The pal include directory is enabled conditionally to `LIBUKPLAT_NATIVE_PAL`, but we still need to come up with a way to apply exportsyms.uk.

plat/native/pal/include/uk/plat/pal/paging.h
michpappas: I don't think there is a reason for that, as `__cplusplus` as it is only evaluated to true when compiling `.cpp` files. Its relative order to `__ASSEMBLY__` should therefore not matter.

lib/ukpaging/Makefile.uk
```diff
@@ -0,0 +1,20 @@
+$(eval $(call addlib_s,libukpaging,$(CONFIG_LIBUKPAGING)))
+
+# Headers always enabled to provide basic definitions like
+# PAGE_SIZE to configurations that don't select LIBUKPAGING.
```
andreittr: I like this, having things like `PAGE_SIZE` available without all the paging logic being built :+1: . Misconfigs can easily be caught at linktime.

lib/ukfs-ramfs/ramfs-file.h
```diff
@@ -179,19 +180,23 @@ ssize_t ramfs_file_range_acquire(struct ramfs_file_data *file_data,
...
+	const size_t npages;
 	UK_SPARSEBUF_EMBED_HEAD(fhead, file_data);
 
+	pgoff = off / UK_PAGING_PAGE_SIZE;
+	pgend = ALIGN_UP(off + len, UK_PAGING_PAGE_SIZE) / UK_PAGING_PAGE_SIZE;
+	npages = pgend - pgoff;
```
andreittr: These later assignments will not compile if the vars are kept declared `const` as before.
I think we can keep the same logic with
```c
const size_t pgoff = off / UK_PAGING_PAGE_SIZE;
const size_t pgend = DIV_ROUND_UP(off + len, UK_PAGING_PAGE_SIZE);
const size_t npages = pgend - pgoff;
```
(the use of `DIV_ROUND_UP` is equivalent but much shorter -- something I hadn't initially considered as `PAGE_SIZE` is quite a short macro :sweat_smile: )

This goes similarly for the `pgend` value in `ramfs_file_range_acquire`.

plat/native/exportsyms.uk
mogasergiu: This goes back to my previous [mention](<https://github.com/unikraft/unikraft/pull/1752#discussion_r2682543789>):
 > While at it, should we maybe have different config menus, exportsyms and Makefile.uk for each architecture? Thinking it might be cleaner to allow each subcomponent to have its own. For example plat/native/exportsyms.uk has uk_plat_native_x86_pg_maxphysaddr which is unnecessarily shared with ARM64.

Unless we add a new `exportsyms`-related build rule we will have to use the classic lib eval/addlib_s we typically do for all other libraries. Something maybe like:
`plat/native/Makefile.uk` (and then each native plat can have its own `Makefile.uk`):
```
ifeq ($(CONFIG_ARCH_X86_64),y)
$(eval $(call addlib_s,libukplat_native_arch_x86_64,$(CONFIG_LIBUKPLAT_NATIVE_ARCH_X86_64))
else
ifeq ($(CONFIG_ARCH_ARM_64),y)
$(eval $(call addlib_s,libukplat_native_arch_arm64,$(CONFIG_LIBUKPLAT_NATIVE_ARCH_ARM_64))
endif
endif
```

And then a `plat/native/Config.uk` (and then each native plat can have its own `Config.uk`):
```
if (ARCH_X86_64)
	source "$(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/Config.uk"
endif
if (ARCH_ARM_64)
	source "$(LIBUKPLAT_NATIVE_BASE)/arch/arm64/Config.uk"
endif
```

This help with separation and avoid having a shared Config.uk/Makefile.uk which in the long run would not be sustainable.

Edit: Though maybe for the `if`/`ifeq` we should use the native platform config not the arch.

lib/ukpal/include/uk/pal/paging.h
michpappas: Yes I know, my editor also keeps reminding me too. It is a carry-over from the original PR. Was considering leaving it for later, but as you also mentioned it now creeps into the PAL, so I'll push a fix in the scope of this PR 👍🏼

lib/ukfs-ramfs/ramfs-file.h
michpappas: Yup, I missed the `const`. Will apply the proposed change, thanks.

plat/native/exportsyms.uk
michpappas: I find per-arch sub-libraries unnecessary at this point, and it's also in conflict with our existing conventions. If complexity increases later we can reconsider. In the last version I push a change that also addresses the per-arch `exportsyms.uk` for the pal symbols. The only thing remaining is addressing inconsistencies in headers, but that's something more applicable to discuss in the scope of #1758.

michpappas:
Changes in this update:
* Fix declarations in `ramfs-file.h`.
* Rename `template` variable.
* Append `pal/exportsyms.uk` to `LIBUKPLAT_NATIVE_EXPORTS`. This logic can be also used for arch-specific pal exports.
* Backport `CONFIG_LIBUKPLAT_PAGING` from #1758 and make conditional to `HAVE_PAGING`.
* Make paging implementation conditional to `CONFIG_LIBUKPLAT_PAGING` (was `HAVE_PAGING`). This allows other platforms that also select `HAVE_PAGING` to provide their own paging implementation, while re-using the core header definitions.

mogasergiu:
Thanks for the descriptive updates! A couple of questions:
- Should we move `CONFIG_HAVE_5LEVEL_PAGING` to `lib/Config.uk`?
- Would it make sense to somehow split paging functions intro subcategories, like I did on uklcpu PR, e.g.: maybe pt.h (for pt and pte funcs), tlb.h, addr.h (for paddr/vaddr) and page.h for the PAGE macros?
- Would it be more consistent with arm64 to maybe use x86_64 namespace instead of just x86?
- I believe we have the opportunity to also bring the `virt2phys` function to ukpaging. Would make sense?
- Not sure if it is a good fit for this time, but maybe we could also move the w_xor_x stuff? I don't think it would require bringing in any additional deps since ukpaging already depends on memregion stuff. Up to you.

EDIT:
Reminder about the fixup commit for `x86 efi_post.c`: <https://github.com/unikraft/unikraft/pull/1767/changes/3867adb51dd0856bdaee410c558c73ebbde19571>

EDIT2:
We no longer need the pal exportsyms.uk. I guess it is on me to add it on the platform side for the sysctx/ectx FNSYMs

plat/native/Makefile.uk
```diff
@@ -0,0 +1,28 @@
...
+LIBUKPLAT_NATIVE_EXPORTS-$(CONFIG_LIBUKPLAT_NATIVE_PAL) += $(LIBUKPLAT_NATIVE_BASE)/pal/exportsyms.uk
+
+# TODO Remove as soon as plat dependencies go away
+LIBUKPLAT_NATIVE_CINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
+
+ifeq ($(CONFIG_ARCH_ARM_64), y)
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_PAGING) += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/paging.c|isr
+else
+ifeq ($(CONFIG_ARCH_X86_64), y)
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/paging.c|isr
```
mogasergiu: I guess this also needs to change to `CONFIG_LIBUKPLAT_NATIVE_PAGING`

plat/kvm/Config.uk
```diff
@@ -14,6 +14,10 @@ menuconfig PLAT_KVM
 	imply LIBUKOFW if ARCH_ARM_64
 	select ELF64_TO_32 if ARCH_X86_64
 	select LIBUKRELOC if OPTIMIZE_PIE
+	select LIBUKPLAT_NATIVE
+	select LIBUKPLAT_NATIVE_PAGING
+	select LIBUKPLAT_NATIVE_PAL
+	select LIBUKPLAT_NATIVE_5LEVEL_PAGING if LIBUKPAGING_5LEVEL
```
mogasergiu: I suppose this will change once we have `CONFIG_HAVE_5LEVEL_PAGING` in `lib/Config.uk` and have `LIBUKPLAT_NATIVE_5LEVEL_PAGING` select it instead. This way ukpaging can change its behavior depending on that and not need to have this select where the driver option is selected depending on the framework option.

michpappas:
> Thanks for the descriptive updates! A couple of questions:
> 
>     * Should we move `CONFIG_HAVE_5LEVEL_PAGING` to `lib/Config.uk`?

Yes 👍🏼  I pushed an update that:
* Adds `CONFIG_HAVE_5LEVEL_PT` (I think this name is more accurate) into `plat/Config.uk` 
* Renames `CONFIG_LIBUKPAGING_5LEVEL_PAGING` to `CONFIG_LIBUKPAGING_5LEVEL_PT` and makes it conditional to `CONFIG_HAVE_5LEVEL_PT`.
* Renames `CONFIG_LIBUKPLAT_NATIVE_5LEVEL_PAGING` to `CONFIG_LIBUKPLAT_NATIVE_5LEVEL_PAGING`.

Notice that `CONFIG_HAVE_5LEVEL_PT` is still not selected by `plat/kvm`.

>     * Would it make sense to somehow split paging functions intro subcategories, like I did on uklcpu PR, e.g.: maybe pt.h (for pt and pte funcs), tlb.h, addr.h (for paddr/vaddr) and page.h for the PAGE macros?

I had a look. On the proposed layout:
* pt and pte definitions need to be in separate headers, for namespacing reasons. Specifically, the implementation defines `uk_pal_pt_xxx`, and `uk_pal_pte_xxx`, deriving from `ukplat_pt_xxx` and `ukplat_pte_xxx` respectively. So we would need `pt.h` and `pte.h`.
* There is also `paging.h` with `uk_paging_init()`.

I'll think a bit more and push a dedicated update.

>     * Would it be more consistent with arm64 to maybe use x86_64 namespace instead of just x86?

Which ones in particular? The only public definition that I am aware of is `uk_plat_native_x86_pg_maxphysaddr()`. Is that what you had in mind, or something else? Private definitions are not so critical, because... well, private.

>     * I believe we have the opportunity to also bring the `virt2phys` function to ukpaging. Would make sense?

Good idea, done 👍🏼

>     * Not sure if it is a good fit for this time, but maybe we could also move the w_xor_x stuff? I don't think it would require bringing in any additional deps since ukpaging already depends on memregion stuff. Up to you.

I was planning to have a look at W^X in the scope of the 2nd part of re-arch. When that happens it is possible that it may be integrated into libukpaging.

> EDIT: Reminder about the fixup commit for `x86 efi_post.c`: [3867adb](https://github.com/unikraft/unikraft/commit/3867adb51dd0856bdaee410c558c73ebbde19571)

Integrated changes in the latest update 👍🏼 
 
> EDIT2: We no longer need the pal exportsyms.uk. I guess it is on me to add it on the platform side for the sysctx/ectx FNSYMs

Removed 👍🏼

michpappas:
Changes in this update:  Deprecate `uk/page.h` macros in favor of `UK_PAGING_PAGE_ALIGN` macros.

michpappas:
Changes in this update: `HAVE_PAGING` to plat/Config.uk, make `libukpaging` dependend on `HAVE_PAGING`, make `UKPLAT_NATIVE_PAGING` depend on `HAVE_PAGING`, select `HAVE_PAGING` by the platform.

mogasergiu:
Btw, I have already done this in the lcpu PR, but should we guard the headers with the paging config? Like I do with the corresponding ctx/lcpu configs
e.g.
```C
#if !CONFIG_LIBUKPLAT_NATIVE_ECTX
#error "Native ectx is not available in current configuration"
#endif

```
in `plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h`

michpappas:
> Btw, I have already done this in the lcpu PR, but should we guard the headers with the paging config? Like I do with the corresponding ctx/lcpu configs e.g.
> 
> ```c
> #if !CONFIG_LIBUKPLAT_NATIVE_ECTX
> #error "Native ectx is not available in current configuration"
> #endif
> ```
> 
> in `plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h`

I think adding these errors is not necessary, as generally we make code conditional to Kconfig directives instead. So maybe we should also remove that from the lcpu PR.

michpappas:
Changes in this update:
* Add fine-grained paging configuration in `libukplat_native`.
* Select `LIBUKPLAT_NATIVE_PAGING` only when the platform selects `LIBUKPAGING`.

Combined changes (2-part force-push):
https://github.com/unikraft/unikraft/compare/48674029497ad2ad40c09e4e5c63907eecf3efe0..f78fd94407221052423d9ebd72262bbc627a7894

mogasergiu (approved):
The only thing I find unpleasant about this is the workaround for the device mapping in the platform bus, but I am not sure if we have time to think for a proper solution right now. Otherwise LGTM.

andreittr (approved):
Looks great, thanks!

andreittr:

PR #1758, opened 2026-01-15
{lib, plat/native/arch/x86}: Add libuklcpu, CPU-related `PAL` ops and implement its native `x86_64` driver 

mogasergiu:
### Description of Changes

Migrate LCPU implementation from plat/* into libuklcpu. The library utilizes the PAL CPU-related API to implement CPU-local operations, and exports them to other libraries via the uk_lcpu API.

In addition, this library will also take ownership of various context register contexts management (ectx, sysctx), exceptions, TLSP setting/getting and hardware memory barriers.

Some operations will simply be wrappers over the PAL implementation.

Define the platform-specific CPU-local operations primitives. This is derived from the LCPU API as well as register contexts, exceptions, TLSP handling and hardware memory barriers code.

An important note is that this commit is x86_64 focused. Some primitives for ARM64 are not yet implemented.

One side addition/improvement is that we also move/migrate architecture-specific primitives and definitions to a centralized placed, under `arch/`

That being said, the whole codebase has been updated accordingly following these namespace/API changes.

Depends-on: #1752 

### Related Work

N/A

### PR Checklist


lib/ukpal/include/uk/pal/lcpu.h
michpappas, line 1: I think it would be a good idea to document the PAL API. 

Also in the commit message:
> Define the platform-specific CPU-local operations primitives.

"operations primitives" sounds redundant, maybe one or the other. In this case I think we should use "operations", as we use "primitives" for `arch/util.h`.

include/uk/asm.h
michpappas, line 1: On the commit message: That intro-dude sounds like a really cool guy... 😎 🤙🏼 😜

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
michpappas, line 1: If migrationg code, it's best to do that in the same commit, especially to preserve copyrights.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+
+#define UK_PLAT_NATIVE_EXCEPT_ERR_CTX_SIZE			48
+
+/* uk_plat_native_except_irq_ctx offsets */
+#define UK_PLAT_NATIVE_EXCEPT_IRQ_CTX_OFFSETOF_REGS		0
+#define UK_PLAT_NATIVE_EXCEPT_IRQ_CTX_OFFSETOF_IRQ		8
+
+#define UK_PLAT_NATIVE_EXCEPT_IRQ_CTX_SIZE			16
+
+#if !__ASSEMBLY__
```
michpappas: As this gets chain-included to the kernel API, you may also want to add a `__cplusplus` guard here.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+UK_CTASSERT(__offsetof(struct uk_plat_native_regs, cs) ==
+	    UK_PLAT_NATIVE_REGS_OFFSETOF_CS);
+UK_CTASSERT(__offsetof(struct uk_plat_native_regs, eflags) ==
+	    UK_PLAT_NATIVE_REGS_OFFSETOF_EFLAGS);
+UK_CTASSERT(__offsetof(struct uk_plat_native_regs, rsp) ==
+	    UK_PLAT_NATIVE_REGS_OFFSETOF_RSP);
+UK_CTASSERT(__offsetof(struct uk_plat_native_regs, ss) ==
+	    UK_PLAT_NATIVE_REGS_OFFSETOF_SS);
+
+struct uk_pal_regs;
```
michpappas: Are the various pal declarations in this file a development leftover?

plat/native/arch/x86_64/ectx.c
```diff
@@ -0,0 +1,416 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2018, NEC Europe Ltd., NEC Corporation. All rights reserved.
+ * Copyright (c) 2021, NEC Laboratories Europe GmbH, NEC Corporation.
+ *                     All rights reserved.
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <stdbool.h>
```
michpappas: Maybe it's better to switch to kernel's boolean types

plat/common/x86/syscall.S
```diff
@@ -102,7 +101,7 @@ ENTRY(_ukplat_syscall)
 	 *         8 bytes| |pushed auxsp | |             |
 	 *                v |-------------| v             |
 	 *                ^ |  struct     | ^             |
-	 *                | |  __regs     | |__REGS_SIZEOF|
+	 *                | |  __regs     | |UK_LCPU_REGS_SIZE|
```
michpappas: You can also update the docstring for `__regs`, `ukarch_sysctx`, and `ukarch_ectx`:
```
+        *                  +--------------+ ^
+        *                  |              | | EXECENV_END_ALIGN
+        *                ^ |<------------>| |                 ^
+        *         8 bytes| | pushed auxsp | |                 |
+        *                v |--------------| v                 |
+        *                ^ |    struct    | ^                 |
+        *                | | uk_lcpu_regs | |UK_LCPU_REGS_SIZE|
+        *                | |--------------| v                 |
+        *                | |    struct    | ^                 |uk_syscall_ctx
+        *      struct    | |uk_lcpu_sysctx| |SYSCTX_SIZE      |
+        *  ukarch_execenv| |--------------| v                 |
+        *                | |              | ^                 |
+        *                | |    struct    | |                 |
+        *                | | uk_lcpu_ectx | |ECTX_SIZE        |
+        *                | |              | |                 |
+        *                | |              | |                 |
+        *                | |              | |                 |
+        *                v +--------------+ v                 v
```

plat/native/arch/x86_64/exceptions.c
```diff
@@ -0,0 +1,670 @@
...
+ * rdi, rax, r8, ...), but it does NOT save any floating-point or SSE/AVX
+ * registers. (This would require figuring out in the trap handler code whether
+ * these are available to not risk a #UD trap inside the trap handler itself.)
+ * Hence, you need to be extra careful not to do anything that clobbers these
+ * registers if you intend to return from the handler. This includes calling
+ * other functions, which may clobber those registers.
+ * Of course, if you end your trap handler with a UK_CRASH, knock yourself out,
+ * it's not like the function you came from will ever have the chance to notice.
+ */
+UK_EVENT(UK_PAL_EXCEPT_EVENT_DEBUG);
```
michpappas: Two things:
1. As is, uk_plat_native does not raise uk_plat_native events.
2. Headers of uk_plat_native should only contain declarations of uk_plat_native (and private). Translation to PAL should happen in uk_pal headers.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+{
+	*(__u64 *)((char *)ic + offset) = val;
+}
+
+#define UK_PLAT_NATIVE_EXCEPT_EVENT_ERR_X86_GP				\
+	except_event_err_gp_fault
+#define UK_PLAT_NATIVE_EXCEPT_EVENT_NMI					\
+	except_event_err_nmi
+
+/* Event to raise on debug */
```
michpappas: I think instead of "to raise on debug" (implies that the API user is expected to do it), that should be "raised on debug". Similarly for the rest.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+#define UK_PLAT_NATIVE_EXCEPT_EVENT_DEBUG				\
+	except_event_debug
```
michpappas: `native_except_event_debug` like the rest?

plat/native/arch/x86_64/lcpu.c
```diff
@@ -0,0 +1,464 @@
...
+
+	/* Unlike the other entries, lcpu_start32 must stay the same
+	 * as it is not part of the start16 section.
+	 */
+	apply_start16_reloc((__u64)x86_start16_addr,
+			    START16_MOV_OFF(lcpu_start32, 4),
+			    (__u64)lcpu_start32 - (__u64)x86_start16_addr, 4);
+}
+
+int uk_plat_native_populate_lcpus(void *arg __unused)
```
michpappas: What's the reason for renaming to `populate_lcpus` and not converting the original naming to `uk_plat_native_mp_init()` (and similarly in the PAL)?

plat/common/arm/lcpu.c
```diff
@@ -336,12 +336,12 @@ int lcpu_arch_mp_init(void *arg)
...
 }
 
-int lcpu_arch_start(struct lcpu *lcpu, unsigned long flags __unused)
+int lcpu_arch_start(struct uk_lcpu *lcpu, unsigned long flags __unused)
 {
 	return cpu_on(lcpu->id, lcpu_start_paddr, lcpu);
 }
 
-int lcpu_arch_run(struct lcpu *lcpu, const struct ukplat_lcpu_func *fn,
+int lcpu_arch_run(struct uk_lcpu *lcpu, const struct uk_lcpu_pal_func *fn,
```
michpappas: Not sure `uk_lcpu_pal_func` is quite right here. Same for `uk_lcpu_pal_entry_t`.

plat/native/arch/x86_64/exceptions.c
```diff
@@ -0,0 +1,670 @@
...
+	FILL_IRQ_GATE(220, 1);
+	FILL_IRQ_GATE(221, 1);
+	FILL_IRQ_GATE(222, 1);
+	FILL_IRQ_GATE(223, 1);
+
+	idtptr.limit = sizeof(cpu_idt) - 1;
+	idtptr.base = (__u64)&cpu_idt;
+}
+
+void traps_lcpu_init(struct uk_lcpu *this_lcpu)
```
michpappas: This should be:
1. Namespaced properly: `uk_plat_native_`.
2. Use `except` instead of `traps`.

Similarly for `traps_lcpu_get_except_stack_base()`.

plat/native/Makefile.uk
```diff
@@ -14,3 +14,6 @@ LIBUKPLAT_NATIVE_CINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_ARM_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/paging.c|isr
 
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/paging.c|isr
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/ectx.c|isr
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/exceptions.c|isr
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_ARCH_X86_64) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/lcpu.c
```
michpappas: The sysctx implementation was previously `isr`, so perhaps it's better to keep it in `sysctx.c|isr`.

plat/native/arch/x86_64/lcpu.c
```diff
@@ -0,0 +1,464 @@
...
+	/* Deassert */
+	apic_send_iipi_deassert();
+
+	return 0;
+}
+
+int uk_plat_native_lcpu_post_start(const __uk_lcpuidx lcpuidx[],
+				   unsigned int *num)
+{
+	__uk_lcpuid this_cpu_id = uk_lcpu_id();
```
michpappas: Should't be calling the uklcpu API here, use `uk_plat_native_lcpu_id()`.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+	else
+		uk_arch_disable_irq();
+}
+
+static inline int uk_plat_native_irqs_disabled(void)
+{
+	return !(uk_arch_eflags_get() & UK_ARCH_EFLAGS_IF);
+}
+
+static inline void uk_plat_native_halt(void)
```
michpappas: Surely halt is not performance critical to justify a `static inline`...

lib/ukpal/arch/x86_64/include/uk/pal/lcpu/arch.h
michpappas, line 1: The API needs to be documented, hopefully there should already be documentation to copy from other headers.

lib/ukpal/include/uk/pal/lcpu.h
```diff
@@ -0,0 +1,218 @@
...
+
+void uk_pal_except_pop_nested(void);
+
+__uptr uk_pal_tlsp_get(void);
+
+void uk_pal_tlsp_set(__uptr tlsp);
+
+void uk_pal_disable_irq(void);
+void uk_pal_enable_irq(void);
+void uk_pal_halt(void);
```
michpappas: Perhaps `uk_pal_lcpu_halt()` and move up along with the rest of power management declarations. Similarly for `uk_pal_lcpu_halt_irq()`.

lib/ukpal/include/uk/pal/lcpu.h
```diff
@@ -0,0 +1,218 @@
...
+
+void uk_pal_ectx_init(struct uk_pal_ectx *state);
+
+void uk_pal_ectx_store(struct uk_pal_ectx *state);
+
+void uk_pal_ectx_load(struct uk_pal_ectx *state);
+
+void uk_pal_ectx_assert_equal(struct uk_pal_ectx *state);
+
+/*
```
michpappas: This grouping is good, and makes the header easier to read. But unfortunately on the second half of the file looks like it's been thrown over the fence :(

lib/ukpal/include/uk/pal/lcpu.h
```diff
@@ -0,0 +1,218 @@
...
+/* Arch defined, use helpers to access */
+struct uk_pal_except_irq_ctx;
+
+__u64 uk_pal_except_irq_ctx_get(const struct uk_pal_except_irq_ctx *ic,
+				__sz offset);
+void uk_pal_except_irq_ctx_set(struct uk_pal_except_irq_ctx *ic,
+			       __sz offset, __u64 val);
+
+int uk_pal_lcpu_init(struct uk_lcpu *this_lcpu);
+void __noreturn uk_pal_jump_to(void *sp, uk_lcpu_entry_t entry);
```
michpappas: Maybe `uk_pal_lcpu_jump_to()` is preferrable, as `uk_pal_jump_to()` is quite ambiguous.

lib/ukpal/include/uk/pal/lcpu.h
```diff
@@ -0,0 +1,218 @@
...
+
+#if CONFIG_HAVE_SMP
+int uk_pal_lcpu_populate_lcpus(void *arg);
+int uk_pal_lcpu_start(struct uk_lcpu *lcpu, unsigned long flags);
+int uk_pal_lcpu_post_start(const __uk_lcpuidx lcpuidx[], unsigned int *num);
```
michpappas: This was defined conditionally to `LCPU_ARCH_MULTI_PHASE_STARTUP`, which was only available for `x86_64`. I think this should be also propagated to the PAL via a config option, similarly to what we do with `5LEVEL_PAGING`.

michpappas (changes requested):
First round! Besides the comments below, maybe consider also moving as much of the pal (should be all except from EQUIV) into `platform/native/pal/` (see the latest state of #1752, maybe rebase too).

plat/native/arch/x86_64/exceptions.c
mogasergiu: yes, this is also dev leftover

plat/native/arch/x86_64/lcpu.c
mogasergiu: There were some different plans initially, we were discussing filling in by reference the id <-> idx map (lcpu lib would repeatedly call the function to fill in the IDs).
Of course, now things are different.

plat/native/arch/x86_64/lcpu.c
mogasergiu: Unfortunately this is part of the current ugly fwd dep with lcpu/core.h...
uk_lcpu_id() just fetches the cached ID, while the uk_plat_native is the platform/hw implementation of fetching a CPU's ID by calling CPUID (slow instruction). This should be removed with the separation of the PM/percpu variables concepts.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
mogasergiu: True, but why not :sweat_smile: :laughing:

lib/ukpal/include/uk/pal/lcpu.h
mogasergiu: Yes, things are currently just dumped into the headers as they were migrated. As I will be approaching a more final versiona as we are also finalizing dicussions, I will bring proper documentation/order to everything.

lib/ukpal/include/uk/pal/lcpu.h
mogasergiu: I actually want to move this to uk_arch or something because I also want to get rid of these uk_lcpu typedefs whose usefulness I don't really see... really nothing "lcpu" to this function that was just migrated, it;s just some inline asm jumping somewhere based on a couple of registers while respecting the ABI.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
```diff
@@ -0,0 +1,447 @@
...
+	uk_arch_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	return (ebx >> 24);
+}
+
+static inline __uk_lcpuidx uk_plat_native_lcpu_idx(void)
+{
+	return uk_arch_rdgsbase32(UK_LCPU_IDX_OFFSET);
+}
+
+static inline void uk_plat_native_enable_irq(void)
```
michpappas: Some operations of the `ukplat_lcpu_irq_` API were wrapping around the `local_irq_` API in `plat/common/include/x86/irq.h`. Should we also merge the remaining operations of the `local_irq_` API into the PAL?

plat/kvm/arm/lcpu.c
```diff
@@ -30,23 +30,23 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 #include <stdint.h>
-#include <uk/plat/lcpu.h>
+#include <uk/lcpu.h>
 #include <arm/irq.h>
 #include <arm/cpu.h>
 
-void ukplat_lcpu_enable_irq(void)
+void uk_lcpu_enable_irq(void)
```
michpappas: Looks like some overzealous search and replace happened here. Maybe worth to look elsewhere in case it changed something else it shouldn't had.

lib/ukpal/arch/x86_64/include/uk/pal/lcpu/arch.h
```diff
@@ -0,0 +1,113 @@
...
+
+#ifndef UK_PAL_SYSCTX_OFFSETOF_FSBASE
+#error "UK_PAL_SYSCTX_OFFSETOF_FSBASE undefined"
+#endif
+
+#ifndef UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_CR2
+#error "UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_CR2 undefined"
+#endif
+
+#ifndef UK_PAL_EXCEPT_EVENT_ERR_X86_GP
```
michpappas: We have some inconsistency here, this is the only item with `X86`. We should decide whether to namespace the architecture for arch-specific items, i.e. (`UK_PAL_X86_...`). Personally I'm in favor of this approach as it easy for someone to understand whether part of the API is arch-agnostic (`UK_PAL_XXX`) or arch-specific (`UK_PAL_X86_YYY`).

lib/uklcpu/include/uk/lcpu/core.h
```diff
@@ -0,0 +1,284 @@
...
+#endif /* !__ASSEMBLY__ */
+#endif /* CONFIG_HAVE_SMP */
+
+#define UK_LCPU_MEMBERS_SIZE						\
+	(UK_LCPU_ARCH_OFFSET  + UK_LCPU_ARCH_SIZE)
+#define UK_LCPU_SIZE							\
+	ALIGN_UP(UK_LCPU_MEMBERS_SIZE, UK_ARCH_CACHE_LINE_SIZE)
+
+#if !__ASSEMBLY__
+struct __align(UK_ARCH_CACHE_LINE_SIZE) uk_lcpu {
```
michpappas, line 114: Ideally we should introduce an additional directive for cache line size on PAL. At least on arm64 it's best to derive that on runtime, as some SoC use an 128-bit cache line. In practice it's not urgent as we haven't come across such a system yet, but I think it will be quicker to do that now rather than debug an obscure issue later.

mogasergiu:
Rebase + address comments fpush (ughhh... did not wait long enough and github merged the Compare...)

lib/ukpal/include/uk/pal/lcpu.h
```diff
@@ -0,0 +1,219 @@
...
+#ifndef UK_PAL_SYSCTX_OFFSETOF_TLSP
+#error "UK_PAL_SYSCTX_OFFSETOF_TLSP undefined"
+#endif
+
+/* Guards for uk_pal_except_err_ctx */
+#ifndef UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_STR
+#error "UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_STR undefined"
+#endif
+
+#ifndef UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_TRAPNR
```
michpappas: `UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_TRAPNR`, `UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_ERROR_CODE`, and `UK_PAL_EXCEPT_ERR_CTX_OFFSETOF_FAULT_ADDR` break the arm64 build. We should move non-common fields to the arch headers.

mogasergiu:
Move the CONFIG_HAVE_CPU_MULTI_PHASE_STARTUP config
+
Forgot to migrate/rename CONFIG_LIBUKINTCTLR_ISR_ECTX_ASSERTIONS -> CONFIG_LIBUKPLAT_NATIVE_ECTX_ISR_ASSERTIONS

mogasergiu:
Pushed per "concept" config + moved stray except-related ops to except.c

lib/uknofault/maccess.c
```diff
@@ -7,10 +7,9 @@
 #include <uk/arch/limits.h>
 #include <uk/nofault.h>
 #include <uk/arch/types.h>
-#include <uk/arch/traps.h>
-#include <uk/arch/paging.h>
 #include <uk/event.h>
 #include <uk/config.h>
+#include <uk/paging.h>
```
michpappas, line 12: Also `#include <uk/lcpu.h>`

drivers/virtio/ring/virtio_ring.c
michpappas, line 1: I would propose to leave `drivers/xen/` out of this PR, and have its drivers handled by the followup patches that will update Xen.

lib/uklcpu/include/uk/lcpu/ectx.h
```diff
@@ -0,0 +1,62 @@
...
+	UK_PAL_ECTX_SANITIZE_FN
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_lcpu_ectx;
+
+static inline void uk_lcpu_ectx_load(struct uk_lcpu_ectx *state)
```
michpappas: We should replace ectx load and store with declarations and move the definition to `lcpu.c`, as these functions are called by assembly in `arm64`.

lib/uklcpu/include/uk/lcpu/sysctx.h
```diff
@@ -0,0 +1,64 @@
...
+	UK_PAL_SYSCTX_STORE_FN
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_lcpu_sysctx;
+
+static inline void uk_lcpu_sysctx_load(struct uk_lcpu_sysctx *sysctx)
```
michpappas: We should replace sysctx load and store with declarations and move the definition to `lcpu.c`, as these functions are called by assembly in `arm64`.

include/uk/arch/lcpu.h
michpappas, line 1: With this is being deleted, any files that included it need to be updated to directly include `uk/asm/lcpu.h`. Example: `plat/kvm/arm/pagetable64.S`.

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
michpappas: Because `static inline` makes code hard to debug in gdb, generates incomprehensible compile errors, and chained recompilation once a header is changed, especially in cases like platform headers where headers are nested at multiple levels.
On the other hand the performance benefits in most cases are questionable at best. We should carefully consider what should be `static inline`.

drivers/ukintctlr/xpic/ukintctlr.c
```diff
@@ -37,8 +45,12 @@ void uk_intctlr_xpic_handle_irq(struct __regs *regs, unsigned int irq)
 #else   /* !CONFIG_LIBUKINTCTLR_APIC */
 	pic_ack_irq(irq);
 #endif /* !CONFIG_LIBUKINTCTLR_APIC */
+
+	return 0;
```
michpappas: What should be the return value of uk_event handlers? 🧐

lib/ukdebug/crash.c
michpappas: I think this should now be using the unhandled exception event from uklcpu.

michpappas (changes requested):
@mogasergiu a few more things I came across while working on the arm64 follow up PR.

plat/common/x86/traps.c
```diff
@@ -23,51 +24,69 @@
...
+	     UK_LCPU_EXCEPT_EVENT_ERR_MATH)
+DECLARE_TRAP_EC(security_error, "control protection",
+		UK_LCPU_EXCEPT_EVENT_ERR_SECURITY)
 
-UK_EVENT(UK_EVENT_UNHANDLED_EXCEPTION);
+UK_EVENT(UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPT);
 
-void do_unhandled_trap(int trapnr, char *str, struct __regs *regs,
-		unsigned long error_code)
+void do_unhandled_trap(int trapnr, char *str, struct uk_lcpu_regs *regs,
```
michpappas: Shouldn't unhandled exceptions also be managed by `plat/native/`?

mogasergiu:
Force push that adds:
- plat native docs
- pal docs
- renaming `UK_LCPU_[ECTX|SYSCTX]_*_FN` -> `UK_LCPU_[ECTX|SYSCTX]_*_FNSYM` to make things clearer
- use UK_ARCH macros for `TRAP_ENTRY` macros in `cpu_vectors_x86_64.S` instead of hardcoded trap numbers (forgot to do this in first revision)

lib/ukpal/include/uk/pal/lcpu.h
mogasergiu: I have removed these macros as they are not used anywhere, not even in plat, unlike thier REGS/SYSCTX counterpart.

mogasergiu:
Force pushed:
- new commit removing apic.h, moving related macro-defs to arch.h, and redistributing used implementations in the correct C files
- undoing some unnecessary overzealous sed changes on ARM64
- obsoleted `plat/common/include/uk/plat/common/cpu.h` + `plat/common/include/x86/cpu.h`

plat/common/x86/traps.c
mogasergiu: Yes this file is unused I forgot to delete it. Thanks.

drivers/virtio/ring/virtio_ring.c
mogasergiu: I think just the drivers directory is fine as the diff would have been the same. It's the `plat/` that needed to be left alone.

mogasergiu:
Force pushed:
- some ARM files to include `uk/asm/lcpu.h`
- fixed the wrong use of the unhandled exception event name
- removed `plat/common/x86/traps.c`
- fix `uk_intctlr_xpic_handle_irq` to return `UK_EVENT_HANDLED`

mogasergiu:
fpush with:
- vga removed macros fixup to make SMP build
- lcpu pm ops commit

mogasergiu:
fpush for:
- make a bunch of trivial functions/signatures __isr as required

mogasergiu:
fpush:
- namespace arch specific regs/sysctx/except defs with `X86_64`

mogasergiu:
fpush:
- (almost) deletes no longer necessary cpu.h

lib/ukpal/include/uk/pal/lcpu.h
michpappas: The comment still applies though to the static inlines, e.g. `uk_plat_native_except_err_ctx_set_trapnr()` (we don't have trap numbers in arm64, but exception ids)

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,316 @@
...
+{
+	uk_arch_disable_irq();
+}
+
+/**
+ * Save current IRQ state and disable interrupts.
+ * Returns 1 if interrupts were enabled, 0 if already disabled.
+ * Use with uk_plat_native_restore_irqf() for critical sections.
+ *
+ * @return 1 if RFLAGS.IF was set, 0 otherwise
```
michpappas, line 222: I think this should return `flags` as-is, to allow the caller to later restore (in fact I wonder how it worked until now on x86). If interpretation is required by the caller, then perhaps we should provide a function to accept `flags` as a parameter, and interpret that.

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,316 @@
...
+	__u64 irq)
+{
+	ctx->irq = irq;
+}
+
+/**
+ * Enable x86_64 interrupts (STI instruction).
+ * Sets RFLAGS.IF, allowing maskable hardware interrupts.
+ */
+__isr static inline void uk_plat_native_enable_irq(void)
```
michpappas, line 202: I think the IRQ and auxsp functions can be in their own files, unless you had a specific reason to have them together with exceptions.

plat/native/exportsyms.uk
```diff
@@ -0,0 +1,81 @@
...
+uk_plat_native_lcpu_run
+uk_plat_native_save_irqf
+uk_plat_native_set_auxsp
+uk_plat_native_lcpu_start
+uk_plat_native_sysctx_load
+uk_plat_native_sysctx_store
+uk_plat_native_tlsp_get
+uk_plat_native_tlsp_set
+uk_plat_native_lcpu_wakeup
+uk_plat_native_wmb
```
michpappas: `uk_plat_native_wmb` looks like a leftover. Similar for the rest of barrier functions both on the scope of `uk_pal` and `uk_plat_native`.

plat/native/arch/x86_64/lcpu.c
```diff
@@ -0,0 +1,506 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/*
+ * Copyright (c) 2022, Karlsruhe Institute of Technology (KIT)
+ *                     All rights reserved.
+ * Copyright (c) 2022, University POLITEHNICA of Bucharest.
+ *                     All rights reserved.
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ *
```
michpappas: No new line after the copyright

plat/native/arch/x86_64/lcpu.c
```diff
@@ -0,0 +1,506 @@
...
+
+/* We only support x2APIC at the moment */
+#define apic_enable		x2apic_enable
+#define apic_send_ipi		x2apic_send_ipi
+#define apic_send_sipi		x2apic_send_sipi
+#define apic_send_iipi		x2apic_send_iipi
+#define apic_send_iipi_deassert x2apic_send_iipi_deassert
+#endif /* CONFIG_HAVE_SMP */
+
+static int plat_native_lcpu_pm_ops_register(void);
```
michpappas, line 208: Nit: A new line after the forward declaration would make the code easier to read.

lib/uklcpu/include/uk/lcpu/pm.h
```diff
@@ -0,0 +1,98 @@
...
+int uk_lcpu_start(const __u32 lcpuidx[],
+		  unsigned int *num,
+		  __u64 sp[], __u64 entry[],
+		  unsigned long flags);
+#endif /* CONFIG_HAVE_SMP */
+
+/**
+ * Halts the current logical CPU.
+ */
+void uk_lcpu_halt(void);
```
michpappas: Should be annotated as `__noreturn`

lib/uklcpu/include/uk/lcpu/core.h
```diff
@@ -0,0 +1,281 @@
...
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_LCPU_CORE_H__
+#define __UK_LCPU_CORE_H__
+
+#include <uk/arch.h>
+#include <uk/essentials.h>
```
michpappas, line 17: Needs to include `uk/arch/util.h` for `UK_ARCH_CACHE_LINE_SIZE`.

lib/ukpal/include/uk/pal/except.h
```diff
@@ -0,0 +1,233 @@
...
+ * @return Auxiliary stack pointer value
+ */
+__uptr uk_pal_get_auxsp(void);
+
+/**
+ * Get exception stack base address.
+ *
+ * @return Base of exception stack for current CPU
+ */
+__isr __uptr uk_pal_traps_get_except_stack_base(void);
```
michpappas: Should we also adapt this function's name to use the `except` convention instead of `traps`?

plat/native/arch/x86_64/exceptions.c
```diff
@@ -215,32 +191,182 @@ void ukarch_pop_nested_exceptions(void)
...
+		uk_pr_crit("Unhandled Trap %d (%s), error code=0x%lx\n",
+			   trapnr, x86_exception_table[trapnr],
+			   error_code);
+
+trap_halt:
+	/* We expect that if there is a handler it won't return,
+	 * but as we can't control that, we place halt outside
+	 * the above conditional.
+	 */
+	uk_lcpu_halt();
```
michpappas: `uk_arch_halt()` more suitable here

plat/native/Makefile.uk
```diff
@@ -0,0 +1,32 @@
...
+# TODO Remove as soon as plat dependencies go away
+LIBUKPLAT_NATIVE_CINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
+
+ifeq ($(CONFIG_ARCH_ARM_64),y)
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_PAGING) += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/paging.c|isr
+else
+ifeq ($(CONFIG_ARCH_X86_64),y)
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_PAGING) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/paging.c|isr
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_ECTX) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/ectx.c|isr
+LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_EXCEPT) += $(LIBUKPLAT_NATIVE_BASE)/arch/x86_64/exceptions.c|isr
```
michpappas: Nit: You can consider renaming to `except.c` to match the header and general convention.

lib/ukdebug/crash.c
```diff
@@ -134,20 +134,21 @@ __noreturn void _uk_crash(struct __regs *regs, struct uk_crash_descr *descr)
...
 	uk_crash_populate_descr(ctx, &descr);
 
-	_uk_crash(ctx->regs, &descr);
+	_uk_crash(uk_lcpu_except_err_ctx_get_regs(ctx),
+		  &descr);
 	UK_BUG(); /* noreturn */
 }
 
-UK_EVENT_HANDLER_PRIO(UK_EVENT_UNHANDLED_EXCEPTION, uk_crash_handler,
+UK_EVENT_HANDLER_PRIO(UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPTION, uk_crash_handler,
```
michpappas: `UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPT`

plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
mogasergiu: This has been moved to lcpu pm ops.

lib/ukpal/include/uk/pal/lcpu.h
mogasergiu: Moved to lcpu pm ops.

mogasergiu:
fpush:
- moved trapnr getter/setter to arch specific header

mogasergiu:
fpush... to solve build error :upside_down_face:

mogasergiu:
fpush to delete some more no longer necessary stuff...

lib/uklcpu/Config.uk
```diff
@@ -0,0 +1,34 @@
...
+
+config LIBUKLCPU_MAXCOUNT
+	int "Maximum number of supported logical CPUs"
+	range 1 256
+	default 1
+
+menuconfig LIBUKLCPU_SMP
+	bool "Multiprocessor Configuration"
+	depends on LIBUKLCPU
+	select HAVE_SMP
```
michpappas: `HAVE_SMP` and generally `HAVE` directives should be selected by the platform's configuration to indicate that the platform has a given capability.

mogasergiu:
fpsh:
- removed unnecessary duplicate inclusions of lcpu.h as a result of aggressive sed-ing

michpappas (changes requested):
@mogasergiu another round, thanks! 🚀

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
mogasergiu: We have `uk_arch_rflags_get` which does an unprivileged instruction. I believe this was meant as something generic that should've probably returned bool instead? Not sure if it's meaningful to change it to return bool in the context of this work. The restore function, as it can be seen, it checks if the restored flags is  0 or not as it is not safe to reuse a previous saved flags due to the other bitflags that may have been changed during execution.

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
mogasergiu: I can do an auxsp header sure. Though not sure about IRQ given it's basically "pausing"/"unpausing" some of the events defined in this very header.

plat/native/exportsyms.uk
mogasergiu: ~~Yes, this exportsyms will only contain ectx/sysctx syms~~

Edit: Actually... I guess it will contain all native syms since they are meant to be reusable.

lib/uklcpu/include/uk/lcpu/pm.h
mogasergiu: oh.. it's only removed in the intermediary commit... oops

lib/uklcpu/include/uk/lcpu/core.h
mogasergiu: it's in uk/arch.h :yum:

lib/ukpal/include/uk/pal/except.h
mogasergiu: Well... it's going to disappear anyway as it is just a hack for now due to missing proper pcpu vars. But yes, I can rename for the sake of this commit

plat/native/arch/x86_64/exceptions.c
mogasergiu: oh yeah definetely oversight, this would have ended up adding another dep on uklcpu

lib/uklcpu/Config.uk
mogasergiu: Actually... I am starting to think that we might have to somehow change this (in ukpaging too?) so that the PAL config actually selects it somehow. The functionality is, after all, registered only if the PAL is active. Otherwise it may at most be just available but not used by the lib that abstracts over it.

mogasergiu:
fpush:
- s/UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPTION/UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPT in ukdebug
- auxsp header in lcpu/pal/plat
- s/uk_pal_traps_get_except_stack_base/uk_pal_except_get_except_stack_base
- use uk_arch_halt() instead of uk_lcpu_halt()
- rm pal + mb syms from exportsyms

mogasergiu:
fpush rebase

mogasergiu:
fpush:
- remove no longer necessary header start16_helpers.h
- add uk/config.h to plat headers
- add select LIBUKLCPU that do use uklcpu
- move CPU MAXCOUNT/MULTI_PHASE_STARTUP back to plat/Config.uk

mogasergiu:
fpush remove unintentionally added files

lib/ukatomic/include/uk/atomic.h
```diff
@@ -167,17 +167,17 @@ extern "C" {
 #define	UK_ACCESS_ONCE(x)			(*(volatile __typeof(x) *)&(x))
 
 #define	UK_WRITE_ONCE(x, v) do {	\
-	barrier();			\
+	__barrier();			\
```
michpappas: Requires also including `uk/compiler.h`. Same for other updates for `__barrier`.

lib/ukpal/include/uk/pal/except.h
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Exception error context (opaque).
+ * Contains CPU state and exception metadata passed to error handlers.
+ * Access via accessor functions.
+ */
+struct uk_pal_except_err_ctx;
+
+/* Exception error context accessors */
+
+__isr const char * uk_pal_except_err_ctx_get_str(
```
michpappas: IIRC we moved accessors to `struct uk_pal_except_err` to arch-specific headers.

lib/uklcpu/Makefile.uk
```diff
@@ -0,0 +1,16 @@
...
+
+ASINCLUDES-$(CONFIG_LIBUKLCPU)  += -I$(LIBUKLCPU_BASE)/include
+CINCLUDES-$(CONFIG_LIBUKLCPU)   += -I$(LIBUKLCPU_BASE)/include
+CXXINCLUDES-$(CONFIG_LIBUKLCPU) += -I$(LIBUKLCPU_BASE)/include
+
+ASINCLUDES-$(CONFIG_LIBUKLCPU)  += -I$(LIBUKLCPU_BASE)/arch/$(CONFIG_UK_ARCH)/include
+CINCLUDES-$(CONFIG_LIBUKLCPU)   += -I$(LIBUKLCPU_BASE)/arch/$(CONFIG_UK_ARCH)/include
+CXXINCLUDES-$(CONFIG_LIBUKLCPU) += -I$(LIBUKLCPU_BASE)/arch/$(CONFIG_UK_ARCH)/include
+
+# FIXME: Remove this and make the library self contained
```
michpappas, line 11: Since we legitimately include `uk/plat/pal/` headers, this statement can be rephrased or removed.

lib/uklcpu/Config.uk
```diff
@@ -0,0 +1,24 @@
+menuconfig LIBUKLCPU
+	bool "uklcpu: generic logical CPU interface"
```
michpappas: `uklcpu: Generic ...` (capital G)  to be aligned with the rest of options in the menuconfig interface.

lib/uklcpu/Config.uk
```diff
@@ -0,0 +1,24 @@
...
+	select LIBUKPAL
+	default y
```
michpappas: I wonder setting a hard default is necessary, or whether we should select this library in the platform. In theory it should be possible to build a bare-bones platform that takes control right after the bootcode / ukboot.

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
michpappas: Okay I understand. I think returning a bool is a bit in conflict with the semantics implied by the function's name (save flag(s)) as you don't truly save the flag(s) but the state (interrupts were enabled or disabled). The `arm64` port interprets that quite literally: It saves DAIF as is. I will push a change on the `arm64` follow up PR to first of all only return / restore the I flag, so that we align semantically across architectures, but I won't treat is a binary flag (just return DAIF masked for the I flag). `x86_64` doesn't need to align to that, it can still treat it as boolean and that will work as long as the implementations of save / restore functions are in sync. So nothing to do here.

lib/uklcpu/include/uk/lcpu/core.h
michpappas: Okay I'll adapt the `arm64` PR instead.

lib/ukdebug/crash.c
michpappas: What are your thoughts on renaming this to `UK_LCPU_EXCEPT_EVENT_UNHANDLED`?

lib/uklcpu/Config.uk
michpappas: So the idea is that the platforms select their capabilities (`HAVE_`), as well as their dependencies (other platforms or libraries) when they are *active* (selected). For example:
```
menuconfig UKPLAT_NATIVE_PAL
   bool "Native Platform"
   # The platform here configures all it needs when it's selected
   HAVE_PAGING
   select UKPLAT_NATIVE
   select LIBUKPAGING
   ...
   
config UKPLAT_NATIVE
  bool
  # The platform here defines various options that can be selected
  # when it's treated as a platform library (not necessarily selected)
   ...
```
Other platforms inheriting from UKPLAT_NATIVE can similarly choose whether to select `HAVE_PAGING` and accordingly will or will not see what depends on it.

That being said, I'm not sure if there is a downside to having `HAVE_PAGING` selected by `LIBUKPLAT_NATIVE_PAGING` instead. Maybe two ways to achieve the same result? The only benefit I can think of at the moment is that under the scheme I described above we have a single place to define platform capabilities (`HAVE`) and dependencies (plat / lib), which is its `menuconfig LIBUK<plat>_PAL`.

lib/uklcpu/include/uk/lcpu/pm.h
```diff
@@ -0,0 +1,98 @@
...
+#include <uk/lcpu/core.h>
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_lcpu_pm_ops {
+#if CONFIG_HAVE_SMP
+	int (*start)(__u64 id);
```
michpappas: PSCI issues the `PSCI_FNID_CPU_ON` SMC with two more  parameters: Entry point, and args. Can we adapt the `start` funtion's declaration to cater for that? `x86_64` can ignore these args as it supports multi-phase startup.

Also, a minor thing: by the coding conventions it's supposedly preferable to `typedef` the functions first and then register them as members of the ops struct.
```c
typedef int (*uk_lcpu_halt_func)(void);
...
```
Personally I don't see why, but still follow the convention nevertheless 🤷🏼‍♂️

michpappas (changes requested):
@mogasergiu I did an end-to-end pass. Unless we discover something later, this should be the last round.

mogasergiu:
Pushed adding entry arg to ->start()

lib/uklcpu/Makefile.uk
mogasergiu: This is for `time_block_until`... :sweat_smile: 
I guess we can only remove after we properly migrate tscclock

lib/uklcpu/include/uk/lcpu/pm.h
mogasergiu: I am not sure I see the technical reason of this either. Perhaps easier refactoring? e.g. if the same signature is reused in a bunch of other structs you only need to change the typedef?
Idk, yeah, anyway, convention is convention :sweat_smile:

mogasergiu:
fpush:
- move error_code get/set to x86 dir
- s/generic/Generic in lib/uklcpu/Config.uk
- s/UK_LCPU_EXCEPT_EVENT_UNHANDLED_EXCEPT/UK_LCPU_EXCEPT_EVENT_UNHANDLED
- use typedefs for lcpu_pm ops

mogasergiu:
fpush:
- #1752 rebase
- pushed the header guards that allow reusing some macros without enabling the implementation (also moved `uk_plat_native_lcpu_id/idx` methods from `except.h` to `lcpu.h` (idk how they ended up there in the first place)
<details>
<summary>Large diff for relevant changes (click to expand)</summary>


```diff
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/auxsp.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/auxsp.h
index 2f37ffcb..856e972a 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/auxsp.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/auxsp.h
@@ -6,10 +6,6 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_AUXSP
-#error "Native auxsp is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_AUXSP_H__
 #define __UK_PLAT_NATIVE_ARCH_AUXSP_H__
 
@@ -20,6 +16,7 @@
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_AUXSP
 /**
  * Get current auxiliary stack pointer.
  * The auxiliary stack is used for exception handling.
@@ -43,6 +40,7 @@ void uk_plat_native_set_auxsp(__uptr auxsp);
  * @return Current auxiliary stack pointer
  */
 __isr __uptr uk_plat_native_get_auxsp_in_except(void);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_AUXSP */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h
index ae6b1866..fbf6c439 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/ectx.h
@@ -6,10 +6,6 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_ECTX
-#error "Native ectx is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_ECTX_H__
 #define __UK_PLAT_NATIVE_ARCH_ECTX_H__
 
@@ -43,6 +39,7 @@
 /* XSAVE/XRSTOR required alignment */
 #define UK_PLAT_NATIVE_ECTX_ALIGN			64
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 /* Assembly-visible function symbols */
 #define UK_PLAT_NATIVE_ECTX_LOAD_FNSYM					\
 	uk_plat_native_ectx_load
@@ -50,12 +47,14 @@
 	uk_plat_native_ectx_store
 #define UK_PLAT_NATIVE_ECTX_SANITIZE_FNSYM				\
 	uk_plat_native_ectx_sanitize
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 /**
  * x86_64 extended context (opaque XSAVE area).
  * Must be UK_PLAT_NATIVE_ECTX_SIZE bytes, 64-byte aligned.
@@ -110,6 +109,7 @@ __isr void uk_plat_native_ectx_init(struct uk_plat_native_ectx *state);
  * @param state  ectx to compare against current CPU state
  */
 __isr void uk_plat_native_ectx_assert_equal(struct uk_plat_native_ectx *state);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
index 4f8beb9b..99184e0c 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
@@ -6,10 +6,6 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_EXCEPT
-#error "Native except is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_EXCEPT_H__
 #define __UK_PLAT_NATIVE_ARCH_EXCEPT_H__
 
@@ -199,6 +195,7 @@ uk_plat_native_except_irq_ctx_set_irq(
 	ctx->irq = irq;
 }
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 /**
  * Enable x86_64 interrupts (STI instruction).
  * Sets RFLAGS.IF, allowing maskable hardware interrupts.
@@ -286,6 +283,7 @@ __isr void uk_plat_native_except_push_nested(void);
  * Must be called with interrupts disabled.
  */
 __isr void uk_plat_native_except_pop_nested(void);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
index cbbdae10..217904af 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/lcpu.h
@@ -6,24 +6,23 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_LCPU
-#error "Native lcpu is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_LCPU_H__
 #define __UK_PLAT_NATIVE_ARCH_LCPU_H__
 
 #include <uk/arch/types.h>
 #include <uk/arch/util.h>
 
+#if CONFIG_LIBUKPLAT_NATIVE_LCPU
 #define UK_PLAT_NATIVE_AP_ENTRY_SYM					\
 	uk_plat_native_x86_64_start16_addr
+#endif /* CONFIG_LIBUKPLAT_NATIVE_LCPU */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_LCPU
 extern __vaddr_t uk_plat_native_x86_64_start16_addr; /* target address */
 
 /**
@@ -83,6 +82,7 @@ int uk_plat_native_lcpu_mp_init(void *arg __unused);
 int uk_plat_native_send_ipi(__u64 id, unsigned long irq);
 
 #endif /* CONFIG_HAVE_SMP */
+#endif /* CONFIG_LIBUKPLAT_NATIVE_LCPU */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/regs.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/regs.h
index 949520f0..7aec5a05 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/regs.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/regs.h
@@ -6,10 +6,6 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_REGS
-#error "Native regs is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_REGS_H__
 #define __UK_PLAT_NATIVE_ARCH_REGS_H__
 
diff --git a/plat/native/arch/x86_64/include/uk/plat/native/arch/sysctx.h b/plat/native/arch/x86_64/include/uk/plat/native/arch/sysctx.h
index c284a03a..8de427e5 100644
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/sysctx.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/sysctx.h
@@ -6,10 +6,6 @@
 
 #include <uk/config.h>
 
-#if !CONFIG_LIBUKPLAT_NATIVE_SYSCTX
-#error "Native sysctx is not available in current configuration"
-#endif
-
 #ifndef __UK_PLAT_NATIVE_ARCH_SYSCTX_H__
 #define __UK_PLAT_NATIVE_ARCH_SYSCTX_H__
 
@@ -35,11 +31,13 @@
 #define UK_PLAT_NATIVE_SYSCTX_OFFSETOF_TLSP				\
 	UK_PLAT_NATIVE_X86_64_SYSCTX_OFFSETOF_FSBASE
 
+#if CONFIG_LIBUKPLAT_NATIVE_SYSCTX
 /* Assembly-visible function symbols */
 #define UK_PLAT_NATIVE_SYSCTX_LOAD_FNSYM				\
 	uk_plat_native_sysctx_load
 #define UK_PLAT_NATIVE_SYSCTX_STORE_FNSYM				\
 	uk_plat_native_sysctx_store
+#endif /* CONFIG_LIBUKPLAT_NATIVE_SYSCTX */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
@@ -110,6 +108,7 @@ uk_plat_native_sysctx_set(struct uk_plat_native_sysctx *sc, __sz offset,
 	*(__u64 *)((char *)sc + offset) = val;
 }
 
+#if CONFIG_LIBUKPLAT_NATIVE_SYSCTX
 /**
  * Function pointers for FS_BASE register access.
  * Dynamically selected at boot based on CPU capabilities:
@@ -160,6 +159,7 @@ __isr void uk_plat_native_sysctx_store(struct uk_plat_native_sysctx *sysctx);
  * @param sysctx  System context to restore
  */
 __isr void uk_plat_native_sysctx_load(struct uk_plat_native_sysctx *sysctx);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_SYSCTX */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/include/uk/plat/native/auxsp.h b/plat/native/include/uk/plat/native/auxsp.h
index c7624100..7900bae3 100644
--- a/plat/native/include/uk/plat/native/auxsp.h
+++ b/plat/native/include/uk/plat/native/auxsp.h
@@ -16,9 +16,11 @@
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_AUXSP
 __uptr uk_plat_native_get_auxsp(void);
 __isr __uptr uk_plat_native_get_auxsp_in_except(void);
 void uk_plat_native_set_auxsp(__uptr auxsp);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_AUXSP */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/include/uk/plat/native/ectx.h b/plat/native/include/uk/plat/native/ectx.h
index eaadb0b1..092b2cc8 100644
--- a/plat/native/include/uk/plat/native/ectx.h
+++ b/plat/native/include/uk/plat/native/ectx.h
@@ -32,6 +32,7 @@
 #error "UK_PLAT_NATIVE_ECTX_ALIGN undefined"
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 #ifndef UK_PLAT_NATIVE_ECTX_LOAD_FNSYM
 #error "UK_PLAT_NATIVE_ECTX_LOAD_FNSYM undefined"
 #endif
@@ -43,12 +44,14 @@
 #ifndef UK_PLAT_NATIVE_ECTX_SANITIZE_FNSYM
 #error "UK_PLAT_NATIVE_ECTX_SANITIZE_FNSYM undefined"
 #endif
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 /**
  * Sanitize extended context memory for first use.
  * Initializes architecture-specific fields that must have valid values
@@ -90,6 +93,7 @@ __isr void uk_plat_native_ectx_init(struct uk_plat_native_ectx *state);
  * @param state  Extended context to compare against current CPU state
  */
 __isr void uk_plat_native_ectx_assert_equal(struct uk_plat_native_ectx *state);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/include/uk/plat/native/except.h b/plat/native/include/uk/plat/native/except.h
index f00d8619..24aecc01 100644
--- a/plat/native/include/uk/plat/native/except.h
+++ b/plat/native/include/uk/plat/native/except.h
@@ -64,11 +64,6 @@ extern "C" {
 #error "UK_PLAT_NATIVE_EXCEPT_EVENT_UNHANDLED undefined"
 #endif
 
-/* LCPU identification (convenience wrappers) */
-
-__u64 uk_plat_native_lcpu_id(void);
-__u32 uk_plat_native_lcpu_idx(void);
-
 /* Exception error context accessors */
 
 __isr const char *uk_plat_native_except_err_ctx_get_str(
@@ -115,6 +110,7 @@ __isr void uk_plat_native_except_irq_ctx_set_irq(
 	struct uk_plat_native_except_irq_ctx *ctx,
 	__u64 irq);
 
+#if CONFIG_LIBUKPLAT_NATIVE_ECTX
 /* IRQ control */
 
 __isr void uk_plat_native_enable_irq(void);
@@ -138,6 +134,7 @@ __isr __uptr uk_plat_native_except_get_except_stack_base(void);
 
 __isr void uk_plat_native_except_push_nested(void);
 __isr void uk_plat_native_except_pop_nested(void);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_ECTX */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/include/uk/plat/native/lcpu.h b/plat/native/include/uk/plat/native/lcpu.h
index d3b59146..8895f36f 100644
--- a/plat/native/include/uk/plat/native/lcpu.h
+++ b/plat/native/include/uk/plat/native/lcpu.h
@@ -12,16 +12,24 @@
 #include <uk/lcpu/core.h>
 #include <uk/plat/native/arch/lcpu.h>
 
+#if CONFIG_LIBUKPLAT_NATIVE_LCPU
 /* Virtual address of Application Processor start */
 #ifndef UK_PLAT_NATIVE_AP_ENTRY_SYM
 #error "UK_PLAT_NATIVE_AP_ENTRY_SYM undefined"
 #endif
+#endif /* CONFIG_LIBUKPLAT_NATIVE_LCPU */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_LCPU
+/* LCPU identification (convenience wrappers) */
+
+__u64 uk_plat_native_lcpu_id(void);
+__u32 uk_plat_native_lcpu_idx(void);
+
 /**
  * Initialize logical CPU structures and per-CPU state.
  * Must be called once per CPU during boot.
@@ -42,6 +50,7 @@ int uk_plat_native_lcpu_init(struct uk_lcpu *this_lcpu);
  */
 int uk_plat_native_mp_init(void *arg __unused);
 #endif /* CONFIG_HAVE_SMP */
+#endif /* CONFIG_LIBUKPLAT_NATIVE_LCPU */
 
 #ifdef __cplusplus
 }
diff --git a/plat/native/include/uk/plat/native/sysctx.h b/plat/native/include/uk/plat/native/sysctx.h
index 0557c73f..ee81a030 100644
--- a/plat/native/include/uk/plat/native/sysctx.h
+++ b/plat/native/include/uk/plat/native/sysctx.h
@@ -32,6 +32,7 @@
 #error "UK_PLAT_NATIVE_SYSCTX_OFFSETOF_TLSP undefined"
 #endif
 
+#if CONFIG_LIBUKPLAT_NATIVE_SYSCTX
 #ifndef UK_PLAT_NATIVE_SYSCTX_LOAD_FNSYM
 #error "UK_PLAT_NATIVE_SYSCTX_LOAD_FNSYM undefined"
 #endif
@@ -39,6 +40,7 @@
 #ifndef UK_PLAT_NATIVE_SYSCTX_STORE_FNSYM
 #error "UK_PLAT_NATIVE_SYSCTX_STORE_FNSYM undefined"
 #endif
+#endif /* CONFIG_LIBUKPLAT_NATIVE_SYSCTX */
 
 #if !__ASSEMBLY__
 #ifdef __cplusplus
@@ -68,6 +70,7 @@ __u64 uk_plat_native_sysctx_get(const struct uk_plat_native_sysctx *sc,
 void uk_plat_native_sysctx_set(struct uk_plat_native_sysctx *sc,
 			       __sz offset, __u64 val);
 
+#if CONFIG_LIBUKPLAT_NATIVE_SYSCTX
 /**
  * Get current thread-local storage pointer.
  * Reads the architecture-specific register used for TLS.
@@ -101,6 +104,7 @@ __isr void uk_plat_native_sysctx_store(struct uk_plat_native_sysctx *sysctx);
  * @param sysctx  System context to restore
  */
 __isr void uk_plat_native_sysctx_load(struct uk_plat_native_sysctx *sysctx);
+#endif /* CONFIG_LIBUKPLAT_NATIVE_SYSCTX */
 
 #ifdef __cplusplus
 }
 ```
 </details>

mogasergiu:
Got rid of the AP_ENTRY macro and CPU ID for arguments in mp.h in favor of keeping struct uk_lcpu pointer for args **for now** (cyclical dep is temporary and already exists through core.h), like it was before the rearch. Ultimately, will be updated to use indexes instead of IDs or struct uk_lcpu when proper per cpu vars will be added, as that is the component that is universally understood as logical/linear and platform/arch agnostic, unlike the hw ID.

michpappas (approved):
Aside from the parts to be clarified by the upcoming `libukpcpuvars`, I think it overall looks good! Checkpatch errors to be revised merging dependencies and rebasing on top of `staging`.

andreittr (approved):
Looks good, looking forward to the pcpuvars changes.

mogasergiu:
fpush:
- Remove unnecessary setting of fs/gs segments during segment reload
- s/ectx/except guards in except.h

mogasergiu:
fpush:
- checkpatch warnings
- another s/ectx/except guards in except.h oops

mogasergiu:
fpush:
- remove no longer necessary `#include <uk/config.h> ` outside the guards. This was previously needed when we wanted to guard the inclusion of the headers themselves with that header's config, but we now guard inside the header itself only whatever really depends on some implementation that wouldn't otherwise be available if the config was disabled. Every translation unit that transitively includes the headers will re-process the #include <uk/config.h> line even after the guard has fired, so it is only detrimental now => move the inclusion inside the guards

mogasergiu:
rebase fpush

mogasergiu:
fpush:
- fix some leftover checkpatch errors...

mogasergiu:
fpush:
- oops another checkpatch warning missed

andreittr:

PR #1763, opened 2026-01-24
lib/posix-futex: Fix lost wakeup race in futex_wait

Aman-Cool:
# Fix lost wakeup race in futex_wait

## Problem

`futex_wait()` checks the futex value **before** acquiring the futex wait-list lock.  
A `futex_wake()` can occur between the value check and enqueue, causing a **lost wakeup**.

This can lead to **permanent deadlocks** in pthread mutexes, condition variables, and other futex-based primitives.

---

## Root Cause

The futex value check and wait-list enqueue are **not atomic** with respect to `futex_wake()`, violating required futex semantics.

---

## Fix

Move the futex value check **inside the futex wait-list critical section**, making the check-and-enqueue atomic.

---

## Impact

- Prevents permanent deadlocks
- Restores correct futex semantics
- Fixes silent hangs in multi-threaded workloads
- Affects all platforms and pthread-based applications

### DIFF
--- a/lib/posix-futex/futex.c
+++ b/lib/posix-futex/futex.c
@@ -102,19 +102,29 @@ static int futex_wait(uint32_t *uaddr, uint32_t val, const __nsec *timeout)
 	struct uk_thread *current = uk_thread_current();
 	struct uk_futex f = {.uaddr = uaddr, .thread = current};
 
+	/*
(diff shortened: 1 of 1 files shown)


Aman-Cool:
@mogasergiu Fixes a lost-wakeup race in `futex_wait()` where a wake could occur before enqueueing, causing permanent pthread deadlocks_

PR #1766, opened 2026-01-31
drivers/virtio/fs: Avoid VLA on stack for sg segments

andutu0:
### Description of Changes

Replace runtime-sized sg segment VLAs in virtiofs_do_request() and virtiofs_do_request_vec() with heap allocations.

This avoids large stack allocations, preventing potential crashes with large virtqueue lengths.

### Related Work

N/A

### PR Checklist

### Note:

The fix was only tested with mocks, because Unikraft’s virtio-pci transport only supports the legacy interface for now, while QEMU exposes virtio-fs using a modern virtio PCI device ID, and I could not find a way to make it work.

### DIFF
--- a/drivers/virtio/fs/virtio_fs.c
+++ b/drivers/virtio/fs/virtio_fs.c
@@ -271,6 +271,7 @@ __ssz virtiofs_do_request_vec(struct virtqueue *vq, __u16 vqlen,
 	struct virtiofs_thread_cookie cookie = { .thread = current };
 	struct uk_sglist sg;
 	struct uk_sglist sg_out;
+	struct uk_sglist_seg *sgsegs = NULL;
 	__u16 nseg;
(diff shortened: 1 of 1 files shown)


razvand:
@andutu0 , could you please clarify what issue can stack allocation cause? Could you provide a demo / PoC of that issue?

andutu0:
### Summary

During further investigation, I found out that in QEMU the queue-size is capped at 1024, even though virtio allows queue sizes up to 32768. As a result, the maximum size of the VLAs is limited to ```1024 * sizeof(struct uk_sglist_seg) = 1024 * 16 bytes = 16 KiB```, which is around 1/4 of a thread’s stack on my system. This means that, without a queue size of 4096, which is currently not available in QEMU, I cannot realistically get it to crash.

PR #1767, opened 2026-02-05
lib: Add system-wide power management library

mogasergiu:
### Description of Changes

Introduce ukpm: a library that replaces bootstrap.h.
ukpm attempts to enhance (and remove from plat) what bootstrap.h
definitions offered, while also adding an additional shutdown request
event and allowing an external driver to register its implementation
of system-wide power management operations.

Note that at this moment only one set of system-wide power management
operations are allowed. In reality, most platforms allow for multiple
shutdown mechanism, so we should, when need arises, enhance the library
by adding the possibility of maintaining an internal, priority-based,
list of all registered operations from all drivers that are able to
perform such actions.

That being said, following the introduction of this new library, all consumers
of bootstrap.h have been updated to use libukpm and previous implementations
of shutdown mechanisms have been moved to drivers.

Depends-on: #1752 
Depends-on: #1758 

### Related Work

N/A

### PR Checklist


mogasergiu:
- Undid ARM PSCI deletion
- efi -> ukefi
- undid the commit that "De-Unikraftified" PS/2 driver

drivers/firmware/ukefi/Makefile.uk
```diff
@@ -0,0 +1,5 @@
+$(eval $(call addlib_s,libfirmware_ukefi,$(CONFIG_LIBFIRMWARE_UKEFI)))
```
michpappas: No strong opinion, but since we have decoupled the 1-1 relation between driver and kernel libraries, we no longer need to prepend the library name with its parent. So we can also call this `LIBUKEFI`.

lib/ukpm/pm.c
```diff
@@ -0,0 +1,150 @@
...
+
+__noreturn void uk_pm_syshalt(void)
+{
+	int rc;
+
+	rc = uk_raise_event(UK_PM_EVENT_SYSHALT, __NULL);
+	if (unlikely(rc < 0))
+		uk_pr_err("Failed system halt event raising: %d\n", rc);
+
+	if (pm_ops && pm_ops->syshalt)
```
michpappas: Here and on the rest of the functions, a few comments would clear any doubts on the behavior while reading the code:
```c
	if (pm_ops && pm_ops->syshalt)
		pm_ops->syshalt(); /* noreturn */

    /* Fallback */	
    uk_lcpu_disable_irq();
	uk_lcpu_halt();
```
On the other hand I only had to have a quick look at the header to confirm that the op does not return so I dunno... 🤷🏼‍♂️

lib/ukpm/pm.c
```diff
@@ -0,0 +1,150 @@
...
+
+UK_EVENT(UK_PM_EVENT_SHUTDOWN_REQ);
+
+__isr void uk_pm_raise_shutdown_event(enum uk_pm_shutdown_op op)
+{
+	int rc __maybe_unused;
+
+	rc = uk_raise_event(UK_PM_EVENT_SHUTDOWN_REQ, (void *)op);
+
+	/* Make sure no handler returns UK_EVENT_HANDLED as that
```
michpappas: It would be useful to document `uk/pm.h` that handlers should not return `UK_EVENT_HANDLED`

lib/ukpm/include/uk/pm.h
```diff
@@ -0,0 +1,90 @@
...
+#include <uk/arch/types.h>
+#include <uk/essentials.h>
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_pm_ops {
+	void (*syshalt)(void) __noreturn;
```
michpappas: These functions should be `typedefs` according to the convention. More importantly: I see that the efi driver implements all four ops as `__isr`. Is that something we expect of all drivers? And if yes, should their definition also include the `__isr` attribute?

drivers/firmware/ukefi/pm.c
```diff
@@ -0,0 +1,69 @@
...
+		return -ENODEV;
+
+	rs = st->runtime_services;
+	if (unlikely(!rs))
+		return -ENODEV;
+
+	return uk_pm_ops_register(&efi_rs_pm_ops);
+}
+
+UK_BOOT_EARLYTAB_ENTRY(efi_rs_register_pm_ops, UK_PRIO_EARLIEST);
```
michpappas, line 114: Since the driver does - correctly - not explicitly select `libukboot` in Kconfig, we should probably make the ops registration to earlytab, `efi_rs_register_pm_ops()`,and associated includes, conditional to `LIBUKBOOT`. And since it doesn't make sense to provide a driver with no public symbols if `libukboot` is not selected, maybe also provide a public API that contains:
* All the currently `static efi_rs_` functions (non-static `uk_efi_rs_`).
* A `uk_efi_init()` function that performs the sanity checks now performed by `efi_rs_register_pm_ops()`.

That way we also allow the implementation to be used on a very bare-bones implementation that doesn't use `libukboot`.

mogasergiu:
Rebase fpush

drivers/input/ukps2/ps2.h
michpappas, line 1: This header is now orphaned (not owned by any library). Can you provide some context on the necessity of having a separate driver library for sysreset as opposed to implementing libukpm handlers within the kbd library?

plat/kvm/x86/qemu.c
michpappas, line 1: Initially I thought whether this code should live under `drivers/kvm/qemu` or so, similarly to `drivers/xen/`. On the other hand, it may be that `plat/kvm` should actually be `plat/qemu-virt/` and in that case this code should live under `drivers/qemu`. Let's leave it as is for this PR and defer the discussion to the time we clean up `plat/kvm`.

include/uk/init.h
```diff
@@ -63,7 +62,6 @@ struct uk_init_ctx {
 };
 
 struct uk_term_ctx {
-	enum ukplat_gstate target;
```
michpappas, line 66: Why are we no longer preserving `target`?

lib/ukboot/boot.c
```diff
@@ -239,9 +239,9 @@ extern char **environ;
 
 void uk_boot_entry(void)
 {
+	enum uk_pm_shutdown_op shutdown_op = UK_PM_SHUTDOWN_OP_SYSCRASH;
 	struct uk_init_ctx ictx = { 0 };
-	/* NOTE: Default target is crash for failed initialization (inittab) */
-	struct uk_term_ctx tctx = { .target = UKPLAT_CRASH };
+	struct uk_term_ctx tctx = { 0 };
```
michpappas: Why are we changing the initialization to zero?

plat/common/include/arm/arm64/cpu.h
```diff
@@ -55,10 +54,6 @@ static inline void _init_cpufeatures(void)
  */
 extern smccc_conduit_fn_t smccc_psci_call;
 
-/* CPU native APIs */
```
michpappas, line 58: Let's have the PSCI driver remove these.

michpappas (changes requested):

plat/xen/shutdown.c
```diff
@@ -1,87 +0,0 @@
-/* SPDX-License-Identifier: BSD-3-Clause */
```
andreittr, line 1: Considering that this source file contains the Xen implementation of system shutdown, and that this PR does not move it anywhere, I think we should leave its removal to the PR where Xen implements the new ukpm API.

andreittr (commented):
Changes look good, modulo comments already raised; one minor issue.

lib/ukpm/include/uk/pm.h
mogasergiu: Good question... I need to have another pass to see that all drivers are consistent with this. But yeah I think ideally this stuff should be __isr, especially when you consider pre-suspend operations.

drivers/input/ukps2/ps2.h
mogasergiu: I wanted to have it as `drivers/ukps2/{kbd.c, reset.c, ps2.h}`... but then we have this thing where subdirs are mandatory.
The PS/2 controller functionality is logically split in more subcomponents that you mux between using the command register (keyboard, CPU reset line, mouse).

So that being said, I thought it would not be suitable to have the reset line in the same driver as the keyboard because it is not necessarily related.
I also wanted to have the ps2.h commonly accessible and I did not know where else to place it.

After further offline discussions, we decided it would be a better idea to have this as part of `drivers/input/ukps`, i.e. create an input devices subsystem where we can place this. This would allow the aforementioned directory layout as well.

include/uk/init.h
mogasergiu: I had 2 choices, because `uk_term_ctx` (and so the entire `init.h`) would otherwise have a ukpm dependency:
- either keep `target` but guard with `ukpm` configs - I don't like this because i want when i look at `init.h` for it to not reference any other library, I don't want it to be a crossroads-like library (like ukboot) + ukpm already raises individual events for each type of termination so if a driver wanted to do something depending on the pm event, why would it register into inittab instead of as a handler for said ukpm event (no such driver currently)
- or remove target entirely and let ukpm events handle it (which is what I chose)

Following offline discussions, picking the second choice would still not solve the mixed semantics of teardown (they were a bit unclear before as well: is it inittab or ukplat_terminate that owns this?), so the decision would be to actually go with the first choice in the end and have `target` guarded with `ukpm` configs together with a `TODO` for later to decide who takes ownership of teardown events.

lib/ukboot/boot.c
mogasergiu: This (initializing `error_code` with `0`)  is actually a bug fix because if `uksched` would not be enabled then `tctx` would remain uninitialized on the `exit:` path and run the handlers with uninitialized `error_code`

mogasergiu:
Rebase fpush (sorry for multiple fpushes, messed it up :sweat_smile: )

mogasergiu:
second fpush undoes the removal of arm64 cpu.h and xen's shutdown.c (sorry, i fpushed the first time before doing `git rebase --continute` :sweat_smile: )

first fpush:
- moves ukps2 to drivers/input
- s/LIBFIRMWARE_UKEFI/UKEFI
- in both ps/2 and uefi pm drivers all the syms are now exported and only registered into ukpm/ukboot_early_tab if they are enabled
- brings back `target` into `uk/init.h`'s `uk_term_ctx`, conditioned by `CONFIG_LIBUKPM`
- adds `typedefs` for `ukpm`'s ops (note I can't use `__isr` because it is a code generation directive, not meaningful to the prototype - otherwise compiler throws: ` 'target' attribute ignored [-Wattributes]`)
- documents that everything ukpm must be ISR and that event handlers should not return `UK_EVENT_HANDLED` + minor `noreturn`/`Fallback` comments

drivers/input/ukps2/kbd.c
```diff
@@ -45,7 +44,7 @@ static int kbd_ps2_irq_handler(void *arg __unused)
 		if (!(lctrl_pressed && lalt_pressed))
 			break;
 
-		uk_raise_event(UKPLAT_SHUTDOWN_EVENT, (void *)UKPLAT_HALT);
+		uk_pm_raise_shutdown_event(UK_PM_SHUTDOWN_OP_SYSHALT);
```
michpappas, line 47: Since we are calling into the API of `libukpm` we should reflect that dependency to this library's `Config.uk`.

drivers/input/ukps2/exportsyms.uk
```diff
@@ -0,0 +1 @@
+uk_ps2_cpu_reset
```
michpappas, line 1: If we are exporting this symbol it should probably also be declared in a public header?

michpappas (changes requested):
Just a few small things left...

mogasergiu:
fpush:
- all ukpm drivers have `depends on LIBUKPM` instead of `select LIBUKPM`
- add forgotten `select HAVE_REBOOT` to UEFI ukpm driver
- remove LIBUKPM select from ukdebug and instead add fallback to lcpu

mogasergiu:
fpush:
- remove no longer necessary `HAVE_REBOOT` selection from the efi stub config, as now it is selected by the uefi ukpm driver

mogasergiu:
fpush:
- exports `uk_ps2_cpu_reset`
- add fallbacks to ukpm and make it always available/importable
- no longer make uk/init.h guard for LIBUKPM: it is now always usable with fallbacks, like ukprint

drivers/input/ukps2/Makefile.uk
```diff
@@ -3,4 +3,9 @@ $(eval $(call addlib_s,libukps2,$(CONFIG_LIBUKPS2)))
 # FIXME: Remove this and make the library self contained
 LIBUKPS2_CINCLUDES-$(CONFIG_LIBUKPS2_KBD) += -I$(UK_PLAT_COMMON_BASE)/include
 
+ASINCLUDES-y  += -I$(LIBUKPS2_BASE)/include
```
michpappas: Are headers unconditionally included by intention?

lib/nolibc/include/assert.h
```diff
@@ -61,7 +73,7 @@ extern "C" {
 				"Assertion failed: %s, file %s, line %u\n", \
 				STRINGIFY(x),				\
 				__FILE__, __LINE__);			\
-			ukplat_terminate(UKPLAT_CRASH);			\
+			nolibc_syscrash();				\
```
michpappas: Now that we have fallbacks in libukpm is the explicit spinwait needed here?

lib/ukdebug/crash.c
```diff
@@ -73,10 +73,20 @@ __noreturn static void crash_shutdown(void)
 		while (until > ukplat_monotonic_clock())
 			; /* do nothing */
 	}
-	ukplat_restart();
+
+#if CONFIG_LIBUKPM
```
michpappas: Now that we have fallbacks in libukpm are the conditionals and explicit fallback in lines 86-88 needed?

michpappas (changes requested):
Last round, caught a few small things.

mogasergiu:
fpush
- added `CONFIG_LIBUKPS` guards to ukps2 header inclusion
- nolibc unconditionally calls ukpm since it is now forever available through fallbacks
- removed unnecessary fallbacks forgotten in ukcrash
- IMPORTANT SEMANTIC CHANGE: now all ukpm drivers are expected to be able to return an error if their method failed to halt/crash/whatever (most suitable I could think of is `EIO`), ukpm logs it and then falls back on busy waiting... this is because in cases where the driver is selected but the related device is not available, then the driver method would silently fail and would continue execution to wherever (because no `ret` instruction)



I added the latter after testing at some point while forgetting to add the `isa-debug-exit` in the QEMU command-line and resulting in infinite exceptions

mogasergiu:
fpush:
- forgot to make QEMU `isa-debug-exit` driver `__isr`

plat/kvm/Config.uk
```diff
@@ -2,6 +2,7 @@ menuconfig PLAT_KVM
 	bool "KVM guest"
 	default n
 	depends on (ARCH_X86_64 || ARCH_ARM_64)
+	select LIBUKPM
```
andreittr: Since `<uk/pm.h>` is always available, maybe platforms should just `imply` it, or leave it out of their config altogether, and let libs select/imply libukpm if needed.

plat/xen/Config.uk
```diff
@@ -10,6 +10,7 @@ menuconfig PLAT_XEN
        select LIBUKBITOPS
        select LIBNOLIBC if !HAVE_LIBC
        select HAVE_FDT if ARCH_ARM_64
+       select LIBUKPM
```
andreittr: Similar argument to KVM, perhaps platforms shouldn't select libukpm.

drivers/firmware/ukefi/pm.c
```diff
@@ -0,0 +1,115 @@
...
+ */
+__isr int uk_efi_rs_syshalt(void)
+{
+	UK_ASSERT(rs);
+
+	rs->reset_system(UK_EFI_RESET_SHUTDOWN, UK_EFI_SUCCESS,
+			 sizeof("UK EFI SYSTEM SHUTDOWN"),
+			 "UK EFI SYSTEM SHUTDOWN");
+
+	return -EIO;
```
andreittr, line 35: I believe PM driver code would read clearer if we'd have a comment here explaining that `return -EIO;` is a fallback if `reset_system` fails for whatever reason. Without enough context this function reads as "do something then fail", which is inaccurate.
This applies to other PM ops in all drivers.

andreittr (commented):
Looks great, I've only a few minor comments regarding recent changes.

mogasergiu:
fpush:
- use imply instead of select, since the platform can live without LIBUKPM active
- add comments before driver error return
- fixup plat/common include in ps2 (should not depend on the kbd only)

mogasergiu:
The comments were a bit fat so make them slimmer

andreittr (approved):
Looks great, thanks!

michpappas (approved):
Nice thanks for the updates! 🚀✨

mogasergiu:
fpush:
- according to coding conventions, use `_func` not `_fn`

mogasergiu:
rebase fpush

mogasergiu:
fpush:
- fix checkpatch warning by using ENOTSUP instead of ENOSYS

michpappas:

PR #1768, opened 2026-02-05
plat/xen: Implement PAL API

andreittr:
### Description of Changes

This changeset implements the PAL interface for the Xen x86 platform, bringing it in line with recents platform rearchitecting developments.

What this PR does NOT (yet) do:
- ~~Add any new features to xen, even if demanded by (part of) the PAL API -- these are left as stubs for future work~~
  - Reworked traps, irq control and `%gs` register use on x86
- Clean up/rearchitect the Xen platform itself
- Cosmetic changes to bring Xen platform code up to current coding standards/conventions
- ~~Remove unused code~~
- Be `checkpatch.uk`-clean -- this will be done as a final step after review rounds nail down the content
- Any significant work on ARM64

### Related Work

This PR is tightly interlinked with the rest of the platform rearch work, consisting of:
- https://github.com/unikraft/unikraft/pull/1752
- https://github.com/unikraft/unikraft/pull/1758
- https://github.com/unikraft/unikraft/pull/1767
- https://github.com/unikraft/unikraft/pull/1774

### PR Checklist


plat/xen/include/xen-x86/irq.h
```diff
@@ -61,7 +61,7 @@
 		_vcpu->evtchn_upcall_mask = 0;                                 \
 		__barrier(); /* unmask then check (avoid races) */             \
 		if (unlikely(_vcpu->evtchn_upcall_pending))                    \
-			ukplat_lcpu_irqs_handle_pending();		       \
+			uk_pal_irqs_handle_pending();		       \
```
mogasergiu: So, the idea is that inside the plat driver we call plat driver stuff, instead of pal stuff. What is more, I would love it if we coul get rid of this header entirely and could redefine these as static inlines e.g. `uk_plat_xen_irq_enable`, like we did on native. I believe this is fine, since this header, as the comment implies, is taken from MiniOS's `os.h`, it's not a public Xen guest OS interface header. So in this case as well, the `uk_plat_xen_irqs_handle_pending` would be called.

plat/xen/events.c
```diff
@@ -352,7 +353,7 @@ struct uk_alloc;
 
 int ukplat_irq_init(struct uk_alloc *a __unused)
```
mogasergiu: This function looks... like it's not called anywhere, should we just remove it?

plat/xen/include/xen-x86/traps.h
```diff
@@ -35,21 +35,63 @@
 #ifndef _TRAPS_H_
 #define _TRAPS_H_
 
+#ifndef __ASSEMBLY__
```
mogasergiu: We try to use `#if CONFIG_*` where possible. I will only mention it here, but I see there are other places that could use this update.

plat/xen/include/xen-x86/traps.h
```diff
@@ -35,21 +35,63 @@
...
+
+#include <uk/arch.h>
 
 #include <xen/xen.h>
 
-#define TRAP_coproc_seg_overrun  9
-#define TRAP_spurious_int        15
-#define TRAP_xen_callback        32
+/* Xen-specific traps */
+#define UK_ARCH_TRAPNUM_COPROC_SEG_OVERRUN 9
```
mogasergiu: I believe you could create a new commit that adds coproc_seg_ovverun + spurious int to uk/arch.h? (or should I add it to arch.h in my PR instead 🤔 )
And the XEN_CALLBACK I believe you can namespace as `UK_PLAT_XEN_TRAPNUM_XEN_CALLBACK` (or maybe `UK_PLAT_XEN_TRAPNUM_HV_CALLBACK` to avoid the repeated "XEN" and also be aligned with its corresponding asm entry?).

plat/xen/x86/traps.c
```diff
@@ -48,48 +47,180 @@ DECLARE_TRAP   (spurious_int,       "spurious interrupt bug",      NULL)
...
+	struct uk_plat_xen_except_err_ctx ctx;
+	int rc;
+
+	ctx = (struct uk_plat_xen_except_err_ctx){
+		.regs = (struct uk_plat_native_regs *)regs,
+		.trapnr = trapnr,
+		.str = x86_exception_table[trapnr],
+		.error_code = error_code,
+		.handler_err = 0,
+		.cr2 = uk_arch_rdcr2(),
```
mogasergiu: Oh, I am not sure this will work as it is a privileged operation. On Xen PV guests, you need to read this from the hypervisor-shared information. So you could do a static inline like `uk_plat_xenpv_rdcr2` that does a `HYPERVISOR_shared_info->vcpu_info[smp_processor_id()].arch.cr2`

plat/xen/x86/arch_events.c
```diff
@@ -47,7 +48,7 @@ void arch_init_events(void)
 #if defined(__x86_64__)
 	asm volatile("movl %0,%%fs ; movl %0,%%gs" :: "r" (0));
 	/* 0xc0000101 is MSR_GS_BASE */
-	wrmsrl(0xc0000101, (uint64_t) &cpu0_pda);
+	uk_arch_wrmsrl(0xc0000101, (uint64_t) &cpu0_pda);
```
mogasergiu: What the... wrmsr for seg registers...?? This is from copy-paste, my hunch this is for HVM(Lite) not PV... yeah.. this was here before but does it ever even get run? Maybe you can add some prints or tests... idk

plat/xen/x86/traps.c
```diff
@@ -48,48 +47,180 @@ DECLARE_TRAP   (spurious_int,       "spurious interrupt bug",      NULL)
...
+	/* We expect that if there is a handler it won't return,
+	 * but as we can't control that, we place halt outside
+	 * the above conditional.
+	 */
+	uk_lcpu_halt();
+}
+
+void uk_plat_xen_enable_irq(void)
+{
+	local_irq_enable();
```
mogasergiu: Yep, I believe you could move those macros' implementations directly here. I am not entirely sure whether it makes sense to have these as static inlines like we did on native or implemented in the C file... up to you to consider.

plat/xen/sysctx.c
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/essentials.h>
+#include <uk/plat/native/sysctx.h>
+
+struct uk_plat_native_sysctx;
+
+void uk_plat_xen_sysctx_store(struct uk_plat_native_sysctx *sysctx __unused)
```
mogasergiu: I think these can also go as static inlines, given that we now have the FNSYM macros

plat/xen/arch/x86_64/include/uk/plat/xen/arch/halt.h
```diff
@@ -0,0 +1,17 @@
...
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_XEN_ARCH_HALT_H__
+#define __UK_PLAT_XEN_ARCH_HALT_H__
+
+#if !__ASSEMBLY__
+
+/* On x86 halt requires a hypercall; implemented in cpu_pv.c */
+void halt(void);
```
mogasergiu: Why do we need this halt()?

plat/xen/arch/arm64/include/uk/plat/xen/arch/halt.h
```diff
@@ -0,0 +1,22 @@
...
+#define __UK_PLAT_XEN_ARCH_HALT_H__
+
+#if !__ASSEMBLY__
+
+/* Xen uses the native halt on ARM64 */
+#include <uk/plat/native/except.h>
+
+static inline void halt(void)
+{
+	uk_plat_native_halt();
```
mogasergiu: ~~This will also not work as `hlt` is privileged. Just like `cr2` I suggest a `HYPERVISOR_sched_op(SCHEDOP_block, 0)`. Note you can find these in `plat/xen/x86/cpu_pv.c`, which I guess you could get rid of now.~~

Edit: Oops, this is ARM64.... yeah this will need some care in the next version.

mogasergiu (commented):
Did a first round. Really good in general. I believe now there are some headers/C files we could even get rid of (that are not part of the public Xen Guest OS interface headers of course).

plat/xen/include/xen-x86/traps.h
andreittr: According to the Intel SDM (Vol 1, Table 6-1 Exceptions and Interrupts) these exception codes are either reserved or free, with Xen piggybacking on top of them; specifically:
- 9, while indeed named, is a legacy code not used after the 386, being de-facto reserved since (the SDM calls it such)
- 15 is marked explicitly as reserved; the name "spurious interrupt" is a Xen construction
- 32 is the first maskable interrupt; the name "xen callback" is (obviously) a Xen construction

Thus, I'd say that _at most_ we could include 9 in the arch header, though considering that we are _x86_64_ (very much past the 386 days), it makes far more sense to leave it reserved for the general arch, and leave it to Xen to give them meaning.

drivers/xen/xencons/console.c
```diff
@@ -90,14 +90,9 @@
 #if (defined __X86_32__) || (defined __X86_64__)
 #include <xen-x86/setup.h>
 #include <xen-x86/mm.h>
-#if defined __X86_32__
-#include <xen-x86/hypercall32.h>
-#elif defined __X86_64__
-#include <xen-x86/hypercall64.h>
-#endif
 #elif (defined __ARM_32__) || (defined __ARM_64__)
+#include <xen-arm/setup.h>
```
oleksiimoisieiev, line 94: This change is against the commit description. You have removed hypercall{64|32} include but added setup.h. In the other case you have added hypercall.h in gnttab.c file. I think this deserves a comment. At least in commit message.

plat/xen/arch/x86_64/include/uk/plat/xen/arch/except.h
```diff
@@ -0,0 +1,265 @@
...
+/* IRQ control */
+
+__isr void uk_plat_xen_irqs_handle_pending(void);
+
+__isr static inline
+void uk_plat_xen_enable_irq(void)
+{
+	vcpu_info_t *vcpu;
+
+	__barrier();
```
oleksiimoisieiev: I'm not sure this barrier is needed. At least if deserves a comment why do you have barrier here and don't have barrier in this function: uk_plat_xen_save_irqf() before getting vcpu_info

andreittr:
Pushed update reworking irq control, the internal xen platform API, and addressing some comments.

andreittr:
Pushed rebase to latest #1758 ; no other changes.

oleksiimoisieiev (changes requested):

plat/xen/arch/x86_64/include/uk/plat/xen/arch/except.h
andreittr: Thanks for taking your time to look over the changes!

I believe the barrier here is intended to prevent reordering w.r.t. clearing the event mask (`vcpu->evtchn_upcall_mask = 0;`) rather than getting vcpu_info. A similar barrier exists in `uk_plat_xen_restore_irqf`, where flags could be 0.

But you bring up a very good point, the use of barriers should be explicitly motivated. I'll add clear comments for each one in the file :+1: .

andreittr:
Pushed a series of changes:
- Adapt to latest PAL API, as mandated by the rebase of latest #1752 & #1758
- Address review comments
- Removed some unused code

andreittr:
Pushed update rebasing on the latest:
- #1752 
- #1758 

Added dependency and rebase on #1767 . No code changes.

andreittr:
Pushed update rebasing on the latest:
- #1758 
- #1767 

Added dependency on #1774 . No code changes.

andreittr:
Pushed major update to the code, reflecting developments in dependency PRs, as well as:
- Reworking Xen's use of GS to reflect ukpcpuvar's use
- Implement ukpm operations for system halt/reboot/crash

andreittr:
Pushed rebase on latest staging, removing all dependent PR commits; taking PR out of draft.

andreittr:
Pushed update with a missed removal of a retired header.

plat/xen/include/xen-x86/traps.h
```diff
@@ -35,21 +35,66 @@
 #ifndef _TRAPS_H_
 #define _TRAPS_H_
 
+#ifndef __ASSEMBLY__
 #include <stdint.h>
-#include <x86/traps.h>
+#endif /* !__ASSEMBLY__ */
+
+#include <uk/arch.h>
```
mogasergiu, line 42: Typo in `plat/xen: Rework trap handling` commit message 😅 s/pecularities/peculiarities

plat/xen/x86/traps.c
```diff
@@ -24,22 +24,28 @@
...
 
-#define TRAP_TABLE_ENTRY(trapname, pl) \
-	{ TRAP_##trapname, pl, __KERNEL_CS, (unsigned long) ASM_TRAP_SYM(trapname) }
+/* Construct a trap table entry for trapname */
+#define TRAP_TABLE_ENTRY(trapname, ring)			\
+	{ UK_ARCH_TRAPNUM_##trapname, (ring), __KERNEL_CS,	\
+	  (unsigned long)ASM_TRAP_SYM(trapname) }
 
 /*
  * Submit a virtual IDT to teh hypervisor. This consists of tuples
```
mogasergiu: Could we also fix this tiny typo as well while at it? s/teh/the

plat/xen/arch/x86_64/include/uk/plat/xen/arch/except.h
```diff
@@ -0,0 +1,279 @@
...
+#define UK_PLAT_XEN_EXCEPT_EVENT_ERR_MATH				\
+	xen_except_event_err_math
+#define UK_PLAT_XEN_EXCEPT_EVENT_ERR_SECURITY				\
+	xen_except_event_err_security
+#define UK_PLAT_XEN_EXCEPT_EVENT_SYSCALL				\
+	xen_except_event_syscall
+#define UK_PLAT_XEN_EXCEPT_EVENT_IRQ					\
+	xen_except_event_irq
+#define	UK_PLAT_XEN_EXCEPT_EVENT_UNHANDLED				\
+	xen_except_unhandled
```
mogasergiu: xen_except_event_unhandled?

plat/xen/x86/traps.c
```diff
@@ -48,48 +54,185 @@ DECLARE_TRAP   (spurious_int,       "spurious interrupt bug",      NULL)
...
+static unsigned long read_cr2(void)
 {
-	UK_WARN_STUBBED();
-	return 0;
+	return HYPERVISOR_shared_info->vcpu_info[smp_processor_id()].arch.cr2;
+}
+
+void uk_plat_xen_except_err_handler(unsigned int trapnr,
+				    struct uk_lcpu_regs *regs,
+				    unsigned long error_code)
```
mogasergiu, line 180: Oh man I just noticed (and this is also in the native plat 🤦‍♂️ ), this is unsigned long and we assign it to an `int error_code` struct member

plat/xen/include/uk/plat/xen/except.h
```diff
@@ -0,0 +1,158 @@
...
+extern "C" {
+#endif
+
+/*
+ * Verify that architecture-specific header defined required event identifiers.
+ * These are used with the uk_event system to register exception/interrupt
+ * handlers.
+ */
+
+#ifndef UK_PLAT_XEN_EXCEPT_EVENT_DEBUG
```
mogasergiu, line 31: I think you are missing the NMI check

plat/xen/x86/entry64.S
```diff
@@ -399,27 +403,28 @@ ENTRY(asm_failsafe_callback)
...
+TRAP_ENTRY(GP_FAULT,            1)
+TRAP_ENTRY(PAGE_FAULT,          1)
+TRAP_ENTRY(SPURIOUS_INT,        1)
+TRAP_ENTRY(COPROC_ERROR,        0)
+TRAP_ENTRY(ALIGNMENT_CHECK,     1)
 /* no Machine Check */
-TRAP_ENTRY simd_error,          0
+TRAP_ENTRY(SIMD_ERROR,          0)
 /* no Virtualization Exception */
+TRAP_ENTRY(SECURITY_ERROR,      0)
```
mogasergiu, line 426: Wait I think the comment above is misleading, I think this was supposed to refer to #CP (Control Exception Error) not #VE (Virtualization Exception). The former does have an error code, while the latter does not.

plat/xen/include/uk/plat/xen/addr.h
```diff
@@ -0,0 +1,27 @@
...
+#include <uk/plat/native/addr.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define UK_PLAT_XEN_VADDR_INV	UK_PLAT_NATIVE_VADDR_INV
+#define UK_PLAT_XEN_PADDR_INV	UK_PLAT_NATIVE_PADDR_INV
+
+#endif /* !__ASSEMBLY__ */
```
mogasergiu: Stray `#endif /* !__ASSEMBLY__ */` with no corresponding `#if !__ASSEMBLY__`.

plat/xen/include/uk/plat/xen/page.h
```diff
@@ -0,0 +1,87 @@
...
+	UK_PLAT_NATIVE_PAGE_HUGE_ALIGN_UP
+
+#define UK_PLAT_XEN_PAGE_HUGE_ALIGN_DOWN	\
+	UK_PLAT_NATIVE_PAGE_HUGE_ALIGN_DOWN
+
+#define UK_PLAT_XEN_PAGE_HUGE_ALIGNED		\
+	UK_PLAT_NATIVE_PAGE_HUGE_ALIGNED
+
+/* Page Protections */
+#define UK_PLAT_XEN_PAGE_ATTR_PROT_NONE		\
```
mogasergiu, line 69: hmm, are these mappings correct? (NONE - READ and WRITE - EXEC)

plat/xen/pal/include/uk/plat/pal/addr.h
```diff
@@ -0,0 +1,30 @@
...
+/* Copyright (c) 2026, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_PAL_ADDR_H__
+#define __UK_PLAT_PAL_ADDR_H__
+
+#include <uk/arch/types.h>
+#include <uk/plat/native/addr.h>
```
mogasergiu: grep-ing for native got me this 😅 I think you meant xen

plat/xen/pal/arch/x86_64/include/uk/plat/pal/arch/except.h
```diff
@@ -0,0 +1,58 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
```
mogasergiu: I think all headers in plat/xen need 2026?

plat/xen/include/xen-x86/mm.h
```diff
@@ -47,13 +47,13 @@
...
 
-#define PAGE_SIZE UK_PAGING_PAGE_SIZE
-#define PAGE_SHIFT UK_PAGING_PAGE_SHIFT
-#define PAGE_MASK UK_PAGING_PAGE_MASK
-#define PAGE_COUNT UK_PAGING_PAGE_COUNT
-#define PAGE_ALIGN_DOWN UK_PAGING_PAGE_ALIGN_DOWN
+#define PAGE_SIZE	UK_PAGING_PAGE_SIZE
+#define PAGE_SHIFT	UK_PAGING_PAGE_SHIFT
+#define PAGE_MASK	UK_PAGING_PAGE_MASK
+#define PAGE_COUNT	UK_PAGING_PAGE_COUNT
```
mogasergiu: Confused a bit about the commit message and the above included header. You are saying you are using xen platform-native definitions, but I see it uses ukpaging ones

plat/xen/shutdown.c
```diff
@@ -48,38 +50,43 @@
...
 
-	ret = EBUSY;
-	/* ret = HYPERVISOR_suspend(virt_to_mfn(start_info_ptr)); */
-	return -ret;
+static int xen_crash(void)
+{
+	return xen_shutdown(SHUTDOWN_crash);
 }
+
+static const struct uk_pm_ops xen_pm_ops = {
```
mogasergiu, line 81: note, all ops need __isr

plat/xen/arch/x86_64/include/uk/plat/xen/arch/sysctx.h
```diff
@@ -0,0 +1,36 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* On x86 reading the segment registers is done as on native */
+
+__isr static inline
+void uk_plat_xen_sysctx_store(struct uk_plat_native_sysctx *sysctx)
+{
+	uk_plat_native_sysctx_store(sysctx);
```
mogasergiu, line 22: Note that one native these call some function pointers (by default MSR-based read/write) that are initialized at runtime based on CPUID. You may have to find a way to initialize these like on native. Perhaps you could export + namespace the native initialization in the x86_64 native header

andreittr: I believe the initialization you mention is already handled by a CTOR [defined by plat/native](https://github.com/unikraft/unikraft/blob/staging/plat/native/arch/x86_64/sysctx_auxsp.c#L112).

plat/xen/include/xen-x86/mm.h
andreittr: Good catch, must've gotten lost in git blackmagic :sweat_smile:

plat/xen/include/uk/plat/xen/page.h
andreittr: Aw crap, mispasting, thanks for the catch! I also seem to have missed defining read & exec altogether :shrug:

plat/xen/include/uk/plat/xen/except.h
andreittr: Is NMI not x86_64-specific? (and thus not subject to checks in the common header)

plat/xen/x86/entry64.S
andreittr: #CP is `SECURITY_ERROR` in the table; the comment about #VE stands in for exception number 20, the one following `SIMD_ERROR`, (19, #XM).

plat/xen/include/uk/plat/xen/except.h
mogasergiu: oh oops you're right, forgot it's the generic header :laughing: :man_facepalming:

plat/xen/x86/entry64.S
mogasergiu: Oh, I somehow got confused while reading the comments itself. It explicitly says "**no** VE" I somehow skipped the "no" and thought it was a comment for the TRAP_ENTRY below not a placeholder for #VE which does not apply to us.

plat/xen/arch/x86_64/include/uk/plat/xen/arch/sysctx.h
mogasergiu: Oh yeah... well... i guess this does work since Xen PV does CPUID faulting/trapping, but otherwise we should use `XEN_EMULATE_PREFIX` in general. I guess it is fine for this plain ISA feature, but this raises a problem in scenarios where you want the implementations from the C file of native, but you probably don't want the auto-registration of certain handlers. While the platform may use the same implementation in some places e.g. store/load, the initialization may look different (e.g. hypercall or synthetic instruction). Going to resolve the comment for now, but I feel like we might bump into this again. We will likely have to add a separate CONFIG for the init handler registration when that happens.

andreittr:

andreittr:
Pushed update dropping commit dd0e018 (plat/xen: Use xen platform API for x86 mm.h), as the macros used by mm.h cannot fully use xen page definitions and must depend on ukpaging for the time being.

andreittr:
Pushed update rebasing on latest staging; no further changes.

andreittr:
Pushed update with required changes after the merging of #1779 .

plat/xen/x86/traps.c
mogasergiu: Following an offline discussion with @andreittr , we concluded on opening a PR after this one is merged which fixes this up in both native/xen code

plat/xen/x86/setup.c
```diff
@@ -182,6 +177,13 @@ static int _init_mem(struct ukplat_bootinfo *const bi)
 	return 0;
 }
 
+static void _init_segbase(void)
+{
+	HYPERVISOR_set_segment_base(SEGBASE_FS, 0);
```
mogasergiu: Note these return `int`

mogasergiu (approved):
One last comment. Either way, LGTM.

andreittr:

michpappas:

PR #1769, opened 2026-02-11
plat/kvm/arm: Fix exception stack switch

michpappas:
Add missing dereference to fix an issue where stack switch would not be performed during exception entry. Notice that a two-step load is necessary, as a direct PC-relative load fails as symbol distance is > 4KiB.

### Related Work

N/A

### PR Checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

PR #1772, opened 2026-02-17
lib/uklcpu: Add arm64 port

michpappas:
### Description of Changes

Add `arm64` port of `libuklcpu`, `libukpal`, `libukplat_native`.

Depends-on: #1752 
Depends-on: #1758 
Depends-on: #1767 
Depends-on: #1769 

### Related Work

N/A

### PR Checklist


lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
```diff
@@ -0,0 +1,52 @@
...
+#error "Do not include this header directly"
+#endif
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_lcpu_except_err_ctx;
+struct uk_pal_except_err_ctx;
```
mogasergiu: I think this is included here by mistake?

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
```diff
@@ -0,0 +1,52 @@
...
+
+__isr static inline int
+uk_lcpu_arm64_except_err_ctx_get_eid(const struct uk_lcpu_except_err_ctx *ctx)
```
mogasergiu, line 34: I think this is enum in the platform implementation?
If I am seeing this right, the definition of that enum is only available within the platform so there would be no point to this getter. Should that enum be defined in the PAL instead? But then how do you define lcpu wrappers...

plat/native/arch/arm64/sysctx.c
```diff
@@ -6,33 +6,29 @@
...
+#include <uk/arch/types.h>
 #include <uk/assert.h>
 
-__uptr ukarch_sysctx_get_tlsp(struct ukarch_sysctx *sysctx)
+__uptr uk_plat_native_tlsp_get(void)
 {
-	UK_ASSERT(sysctx);
-
-	return sysctx->tpidr_el0;
+	return UK_ARCH_ARM64_SYSREG_READ(tpidr_el0);
```
mogasergiu: Given how trivial these are perhaps you could have them as static inlines like it is the case on x86_64

plat/native/arch/arm64/include/uk/plat/native/arch/sysctx.h
```diff
@@ -0,0 +1,99 @@
...
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_NATIVE_ARCH_SYSCTX_H__
+#define __UK_PLAT_NATIVE_ARCH_SYSCTX_H__
+
+#include <uk/arch.h>
+#include <uk/arch/types.h>
+
+#define UK_PLAT_NATIVE_ARM64_SYSCTX_OFFSETOF_TPIDR_EL0		0
```
mogasergiu, line 14: Why not have this added to the PAL/LCPU as well, like on x86_64.

plat/native/arch/arm64/include/uk/plat/native/arch/sysctx.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Set thread-local storage pointer.
+ *
+ * @param tlsp  New TLS pointer value (typically base of TLS area)
+ */
+__isr void uk_plat_native_tlsp_set(__uptr tlsp);
+
+/**
+ * Save current CPU's system context (FS_BASE and GS_BASE) to memory.
```
mogasergiu: FS_BASE/GS_BASE? 😅 (other places as well, seems to have been copied from x86 and forgot to adapt it)

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,309 @@
...
+
+/**
+ * Handle pending interrupts.
+ */
+__isr static inline
+void uk_plat_native_irqs_handle_pending(void) { }
+
+/* Nested exception context management */
+extern __uk_pcpuvar __u8 uk_plat_native_except_switch_stack;
+extern __uk_pcpuvar __uptr uk_plat_native_except_stack_base;
```
mogasergiu, line 288: Why does this have to be per-CPU?

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,298 @@
...
+{
+	__uptr except_stack_base;
+	__sz except_stack_offs;
+	__sz pcpu_slot_offs;
+	__u32 this_cpu_idx;
+	__u64 this_cpu_id;
+	__u64 boot_cpu_id;
+	int ret;
+
+	UK_ARCH_ARM64_SYSREG_WRITE(TPIDR_EL1, (__u64)_uk_pcpuvar_base); /* FIXME */
```
mogasergiu: > FIXME

True 😅 I will send a PR for this, as currently the pcpuvar for ARM wrongly considers TPIDR_EL1 as holding the pcpuvar section base instead of cpu index * section size

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,298 @@
...
+	boot_cpu_id = uk_pcpuvar_lval(0, uk_pcpuvar_cpu_id);
+
+	if (this_cpu_id != boot_cpu_id) {
+		/* Initialize the interrupt controller */
+		ret = gic->ops.initialize();
+		if (unlikely(ret))
+			return ret;
+
+		/* Initialize TPIDR_EL1 with this cpu's pcpu base */
+		pcpu_slot_offs = this_cpu_idx * ((__sz)&_uk_pcpuvar_tmpl_size);
```
mogasergiu: Yeah so I had a macro for this multiplication at some point but I removed it because I though it wouldn't be used as all platforms will end up initializing pcpuvar early.
So then, should we add a macro fro this in ukpcpuvar?
Or should this be initialized in the arm64 entry instead? I personally favor the latter as i find it more assuring to know there is something that initializes it for me before entering C code. Furthermore, exceptions are intiialized in the entry and the exception code uses pcpuvar, meaning any exception preceding this function call will break.

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,298 @@
...
+		UK_ARCH_ARM64_SYSREG_WRITE64(tpidr_el1, (__uptr)pcpu_slot_offs);
+	}
+
+	/* Initialize this cpu's except stack base */
+	except_stack_offs = this_cpu_idx * CPU_EXCEPT_STACK_SIZE * 3;
+	except_stack_base = (__uptr)&lcpu_except_stack[except_stack_offs];
+
+	uk_pcpuvar_current_set(uk_plat_native_except_stack_base, except_stack_base);
```
mogasergiu, line 287: Ah ok so I think there is a misunderstanding about what the stack_base is for. Currently it is a temporary workaround in x86_64 because of being unable to safely get the current CPU index in an exception. ARM64 of course does not have this problem as TPIDR_EL1 is safe unlike GSBASE on x86_64. But I believe that here you meant for it to replace the way we were previously using SP_EL0. Problem is this except_stack_base has the former meaning in PAL/LCPU.
This means that you will want to define a differently named pcpuvar for this that is private to the native arm64 platform. As far as except_stack_base goes (which will get removed soon hopefully), it is meant to represent the base of the entire exception stack array, all for the purpose of:
```C
__isr __u64 uk_lcpu_get_current_idx_in_except(void)
{
	return (uk_arch_read_sp() - uk_pal_except_get_except_stack_base()) /
		(CPU_EXCEPT_STACK_SIZE * 3);
}
```

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/except.h
```diff
@@ -0,0 +1,54 @@
...
+#include <uk/arch/types.h>
+#include <uk/plat/native/arch/except.h>
+
+#if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct uk_pal_except_err_ctx;
+struct uk_plat_native_except_err_ctx;
```
mogasergiu: Is this included by mistake?

plat/native/arch/arm64/include/uk/plat/native/arch/auxsp.h
```diff
@@ -0,0 +1,32 @@
...
+#define UK_PLAT_NATIVE_AUXSP_SYM	uk_plat_native_auxsp
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if !__ASSEMBLY__
+
+extern
+__align(UK_ARCH_CACHE_LINE_SIZE) __uk_pcpuvar __uptr uk_plat_native_auxsp;
```
mogasergiu: The cache alignment is not necessary. x86 also removed it before merge.
Also guards.

plat/native/arch/arm64/include/uk/plat/native/arch/sysctx.h
```diff
@@ -0,0 +1,99 @@
...
+#include <uk/arch/types.h>
+
+#define UK_PLAT_NATIVE_ARM64_SYSCTX_OFFSETOF_TPIDR_EL0		0
+
+#define UK_PLAT_NATIVE_SYSCTX_SIZE				16
+#define UK_PLAT_NATIVE_SYSCTX_OFFSETOF_TLSP			\
+	UK_PLAT_NATIVE_ARM64_SYSCTX_OFFSETOF_TPIDR_EL0
+
+/* Assembly-visible function symbols */
+#define UK_PLAT_NATIVE_SYSCTX_LOAD_FNSYM			\
```
mogasergiu, line 22: You need to guard some of these like on x86

plat/native/arch/arm64/include/uk/plat/native/arch/ectx.h
```diff
@@ -0,0 +1,92 @@
...
+ * Copyright (c) 2018, NEC Europe Ltd., NEC Corporation. All rights reserved.
+ * Copyright (c) 2021, NEC Laboratories Europe GmbH, NEC Corporation.
+ *                     All rights reserved.
+ * Copyright (c) 2022, OpenSynergy GmbH All rights reserved.
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#if !CONFIG_LIBUKPLAT_NATIVE_ECTX
```
mogasergiu: You can remove this.

plat/native/arch/arm64/include/uk/plat/native/arch/ectx.h
```diff
@@ -0,0 +1,92 @@
...
+#define __UK_PLAT_NATIVE_ARCH_ECTX_H__
+
+#include <uk/arch/types.h>
+#include <uk/arch/util.h>
+
+#define UK_PLAT_NATIVE_ECTX_SIZE			520
+#define UK_PLAT_NATIVE_ECTX_ALIGN			16
+
+/* Assembly-visible function symbols */
+#define UK_PLAT_NATIVE_ECTX_LOAD_FNSYM					\
```
mogasergiu, line 24: Guard

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,309 @@
+
+#if !CONFIG_LIBUKPLAT_NATIVE_EXCEPT
```
mogasergiu: You can remove this. And don't forget aobut the license 🤓 ☝️

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,309 @@
...
+#include <uk/arch/types.h>
+#include <uk/arch/util.h>
+#include <uk/plat/native/arch/regs.h>
+#include <uk/compiler.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define UK_PLAT_NATIVE_EXCEPT_SWITCH_STACK_SYM				\
```
mogasergiu, line 18: Guards.

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,298 @@
...
+#include <uk/arch/types.h>
+#include <uk/arch/util.h>
+#include <uk/assert.h>
+#include <uk/event.h>
+#include <uk/intctlr/gic.h>
+#include <uk/pcpuvar.h>
+#include <uk/plat/config.h>
+#include <uk/print.h>
+
+#ifdef CONFIG_ARM64_FEAT_MTE
```
mogasergiu: #if

plat/native/arch/arm64/include/uk/plat/native/arch/start.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if CONFIG_LIBUKPLAT_NATIVE_START
+
+#if __ASSEMBLY__
+
+#define UK_PLAT_NATIVE_SENTRY_SYM					\
```
mogasergiu, line 19: I think it is fine to have these accessible from C as well

plat/native/Makefile.uk
```diff
@@ -24,9 +24,14 @@ ifeq ($(CONFIG_ARCH_ARM_64), y)
 LIBUKPLAT_NATIVE_SRCS-y                                 += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/auxsp.c
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_ECTX)   += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/ectx.c|isr
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_EXCEPT) += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/except.c|isr
+LIBUKPLAT_NATIVE_SRCS-y                                 += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/lcpu.c
```
mogasergiu: Guard?

plat/native/Makefile.uk
```diff
@@ -21,6 +21,7 @@ CXXINCLUDES-$(CONFIG_LIBUKPLAT_NATIVE_PAL) += -I$(LIBUKPLAT_NATIVE_BASE)/pal/arc
 LIBUKPLAT_NATIVE_CINCLUDES-y += -I$(UK_PLAT_COMMON_BASE)/include
 
 ifeq ($(CONFIG_ARCH_ARM_64), y)
+LIBUKPLAT_NATIVE_SRCS-y                                 += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/auxsp.c
```
mogasergiu: Guard

plat/native/arch/arm64/paging.c
```diff
@@ -11,6 +11,8 @@
 #include <uk/assert.h>
 #include <uk/plat/native/addr.h>
 
+#include <uk/asm/lcpu.h> /* FIXME tcr_ips_bits */
```
mogasergiu, line 14: Why can't we move lcpu.h right now?

drivers/firmware/uksmccc/smccc.c
```diff
@@ -0,0 +1,107 @@
...
+		conduit = uk_smccc_hvc;
+	else
+		conduit = uk_smccc_smc;
+
+	_version = config->version;
+}
+
+void uk_smccc_invoke(struct uk_smccc_args *args)
+{
+#ifndef CONFIG_FPSIMD
```
mogasergiu: Maybe #if with this occasion and `/* !CONFIG_FPSIMD *` for the #endif

drivers/firmware/ukpsci/psci.c
```diff
@@ -0,0 +1,192 @@
...
+
+	smccc_arguments.a0 = PSCI_FNID_CPU_ON;
+	smccc_arguments.a1 = id;
+
+	smccc_psci_call(&smccc_arguments);
+
+	return (int) smccc_arguments.a0;
+}
+
+#ifdef CONFIG_UKPLAT_ACPI
```
mogasergiu: I think you may need #if and updating the guard as it is now LIBUKACPI

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -2,29 +2,230 @@
...
-#if CONFIG_LIBUKOFW
+#if CONFIG_LIBUKACPI
+union {
+	struct uk_acpi_madt_gicc *gicc;
+	struct uk_acpi_subsdt_hdr *h;
+} m;
 
-#include <libfdt.h>
-#include <uk/ofw/fdt.h>
+static int enumerate_lcpu(void)
```
mogasergiu: Should this also be guarded for SMP config and why does this fit in the GIC driver? Perhaps we could also rename it to enumerate_cpus or something other than something to hint at the lcpu lib

plat/native/arch/arm64/include/uk/plat/native/arch/regs.h
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * Read a register value by byte offset.
+ *
+ * @param regs    Register context to read from
+ * @param offset  Byte offset of register (use
+ *                UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_*)
+ * @return Register value as 64-bit unsigned integer
+ */
+static inline
+__u64 uk_plat_native_regs_get(const struct uk_plat_native_regs *regs,
```
mogasergiu: Missing __isr for both setter and getter.

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/regs.h
```diff
@@ -0,0 +1,79 @@
...
+	UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_X28
+#define UK_PAL_ARM64_REGS_OFFSETOF_X29					\
+	UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_X29
+#define UK_PAL_ARM64_REGS_OFFSETOF_LR					\
+	UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_LR
+#define UK_PAL_ARM64_REGS_OFFSETOF_ELR_EL1				\
+	UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_ELR_EL1
+#define UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1				\
+	UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SPSR_EL1
+#define UK_PAL_ARM64_REGS_OFFSETOF_ESR_EL1				\
```
mogasergiu, line 77: Forgot `UK_PAL_ARM64_REGS_OFFSETOF_SP`

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/sysctx.h
```diff
@@ -0,0 +1,9 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_PAL_SYSCTX_H__
+#error "Do not include this header directly"
```
mogasergiu, line 8: Maybe guard for TPIDR. Ditto PAL/LCPU

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/except.h
```diff
@@ -0,0 +1,54 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_PAL_EXCEPT_H__
+#error "Do not include this header directly"
+#endif
+
+#include <uk/arch/types.h>
+#include <uk/plat/native/arch/except.h>
```
mogasergiu: Don't think this is needed, given neither x86 nor the other ARM64  headers need to include such headers

drivers/firmware/uksmccc/smccc_invoke.S
```diff
@@ -0,0 +1,58 @@
...
+
+.macro uk_smccc conduit
+#if CONFIG_ARM64_FEAT_PAUTH
+	/* Also covers the case when both PAuth and BTI
+	 * are enalbled (ARM DDI 0487H.a Sect. D5.4.4)
+	 */
+	paciasp
+#elif CONFIG_ARM64_FEAT_BTI
+	bti jc
+#endif
```
mogasergiu: `/* CONFIG_ARM64_FEAT_PAUTH && CONFIG_ARM64_FEAT_BTI */`

drivers/firmware/uksmccc/smccc_invoke.S
```diff
@@ -0,0 +1,58 @@
...
+	str x19, [sp, #-16]!
+
+	mov x19, x0
+
+	ldp x0, x1,   [x19]
```
mogasergiu: I think we can use this opportunity to update this weird coding style to be consistent with other parts of the codebase: one tap between instruction and first operand and whitespace between operands

plat/native/arch/arm64/ectx.c
```diff
@@ -0,0 +1,180 @@
...
+
+static inline void save_extregs(void *ectx __unused)
+{
+}
+
+static inline void restore_extregs(void *ectx __unused)
+{
+}
+
+#endif /* CONFIG_FPSIMD */
```
mogasergiu: `!CONFIG_FPSIMD`

plat/native/arch/arm64/ectx.c
```diff
@@ -0,0 +1,180 @@
...
+	UK_ASSERT(IS_ALIGNED((__uptr)state, UK_PLAT_NATIVE_ECTX_ALIGN));
+
+	/* Initialize extregs area:
+	 * Zero out and then save a valid layout to it.
+	 */
+	memset_isr(state, 0, sizeof(struct fpsimd_state));
+	uk_plat_native_ectx_store(state);
+}
+
+__isr void uk_plat_native_ectx_assert_equal(struct uk_plat_native_ectx *state)
```
mogasergiu, line 155: I think this needs guarding with the ECTX assert config

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,298 @@
...
+/*
+ *  CPU_EXCEPT_STACK_SIZE  CPU_EXCEPT_STACK_SIZE  CPU_EXCEPT_STACK_SIZE
+ *<--------------------><---------------------><-------------------->
+ *|=================================================================|
+ *|                     |                     |                     |
+ *|     crit stack      |       trap stack    |        IRQ stack    |
+ *|                     |                     |                     |
+ *|=================================================================|
+ * ^
+ * SP_EL0/lcpu_except_stack
```
mogasergiu: I guess this SP_EL0 is no longer valid. THe comment probably needs an update.

plat/native/arch/arm64/lcpu.c
```diff
@@ -0,0 +1,102 @@
...
+	 * We have to provide the physical address of the start routine when
+	 * starting secondary CPUs. We thus do the translation here once and
+	 * cache the result.
+	 */
+	lcpu_start_paddr = uk_paging_virt_to_phys((__vaddr_t)lcpu_start);
+
+	return do_arch_mp_init(arg);
+}
+
+int lcpu_arch_start(struct lcpu *lcpu, unsigned long flags __unused)
```
mogasergiu: I think you can drop the lcpu_arch and use plat_native prefix. Maybe we can also rename this as lcpu_pm.c and have ti aligned like on x86

mogasergiu (changes requested):

michpappas:
Addressed copilot review comments

drivers/firmware/uksmccc/smccc_invoke.S
michpappas: Should be `#endif /* CONFIG_ARM64_FEAT_BTI */` as the closing condition only applies to BTI. Will update.

plat/native/arch/arm64/lcpu.c
michpappas: `x86_64` is not, neither should be a point of reference. Neither reviews should attempt to make the code look as if the reviewer wrote it, especially if the reviewer is not the maintainer. I left it named `lcpu.c` for now given that the SMP part is not complete yet, and I don't know if the contents of the file will be exclusively related to PM ops.

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
michpappas: Yup, I just missed it while iterating.

plat/native/arch/arm64/include/uk/plat/native/arch/ectx.h
michpappas: Yup, also a relic from a previous iteration

plat/native/arch/arm64/lcpu.c
mogasergiu: Of course, I agree!
Pardon my wording, what I meant was this is not aligned with what we discussed during the review of the x86_64 patches. There we discussed that we can namespace these accordingly and have them exported in the exportsyms.uk (you don't do that for these, even though they are not static, so further reason for me to think that you just forgot to rename/adapt them) for usecases where you may want to be able to call them directly in external code instead of relying on LCPU API. Furthermore, even more reason to think they were just forgotten to be adapted is the fact that they 100% don't compile, given they use `struct lcpu` instead of `struct uk_lcpu`.
But I did not know that this part was just not complete yet, which makes sense given that it wouldn't compile.

plat/native/Makefile.uk
michpappas: I guess this is the same as https://github.com/unikraft/unikraft/pull/1772#discussion_r2964653106. Leaving it open for now until I also implement what's needed for SMP.

plat/native/arch/arm64/paging.c
michpappas: The original placement to `lcpu.h` is rather unfortunate. In practice TCR IPS bits control the max PA size and therefore belong to (or should be derived from) platform configuration, so I decided to do it later in the scope of boot-code related rearch.

plat/native/arch/arm64/include/uk/plat/native/arch/start.h
michpappas: Can you elaborate on the reason? AFAIK the symbols were introduced for asm, so I don't see a reason why they should be accessible from C.

plat/native/arch/arm64/ectx.c
michpappas: I don't understand the "assert" part in the comment, can you please elaborate?

mogasergiu: I am referring to `CONFIG_LIBUKPLAT_NATIVE_ECTX_ISR_ASSERTIONS`

plat/native/arch/arm64/include/uk/plat/native/arch/start.h
mogasergiu: Hmm, maybe you're right :thinking:, I don't know. When I did it on x86 I made them accessible from C as well because I thought it could allow them to be macronized in inline assembly as well as I didn't see a reason to restrict to assembly, but it's true that maybe it is better to restrict them to assembly. Really not sure. But that means that we will also have to make these updates in the PAL, native x86 and xen x86 if we choose to make them inaccessible in C, otherwise having them defined here for assembly only will be wrong as per PAL, which does allow it. So I think we should pick one or the other.

plat/native/arch/arm64/ectx.c
michpappas: Ah sure, will update

plat/native/arch/arm64/include/uk/plat/native/arch/start.h
michpappas: Ok, it's that critical anyway. I'll adapt this PR to avoid updating everything else.

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
michpappas: It was there to avoid compiler warnings when casting from the `uk_pal` type to the `uk_lcpu` type, but you're right it's actually not needed.

plat/native/arch/arm64/include/uk/plat/native/arch/sysctx.h
michpappas: Because my main criterion is not symmetry with x86 😉 Actually the rationale is that the information carried by `TPIDR_EL0` is already provided to the PAL / lcpu via the `sysctx` accessors, so additionally exporting `TPIDR_EL0` would be redundant.

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
michpappas: Previously we stored a cpu's exception stack base in `SP_EL0`, but now we have to make that register available as a 2nd scratch register as the pcpu macros require a temp register. So I moved each cpu's exception stack base to a pcpu var that I'm able to derive any time with only 2 scratch registers available.

plat/native/arch/arm64/except.c
michpappas: Yes, I'm also planning to initialize that earlier. I'll do it while working the changes for SMP.

plat/native/arch/arm64/except.c
michpappas: I will do it early, it's there temporarily.

plat/native/arch/arm64/except.c
michpappas: > But I believe that here you meant for it to replace the way we were previously using SP_EL0. 

Yes exactly (see also https://github.com/unikraft/unikraft/pull/1772#discussion_r2999398715). 

> This means that you will want to define a differently named pcpuvar for this that is private to the native arm64 platform.

I see, okay. In practice though the problem is the redundancy of the var name right? Other than that, I think the implementation is functionally correct in both the per cpu stack and the return value of `uk_plat_native_except_get_stack_base()`.

drivers/ukintctlr/gic/ukintctlr.c
michpappas: > Should this also be guarded for SMP config

Yes, the rest are guarded already, I just missed this one. Will update.

>  why does this fit in the GIC driver? 

It's more of a temporary thing - part of the PR's draft status - as I haven't concluded on what's the best library to host enumeration. Despite the comments and assertions, there is no strong reason to have an initialized GIC for enumeration, the requirement is only for *initializing* secondary cores as they call into `gic->ops`.

> Perhaps we could also rename it to enumerate_cpus or something other than something to hint at the lcpu lib

Yeah sure we can do that to match the name of the initialzied pcpu vars.

plat/native/arch/arm64/ectx.c
michpappas: So this causes the CI build to fail in app-helloworld due to a warning, as the prototype in the PAL is not conditional to any config directives. Interestingly on x86_64 the relevant function is not conditional either, only calls to it.

plat/native/arch/arm64/include/uk/plat/native/arch/sysctx.h
mogasergiu: So then why does this macro have to exist if we guarantee that the accessors will always access this field through  TLSP rather than TPIDR_EL0? This is also why the arch specific sysctx.h headers in lcpu/pal are empty

plat/native/arch/arm64/except.c
mogasergiu: I think maybe we could try to give a different name to this then so that it does not create confusion. Something that might imply it is per-lcpu or something or somehow not the same thing as `lcpu_except_stack`?

plat/native/arch/arm64/ectx.c
mogasergiu: Oh well... maybe then it makes sense to have it available as it will be optimized out anyway. What doesn't make sense I think now is the fact that it is also defined by the PAL? Nothing uses it outside the platform... Should we just let it be unguarded and only guard the interrupt handling part where it is called? Moreover, should we remove it from PAL as well?

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
mogasergiu: Does this still need to be included here?

lib/ukpal/arch/arm64/include/uk/pal/arch/regs.h
```diff
@@ -0,0 +1,145 @@
...
+
+#ifndef UK_PAL_ARM64_REGS_OFFSETOF_ELR_EL1
+#error "UK_PAL_ARM64_REGS_OFFSETOF_ELR_EL1 undefined"
+#endif
+
+#ifndef UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1
+#error "UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1 undefined"
+#endif
+
+#ifndef UK_PAL_ARM64_REGS_OFFSETOF_ESR_EL1
```
mogasergiu, line 143: SP?

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h
```diff
@@ -0,0 +1,101 @@
...
+	UK_PAL_ARM64_REGS_OFFSETOF_X28
+#define UK_LCPU_ARM64_REGS_OFFSETOF_X29					\
+	UK_PAL_ARM64_REGS_OFFSETOF_X29
+#define UK_LCPU_ARM64_REGS_OFFSETOF_LR					\
+	UK_PAL_ARM64_REGS_OFFSETOF_LR
+#define UK_LCPU_ARM64_REGS_OFFSETOF_ELR_EL1				\
+	UK_PAL_ARM64_REGS_OFFSETOF_ELR_EL1
+#define UK_LCPU_ARM64_REGS_OFFSETOF_SPSR_EL1				\
+	UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1
+#define UK_LCPU_ARM64_REGS_OFFSETOF_ESR_EL1				\
```
mogasergiu, line 77: SP again? Though it also seems a bit redundant in a way given that we also have SP as the generic one... maybe for consistency only since it's still a struct member?

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,313 @@
...
+#define __UK_PLAT_NATIVE_ARCH_EXCEPT_H__
+
+#include <uk/arch/types.h>
+#include <uk/arch/util.h>
+#include <uk/plat/native/arch/regs.h>
+#include <uk/compiler.h>
+#include <uk/config.h>
+
+#if CONFIG_LIBUKPLAT_NATIVE_EXCEPT
+#define UK_PLAT_NATIVE_EXCEPT_SWITCH_STACK_SYM				\
```
mogasergiu, line 18: Do we need this and the one below it to have SYM macros? We only needed them because of PAL introducing an additional layer and we wouldn't be able to use .equiv/.set.
The users of these symbols using the UK_PLAT_NATIVE macro is kind of redundant. The only difference is that they use uppercase lettters to access the symbol but the namespace is the same, it no longer has the reasoning that it is required to be forwarded through the PAL.

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
```diff
@@ -0,0 +1,313 @@
...
+/* Nested exception context management */
+extern __uk_pcpuvar __u8 uk_plat_native_except_switch_stack;
+extern __uk_pcpuvar __uptr uk_plat_native_except_stack_base;
+
+__isr __uptr uk_plat_native_except_get_except_stack_base(void);
+
+static inline
+void uk_plat_native_except_push_nested(void)
+{
+	uk_plat_native_except_switch_stack++;
```
mogasergiu: If you do go with per-CPU variables for this, I think this increment/decrement need to use pcpuvar accessors?

plat/native/arch/arm64/except.c
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * Why can one stack not corrupt the other/Why is trap stack before IRQ stack?
+ * During a trap IRQs are disabled. So only the trap stack could potentially
+ * corrupt the IRQ stack if they were the other way around. The ordering of the
+ * stacks is made so that this cannot happen. Same thing goes for the crit
+ * stack as a crit exception may happen during a trap.
+ *
+ */
+static __align(UKARCH_SP_ALIGN)
+__u8 lcpu_except_stack[CONFIG_UKPLAT_CPU_MAXCOUNT][3 * CPU_EXCEPT_STACK_SIZE];
```
mogasergiu: Perhaps we could rename this to not hint at the lcpu library

plat/native/exportsyms.uk
```diff
@@ -23,18 +22,23 @@ uk_plat_native_ectx_init
...
 uk_plat_native_except_get_except_stack_base
+uk_plat_native_except_init
+uk_plat_native_except_inv_handler
+uk_plat_native_except_irq_handler
+uk_plat_native_except_push_nested
+uk_plat_native_except_pop_nested
+uk_plat_native_except_send_ipi
+uk_plat_native_except_switch_stack
+uk_plat_native_except_stack_base
+traps_table_init
```
mogasergiu: I think this can be dropped

plat/native/arch/arm64/auxsp.c
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#include <uk/arch/types.h>
+#include <uk/arch.h>
+#include <uk/pcpuvar.h>
+
+__uk_pcpuvar __align(UK_ARCH_CACHE_LINE_SIZE) __uptr uk_plat_native_auxsp;
```
mogasergiu: Forgot to remove the cache line alignment

plat/native/arch/arm64/lcpu_pm.c
```diff
@@ -0,0 +1,73 @@
...
+
+static int plat_native_lcpu_start(__u64 idx)
+{
+	__u64 cpuid;
+	__u64 entry;
+
+	cpuid = uk_pcpuvar_lval(idx, uk_pcpuvar_cpu_id);
+	entry = uk_paging_virt_to_phys((__vaddr_t)lcpu_start);
+
+	return uk_psci_cpu_on(cpuid, entry, idx);
```
mogasergiu, line 37: This introduces a dependency of the platform on PSCI.

plat/native/arch/arm64/lcpu.c
```diff
@@ -0,0 +1,73 @@
...
+	 * See Linux `cpu_do_idle(void)` implementation
+	 */
+	uk_arch_arm64_halt();
+}
+
+static const struct uk_lcpu_pm_ops plat_native_pm_ops = {
+#if CONFIG_HAVE_SMP
+	.start = plat_native_lcpu_start,
+#if CONFIG_HAVE_CPU_MULTI_PHASE_STARTUP
+	.post_start = NULL,
```
mogasergiu: I am not sure this is necessary. Probably leftover from the x86 code.

plat/native/arch/arm64/lcpu_pm.c
```diff
@@ -0,0 +1,73 @@
...
+	.halt_irq = plat_native_lcpu_halt_irq,
+};
+
+__isr static
+int plat_native_lcpu_pm_ops_register(struct ukplat_bootinfo *bi __unused)
+{
+	return uk_lcpu_pm_ops_register(&plat_native_pm_ops);
+}
+
+UK_BOOT_EARLYTAB_ENTRY(plat_native_lcpu_pm_ops_register, UK_PRIO_EARLIEST);
```
mogasergiu, line 73: Similar to x86 and according to your own suggestion, we could guard this with CONFIG_LIBUKBOOT so that we avoid ukboot dep

plat/native/Makefile.uk
```diff
@@ -24,9 +24,14 @@ ifeq ($(CONFIG_ARCH_ARM_64), y)
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_AUXSP)   += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/auxsp.c
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_ECTX)   += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/ectx.c|isr
 LIBUKPLAT_NATIVE_SRCS-$(CONFIG_LIBUKPLAT_NATIVE_EXCEPT) += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/except.c|isr
+LIBUKPLAT_NATIVE_SRCS-y                                 += $(LIBUKPLAT_NATIVE_BASE)/arch/arm64/lcpu.c
```
mogasergiu: CONFIG_LIBUKPLAT_NATIVE_LCPU_PM? Oh wait you included it again below. I think this line is added twice by mistake right?:

plat/kvm/arm/entry64.S
```diff
@@ -129,6 +129,11 @@ ENTRY(_libkvmplat_entry)
 	/* Enable the mmu */
 	bl start_mmu
 
+	/* Set TPIDR_EL1 for BSP (idx=0): base + 0 * size = base */
+	adrp	x1, _uk_pcpuvar_base
+	add	x1, x1, #:lo12:_uk_pcpuvar_base
```
mogasergiu: ur_addr?

drivers/firmware/uksmccc/Config.uk
```diff
@@ -0,0 +1,5 @@
+menuconfig LIBUKSMCCC
```
mogasergiu: Do we use menuconfigs even if the menu is entry?

drivers/firmware/uksmccc/smccc.c
```diff
@@ -0,0 +1,107 @@
...
+void uk_smccc_arch_workaround_2(void)
+{
+	struct uk_smccc_args args = {0};
+
+	args.a0 = SMCCC_FID_ARCH_WORKAROUND_2;
+
+	uk_smccc_invoke(&args);
+}
+
+unsigned long uk_smccc_svc_query(unsigned long service, unsigned long type)
```
mogasergiu: Should we use a more explicit sized type instead of unsigned long? Ditto other unsigned longs here.

drivers/ukintctlr/gic/ukintctlr.c
```diff
@@ -2,23 +2,241 @@
...
+			/* Handle gracefully */
+			uk_pr_warn("Maximum number of cores exceeded.\n");
+			return 0;
+		}
+	}
+	UK_ASSERT(bsp_found);
+
+	return 0;
+}
+#endif /* CONFIG_HAVE_SMP */
```
mogasergiu: Why not guard with a single CONFIG_HAVE_SMP both the CONFIG_LIBUKACPI and CONFIG_LIBUKOFW arms

plat/common/arm/lcpu_start.S
```diff
@@ -47,12 +47,19 @@ ENTRY(lcpu_start)
 	/* Disable interrupts */
 	msr	daifset, #2
 
-	/* Set the context id */
-	msr	contextidr_el1, xzr
-
 	/* Preserve CPU index across start_mmu (x19 is not clobbered) */
 	mov	x19, x0
 
+#ifdef CONFIG_FPSIMD
```
mogasergiu: #if?

plat/native/arch/arm64/lcpu_pm.c
michpappas: Yes, but there's not much we can do about it. Actually I have been considering lately whether cpu pm ops should be part of the libukpm, but that's something out of scope of this PR. Unless you mean we should express that dependency in Kconfig? Can you please clarify what is the suggestion here?

drivers/firmware/uksmccc/Config.uk
michpappas: I'll switch to `config`

drivers/firmware/uksmccc/smccc.c
michpappas: Yes, back when this was written it supported both the 32-bit and 64-bit variants of the spec. Now we have no reason for that, I'll switch to `__u64`.

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
michpappas: This is done already

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
michpappas: I'm okay with it, it's easier for users of the API to have to deal with a single convention.

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h
michpappas: I don't understand what you mean here. OFFSETOF_SP is defined right above.

michpappas:
Addressed comments in @mogasergiu's last review round

plat/native/arch/arm64/except.c
michpappas: Done

michpappas:
Addressed last round of copilot comments

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/sysctx.h
michpappas: I don't understand this comment, please elaborate

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h
mogasergiu: What I mean is that you have
UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP
UK_PAL_ARM64_REGS_OFFSETOF_SP
but you don't have the 
UK_LCPU_ARM64_REGS_OFFSETOF_SP

plat/native/arch/arm64/include/uk/plat/native/arch/except.h
mogasergiu: I am not sure I understand the convention. As it is now it is clear that ECTX/SYSCTX are treated as an exception for symbols that we want to generically export through PAL so that they are accessible in assembly and inline assembly that must not be aware of the platform code and is a result of the unforeseen PAL limitation as an intermediate layer. But having these two here actually make it inconsistent because what is the justification for having these (which are not necessary to be accessible outside the platform) as SYM macros but not all the other syms?
And, more generically, why is this not the case for every single other symbol? But maybe I should assume your convention is targeting symbols that are only accessed in assembly? If that is the case you may want to also update:
```
plat/kvm/arm/exceptions.S:21:.extern uk_plat_native_except_switch_stack
plat/kvm/arm/exceptions.S:22:.extern uk_plat_native_except_stack_base
```
to use the uppercased macros.
But then if my previous assembly-only assumption is not wrong then why does this convention not apply to other symbols as well. For example, why is this not the case for:
- uk_plat_native_except_err_handler
- uk_plat_native_except_irq_handler
- uk_plat_native_except_inv_handler

Moreover, the already existing symbols with the similar assumed targeted usage also become inconsistent:
```
$ git grep -n uk_plat | grep -E ".S:"
plat/kvm/arm/exceptions.S:21:.extern uk_plat_native_except_switch_stack
plat/kvm/arm/exceptions.S:22:.extern uk_plat_native_except_stack_base
plat/kvm/arm/exceptions.S:279:  bl uk_plat_native_except_err_handler
plat/kvm/arm/exceptions.S:287:  bl uk_plat_native_except_irq_handler
plat/kvm/arm/exceptions.S:304:  b   uk_plat_native_except_inv_handler;  \
plat/kvm/x86/cpu_vectors_x86_64.S:112:  call uk_plat_native_except_err_handler
plat/kvm/x86/cpu_vectors_x86_64.S:155:  call uk_plat_native_except_irq_handler
plat/kvm/x86/lcpu_start.S:54:.globl uk_plat_native_x86_64_start16_addr
plat/kvm/x86/lcpu_start.S:55:uk_plat_native_x86_64_start16_addr:
plat/kvm/x86/lxboot.S:54:       movq    %rax, uk_plat_native_sstackp(%rip)
plat/kvm/x86/lxboot.S:57:       movq    %rax, uk_plat_native_sentry(%rip)
plat/kvm/x86/lxboot.S:59:       movq    %rsi, uk_plat_native_sarg(%rip)
plat/kvm/x86/multiboot.S:73:    ur_mov  uk_plat_native_sstackp, %esi, 4, _phys
plat/kvm/x86/multiboot.S:77:    ur_mov  uk_plat_native_sentry, %esi, 4, _phys
plat/kvm/x86/multiboot.S:80:    ur_mov  uk_plat_native_sarg, %esi, 4, _phys
plat/xen/x86/entry64.S:279:     call uk_plat_xen_except_err_handler
plat/xen/x86/entry64.S:313:     cmovzq UK_PCPUVAR_X86_64_GSREL(uk_plat_xen_auxsp), %rax
```
Should we update all of these as well? Since this is code that is meant to be platform-aware or even part of the platform I am not sure I see the benefit and it feels loose: what decides whether a symbol will be used in assembly or not - this sounds like something incremental, i.e. you end up creating a macro every time you realise you want to access a symbol in assembly within the same platform it is defined. Or do you mean that this is meant to be a convention that is ARM64 native only? In that case how should it be documented to be consistent?

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h
michpappas: From what I see `UK_LCPU_ARM64_REGS_OFFSETOF_SP` is defined on line 80

plat/native/pal/arch/arm64/include/uk/plat/pal/arch/sysctx.h
mogasergiu: This ties to my comment/our discussion around the TPIDR_EL0 macro
https://github.com/unikraft/unikraft/pull/1772#discussion_r3023479701

> So then why does this macro have to exist if we guarantee that the accessors will always access this field through TLSP rather than TPIDR_EL0? This is also why the arch specific sysctx.h headers in lcpu/pal are empty

The reason I had these on x86_64 is:
1. Be consistent with every other struct and accessors, being able to access the struct member by its architecturally-defined name: why can we access the members by their name in other architecture specific structs e.g. x20 in regs, but not for sysctx.
2. Be able to access the struct member by its low level meaning: the register name, not its definition as mandated by the userspace ABI. This helps low-level architecture-specific code that doesn't need to know the higher-level meaning/name of the register (e.g. TLS register or per-CPU variables register) access individual fields of the struct for what they are: the raw register.

This comment was part of the first round of review where I assumed you wrongly forgot to add PAL/LCPU wrappers for TPIDR_EL0 offset macro, because:
1. You ended up with 2 empty sysctx.h registers
2. You still define the macro but it is never used for anything other than to define the TLS offset macro one line below it.

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/except.h
michpappas: Hmmm good point, I'll push an update to define these on the PAL and provide wrappers on libuklcpu.

lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h
mogasergiu: Yes, sorry, bruh I misread it as spsr when grepping :man_facepalming: 
```
$ git grep -n _ARM64_REGS_OFFSETOF_SP
lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h:75:#define UK_LCPU_ARM64_REGS_OFFSETOF_SPSR_EL1                         \
lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h:76:   UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1
lib/uklcpu/arch/arm64/include/uk/lcpu/arch/regs.h:80:#define UK_LCPU_ARM64_REGS_OFFSETOF_SP                                       \
lib/ukpal/arch/arm64/include/uk/pal/arch/regs.h:139:#ifndef UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1
lib/ukpal/arch/arm64/include/uk/pal/arch/regs.h:140:#error "UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1 undefined"
plat/kvm/arm/exceptions.S:46:    * aligned sp[UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP]
plat/kvm/arm/exceptions.S:49:   str x0, [x1, #UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP]
plat/kvm/arm/exceptions.S:58:   str x0, [x0, #UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP]
plat/kvm/arm/exceptions.S:189:  str     x0, [sp, #UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP]  /* Store old SP in auxiliary stack */
plat/kvm/arm/exceptions.S:259:  ldr     x18, [sp, #UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP]
plat/native/arch/arm64/include/uk/plat/native/arch/regs.h:50:#define UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SPSR_EL1  256
plat/native/arch/arm64/include/uk/plat/native/arch/regs.h:52:#define UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP                272
plat/native/arch/arm64/include/uk/plat/native/arch/regs.h:58:   UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP
plat/native/arch/arm64/include/uk/plat/native/arch/regs.h:182:      UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SPSR_EL1);
plat/native/arch/arm64/include/uk/plat/native/arch/regs.h:186:      UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP);
plat/native/pal/arch/arm64/include/uk/plat/pal/arch/regs.h:75:#define UK_PAL_ARM64_REGS_OFFSETOF_SPSR_EL1                         \
plat/native/pal/arch/arm64/include/uk/plat/pal/arch/regs.h:76:  UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SPSR_EL1
plat/native/pal/arch/arm64/include/uk/plat/pal/arch/regs.h:79:#define UK_PAL_ARM64_REGS_OFFSETOF_SP                                       \
plat/native/pal/arch/arm64/include/uk/plat/pal/arch/regs.h:80:  UK_PLAT_NATIVE_ARM64_REGS_OFFSETOF_SP
```
Yes it is defined everywhere accordingly.

michpappas:
Address https://github.com/unikraft/unikraft/pull/1772#discussion_r2982401397 and `s/menuconfig/config` in `drivers/firmware/ukpsci/Config.uk`.

michpappas:
Rebased from `staging` (#1807), adapted `irq` to `__u32`, updated `uksmccc` function typedefs to use `_func`, removed kconfig conditional for ectx assert equal.

mogasergiu (approved):
LGTM!

PR #1774, opened 2026-02-22
lib: Introduce per-CPU variables library

mogasergiu:
### Description of Changes

This series introduces libukpcpuvar, a library providing efficient per-CPU variable support for Unikraft systems, and migrates the rest of the codebase to use it.

Per-CPU variables allow each CPU to maintain its own copy of data without locks. They are particularly useful in boot code, "risky" assembly (where you cannot touch many registers, e.g. exception entry) and execution paths where false sharing between CPUs would otherwise degrade performance. Variables are declared with `__uk_pcpuvar`, placing them in a dedicated `.uk_pcpuvar` ELF section. The linker script reserves space for `CONFIG_UKPLAT_CPU_MAXCOUNT` copies at build time, and a post-build script (`mkukpcpuvar.py`) replicates the template into the binary after linking but before stripping. Because variables are image built-in, they are always accessible regardless of boot stage, with no allocator required.

Following the introduction of the library, the series migrates `lib/` and `plat/` to use the new accessors, replacing the legacy `UK_PER_LCPU_*/uk_per_lcpu_*` macros in libuklcpu. Early boot assembly in `plat/kvm/x86` is updated to use the `uk_pcpuvar_x86_64_gsval` macro for computing the per-CPU slot offset. Platform data that was previously stored as per-CPU variables but is more naturally indexed by `lcpu->idx` (such as `cpu_idt`, `cpu_tss`, and `cpu_gdt64`) is replaced with plain `CONFIG_UKPLAT_CPU_MAXCOUNT` C arrays.

The series also removes `idx`, `id`, and `auxsp` and `lcpu_sargs`'s `entry`/`stackp` from the lcpu structure, defining them instead as proper per-CPU variables owned by libukpcpuvar and PAL respectively. The `lcpu_sargs` struct is removed as a consequence. `uk_pal_lcpu_init` is renamed to `uk_pal_except_init` to reflect that per-CPU base register initialization is the platform's responsibility and not part of the PAL contract. The series further folds in ACPI library refactoring (`UKPLAT_ACPI` → `LIBUKACPI`) and miscellaneous cleanups made possible by the removal of the legacy per-LCPU infrastructure.

This library intentionally provides only the section, its post-build duplication, and the variable accessors. Per-CPU register initialization and variable constructors are out of scope by design, as the variety of platform mechanisms (e.g. MSR write on KVM vs. hypercall on Xen PV) and the possibility that initialization must occur before C code can run make a generic initialization API infeasible. A pcpuvar-aware component such as platform boot code is expected to own this.

### Related Work

N/A

### PR Checklist


lib/ukpcpuvar/README.md
michpappas, line 1: Let's follow the "one sentence per line" rule to make it easier to review and read.

lib/ukpcpuvar/Makefile.uk
```diff
@@ -0,0 +1,12 @@
+$(eval $(call addlib_s,libukpcpuvar,$(CONFIG_LIBUKPCPUVAR)))
+
+ASINCLUDES-y  += -I$(LIBUKPCPUVAR_BASE)/arch/$(CONFIG_UK_ARCH)/include
```
michpappas: Are the headers unconditionally available by intention?

lib/ukpcpuvar/Makefile.uk
```diff
@@ -0,0 +1,12 @@
...
+CINCLUDES-y   += -I$(LIBUKPCPUVAR_BASE)/arch/$(CONFIG_UK_ARCH)/include
+CXXINCLUDES-y += -I$(LIBUKPCPUVAR_BASE)/arch/$(CONFIG_UK_ARCH)/include
+
+ASINCLUDES-y  += -I$(LIBUKPCPUVAR_BASE)/include
+CINCLUDES-y   += -I$(LIBUKPCPUVAR_BASE)/include
+CXXINCLUDES-y += -I$(LIBUKPCPUVAR_BASE)/include
+
+LIBUKPCPUVAR_SRCS-$(CONFIG_LIBUKPCPUVAR) += $(LIBUKPCPUVAR_BASE)/pcpuvar.lds.S
```
michpappas: Normally C / asm files are declared unconditionally, i.e. `LIBUKPCPUVAR_SRCS-y`.

lib/ukpcpuvar/arch/arm64/include/uk/pcpuvar/arch.h
```diff
@@ -0,0 +1,225 @@
...
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PCPUVAR_H__
+#error "Do not include this header directly"
+#endif
+
+#include <uk/essentials.h>
+
```
michpappas, line 12: I believe the C++ guard needs to move here, and then alternate between asm / non-asm directives

lib/ukpcpuvar/include/uk/pcpuvar.h
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @param _idx
+ *   CPU index
+ * @param _sym
+ *   Symbol name
+ * @return
+ *   Lvalue reference (can read or write)
+ */
+#define __uk_pcpuvar_lval(_idx, _sym)					\
+	(*(__typeof__(_sym) *)						\
```
michpappas, line 50: Uhhh although `__typeof__` fits perfectly here, it's a C23 feature 😬

lib/uksched/include/uk/sched_impl.h
```diff
@@ -102,11 +102,11 @@ void uk_sched_thread_switch(struct uk_thread *next)
 {
 	struct uk_thread *prev;
 
-	prev = uk_per_lcpu_current(__uk_sched_thread_current);
+	prev = uk_pcpuvar_current_get(__uk_sched_thread_current);
```
michpappas, line 105: I guess `libuksched` needs to also select `LIBUKPCPUVAR` in its `Config.uk`

mogasergiu:
Previous fpush:
- add CONFIG_LIBUKPLAT_NATIVE_LCPU_PM config so that different platforms can individually (not) select native's lcpu_pm_ops implementation

Second fpush:
- forgot to update `_ukplat_entry` signature in `efi_post.c`

mogasergiu:
fpush:
- Forgot to config-guard ukpcpuvar includes (thanks @andreittr )

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+#define UK_LCPU_STATE_HALTED		-1 /* CPU stopped, needs reset */
+#define UK_LCPU_STATE_OFFLINE		0 /* CPU not started */
+#define UK_LCPU_STATE_INIT		1 /* CPU started, init not finished */
+#define UK_LCPU_STATE_IDLE		2 /* CPU is idle */
+#define UK_LCPU_STATE_BUSY0		3 /* >= CPU is busy */
+
+#if !__ASSEMBLY__
+struct __align(UK_ARCH_CACHE_LINE_SIZE) uk_lcpu {
+	/* Current CPU state (UK_LCPU_STATE_*).
+	 * Working on it with atomic instructions - must be 8-byte aligned
```
michpappas: Gramatically "Working on it with atomic instructions" makes no sense. Perhaps it means "Access using atomic instructions?". Also is the alignment comment necessary given that the struct is already aligned with the `__align` attribute?

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+ * Logical CPU (LCPU) Structure
+ */
+#define UK_LCPU_STATE_OFFSET		0x00
+#define UK_LCPU_ERR_OFFSET		(UK_LCPU_STATE_OFFSET + 0x4)
+#if CONFIG_HAVE_SMP
+#define UK_LCPU_FN_OFFSET		(UK_LCPU_ERR_OFFSET + 0x4)
+#endif /* CONFIG_HAVE_SMP */
+
+#if CONFIG_HAVE_SMP
+#define UK_LCPU_MEMBERS_SIZE						\
```
michpappas: Given the `+4` in the non-SMP case to me it looks like this should be named `UK_LCPU_STRUCT_SIZE`, as `MEMBERS_SIZE` implies we don't take padding into consideration. Also overall besides the CTASSERTs, is the `UK_LCPU_SIZE` intended to be used by anything? In other words, are these macros and CTASSERTs useful for something?

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+
+UK_CTASSERT(sizeof(struct uk_lcpu) == UK_LCPU_SIZE);
+UK_CTASSERT(UK_LCPU_MEMBERS_SIZE <= UK_LCPU_SIZE);
+#endif /* !__ASSEMBLY__ */
+
+#if !__ASSEMBLY__
+#include <uk/arch/time.h>
+#include <uk/essentials.h>
+
+#ifdef __cplusplus
```
michpappas: You should need the C++ guard to the top, as currently there is unguarded C code. Then inside the C++ guard you can have the asm part followed by the non-asm part. Also there is no reason for the headers to be inside any guards, they can be moved to the top.

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+static inline struct uk_lcpu *uk_lcpu_get_bsp(void)
+{
+	/* The BSP is always index 0 */
+	return &uk_pcpuvar_lval(0, uk_lcpus);
+}
+
+/**
+ * Return 1 if the supplied UK_LCPU is the boottrap processor, 0 otherwise
+ */
+static inline int uk_lcpu_is_bsp(struct uk_lcpu *lcpu)
```
michpappas: The function essentially returns `__bool`, should we update the return type to reflect that?

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+{
+	return (lcpu == uk_lcpu_get_bsp());
+}
+
+/**
+ * Return 1 if the executed on the bootstrap processor, 0 otherwise
+ */
+static inline int uk_lcpu_current_is_bsp(void)
```
michpappas: The function essentially returns `__bool`, should we update the return type to reflect that?

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+ * @param run_irq the IRQ vector to use for running remote functions
+ * @param wakeup_irq the IRQ vector to use for waking up CPUs
+ *
+ * @return 0 on success, -errno otherwise
+ */
+int uk_lcpu_mp_init(unsigned long run_irq, unsigned long wakeup_irq);
+
+/**
+ * Default entry function for secondary logical CPUs. Will call lcpu_init() and
+ * If the logical CPU's startup arguments supply an entry function, the
```
michpappas: small-case `if` as you continue a senctence from a previous line

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+ * original stack pointer will be restored and execution continues in the
```
michpappas: Maybe "... original stack pointer *is* restored" to match the rest of the paragraph.

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * @param run_irq the IRQ vector to use for running remote functions
+ * @param wakeup_irq the IRQ vector to use for waking up CPUs
+ *
+ * @return 0 on success, -errno otherwise
+ */
+int uk_lcpu_mp_init(unsigned long run_irq, unsigned long wakeup_irq);
+
+/**
+ * Default entry function for secondary logical CPUs. Will call lcpu_init() and
```
michpappas: "Calls lcpu_init()...", see https://github.com/unikraft/unikraft/pull/1774/changes#r2844736395 for reasoning.

lib/uklcpu/lcpu.c
```diff
@@ -59,177 +32,124 @@
...
-	return lcpu;
-}
-
-__u32 ukplat_lcpu_count(void)
-{
-	return lcpu_count;
-}
-#else
-#define lcpu_count	(1)
+__uk_pcpuvar struct uk_lcpu uk_lcpus;
```
michpappas: If this is an array containing all lcpus, why does it have to be a per-cpu variable and not a global?

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
```diff
@@ -288,6 +288,17 @@ __isr void uk_plat_native_except_push_nested(void);
...
+
+#if CONFIG_HAVE_SMP
+/**
+ * Send an Interprocessor Interrupt.
+ *
+ * @param id The CPU ID to send the IPI to
+ * @param irq The IRQ to send to the CPU
+ * @return 0 on success, negative errno on failure
+ */
+int uk_plat_native_except_send_ipi(__u64 id, unsigned long irq);
```
michpappas, line 300: Quoting the commit message:
> The `lcpu.h` is an unpleasant reference to a library the PAL is not
> supposed to be aware of and whose existence is no longer necessary:
> it only contains the IPI sending prototype. Therefore, remove it
> and move the IPI sending prototype to the exception namespace, as it
> essentially implies synchronously sending an exception to another CPU
> anyway.

I don't have a strong opinion, but it feels a bit strange to see IPI under the except API, as it is predominantly a inter-processor communication mechanism. The migration signifies a mechanism-based approach. It also increases the scope of the except API from handling exceptions to also triggering exceptions. So if in the future we had to abstract other operations that cause exceptions -  BRK, HVC, SMC, SVC  on arm64 - we would naturally need to include them into the except API too.

PS:

> synchronously sending an exception

At least on arm64 IRQs are considered asynchronous exceptions 😛

lib/uklcpu/lcpu.c
```diff
@@ -364,34 +280,41 @@ void __weak __noreturn lcpu_entry_default(struct lcpu *this_lcpu)
...
-	int rc = 0;
-#ifdef LCPU_ARCH_MULTI_PHASE_STARTUP
-	int rc2;
-#endif /* LCPU_ARCH_MULTI_PHASE_STARTUP */
+	const int new = UK_LCPU_STATE_INIT;
+	int old, rc = 0, rc2;
 
-	UK_ASSERT(((lcpuidx) && (num)) || ((!lcpuidx) && (!num)));
+	UK_ASSERT(lcpuidx);
+	UK_ASSERT(num);
```
andreittr, line 304: These 2 asserts change the semantics of the function; the [documentation](https://github.com/unikraft/unikraft/pull/1774/changes#diff-0bc203f1335285055f9cda93f0ca2116519c2085d6e6fc616cd3602ceef18d5d) still suggests that NULLs are accepted as args (and, AFAICT, the rest of the implementation still follows that logic).

lib/uklcpu/lcpu.c
```diff
@@ -489,35 +419,39 @@ static inline int lcpu_transition_safe(struct lcpu *lcpu, int incr)
...
-	struct lcpu *lcpu;
+	__u64 this_cpu_id = uk_pcpuvar_current_get(uk_pcpuvar_cpu_id);
+	struct uk_lcpu *lcpu;
 	unsigned int n, i;
-	int rc;
+	int rc = 0;
 
-	UK_ASSERT(((lcpuidx) && (num)) || ((!lcpuidx) && (!num)));
+	UK_ASSERT(lcpuidx);
+	UK_ASSERT(num);
```
andreittr, line 427: Similar argument to `uk_lcpu_start`: the documentation still suggests that NULLs are accepted.

andreittr (commented):
Great work! Just one small oversight in lcpu patches.

mogasergiu:
fpush:
- remove traces of previoud lcpu_[start|wake|run] `NULL` as arg functionality
<details>
<summary>Click to expand diff</summary>

```diff
diff --git a/lib/uklcpu/include/uk/lcpu.h b/lib/uklcpu/include/uk/lcpu.h
index 0b8e24a6..32b3b51c 100644
--- a/lib/uklcpu/include/uk/lcpu.h
+++ b/lib/uklcpu/include/uk/lcpu.h
@@ -240,12 +240,11 @@ int uk_lcpu_fn_enqueue(struct uk_lcpu *lcpu, const struct uk_lcpu_func *fn);
  * Waits for the specified logical CPUs to enter idle state, or until the
  * timeout expires.
  *
- * @param lcpuidx array with the indices of the logical CPUs to wait for. Can
- *   be NULL to include all logical CPUs except the one executing the function
- * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
- *   elements in lcpuidx, and [OUT] the number of CPUs in idle state until the
- *   timeout expired in sequential order of lcpuidx. If the call succeeds,
- *   input and output values are equal. Must be NULL if lcpuidx is NULL
+ * @param lcpuidx array with the indices of the logical CPUs to wait for.
+ * @param[inout] num provides [IN] the number of elements in lcpuidx,
+ *   and [OUT] the number of CPUs in idle state until the timeout expired
+ *   in sequential order of lcpuidx. If the call succeeds, input and output
+ *   values are equal
  * @param timeout timeout in nanoseconds for the wait to be satisfied. Can be
  *   0 to wait indefinitely
  *
@@ -261,13 +260,11 @@ int uk_lcpu_wait(const __u32 lcpuidx[], unsigned int *num,
  * at the same time without having to wait for their completion.
  *
  * @param lcpuidx array with the indices of the logical CPUs that should
- *   execute the function. Can be NULL to execute the function on all logical
- *   CPUs except the current one
- * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
- *   elements in lcpuidx, and [OUT] the number of CPUs on which the function has
+ *   execute the function.
+ * @param[inout] num provides [IN] the number of elements in lcpuidx,
+ *   and [OUT] the number of CPUs on which the function has
  *   been successfully queued in sequential order of lcpuidx. If the call
- *   succeeds, input and output values are equal. Must be NULL if lcpuidx is
- *   NULL
+ *   succeeds, input and output values are equal.
  * @param fn the function to be executed
  * @param flags (architecture-dependent) flags that specify how the function
  *   should be executed (see UK_LCPU_RFLG_* flags)
@@ -284,11 +281,11 @@ int uk_lcpu_run(const __u32 lcpuidx[], unsigned int *num,
  * Wakes up the specified logical CPUs from a halt or low-power sleep state.
  *
  * @param lcpuidx array with the indices of the logical CPUs that should be
- *   woken up. Can be NULL to wakeup all logical CPUs except the current one
- * @param[inout] num if lcpuidx is not NULL, provides [IN] the number of
- *   elements in lcpuidx, and [OUT] the number of successfully woken up CPUs in
- *   sequential order of lcpuidx. If the call succeeds, input and output values
- *   are equal. Must be NULL if lcpuidx is NULL
+ *   woken up
+ * @param[inout] num provides [IN] the number of elements in lcpuidx,
+ *   and [OUT] the number of successfully woken up CPUs in sequential
+ *   order of lcpuidx. If the call succeeds, input and output values
+ *   are equal.
  *
  * @return 0 on success, an errno-type error value otherwise
  */
diff --git a/lib/uklcpu/lcpu.c b/lib/uklcpu/lcpu.c
index 92c64dd2..c8ef0d41 100644
--- a/lib/uklcpu/lcpu.c
+++ b/lib/uklcpu/lcpu.c
@@ -295,10 +295,10 @@ int uk_lcpu_start(const __u32 lcpuidx[], unsigned int *num,
 		  unsigned long flags __unused)
 {
 	__u64 this_cpu_id = uk_pcpuvar_current_get(uk_pcpuvar_cpu_id);
-	struct uk_lcpu *lcpu;
-	unsigned int i, n, argi = 0;
 	const int new = UK_LCPU_STATE_INIT;
+	struct uk_lcpu *lcpu;
 	int old, rc = 0, rc2;
+	unsigned int i, n;
 
 	UK_ASSERT(lcpuidx);
 	UK_ASSERT(num);
@@ -312,20 +312,10 @@ int uk_lcpu_start(const __u32 lcpuidx[], unsigned int *num,
 		return -ENODEV;
 #endif /* CONFIG_HAVE_CPU_MULTI_PHASE_STARTUP */
 
-	for (i = 0, n = *num; i < n; i++, *num = i) {
+	for (i = 0, n = *num; i < n; i++) {
 		if (uk_pcpuvar_lval(lcpuidx[i],
-				    uk_pcpuvar_cpu_id) == this_cpu_id) {
-			/* If the caller did not supply an index array, we
-			 * assume that we do not have parameters and a stack
-			 * for the executing CPU. Otherwise, i.e., if the
-			 * caller explicitly put the executing CPU in, we still
-			 * ignore it but need to skip the parameters.
-			 */
-			if (lcpuidx)
-				argi++;
-
+				    uk_pcpuvar_cpu_id) == this_cpu_id)
 			continue;
-		}
 
 		lcpu = &uk_pcpuvar_lval(lcpuidx[i], uk_lcpus);
 retry:
@@ -336,9 +326,6 @@ retry:
 			uk_pr_warn("Failed to start CPU 0x%lx: not offline\n",
 				   uk_pcpuvar_lval(lcpuidx[i],
 						   uk_pcpuvar_cpu_id));
-
-			/* Skip CPU and its arguments*/
-			argi++;
 			continue;
 		}
 @@ -344,10 +344,10 @@ retry:
 		 * separate counter to index the arguments.
 		 */
 		uk_pcpuvar_lval(lcpuidx[i], UK_LCPU_SENTRY_SYM) =
-			(entry && entry[argi]) ?
-				entry[argi] :
+			(entry && entry[i]) ?
+				entry[i] :
 				(__uptr)uk_lcpu_entry_default;
-		uk_pcpuvar_lval(lcpuidx[i], UK_LCPU_SSTACKP_SYM) = sp[argi];
+		uk_pcpuvar_lval(lcpuidx[i], UK_LCPU_SSTACKP_SYM) = sp[i];
 		uk_pcpuvar_lval(lcpuidx[i], UK_LCPU_SARG_SYM) = (__uptr)lcpu;
 
 		/* Ensure that the startup arguments have been written back
@@ -378,9 +365,6 @@ retry:
 			 */
 			break;
 		}
-
-		/* Move to next argument */
-		argi++;
 	}
 
 	*num = i;
@@ -392,10 +376,8 @@ retry:
 	 */
 	rc2 = pm_ops->post_start(lcpuidx, &i);
 	if (unlikely(rc2)) {
-		if (num) {
-			UK_ASSERT(i <= *num);
-			*num = i;
-		}
+		UK_ASSERT(i <= *num);
+		*num = i;
 
 		/* Return the first error */
 		return (rc) ? rc : rc2;
@@ -445,7 +427,7 @@ int uk_lcpu_run(const __u32 lcpuidx[], unsigned int *num,
 	UK_ASSERT(num);
 	UK_ASSERT(fn);
 
-	for (i = 0, n = *num; i < n; i++, *num = i) {
+	for (i = 0, n = *num; i < n; i++) {
 		if (uk_pcpuvar_lval(lcpuidx[i],
 				    uk_pcpuvar_cpu_id) == this_cpu_id)
 			continue;
@@ -508,7 +490,7 @@ int uk_lcpu_wait(const __u32 lcpuidx[], unsigned int *num,
 	if (timeout > 0)
 		end = ukplat_monotonic_clock() + timeout;
 
-	for (i = 0, n = *num; i < n; i++, *num = i) {
+	for (i = 0, n = *num; i < n; i++) {
 		if (uk_pcpuvar_lval(lcpuidx[i],
 				    uk_pcpuvar_cpu_id) == this_cpu_id)
 			continue;
@@ -552,7 +534,7 @@ int uk_lcpu_wakeup(const __u32 lcpuidx[], unsigned int *num)
 	UK_ASSERT(lcpuidx);
 	UK_ASSERT(num);
 
-	for (i = 0, n = *num; i < n; i++, *num = i) {
+	for (i = 0, n = *num; i < n; i++) {
 		if (uk_pcpuvar_lval(lcpuidx[i],
 				    uk_pcpuvar_cpu_id) == this_cpu_id)
 			continue;
```
</details>

drivers/firmware/ukacpi/arch/x86_64/madt.c
```diff
@@ -0,0 +1,94 @@
...
+		if (bsp_cpu_id == cpu_id) {
+			UK_ASSERT(!bsp_found);
+
+			bsp_found = 1;
+			continue;
+		}
+
+		uk_pcpuvar_lval(idx, uk_pcpuvar_cpu_id) = cpu_id;
+		uk_pcpuvar_lval(idx, uk_pcpuvar_cpu_idx) = idx;
+		idx++;
```
michpappas, line 79: We needs a bounds check against `CONFIG_MAX_LCPUS`:
```c
		/* Ignore cores that exceed max configured value */
		if (unlikely(idx == CONFIG_MAX_LCPUS)) {
			uk_pr_warn("Maximum number of cores exceeded.\n");
			return 0;
		}
```

lib/ukpal/include/uk/pal/auxsp.h
```diff
@@ -16,33 +16,21 @@
...
 extern "C" {
 #endif
 
-/**
- * Set auxiliary stack pointer.
- * Configures the stack used for exception handling.
- *
- * @param auxsp  New auxiliary stack pointer
- */
 void uk_pal_set_auxsp(__uptr auxsp);
```
michpappas, line 32: Now the `set()` operation is a bit ambiguous, and one would expect that there is also a `get()`. Shall we rename this as `uk_pal_auxsp_init()` (for initialize) or similar?

plat/native/pal/include/uk/plat/pal/start.h
```diff
@@ -0,0 +1,20 @@
...
+ * You may not use this file except in compliance with the License.
+ */
+
+#ifndef __UK_PLAT_PAL_START_H__
+#define __UK_PLAT_PAL_START_H__
+
+#include <uk/arch/types.h>
+#include <uk/plat/native/arch/start.h>
+
+#define UK_PAL_SENTRY_SYM						\
```
michpappas, line 13: Nit, newline escapes can be aligned.

lib/uklcpu/include/uk/lcpu.h
```diff
@@ -78,12 +56,18 @@ UK_CTASSERT(sizeof(struct uk_lcpu_func) == UK_LCPU_FUNC_SIZE);
...
 #define UK_LCPU_STATE_OFFSET		0x00
-#define UK_LCPU_ENTRY_OFFSET		(UK_LCPU_STATE_OFFSET + 0x08)
-#define UK_LCPU_STACKP_OFFSET		(UK_LCPU_ENTRY_OFFSET + 0x08)
-#define UK_LCPU_ERR_OFFSET		(UK_LCPU_ENTRY_OFFSET + 0x0)
+#define UK_LCPU_ERR_OFFSET		(UK_LCPU_STATE_OFFSET + 0x4)
+#if CONFIG_HAVE_SMP
+#define UK_LCPU_FN_OFFSET		(UK_LCPU_ERR_OFFSET + 0x4)
+#endif /* CONFIG_HAVE_SMP */
 
+#if CONFIG_HAVE_SMP
```
michpappas: Consecutive conditionals can be merged (this CONFIG_HAVE_SMP block with the one right above).

lib/ukpal/include/uk/pal/start.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef __UK_PAL_START_H__
+#define __UK_PAL_START_H__
+
+#include <uk/arch/types.h>
+
+/**
+ * Start elements - Platform Abstraction Layer
+ *
+ * Provides definitions to start elements that a platform may have to define
```
michpappas: I really can't make anything out of this sentence...

lib/ukpal/include/uk/pal/start.h
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * The platform must define per-CPU variables for the CPU start entry, stack
+ * and argument and export their symbols through the PAL by aliasing these
+ * macros to them.
+ *
+ * These are all of type __uptr and must be accessed through the per-CPU
+ * variables API as such.
+ */
+#ifndef UK_PAL_SENTRY_SYM
```
michpappas, line 27: Since a platform may not support SMP, maybe these definitions should be conditional to `CONFIG_HAVE_SMP`?

plat/native/arch/x86_64/include/uk/plat/native/arch/start.h
michpappas, line 1: This header will look identical in arm64, maybe we can move it outside the arch layer.
Same for `start.c`...

lib/uklcpu/include/uk/lcpu/pm.h
```diff
@@ -16,7 +16,7 @@ extern "C" {
 #endif
 
 #if CONFIG_HAVE_SMP
-typedef int (*uk_lcpu_pm_start_t)(struct uk_lcpu *lcpu);
+typedef int (*uk_lcpu_pm_start_t)(__u32 idx);
```
michpappas: With the opportunity of being updated, maybe this should be renamed to comply to current conventions, as `uk_lcpu_pm_start_t` sounds like a type, and not a function. Also semantically speaking `__sz` is a more suitable type for the index.

lib/ukpcpuvar/Makefile.uk
michpappas, line 1: We're missing `exportsyms.uk`.

michpappas (changes requested):

lib/ukpcpuvar/include/uk/pcpuvar.h
mogasergiu: Ughh... yeah, its availabiliy does ineed predate C23 though (for both GCC/clang). But it's also being used in other parts of the codebase since the project's inception as well... maybe let's defer until we delve deeper into the standard we should stick to

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
mogasergiu: No strong opinion either, I just thought that was the only remaining sensible place given it's still exception related. Maybe we can defer this discussion until we have to add stuff into the PAL like what you mentioned, as that wold likely mean also making it arch-specific.

> At least on arm64 IRQs are considered asynchronous exceptions 

Yes, It's the sending I was referring to that is synchronous :stuck_out_tongue:

lib/ukpal/include/uk/pal/auxsp.h
mogasergiu: Yes, unfortunately, it can't be `init` either as it is called whenever the scheduler switches threads as well so that the current CPUs auxsp stays updated to point to the auxsp of the current thread. Given this, I would leave as is for now.

lib/ukpal/include/uk/pal/start.h
mogasergiu: Note these are used in boot protocol/entry for the bootstrap CPU as well so it's not necessarily SMP related.

plat/native/arch/x86_64/include/uk/plat/native/arch/start.h
mogasergiu: As you also said in an offline discussion, this change probably fits better into the ARM64 PR that adapts the ARM64 native code to ukpcpu :thinking:

lib/uklcpu/include/uk/lcpu/pm.h
mogasergiu: Hmm, yes `__sz` makes sense as well. In fact `__u32` is inherently wrong because that is not the type that ukpcpuvar (the library that owns indexes now) uses (it uses `__u64`). Between the two, since there is no strong opinion, I think I will stick to `__u64` for now.

What should the name be? What about `uk_lcpu_pm_kick_t` and `uk_lcpu_pm_post_kick_t`?

lib/uklcpu/include/uk/lcpu.h
mogasergiu: So this comment, like all other comments from lcpu.h are just moved from the original lcpu.h. But it is indeed a good opportunity to improve it.

mogasergiu:
fpush:
- one sentence per line in the pcpuvar README
- unconditionally declare C/lds file in pcpuvar
- move cplusplus guards up/down to contain everything in pcpuvar headers
- add exportsysms to pcpuvar
- add uksched dep on pcpuvar
- improve docs in uklcpu
- use __bool instead of int in uklcpu
- move cplusplus guards up/down in lcpu.h header
- merged consecutive HAVE_SMP conditionals in lcpu.h
- use _func instead of _t for lcpu/pm.h
- fixed up forgotten lcpuidx references - all idxs now 64-bit
- add `CONFIG_UKPLAT_CPU_MAXCOUNT` bounds check in MADT parser
- fix UK_PAL_SARG_SYM backslash misalignment

mogasergiu:
fpush:
- use break instead of return in madt parser bounds check so that we still assert for `bsp_found`

michpappas (approved):
Really cool work! 🚀 

lib/uklcpu/include/uk/lcpu/pm.h
```diff
@@ -0,0 +1,155 @@
...
+ *   Can be NULL, otherwise if lcpuidx is NULL, must contain
+ *   uk_lcpu_count() - 1 function pointers. Provided functions must not
+ *   return. If the parameter or individual function pointers are NULL the
+ *   respective logical CPUs enter a low-power wait state after startup
+ * @param flags (architecture-dependent) flags that specify how to start the
+ *   CPUs (see UK_LCPU_SFLG_* flags if available)
+ *
+ * @return 0 on success, an errno-type error value otherwise
+ */
+int uk_lcpu_start(const __u64 lcpuidx[],
```
andreittr, line 118: The docs of `uk_lcpu_start` are still written as if `lcpuidx` can be NULL; these should be updated to reflect the implementation.

mogasergiu:
fpush:
- fix checkpatch warning for Local SAPIC (we will likely never use that anyway)

mogasergiu:
fpush:
- fix Python script linter errors (uk-libinfo.py errors are unrelated to this PR)

mogasergiu:
fpus:
- Introduced fallback `__uk_pcpuvar_arch_current_ptr_get` implementation for when `CR4.FSGSBASE` is not available

plat/native/arch/x86_64/sysctx_auxsp.c
```diff
@@ -0,0 +1,142 @@
...
+__isr void uk_plat_native_sysctx_load(struct uk_plat_native_sysctx *sysctx)
+{
+	UK_ASSERT(sysctx);
+
+	wrgsbasefn(sysctx->gsbase);
+	uk_plat_native_wrfsbasefn(sysctx->fsbase);
+}
+
+#if CONFIG_LIBUKPLAT_NATIVE_AUXSP
+__uk_pcpuvar __align(UK_ARCH_CACHE_LINE_SIZE) __uptr uk_plat_native_auxsp;
```
andreittr: (I'm going to assume you placed this alignment to prevent false sharing)
I believe we don't need to align the auxsp separately from the rest of the pcpu area, as that is already cacheline-aligned, and thus isolated in the cache from competing cores. Having extra alignment only adds padding without any benefit.

mogasergiu: Yes, you are right! Good catch! Thanks!

andreittr (approved):
Looks great, thanks!

andreittr:

PR #1775, opened 2026-02-22
plat/kvm/x86/tscclock: Add KVM paravirtual wall clock support

### LINKED ISSUE #1759: System time is 1999
### Describe the bug

`time.time()` returns 946598400 (December 31st, 1999) instead of current time.

Output:
```
time.time() = 946598400.1024494
datetime    = 1999-12-31 00:00:00.102449
monotonic   = 0.203227443
````

### Steps to reproduce

I've created a small example in https://github.com/dutt/unikraft_debugging/tree/main/system_time

Dockerfile:
```
FROM python:3.12-slim AS build

# Run ldconfig to update library cache
RUN /sbin/ldconfig /usr/local/lib

# Final minimal image
FROM scratch

# Copy Python runtime and required libraries
COPY --from=build /usr/local/lib /usr/local/lib
COPY --from=build /usr/local/bin/python3.12 /usr/bin/python3
COPY --from=build /lib/x86_64-linux-gnu/libc.so.6 /lib/x86_64-linux-gnu/libc.so.6
COPY --from=build /lib/x86_64-linux-gnu/libm.so.6 /lib/x86_64-linux-gnu/libm.so.6
COPY --from=build /lib/x86_64-linux-gnu/libz.so.1 /lib/x86_64-linux-gnu/libz.so.1
COPY --from=build /lib64/ld-linux-x86-64.so.2 /lib64/ld-linux-x86-64.so.2
COPY --from=build /etc/ld.so.cache /etc/ld.so.cache
COPY ./main.py .
```

Kraftfile:
```
spec: v0.6

runtime: unikraft.org/python:3.14

rootfs: ./Dockerfile

cmd: ["/usr/bin/python3", "main.py"]

targets:
  - name: default
    platform: fc
    architecture: x86_64
```

main.py:
```
import time
from datetime import datetime

ts = time.time()
print(f"time.time() = {ts}")
print(f"datetime    = {datetime.fromtimestamp(ts)}")
print(f"monotonic   = {time.monotonic()}")
```

### Expected behavior

Output:
````
time.time() = 1736280000
datetime    = 2025-01-07 18:00:00
monotonic   = 0.001234
````

x86_64

Firecracker

### Relevant log output

```shell

```

YashSuthar983:
fix: #1759

Firecracker does not emulate the legacy CMOS , causing I/O port reads to fail.

Switched the initial boot time fetch to try the new KVM PV clock logic first.fall back to the old, legacy CMOS only if the PV clock isn't available.

### DIFF
--- a/plat/kvm/x86/tscclock.c
+++ b/plat/kvm/x86/tscclock.c
@@ -94,6 +94,54 @@
 /* RTC wall time offset at monotonic time base. */
 static __u64 rtc_epochoffset;
 
+/* KVM paravirtual wall clock MSRs */
+#define MSR_KVM_WALL_CLOCK 0x11
+#define MSR_KVM_WALL_CLOCK_NEW 0x4b564d00
+#define KVM_FEATURE_CLOCKSOURCE (1 << 0)
+#define KVM_FEATURE_CLOCKSOURCE2 (1 << 3)
+
+struct pvclock_wall_clock {
+	volatile __u32 version;
+	volatile __u32 sec;
+	volatile __u32 nsec;
+} __attribute__((__packed__));
+
+static struct pvclock_wall_clock wall_clock_data __align(4);
+
+static __u64 pvclock_read_wall_clock(void)
+{
+	__u32 eax, ebx, ecx, edx;
+	__u32 features, msr, version;
+	__u64 wall_nsec;
+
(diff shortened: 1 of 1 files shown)


YashSuthar983:
@razvand @StefanJum could you please review this whenever you have time?

PR #1779, opened 2026-03-02
arch: Introduce new layout

michpappas:
### Description of Changes

This patchset introduces the new `arch/` conventions and adapts the codebase accordingly.

`arch/` is now organized under the following scheme:
    
### uk/arch.h ###
    
**Scope:** 

Generic arch helpers that can be used by arch-agnostic code. This includes non-privileged, non-platform-specific declarations for barriers, MMIO accessors, etc. Defines the uk_arch_ API. Must implemented by all platforms.
    
**Namespace:** `uk_arch_`
    
### uk/arch/arch.h ###
    
**Scope:** Arch-specific implementation of the uk_arch_ API.
    
**Namespace:** `uk_arch_`

### uk/arch/<arch>.h ###
    
**Scope:**

Core arch definitions. These are strictly limited to definitions deriving from the architecture's reference documentation (Arm ARM, Intel 64 SDM).
    
**Namespace:** `UK_ARCH_<arch>_`
    
### uk/arch/util.h ###
    
**Scope:**

Non-privileged, non-platform-specific helpers that are arch-specific, and thus cannot be part of the generic uk_arch API.
    
**Namespace:** `UK_ARCH_<arch>_`

### Related Work

N/A

### PR Checklist


arch/arm64/include/uk/arch/util.h
```diff
@@ -183,6 +183,26 @@ void uk_arch_arm64_spinwait(void)
 	/* Intelligent busy wait not supported on arm64. */
 }
 
+static inline
+void __noreturn uk_arch_arm64_jump_to(__u64 sp, __u64 entry)
+{
+	__asm__ __volatile__ ("mov	x29, xzr\n" /* reset frame pointer */
```
mogasergiu: Nit, you moved the first inline instruction on the asme line as __asm__ __volatile__, unlike in the original snipper. Also, the call to this in `lcpu_arch_jump_to` still calls `uk_arch_arm64_lcpu_jump_to` (notice the `_lcpu_` in the namespace) and might need a cast for the args.

arch/arm64/include/uk/arch/arm64.h
```diff
@@ -1,37 +1,19 @@
...
- * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
- * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
- * POSSIBILITY OF SUCH DAMAGE.
+/* Copyright (c) 2022, OpenSynergy GmbH. All rights reserved.
+ * Copyright (c) 2025, Unikraft GmbH and The Unikraft Authors.
+ * Licensed under the BSD-3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
  */
 
 #include <uk/asm.h>
```
mogasergiu, line 8: Nit, while not significant, would probably be better and more consistent to have this include inside the guard.

arch/arm64/include/uk/arch/arm64.h
```diff
@@ -15,411 +15,452 @@ extern "C" {
...
+#define UK_ARCH_ARM64_PTE_ATTR_nG			(1 << 11)
+#define UK_ARCH_ARM64_PTE_ATTR_AF			(1 << 10)
+#define UK_ARCH_ARM64_PTE_ATTR_SH(x)			((x) << 8)
+#define UK_ARCH_ARM64_PTE_ATTR_SH_MASK			UK_ARCH_ARM64_PTE_ATTR_SH(3)
+#define UK_ARCH_ARM64_PTE_ATTR_SH_NS			0 /* Non-shareable */
+#define UK_ARCH_ARM64_PTE_ATTR_SH_OS			2 /* Outer-shareable */
+#define UK_ARCH_ARM64_PTE_ATTR_SH_IS			3 /* Inner-shareable */
+#define UK_ARCH_ARM64_PTE_ATTR_AP_RW_BIT		(1 << 7)
+#define UK_ARCH_ARM64_PTE_ATTR_AP(x)			((x) << 6)
+#define UK_ARCH_ARM64_PTE_ATTR_AP_MASK			UK_ARCH_ARM64_ATTR_AP(3)
```
mogasergiu: Typo? missing `PTE_`

arch/arm64/include/uk/arch/util.h
```diff
@@ -118,6 +118,71 @@ void uk_arch_arm64_ioreg_write64(const volatile __u64 *address,
...
 }
 
+/*
+ * Instruction Synchronization Barrier flushes the pipeline in the
+ * processor, so that all instructions following the ISB are fetched
+ * from cache or memory, after the instruction has been completed.
+ */
+static inline void uk_arch_arm64_isb(void)
+{
+	__asm__ __volatile("isb" ::: "memory");
```
mogasergiu: Noticed this cus GH gave it a different color, i think this should've been `__volatile__`

arch/arm64/include/uk/arch/arch.h
```diff
@@ -0,0 +1,57 @@
...
+{
+	uk_arch_arm64_dsb(ld);
+}
+
+static inline void uk_arch_wmb(void)
+{
+	uk_arch_arm64_dsb(st);
+}
+
+static inline unsigned long uk_arch_read_sp(void)
```
mogasergiu: I noticed this is returning `unsigned long`, while the implementation returns `__u64` (on x86_64 also unsigned long). Should probably stick to one, maybe `__u64` since it is already in `uk_arch_jump_to`. Otherwise I think all `__u64`s should be replaced with `unsigned long`

plat/common/w_xor_x.c
```diff
@@ -1,44 +1,21 @@
...
+
+#include <uk/arch/util.h>
 #include <uk/config.h>
 #include <uk/paging.h>
 #include <uk/plat/memory.h>
 #include <uk/print.h>
 
 #ifdef CONFIG_ARCH_ARM_64
 #include <arm/arm64/cpu.h>
+#include <uk/arch/util.h>
```
mogasergiu, line 17: oof, sed did this to me several times in my lcpu PR as well: duplicate inclusion.

lib/ukvmem/arch/arm/pagefault64.c
```diff
@@ -25,16 +25,16 @@ static int vmem_arch_pagefault(void *data)
 	struct uk_vas *vas;
 	int rc;
 
-	if (ctx->esr & ESR_ISS_ABRT_WnR_SHIFT)
+	if (ctx->esr & UK_ARCH_ARM64_ESR_ISS_ABRT_WnR_SHIFT)
```
mogasergiu: oh man... this is a bug (introduced by me), it should've been `ESR_ISS_ABRT_WnR_MASK`, right? 💀

arch/arm64/include/uk/arch/arch.h
michpappas: Good point. I'll replace with `__u64` as at the low level it's best to be explicit.

arch/arm64/include/uk/arch/arm64.h
michpappas: Consistent with? Also generally speaking headers should be self-guarded to avoid exactly that, and this header is guarded indeed.

arch/arm64/include/uk/arch/util.h
michpappas: > Nit, you moved the first inline instruction on the asme line as asm volatile, unlike in the original snipper.

Yeah it was checkpatch that complained about the open brace, but after your comment I noticed that we start multi-line inline asm statements on a new line. Will revert.

> Also, the call to this in lcpu_arch_jump_to still calls uk_arch_arm64_lcpu_jump_to (notice the _lcpu_ in the namespace) and might need a cast for the args.

Good catch, will add a cast.

arch/arm64/include/uk/arch/arch.h
```diff
@@ -0,0 +1,57 @@
...
+}
+
+static inline void uk_arch_rmb(void)
+{
+	uk_arch_arm64_dsb(ld);
+}
+
+static inline void uk_arch_wmb(void)
+{
+	uk_arch_arm64_dsb(st);
```
andreittr: AFAICT arm64 provides namespaced equivalents for mb/rmb/wmb (all implemented as _dsb, same as here); I think having these generic arch mb/rmb/wmb implementations delegate to their arm64 equivalents, rather than duplicating the lower-level implementations, would help keep a single source of truth for how barriers are implemented on arm.

plat/native/arch/x86_64/except.c
```diff
@@ -677,24 +659,28 @@ static inline int x2apic_enable(void)
...
-	uk_arch_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
-	if (!(ecx & UK_ARCH_CPUID1_ECX_X2APIC))
+	uk_arch_x86_64_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	if (!(ecx & UK_ARCH_X86_64_CPUID1_ECX_X2APIC))
 		return -ENOTSUP;
 
 	/* Check if APIC is active */
-	uk_arch_rdmsr(UK_ARCH_APIC_MSR_BASE, &eax, &edx);
-	if (!(eax & UK_ARCH_APIC_BASE_EN))
+	uk_arch_x86_64_x86_64_uk_arch_rdmsr(UK_ARCH_X86_64_APIC_MSR_BASE,
```
andreittr: I believe this is a case of pattern replacement going awry, most likely `uk_arch_x86_64_rdmsr` was meant 😄
There are several cases within this file with the same pattern, I won't add a comment for each, but line numbers are: 674, 678, 682, 706.

plat/native/arch/x86_64/lcpu_pm.c
```diff
@@ -27,23 +28,27 @@ static inline void x2apic_send_sipi(__vaddr_t addr, int dest)
...
-	eax = UK_ARCH_APIC_ICR_TRIGGER_LEVEL | UK_ARCH_APIC_ICR_LEVEL_ASSERT |
-	      UK_ARCH_APIC_ICR_DESTMODE_PHYSICAL | UK_ARCH_APIC_ICR_DMODE_SUP |
+	eax = UK_ARCH_X86_64_APIC_ICR_TRIGGER_LEVEL |
+	      UK_ARCH_X86_64_APIC_ICR_LEVEL_ASSERT |
+	      UK_ARCH_X86_64_APIC_ICR_DESTMODE_PHYSICAL |
+	      UK_ARCH_X86_64_APIC_ICR_DMODE_SUP |
 	      (addr >> __PAGE_SHIFT);
 
-	uk_arch_wrmsr(UK_ARCH_APIC_MSR_ICR, eax, dest);
+	uk_arch_x86_64_x86_64_uk_arch_wrmsr(UK_ARCH_X86_64_APIC_MSR_ICR, eax, dest);
```
andreittr: Pattern replacement hiccup; also on line 51 in this file.

plat/kvm/x86/efi_post.c
```diff
@@ -51,7 +53,7 @@ static inline void lapic_timer_disable(void)
 	__u32 eax, edx;
 
 	/* Check if APIC is active */
-	uk_arch_rdmsr(LAPIC_MSR_BASE, &eax, &edx);
+	uk_arch_x86_64_x86_64_uk_arch_rdmsr(LAPIC_MSR_BASE, &eax, &edx);
```
andreittr: Pattern replacement hiccup.

drivers/ukintctlr/xpic/ukintctlr.c
```diff
@@ -2,39 +2,47 @@
...
-	uk_arch_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
-	if (!(ecx & UK_ARCH_CPUID1_ECX_X2APIC))
+	uk_arch_x86_64_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	if (!(ecx & UK_ARCH_X86_64_CPUID1_ECX_X2APIC))
 		return -ENOTSUP;
 
 	/* Check if APIC is active */
-	uk_arch_rdmsr(UK_ARCH_APIC_MSR_BASE, &eax, &edx);
-	if (!(eax & UK_ARCH_APIC_BASE_EN))
+	uk_arch_x86_64_x86_64_uk_arch_rdmsr(UK_ARCH_X86_64_APIC_MSR_BASE,
```
andreittr: Pattern replace hiccup; repeats on lines 36, 40, 44, 59.

lib/posix-process/arch/x86_64/execve.c
```diff
@@ -4,7 +4,7 @@
  * You may not use this file except in compliance with the License.
  */
 
-#include <uk/arch.h>
+#include <uk/arch/x86_64.h>
```
andreittr, line 7: (Related to commit message itself): I believe the title was intended to begin with `lib:` to reflect the scope of the commit.

andreittr (changes requested):
Looks good overall, only two minor issues.

plat/native/arch/x86_64/except.c
michpappas: Yikes! How did it even compile?!

arch/arm64/include/uk/arch/arch.h
michpappas: Of course 👍🏼

michpappas:
Addressed @andreittr's review comments

andreittr (approved):
Thanks for the changes, looks great!

arch/arm64/include/uk/arch/arm64.h
mogasergiu: Yes it is self guarded. I meant consistent with most of our other headers (if not all, but I can't see one that includes a header outside the guards).

arch/arm/arm64/include/uk/asm/lcpu.h
```diff
@@ -35,8 +35,6 @@
 #include <uk/config.h>
 #include <uk/essentials.h>
 
-#define CACHE_LINE_SIZE		64
```
mogasergiu, line 38: Typo in first commit message:
s/Must implemented by all platforms./Must **be** implemented by all platforms.

arch/arm64/include/uk/arch/arm64.h
```diff
@@ -15,411 +15,452 @@ extern "C" {
...
+#define UK_ARCH_ARM64_SCTLR_EL1_SED_BIT			(_AC(1, UL) << 8)
+#define UK_ARCH_ARM64_SCTLR_EL1_UMA_BIT			(_AC(1, UL) << 9)
+#define UK_ARCH_ARM64_SCTLR_EL1_EnRCTX_BIT		(_AC(1, UL) << 10)
+#define UK_ARCH_ARM64_SCTLR_EL1_EOS_BIT			(_AC(1, UL) << 11)
+#define UK_ARCH_ARM64_SCTLR_EL1_I_BIT			(_AC(1, UL) << 12)
+#define UK_ARCH_ARM64_SCTLR_EL1_EnDB_BIT		(_AC(1, UL) << 13)
+#define UK_ARCH_ARM64_SCTLR_EL1_DZE_BIT			(_AC(1, UL) << 14)
+#define UK_ARCH_ARM64_SCTLR_EL1_UCT_BIT			(_AC(1, UL) << 15)
+#define UK_ARCH_ARM64_SCTLR_EL1_nTWI_BIT		(_AC(1, UL) << 16)
+#define UK_ARCH_ARM64_SCTLR_EL1_RES0_27_BIT		(_AC(1, UL) << 17)
```
mogasergiu: I just noticed this, I think this should've been `17` not `27`, right? I guess this typo has always existed.

drivers/ukintctlr/gic/gic-v3.c
```diff
@@ -154,12 +156,12 @@ static void wait_for_rwp(__u64 offset)
 static __u32 get_cpu_affinity(void)
 {
 	__u64 aff;
-	__u64 mpidr = SYSREG_READ64(MPIDR_EL1);
+	__u64 mpidr = SYSREG_READ64(UK_ARCH_ARM64_MPIDR_EL1);
```
mogasergiu: Oops, oversedded here, `MPIDR_EL` was fine I think. But also I think this should've been `UK_ARCH_ARM64_SYSREG_READ` right?

plat/common/arm/lcpu.c
```diff
@@ -94,17 +94,7 @@ __lcpuid lcpu_arch_id(void)
...
-		"mov	x30, xzr\n" /* reset link register */
-		"mov	sp, %0\n" /* set the sp  */
-		"br	%1\n"     /* branch to the entry function */
-		:
-		: "r"(sp), "r"(entry)
-		: /* sp not needed */);
-
-	/* just make the compiler happy about returning function */
-	__builtin_unreachable();
+	uk_arch_arm64_lcpu_jump_to(sp, entry);
```
mogasergiu: typo, remove the "_lcpu"

arch/arm64/include/uk/arch/arch.h
```diff
@@ -0,0 +1,57 @@
...
+{
+	uk_arch_arm64_rmb();
+}
+
+static inline void uk_arch_wmb(void)
+{
+	uk_arch_arm64_wmb();
+}
+
+static inline unsigned long uk_arch_read_sp(void)
```
mogasergiu: I think you meant `__u64`?

lib/ukring/include/uk/ring.h
```diff
@@ -337,7 +337,7 @@ uk_ring_peek_clear_sc(struct uk_ring *br)
 	#error "unsupported: atomic_thread_fence_acq()"
 	/* TODO atomic_thread_fence_acq(); */
 #elif defined(CONFIG_ARCH_ARM64)
-	dmb();
+	uk_arch_arm64_dmb();
```
mogasergiu: This is documented to require some argument. But what is worse, and I am just noticing now, there is a typo: `s/CONFIG_ARCH_ARM64/CONFIG_ARCH_ARM_64/`

arch/x86_64/include/uk/arch/x86_64.h
```diff
@@ -8,272 +8,272 @@
  * You may not use this file except in compliance with the License.
  */
 
-#ifndef __UK_ARCH_H__
-#define __UK_ARCH_H__
+#ifndef __UK_ARCH_X86_64_H__
```
mogasergiu, line 11: Typo in `arch/x86_64: Namespace x86_64.h` commit message: `s/Checkptch/Checkpatch`

arch/x86_64/include/uk/arch/x86_64.h
```diff
@@ -756,9 +756,9 @@ struct uk_arch_tss64 {
 	/** I/O map base address */
 	__u16 iomap_base;
 } __packed;
-#endif /* !__ASSEMBLY__ */
 
 #ifdef __cplusplus
 }
 #endif
-#endif /* __UK_ARCH_H__ */
+#endif /* !__ASSEMBLY__ */
```
mogasergiu: I think this needs to be inslude the cplusplus guard given it is also opened inside id.

arch/x86_64/include/uk/arch/util.h
```diff
@@ -67,7 +67,7 @@ static inline void uk_arch_cpuid(__u32 fn, __u32 subfn,
  *
  * @return Current RFLAGS value
  */
-static inline unsigned long uk_arch_rflags_get(void)
+static inline unsigned long uk_arch_x86_64_rflags_get(void)
```
mogasergiu: Ditto about the `__u64`, i see same goes for the read_sp/rdcr2/wrcr3 in this file as well

arch/x86_64/include/uk/arch/arch.h
```diff
@@ -0,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+/* Copyright (c) 2018, Arm Ltd.
```
mogasergiu: Thank you Arm for you x86_64 contribution? 😅

plat/native/arch/x86_64/except.c
```diff
@@ -677,24 +659,28 @@ static inline int x2apic_enable(void)
...
-	uk_arch_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
-	if (!(ecx & UK_ARCH_CPUID1_ECX_X2APIC))
+	uk_arch_x86_64_cpuid(1, 0, &eax, &ebx, &ecx, &edx);
+	if (!(ecx & UK_ARCH_X86_64_CPUID1_ECX_X2APIC))
 		return -ENOTSUP;
 
 	/* Check if APIC is active */
-	uk_arch_rdmsr(UK_ARCH_APIC_MSR_BASE, &eax, &edx);
-	if (!(eax & UK_ARCH_APIC_BASE_EN))
+	uk_arch_x86_64_rdmsr(UK_ARCH_X86_64_APIC_MSR_BASE,
```
mogasergiu, line 667: I think all MSR ops in `x2apic_enable`  have their last two args misaligned.

plat/kvm/x86/setup.c
```diff
@@ -61,27 +63,33 @@ static inline void _init_syscall(void)
...
 
 	/* Enable and program syscall/sysret */
-	uk_arch_wrmsrl(UK_ARCH_MSR_EFER,
-		       uk_arch_rdmsrl(UK_ARCH_MSR_EFER) |
-		       UK_ARCH_EFER_LMA | UK_ARCH_EFER_LME | UK_ARCH_EFER_SCE);
-	uk_arch_wrmsrl(UK_ARCH_MSR_STAR, (0x08ULL << 48) | (0x08ULL << 32));
-	uk_arch_wrmsrl(UK_ARCH_MSR_LSTAR, (__uptr)_ukplat_syscall);
+	uk_arch_x86_64_wrmsrl(UK_ARCH_X86_64_MSR_EFER,
+			      uk_arch_x86_64_rdmsrl(UK_ARCH_X86_64_MSR_EFER) |
+			      UK_ARCH_X86_64_EFER_LMA |
```
mogasergiu, line 78: Wait what I just noticed... why are we writing a RO bit? cc @andreittr

arch/arm64/include/uk/arch/arm64.h
michpappas: Yeah it's harmless as it's one of the two reserved bits, but will fix anyway. Thanks for catching.

drivers/ukintctlr/gic/gic-v3.c
michpappas: Seems that I updated these the other way around during some conflict resolution.

plat/kvm/x86/setup.c
andreittr: :shrug: guess it's being silently ignored by the CPU. Either way, changing wrmsr semantics is IMO outside the scope of this work (which is just namespacing/renaming).

michpappas:
Addressed review comments

michpappas:
Rebased and improved formatting of 1st commit (no content changes).

mogasergiu (approved):

PR #1781, opened 2026-03-03
include/uk: fix C++ compatibility in kernel headers

harimishal1:
### Description of Changes

Unikraft's internal headers are exposed to C++ translation units when libraries such as lib-click compile with -std=c++11.

- uk/list.h: Replaced typeof with __typeof__ in all list-iteration macros. typeof is a GNU C extension not available in strict C++ mode; __typeof__ is the portable GCC built-in that works in both C and C++.

- uk/arch/types.h: Define __NULL as nullptr in C++ context. Assigning ((void *)0) to a typed pointer is valid in C but illegal in C++11.

- plat/native/pal/include/uk/plat/pal/ectx.h: Added extern "C" guards. Without them, the five __isr static inline functions acquire C++ linkage, conflicting with their C-linkage re-declaration in lib/ukpal/include/uk/pal/ectx.h.

Tested by building catalog-core/click and catalog-core/cpp-hello which both previously failed to compile.

### Related Work

N/A

### PR Checklist

 - [X] Updated relevant documentation. (I don't think a change to documentation is necessary.)

### DIFF
--- a/include/uk/arch/types.h
+++ b/include/uk/arch/types.h
@@ -248,7 +248,11 @@ typedef __u64 __paddr_t;
 #endif
 
 #ifndef __NULL
+#ifdef __cplusplus
+#define __NULL nullptr
+#else
 #define __NULL ((void *) 0)
+#endif 
 #endif
 
 typedef char *caddr_t; /* core address */
--- a/plat/native/pal/include/uk/plat/pal/ectx.h
+++ b/plat/native/pal/include/uk/plat/pal/ectx.h
@@ -23,6 +23,10 @@
 	UK_PLAT_NATIVE_ECTX_ALIGN
 
 #if !__ASSEMBLY__
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
 struct uk_pal_ectx;
 
 __isr static inline void uk_pal_ectx_sanitize(struct uk_pal_ectx *state)
@@ -50,5 +54,8 @@ __isr static inline void uk_pal_ectx_assert_equal(struct uk_pal_ectx *state)
 	uk_plat_native_ectx_assert_equal((struct uk_plat_native_ectx *)state);
 }
 
+#ifdef __cplusplus
+} /* extern "C" */
+#endif
 #endif /* !__ASSEMBLY__ */
 #endif /* __UK_PLAT_PAL_ECTX_H__ */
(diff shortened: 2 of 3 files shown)


include/uk/arch/types.h
```diff
@@ -248,8 +248,12 @@ typedef __u64 __paddr_t;
 #endif
 
 #ifndef __NULL
+#ifdef __cplusplus
+#define __NULL nullptr
+#else
 #define __NULL ((void *) 0)
-#endif
+#endif /* __cplusplus */
```
mogasergiu: Similar to other cpp guards and unlike our typical header guards, no need for these `/* __cplusplus */` comments.

plat/native/pal/include/uk/plat/pal/ectx.h
```diff
@@ -50,5 +54,8 @@ __isr static inline void uk_pal_ectx_assert_equal(struct uk_pal_ectx *state)
 	uk_plat_native_ectx_assert_equal((struct uk_plat_native_ectx *)state);
 }
 
+#ifdef __cplusplus
+} /* extern "C" */
```
mogasergiu, line 58: Ditto extern C

mogasergiu (changes requested):
Lgtm. Just a nitpick and could you please split this commit into 3 separate ones based on the subsystem they touch?

plat/native/pal/include/uk/plat/pal/ectx.h
harimishal1: Split the original commit into three subsystem-specific commits and addressed the review comments.

include/uk/arch/types.h
harimishal1: I have split the original commit into three subsystem-specific commits and addressed the review comments.

plat/native/pal/include/uk/plat/pal/ectx.h
mogasergiu, line 58: Forgot to remove the added comments in this file

mogasergiu (changes requested):

include/uk/arch/types.h
```diff
@@ -248,7 +248,11 @@ typedef __u64 __paddr_t;
 #endif
 
 #ifndef __NULL
+#ifdef __cplusplus
+#define __NULL nullptr
+#else
 #define __NULL ((void *) 0)
+#endif
```
mogasergiu, line 255: Trailing white space?

PR #1794, opened 2026-03-17
lib/ukprint: Add interrupt-safe printing

mogasergiu:
### Description of Changes

In some execution contexts — interrupt handlers, exception paths, early boot before the scheduler is live — the normal console output path is unsafe. It may perform DMA, take locks without IRQ masking, or call into code that assumes a cooperative scheduling environment. This series closes that gap by introducing a dedicated emergency output path that runs end-to-end without any of those constraints.

At the driver level, a new emerg_out operation is added to uk_console_ops. Drivers that can produce output using only simple register writes (port I/O, MMIO) implement this op and are automatically promoted to emergency STDOUT. All existing UART drivers — ns16550, pl011, and the Xen emergency console — are updated accordingly, with their output functions annotated __isr throughout.

At the ukconsole level, the emerg_out op is surfaced through a matching set of uk_console_emerg_out[_direct][_all] functions and a new UK_CONSOLE_FLAG_EMERG_STDOUT flag, alongside blocking uk_console_[in|out]_direct_all variants for callers that cannot tolerate short transfers.

At the ukprint level, the entire formatting stack gains __isr twins compiled via the |isr translation unit mechanism, routing output through uk_console_emerg_out rather than uk_console_out. A new CONFIG_LIBUKPRINT_NOCARRIAGERETURN option suppresses the \r\n substitution for transports that do not need it, and a pre-existing null pointer arithmetic bug in the newline scan loop is fixed along the way. The result is a full uk_pr_{debug,info,warn,err,crit}[_once]_isr macro family usable from any context.

### Related Work

N/A

### PR Checklist


lib/ukconsole/include/uk/console.h
```diff
@@ -72,22 +72,26 @@ __ssz uk_console_out(const char *buf, __sz len);
...
-  * @param buf
-  *   Source of bytes to write
-  * @param len
-  *   Number of bytes to write starting at buf
-  * @return
-  *   - (>=0): Number of bytes written
-  *   - (<0): Error
-  */
+/**
+ * Write bytes to the given console device.
```
michpappas, line 76: Would be worth to elaborate a bit on the behavior of `uk_console_[in,out]_direct()`, so that the difference between the `_all()` variants is clear.

Also in the commit message it would be useful to state why the change is required.

lib/ukconsole/include/uk/console.h
```diff
@@ -72,6 +72,24 @@ __ssz uk_console_out(const char *buf, __sz len);
  */
 __ssz uk_console_in(char *buf, __sz len);
 
+/**
+ * Write bytes to the STDOUT default console(s) in emergency mode.
+ *
+ * Write bytes to all devices that have the UK_CONSOLE_FLAG_EMERG_STDOUT
+ * flag set ("emergency STDOUT devices"). Output is best-effort: the
```
michpappas, line 79: > Output is best-effort: the
> * function iterates all eligible devices regardless of individual
> * device errors.

The condition also applies to `uk_console_in()` and `uk_console_out()`. As is, the documentation raises questions on the behavior of these functions to someone reading the API, so maybe it would be worth updating the documentation of these functions to include the same statement.

More importantly, the `emerg_out` op is optional and the calling operations silently print nothing if not implemented. Should it be mandatory? Should we have a silent failure or return an error? If we keep the current behavior let's at least document it both in this header and also in any calling functions that use `emerg_out`.

lib/ukconsole/include/uk/console.h
```diff
@@ -72,6 +72,24 @@ __ssz uk_console_out(const char *buf, __sz len);
...
+ *
+ * @param buf
+ *   Source of bytes to write
+ * @param len
+ *   Number of bytes to write starting at buf
+ * @return
+ *   - (>=0): Number of bytes written
+ *   - (<0): Error
+ */
+__ssz uk_console_emerg_out(const char *buf, __sz len);
```
michpappas, line 91: Shouldn't both emerg functions as well as the driver op be marked as `__isr`?

lib/ukprint/include/uk/print.h
```diff
@@ -143,6 +191,42 @@ static inline void uk_printk_once(int flags, const char *fmt, ...) __printf(2, 3
...
+				 const char *fmt __unused, ...)
+{}
+
+static inline void uk_vprintk_once_isr(int flags __unused, const char *fmt __unused,
+				       va_list ap __unused)
+{}
+
+static inline void uk_printk_once_isr(int flags,
+				      const char *fmt, ...) __printf(2, 3);
+static inline void uk_printk_once_isr(int flags __unused,
```
michpappas, line 227: I think the function body`{}` is missing here.

lib/ukconsole/include/uk/console.h
```diff
@@ -135,6 +167,23 @@ int uk_console_out_direct_all(struct uk_console *dev,
  */
 int uk_console_in_direct_all(struct uk_console *dev, char *buf, __sz len);
 
+/**
+ * Tries to write all bytes to the given console device in emergency mode and
+ * blocks until it achieves to do so.
+ *
+ * @param dev
+ *   Console device to read from
```
michpappas, line 175: "to write to"

michpappas (changes requested):
This is a much needed set of changes, and overall it looks great. Only a few minor comments from me.

lib/ukconsole/include/uk/console.h
mogasergiu: I don't think it should be mandatory, I intended for it to be the exact same as in/out. I think I could just copy-paste the message to in/out? Would that be enough?

michpappas: Sure, IMO either approach could be justified as long as it's clearly documented in the API.

mogasergiu:
Force pushed rebase

mogasergiu:
Pushed updates according to comments

michpappas (approved):

michpappas:

PR #1795, opened 2026-03-17
lib/posix-tty: Auto-create console devices devfs entries

mogasergiu:
### Description of Changes

This series makes the relationship between console devices and their userspace-visible tty representations explicit and configurable, replacing the previous model where posix-tty unconditionally used a single statically allocated serial file backed by whichever console happened to be STDOUT.

A new cclass field in struct uk_console classifies each registered device as a UART, hypervisor console, or framebuffer device. The classification drives both runtime selection and devfs naming: at rootfs init time, every registered console with an assigned class automatically gets a devfs node under /dev using the following naming convention: hvc<n>, ttyS<n>, tty<n>. A new posix_tty.console kernel argument lets the application select which device backs stdin/stdout using the same naming scheme, with the default being the first registered device as before.

The per-console file representation is now dynamically allocated (struct uk_console_file) rather than a single static uk_file_serial, allowing each device to have an independent file with its own refcount and state. The selected tty console's file is allocated during init_tty_cons and reused by init_posix_tty_devfs_consnodes to avoid a duplicate allocation for the common case. uk_console_create_devnode() is exposed as a public API for drivers that need to create devfs nodes on demand rather than at bulk init time.

Kconfig options for suppressing carriage-return insertion on output and echo-back on input are added alongside this, since both are unnecessary when the console transport is not a raw terminal. A fix to posix-vfs is included to correctly short-circuit symlink resolution when the target is a non-filesystem leaf node, which is a prerequisite for the devfs symlink paths used by the new console nodes.

Depends-on: #1794

### Related Work

N/A

### PR Checklist


lib/ukconsole/include/uk/console/driver.h
```diff
@@ -16,6 +16,14 @@
 extern "C" {
 #endif
 
+enum uk_console_class {
+	UK_CONSOLE_CLASS_NONE,
```
andreittr, line 20: Perhaps a better name would be "console device class" or "console device type" (`uk_console_devclass`/`uk_console_devtype` & similar reflected across names) to better reflect what it tells us about the console.

lib/ukconsole/include/uk/console/driver.h
```diff
@@ -16,6 +16,14 @@
...
+	UK_CONSOLE_CLASS_UART,
+	UK_CONSOLE_CLASS_HVC,
+	UK_CONSOLE_CLASS_FB,
```
andreittr, line 23: The explanations for classes in the commit message are welcome, and would be nice to add to the header too, e.g. as inline comments.

Also: please document what the `NONE` class represents.

lib/posix-tty/serial.c
```diff
@@ -104,25 +122,22 @@ static ssize_t serial_read(const struct uk_file *f,
...
-		if (!bytes_read)
-			break;
-		if (unlikely(bytes_read < 0))
-			return bytes_read;
+		rc = uk_console_in_direct(tty_cons, buf, len);
+		if (unlikely(rc < 0))
+			return (ssize_t)rc;
 
-		total += bytes_read;
+		total += len;
```
andreittr, line 133: This change assumes `uk_console_in_direct` will always write out `len` bytes into `buf` unless it errors; the ukconsole API however (even post #1794 ) [explicitly states](https://github.com/mogasergiu/unikraft/blob/smoga/%231794/ukprint-isr/lib/ukconsole/include/uk/console.h#L118) that the number of bytes read may be less than `len`.

Did you mean to use [`uk_console_in_direct_all`](https://github.com/mogasergiu/unikraft/blob/smoga/%231794/ukprint-isr/lib/ukconsole/include/uk/console.h#L168)? The type of `rc` suggests so.
Though even then I would argue this change is not correct; callers may pass arbitrarily large buffers to read stdin, and you don't want to block (possibly indefinitely) until such buffers fill before returning.

lib/posix-tty/serial.c
```diff
@@ -149,16 +165,15 @@ static ssize_t serial_write(const struct uk_file *f __maybe_unused,
...
 		size_t len = iov[i].iov_len;
-		int bytes_written;
 
 		if (unlikely(!buf && len))
 			return  -EFAULT;
 
-		bytes_written = _console_out(buf, len);
-		if (unlikely(bytes_written < 0))
-			return bytes_written;
+		rc = (int)_console_out(buf, len);
```
andreittr, line 172: Since `_console_out` is private and either writes everything or errors out, better to change its return behavior to match that of `uk_console_*_all` rather than this awkward cast.

lib/posix-tty/serial.c
```diff
@@ -278,3 +293,71 @@ const struct uk_file uk_file_serial = {
...
+		tty_cclass = UK_CONSOLE_CLASS_FB;
+		cclass_str_len = 3;
+	} else {
+		uk_pr_err("Invalid console argument\n");
+		return -EINVAL;
+	}
+
+	UK_ASSERT(tty_cclass != UK_CONSOLE_CLASS_NONE &&
+		  tty_cclass != UK_CONSOLE_CLASS_MAX);
+	UK_ASSERT(cclass_str_len);
```
andreittr, line 326: What is the purpose of these two runtime asserts?
- `tty_cclass` is a local variable assigned to immediately above; the only way it can take the two values you check against is if they are assigned mere lines above; I don't think a runtime check is remotely warranted for such a blatant dev oops
- are they to guard against uninitialized use? (i.e., a branch that forgets to set the length?) I think compiler uninitialized use warnings during dev already solve that problem.

lib/posix-tty/serial.c
```diff
@@ -278,3 +293,71 @@ const struct uk_file uk_file_serial = {
...
+
+	str = tty_cons_str + cclass_str_len;
+	idx = 0;
+	for (i = 0; str[i] != '\0'; i++) {
+		if (unlikely(str[i] < '0' || str[i] > '9')) {
+			uk_pr_err("Invalid console index\n");
+			return -EINVAL;
+		}
+
+		idx = idx * 10 + (str[i] - '0');
```
andreittr, line 336: I find it hard to believe that we have no code we can call here to do decimal parsing for us. Doing this by hand in a _posix_ layer library is very ugly (not to mention incorrect, as it makes assumptions about the char encoding).

lib/posix-tty/serial.c
```diff
@@ -278,3 +293,71 @@ const struct uk_file uk_file_serial = {
...
+
+		if (cons->cclass != tty_cclass)
+			continue;
+
+		if (!idx) {
+			tty_cons = cons;
+			break;
+		}
+
+		idx--;
```
andreittr, line 352: While I understand the more unique logic of this loop (ukconsole enumerates all consoles, but we need to find the `idx`th _of a particular class_), I believe its readability would benefit from some documenting comments.
Also, removing the newline between `if (!idx) {` and `idx--;` would better suggest these are tightly related (which they are, it's the main looping logic).

lib/posix-tty/serial.c
```diff
@@ -13,9 +13,23 @@
...
 
+#if CONFIG_LIBUKLIBPARAM
+#include <uk/libparam.h>
+#endif /* CONFIG_LIBUKLIBPARAM */
+
+static struct uk_console *tty_cons;
+static char *tty_cons_str;
+
+#if CONFIG_LIBUKLIBPARAM
+UK_LIBPARAM_PARAM_ALIAS(console, &tty_cons_str, charp,
```
andreittr, line 29: The `_str` suffix does not add to the variable's meaning beyond what we can gleam from looking at its type.
Consider either a more descriptive suffix (e.g., `_arg` to reflect it comes from the kernel cmdline, or `_name` to reflect it's the console's _name_ we're talking about), or remove it entirely, and call it `tty_console` consistent with its kernel cmdline argument.

lib/posix-tty/serial.c
```diff
@@ -278,3 +293,71 @@ const struct uk_file uk_file_serial = {
...
+
+	if (unlikely(!tty_cons)) {
+		uk_pr_err("Requested console not found\n");
+		return -ENOENT;
+	}
+
+	return 0;
+}
+
+uk_late_initcall(init_tty_cons, 0x0);
```
andreittr, line 363: I believe late init is too late for this to be called. Not only are the serial files available publicly through `posix-tty`'s API, but they are present and exported through devfs and mounted on userspace-visible VFS `/dev/*` by end of rootfs init. Calling into them at any of these points will likely crash, as `uk_console_*` is handed a NULL.

A better place would be at the tail of lib init (`uk_lib_initcall`) where the file stack is not yet assumed to work.

lib/posix-tty/serial.c
```diff
@@ -115,16 +117,20 @@ static ssize_t serial_read(const struct uk_file *f,
...
 	UK_ASSERT(f->vol == SERIAL_VOLID);
+	UK_ASSERT(f->node);
+
 	if (unlikely(off))
 		return -ESPIPE;
 
 	if (!uk_file_poll_immediate(f, UKFD_POLLIN))
 		return 0;
 
+	cons = (struct uk_console *)f->node;
```
andreittr, line 133: Why do you wait until here to assign? It's much more readable to see how a ukfile's fields are interpreted & cast into locals at the top of the function where these are declared.
Also: `void *` naturally casts to compatible pointer types in C and it's safer to _not_ explicitly cast (e.g., to not accidentaly cast away `const` on an API change).

Related: after moving the assignment to the top, the assert slightly above will reflect your intention much clearer: `UK_ASSERT(cons)`.

lib/posix-tty/serial.c
```diff
@@ -160,25 +166,29 @@ static ssize_t serial_read(const struct uk_file *f,
...
 	ssize_t total = 0;
 	int rc;
 
 	UK_ASSERT(f->vol == SERIAL_VOLID);
+	UK_ASSERT(f->node);
+
 	if (unlikely(off))
 		return -ESPIPE;
 
+	cons = (struct uk_console *)f->node;
```
andreittr, line 183: Same argument about late-assignment as for `_read`.

lib/posix-tty/serial.c
```diff
@@ -289,19 +299,71 @@ static const struct uk_file_ops serial_ops = {
...
-static uk_file_refcnt serial_ref = UK_FILE_REFCNT_INITIALIZER(serial_ref);
+struct uk_console_file {
+	/* Allocator used to create this console file */
+	struct uk_alloc *a;
+	/* Console file refcount */
+	uk_file_refcnt frefcnt;
+	/* Console file state */
+	struct uk_file_state fstate;
+	/* Base file structure */
+	struct uk_file f;
```
andreittr, line 310: A nitpick about terminology/naming within the ukfile stack:
- a "X file" always refers to a `struct uk_file` instance of driver "X"; "console file" naturally parses as "a `struct uk_file` instance backed by the console driver"
- (optional) the driver-private data struct that houses a file's state is called a "node" (b/c it's usually referenced from `f->node`)
- the memory used to store the `struct uk_file`, node data, and any other state of a file (if singular) is commonly called an "allocation" or alloc

Here are some examples across the (non-fs) file stack: [eventfd](https://github.com/unikraft/unikraft/blob/staging/lib/posix-eventfd/eventfd.c#L31), [timerfd](https://github.com/unikraft/unikraft/blob/staging/lib/posix-timerfd/timerfd.c#L40), [pipe](https://github.com/unikraft/unikraft/blob/staging/lib/posix-pipe/pipe.c#L70).

It helps to keep the terminology and naming consistent across a stack.

Also: please drop the `uk_` prefix, as (1) it's not a public symbol, and (2) even if it were, the `uk_console_*` namespace belongs to a different library.

lib/posix-tty/serial.c
```diff
@@ -289,19 +299,71 @@ static const struct uk_file_ops serial_ops = {
...
+static void consf_release(const struct uk_file *f, int what)
+{
+	struct uk_console_file *consf;
+
+	UK_ASSERT(f->vol == SERIAL_VOLID);
+
+	if (!(what & UK_FILE_RELEASE_OBJ))
+		return;
+
+	UK_ASSERT(f->node);
```
andreittr, line 327: I don't think this assert serves any purpose in the destructor. We already sanity-check the file type in `f->vol`, and no later code touches `f->node`; freeing the allocation works just as well regardless of what value is stored there.

lib/posix-tty/serial.c
```diff
@@ -289,19 +299,71 @@ static const struct uk_file_ops serial_ops = {
...
+		.refcnt = &consf->frefcnt,
+		.state = &consf->fstate,
+		.ops = &serial_ops,
+		._release = consf_release,
+	};
+
+	return consf;
+}
+
+static struct uk_console_file *tty_consf;
```
andreittr, line 365: Is there a reason to have this as a global? It's redundant, as we already have `tty_f`, and it's generally risky to keep an uncounted reference to a file's allocation around.

lib/posix-tty/include/uk/posix-serialfile.h
```diff
@@ -10,6 +10,16 @@
...
 #define UK_POSIX_TTY_CONSOLE_PRIO		UK_FS_PRIO_FSAVAIL
 
+#if CONFIG_LIBPOSIX_TTY_DEVFS
+
+#include <uk/alloc.h>
+#include <uk/console.h>
+#include <uk/file.h>
+
+int uk_console_create_devnode(struct uk_console *cons,
+			      const char *name, __sz name_len);
```
andreittr, line 20: I don't think this function should exist, as it breaks all sorts of conventions and separations:
- `lib/posix-tty` does not own the `uk_console_*` namespace
- it takes a unikraft-internal console object and exposes it at the posix filesystem level, implemented in the posix-tty layer and exported as part of the serial console ukfile; a near full vertical slice through the stack
- the only bit of posix-tty in actual use here is the console file creation; how about exporting that instead, and let the caller worry about devfs?

But this brings me to a bigger point: the file implementations in `posix-tty` never truly belonged there, we just kept them in to reduce lib spam for as long as it made sense. The pseudofiles (null, zero, void) were already moved out into `lib/ukfile-pseudo`, a library which handles the files' type, lifetime, and devfs autoregistration; I think the time has come for console files to get their own `lib/ukfile-console`, and leave `posix-tty` with clear responsibilities:
- manage the "system console", as dictated by cmdline or defaults
- initialize stdin/out/err (optionally) along with their devfs nodes

lib/posix-tty/serial.c
```diff
@@ -438,3 +446,133 @@ static int init_tty_cons(struct uk_init_ctx *ictx __unused)
...
+
+	return 0;
+}
+
+static int init_posix_tty_devfs_consnodes(struct uk_init_ctx *ictx __unused)
+{
+	__sz i, ccount, name_len, hvc_idx = 0, uart_idx = 0, fb_idx = 0;
+	struct uk_console_file *consf;
+	struct uk_console *cons;
+	char name[32];  /* Somewhat pragmatic size for tty devfs nodes */
```
andreittr, line 512: We can be even more pragmatic about this, as we know what templates we snprintf in it:
- max 4 chars for the devtype ("ttyS")
- 20 chars (length of max 64bit uint in decimal)
- 1 NUL

=> 25 chars or more should never overflow. If kept as 32 (which I don't mind), we can turn the `(__sz)rc >= sizeof(name)` check below into an assert, as it could only ever trigger if our snprintf is busted.

lib/posix-tty/serial.c
```diff
@@ -438,3 +446,133 @@ static int init_tty_cons(struct uk_init_ctx *ictx __unused)
...
+	consf = consf_create(cons);
+	if (unlikely(PTRISERR(consf))) {
+		uk_pr_err("Failed to create console file for %s\n", name);
+		return PTR2ERR(consf);
+	}
+
+	rc = consf_create_devnode(consf, cons, name, name_len);
+	if (unlikely(rc)) {
+		uk_pr_err("Failed to create the devfs entry for %s\n", name);
+		uk_free(consf->a, consf);
```
andreittr, line 500: This is extremely poor form to deal with a file's allocation directly, even if you are its owning lib; that's what we have refcounting and destructors for.

lib/posix-tty/serial.c
```diff
@@ -438,3 +446,133 @@ static int init_tty_cons(struct uk_init_ctx *ictx __unused)
 }
 
 uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_POSIX_TTY_CONSOLE_PRIO);
+
+#if CONFIG_LIBPOSIX_TTY_DEVFS
+static int consf_create_devnode(struct uk_console_file *consf,
```
andreittr, line 451: Never pass file backing allocations across functions outside of creation and destruction; it's very poor form and a huge risk for memory leaks or reference miscounting. Use the `const struct uk_file *` references directly.

lib/posix-tty/serial.c
```diff
@@ -438,3 +446,133 @@ static int init_tty_cons(struct uk_init_ctx *ictx __unused)
...
+			/* Not tracking previously allocated console files to
+			 * avoid leaks because they will be automatically
+			 * dropped on devfs drop during shutdown.
+			 */
+			uk_pr_err("Failed to create devfs node for %s\n",
+				  name);
+
+			/* never free the shared tty console file */
+			if (consf != tty_consf)
+				uk_free(consf->a, consf);
```
andreittr, line 568: This is wrong. You are freeing the backing allocation of files whose refcount you have no idea about. These were borrowed via `consf_create_devnode` (through the uncounted `consf` reference) to `uk_fs_createat` who correctly takes a refcount before returning => you are freeing memory with live references elsewhere.
Please never use a file's backing allocation directly outside of creation and destruction, use the file reference along with its refcounting mechanism.

andreittr (changes requested):

lib/posix-tty/serial.c
mogasergiu: Yeah, during prototyping I thought this was trivial enough to not need `sscanf` for, as I want to ideally avoid stuff that touches `errno`. But as you can see during devfs nodes creation I ended up using `snprintf` anyway not wanting to bother with it. So yeah will switch to `sscanf`. However we do need some `string.h` routines that promise not to touch `errno`. We do have some `uk_*` prefixes variants in `lib/ukprint` but they are not exported. Ideally we might move them to some internal lib at some point and do a cleanup wherever we can.

mogasergiu:
Force pushed untested changes

lib/posix-tty/serial.c
```diff
@@ -13,6 +13,10 @@
 #include <uk/console/driver.h>
 #include <uk/file.h>
 #include <uk/file-console.h>
+#include <uk/fs.h>
+#include <uk/devfs.h>
+#include <uk/prio.h>
```
andreittr, line 22: I believe these 3 includes should be guarded behind `CONFIG_LIBUKFS_DEVFS`, as they are used only for registering devfs entries.

andreittr (commented):
Thanks for the update! Look good overall, only one minor comment.

andreittr (approved):
Thanks for the extensive changes, looks good!

andreittr:

PR #1796, opened 2026-03-17
drivers/virtio: Add virtio-console support

mogasergiu:
### Description of Changes

This series introduces interrupt-driven console I/O as a first-class concept in the Unikraft console stack, and delivers the first driver that actually uses it: virtio-console.

At the ukconsole level, a new uk_console_async type extends uk_console with a callback list. Drivers that support interrupt-driven I/O embed uk_console_async, set UK_CONSOLE_FLAG_ASYNC_RX and/or UK_CONSOLE_FLAG_ASYNC_TX, and call uk_console_async_in_handle() / uk_console_async_out_handle() from their ISRs to fan out to all registered callbacks. Alongside this, uk_console_unregister() is introduced to allow devices to be cleanly removed at runtime — something previously impossible. Device IDs are now managed via a bitmap so that slots freed by unregistration can be reused, and the stdout/stdin/emerg_stdout auto-assignment flags are re-evaluated after every removal so that subsequent registrations behave correctly.

At the posix-tty level, the async callbacks are wired into the file event system. For each async-capable console file, ISR-safe callbacks atomically post UKFD_POLLIN / UKFD_POLLOUT bits into a per-file pending_events field and up a semaphore. A dedicated "posix-tty-evmon" thread blocks on that semaphore, atomically drains each async file's pending_events, and delivers them via uk_file_event_set. Separate serial_read_async and serial_write_async I/O ops are selected at file-creation time based on the device's async flags, allowing synchronous and asynchronous devices to coexist without branching in the hot path. A _console_out_nb non-blocking output helper is added for the async TX write path, which returns immediately when the TX queue is full and clears POLLOUT so the caller blocks until the next TX completion interrupt re-arms it.

Several prerequisites for the virtio-console driver are addressed in the process: virtio_config_set is added as the write-side counterpart of virtio_config_get; the virtio ring library is updated to use _isr printing variants to prevent infinite recursion through a virtio-backed console; and legacy virtio devices are now always set up with the full device-reported queue size rather than a driver-selected cap, since the queue_size register is read-only in legacy mode.

The virtio-console driver itself (VIRTIO_ID_CONSOLE) implements uk_console out/in ops on top of port 0's TX/RX virtqueue pair. RX buffers are pre-allocated at probe time so the host can deliver data immediately after DRIVER_OK. TX buffers are allocated per write and reclaimed lazily before each new enqueue. Emergency output is supported when the device advertises VIRTIO_CONSOLE_F_EMERG_WRITE, writing one byte at a time through the emerg_wr configuration register, bypassing the virtqueues entirely. RX interrupts are forwarded through uk_console_async_in_handle to wake the posix-tty event monitor.

The final patch extends the driver with VIRTIO_CONSOLE_F_MULTIPORT support. The device internals are refactored around a struct vtcons_port that owns its own RX/TX virtqueue pair and uk_console_async registration. Two control virtqueues are added and a dedicated control thread dispatches PORT_ADD, PORT_REMOVE, CONSOLE_PORT, and PORT_NAME events, creating and tearing down uk_console registrations and devfs nodes dynamically as the host adds or removes ports. For multiport devices, non-console ports appear under /dev/vportXportY with a named symlink in /dev/virtio-ports/; console ports are exposed through the class-based tty nodes created by lib/posix-tty. Devices that do not advertise multiport fall back to the original single-port path.

Depends-on: #1795

### Related Work

N/A

### PR Checklist


drivers/virtio/blk/virtio_blk.c
```diff
@@ -524,7 +524,16 @@ static int virtio_blkdev_vqueue_setup(struct uk_blkdev_queue *queue,
...
 	}
 
-	nr_desc = (nr_desc) ? nr_desc : max_desc;
+	/* Legacy virtio has queue_size register as RO, while modern has it RW.
+	 * This means the vring layout is immutable and setting up the queues
+	 * with anything less than the reported vq size will result in layout
+	 * mismatch between the host paravirtualized device and the
+	 * guest device driver.
+	 */
+	if (!nr_desc ||
```
michpappas, line 533: Except from the two places addressed in this commit, this logic also repeats, in a slightly different way, a few times in `virtio_console.c`. So - if I'm not mistaken - adding a small helper similar to the one below to `virtio_bus.h` would avoid repetition:
```c
static inline __u16
virtio_vqueue_get_effective_size(struct virtio_dev *vdev, __u16 desired, __u16 max)
{
    if (VIRTIO_FEATURE_HAS(vdev->features, VIRTIO_F_VERSION_1))
        return MIN(desired, max);
    return max;
}
```

lib/ukconsole/console.c
```diff
@@ -313,5 +314,119 @@ void uk_console_register(struct uk_console *dev)
...
+		return -ENOMEM;
+	}
+
+	cb->cookie = cookie;
+	cb->handler = handler;
+	cb->evflags = event;
+	UK_INIT_LIST_HEAD(&cb->_list);
+
+	async_dev = __containerof(dev, struct uk_console_async, cons);
+	uk_list_add_tail(&cb->_list, &async_dev->_cb_list);
```
michpappas, line 366: Although we now register/unregister callbacks sequentially, perhaps it would be future-proof to protect the `cb_list` with a lock?

drivers/virtio/console/virtio_console.c
```diff
@@ -0,0 +1,484 @@
...
+	UK_ASSERT(vdev);
+
+	rc = vtcons_dev_negotiate(vdev);
+	if (unlikely(rc))
+		return rc;
+
+	dev = uk_calloc(vtcons_a, 1, sizeof(*dev));
+	if (unlikely(!dev)) {
+		uk_pr_err("Failed to allocate device struct\n");
+		return -ENOMEM;
```
michpappas, line 419: According to the virtio spec (3.1.1 Driver Requirements: Device Initialization):

"If any of these steps go irrecoverably wrong, the driver SHOULD set the FAILED status bit to indicate that it has given up on the device (it can reset the device later to restart if desired). The driver MUST NOT continue initialization in that case."

So we should set the `FAILED` bit in the case of failure in `vtcons_negotiate()`, `uk_calloc()`, `vqueue_setup()`.

drivers/virtio/console/virtio_console.c
```diff
@@ -0,0 +1,484 @@
...
+
+		/* The host may have filled fewer bytes than VTCONS_RX_BUFSZ;
+		 * also clamp to the remaining space in the caller's buffer.
+		 */
+		rxlen = (__u32)MIN((__sz)rxlen, len - total);
+		memcpy(buf + total, rxbuf->data, rxlen);
+		total += rxlen;
+		reclaimed++;
+
+		uk_free(vtcons_a, rxbuf);
```
michpappas, line 197: Correct me if I'm wrong, but I think if `MIN(`rxlen`, `len - total`)` returns `len - total`, then we lose the remaining bytes by freeing the entire `rxbuf`?

drivers/virtio/console/virtio_console.c
```diff
@@ -0,0 +1,484 @@
...
+
+	/* sg.sg_nseg readable descriptors (guest -> host), 0 writable.
+	 * txbuf is the cookie returned by virtqueue_buffer_dequeue during
+	 * reclaim, so we know what to free.
+	 */
+	rc = virtqueue_buffer_enqueue(dev->txq.vq, txbuf,
+				      &sg, sg.sg_nseg, 0);
+	if (unlikely(rc < 0)) {
+		uk_free(vtcons_a, txbuf);
+		return 0;
```
michpappas, line 251: Is the return value correct? From the caller's perspective it looks more like zero bytes were written. Would something like `EIO` be more suitable?

drivers/virtio/console/virtio_console.c
```diff
@@ -401,64 +1051,285 @@ static int vtcons_dev_negotiate(struct virtio_dev *vdev)
...
+	}
+	if (unlikely(filled < dev->ports[0].rxq.nb_desc))
+		uk_pr_warn_isr("partial port 0 RX prefill (%u/%u) — heap may be low\n",
+			       filled, dev->ports[0].rxq.nb_desc);
 
-	/* Set device DRIVER_OK — host may start sending data from here. */
-	virtio_dev_drv_up(vdev);
+	if (multiport) {
+		rc = vtcons_multiport_start(dev);
+		if (unlikely(rc)) {
```
michpappas, line 1308: I think you need to release queues if `vtcons_multiport_start()` fails.

drivers/virtio/console/virtio_console.c
```diff
@@ -100,59 +102,131 @@ static const struct uk_console_ops vtcons_ops_emerg = {
...
+			break;
+		}
+
+		/* 0 readable segments (host writes into this buffer),
+		 * sg.sg_nseg writable segments.
+		 */
+		rc = virtqueue_buffer_enqueue(dev->ctlrxq.vq, ctlbuf,
+					      &sg, 0, sg.sg_nseg);
+		if (unlikely(rc < 0)) {
+			uk_free(vtcons_a, ctlbuf);
```
michpappas, line 206: Nit: Maybe an error message here similarly to `vtcons_ctlrxq_fillup()`?

michpappas (changes requested):
Very nice set of changes. My review is focused mostly on `virtio` and `uk_console`, didn't look at `posix-tty`.

lib/ukconsole/console.c
```diff
@@ -313,5 +314,119 @@ void uk_console_register(struct uk_console *dev)
...
+};
+
+int uk_console_async_register_callback(struct uk_console *dev,
+				       uk_console_async_handler_func handler,
+				       void *cookie, __u32 event)
+{
+	struct uk_console_async_callback *cb;
+	struct uk_console_async *async_dev;
+
+	if (unlikely(!dev))
```
andreittr, line 333: Does it make sense to error out on null values for `dev` and/or `handler` instead of e.g., asserting them? I would imagine that if calling code supplied null for either of these, it is indicative of programmer error and not an exceptional runtime condition.

(Similar argument for unregister callback)

lib/ukconsole/console.c
```diff
@@ -323,6 +337,51 @@ struct uk_console_async_callback {
...
+	}
+}
+
+void uk_console_unregister(struct uk_console *dev)
+{
+	struct uk_console *cons;
+
+	UK_ASSERT(dev);
+
+	uk_list_del_init(&dev->_list);
```
andreittr, line 358: I believe we need to sync changes to `uk_console_device_list` unless we can be sure that all console registrations/unregistrations happen sequentially.

lib/posix-tty/serial.c
```diff
@@ -174,6 +263,78 @@ static ssize_t serial_read(const struct uk_file *f,
...
+	ssize_t total = 0;
+	__ssz rc;
+
+	UK_ASSERT(f->node);
+
+	if (unlikely(off))
+		return -ESPIPE;
+
+	if (!uk_file_poll_immediate(f, UKFD_POLLIN))
+		return -EAGAIN;
```
andreittr, line 292: I don't think you need this here; `read()` should return -EAGAIN only after checking and confirming the console's buffer to be empty. As it is now, there is the risk of missed data if a 2nd read comes immediately after a 1st read that partially drains the buffer, but without any intervening RX events to set POLLIN back.

lib/posix-tty/serial.c
```diff
@@ -174,6 +263,78 @@ static ssize_t serial_read(const struct uk_file *f,
...
+
+	/*
+	 * Clear POLLIN before draining.  The driver re-arms its interrupt
+	 * inside in(); new data arriving after re-arm triggers a fresh
+	 * interrupt and the evmon will re-set POLLIN.
+	 */
+	uk_file_event_clear(f, UKFD_POLLIN);
```
andreittr, line 299: This might be problematic if the current read does not fully drain the buffer and no further RX events happen: there are bytes left to read, but the file does not poll readable.
A solution around this would either be to set POLLIN back if not all data was drained, or make async console files explicitly polled (so you don't have to worry about _when_ to place the falling edge of POLLIN).

lib/posix-tty/serial.c
```diff
@@ -206,6 +367,62 @@ static ssize_t serial_write(const struct uk_file *f,
...
+	ssize_t total = 0;
+	__ssz rc;
+
+	UK_ASSERT(f->node);
+
+	if (unlikely(off))
+		return -ESPIPE;
+
+	if (!uk_file_poll_immediate(f, UKFD_POLLOUT))
+		return -EAGAIN;
```
andreittr, line 395: Similar argument to async read(): write() should EAGAIN only after actually checking that the TX buffer is full. While not (immediately to my eyes) problematic like the read case, I think it's redundant here.

lib/posix-tty/serial.c
```diff
@@ -206,6 +367,62 @@ static ssize_t serial_write(const struct uk_file *f,
...
+			uk_file_event_clear(f, UKFD_POLLOUT);
+			break;
+		}
+		if (unlikely(rc < 0))
+			return (ssize_t)rc;
+
+		total += rc;
+		if ((__sz)rc < len) {
+			uk_file_event_clear(f, UKFD_POLLOUT);
+			break;
```
andreittr, line 419: This branch can be combined with the `rc == 0` one above, as both do the same thing: handle the case where TX buffer is full.

drivers/virtio/console/include/virtio/virtio_console.h
andreittr, line 1: Since this file is imported verbatim from Linux, please add provenance information to the commit message (repo, tag, path) too. Also might be worth having the raw UAPI header import as its own commit, followed by work actually making use of it.

drivers/virtio/console/Config.uk
```diff
@@ -0,0 +1,21 @@
...
+	select LIBVIRTIO_BUS
+	select LIBUKSGLIST
+	select LIBUKALLOC
+	help
+		Driver for Virtual I/O Console devices.
+
+if LIBVIRTIO_CONSOLE
+
+config LIBVIRTIO_CONSOLE_NB_DESC
+	int "Default descriptor number per virtuqueue"
```
andreittr, line 13: Typo: `virtuqueue` -> `virtqueue`

drivers/virtio/console/virtio_console.c
```diff
@@ -0,0 +1,484 @@
...
+
+	while (filled < nb) {
+		if (virtqueue_is_full(dev->rxq.vq))
+			break;
+
+		rxbuf = uk_malloc(vtcons_a, sizeof(*rxbuf));
+		if (unlikely(!rxbuf))
+			break;
+
+		uk_sglist_init(&sg, 1, sgsegs);
```
andreittr, line 123: I believe this assumes `rxbuf` to be wholly contained within contiguous physical memory, something that `uk_malloc` cannot guarantee.

drivers/virtio/console/virtio_console.c
```diff
@@ -0,0 +1,484 @@
...
+
+	vtcons_txq_reclaim(dev);
+
+	txbuf = uk_malloc(vtcons_a, len);
+	if (unlikely(!txbuf))
+		return -ENOMEM;
+
+	memcpy(txbuf, buf, len);
+
+	uk_sglist_init(&sg, 1, sgsegs);
```
andreittr, line 236: Similar argument as for `rxq_fillup`: this assumes `buf[len]` fits wholly inside contiguous physical pages; `len` itself might be larger than pagesize, making the issue even more pronounced.

drivers/virtio/console/virtio_console_devfs.c
andreittr, line 1: I believe the devfs functionality doesn't belong in the virtio console driver, and is likely to be moved into `ukfile-console`, as consequence of https://github.com/unikraft/unikraft/pull/1795#discussion_r3015659531.

I will do a full pass on it after the move.

drivers/virtio/console/virtio_console_priv.h
andreittr, line 1: Similar argument as for `virtio_console_devfs.c`: it seems quite a bit of this file is dedicated to devfs support, which I don't think belongs in a virtio driver; will review after refactor of #1795 .

drivers/virtio/console/virtio_console.c
```diff
@@ -100,59 +102,131 @@ static const struct uk_console_ops vtcons_ops_emerg = {
...
+ * Post up to @nb fresh writable descriptors to the RX virtqueue of @port.
  * Notifies the host if @notify is non-zero and at least one buffer was added.
  * Returns the number of descriptors successfully posted.
  */
-static __u16 vtcons_rxq_fillup(struct vtcons_dev *dev, __u16 nb, int notify)
+static __u16 vtcons_port_rxq_fillup(struct vtcons_port *port, __u16 nb,
+				    int notify)
 {
-	struct uk_sglist_seg sgsegs[1];
+	/* Two segments in case the allocation crosses the page boundary */
```
andreittr, line 112: Ah, I see you've fixed the sgsegs in the last commit 👍 .

drivers/virtio/console/virtio_console.c
```diff
@@ -219,21 +341,22 @@ static __ssz vtcons_in(struct uk_console *con, char *buf, __sz len)
...
+	vtcons_txq_reclaim(port);
 
 	txbuf = uk_malloc(vtcons_a, len);
 	if (unlikely(!txbuf))
 		return -ENOMEM;
 
 	memcpy(txbuf, buf, len);
 
-	uk_sglist_init(&sg, 1, sgsegs);
+	uk_sglist_init(&sg, 2, sgsegs);
```
andreittr, line 359: 2 segments might not be enough here, as `len` can be arbitrarily large; perhaps a better value would be `DIV_ROUND_UP(len, PAGE_SIZE) + 1`.

drivers/virtio/console/virtio_console.c
```diff
@@ -290,33 +413,504 @@ static __ssz vtcons_emerg_out(struct uk_console *con,
...
+	__sz namelen;
+
+	if (ctlbuf->id >= dev->max_nr_ports)
+		return;
+
+	port = &dev->ports[ctlbuf->id];
+
+	namelen = (rxlen > sizeof(*ctlbuf)) ? rxlen - sizeof(*ctlbuf) : 0;
+	if (namelen > VTCONS_PORT_NAME_LEN)
+		namelen = VTCONS_PORT_NAME_LEN;
```
andreittr, line 487: Is this truncation expected/observed behavior, or would it make more sense to error out on receiving a name that's too long?

drivers/virtio/console/virtio_console_priv.h
```diff
@@ -0,0 +1,146 @@
...
+
+	/* Per-port state (dynamically allocated array of max_nr_ports
+	 * entries).
+	 */
+	struct vtcons_port *ports;
+
+	/* All queue descriptor sizes returned by virtio_find_vqs, kept for
+	 * lazy per-port queue setup.
+	 */
+	__u16 *all_qdesc_sizes;
```
andreittr, line 111: Nit: "all" is a bit redundant here; `qdesc_sizes[]` or `qdesc_size[]` would IMHO read cleaner.

andreittr (changes requested):
1st pass focusing on the implementation itself, mostly skipping over devfs registration, as that will change as part of updates to #1795 .

drivers/virtio/blk/virtio_blk.c
mogasergiu: Thanks for the suggestion! I'd prefer to keep it as-is though, for a couple of reasons:

The `nr_desc == 0` convention is a framework-level contract (`ukblkdev`/`uknetdev`), while the `VIRTIO_F_VERSION_1` clamping is a virtio protocol constraint. Mixing the two in a shared helper or comment makes the rationale murkier, not clearer.

More practically: the helper doesn't actually remove the zero check, the caller still needs to handle `desired == 0` before invoking it, since `MIN(0, max)` would silently return `0`. At that point the helper is just wrapping a single `MIN()` call behind a feature flag check, which doesn't seem worth the indirection.

So, for example, a complete helper to justify adding maybe would be
```C
static inline __u16
virtio_vqueue_effective_size(struct virtio_dev *vdev, __u16 requested, __u16 host_max)
{
    __u16 desired = requested ? requested : host_max;

    if (VIRTIO_FEATURE_HAS(vdev->features, VIRTIO_F_VERSION_1))
        return MIN(desired, host_max);
    return host_max;
}
```
But then it mixes the framework constraints (0-check, which virtio does not need to know about) and virtio constraints.

drivers/virtio/console/virtio_console.c
mogasergiu: Fair point. I did not get the chance to test out the behavior of other OSes on this. For now, I would propose we do not error out, but do throw a warning message.

drivers/virtio/console/virtio_console_priv.h
mogasergiu: So, unfortunately, even though we discussed to go in that direction, it turns out that the devfs naming/procedure for virtio-console is so specific to this device that it makes it hard to handle it through generic event handlers outside the driver itself. For example, you not only need to know the name of the devfs node, but you may also need to know whether to create it in the `virtio-ports` subdirectory or not or if to create a symlink to it...

drivers/virtio/console/virtio_console_devfs.c
mogasergiu: See https://github.com/unikraft/unikraft/pull/1796/changes#r3126330358

mogasergiu:
All comments addressed, except for https://github.com/unikraft/unikraft/pull/1796#discussion_r3078419565, which I will solve tomorrow. Besides that, I noticed that some early boot printings are skipped because the driver does not register its emergency writing functionality as an early console so I will solve that too tomorrow.

EDIT: I will only solve the first issue now, as the early console registration requires further architectural discussion (not as simple as I presumed).

mogasergiu:
Pushed update addressing https://github.com/unikraft/unikraft/pull/1796#discussion_r3078419565

drivers/virtio/console/virtio_console_priv.h
andreittr: That's fair, let's leave it here for the time being.

andreittr (approved):
Looks great, thanks!

michpappas (approved):

andreittr:

PR #1799, opened 2026-03-26
 lib/ukpcpuvar/arch/arm64: Use CPU idx * section size not section base 

mogasergiu:
### Description of Changes

On the per-CPU variable side, the ARM64 macros are corrected to stop subtracting _uk_pcpuvar_base from symbol addresses, since TPIDR_EL1 holds cpu_idx * per_cpu_area_size, not the section base.
### Related Work

N/A

### PR Checklist

### DIFF
--- a/lib/ukpcpuvar/arch/arm64/include/uk/pcpuvar/arch.h
+++ b/lib/ukpcpuvar/arch/arm64/include/uk/pcpuvar/arch.h
@@ -18,16 +18,16 @@ extern "C" {
 /**
  * Read per-CPU variable into register.
  *
- * @dst: Destination register (will contain the value)
- * @sym: Symbol name
- * @tmp: Scratch register
+ * @param dst
+ *   Destination register (will contain the value)
+ * @param sym
+ *   Symbol name
+ * @param tmp
+ *   Scratch register (will hold &sym)
  */
 .macro uk_pcpuvar_arm64_ldr dst:req, sym:req, tmp:req
 	adrp	\tmp, \sym
 	add	\tmp, \tmp, #:lo12:\sym
-	adrp	\dst, _uk_pcpuvar_base
-	add	\dst, \dst, #:lo12:_uk_pcpuvar_base
-	sub	\tmp, \tmp, \dst
 	mrs	\dst, tpidr_el1
 	ldr	\dst, [\dst, \tmp]
 .endm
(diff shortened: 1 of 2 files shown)


drivers/ukintctlr/xpic/ukintctlr.c
```diff
@@ -75,7 +75,7 @@ static int configure_irq(struct uk_intctlr_irq *irq __unused)
 static int uk_intctlr_xpic_handle_irq(void *data)
 {
 	struct uk_lcpu_except_irq_ctx *ctx;
-	unsigned long irq;
+	__u64 irq;
```
michpappas: On the commit message:

"Since we exclusively only support" should either use "exclusively" or "only". But more importantly, 
the irq type change to `__u64` is inconsistent with the GIC ops which use `__u32` throughout, and will force propagating the change into the GIC driver, or force the caller make ugly casts. As the IRQ number is inherently a small integer (10 bit on `arm64` , and 4 bits for SGIs) — `__u32` or `unsigned int` better reflects the hardware.  I'd suggest splitting the irq type change from the error_code type change, and keeping irq at its current width. If you still want to use an explicit type nevertheless (personally I like explicit types when working with hardware), perhaps  `__u32` would be a more suitable type.

For the `error_code` change to `__u64` I don't have an opinion as it's `x86_64` (the `ESR` part in the `arm64` structure is `__u64` already).

lib/ukpcpuvar/arch/arm64/include/uk/pcpuvar/arch.h
```diff
@@ -108,7 +103,7 @@ extern "C" {
  */
 #define __uk_pcpuvar_arch_current_get(_sym)				\
 	({								\
-		__u64 _offset = (__u64)&(_sym) - (__u64)_uk_pcpuvar_base; \
+		__u64 _offset = (__u64)&(_sym);				\
```
michpappas, line 106: Although both implementations yield the same result, I'm skeptical about this change:

In the current implementation `TPIDR_EL1` holds the absolute virtual address of this CPU's per-cpu data slot. The loads then are made from `[TPIDR_EL1, offset]`. This reads naturally as "base + offset" — the standard addressing pattern.

With the proposed change, `TPIDR_EL1` holds just an index, not a pointer. The macros use `&sym` directly (the full VA) and load from `[TPIDR_EL1, &sym]`, where the "base register" holds a small number and the "offset register" holds the actual VA. This is functionally correct but reads backwards compared to the conventional "base + offset" pattern.

I'd prefer keeping `TPIDR_EL1` as an absolute VA because:

1. It's more intuitive — `TPIDR_EL1` is a pointer to this CPU's data, consistent with how per-cpu base registers are typically used.
2. The `[base_reg, offset_reg]` addressing reads naturally when the base register actually holds the base address.

mogasergiu:
Addressed copilot comments

lib/ukpcpuvar/arch/arm64/include/uk/pcpuvar/arch.h
```diff
@@ -18,26 +18,23 @@ extern "C" {
...
  */
 .macro uk_pcpuvar_arm64_ldr dst:req, sym:req, tmp:req
 	adrp	\tmp, \sym
 	add	\tmp, \tmp, #:lo12:\sym
-	adrp	\dst, _uk_pcpuvar_base
-	add	\dst, \dst, #:lo12:_uk_pcpuvar_base
-	sub	\tmp, \tmp, \dst
 	mrs	\dst, tpidr_el1
 	ldr	\dst, [\dst, \tmp]
 .endm
```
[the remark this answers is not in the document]
mogasergiu: To be fixed by https://github.com/unikraft/unikraft/pull/1772

michpappas (changes requested):

mogasergiu:
All commits but the current one migrated to #1807

PR #1800, opened 2026-03-26
drivers/virtio: Add `virtio-vsock` support

mogasergiu:
### Description of Changes

This patch set adds support for AF_VSOCK sockets to Unikraft.

The core of the work is a new libukvsockdev library that provides the complete guest-side socket infrastructure: a circular receive buffer, a connection state machine, an O(log n) rbtree for socket lookup, and a POSIX_SOCKET-compatible AF_VSOCK family driver. The family driver implements the full set of POSIX socket operations — connect, bind, listen, accept, send, recv, shutdown, getsockname, getpeername, and getsockopt(SO_ERROR) — with implicit bind on listen and automatic ephemeral port allocation on connect. An optional dispatcher thread model decouples interrupt delivery from socket event processing.

The actual transport is provided by a new libvirtio_vsock driver that implements the ukvsockdev interface on top of three virtqueues (RX, TX, EVENT), with credit-based flow control, virtio feature negotiation, and correct handling of the EVENT queue for transport-level notifications.

Live migration and snapshot/restore are handled through a transport reset path that tears down established connections while preserving sockets bound to CID_ANY, integrated with libukpm via a new system migration event that allows registered handlers to react when the guest is migrated.

### Related Work

N/A

### PR Checklist


lib/ukvsockdev/include/linux/vm_sockets.h
andreittr, line 1: Since this is a Linux UAPI header, we should import it with minimal changes along with provenance information in the commit message, as that will make for easier updates from upstream in the future.

lib/ukvsockdev/include/uk/vsock.h
```diff
@@ -0,0 +1,337 @@
...
+#include <uk/netbuf.h>
+#include <uk/tree.h>
+#include <uk/socket.h>
+
+#define UK_VSOCK_CID_ANY			-1U
+#define UK_VSOCK_CID_HYPERVISOR			0U
+#define UK_VSOCK_CID_LOCAL			1U
+#define UK_VSOCK_CID_HOST			2U
+
+#define UK_VSOCK_PORT_ANY			-1U
```
andreittr, line 20: Could we use the constants provided by `vm_sockets.h` here instead of duplicating the values?

lib/ukvsockdev/include/uk/vsock.h
```diff
@@ -0,0 +1,337 @@
...
+};
+
+UK_STAILQ_HEAD(uk_vsock_accept_queue, struct uk_vsock_accept_entry);
+
+/**
+ * Generic representation of a VSOCK socket.
+ * Drivers are encouraged to embed this structure in a driver specific socket
+ * data structure.
+ *
+ * Unless otherwise indicates the fields should be considered read-only for the
```
andreittr, line 58: Typo: `indicates` -> `indicated`

lib/ukvsockdev/include/uk/vsock.h
```diff
@@ -0,0 +1,337 @@
...
+/**
+ * Initialize a vsock buffer with the specified capacity.
+ * Allocates memory for the circular buffer and initializes all tracking fields.
+ * The buffer uses a circular design for efficient memory utilization and
+ * automatic wraparound handling.
+ *
+ * @param buf The buffer structure to initialize
+ * @param a The memory allocator to use for buffer allocation
+ * @param bufsize The size of the buffer to allocate in bytes
+ * @return 0 on success, non-zero on error
```
andreittr, line 109: Minor nitpick for most functions in this file: we should more clearly document the error return case: whether it returns a negative errno specific to the failure or an arbitrary generic non-zero value. (Looking at the impl, it seems to be the former)

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+	uk_pr_debug("buf=%p requested=%zu available=%zu\n",
+		    buf, size, buf->count);
+
+	size = MIN(size, buf->count);
+
+	start = buf->head;
+	end = start + size;
+
+	to_copy = MIN(end, buf->capacity) - MIN(start, buf->capacity);
+	if (to_copy > 0) {
```
andreittr, line 95: Since `to_copy` is unsigned `__sz`, this branch will always trigger for any non-zero value, probably not intended.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+		memcpy(out, &buf->buf[start], to_copy);
+	}
+	if (end > buf->capacity) {
+		uk_pr_debug("wrap-around read: %zu bytes from buf[0]\n",
+			    end - start - to_copy);
+		memcpy(out + to_copy, &buf->buf[0], end - start - to_copy);
+	}
+
+	uk_pr_debug("returning %zu bytes\n", end - start);
+	return end - start;
```
andreittr, line 107: Nit: Since `end` and `start` do not change after being assigned, I believe it'd be clearer to just return `size` here.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+/* rb tree that stores the nodes according to their local port number and type
+ */
+UK_RB_HEAD(uk_vsockets, uk_vsock);
+UK_RB_KEY_GENERATE_STATIC(uk_vsockets, uk_vsock, rb_entry,
+			  uk_vsock_sockaddr_cmp, uk_vsock_sockaddr_key);
+
+/* List of all active sockets */
+static struct uk_vsockets sockets;
+
+/* State for port searching (shared across socket types for simplication) */
```
andreittr, line 188: Typo: `simplication` -> `simplicity` (likely what was meant; either that or "simplification")

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+	};
+}
+
+/* rb tree that stores the nodes according to their local port number and type
+ */
+UK_RB_HEAD(uk_vsockets, uk_vsock);
+UK_RB_KEY_GENERATE_STATIC(uk_vsockets, uk_vsock, rb_entry,
+			  uk_vsock_sockaddr_cmp, uk_vsock_sockaddr_key);
+
+/* List of all active sockets */
```
andreittr, line 185: "List" is a bit misleading here, as the structure is a tree. "All active sockets" is descriptive enough while being agnostic to the datastruct in use.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+
+int uk_vsockets_init(void)
+{
+	uk_pr_debug("initializing vsockets RB tree\n");
+	UK_RB_INIT(&sockets);
+	uk_pr_debug("vsockets RB tree initialized\n");
+	return 0;
+}
+
+static __s32 vsock_alloc_port(int type)
```
andreittr, line 199: This function touches a global var, but can be called concurrently from `bind` or `connect` on different sockets (bind/connect are serialized per-socket by posix-socket, but not globally). You should take care that accesses to `port_alloc_next` are consistent in this case.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+int uk_vsock_init(struct uk_vsock *sock, struct uk_alloc *a, int sock_type)
+{
+	int rc;
+
+	uk_pr_debug("sock=%p sock_type=%d\n", sock, sock_type);
+
+	memset(sock, 0, sizeof(*sock));
+
+	rc = uk_vsock_buffer_init(&sock->rx, a,
+				  4096 /* FIXME: Proper buffer size */);
```
andreittr, line 243: What would a proper buffer size be in this case? A larger value, or something configurable?
(I'm not asking to fix the value, but the comment needs to be more descriptive of what exactly needs "fixing" about 4096).

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+		}
+		uk_pr_debug("removing WAIT_CONNECT sock=%p from tree\n", sock);
+		UK_RB_REMOVE(uk_vsockets, &sockets, sock);
+		break;
+	case UK_VSOCK_STATE_READY:
+	case UK_VSOCK_STATE_CLOSED:
+		/* nothing to do */
+		uk_pr_debug("sock=%p already in state=%d, nothing to do\n",
+			    sock, sock->state);
+		return 0;
```
andreittr, line 341: I understand the logic of doing nothing if the socket is already in a closed state, but why have the same for `READY`?
AFAICT if this function succeeds the socket is expected to be in the CLOSED state with the shutdown flags set; that happens for all cases except `READY` sockets, is that on purpose?

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+				 struct uk_vsockaddr local,
+				 struct uk_vsockaddr peer)
+{
+	struct uk_vsock *sock;
+
+	uk_pr_debug("type=%d local={cid=%u port=%u} peer={cid=%u port=%u}\n",
+		    sock_type, local.cid, local.port, peer.cid, peer.port);
+
+	/* Try to find an exact match */
+	sock = uk_vsockets_RB_FIND(&sockets, (struct uk_vsock_conn_id){
```
andreittr, line 398: You should use the macros provided by `<uk/tree.h>` to call auto-generated functions, rather than writing their name out by hand: `UK_RB_FIND(uk_vsockets, ...)`

(applies in several places in this file)

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+						.type = sock_type,
+						.local = local,
+						.peer = peer,
+						.mask = 0,
+					     });
+	if (likely(sock)) {
```
andreittr, line 404: I'm not sure we have a good enough reason to have this branch marked as "likely"; unless we can argue that exact lookups will be the vast majority of successful uses, I think we can leave this branch untagged.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+		return ERR2PTR(rc);
+	}
+
+	/* Mark the socket as open and register it in the sockets list */
+	out->state = UK_VSOCK_STATE_OPEN;
+
+	uk_pr_debug("accepted new sock=%p peer={cid=%u port=%u}, inserting into tree\n",
+		    out, out->peer_addr.cid, out->peer_addr.port);
+
+	UK_RB_INSERT(uk_vsockets, &sockets, out);
```
andreittr, line 777: Possible concurrent inserts on global `&sockets`; similar concern with port number lookup; we should ensure some sort of synchronization here.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+		memcpy(addr, &svm, size);
+		*addr_len = size;
+		uk_pr_debug("returning peer addr cid=%u port=%u\n",
+			    svm.svm_cid, svm.svm_port);
+	}
+
+	if (UK_STAILQ_EMPTY(&vsock->accept_queue)) {
+		uk_pr_debug("accept queue now empty, clearing EPOLLIN on vsock=%p\n",
+			    vsock);
+		posix_sock_event_clear(sock, EPOLLIN);
```
andreittr, line 794: If we clear POLLIN here, we must ensure that handling of any incoming connection requests are serialized w.r.t. `accept4` calls, otherwise we risk losing events. And more generally ensure that concurrent accesses to the accept queue make sense.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+			uk_pr_debug("port allocation failed: %d\n", rc);
+			return rc;
+		}
+		port = rc;
+		uk_pr_debug("allocated port=%u\n", port);
+	}
+
+	uk_pr_debug("checking if port=%u is in use\n", port);
+
+	other = uk_vsockets_RB_FIND(&sockets,
```
andreittr, line 874: Is this lookup still needed if we're passed `PORT_ANY` and we allocate the port dynamically through `vsock_alloc_port`?

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+
+	uk_pr_debug("sock=%p backlog=%d\n", sock, backlog);
+
+	vsock = posix_sock_get_data(sock);
+
+	uk_pr_debug("vsock=%p state=%d\n", vsock, vsock->state);
+
+	if (vsock->state == UK_VSOCK_STATE_LISTENING) {
+		/* Allow re-calling listen() to update backlog */
+		vsock->accept_queue_limit = backlog ? backlog : 128;
```
andreittr, line 921: What does 128 stand for here?
Judging by `listen(2)` backlog should be capped to max `SOMAXCONN` (which used to be 128 in older Linux kernels).

That said, I don't think the logic here is correct for a few reasons:
- no validation is done against negative values for backlog
- `backlog == 0` should not set it to `SOMAXCONN`/128, rather leave it as-is
- backlog is not bounded to max `SOMAXCONN`/128

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+				sizeof(bind));
+		if (unlikely(rc)) {
+			uk_pr_debug("implicit bind failed: %d\n", rc);
+			return rc;
+		}
+		uk_pr_debug("implicit bind succeeded\n");
+	}
+	UK_ASSERT(vsock->state == UK_VSOCK_STATE_BOUND);
+
+	vsock->accept_queue_limit = backlog ? backlog : 128;
```
andreittr, line 950: Duplicated logic for validating/capping out `backlog`; this should be common with the one under `STATE_LISTENING`.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+
+	if (vsock->state != UK_VSOCK_STATE_BOUND &&
+	    vsock->state != UK_VSOCK_STATE_READY) {
+		uk_pr_debug("vsock=%p already connected or invalid state=%d\n",
+			    vsock, vsock->state);
+		return -EISCONN;
+	}
+
+	vsock->peer_addr.cid = svm->svm_cid;
+	vsock->peer_addr.port = svm->svm_port;
```
andreittr, line 1072: These lines change key fields of `vsock` while it is still live in the `sockets` RBTree, leading to an inconsistent tree state before doing the REINSERT a few lines down. If other threads are not excluded from touching the RBTree in the meantime, they will walk an invalid tree, with unpredictable results.

lib/ukvsockdev/vsock.c
```diff
@@ -0,0 +1,1547 @@
...
+		     vsock->state != UK_VSOCK_STATE_CLOSED)) {
+		uk_pr_debug("vsock=%p not in OPEN/CLOSED state (state=%d)\n",
+			    vsock, vsock->state);
+		return -ENOTCONN;
+	}
+
+	for (ioi = 0; ioi < iovcnt; ioi++) {
+		uk_pr_debug("reading into iov[%zu] base=%p len=%zu\n",
+			    ioi, iov[ioi].iov_base, iov[ioi].iov_len);
+		res = uk_vsock_buffer_read(&vsock->rx,
```
andreittr, line 1408: Keep in mind that multiple read/recv/send/write ops can happen concurrently on a socket (posix-sockets manages locking for sockets, and all I/O ops are protected by a reader lock, with only control ops grabbing an exclusive lock), and thus buffer accesses should be synced at the driver level.

(applies to all I/O ops)

andreittr (changes requested):
Review of only the vsock socket implementation; virtio-vsock driver coming up.

drivers/virtio/vsock/include/virtio/virtio_vsock.h
andreittr, line 1: The contents of this file seem to be (partially or totally) imported from Linux's `virtio_vsock.h` UAPI header. As is the case with other UAPI headers, we should import them with provenance info and minimal changes to facilitate future updates from upstream.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2029 @@
...
+	if (ret < 0) {
+		uk_pr_debug("RX vq empty on vv=%p\n", vv);
+		*pkt = __NULL;
+		return queue->nb_desc;
+	}
+
+	if (unlikely(len > buf->len)) {
+		uk_pr_err("VMM sent bigger packet than the buffer size\n");
+		*pkt = __NULL;
+		return -ENOSPC;
```
andreittr, line 307: It seems that the dequeued packet is lost if the supplied buffer is not large enough; is this intended behavior, or should the packet be re-queued for a later dequeue with a bigger buffer?

lib/ukvsockdev/vsock.c
mogasergiu: uh, yeah, somehow I think this was meant to avoid a 0-sized memcpy, but it looks like it can cause confusion. what is more, it is debateable whether an if on every execution is better than an unnecessary function call on certain cases, will remove the `if` 👍

lib/ukvsockdev/vsock.c
mogasergiu: Agreed; 4096 is just an initial conservative default. I’ll update the comment to state that the RX buffer size should ideally be configurable or derived from transport credit / queue sizing rather than being a hard‑coded 4 KiB.

lib/ukvsockdev/vsock.c
mogasergiu: That’s a good catch. There’s no strong reason to keep READY sockets in READY after a successful `uk_vsock_close()`. I’ll drop READY from the early‑return case so READY sockets also transition to CLOSED and have their shutdown flags set.

lib/ukvsockdev/vsock.c
mogasergiu: Correct: once vsock_alloc_port() has returned a port, the extra lookup is redundant for the PORT_ANY case. I’ll skip the RB lookup in that branch and keep it only when the user supplied an explicit port.

drivers/virtio/vsock/virtio_vsock.c
mogasergiu: Good observation. If the VMM sends more data than fits into our RX buffer, that’s a configuration error, and the packet should be dropped rather than re-queued (we don’t have a larger buffer to offer). I’ll make that explicit by logging, freeing the dequeued buffer, and documenting that packets larger than the configured RX buffer size are dropped with `-ENOSPC`.

mogasergiu:
Force push rebase

lib/ukvsockdev/vsock.c
mogasergiu: So, according to https://www.man7.org/linux/man-pages/man3/listen.3p.html
>  The implementation may have an upper limit on the length of the listen queue—either global or per accepting socket. If backlog exceeds this limit, the length of the listen queue is set to the limit.
If listen() is called with a backlog argument value that is less than 0, the function behaves as if it had been called with a backlog argument value of 0.
A backlog argument of 0 may allow the socket to accept connections, in which case the length of the listen queue may be set to an implementation-defined minimum value.

After an offline discussion we agreed to proceed with:
```C
if (backlog <= 0)
    backlog = 1;
else
    backlog = MIN(backlog, UK_VSOCK_BACKLOG_MAX);
```

mogasergiu:
Rebase fpush

lib/ukvsockdev/vsock.c
andreittr: Yeah, my bad, I had only read the listen(2) page and (wrongly) assumed behavior from there :facepalm: ; implementing the POSIX standards is the actual correct way to go :+1: .

mogasergiu:
fix commit message type: s/icnlude/include

mogasergiu:
pushed rebase

andreittr (approved):
Thanks for all the changes, looks great!

NatsuCamellia:
Hi, I'm an undergraduate CS student from Taiwan researching vsock. I've been trying to run your implementation and noticed that Unikraft currently only supports legacy virtio PCI, while the Linux vhost-vsock driver requires modern virtio PCI — so I haven't been able to get it working yet.

Could you share how you run the virtio-vsock driver in your setup? Any hints would be greatly appreciated. Thank you!

mogasergiu:
> Hi, I'm an undergraduate CS student from Taiwan researching vsock. I've been trying to run your implementation and noticed that Unikraft currently only supports legacy virtio PCI, while the Linux vhost-vsock driver requires modern virtio PCI — so I haven't been able to get it working yet.
> 
> Could you share how you run the virtio-vsock driver in your setup? Any hints would be greatly appreciated. Thank you!

Hi, we used Firecracker's modern virtio-mmio-based vsock for testing/developing.

NatsuCamellia:
> > Hi, I'm an undergraduate CS student from Taiwan researching vsock. I've been trying to run your implementation and noticed that Unikraft currently only supports legacy virtio PCI, while the Linux vhost-vsock driver requires modern virtio PCI — so I haven't been able to get it working yet.
> > Could you share how you run the virtio-vsock driver in your setup? Any hints would be greatly appreciated. Thank you!
> 
> Hi, we used Firecracker's modern virtio-mmio-based vsock for testing/developing.

Thank you! I switched to virtio-mmio-based vsock and got it working. Really appreciate your help!

mogasergiu:
Pushed some fixups:
- use `sock->state` not `hdr->op` when checking for listening state during RX work
- declare `UK_PM_EVENT_SYSMIGRATION` before the function that does the actual raising of the event
- make vsock device getter available globally, similar to how blkdev/netdev do
- don't assert for buffer while holding lock in `uk_vsock_buffer_destroy`
- don't forget to release spinlock in `uk_vsock_buffer_append` on error
- we have 32-bit range for vsock ports so use `__U32_MAX` instead of `__U16_MAX`
- `uk_vsock_close` is called while holding a lock in `uk_vsockets_term` so make sure to only try to acquire the lock when removing a socket if needed
- POSIX says `recvfrom` should just ignore the `from` parameter for connected sockets or fill it in with the peer address so fill it in if needed instead of returning `-EISCONN`

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+	rc = virtio_config_get(vv->vdev,
+			       __offsetof(struct virtio_vsock_config,
+					  guest_cid),
+			       &guest_cid, sizeof(guest_cid), 1);
+	if (unlikely(rc)) {
+		uk_pr_err("Unable to read guest CID from virtio-vsock device: read %d bytes but expected %zu bytes\n",
+			  rc, sizeof(guest_cid));
+		return -EAGAIN;
+	}
+	if (unlikely(guest_cid < 3)) {
```
michpappas: Since you're doing the check, you can also consider rejecting `VSOCK_CID_ANY`.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+		} else {
+			uk_pr_debug("got unexpected op %u to listening socket\n",
+				    hdr->op);
+		}
+		goto out_free;
+	}
+
+	switch ((enum virtio_vsock_op)hdr->op) {
+	case VIRTIO_VSOCK_OP_REQUEST:
+		uk_pr_warn("got unexpected connection request to non-listening socket\n");
```
michpappas, line 1076: I think you also need to reply with `VIRTIO_VSOCK_OP_RST` here. According to https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.html: "A VIRTIO_VSOCK_OP_RST reply is sent if a listening socket does not exist on the destination or the destination has insufficient resources to establish the connection ".

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+	.shutdown = virtio_vsock_shutdown,
+	.reset = virtio_vsock_reset,
+	.destroy = virtio_vsock_destroy,
+};
+
+static struct uk_alloc *drv_alloc;
+
+static inline __u32 virtio_vsock_peer_rx_free(struct virtio_vsock *vs)
+{
+	return vs->peer_buf_alloc - (vs->tx_count - vs->peer_fwd_count);
```
michpappas: Notice here that if the peer provides a value of `peer_fwd_count` so that `vs->tx_count < vs->peer_fwd_count`, this calculation will result into an integer overflow. I didn't manage to find an exploitable path, but notice how close is `virtio_vsock_send()` is: by having this function to retur a value near `UINT32_MAX`, if it wasn't for the `min()` check, an attacker could could have `pkt_len` set to a very large value, resulting into wrap around to a small value when allocating the buffer, and immediately followed by a `memcpy()` of a large `pkt_len` bytes of attacker controlled data 💀. I also didn't look much at other consequences, but it's better to perform some sanity checks.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+/* Raw packet transmit with retry for full ring */
+static int virtio_vsockdev_xmit_retry(struct virtio_vsockdev *vv,
+				      struct uk_netbuf *pkt)
+{
+	int rc;
+
+	uk_pr_debug("vv=%p pkt=%p\n", vv, pkt);
+
+	rc = virtio_vsockdev_xmit(vv, pkt);
+	while (rc == -EAGAIN) {
```
michpappas: Assuming a misbehaving or malicious peer, should we give up at some point?

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+	default:
+		/* FIXME: The virtio spec requires sending a RST answer */
+		uk_pr_err("Got invalid virtio operation: %x\n", hdr->op);
+		break;
+	}
+
+out_free:
+	if (pkt)
+		uk_netbuf_free(pkt);
+	if (unlikely(rc < 0))
```
michpappas, line 1125: Uhhh, this looks like bad form. Without going through the code extensively, it looks like there are cases like line 958 where we fall to `out_free` with the value of `rc` being set by happenstance?

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+					.port = hdr->dst_port,
+				},
+				(struct uk_vsockaddr){
+					.cid = hdr->src_cid,
+					.port = hdr->src_port,
+				});
+	if (unlikely(!sock)) {
+		uk_pr_info("Got packet to port %u without socket\n",
+			   hdr->dst_port);
+		goto out_free;
```
michpappas, line 1038: I think you also need to send back a `VSOCK_OP_RST`? May be missing elsewhere too.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+	ret = virtqueue_buffer_dequeue(queue->vq, (void **)&buf, &len);
+	if (ret < 0) {
+		uk_pr_debug("event vq empty on vv=%p\n", vv);
+		*event = __NULL;
+		return queue->nb_desc;
+	}
+
+	if (unlikely(len != sizeof(*buf))) {
+		uk_pr_err("VMM sent unexpected vsock event structure\n");
+		*event = __NULL;
```
michpappas, line 540: If I understand this correctly, I think you need to free the buffer here?

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+		return (rc == -ECONNREFUSED) ? 0 : rc;
+	}
+
+	uk_pr_debug("conn request accepted, entry=%p enqueued\n",
+		    accept_entry);
+
+	return 0;
+}
+
+static int virtio_vsockdev_work_rx(struct virtio_vsockdev *vv)
```
michpappas: Nit: The name `vv` is used in different places for `struct virtio_vsockdev` and `struct virtio_vsock`. This is quite confusing.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2032 @@
...
+			   (unsigned long long)hdr->dst_cid, hdr->dst_port);
+		goto out_free;
+	}
+
+	switch (hdr->type) {
+	case VIRTIO_VSOCK_TYPE_STREAM:
+		socket_type = SOCK_STREAM;
+		break;
+	case VIRTIO_VSOCK_TYPE_SEQPACKET:
+		socket_type = SOCK_SEQPACKET;
```
michpappas, line 999: Maybe check if `SEQPACKET` was negotiated?

michpappas (changes requested):
@mogasergiu I only went through the driver. Overall looks great, but I think it needs more validation on data supplied from the peer, which can not only be the vmm, but also an untrusted VM. I only provide an example, but it would be worth having a deeper look.

mogasergiu:
Rebase fpush

mogasergiu:
fpush addressing comments

Static analysis results don't seem to apply?

drivers/virtio/vsock/virtio_vsock.c
mogasergiu: Agreed and pushed, seems to have been like this since the version prior to me picking up the patches for rebasing/refactoring. A simple sed wouldn't work, it was that messy and all over the place. You also have `struct uk_vsock` mixing up in there so I ended up mostly doing it manually, so I would appreciate an attentive look to make sure I did not misrenamed :laughing: .

mogasergiu:
fpushed a check to avoid DoS in case `peer_buf_alloc` is unreasonably large

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2145 @@
...
+	}
+	if (!(status & VIRTIO_VSOCKDEV_STATUS_SUCCESS)) {
+		uk_pr_debug("no packet available (status=0x%x)\n", status);
+		return 0;
+	}
+	UK_ASSERT(pkt);
+
+	hdr = pkt->data;
+	rc = uk_netbuf_header(pkt, -(__u16)sizeof(struct virtio_vsock_hdr));
+	UK_ASSERT(rc == 1);
```
michpappas: I think `UK_ASSERT(rc == 1)` should be replaced with an actual check, as disabling assertions skips a check on untrusted data (consider the case where the other side supplies a `pkt->len` smaller than  `sizeof(struct virtio_vsock_hdr))`).

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2145 @@
...
+static int virtio_vsock_destroy(struct uk_vsockdev *vd __unused,
+				struct uk_vsock *sock)
+{
+	struct virtio_vsock *vv;
+
+	vv = to_virtio_vsock(sock);
+
+	uk_pr_debug("sock=%p vv=%p freeing\n", sock, vv);
+
+	uk_free(drv_alloc, vv);
```
michpappas, line 1713: Maybe something that prevents UAF caused by races would be useful here?

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2145 @@
...
+
+static struct uk_alloc *drv_alloc;
+
+/**
+ * Return the effective TX window (in bytes) we are willing to use
+ * for this connection. This is the minimum of:
+ * - peer_buf_alloc: what the peer advertises via buf_alloc, and
+ * - our own vsock RX buffer capacity (uk_vsock_buffer_capacity()).
+ */
+static inline __u32 virtio_vsock_tx_buf_size(struct virtio_vsock *vv)
```
michpappas, line 153: I don't understand the rationale for using the rx buffer to cap tx buf size. If it's correct please document. If incorrect, and the intention is to protect from DoS, maybe we can use our tx buffer size and / or some configurable defaults? In the scope of DoS maybe also configurable defaults for max connections per peer would also make sense.

michpappas (changes requested):
Another round, hopefully last one!

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2145 @@
...
+
+	uk_pr_debug("peer_rx_free=%u requested=%zu effective pkt_len=%u\n",
+		    virtio_vsock_peer_rx_free(vv), size, pkt_len);
+
+	if (pkt_len == 0) {
+		uk_pr_debug("peer RX buffer full, returning EAGAIN\n");
+		return -EAGAIN;
+	}
+
+	pkt = uk_netbuf_alloc_buf(drv_alloc, pkt_len + sizeof(*hdr), 8,
```
michpappas, line 1496: Notice that the cap in the implementation of `virtio_vsock_peer_rx_free()` does still not protect us in this function, as `pkt_len` is `__u32`, and the additon of `sizeof(*hdr)` to its high capped value (`UINT32_MAX`) would cause an overflow, which would result into small allocation, and a buffer overflow when we use the large `pkt_len` later in `memcpy()`.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2171 @@
...
+	if (!(status & VIRTIO_VSOCKDEV_STATUS_SUCCESS)) {
+		uk_pr_debug("no packet available (status=0x%x)\n", status);
+		return 0;
+	}
+	UK_ASSERT(pkt);
+
+	hdr = pkt->data;
+	rc = uk_netbuf_header(pkt, -(__u16)sizeof(struct virtio_vsock_hdr));
+	if (unlikely(rc != 1))
+		UK_CRASH("Not enough room to prepend virtio vsock header\n");
```
michpappas: A crash looks quite risky for DoS. If possible perhaps we should handle such cases gracefully.

drivers/virtio/vsock/virtio_vsock.c
```diff
@@ -0,0 +1,2171 @@
...
+};
+
+static struct uk_alloc *drv_alloc;
+
+/*
+ * Return the effective TX window (in bytes) for this connection, i.e. the
+ * minimum of:
+ * - peer_buf_alloc: the peer's advertised RX buffer (buf_alloc), and
+ * - our own local buffer size (uk_vsock::buf_size), the value we ourselves
+ *   advertise to the peer and which is clamped to max_buf_size. While the RX
```
michpappas: > While the RX buffer is independent from the TX buffer, we just reuse the same cap.

Now that we use the same buffer size for both queues I think this sentence is redundant.

michpappas (changes requested):
Last round!

michpappas (approved):
Thanks for putting up with the multiple review rounds!

andreittr:

Issue #956, opened 2023-06-22
Handle `CACHE_LINE_SIZE` constant in a way that reflects the underlying physical CPU.


Starnox:
### Feature request summary

Right now, under each specific `arch` in `lcpu.h` a constant named `CACHE_LINE_SIZE` is specified with its value set to the arch bit size (eg. for `x86_64` the constant is set to `64`), though this is not always the case, as this variable depends on the physical CPU that the kernel runs on. Being able to automatically detect and set this would provide performance benefits, taking advantage of the underlying hardware.

### Related architectures

None

### Related platforms

None

Ecazares15:
Hello!

We are students from the University of Texas at Austin taking a virtualization course (cs360v) looking for opportunities to contribute to an open source project for class credit.

Could I be assigned to this?

PR #1801, opened 2026-03-29
arch/include: detect cache line size at compile time

lucachot:
### Description of Changes

Fixes issue #956. Previously, `UK_ARCH_CACHE_LINE_SIZE` was hardcoded to 64
bytes. This value may not hold true for all CPU architectures, potentially
incurring performance costs due to false sharing. As this macro is used to
define structs and ELF section layout at compile time, a run-time approach is
not appropriate.

This commit udpates the definition of `UK_ARCH_CACHE_LINE_SIZE` to
use `__GCC_DESTRUCTIVE_SIZE`, a macro produced by GCC 12+ and Clang 15+. Its value
is derived from the `-march` and `-mpcu` compiler flags, allowing
`UK_ARCH_CACHE_LINE_SIZE` to reflect the target CPU's cache line size.

**Note:** `__GCC_DESTRUCTIVE_SIZE` reflects a conservative upper bound for
false sharing prevention rather than the exact cache line size. For generic
targets (e.g. `-march=armv8-a`), this may be larger than the actual hardware
value (e.g. 256 bytes on ARM64). For named CPU targets (e.g.
`-mcpu=cortex-a76`), the value is accurate. The fallback of 64 bytes is
preserved for compilers that do not define `__GCC_DESTRUCTIVE_SIZE`. 

Tested with preprocessor across relevant targets:
```
$ echo "#ifdef __GCC_DESTRUCTIVE_SIZE
#define UK_ARCH_CACHE_LINE_SIZE __GCC_DESTRUCTIVE_SIZE
#else
#define UK_ARCH_CACHE_LINE_SIZE 64
#endif
int x = UK_ARCH_CACHE_LINE_SIZE;" | aarch64-linux-gnu-gcc -mcpu=cortex-a76 -E -
  | grep "^int"
int x = 64;

$ echo "..." | aarch64-linux-gnu-gcc -march=armv8-a -E - | grep "^int"
int x = 256;

$ echo "..." | gcc -march=corei7 -E - | grep "^int"
int x = 64;
```

### Related Work

N/A

### PR Checklist

### DIFF
--- a/arch/include/uk/arch.h
+++ b/arch/include/uk/arch.h
@@ -13,10 +13,16 @@
 extern "C" {
 #endif
 
-/* FIXME: Hardcode for now. In practice this can be
- *        variable, define per platform instead.
+/**
+ * Use the compiler's knowledge of the target CPU cache line size if available.
+ * __GCC_DESTRUCTIVE_SIZE is set by GCC 12+ and Clang 15+ based on the
+ * -march/-mcpu flags. Falls back to previous hard-coded value of 64 bytes.
  */
+#ifdef __GCC_DESTRUCTIVE_SIZE
+#define UK_ARCH_CACHE_LINE_SIZE		__GCC_DESTRUCTIVE_SIZE
+#else
 #define UK_ARCH_CACHE_LINE_SIZE		64
+#endif
 
 #if !__ASSEMBLY__
 


michpappas:
Hi, thanks for the PR. Unfortunately I won't be able to merge as deriving the cache line size at compile time is a severely limited approach as it implies that execution happens on a system with the same specs as the build system, which in practice is seldomly the case. The proper way to address this issue requires querying the SoC for the cache-line size at runtime instead.

PR #1804, opened 2026-04-01
lib/posix-time: Provide an implementation for the times function

andreittr:
### Description of Changes

This change adds an implementation of the `times()` function/sycall, along with the corresponding header `<sys/times.h>` imported from musl.

Based on previous work by @jschlumpp .

### Related Work

N/A

### PR Checklist

### DIFF
--- a/lib/posix-time/time.c
+++ b/lib/posix-time/time.c
@@ -37,6 +37,7 @@
 #include <time.h>
 #include <unistd.h>
 #include <sys/time.h>
+#include <sys/times.h>
 #include <uk/plat/time.h>
 #include <uk/config.h>
 #include <uk/print.h>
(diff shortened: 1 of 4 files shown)


lib/posix-time/time.c
```diff
@@ -273,9 +274,20 @@ UK_SYSCALL_R_DEFINE(int, clock_nanosleep, clockid_t, clockid, int, flags,
...
+	/* We don't really spent time in the kernel ¯\_(ツ)_/¯ */
+	buf->tms_stime = 0;
+	/* Assumption: NSEC_PER_SEC is divisible by CLOCKS_PER_SEC */
+	divisor = UKARCH_NSEC_PER_SEC / CLOCKS_PER_SEC;
+	buf->tms_utime = ukplat_monotonic_clock() / divisor;
+	/* TODO: Add times of terminated waited-on children in the future */
+	buf->tms_cutime = 0;
+	buf->tms_cstime = 0;
+
+	return 0;
```
michpappas: Not sure why the return value is zero, as that would return nonsensical values for callers that want to calculate the delta between subsequent calls. According to times(2):
```
RETURN VALUE
       times() returns the number of clock ticks that have elapsed since an arbitrary point
       in the past. The return value may overflow the possible range of type  clock_t. On 
       error, (clock_t) -1 is returned, and errno is set appropriately.

...

NOTES
       ...
       On Linux, the "arbitrary point in the past" from which the return value of times()
       is measured has varied across kernel versions.  On Linux 2.4 and earlier, this point
       is the moment the system was booted.  Since Linux 2.6, this point is (2^32/HZ) - 300
       seconds before system boot time.  This variability across kernel versions (and across
       UNIX implementations), combined with the fact that the returned value may overflow
       the range of clock_t, means that a portable application would be wise to avoid using
       this value. To measure changes in elapsed time, use clock_gettime(2) instead.
```
Would it make more sense to return a monotinic value like `(clock_t)buf->tms_utime`?

michpappas (changes requested):

lib/posix-time/time.c
```diff
@@ -273,9 +274,20 @@ UK_SYSCALL_R_DEFINE(int, clock_nanosleep, clockid_t, clockid, int, flags,
...
 }
 
-UK_SYSCALL_R_DEFINE(int, times, struct tm *, buf)
+UK_SYSCALL_R_DEFINE(clock_t, times, struct tms *, buf)
 {
-	return -ENOTSUP;
+	uint64_t divisor;
+
+	/* We don't really spent time in the kernel ¯\_(ツ)_/¯ */
+	buf->tms_stime = 0;
```
michpappas: It would be worth to add an assertion for `buf`.

andreittr:
Pushed update addressing comments + minor refactor to bring `times()` in line with other syscalls & more faithfully implement the manpage.

michpappas (approved):

PR #1805, opened 2026-04-01
lib/posix-time: Provide implementation of CLOCK_THREAD_CPUTIME_ID

andreittr:
### Description of Changes

This change adds support for the `CLOCK_THREAD_CPUTIME_ID` clock type to `posix-time`.

Based on previous work by @jschlumpp .

### Related Work

N/A

### PR Checklist

### DIFF
--- a/lib/posix-time/time.c
+++ b/lib/posix-time/time.c
@@ -178,10 +178,14 @@ int uk_sys_clock_getres(clockid_t clk_id, struct timespec *tp)
 
 	switch (clk_id) {
 	case CLOCK_MONOTONIC:
+	case CLOCK_MONOTONIC_RAW:
 	case CLOCK_MONOTONIC_COARSE:
 	case CLOCK_REALTIME:
 	case CLOCK_REALTIME_COARSE:
 	case CLOCK_BOOTTIME:
+#if CONFIG_HAVE_SCHED
+	case CLOCK_THREAD_CPUTIME_ID:
+#endif /* CONFIG_HAVE_SCHED */
 		if (tp) {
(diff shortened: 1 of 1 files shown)


lib/posix-time/time.c
```diff
@@ -225,6 +225,9 @@ int uk_sys_clock_gettime(clockid_t clk_id, struct timespec *tp)
 	case CLOCK_REALTIME_COARSE:
 		now = ukplat_wall_clock();
 		break;
+	case CLOCK_THREAD_CPUTIME_ID:
```
michpappas, line 233: Should be conditional to `CONFIG_LIBUKSCHED` because of `uk_thread_current()`. Also `exec_time` is updated on schedule time, so we are missing the elapsed time since the current thread (caller) was scheduled. Perhaps offset with something like `+ (ukplat_monotonic_clock() - sched->ts_prev_switch)`.

andreittr: You're right, though unfortunately the previous switch timestamp is private to the [coop scheduler](https://github.com/unikraft/unikraft/blob/staging/lib/ukschedcoop/schedcoop.h#L18), and I found no equivalent to do this with the public scheduler API. A clean fix would be to expose prev switch time (or even "current scheduled runtime") as part of the uksched/ukthread API.

I could for now add a comment documenting this as a known issue; or postpone this feature until the required changes to uksched.

andreittr:
Pushed update addresssing comment + 1 trivial fix commit that adds missing clocks to `getres`.

michpappas (approved):
Thanks!

PR #1807, opened 2026-04-02
{plat, lib, drivers}: Fixups post-lcpu/pcpuvar rework

mogasergiu:
### Description of Changes

This series collects six minor fixups following the recent uklcpu/pcpuvar rework.

The comment for the lcpu pm start op is updated and the halting loop is fixed to correctly just halt. Before it would also implicitly disable IRQs prior to halting which was wrong and would completely kill any activity on the CPU.

Since Unikraft exclusively targets 64-bit platforms, `error_code` fields and parameters across exception structs are unified to __u64, replacing the inconsistent mix of unsigned long and int. Similarly done to `irq`, we resort to a more lower level and precise type, `__u32`. A strace build regression introduced by fb7c796534a2 — which updated the binary syscall handler but missed the strace path — is also fixed by applying the _UK_EXECENV_REGS_GET accessors consistently.

Finally, an indirect pointer _uk_pcpuvar_tmpl_size_ptr is introduced for accessing the template size, avoiding R_AARCH64_ADR_PREL_PG_HI21 truncation errors when the image is linked at a high virtual address (like is the case for Xen ARM64 PVH).

### Related Work

N/A

### PR Checklist

### DIFF
--- a/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
+++ b/plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
@@ -52,7 +52,7 @@ struct uk_plat_native_except_err_ctx {
 	/* x86 exception vector number (0-31) */
 	__u32 trapnr;
 	/* Hardware error code (only valid for #PF, #GP, #TS, etc.) */
-	int error_code;
+	__u64 error_code;
 	/* Set by handler if exception unrecoverable */
 	int handler_err;
 	/* CPU registers at exception time */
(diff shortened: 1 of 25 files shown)


mogasergiu:
Last 4 commits migrated from https://github.com/unikraft/unikraft/pull/1799

plat/native/arch/x86_64/include/uk/plat/native/arch/except.h
michpappas, line 110: The getter and setter also need to return `__u64`. I think also in Xen? (`plat/xen/arch/x86_64/include/uk/plat/xen/arch/except.h`, `plat/xen/pal/arch/x86_64/include/uk/plat/pal/arch/except.h`).

michpappas (changes requested):

michpappas (approved):

PR #1809, opened 2026-04-15
support: Generate compilation database from .cmd files

michpappas:
Replace the clang-specific `-MJ` flag approach for `compile_commands.json` generation with a compiler-agnostic method based on the already existing `.cmd` files. This enables the generation of a compile database when compiling with GCC.

### Related Work

#1751 

### PR Checklist

 - [x] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [x] Tested your changes against relevant architectures and platforms;
 - [x] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [ ] Updated relevant documentation.

PR #1811, opened 2026-04-17
.github/workflows: Introduce static-analysis workflow

michpappas:
### Description of Changes

Add a workflow that runs static analysis with CodeChecker. The workflow can be executed standalone, although it's main purpose to act as a CI gate for new findings on PRs. This very first implementation is limited to `app-elfloader` on `qemu/x86_64` built with Clang.

Identifier checks (`clang-diagnostic-reserved-identifier` / `clang-diagnostic-reserved-macro-identifier`) are disabled by the newly introduced `.codechecker.yml`, as unikraft uses the underscore prefix to indicate private scope.

Internally, once static analysis is complete the results are checked against those stored in the CodeChecker remote database for the same configuration. If new findings are introduced the job is failed and the results are shown in its logs as shown below:

image

Upon merge, the results are stored in the remote database resulting into an updated state.

### Related Work

#1751, #1809

### PR Checklist

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -1,0 +1,249 @@
+name: static-analysis
+
+on:
+  workflow_dispatch:
+
+  workflow_run:
+    workflows: [integration]
+    types: [completed]
+
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  statuses: write
+
+jobs:
+  analyze:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'schedule' ||
+      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Set commit status: pending"
+        if: ${{ github.event_name == 'workflow_run' }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api repos/${{ github.repository }}/statuses/${{ github.event.workflow_run.head_sha }} \
+            -f state=pending \
+            -f context="static analysis" \
+            -f description="Running static analysis..." \
+            -f target_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || 'staging' }}
+          submodules: recursive
+
+      - name: "Fetch clang"
+        uses: KyleMayes/install-llvm-action@v2
+        with:
+          arch: 'x64'
+          version: '21'
+
+      - name: "Fetch elfloader"
+        uses: actions/checkout@v4
+        with:
+          repository: unikraft/app-elfloader
+          fetch-depth: 1
+          path: _elfloader
+
+      - name: "Build elfloader"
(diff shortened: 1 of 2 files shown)


michpappas:
Current static analysis failure due to database server domain not registered. Please do not merge until resolved.

.github/workflows/static-analysis.yaml
```diff
@@ -0,0 +1,195 @@
+name: static-analysis
+
+on:
+  workflow_dispatch:
+
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches: [staging]
+    paths-ignore: ['*.md']
```
andreittr, line 9: Minor question: I see we are explicitly excluding docs changes from static analysis runs (which I competely agree with); do you think it's worth excluding more file types (e.g., `*.S`), or to trigger only on source files on which the static analyzer can run? Or would it be too much hassle for negligible benefit?

.github/workflows/static-analysis.yaml
```diff
@@ -0,0 +1,195 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: "Fetch clang"
+        uses: KyleMayes/install-llvm-action@v2
+        with:
+          arch: 'x64'
+          version: '21'
```
andreittr, line 23: Just confirming: since this is for static analysis, we would always want to use the newest clang/llvm that we support, and there would be no benefit to going back to an older version, right? So in principle this field would be periodically bumped once a new llvm comes out and we've confirmed it working.

andreittr (approved):
Looks good, I've only some clarifying questions.
(AB & merging blocked pending database server)

.github/workflows/static-analysis.yaml
michpappas: Yes for sure there is no hassle there. The only benefit in the current state where `app-elfloader` is only compiled as part of static analysis, is the implicit compile check. So maybe we can defer that optimization to when we have a separate build check for `app-elfloader`? But in principle I think the point is valid.

.github/workflows/static-analysis.yaml
michpappas: We need to put a bit of thought into this. In this PR I set the version to an arbitrarily recent on (currently clang is at 22 but I chose to be conservative), with the intention to mainly enable proper static analysis in our PRs. Once we have that, there are a few things to determine, such as whether we need to match the compiling version with the analyzer version, the compiler version of the kraft container (currently I am installing clang outside the container), the deployment compiler version, and possibly others. So consider this as a very first implementation that we should incrementally improve as we go.

michpappas:
Updated `static-analysis.yaml` to trigger on `workflow_run` after `integration` is complete, as on the `pull_request` trigger GH for PRs originating from remote forks (the norm in Unikraft) blocks access to secrets for security reasons.

Also added a status line in the PR using the GH API, as normally `workflow_run` triggers don't show on the PR's checks. 

Finally, the workflow now runs static analysis on `staging` and stores the results into the database periodically overnight, and also if triggered manually.

Notice that:
1. The custom check only appears once static analysis starts.
2. The static analysis check does not appear in this PR because it does not exist on `staging`.

@andreittr if this goes well we can later generalize and rename the static analysis workflow to build and analyze an arbitrary application, arch, plat matrix and add helloworld and possibly others.

michpappas:
Changes in last update: Display different message on fail and cancel.

andreittr (approved):

andreittr:

PR #1813, opened 2026-04-17
lib/posix-process: Do not run `brk` handler on non-`vfork` calls

mogasergiu:
Following commit bd191c39c196 ("lib/posix-process: Rework posix-process events"), clone handlers are now run on every clone syscall and is therefore their responsibility to adapt their behavior depending on clone flags.

If the clone invocation that lead to the call of the `brk` initialization does not have `CLONE_VFORK` set, then do not run the `brk` handler since we only want this to be run once, when a process is created through `vfork`/clone(..., CLONE_VFORK).

### Related Work

N/A

### PR Checklist


michpappas (approved):

PR #1855, opened 2026-06-30
lib/posix-tty: Fix stdio initcalls without ukfs devfs

### LINKED ISSUE #1838: lib/posix-tty: stdio initcalls are discarded when CONFIG_LIBUKFS_DEVFS=n
### Describe the bug

With `CONFIG_LIBPOSIX_TTY_SERIAL=y`, `CONFIG_LIBPOSIX_TTY_STDIO=y`, and
`CONFIG_LIBUKFS_DEVFS=n`, the POSIX TTY initialization entries are emitted into
non-numeric `.uk_inittab` sections and discarded by the final link.

As a result, fd 0, 1, and 2 are never initialized. Linux binary-compatibility
applications execute, but writes to stdout/stderr produce no output.

This currently affects the catalog `base:latest` runtime. For example,
`catalog/examples/helloworld-gcc13.2` boots and exits after the minimal
Unikraft banner without printing `Bye, World!`.

I reproduced this against the current `staging` sources on June 10, 2026.

### Steps to reproduce

Use a configuration containing:

```text
CONFIG_LIBPOSIX_TTY=y
CONFIG_LIBPOSIX_TTY_SERIAL=y
CONFIG_LIBPOSIX_TTY_STDIO=y
CONFIG_LIBPOSIX_TTY_STDIN_NULL=y
CONFIG_LIBPOSIX_TTY_STDOUT_SERIAL=y
# CONFIG_LIBUKFS_DEVFS is not set
```

Build the catalog `base` runtime and run
`catalog/examples/helloworld-gcc13.2` with it.

Observed guest output:

```text
Booting from ROM..Powered by Unikraft Ijiraq (0.21.0~87240a2)
```

The guest then exits without printing the application's output.

The problem reproduces with both KVM and QEMU TCG, and with a valid
approximately 2.1 MiB initramfs. A test ELF writing directly to COM1 prints
successfully, while a Linux `write(1, ...)` does not, isolating the failure to
standard fd initialization.

### Expected behavior

The application should print:

```text
Bye, World!
```

### Root cause

`lib/posix-tty/serial.c` includes `<uk/fs/prio.h>` only in the
`CONFIG_LIBUKFS_DEVFS` branch, but the non-DEVFS branch registers
`init_tty_cons` using `UK_FS_PRIO_FSAVAIL`:

```c
uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_FS_PRIO_FSAVAIL);
```

`lib/posix-tty/tty.c` has the same conditional include, while the serial stdio
path uses:

```c
UK_PRIO_AFTER(UK_PRIO_AFTER(UK_FS_PRIO_FSAVAIL))
```

When DEVFS is disabled, `UK_FS_PRIO_FSAVAIL` is undefined. Because the priority
is used to construct a section name, compilation succeeds but produces
sections such as:

```text
.uk_inittab4UK_FS_PRIO_FSAVAIL
.uk_inittab4__UK_PRIO_AFTER___UK_PRIO_AFTER_UK_FS_PRIO_FSAVAIL
```

The final linker script retains only numeric sections matching
`.uk_inittab[1-6][0-9]`. Therefore `init_tty_cons` and `init_posix_tty` are
garbage-collected.

This appears to be a regression from commit
`74555adf2583f3313250e4923d246d000cf77829`, which moved the non-DEVFS
console-file initialization from a library initcall to
`UK_FS_PRIO_FSAVAIL`.

### Verified workaround / proposed fix

Restore the pre-regression initialization classes when DEVFS is disabled:

```diff
--- a/lib/posix-tty/serial.c
+++ b/lib/posix-tty/serial.c
@@
-uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_FS_PRIO_FSAVAIL);
+uk_lib_initcall_prio(init_tty_cons, 0x0, UK_PRIO_LATEST);
```

Keep the filesystem priority for DEVFS builds, but use the normal rootfs
initcall otherwise:

```diff
 #if CONFIG_LIBPOSIX_TTY_SERIAL
+#if CONFIG_LIBPOSIX_TTY_DEVFS
 uk_rootfs_initcall_prio(init_posix_tty, 0x0,
                         UK_PRIO_AFTER(UK_PRIO_AFTER(UK_FS_PRIO_FSAVAIL)));
+#else
+uk_rootfs_initcall(init_posix_tty, 0x0);
+#endif
```

After applying this change, the generated sections are numeric:

```text
.uk_inittab39
.uk_inittab49
```

Both functions are retained in the final kernel:

```text
init_tty_cons
init_posix_tty
```

The original `helloworld-gcc13.2` rootfs then prints:

```text
Bye, World!
```

### Environment

```text
Architecture: x86_64
Platform: QEMU/KVM and QEMU/TCG
Unikraft: staging, reported runtime version 0.21.0~87240a2
Catalog commit: 87240a293c055da12b9d999da7f339cb223a627b
KraftKit: 0.12.13 (8f5e5dd34ac04c2ac0f43c6499548f2db238e845)
QEMU: 10.2.1
GCC: 15.2.0 (build performed with UK_CFLAGS=-std=gnu17)
Host: Ubuntu 26.04 under WSL2
```

I discovered this issue while working with GPT5.5 on Unikraft. Please let me know if there's anything wrong or unclear about this.

LetsWalkInLine:
### Description of Changes

When `CONFIG_LIBUKFS_DEVFS` is disabled, `lib/posix-tty` still uses
`UK_FS_PRIO_FSAVAIL` to register the serial console and standard I/O
initializers. The priority macro is not included in this configuration, so its
token is pasted into non-numeric `.uk_inittab` section names. The final link
does not retain those sections, leaving file descriptors 0, 1, and 2
uninitialized.

Keep both initializers in the rootfs init class for the non-DEVFS path.
Register the console file initializer immediately before the latest rootfs
priority and the standard I/O initializer at the latest priority. This ensures
that library initialization has completed while preserving the required order
between console file creation and opening the standard descriptors.

The DEVFS-specific initialization order remains unchanged.

Fixes #1838.

### Related Work

The regression was introduced by #1795.

### Testing

- Built the Catalog `base` configuration for QEMU/x86_64 with
  `CONFIG_LIBUKFS_DEVFS=n`.
- Verified numeric `.uk_inittab48` and `.uk_inittab49` sections and retained
  `init_tty_cons` and `init_posix_tty` symbols.
- Ran `helloworld-gcc13.2` under QEMU TCG and observed `Bye, World!`.
- Built the same configuration with `CONFIG_LIBUKFS_DEVFS=y` and verified
  numeric `.uk_inittab43` and `.uk_inittab44` sections.
- Ran `support/scripts/checkpatch.uk` against the commit: 0 errors, 0 warnings.

### PR Checklist

- [x] Read the contribution guidelines regarding submitting new changes to the
  project.
- [x] Tested the changes against the relevant architecture, platform, and
  configuration variants.
- [x] Ran `checkpatch.uk` on the commit series before opening this PR.
- [x] Updated relevant documentation (not applicable: no public API or
  configuration change).

### Note

I investigated this issue and prepared the fix in collaboration with OpenAI
Codex (GPT-5.5). I reviewed and tested the resulting change. I am still
learning Unikraft internals and the project's contribution workflow, so please
point out anything incorrect or suggest a more appropriate implementation or
testing approach.

### DIFF
--- a/lib/posix-tty/serial.c
+++ b/lib/posix-tty/serial.c
@@ -168,5 +168,6 @@ setup_tty_globals:
 	return 0;
 }
 
-uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_FS_PRIO_FSAVAIL);
+uk_rootfs_initcall_prio(init_tty_cons, 0x0,
+			UK_PRIO_BEFORE(UK_PRIO_LATEST));
 #endif /* !CONFIG_LIBUKFS_DEVFS */
--- a/lib/posix-tty/tty.c
+++ b/lib/posix-tty/tty.c
@@ -145,9 +145,13 @@ static int init_posix_tty(struct uk_init_ctx *ictx __unused)
 }
 
 #if CONFIG_LIBPOSIX_TTY_SERIAL
+#if CONFIG_LIBPOSIX_TTY_DEVFS
 /* Must come after the handler in serial.c */
 uk_rootfs_initcall_prio(init_posix_tty, 0x0,
 			UK_PRIO_AFTER(UK_PRIO_AFTER(UK_FS_PRIO_FSAVAIL)));
+#else /* !CONFIG_LIBPOSIX_TTY_DEVFS */
+uk_rootfs_initcall_prio(init_posix_tty, 0x0, UK_PRIO_LATEST);
+#endif /* !CONFIG_LIBPOSIX_TTY_DEVFS */
 #else /* !CONFIG_LIBPOSIX_TTY_SERIAL */
 #if CONFIG_LIBPOSIX_TTY_DEVFS
 uk_rootfs_initcall_prio(init_posix_tty, 0x0, UK_FS_PRIO_FSAVAIL);


lib/posix-tty/serial.c
```diff
@@ -168,5 +168,5 @@ static int init_tty_cons(struct uk_init_ctx *ictx __unused)
 	return 0;
 }
 
-uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_FS_PRIO_FSAVAIL);
+uk_lib_initcall_prio(init_tty_cons, 0x0, UK_PRIO_LATEST);
```
andreittr: Changing the init class from `rootfs-2` (what `UK_FS_PRIO_FSAVAIL` resolves to) to `lib-latest` does indeed ensure that this runs before the corresponding init func in `tty.c`, however it does not ensure that all `lib` init funcs have successfully finished (which the previous `rootfs` class ensured) and thus changes semantics in a possibly unwanted way.

I believe you could achieve the same effect by registering like
```
uk_rootfs_initcall_prio(init_tty_cons, 0x0, UK_PRIO_BEFORE(UK_PRIO_LATEST));
```

(with appropiate update to the commit message)

lib/posix-tty/tty.c
```diff
@@ -145,9 +145,13 @@ static int init_posix_tty(struct uk_init_ctx *ictx __unused)
 }
 
 #if CONFIG_LIBPOSIX_TTY_SERIAL
+#if CONFIG_LIBPOSIX_TTY_DEVFS
 /* Must come after the handler in serial.c */
 uk_rootfs_initcall_prio(init_posix_tty, 0x0,
 			UK_PRIO_AFTER(UK_PRIO_AFTER(UK_FS_PRIO_FSAVAIL)));
+#else /* !CONFIG_LIBPOSIX_TTY_DEVFS */
+uk_rootfs_initcall(init_posix_tty, 0x0);
```
andreittr: Minor nitpick (in context of previous comment): it would help to be explicit here that we are registering with LATEST prio:
```
uk_rootfs_initcall_prio(init_posix_tty, 0x0, UK_PRIO_LATEST);
```
(these are ofc equivalent, but being explicit makes reasoning about the code easier in the future)

andreittr (changes requested):
Thanks for catching this issue and submitting a fix! The logic is sound, but I have one subtle point I'd like you to address.

lib/posix-tty/serial.c
LetsWalkInLine: Thanks, that makes sense. Updated as suggested and adjusted the commit message accordingly.

lib/posix-tty/tty.c
LetsWalkInLine: Agreed. Updated to use the explicit `UK_PRIO_LATEST` priority.

andreittr (approved):
Thank you for the patch and changes!

andreittr:

PR #1865, opened 2026-07-09
lib/posix-futex: implement futex_requeue

kyungha-kim:
Tied to discussion [1844: FUTEX_REQUEUE](https://github.com/orgs/unikraft/discussions/1844)

### Description of Changes

- lib/posix-futex/futex.c:
    - Add support for FUTEX_REQUEUE operation which wakes up to val waiters on uaddr and requeues up to val2 remaining waiters to uaddr2.
    - Refactor futex_wait, futex_wake, and futex_cmp_requeue to use uk_spin_lock_irqsave/uk_spin_unlock_irqrestore instead of the separate uk_lcpu_save_irqf + uk_spin_lock pattern.

### Related Work

N/A

### PR Checklist

 - [✅] Read the [contribution guidelines](https://unikraft.org/docs/contributing/) regarding submitting new changes to the project;
 - [✅] Tested your changes against relevant architectures and platforms;
 - [✅] Ran the [`checkpatch.uk`](https://github.com/unikraft/unikraft/blob/staging/support/scripts/checkpatch.uk) on your commit series before opening this PR;
 - [✅] Updated relevant documentation.

### DIFF
--- a/lib/posix-futex/futex.c
+++ b/lib/posix-futex/futex.c
@@ -266,6 +266,72 @@ static int futex_cmp_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
 }
 
 /**
+ * Requeue waiters from uaddr to uaddr2.
+ *
+ * Requeue waiters on uaddr to uaddr2. Wakes up a maximum of val waiters that
+ * are waiting on the futex at uaddr.  If there are more than val waiters, then
+ * the remaining waiters are removed from the wait queue of the source futex at
+ * uaddr and added to the wait queue of the target futex at uaddr2. The val2
+ * argument specifies an upper limit on the number of waiters that are requeued
+ * to the futex at uaddr2.
+ *
+ * @param uaddr		Source futex user address
+ * @param val		Number of waiters to wake
+ * @param val2		Number of waiters to requeue (0-INT_MAX)
+ * @param uaddr2	Target futex user address
+ *
+ * @return
+ *	>=0: on success, the number of tasks woken;
+ *	<0: on error
+ */
+static int futex_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
+			     uint32_t *uaddr2)
+{
+	unsigned long irqf;
+	struct uk_list_head *itr, *tmp;
+	struct uk_futex *f;
+	int woken_uaddr1 = 0;
+	uint32_t waiters_uaddr2 = 0;
+
+	/* Wake up val waiters on uaddr */
+	if (val)
+		woken_uaddr1 = futex_wake(uaddr, val);
+
+	if (!val2)
+		return woken_uaddr1;
+
+	if (uaddr == uaddr2)
+		return woken_uaddr1;
+
+	uk_spin_lock_irqsave(&futex_list_lock, irqf);
+
+	/* Requeue val2 waiters on uaddr2 */
+	uk_list_for_each_safe(itr, tmp, &futex_list) {
+		f = uk_list_entry(itr, struct uk_futex, list_node);
+
+		if (f->uaddr == uaddr) {
+			/* Requeue thread to uaddr2.
+			 * Re-added node may be revisited during
+			 * iteration, but f->uaddr is now uaddr2
+			 * so it won't match again.
+			 * uaddr == uaddr2 case is guarded above.
+			 */
+			uk_list_del(&f->list_node);
+			f->uaddr = uaddr2;
+			uk_list_add_tail(&f->list_node, &futex_list);
+
+			/* Requeue at most val2 threads */
+			if (++waiters_uaddr2 >= val2)
+				break;
+		}
+	}
+
+	uk_spin_unlock_irqrestore(&futex_list_lock, irqf);
+
+	return woken_uaddr1;
+}
+
+/**
  * According to man pages, there exists no libc wrapper for futex
  *
  * @param uaddr		Source futex user address
@@ -332,8 +398,11 @@ UK_LLSYSCALL_R_DEFINE(int, futex, uint32_t *, uaddr, int, futex_op,
 		return futex_wake(uaddr, val);
 
 	case FUTEX_FD:
-	case FUTEX_REQUEUE:
 		return -ENOSYS;
+
+	case FUTEX_REQUEUE:
+		return futex_requeue(uaddr, val,
+				     (unsigned long)timeout, uaddr2);
 
 	case FUTEX_CMP_REQUEUE:
 		return futex_cmp_requeue(uaddr, val, (unsigned long)timeout,
--- a/lib/posix-futex/futex.c
+++ b/lib/posix-futex/futex.c
@@ -113,11 +113,9 @@ static int futex_wait(uint32_t *uaddr, uint32_t val, const __nsec *timeout)
 			val, uaddr);
 
 	/* Enqueue thread to wait list */
-	irqf = uk_lcpu_save_irqf();
-	uk_spin_lock(&futex_list_lock);
+	uk_spin_lock_irqsave(&futex_list_lock, irqf);
 	uk_list_add_tail(&f.list_node, &futex_list);
-	uk_spin_unlock(&futex_list_lock);
-	uk_lcpu_restore_irqf(irqf);
+	uk_spin_unlock_irqrestore(&futex_list_lock, irqf);
 
 	if (timeout) {
 		/* Block at most until `timeout` nanosecs */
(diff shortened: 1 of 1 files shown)


lib/posix-futex/futex.c
```diff
@@ -259,8 +251,73 @@ static int futex_cmp_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
...
+
+			/* Requeue at most val2 threads */
+			if (++waiters_uaddr2 >= val2)
+				break;
+		}
+	}
+
+	uk_spin_unlock_irqrestore(&futex_list_lock, irqf);
 
 	return woken_uaddr1 + waiters_uaddr2;
```
andreittr: `FUTEX_REQUEUE`, as opposed to `FUTEX_REQUEUE_CMP` only returns the [number of woken threads](https://www.man7.org/linux/man-pages/man2/FUTEX_REQUEUE.2const.html), not [woken + requeued](https://www.man7.org/linux/man-pages/man2/FUTEX_CMP_REQUEUE.2const.html) (I know, weirdly inconsistent API); we should return only `woken_uaddr1` here.

andreittr (changes requested):
Thanks for the sumission, looks good overall, only one small comment.

lib/posix-futex/futex.c
kyungha-kim: Fixed(`futex_requeue` now returns only `woken_uaddr1`). Thanks for catching this and for the man page reference. It was very helpful in clarifying that FUTEX_REQUEUE returns only the number of woken threads, unlike FUTEX_CMP_REQUEUE. Updated the original commit and force-pushed.

andreittr (approved):
Looks good, thanks!

lib/posix-futex/futex.c
```diff
@@ -259,8 +251,7 @@ static int futex_cmp_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
 		}
 	}
 
-	uk_spin_unlock(&futex_list_lock);
-	uk_lcpu_restore_irqf(irqf);
+	uk_spin_unlock_irqrestore(&futex_list_lock, irqf);
```
mogasergiu, line 254: Should we also update `pthread_exit_handler` with this pattern?

lib/posix-futex/futex.c
```diff
@@ -265,6 +265,72 @@ static int futex_cmp_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
...
+			     uint32_t *uaddr2)
+{
+	unsigned long irqf;
+	struct uk_list_head *itr, *tmp;
+	struct uk_futex *f;
+	int woken_uaddr1 = 0;
+	uint32_t waiters_uaddr2 = 0;
+
+	/* Wake up val waiters on uaddr */
+	if (val)
```
mogasergiu, line 297: This is an almost copy of `futex_cmp_requeue` but comparing the two there seem to be some differences that I believe are actual  bugs that `futex_cmp_requeue` has and `futex_requeue` doesn't:
- `futex_cmp_requeue` calls `futex_wake` unconditionally
- no `uaddr == uaddr2` guard

But also I think the newly added `futex_requeue` has another thing `futex_cmp_requeue` doesn't have: it returns `woken_uaddr1` instead of `woken_uaddr1 + waiters_uaddr2`. The spec says, for each:
```
FUTEX_REQUEUE       Returns the number of waiters that were woken up.
FUTEX_CMP_REQUEUE   Returns the total number of waiters that were woken up
                    or requeued to the futex for the futex word at uaddr2.
```
I ran a small test on Linux to see which one the kernel actually does:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdint.h>
#include <limits.h>
#include <pthread.h>
#include <sched.h>
#include <time.h>

#define FUTEX_WAIT            0
#define FUTEX_WAKE            1
#define FUTEX_REQUEUE         3
#define FUTEX_CMP_REQUEUE     4

/* raw syscall: we want the kernel's return value, not libc's -1/errno */
static long fx(volatile uint32_t *uaddr, long op, long val, const void *timeout,
             volatile uint32_t *uaddr2, long val3)
{
      long r;
      register long r10 __asm__("r10") = (long)timeout;
      register long r8  __asm__("r8")  = (long)uaddr2;
      register long r9  __asm__("r9")  = val3;

      __asm__ volatile ("syscall" : "=a"(r)
                        : "a"(202L), "D"(uaddr), "S"(op), "d"(val),
                          "r"(r10), "r"(r8), "r"(r9)
                        : "rcx", "r11", "memory");
      return r;
}

static volatile uint32_t src, dst;    /* the two futex words */
static volatile int ready;

static void *waiter(void *unused __attribute__((unused)))
{
      __atomic_fetch_add(&ready, 1, __ATOMIC_SEQ_CST);
      fx(&src, FUTEX_WAIT, 0, NULL, NULL, 0);
      return NULL;
}

static void run(const char *name, long op, long doc)
{
      struct timespec ms200 = { 0, 200000000 };
      pthread_t th[4];
      long ret, on_dst, on_src;
      int i;

      src = dst = 0; ready = 0;
      for (i = 0; i < 4; i++)         /* four threads all blocked on src */
              pthread_create(&th[i], NULL, waiter, NULL);
      while (ready < 4)
              sched_yield();
      nanosleep(&ms200, NULL);                /* let all four block */

      /* nr_wake = 1, nr_requeue = 2, src -> dst, expected *src == 0 */
      ret = fx(&src, op, 1, (void *)2UL, &dst, 0);

      /* drain both futex words: this is what actually happened, independent
       * of whatever the op claims to have returned
       */
      on_dst = fx(&dst, FUTEX_WAKE, INT_MAX, NULL, NULL, 0);
      on_src = fx(&src, FUTEX_WAKE, INT_MAX, NULL, NULL, 0);

      printf("%-18s returned %ld | futex(2) says %ld | really: woke 1, "
             "requeued %ld, %ld still on src\n", name, ret, doc, on_dst, on_src);

      for (i = 0; i < 4; i++)
              pthread_join(th[i], NULL);
}

int main(void)
{
      setvbuf(stdout, NULL, _IONBF, 0);
      run("FUTEX_REQUEUE",     FUTEX_REQUEUE,     1); /* "number woken" */
      run("FUTEX_CMP_REQUEUE", FUTEX_CMP_REQUEUE, 3); /* "woken or requeued" */
      return 0;
}
```

Four threads block on `src`, then we ask for one wake and two requeues onto `dst`.
The two `FUTEX_WAKE` drains afterwards are the control: they show where the waiters
really ended up, regardless of what the op reported.

On Linux 7.2.5:

```
FUTEX_REQUEUE      returned 3 | futex(2) says 1 | really: woke 1, requeued 2, 1 still on src
FUTEX_CMP_REQUEUE  returned 3 | futex(2) says 3 | really: woke 1, requeued 2, 1 still on src
```

And the same binary (built with `musl-gcc -O1 -fPIE -static-pie`) on this branch
under app-elfloader:

```
FUTEX_REQUEUE      returned 1 | futex(2) says 1 | really: woke 1, requeued 2, 1 still on src
FUTEX_CMP_REQUEUE  returned 3 | futex(2) says 3 | really: woke 1, requeued 2, 1 still on src
```
So there is a discrepancy between kernel and manual semantics? I realised a bit later after testing that @andreittr has already mentioned this and to go according to the manual, but I am not sure if he was aware of the discrepancy between what the manual says and what the kernel actually does.

Either way, maybe we could avoid keeping two near-identical copies around? They've
already drifted apart in the ways above and I only noticed by putting them side by
side. Would something like this work as one implementation that both opcodes call
into?

```c
static int futex_requeue(uint32_t *uaddr, uint32_t val, uint32_t val2,
                         uint32_t *uaddr2, const uint32_t *expected,
                         uint32_t *requeued_out);
```

- `expected` is the value we expect to find at `uaddr`, i.e. what `val3` is compared
  against today in `if (!((uint32_t)val3 == uk_load_n(uaddr))) return -EAGAIN;`.
  Pass `NULL` to skip that comparison entirely. It has to be a pointer rather than a
  plain `uint32_t` since any `uint32_t` is a valid futex word, so there's no leftover
  value we could use to mean "don't compare". `FUTEX_REQUEUE` passes `NULL`,
  `FUTEX_CMP_REQUEUE` passes `&val3`.
- `requeued_out` is an out param for the requeued count, which right now gets counted
  into `waiters_uaddr2` and then thrown away. `NULL` if the caller doesn't need it.
- the return is just the number woken, or a negative errno. That way the `if (val)`
  guard around `futex_wake()` and the `uaddr == uaddr2` guard live in one place and
  cover both opcodes.

Then each case adds up whatever it wants to report, right next to the opcode name
that documents it:

```c
case FUTEX_REQUEUE:
        return futex_requeue(uaddr, val, ..., uaddr2, NULL, NULL);

case FUTEX_CMP_REQUEUE:
        woken = futex_requeue(uaddr, val, ..., uaddr2, &val3, &requeued);
        return woken < 0 ? woken : woken + requeued;
```

That also keeps the kernel vs manual question at the call site instead of hidden in
the helper. If we decide to match the kernel and return `woken + requeued` for both,
it's a one line change in the `FUTEX_REQUEUE` case. No strong preference from me,
I'd just like it to be a conscious pick with a line in the commit message saying why. Though I think it would be better to be aligned with the actual kernel runtime than a line in the manual.

mogasergiu (changes requested):

PR #1877, opened 2026-07-26
lib/posix-sysinfo: Handle _SC_CLK_TCK in sysconf()

RaduAndreiTudorica:
sysconf() did not handle _SC_CLK_TCK and fell through to the default return value of 0. CPython 3.13 queries this value while initializing the posix module and fails fatally when it is not positive, which prevents any Python application from starting:

  RuntimeError: cannot read ticks_per_second
  Fatal Python error: _PyImport_InitExternal: external importer
  setup failed

Return 100, which is the value musl reports for _SC_CLK_TCK on Linux.

### Description of Changes

`sysconf()` in `lib/posix-sysinfo` handles a small set of names and falls
through to a default return value of 0 for everything else. `_SC_CLK_TCK`
was not among the handled names, so it returned 0.

CPython 3.13 reads `_SC_CLK_TCK` while initializing the `posix` module and
treats a non-positive result as a fatal error. As a result, no Python
application can start:

```text
RuntimeError: cannot read ticks_per_second
Fatal Python error: _PyImport_InitExternal: external importer setup failed
```

This adds a case for `_SC_CLK_TCK` returning 100, matching the value musl
reports on Linux. The change is two lines and does not affect any other
`sysconf()` name.

I found this while upgrading `lib-python3` to 3.13 as part of GSoC'26.
Reproduced with `catalog-core/python3-hello` on `qemu/x86_64`: before the
change the unikernel boots but Python aborts during interpreter startup,
after it `hello.py` runs and prints its output.

If deriving the value from a configuration option is preferred over a
constant, I'm happy to rework it.

### Related Work

N/A

### PR Checklist

### DIFF
--- a/lib/posix-sysinfo/sysinfo.c
+++ b/lib/posix-sysinfo/sysinfo.c
@@ -36,6 +36,7 @@
 #include <errno.h>
 #include <string.h>
 #include <sys/utsname.h>
+#include <time.h>
 #include <uk/essentials.h>
 #include <uk/config.h>
 #include <sys/sysinfo.h>
@@ -125,6 +126,8 @@ long sysconf(int name)
 
 	if (name == _SC_PAGESIZE)
 		return __PAGE_SIZE;
+	if (name == _SC_CLK_TCK)
+		return CLOCKS_PER_SEC;
 
 #ifdef CONFIG_LIBPOSIX_USER
 	if (name == _SC_GETPW_R_SIZE_MAX)


StefanJum:
Does musl return a hardcoded value? Couldn't we return CONFIG_HZ?

RaduAndreiTudorica:
> Does musl return a hardcoded value? Couldn't we return CONFIG_HZ?

musl hardcodes it `[_SC_CLK_TCK] = 100` in `src/conf/sysconf.c`. On Linux
the value is USER_HZ, which is deliberately fixed at 100 regardless of the
kernel's `CONFIG_HZ`, so the userspace ABI doesn't shift when the timer
frequency changes. That's why I copied it.

But neither value is right. `uk_sys_times()` in `lib/posix-time/time.c` computes ticks
as `ukplat_monotonic_clock() / (UKARCH_NSEC_PER_SEC / CLOCKS_PER_SEC)`,
and `CLOCKS_PER_SEC` is `1000000L`. So `times()` reports microseconds,
and returning 100 from `sysconf()` would make anything dividing one by the
other off by a factor of 10000.

Reworking it to return `CLOCKS_PER_SEC` so the two agree. Thanks for
catching it! Python only checks that the value is positive at startup, so
this wouldn't have shown up until something actually called `os.times()`.

andreittr (approved):

andreittr:

NgoHuy:
`_SC_CLK_TCK` is undefined. Build process will fail. This should be back port in unistd.h.

vTusharr:
```
 W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c: In function ‘sysconf’:                                                                      
 W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c:129:21: error: ‘_SC_CLK_TCK’ undeclared (first use in this function)                         
 W    129 |         if (name == _SC_CLK_TCK)         
 W        |                     ^~~~~~~~~~~          
 W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c:129:21: note: each undeclared identifier is reported only once for each function it appears i
 W  make[2]: *** [/home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/support/build/Makefile.build:27: /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikr
 W  make[1]: *** [Makefile:1175: sub-make] Error 2   
 W  make: *** [Makefile:41: _all] Error 2            

 E  could not complete build: build failed: exit status 2

```

As pointed by @NgoHuy

NgoHuy:
> ```
>  W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c: In function ‘sysconf’:                                                                      
>  W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c:129:21: error: ‘_SC_CLK_TCK’ undeclared (first use in this function)                         
>  W    129 |         if (name == _SC_CLK_TCK)         
>  W        |                     ^~~~~~~~~~~          
>  W  /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/lib/posix-sysinfo/sysinfo.c:129:21: note: each undeclared identifier is reported only once for each function it appears i
>  W  make[2]: *** [/home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikraft/unikraft/support/build/Makefile.build:27: /home/unikraft/andreip/reviews/lighttpd-review/library/lighttpd/1.4/.unikr
>  W  make[1]: *** [Makefile:1175: sub-make] Error 2   
>  W  make: *** [Makefile:41: _all] Error 2            
> 
>  E  could not complete build: build failed: exit status 2
> ```
> 
> As pointed by @NgoHuy

You need to add `#define _SC_CLK_TCK 100` in unistd.h to make this build.

RaduAndreiTudorica:
Thanks both, you are right that the build breaks with nolibc. I only tested against ports that build against musl (Python, Redis, Nginx), where unistd.h already defines the constant, so I did not hit this.

One thing before I push the fix: I think the value should be 2 rather than 100. _SC_CLK_TCK is the name passed to sysconf(), while 100 is what sysconf() returns for it. lib/nolibc/include/unistd.h already follows the Linux/glibc numbering for the names it defines (_SC_OPEN_MAX 4, _SC_PAGE_SIZE 30, _SC_GETPW_R_SIZE_MAX 70, _SC_NPROCESSORS_CONF 83, _SC_PHYS_PAGES 85), and in both glibc and musl _SC_CLK_TCK is 2. In that numbering 100 is already _SC_XOPEN_XPG4, so defining it that way would make nolibc disagree with musl on the same constant.

So the follow-up would be #define _SC_CLK_TCK 2 in lib/nolibc/include/unistd.h, with sysconf() still returning CLOCKS_PER_SEC as in this PR. Does that match what you had in mind? I will test with a nolibc application this time.

RaduAndreiTudorica:
Pushed as a separate commit. Tested with catalog-core's `c-hello`, which
builds against nolibc, with `LIBPOSIX_SYSINFO=y`: it builds and boots fine
with the define, and without it fails exactly as in your log `sysinfo.c:129:21: error: '_SC_CLK_TCK' undeclared`.

Thanks @NgoHuy for spotting it.

NgoHuy:
new commit is not included in this PR, please create new PR.

Issue #1882, opened 2026-08-06
boot fails with pl011 console


garethodickin:
### Describe the bug

On Xen/ARM64 dom0less boot, the unikernel crashes immediately with repeated data abort traps before any console output appears.

Root causes:
  1. In dom0less environment, HYPERVISOR_start_info is not initialized (remains 0x0). The xenconsole driver attempts to write pointer derived from start_info, which is NULL, triggering a data abort trap on every uk_pr_* / printk call.

  2. The PL011 UART driver uses the physical/IPA address from the FDT reg property directly as a virtual address (dev->base = reg_base). Under Xen, guest IPA is not identity-mapped to VA, so UART register accesses hit an incorrect virtual address.

### Steps to reproduce

1. Build a Unikraft application with CONFIG_PLAT_XEN and CONFIG_LIBXENCONS enabled, CONFIG_LIBUKCONSOLE_PL011 disabled on AArch64
2. Boot the unikernel as a Xen dom0less domain (not dom0)
3. Observe repeated trap messages from xenconsole accessing NULL start_info
4. Disable CONFIG_LIBXENCONS, enable CONFIG_LIBUKCONSOLE_PL011 and rebuild
5. Boot again as dom0less domain
6. Observe data abort on first UART register access due to IPA being used directly as VA without fixmap mapping

### Expected behavior

(d1) - Unikraft booting -
(d1) - Setup CPU -
(d1) - Setup booting pagetable -
(d1) - MMU on -
(d1) - Setup stack -
(d1) - Jumping to C entry -
(XEN) DOM1: Powered by
(XEN) DOM1: o.   .o       _ _               __ _
(XEN) DOM1: Oo   Oo  ___ (_) | __ __  __ _ ' _) :_
(XEN) DOM1: oOo oOO| | | | |   (| | | (_) |  _) :_
(XEN) DOM1: oO   oO ' _ `| | |/ /  _)' _` | |_|  _)
(XEN) DOM1:  OoOoO ._, ._:_:_,\_._,  .__,_:_, \___)
(XEN) DOM1:                   Ijiraq 0.21.0~e34d9ba
(XEN) DOM1: Hello world

### Which architectures were you using or does this bug affect?

arm64

### Which platforms were you using or does this bug affect?

xen

### Relevant log output

```shell
[...]

PR #1885, opened 2026-08-13
plat/xen, drivers/pl011: Fix PL011 UART mapping on Xen/arm64

kang-yoongu:
On Xen/ARM64 dom0less boot, the unikernel crashes immediately with repeated data abort traps before any console output appears.

Root causes:

* In dom0less environment, HYPERVISOR_start_info is not initialized (remains 0x0). The xenconsole driver attempts to write pointer derived from start_info, which is NULL, triggering a data abort trap on every uk_pr_* / printk call.

* The PL011 UART driver uses the physical/IPA address from the FDT reg property directly as a virtual address (dev->base = reg_base). Under Xen, guest IPA is not identity-mapped to VA, so UART register accesses hit an incorrect virtual address.

### Description of Changes

This series fixes PL011 UART initialization on Xen/arm64 dom0less guests.

On Xen, IPA cannot be used directly as a VA. The PL011 driver was using reg_base (IPA) directly as dev->base, causing a fault when accessing UART MMIO.

Patch 1 adds arm,sbsa-uart to the compatible string list so the PL011 driver can match the virtual PL011 node inserted by Xen in the guest DTB.

Patch 2 maps the UART MMIO through the fixmap page table when running on Xen, consistent with how other devices are mapped

### Related Work
Issue : https://github.com/unikraft/unikraft/issues/1882

### PR Checklist

### DIFF
--- a/drivers/ukconsole/pl011/pl011.c
+++ b/drivers/ukconsole/pl011/pl011.c
@@ -16,6 +16,9 @@
 #include <uk/console/driver.h>
 #include <uk/compiler.h>
 #include <uk/errptr.h>
+#if CONFIG_PLAT_XEN
+#include <xen-arm/mm.h>
+#endif /* CONFIG_PLAT_XEN */
 
 #if CONFIG_LIBUKALLOC
 #include <uk/alloc.h>
@@ -284,7 +287,13 @@ static int fdt_get_device(struct pl011_device *dev, const void *dtb,
 
 	uk_console_init(&dev->dev, "PL011", &pl011_ops, 0,
 			UK_CONSOLE_CLASS_UART);
+#if CONFIG_PLAT_XEN
+	set_pgt_entry(&fixmap_pgtable[l2_pgt_idx(FIX_PL011_START)],
+		      ((reg_base & L2_MASK) | BLOCK_DEV_ATTR | L2_BLOCK));
+	dev->base = FIX_PL011_START + (reg_base & L2_OFFSET);
+#else
 	dev->base = reg_base;
+#endif /* CONFIG_PLAT_XEN */
 	dev->size = reg_size;
 
 	return 0;
--- a/drivers/ukconsole/pl011/pl011.c
+++ b/drivers/ukconsole/pl011/pl011.c
@@ -70,7 +70,7 @@
 #define REG_UARTICR_OFFSET	0x44
 
 static const char * const fdt_compatible[] = {
-	"arm,pl011", NULL
+	"arm,pl011", "arm,sbsa-uart", NULL
 };
 
 struct pl011_device {
--- a/plat/xen/include/xen-arm/mm.h
+++ b/plat/xen/include/xen-arm/mm.h
@@ -66,6 +66,7 @@ extern paddr_t _libxenplat_paddr_offset;
 
 #define SZ_2M           0x00200000
 
+#define FIXMAP_ENTRIES  (PAGE_SIZE / sizeof(lpae_t))
 #define PAGE_OFFSET     ((0xffffffffffffffff << (VA_BITS - 1))\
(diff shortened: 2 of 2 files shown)


drivers/ukconsole/pl011/pl011.c
```diff
@@ -70,7 +70,7 @@
 #define REG_UARTICR_OFFSET	0x44
 
 static const char * const fdt_compatible[] = {
-	"arm,pl011", NULL
+	"arm,pl011", "arm,sbsa-uart", NULL
```
mogasergiu: I notice that the vpl011 implementation in Xen does not implement some registers that the current probe method touches (`UARTCR`/`UARTLCR_H`) which I think generates some error messages in the log? If we had a proper driver model then we could have register a separate probe method with separate compatible list but unfortunately we don't. If this does not bother you I guess it is fine, but if yes, then maybe we can add a check if the device is SBSA and if it is then we set a boolean to skip the register writes after `REG_UARTICR_OFFSET` in `pl011_setup`.

drivers/ukconsole/pl011/pl011.c
mogasergiu: If early console is enabled, will this break since it's comparing IPA with a fixmap mapping?

drivers/ukconsole/pl011/pl011.c
```diff
@@ -284,7 +287,13 @@ static int fdt_get_device(struct pl011_device *dev, const void *dtb,
 
 	uk_console_init(&dev->dev, "PL011", &pl011_ops, 0,
 			UK_CONSOLE_CLASS_UART);
+#if CONFIG_PLAT_XEN
+	set_pgt_entry(&fixmap_pgtable[l2_pgt_idx(FIX_PL011_START)],
```
mogasergiu: What if you have two PL011's?

drivers/ukconsole/pl011/pl011.c
```diff
@@ -16,6 +16,9 @@
 #include <uk/console/driver.h>
 #include <uk/compiler.h>
 #include <uk/errptr.h>
+#if CONFIG_PLAT_XEN
+#include <xen-arm/mm.h>
```
mogasergiu: Not sure about page-table writes in a generic driver 🤔 . `#include <xen-arm/mm.h>` into `drivers/ukconsole/` drags ~60 unnamespaced macros (`PAGE_SIZE`, `PAGE_ALIGN`, `SCTLR_*`, `to_virt`…) into the TU... wondering if we could somehow use `uk_bus_pf_devmap` instead which could also help us get rid of the `CONFIG_PLAT_XEN` guards.

plat/xen/include/xen-arm/mm.h
```diff
@@ -215,6 +218,7 @@ extern paddr_t _libxenplat_paddr_offset;
 #define map_frames(f, n) (NULL)
 
 #ifndef __ASSEMBLY__
+extern lpae_t fixmap_pgtable[FIXMAP_ENTRIES];
```
mogasergiu: I think you can now also delete the externs in `plat/xen/arm/mm.c`, `plat/xen/arm/setup64.c` and `drivers/xen/xengnttab/arch/arm64/gnttab.c`?

drivers/ukconsole/pl011/pl011.c
```diff
@@ -16,6 +16,9 @@
 #include <uk/console/driver.h>
```
mogasergiu: The first sentence of the commit message sounds a bit vague. How about (and correct me if i am wrong, not as experienced on xen arm64 😄 ): "On xen/arm64 the boot page tables map only the kernel image, the RAM direct map and the fixmap window, so device MMIO has no VA". My understanding is that we do have an identity mapping somewhere in the lower half/TTBR0.

PR #1894, opened 2026-08-26
plat/xen: Add Xen static shared memory

kang-taekyung:
Tied to discussion [Xen Static Shared Memory #1892](https://github.com/orgs/unikraft/discussions/1892)

### Description of Changes

-   Introduce support for Xen static shared memory on ARM64.
    Shared memory regions declared in the device tree with
    compatible "xen,shared-memory-v1" are parsed at boot time
    and mapped into a dedicated virtual address range using
    pre-allocated L3 page tables.

 -  A new API uk_shm_map() is provided to obtain a virtual
    address for a sub-range of a shared memory region and to
    control read/write permissions at page granularity.

### Related Work

N/A

### PR Checklist


plat/xen/Config.uk
```diff
@@ -74,4 +74,33 @@ config XEN_PV_BUILD_P2M
 		Create and initialize physical to machine (p2m) table on a PV
 		xen host
 
+config XEN_STATIC_SHM
+	bool "Xen static shared memory"
+	default n
```
michpappas: `default n` is redundant, as Kconfig boolean options default to `n`.

plat/xen/arm/setup64.c
```diff
@@ -150,6 +166,83 @@ void get_xenbus(void)
...
+
+	node = -1;
+	while (1) {
+		node = fdt_node_offset_by_compatible(dtb, node,
+						     "xen,shared-memory-v1");
+		if (node < 0)
+			break;
+
+		parent = fdt_parent_offset(dtb, node);
+		if (parent < 0) {
```
michpappas: Here and below, it's a good idea to wrap the error path with `unlikely()`.

plat/xen/arm/setup64.c
```diff
@@ -150,6 +166,83 @@ void get_xenbus(void)
...
+			return;
+		}
+
+		vaddr = create_mapping(paddr, paddr + size,
+				       BLOCK_DEF_ATTR | ATTR_AP(ATTR_AP_RO)
+				       | ATTR_XN, &shm_pool);
+		if (!vaddr) {
+			uk_pr_err("%s: create_mapping failed pa=0x%lx\n",
+				  __func__, (unsigned long)paddr);
+			return;
```
michpappas: The enclosing `while()` loop returns on failure, abandoning any mappings created by that point. You may consider jumping to a label that implements rollback.

plat/xen/arm/setup64.c
```diff
@@ -150,6 +166,83 @@ void get_xenbus(void)
...
+				  __func__, id, UK_SHM_ID_MAXLEN);
+			continue;
+		}
+
+		if (shm_count >= UK_SHM_MAX_ENTRIES) {
+			uk_pr_err("%s: too many shm entries\n", __func__);
+			return;
+		}
+
+		vaddr = create_mapping(paddr, paddr + size,
```
michpappas: There is no check in this function for the value of `xen,offset` so once Xen imlements it the implementation will fail. Proposed solutions in order of preference (1) Implement `xen,offset` properly (2) Add a check against zero and fail on a non-zero value.

plat/xen/arm/setup64.c
```diff
@@ -150,6 +166,83 @@ void get_xenbus(void)
...
+			break;
+
+		parent = fdt_parent_offset(dtb, node);
+		if (parent < 0) {
+			uk_pr_err("%s: cannot find parent node\n", __func__);
+			continue;
+		}
+		naddr = fdt_address_cells(dtb, parent);
+		nsz = fdt_size_cells(dtb, parent);
+		if (naddr != 2 || nsz != 2) {
```
michpappas: Is there any reason for the hard requirement of 2 cells? In theory you should fetch the number of cells from the parent and adapt the parsing code to use these values. Generally speaking a sanity check is always a good idea, but one that makes sure that the value is between 0 and 2 makes more sense than excluding a cell size of 1.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+	} while (vaddr < vend);
+}
+
+unsigned long create_mapping(paddr_t start_paddr, paddr_t end_paddr,
+			     __u64 attr, struct pgtable_pool *pool)
+{
+	unsigned long start_vaddr, end_vaddr;
+	lpae_t *pgd;
+
+	if (pool->next_va >= pool->top_va) {
```
michpappas: The error path should be wrapped around  `unlikely()` defined in `uk/compiler.h`.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+		uk_pr_err("[%s] no VA space available\n", __func__);
+		return 0;
+	}
+
+	start_vaddr = pool->next_va;
```
michpappas: Although there is some contract between the caller and the implementation that pool regions must be aligned, it wouldn't hurt adding a calls to `UK_ASSERT()` just to be safe.

Also it's worth checking that the mapping length is non-zero, otherwise you unnecessarily create entries on L3.

plat/xen/Config.uk
```diff
@@ -74,4 +74,33 @@ config XEN_PV_BUILD_P2M
...
+	depends on PLAT_XEN_ARM64
+	help
+		Enable Xen static shared memory regions.
+		When enabled, shared memory regions declared in the device tree
+		(compatible = "xen,shared-memory-v1") are mapped at boot time
+		and accessible via the uk_shm_map() API.
+
+config XEN_SHM_MAX_ENTRIES
```
michpappas: There's nothing wrong with `depends on` here, but ust for uniformity with the rest of the code base, you can wrap `XEN_SHM_MAX_ENTRIES` and `XEN_SHM_REGION_SIZE_MB` around an `if XEN_STATIC_SHM` instead.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+
+static void alloc_init_pmd(lpae_t *pgd, unsigned long vaddr,
+			   unsigned long vend, paddr_t phys, __u64 attr,
+			   struct pgtable_pool *pool)
+{
+	lpae_t *pmd;
+	unsigned long next_va;
+	unsigned long l2_pa;
+
+	if (!(*pgd)) {
```
michpappas: Besides checking that the entry exists, you should also check that the entry points to a table. Otherwise you risk overwriting a block entry.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+		alloc_init_pte(pmd, vaddr, next_va, phys,
+			       attr, pool);
+
+		phys += (next_va - vaddr);
+		vaddr = next_va;
+		pmd++;
+	} while (vaddr < vend);
+}
+
+unsigned long create_mapping(paddr_t start_paddr, paddr_t end_paddr,
```
michpappas: A general comment about the types: You should prefer kernel types defined in `include/uk/types.h`. That is `__paddr_t`, `__vaddr_t`, `__size_t` should be used instead fo `paddr_t`, `vaddrt_t`, `size_t` and so on. Also I noticed in some occasions VAs use `unsigned long`, you can use `__vaddr_t` here too.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+			   struct pgtable_pool *pool)
+{
+	lpae_t *pte;
+	int l3_idx;
+	unsigned long next_va;
+	unsigned long pte_pa;
+
+	l3_idx = (vaddr - pool->base_va) >> L2_SHIFT;
+
+	if (!*pmd) {
```
michpappas: Besides checking that the entry exists, you should also check that the entry points to a table. Otherwise you risk overwriting a block entry.

plat/xen/arm/mm.c
```diff
@@ -100,6 +100,92 @@ static void build_pagetable(unsigned long start_pfn, unsigned long max_pfn)
...
+		set_pgt_entry(pmd, (pte_pa & (~ATTR_MASK_L)) | L2_TABLE);
+	}
+
+	do {
+		next_va = pte_addr_end(vaddr, vend);
+
+		l3_idx = (vaddr - pool->base_va) >> L2_SHIFT;
+		pte = &pool->l3_tables[l3_idx][l3_pgt_idx(vaddr)];
+
+		set_pgt_entry(pte, (phys & (~ATTR_MASK_L)) | attr | L3_PAGE);
```
michpappas: It's probably a good idea to check if the entry already exists - and fail if so.

plat/xen/arm/setup64.c
```diff
@@ -150,6 +166,83 @@ void get_xenbus(void)
...
+			continue;
+		}
+
+		if (shm_count >= UK_SHM_MAX_ENTRIES) {
+			uk_pr_err("%s: too many shm entries\n", __func__);
+			return;
+		}
+
+		vaddr = create_mapping(paddr, paddr + size,
+				       BLOCK_DEF_ATTR | ATTR_AP(ATTR_AP_RO)
```
michpappas: `BLOCK_DEF_ATTR` technically works, but is a bit misleading given that you're mapping at the page level. You can amend that by adding a `#define PAGE_DEF_ATTR    BLOCK_DEV_ATTR` in `include/xen-arm/mm.h`.

plat/xen/include/xen-arm/mm.h
```diff
@@ -217,6 +239,44 @@ extern paddr_t _libxenplat_paddr_offset;
...
+
+/*
+ * Pool-based page table mapping functions.
+ * Used for mapping reserved memory regions (e.g. Xen static shared memory)
+ * with pre-allocated L2/L3 page tables.
+ */
+unsigned long create_mapping(paddr_t start_paddr, paddr_t end_paddr,
+			     __u64 attr, struct pgtable_pool *pool);
+
+#ifdef CONFIG_XEN_STATIC_SHM
```
michpappas: The convention we use is:
```
#if CONFIG_...
...
#endif /* CONFIG_... */
```

or
```
#if CONFIG_...
...
#else /* !CONFIG_...
...
#endif /* !CONFIG_... */
```
`plat/xen/` is arguably not the most compliant part of the codebase :)

plat/xen/include/xen-arm/shm.h
```diff
@@ -0,0 +1,51 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+#ifndef __XEN_ARM_SHM_H__
+#define __XEN_ARM_SHM_H__
+
+#include <uk/essentials.h>
+#include <uk/arch/types.h>
+
+#define UK_SHM_RW 0x0
```
michpappas: Maybe we shouldn't reserve `UK_SHM_` / `uk_shm_` for Xen's static shm, as it's quite a generic name. Prefer `UK_PLAT_XEN_SHM_` / `uk_plat_xen_shm_` instead. I'm aware identifiers can occasionally get pretty long, but it's something we need to deal with given Unikraft's nature.

plat/xen/arm/shm.c
```diff
@@ -0,0 +1,101 @@
...
+#include <stdint.h>
+#include <string.h>
+#include <xen-arm/shm.h>
+#include <xen-arm/mm.h>
+#include <uk/print.h>
+#include <uk/spinlock.h>
+
+static struct uk_spinlock shm_lock = UK_SPINLOCK_INITIALIZER();
+
+static inline void tlb_flush_va(unsigned long vaddr)
```
michpappas: Let's avoid code duplication. This one should instead should be taken from Xen's PAL implementation. Right now it's nop but you can create per-architecture variants, and on arm64 that wrap around `plat/native`'s helper.

plat/xen/arm/shm.c
```diff
@@ -0,0 +1,101 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+#include <stdint.h>
```
michpappas: Since you (correctly) use kernel types I think `stdint.h` is not used and can be removed.

plat/xen/arm/shm.c
```diff
@@ -0,0 +1,101 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
michpappas: Please add the license block defined in https://unikraft.org/docs/contributing/coding-conventions#license-header

plat/xen/arm/shm.c
```diff
@@ -0,0 +1,101 @@
...
+	struct uk_shm_entry *entry = __NULL;
+	unsigned long flags;
+	unsigned long addr, end;
+	void *result;
+	int l3_idx;
+	lpae_t *pte;
+	lpae_t val;
+	int i;
+
+	if (!id) {
```
michpappas: It's probably a good idea to wrap the error path with `unilikely()` defined in `uk/compiler.h`.

plat/xen/include/xen-arm/shm.h
```diff
@@ -0,0 +1,51 @@
...
+
+#else /* !CONFIG_XEN_STATIC_SHM */
+
+static inline void *uk_shm_map(const char *id __unused, __sz len __unused,
+			       __u16 perm __unused, __off offset __unused)
+{
+	return __NULL;
+}
+
+#endif /* CONFIG_XEN_STATIC_SHM */
```
michpappas: Should be `#endif /* !CONFIG_XEN_STATIC_SHM */`

plat/xen/include/xen-arm/shm.h
```diff
@@ -0,0 +1,51 @@
...
+ * @param perm
+ *   UK_SHM_RW or UK_SHM_RO
+ * @param offset
+ *   Byte offset from the start of the shm region (must be page-aligned)
+ * @return
+ *   Virtual address at (region_base + offset), or NULL on error
+ */
+void *uk_shm_map(const char *id, __sz len, __u16 perm, __off offset);
+
+#else /* !CONFIG_XEN_STATIC_SHM */
```
michpappas: TODO: Don't define stub if the feature is not enabled.

michpappas (changes requested):
Hi @kang-taekyung, overall it looks good, please see inline comments. Regarding the new functions introduced for page tables management: To avoid code duplication I opened #1908 that introduces `libukpt` to decouple low-level page table management from `libukpaging`, and #1909 to adapt Xen. The idea is that this PR and subsequent work use `libukpt`. Both PRs are in draft state at the moment not so much due from the design perspective, but due to the lack of in-depth testing. You can expect that the merge state to be (hopefully) very close to the current one, so feel free to have a look and adapt this PR, or leave comments in #1908 / #1909.

PR #1898, opened 2026-09-02
{lib/ukdebug, support/scripts}: Fix uk tracepoint bugs

### LINKED ISSUE #1426: lib/ukdebug: make fails when building `app-helloworld` with `tracepoints` enabled on `arm64`
### Describe the bug

When building the image for `app-helloworld` with `tracepoints` enabled on the `amd64` platform, `make` fails with following error message:

```
...
  CC      libukdebug: print.o
  CC      libukdebug: outf.o
  CC      libukdebug: hexdump.o
  CC      libukdebug: trace.o
/var/folders/35/p0d3n8494dvd7kv5hlj0pg1m0000gn/T//ccjJtg6t.s: Assembler messages:
/var/folders/35/p0d3n8494dvd7kv5hlj0pg1m0000gn/T//ccjJtg6t.s:26: Error: junk at end of line, first unrecognized character is `#'
gmake[2]: *** [/Users/huyang/code/workdir/helloworld/workdir/unikraft/support/build/Makefile.build:27: /Users/huyang/code/workdir/helloworld/workdir/build/libukdebug/trace.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [Makefile:1170: sub-make] Error 2
gmake[1]: Leaving directory '/Users/huyang/code/workdir/helloworld/workdir/unikraft'
gmake: *** [Makefile:7: all] Error 2
```

### Steps to reproduce

- Clone the `app-helloworld` and `unikraft` repos and form a working directory [as you normally do](https://unikraft.org/guides/internals)
- Select `Enable tracepoints` in `make menuconfig` -> `Library Configuration` -> `ukdebug`:

image

- run `make -j$(nproc)`

### Expected behavior

The build process is expected to succeed.

arm64

kvm

### Relevant log output

_No response_

### LINKED ISSUE #1897: UK_TRACEPOINT
### Describe the bug

UK_TRACEPOINT infrastructure does not work on Xen/ARM64 (dom0less).
Six issues prevent tracepoint data from being collected and parsed:

1. Section attribute syntax: The `.uk_tracepoints_list` and `.uk_trace_keyvals`
   sections use `,"",@note#` which is a GNU assembler x86 trick where `#` starts a comment.
   ARM64 assembler rejects this with:
   "Error: junk at end of line, first unrecognized character is '#'"

2. `uk-gdb.py` cannot import uk_trace, causing ModuleNotFoundError: No module named 'uk_trace'

3. String value is treated as integer, when a `pickle` format file (which is saved via `uk trace save` in GDB helper script) is loaded:
   `parse.py` uses `elf` (which is an elf file path) in a `%d` formatting operation 
   This causes the parser to fail with: TypeError: %d format: a real number is required, not str

4. Tracepoint definition parser: `parse.py` assumes 32-byte alignment between
   entries in `.uk_tracepoints_list`. This is not guaranteed by any attribute or
   linker script, and fails on ARM64 where GCC uses different padding.

5. Signed integers printed as unsigned: All integer trace arguments are
   unpacked using unsigned struct formats (B/H/I/Q), causing signed values
   like -1 to be displayed as 4294967295. The tracepoint type system only
   distinguishes INT vs STRING, so signed-ness must be determined from the
   format string.

6. Format string conversion causes Python ValueError: `parse.py` only converted
   `%p` via str.replace() Specifiers with flags, width, length, or precision
   (e.g., `%#08lx`, `%016llX`, `%04lld`) were not handled, causing Python ValueError 
   at runtime because Python's `%` formatting does not recognize C length
   modifiers.
Developer info
gyunseo.lee@boeing.com

### Steps to reproduce

1. Add UK_TRACEPOINT() calls with %ld format specifiers
2. Build Unikraft for Xen/ARM64 with CONFIG_LIBUKDEBUG_TRACEPOINTS=y
3. Attach to the running the Xen hypervisor with GDB on a host machine.
(gdb [unstripped_image_with_debug_symbols].elf.dbg -iex "set auto-load safe-path /" -x "{unikraft_repo_path}/support/scripts/uk-gdb.py")
4. Save a tracefile via `uk trace save [tracefile path]`
5. Parse traces: python3 trace.py list [tracefile path]

Issue 1: Build fails at assembler stage with section attribute error
Issue 2: uk-gdb.py raises ModuleNotFoundError: No module named 'uk_trace'
Issue 3: prase.py raises TypeError: %d format: a real number is required, not str
Issue 4: parse.py raises "Wrong tracepoint definition magic", UnicodeDecodeError, or KeyError
Issue 5: Signed values displayed as large unsigned numbers
Issue 6: parse.py raises ValueError on format string application

### Expected behavior

Tracepoints should compile on ARM64, uk-gdb.py should be correctly load uk_trace module and save a tracefile,
and trace.py should correctly parse and display all tracepoint events with formatted arguments, 
including proper signed/unsigned rendering.

arm64

xen

### Relevant log output

```shell
--- Issue 1: Build error ---
/tmp/ccXEO4Fp.s: Assembler messages:
/tmp/ccXEO4Fp.s:1859: Error: junk at end of line, first unrecognized character is `#'

--- Issue 2: ModuleNotFoundError: No module named 'uk_trace'----
I replaced the actual path in the log with {uk-gdb.py path}.
Traceback (most recent call last):
  File "{uk-gdb.py path}", line 39, in <module>
    import uk_trace.parse as parse
ModuleNotFoundError: No module named 'uk_trace'
Traceback (most recent call last):
  File "{uk-gdb.py path}", line 39, in <module>
    import uk_trace.parse as parse
ModuleNotFoundError: No module named 'uk_trace'


--- Issue 3: String value is treated as integer when a `pickle` format file ---
I replaced the actual path in the log with {unikraft_repo_path}.
Traceback (most recent call last):
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 145, in <module>
    cli()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in list
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 65, in parse_tf
    return parse.sample_parser(keyvals, tp_defs, trace_buff, ptr_size)
  File {unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 93, in __init__
    self.tps = get_tp_definitions(tp_defs_data, ptr_size)
  File {unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 173, in get_tp_definitions
    ptr_fmt = "0x%0" + "%dx" % (ptr_size * 2)
TypeError: %d format: a real number is required, not str

--- Issue 4: Parser error ---
I replaced the actual path in the log with {unikraft_repo_path}.
  
Traceback (most recent call last):
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 145, in <module>
    cli()
  File "~/.local/lib/python3.10/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
  File "~/.local/lib/python3.10/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "~/.local/lib/python3.10/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "~/.local/lib/python3.10/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "~/.local/lib/python3.10/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in list
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 65, in parse_tf
    return parse.sample_parser(keyvals, tp_defs, trace_buff, ptr_size)
  File "{unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 95, in __init__
    self.tps = get_tp_definitions(tp_defs_data, ptr_size)
  File "{unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 188, in get_tp_definitions
    magic = magic.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 0: invalid start byte

--- Issue 4: Parser error ---
Traceback (most recent call last):
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 145, in <module>
    cli()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in list
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 65, in parse_tf
    return parse.sample_parser(keyvals, tp_defs, trace_buff, ptr_size)
  File "/home/ubuntu/blinx/catalog-core/repos/unikraft/support/scripts/uk_trace/parse.py", line 93, in __init__
    self.tps = get_tp_definitions(tp_defs_data, ptr_size)
  File "/home/ubuntu/blinx/catalog-core/repos/unikraft/support/scripts/uk_trace/parse.py", line 196, in get_tp_definitions
    raise Exception("Wrong tracepoint definition magic")
Exception: Wrong tracepoint definition magic

--- Issue 4: Parser error ---
Traceback (most recent call last):
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 145, in <module>
    cli()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in list
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in <listcomp>
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 111, in __next__
    tp = self.tps[cookie]
KeyError: 1568960

--- Issue 5: Signed values displayed as large unsigned numbers ---

I defined the following trace point function.  
  
UK_TRACEPOINT(signed_int_print_test, "-1 must be %d", int);
  
And, I inserted it into a test app as follows.  

signed_int_print_test(-1);

The trace log should be "-1 must be -1", but it shows 4294967295. (which is 2^32 - 1, 32bit unsigned int max)  
The following log is captured by running `python3 trace.py list [tracefile path]`.  

       time  tp_name                       msg
-----------  ----------------------------  ----------------------------------------------------------------------
   31556783  signed_int_print_test         -1 must be 4294967295

--- Issue 6: parse.py raises ValueError on format string application ---
I defined the following trace point function.  
  
UK_TRACEPOINT(long_long_print_test, "%lld", long long);
  
And, I inserted it into a test app as follows.  

long_long_print_test(0x7FFFFFFFFFFFFFFF);

The trace log should be "9223372036854775807" (which is 2^63 - 1, 64bit signed int max), but it shows the following error.
I ran `python3 trace.py list [tracefile path]`   

I replaced the actual path in the log with {unikraft_repo_path}.
  
Traceback (most recent call last):
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 145, in <module>
    cli()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in list
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/trace.py", line 74, in <listcomp>
    print_data = [x.tabulate_fmt() for x in parse_tf(trace_file)]
  File "{unikraft_repo_path}/support/scripts/uk_trace/parse.py", line 66, in tabulate_fmt
    return [self.time, self.tp.name, (self.tp.fmt % self.args)]
ValueError: unsupported format character 'l' (0x6c) at index 13
```

gyunseo:
### Description of Changes

(NOTE: Tied to the issue #1897)  
  
Fix UK_TRACEPOINT infrastructure bugs  
  
Following 7 issues prevent tracepoint data from being collected and parsed:  
  
1. Section attribute syntax: The `.uk_tracepoints_list` and `.uk_trace_keyvals`
   sections use `,"",@note#` which is a GNU assembler x86 trick where `#` starts a comment.
   ARM64 assembler rejects this with:
   "Error: junk at end of line, first unrecognized character is '#'"

2. `uk-gdb.py` cannot import uk_trace, causing ModuleNotFoundError: No module named 'uk_trace'

3.  The tracepoint section retriever uses `objcopy -O binary` to extract the `.uk_tracepoints_list` section. 
    However, this silently skips the section because it is a GNU-as NOTE section, which has the NON-ALLOC attribute.

4. String value is treated as integer, when a `pickle` format file (which is saved via `uk trace save` in GDB helper script) is loaded:
   `parse.py` uses `elf` (which is an elf file path) in a `%d` formatting operation 
   This causes the parser to fail with: TypeError: %d format: a real number is required, not str

5. Tracepoint definition parser: `parse.py` assumes 32-byte alignment between
   entries in `.uk_tracepoints_list`. This is not guaranteed by any attribute or
   linker script, and fails on ARM64 where GCC uses different padding.

6. Signed integers printed as unsigned: All integer trace arguments are
   unpacked using unsigned struct formats (B/H/I/Q), causing signed values
   like -1 to be displayed as 4294967295. The tracepoint type system only
   distinguishes INT vs STRING, so signed-ness must be determined from the
   format string.

7. Format string conversion causes Python ValueError: `parse.py` only converted
   `%p` via str.replace() Specifiers with flags, width, length, or precision
   (e.g., `%#08lx`, `%016llX`, `%04lld`) were not handled, causing Python ValueError 
   at runtime because Python's `%` formatting does not recognize C length
   modifiers.
  
This PR resolves the above issues.   

- Regarding [1] (`trace.h`, `trace.c`): Added `__UK_TRACEPOINTS_SECTION` and `__UK_TRACE_KEYVALS_SECTION` macros that use a GNU-as arm trick where `//` starts a comment on ARM/ARM64 and retain the `@note#` syntax on x86.
- Regarding [2] (`uk-gdb.py`): Moved `import uk_trace` after appending the script dir to sys.path. 
- Regarding [3] (`parse.py`): Used `objcopy --dump-section` instead, which extracts the section contents regardless of its ALLOC/NON-ALLOC attribute. 
- Regarding [4] (`uk-gdb.py`): The ELF path was not used after loading, so removed it from the pickle dump. 
- Regarding [5] (`parse.py`): Removed `__STRUCT_ALIGNMENT` constant. Made a logic scanning for `TP_DEF_MAGIC` to locate the first entry and use each entry's size field to jump to the next. Also, if there is any zero-padding, the parser consumes it.  
- Regarding [6] (`parse.py`): Parsed each argument's format specifier the per `printf(3)` man page to determine signed-ness: `d`/`i` are signed, `u`/`o`/`x`/`X`/`p` are unsigned. Added a 'signed' parameter to `unpack_int()` and stored per-argument signedness in `tp_definition`. Validated that the number of format specifiers matches `args_nr`. 
- Regarding [7] (`parse.py`): Replaced `str.replace()` format conversion with regex-based substitution that strips length modifiers (`hh`/`h`/`l`/`ll`/`z`/`t`/`j`/`q`) while preserving flags, width, and precision. Handled `%p` separately since it has no length modifier per the man page spec.
  
I believe this PR\
closes #1897, and\
closes #1426.  

### Related Work

N/A  
No any other PRs that this PR depends on or it will conflict with.  

### PR Checklist

### DIFF
--- a/lib/ukdebug/include/uk/trace.h
+++ b/lib/ukdebug/include/uk/trace.h
@@ -169,11 +169,18 @@ static inline void __uk_trace_save_arg(char **pbuff,
 #define __UK_TRACE_ARG_TYPES(n, ...)					\
 	{ UK_FOREACH(__UK_TRACE_GET_TYPE_FOREACH, __VA_ARGS__) }
 
+#if defined(__arm__) || defined(__aarch64__)
+#define __UK_TRACEPOINTS_SECTION \
+	__attribute((__section__(".uk_tracepoints_list,\"\",%note//")))
+#else
+#define __UK_TRACEPOINTS_SECTION \
+	__attribute((__section__(".uk_tracepoints_list,\"\",@note#")))
+#endif
+
 #define __UK_TRACE_REG(NR, regname, trace_name, fmt, ...)	\
 	UK_CTASSERT(sizeof(#trace_name) < 255);			\
 	UK_CTASSERT(sizeof(fmt) < 255);				\
-	__attribute((__section__(				\
-		".uk_tracepoints_list,\"\",@note#")))		\
+	__UK_TRACEPOINTS_SECTION				\
 	static struct {						\
 		uint32_t magic;					\
 		uint32_t size;					\
--- a/support/scripts/uk-gdb.py
+++ b/support/scripts/uk-gdb.py
@@ -36,10 +36,11 @@ import os
 import sys
 import tempfile
 import shutil
-import uk_trace.parse as parse
 
 scripts_dir = os.path.dirname(os.path.realpath(__file__))
 sys.path.append(scripts_dir)
+
+import uk_trace.parse as parse
 
 type_char = gdb.lookup_type("char")
 type_void = gdb.lookup_type("void")
@@ -80,7 +81,6 @@ def save_traces(out):
     # versions should just have modifications at the very end to keep
     # compatibility with previously collected data.
     pickler.dump(parse.get_keyvals(elf))
-    pickler.dump(elf)
     pickler.dump(PTR_SIZE)
     # We are saving raw trace buffer here. Another option is to pickle
     # already parsed samples. But in the chosen case it is a lot
--- a/support/scripts/uk_trace/parse.py
+++ b/support/scripts/uk_trace/parse.py
@@ -207,9 +207,10 @@ def get_tp_definitions(tp_data, ptr_size):
 
 def get_tp_sections(elf):
     f = tempfile.NamedTemporaryFile()
-    objcopy_cmd = "objcopy -O binary %s " % elf
-    objcopy_cmd += "--only-section=.uk_tracepoints_list " + f.name
-    objcopy_cmd = objcopy_cmd.split()
+    # NOTE: When reading a cross-architecture ELF file, make sure the objcopy utility supports that architecture.
+    objcopy_cmd = ["objcopy",
+                   "--dump-section", ".uk_tracepoints_list=" + f.name,
+                   elf, "/dev/null"]
     subprocess.check_call(objcopy_cmd)
     return f.read()
 
--- a/support/scripts/uk_trace/parse.py
+++ b/support/scripts/uk_trace/parse.py
@@ -39,8 +39,6 @@ TP_HEADER_MAGIC = "TRhd"
 TP_DEF_MAGIC = "TPde"
 UK_TRACE_ARG_INT = 0
 UK_TRACE_ARG_STRING = 1
-# Not sure why gcc aligns data on 32 bytes
-__STRUCT_ALIGNMENT = 32
 
 FORMAT_VERSION = 1
 
@@ -116,7 +114,7 @@ class sample_parser:
             if tp.types[i] == UK_TRACE_ARG_STRING:
                 args += [self.data.unpack_string()]
             else:
-                args += [self.data.unpack_int(tp.sizes[i])]
+                args += [self.data.unpack_int(tp.sizes[i], tp.signed[i])]
 
         return tp_sample(tp, time, tuple(args))
 
@@ -142,15 +140,15 @@ class unpacker:
         (ret,) = self.unpack(fmt)
         return ret.decode()
 
-    def unpack_int(self, size):
+    def unpack_int(self, size, signed=False):
         if size == 1:
-            fmt = "B"
+            fmt = "b" if signed else "B"
         elif size == 2:
-            fmt = "H"
+            fmt = "h" if signed else "H"
         elif size == 4:
-            fmt = "I"
+            fmt = "i" if signed else "I"
         elif size == 8:
-            fmt = "Q"
+            fmt = "q" if signed else "Q"
         (ret,) = self.unpack(fmt)
         return ret
 
@@ -159,12 +157,13 @@ class unpacker:
 
 
 class tp_definition:
-    def __init__(self, name, args_nr, fmt, sizes, types):
+    def __init__(self, name, args_nr, fmt, sizes, types, signed=None):
         self.name = name
         self.args_nr = args_nr
         self.fmt = fmt
         self.sizes = sizes
         self.types = types
+        self.signed = signed if signed is not None else [False] * args_nr
 
     def __str__(self):
         return "%s %s" % (self.name, self.fmt)
@@ -176,10 +175,17 @@ def get_tp_definitions(tp_data, ptr_size):
 
     ret = dict()
 
+    # Find the first entry by scanning for the TP_DEF_MAGIC string
+    while data.pos < len(data.data) - 4:
+        if data.data[data.pos:data.pos+4] == TP_DEF_MAGIC.encode():
+            break
+        data.pos += 1
+
     while True:
-        data.align_pos(__STRUCT_ALIGNMENT)
+        entry_start = data.pos
+
         try:
-            magic, _, cookie, args_nr, name_len, fmt_len = data.unpack(
+            magic, size, cookie, args_nr, name_len, fmt_len = data.unpack(
                 "4sIQBBB"
             )
         except EndOfBuffer:
@@ -197,10 +203,86 @@ def get_tp_definitions(tp_data, ptr_size):
         name = name[:-1].decode()
         fmt = fmt[:-1].decode()
 
-        # Convert from c-printf format into python one
+        # Determine signed-ness of each integer argument from the format string
+        # Per printf(3): %[flags][width][.precision][length]specifier
+        #
+        # Length modifiers: hh, h, l, ll, z, t, j, q
+        # Signed specifiers: d, i
+        # Unsigned specifiers: u, o, x, X
+        # Pointer: p (always unsigned, size = ptr_size)
+
+        _FLAGS = r"[#0\- +']*"
+        _WIDTH = r'(?:\*|\d+)?'
+        _PRECISION  = r'(?:\.(?:\*|\d+))?'
+        _LEN   = r'(?:hh|h|ll|l|z|t|j|q)?'
+        _FWP   = _FLAGS + _WIDTH + _PRECISION  # flags + width + precision (optional)
+
+        _FMT_SPEC_RE = re.compile(
+            r'%%'                       # literal %% (skip, consumes no arg) (group 0)
+            r'|%' + _FWP +             # optional (group 0)
+            r'(hh|h|ll|l|z|t|j|q)?'    # length modifier (group 1)
+            r'([diouxXs])'              # conversion specifier (group 2)
+            r'|%p'                      # pointer (no length modifier) (group 0)
+        )
+
+        # Validate: the number of format specifiers must match args_nr
+        num_specifiers = sum(
+            1 for m in _FMT_SPEC_RE.finditer(fmt)
+            if m.group(0) != '%%'
+        )
+        assert num_specifiers == args_nr, (
+            "Format specifier count mismatch: fmt has %d, args_nr is %d for '%s'"
+            % (num_specifiers, args_nr, name)
+        )
+
+        # Now safely determine signed-ness for each argument
+        is_arg_signed = [False for _ in range(args_nr)]
+        fmt_tmp = fmt
+        for i in range(args_nr):
+            # Skip any literal %% before looking for the real specifier
+            while True:
+                m = _FMT_SPEC_RE.search(fmt_tmp)
+                if not m:
+                    break
+                if m.group(0) == '%%':
+                    fmt_tmp = fmt_tmp[m.end():]
+                    continue
+                break
+
+            if types[i] == UK_TRACE_ARG_STRING:
+                if m and m.group(2) == 's':
+                    fmt_tmp = fmt_tmp[m.end():]
+            else:
+                if m:
+                    conv = m.group(2)  # None if matched %p branch
+                    if conv and conv != 's':
+                        is_arg_signed[i] = conv in ('d', 'i')
+                    else: # %p case
+                        is_arg_signed[i] = False
+                    fmt_tmp = fmt_tmp[m.end():]
+
+        # Convert C printf format specifiers into Python equivalents.
+        # Strip length modifiers while preserving flags, width, and precision.
+        _FWP_CAP = '(' + _FLAGS + _WIDTH + _PRECISION + ')'  # capturing version
+
+        # Hex/octal: strip length modifier, keep the rest
+        fmt = re.sub(r'%' + _FWP_CAP + _LEN + r'([xXo])', r'%\1\2', fmt)
+
+        # Decimal (d, i, u): strip length modifier, normalize to %d
+        fmt = re.sub(r'%' + _FWP_CAP + _LEN + r'[diu]', r'%\1d', fmt)
+
+        # %p has no length modifier per spec; simple replace at the end
         fmt = fmt.replace("%p", ptr_fmt)
 
-        ret[cookie] = tp_definition(name, args_nr, fmt, sizes, types)
+        ret[cookie] = tp_definition(name, args_nr, fmt, sizes, types,
+                                    is_arg_signed)
+
+        # Use size field to jump to next entry
+        data.pos = entry_start + size
+
+        # Skip any zero-padding (compiler alignment between entries)
+        while data.pos < len(data.data) and data.data[data.pos] == 0:
+            data.pos += 1
 
     return ret
 
--- a/lib/ukdebug/trace.c
+++ b/lib/ukdebug/trace.c
@@ -42,7 +42,6 @@ char uk_trace_buffer[CONFIG_LIBUKDEBUG_TRACE_BUFFER_SIZE];
 size_t uk_trace_buffer_free = CONFIG_LIBUKDEBUG_TRACE_BUFFER_SIZE;
 char *uk_trace_buffer_writep = uk_trace_buffer;
 
-
 /* Store a string in format "key = value" in the section
  * .uk_trace_keyvals. This can be anything what you want trace.py
  * script to know about, and what you do not want to consume any space
@@ -53,10 +52,16 @@ char *uk_trace_buffer_writep = uk_trace_buffer;
  *
  *     $ readelf -p .uk_trace_keyvals <your_image.gdb>
  */
-#define TRACE_DEFINE_KEY(key, val)			\
-	__attribute((__section__(			\
-		".uk_trace_keyvals,\"\",@note#")))	\
-	static const char key[] __used =		\
-		#key " = " #val
+#if defined(__arm__) || defined(__aarch64__)
+#define __UK_TRACE_KEYVALS_SECTION                                             \
+	__attribute((__section__(".uk_trace_keyvals,\"\",%note//")))
+#else
+#define __UK_TRACE_KEYVALS_SECTION                                             \
+	__attribute((__section__(".uk_trace_keyvals,\"\",@note#")))
+#endif
+
+#define TRACE_DEFINE_KEY(key, val)                                             \
+	__UK_TRACE_KEYVALS_SECTION                                             \
+	static const char key[] __used = #key " = " #val
 
 TRACE_DEFINE_KEY(format_version, 1);


gyunseo:
```bash
git diff HEAD~4..HEAD | ./support/scripts/checkpatch.uk
total: 0 errors, 0 warnings, 0 checks, 238 lines checked

Your patch has no obvious style problems and is ready for submission.

NOTE: Ignored message types: ASSIGN_IN_IF AVOID_BUG C99_COMMENT_TOLERANCE FILE_PATH_CHANGES NEW_TYPEDEFS OBSOLETE SYMBOLIC_PERMS
```

gyunseo:
```bash
./support/scripts/checkpatch.uk --git HEAD~4..HEAD
---------------------------------------------------------------------------
Commit 8cef802feeec ("lib/ukdebug: Fix section attribute syntax for ARM64")
---------------------------------------------------------------------------
total: 0 errors, 0 warnings, 0 checks, 48 lines checked

Commit 8cef802feeec ("lib/ukdebug: Fix section attribute syntax for ARM64") has no obvious style problems and is ready for submission.
----------------------------------------------------------------------------------
Commit 839b06c04d7c ("support/scripts/uk_trace: Fix tracepoint section retrieval")
----------------------------------------------------------------------------------
total: 0 errors, 0 warnings, 0 checks, 13 lines checked

Commit 839b06c04d7c ("support/scripts/uk_trace: Fix tracepoint section retrieval") has no obvious style problems and is ready for submission.
----------------------------------------------------------------------------------
Commit c210e1a2c40d ("support/scripts/uk_trace: Fix tracepoint definition parser")
----------------------------------------------------------------------------------
total: 0 errors, 0 warnings, 0 checks, 157 lines checked

Commit c210e1a2c40d ("support/scripts/uk_trace: Fix tracepoint definition parser") has no obvious style problems and is ready for submission.
--------------------------------------------------------------
Commit 1df1c9a9f962 ("support/scripts: Fix gdb helper script")
--------------------------------------------------------------
total: 0 errors, 0 warnings, 0 checks, 20 lines checked

Commit 1df1c9a9f962 ("support/scripts: Fix gdb helper script") has no obvious style problems and is ready for submission.

NOTE: Ignored message types: ASSIGN_IN_IF AVOID_BUG C99_COMMENT_TOLERANCE FILE_PATH_CHANGES NEW_TYPEDEFS OBSOLETE SYMBOLIC_PERMS
```

support/scripts/uk-gdb.py
```diff
@@ -80,7 +81,6 @@ def save_traces(out):
     # versions should just have modifications at the very end to keep
     # compatibility with previously collected data.
     pickler.dump(parse.get_keyvals(elf))
-    pickler.dump(elf)
     pickler.dump(PTR_SIZE)
```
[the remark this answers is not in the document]
mogasergiu: I think this one is backwards. The reader already dropped this field: `trace.py:54` has had `# elf = unpickler.load()` commented out since e6814150, while the writer kept dumping it. That desync is exactly the `TypeError` the commit message quotes.

So old trace files are not readable today either. To check, I took a tracefile this branch produces and re-pickled it in the old five-slot layout (keyvals, elf, ptr_size, defs, buf), which is exactly what a pre-1df1c9a9 `uk-gdb.py` writes:

```python
import pickle
with open("tracefile", "rb") as f:
    u = pickle.Unpickler(f)
    kv, ps, td, tb = u.load(), u.load(), u.load(), u.load()
with open("tracefile.oldfmt", "wb") as f:
    p = pickle.Pickler(f)
    p.dump(kv); p.dump("/path/to/image.dbg"); p.dump(ps); p.dump(td); p.dump(tb)
```

Then read it back with both versions of the scripts (`staging/` is a checkout of `ee652d8d`):

```
$ cd staging/support/scripts/uk_trace && python3 trace.py list --no-tabulate ../../../../tracefile.oldfmt
  File ".../staging/support/scripts/uk_trace/parse.py", line 174, in get_tp_definitions
    ptr_fmt = "0x%0" + "%dx" % (ptr_size * 2)
                       ~~~~~~^~~~~~~~~~~~~~~~
TypeError: %d format: a real number is required, not str

$ cd support/scripts/uk_trace && python3 trace.py list --no-tabulate .../tracefile.oldfmt
  File ".../support/scripts/uk_trace/parse.py", line 173, in get_tp_definitions
    ptr_fmt = "0x%0" + "%dx" % (ptr_size * 2)
                       ~~~~~~^~~~~~~~~~~~~~~~
TypeError: %d format: a real number is required, not str
```

Same failure on staging and on this branch, so there is no readable old data whose compatibility this could break.

What you did looks right to me, ignore the copilot comment. Could we also delete the stale commented-out line in `trace.py` in this same commit, so both sides visibly agree? Otherwise the next person reading `parse_tf` sees a slot that looks like it might still be there and has to dig through git history to find out it is not.

```diff
--- a/support/scripts/uk_trace/trace.py
+++ b/support/scripts/uk_trace/trace.py
@@ -51,7 +51,6 @@
             unpickler = pickle.Unpickler(tf)
 
             keyvals = unpickler.load()
-            # elf = unpickler.load()
             ptr_size = unpickler.load()
             tp_defs = unpickler.load()
             trace_buff = unpickler.load()
```

support/scripts/uk_trace/parse.py
```diff
@@ -197,19 +203,96 @@ def get_tp_definitions(tp_data, ptr_size):
...
-        # Convert from c-printf format into python one
+        # Determine signed-ness of each integer argument from the format string
+        # Per printf(3): %[flags][width][.precision][length]specifier
+        #
+        # Length modifiers: hh, h, l, ll, z, t, j, q
+        # Signed specifiers: d, i
+        # Unsigned specifiers: u, o, x, X
+        # Pointer: p (always unsigned, size = ptr_size)
+
+        _FLAGS = r"[#0\- +']*"
```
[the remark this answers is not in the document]
mogasergiu: Yeah, this one is real, and it is simpler than it looks: Python's `%` operator has no `'` flag at all.

```
$ python3 -c "print(\"%'d\" % 1234567)"
    print("%'d" % 1234567)
          ~~~~~~^~~~~~~~~
ValueError: unsupported format character ''' (0x27) at index 1
```

So having `'` in `_FLAGS` only defers the failure to render time.

`*` in `_WIDTH` and `_PRECISION` is the same shape of problem from the same place: `%*d` matches as a single specifier, but in C it consumes two varargs, so it fails the new specifier count check rather than rendering.

Feels like the regex came from the printf(3) grammar rather than from the intersection of what Python's `%` operator accepts and what the tracepoint ABI can express. Should we drop `'` from the flags class and `\*` from the width and precision classes, so both land on the same "unsupported, warn and carry on" path instead of half parsing?

support/scripts/uk_trace/parse.py
```diff
@@ -197,19 +203,96 @@ def get_tp_definitions(tp_data, ptr_size):
...
+            r'([diouxXs])'              # conversion specifier (group 2)
+            r'|%p'                      # pointer (no length modifier) (group 0)
+        )
+
+        # Validate: the number of format specifiers must match args_nr
+        num_specifiers = sum(
+            1 for m in _FMT_SPEC_RE.finditer(fmt)
+            if m.group(0) != '%%'
+        )
+        assert num_specifiers == args_nr, (
```
mogasergiu: This aborts `get_tp_definitions()` entirely, so one bad format anywhere in the image costs you every tracepoint, not just that one. Easiest way to see it is a format and argument count typo, which the compiler cannot catch. I added one line to `lib/ukintctlr/ukintctlr.c`:

```c
UK_TRACEPOINT(trace_tmp_typo, "a=%d b=%d", int);   /* 2 specifiers, 1 arg */
```

Kernel builds fine. Then, against a live elfloader guest that had already recorded four socket tracepoints:

```
(gdb) uk trace
Python Exception <class 'AssertionError'>: Format specifier count mismatch: fmt has 2, args_nr is 1 for 'trace_tmp_typo'
Error occurred in Python: Format specifier count mismatch: fmt has 2, args_nr is 1 for 'trace_tmp_typo'
```

Nothing else. All 47 definitions and all 4 captured samples gone, because of one unrelated tracepoint in a library I was not even looking at. That is a big blast radius for a debug tool, and it will fire on out of tree libs and apps we do not control.

The other thing is that `assert` gets stripped under `python -O`, so instead of failing loudly you silently get the wrong answer:

```
$ python3 -O fmtcheck.py "%c" 1:0 -- 65
python fmt: '%c'  signed: [False]
rendered  : 'A'
```

Maybe we could print a warning to stderr, fall back to unsigned for that one definition, and keep going? Same image, same guest, with that change:

```
(gdb) uk trace
Warning: tracepoint 'trace_tmp_typo' has 2 format specifiers but 1 arguments, assuming unsigned
0000000165703137 trace_posix_socket_create: 2 1 0
0000000165730544 trace_posix_socket_create_ret: 3
0000000165733464 trace_posix_socket_create: -1 -1 -1
0000000165733762 trace_posix_socket_create_err: -97
```

gdb does surface Python's stderr, so the warning is not lost. Worth knowing that `trace.py fetch` pipes gdb's output and only prints it on failure or `--verbose`, so it would be hidden there.

Since we would be touching this anyway, it might be worth hoisting the regexes to module level (they get rebuilt for every entry right now) and collapsing the three alternatives into one so `group(2)` is always the conversion. Then the whole C to Python conversion becomes a single `re.sub` with a callback, which also fixes literal text getting chewed up:

```
$ ./fmtcheck.py "%%lx %d" 4:0 -- 7
python fmt: '%%x %d'  signed: [True]
rendered  : '%x 7'
```

The `l` there was never part of a specifier.

Also, I think on `types[i] == STRING` when the matched spec is not `s`, `fmt_tmp` never advances, so every later argument inherits the previous one's specifier. Should this be a azip over the matches instead?

`%c`, `%f`, `%#p`, `%*d` and `%.*d` all trip the assert the same way. Those I checked with a throwaway test script below.

```python
#!/usr/bin/env python3
"""usage: fmtcheck.py '<fmt>' <size>:<type> ... -- <value> ...
   e.g. fmtcheck.py '%d %s' 4:0 1:1 -- -1 hello
Run from support/scripts/."""
import struct, sys
sys.path.insert(0, ".")
import uk_trace.parse as p

fmt = sys.argv[1]
sep = sys.argv.index("--")
spec = [a.split(":") for a in sys.argv[2:sep]]
sizes = [int(s) for s, _ in spec]
types = [int(t) for _, t in spec]
vals = [int(v, 0) if v.lstrip("-").isdigit() or v.startswith("0x") else v
        for v in sys.argv[sep + 1:]]

n = b"tp\0"; f = fmt.encode() + b"\0"
body = struct.pack("<BBB", len(sizes), len(n), len(f)) + bytes(sizes) + bytes(types) + n + f
size = (16 + len(body) + 7) & ~7
blob = struct.pack("<4sIQ", b"TPde", size, 0x1000) + body
blob += b"\0" * (size - len(blob))

tp = list(p.get_tp_definitions(blob, 8).values())[0]
print("python fmt: %r  signed: %s" % (tp.fmt, tp.signed))
print("rendered  : %r" % (tp.fmt % tuple(vals),))
```

support/scripts/uk_trace/parse.py
```diff
@@ -197,19 +203,96 @@ def get_tp_definitions(tp_data, ptr_size):
...
+
+        # %p has no length modifier per spec; simple replace at the end
         fmt = fmt.replace("%p", ptr_fmt)
 
-        ret[cookie] = tp_definition(name, args_nr, fmt, sizes, types)
+        ret[cookie] = tp_definition(name, args_nr, fmt, sizes, types,
+                                    is_arg_signed)
+
+        # Use size field to jump to next entry
+        data.pos = entry_start + size
```
mogasergiu: `data.pos = entry_start + size` is the only thing advancing the cursor, and the zero padding skip will not step over a `TPde`, so a `size` that does not cover the fields we already read spins here forever. Took the real 3176 byte section out of an elfloader build and zeroed one entry's size field:

```
$ timeout 10 python3 corrupt.py elfloader_qemu-x86_64.dbg
real section: 3176 bytes, first entry size 64 -> 0
parsing ...
Terminated
```

Never returns. The old `align_pos()` loop could not do this because `unpack()` always moved forward.

Should we bail out before jumping if `entry_start + size` is behind where we already are? That guarantees `data.pos` strictly increases every iteration. Same input with that change:

```
$ timeout 10 python3 corrupt.py elfloader_qemu-x86_64.dbg
real section: 3176 bytes, first entry size 64 -> 0
parsing ...
Warning: tracepoint 'trace_posix_socket_create' has a bogus size of 0, stopping
returned
```

It has to be `<` and not `<=`, since `sizeof()` is often exactly the bytes we read with no trailing padding. I tried `<=` first and it silently dropped 35 of the 46 tracepoints in that same image.

While in here, `magic.decode()` a few lines up throws `UnicodeDecodeError` on non zero padding, before the magic check that is meant to catch that. We should probably compare the raw bytes instead.

The `corrupt.py` script used for testing below:

```python
# Run from support/scripts/ with a .dbg image as argv[1].
import struct, sys
sys.path.insert(0, ".")
import uk_trace.parse as p

sec = bytearray(p.get_tp_sections(sys.argv[1]))
# first entry's size field lives at offset 4, right after the "TPde" magic
orig = struct.unpack_from("<I", sec, 4)[0]
struct.pack_into("<I", sec, 4, 0)
print("real section: %d bytes, first entry size %d -> 0" % (len(sec), orig), flush=True)
print("parsing ...", flush=True)
p.get_tp_definitions(bytes(sec), 8)
print("returned")
```

support/scripts/uk_trace/parse.py
```diff
@@ -176,10 +175,17 @@ def get_tp_definitions(tp_data, ptr_size):
 
     ret = dict()
 
+    # Find the first entry by scanning for the TP_DEF_MAGIC string
+    while data.pos < len(data.data) - 4:
```
mogasergiu: Tiny thing: `len(data.data) - 4` means a magic sitting at the last valid offset never gets checked. Should this be `data.pos + 4 <= len(data.data)` instead?

support/scripts/uk_trace/parse.py
```diff
@@ -197,19 +203,96 @@ def get_tp_definitions(tp_data, ptr_size):
...
 
     return ret
 
 
 def get_tp_sections(elf):
     f = tempfile.NamedTemporaryFile()
-    objcopy_cmd = "objcopy -O binary %s " % elf
-    objcopy_cmd += "--only-section=.uk_tracepoints_list " + f.name
-    objcopy_cmd = objcopy_cmd.split()
+    # NOTE: When reading a cross-architecture ELF file, make sure the objcopy utility supports that architecture.
```
mogasergiu: This works, but it makes objcopy serialize the entire image into `/dev/null` just to pull out 3 KiB, and the leftover warning lands in the middle of every `uk trace`:

```
$ objcopy --dump-section .uk_tracepoints_list=tp.bin elfloader.dbg /dev/null
objcopy: /dev/null: section .bss lma 0x226000 adjusted to 0x22a204
exit=0
```

```
(gdb) uk trace
objcopy: /dev/null: section .bss lma 0x226000 adjusted to 0x22a204
0000000162047325 trace_posix_socket_create: 2 1 0
```

The cross architecture NOTE makes sense and I can see how a host objcopy hard fails, and a gdb session against a cross built image is kind of the main case here:

```
$ objcopy --dump-section .uk_tracepoints_list=tp.bin hello-arm64.dbg /dev/null
objcopy: Unable to recognise the architecture of the input file `hello-arm64.dbg'
exit=1

$ python3 -c "import uk_trace.parse as p; print(len(p.get_tp_sections('hello-arm64.dbg')))"
subprocess.CalledProcessError: Command '['objcopy', '--dump-section',
'.uk_tracepoints_list=/tmp/tmpalupu709', 'hello-arm64.dbg', '/dev/null']'
returned non-zero exit status 1.
```

`readelf` has no such problem on the same file, and we already shell out to it in `get_keyvals()`:

```
$ readelf -p .uk_trace_keyvals hello-arm64.dbg

String dump of section '.uk_trace_keyvals':
  [     0]  format_version = 1
```

Should we try walking the section header table ourselves instead? It is probably not that many lines of code and it drops all three awkward bits at once: the subprocess, the unclosed `NamedTemporaryFile`, and the fact that we read back through a handle opened before objcopy truncated and rewrote that same path. That last one happens to work on Linux because objcopy reopens the same inode, but it is not something I would want to rely on.

`readelf -x` would also work cross arch if you prefer a two liner, but parsing its hex columns is uglier than just reading the ELF. No strong opinion here, but FYI it does not handle `e_shnum == 0` / `SHN_XINDEX` (more than 64k sections), though of course this is an impossible case to ever reach for our unikernel image, so up to you.

support/scripts/uk-gdb.py
mogasergiu: Noticed this but not introduced here, but it is in the file this commit is fixing so maybe worth picking up. Any image where tracing is on but nothing has fired yet:

```
$ gdb -nh -batch hello-x86.dbg -ex "target remote :1234" \
      -ex "source uk-gdb.py" \
      -ex "print uk_trace_buffer_writep" -ex "print &uk_trace_buffer" \
      -ex "uk trace" -ex "uk trace save out.trace"

$1 = 0x147320 <uk_trace_buffer> ""
$2 = (char (*)[16384]) 0x147320 <uk_trace_buffer>
Python Exception <class 'ValueError'>: Argument 'count' should be greater than zero
Error occurred in Python: Argument 'count' should be greater than zero
Saving traces to out.trace ...
Python Exception <class 'ValueError'>: Argument 'count' should be greater than zero
Error occurred in Python: Argument 'count' should be greater than zero

$ ls out.trace
ls: cannot access 'out.trace': No such file or directory
```

The `trace_buff_writep == 0` guard only covers the case where the compiler optimised the variable away. When tracing is on but idle, `writep` equals `&uk_trace_buffer`, so `used` is 0 and `read_memory(addr, 0)` throws. `save` dies too, so you end up with no tracefile at all rather than an empty one.

Could we just return `b""` early when `used == 0` instead of calling `read_memory()`?

support/scripts/uk_trace/parse.py
```diff
@@ -197,19 +203,96 @@ def get_tp_definitions(tp_data, ptr_size):
...
+        # Length modifiers: hh, h, l, ll, z, t, j, q
+        # Signed specifiers: d, i
+        # Unsigned specifiers: u, o, x, X
+        # Pointer: p (always unsigned, size = ptr_size)
+
+        _FLAGS = r"[#0\- +']*"
+        _WIDTH = r'(?:\*|\d+)?'
+        _PRECISION  = r'(?:\.(?:\*|\d+))?'
+        _LEN   = r'(?:hh|h|ll|l|z|t|j|q)?'
+        _FWP   = _FLAGS + _WIDTH + _PRECISION  # flags + width + precision (optional)
```
mogasergiu: `.github/workflows/pychecks.yaml` runs `black` over `support/scripts`, and this file was clean before:

```
$ black --check --line-length 80 support/scripts        # staging
would reformat support/scripts/uk-libinfo.py
1 file would be reformatted, 13 files would be left unchanged.

$ black --check --line-length 80 support/scripts        # this branch
would reformat support/scripts/uk_trace/parse.py
would reformat support/scripts/uk-libinfo.py           # unrelated to PR: ignore
2 files would be reformatted, 12 files would be left unchanged.
```

The new code is single quoted while the rest of the file is double quoted (e6814150 black formatted it on purpose), plus six lines over 79 columns, the worst being the 113 char cross architecture NOTE comment in `get_tp_sections()`.

`black -l 80 support/scripts/uk_trace/parse.py` should sort it.

(`uk-libinfo.py` is pre-existing, not yours. ruff is unchanged at 13 findings for this file and already red repo wide, so I would ignore this one.)

lib/ukdebug/include/uk/trace.h
```diff
@@ -169,11 +169,18 @@ static inline void __uk_trace_save_arg(char **pbuff,
 #define __UK_TRACE_ARG_TYPES(n, ...)					\
 	{ UK_FOREACH(__UK_TRACE_GET_TYPE_FOREACH, __VA_ARGS__) }
 
+#if defined(__arm__) || defined(__aarch64__)
```
mogasergiu: So, as a general note/convention we have for guards, for such guards we prefer using the arch config (e.g. `CONFIG_ARCH_ARM_64` or `CONFIG_ARCH_X86_64`) and the `#else` and `#endif` would then have the opposite condition, e.g.:
```C
#if CONFIG_ARCH_ARM_64 || CONFIG_ARCH_ARM_32
#define __UK_TRACEPOINTS_SECTION \
	__attribute((__section__(".uk_tracepoints_list,\"\",%note//")))
#else /* !CONFIG_ARCH_ARM_64 && !CONFIG_ARCH_ARM_32 */
#define __UK_TRACEPOINTS_SECTION \
	__attribute((__section__(".uk_tracepoints_list,\"\",@note#")))
#endif /* !CONFIG_ARCH_ARM_64 && !CONFIG_ARCH_ARM_32 */
```
Ditto for keyvals below.

But that aside, the fix itself is right, reverting these two files does break the arm64 build with exactly the error you quote.

However, the conditional is on arch, but what actually differs is the assembler. clang's integrated assembler quotes the whole string as a section *name*, on both branches of this `#if`. Plain helloworld, `CC=clang`, builds fine and then:

```
$ readelf -SW build-clang/helloworld_qemu-x86_64 | grep -i uk_trace
  [ 3] .uk_trace_keyvals,"",@note#     PROGBITS 001339e0 ...  A  0 0 16
  [13] .uk_tracepoints_list,"",@note#  PROGBITS 00142288 ... WA  0 0  8
```

That is the stripped boot image, so with clang the sections are ALLOC, sit in LOAD segments, survive `-R` and ship, and `parse.py` finds nothing. Entirely pre-existing, I get the same result building `ee652d8d`, so not saying this PR introduces this bug.

Simplest thing would be `__GNUC__ && !__clang__` rather than an arch check, so clang at least lands in a branch we have thought about.

But I wonder if we could drop the assembler trick altogether. The only property we seem to want is NON-ALLOC, so the section costs nothing at runtime and `-R` can drop it without leaving a hole. The *type* looks irrelevant: `--dump-section`, `readelf -p` and `parse.py` all ignore it, and `@note` is arguably a lie anyway, these are not ELF notes and `readelf -n` already says so on the current build:

```
Displaying notes found in: .uk_tracepoints_list
readelf: Warning: Corrupt note: alignment 32, expecting 4 or 8
```

If that holds, asking the linker for NON-ALLOC instead of the assembler gets the same result with no arch or toolchain conditional at all:

```diff
--- a/lib/ukdebug/include/uk/trace.h
+++ b/lib/ukdebug/include/uk/trace.h
-#if defined(__arm__) || defined(__aarch64__)
-#define __UK_TRACEPOINTS_SECTION \
-	__attribute((__section__(".uk_tracepoints_list,\"\",%note//")))
-#else
+/* NON-ALLOC comes from the (INFO) output section in trace.ld */
 #define __UK_TRACEPOINTS_SECTION \
-	__attribute((__section__(".uk_tracepoints_list,\"\",@note#")))
-#endif
+	__attribute((__section__(".uk_tracepoints_list")))

--- a/lib/ukdebug/trace.c
+++ b/lib/ukdebug/trace.c
-#if defined(__arm__) || defined(__aarch64__)
+/* NON-ALLOC comes from the (INFO) output section in trace.ld */
 #define __UK_TRACE_KEYVALS_SECTION                                             \
-	__attribute((__section__(".uk_trace_keyvals,\"\",%note//")))
-#else
-#define __UK_TRACE_KEYVALS_SECTION                                             \
-	__attribute((__section__(".uk_trace_keyvals,\"\",@note#")))
-#endif
+	__attribute((__section__(".uk_trace_keyvals")))

--- a/lib/ukdebug/trace.ld
+++ b/lib/ukdebug/trace.ld
-	.uk_tracepoints_list : {
+	.uk_tracepoints_list (INFO) : {
 		uk_tracepoints_start = .;
 		KEEP (*(.uk_tracepoints_list));
 		uk_tracepoints_end = .;
 	}
-	.uk_trace_keyvals : {
+	.uk_trace_keyvals (INFO) : {
 		KEEP (*(.uk_trace_keyvals));
 	}
```

I ran that through gcc x86_64, gcc arm64 and clang x86_64. Sections stay NON-ALLOC and get stripped in all three, gcc boot images come out byte for byte the same size as this branch, and `uk trace` still works end to end, including on clang where it currently returns nothing at all.

The nice side effect, I guess, is the arch/toolchain conditional disappears. No strong opinion here, up to you which approach you think is better.

mogasergiu (changes requested):
Looks like a genuine improvement. Took it out for a spin and bumped into a couple of things mentioned below.

Issue #1896, opened 2026-09-01
uk_pr_* log body dropped on Xen/ARM64 dom0less (vpl011)


garethodickin:
### Describe the bug

On Xen/ARM64 dom0less, kernel logs from uk_pr\_\* (uk_pr_crit, uk_pr_err, ...) are broken, while the application's own printf works fine. The bug is in lib/ukprint/console.c, in how it prints ANSI color codes (LVLC\_\*) around each log line. 1. Message body dropped: color-code lengths were measured with sizeof(...), which also counts the string's NUL terminator (sizeof("") == 1). So an extra '\\0' byte was written before the body. In dom0less the hypervisor flushes the guest buffer with printk("%s", ...), and "%s" stops at that '\\0' -- dropping the rest of the message. (dom0 copies bytes raw, so it is unaffected.) 2. Color leaks to later lines (color ON): the reset code was sent after the '\\n'. Since dom0less flushes one line per '\\n', the reset slipped into the next line, leaving everything after a CRIT line (incl. (XEN) lines) red.

Developer Info
taekyung.kang@boeing.com

### Steps to reproduce

1. Build a Unikraft application for Xen/ARM64 with kernel logging enabled, using the PL011 console.
2. Boot the unikernel as a Xen dom0less domain (vpl011 backend served by the hypervisor).
3. Invoke any uk_pr_* log (e.g. uk_pr_crit / uk_pr_err).
4. Observe that the header (CRIT:, [libukboot], <boot.c @ ...>) prints but the message body is missing.

### Expected behavior

- The full uk_pr_* message body is printed on dom0less vpl011.
- With ANSI color on, each log line's color is reset within the same line; no color bleeds into subsequent lines.

### Which architectures were you using or does this bug affect?

arm64

### Which platforms were you using or does this bug affect?

xen

### Relevant log output

```shell
Actual pre-fix output (dom0less vpl011): the CRIT header prints, 
but the message body is cut off at the stray '\0' and the "%s" flush 
then prints only garbage/repeated "DOM1:" fragments afterwards:

(XEN) DOM1: [    0.059734] CRIT: [libmusl] <abort.c @    7> DOM1: DOM1: (XEN) Freed 424kB init memory.
DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1: DOM1:
```

PR #1899, opened 2026-09-03
lib/ukprint: Fix uk_pr_* logs on Xen dom0less vpl011

kang-taekyung:
Tied to discussion [uk_pr_* log body dropped on Xen/ARM64 dom0less (vpl011)](https://github.com/unikraft/unikraft/issues/1896)

### Description of Changes

- Stray NUL: string-literal lengths used sizeof(...),  which adds an extra '\0' (sizeof("") == 1).
  As a result a stray NUL was emitted on every console write, affecting all platforms and drivers.
  In particular, on Xen dom0less vpl011, the NUL cuts off the output in the printk("%s") path.
  Update to use sizeof(...) - 1.

- Color leak with ANSI on: the reset was written after '\n' on every console. 
  On consoles that flush on newline (e.g. vpl011) this moved the reset to the next line, leaking colors.
  Write in the order body -> reset -> '\n'.

### Related Work

N/A

### PR Checklist


lib/ukprint/console.c
```diff
@@ -410,18 +412,24 @@ void uk_print_console_write(struct uk_print_msg *msg)
...
-
+		/* Write the reset before the trailing '\n' so consoles that
+		 * flush on newline (e.g. Xen vpl011) keep the reset on the
+		 * same line and colors do not bleed to the next line.
+		 * Order: body -> reset -> '\n'.
+		 */
+		vprint_cout(cons, (char *)lptr, nlptr ? llen - 1 : llen);
+		vprint_cout(cons, LVLC_RESET, sizeof(LVLC_RESET) - 1);
+		if (nlptr)
+			vprint_cout(cons, "\n", 1);
```
mogasergiu, line 432: nit, I would add a newline here to be consistent with line 566.

mogasergiu (commented):
LGTM just one tiny nit and I think the commit message could use some rewording as this seems to be a more generic fix that applies to all platforms/drivers. It just so happens to have more of an impact on vpl011. Those stray NULs go out on every console: vgacons actually renders them as character cells, and they're there on plain QEMU/KVM serial too.

lib/ukprint/console.c
kang-taekyung: I added a newline. Thanks.

kang-taekyung:
> LGTM just one tiny nit and I think the commit message could use some rewording as this seems to be a more generic fix that applies to all platforms/drivers. It just so happens to have more of an impact on vpl011. Those stray NULs go out on every console: vgacons actually renders them as character cells, and they're there on plain QEMU/KVM serial too.

Thanks for the review! I've reworded the commit message to reflect your comment (generic fix).

mogasergiu (dismissed):
(edited out and resubmitted to deal with a new CI bug)

mogasergiu (approved):

mogasergiu (approved):

